PR #201, opened 2024-06-03
Downgrade Electron Forge to version `7.3.1`

fluiddot:
This PR is a follow-up of https://github.com/Automattic/studio/pull/198 to address a failure on Windows when generating the builds (https://github.com/Automattic/studio/pull/135#issuecomment-2144811867).

## Proposed Changes

- Downgrade Electron Forge packages to version `7.3.1` to avoid using the version `>= 18.2.0` of the package`electron/packager`. This is mainly to avoid [this change](https://github.com/electron/packager/commit/d9655d4bab5a9328391e24c98235ef0d241f1e71) in version `7.4.0` that requires that build versions are in number format (e.g. `a.b.c.d`).

## Testing Instructions

- Observe that all CI jobs pass.
- Observe that builds work as expected in the different architectures.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #782, opened 2025-01-07
Update Electron to latest version 33.3.1

nightnei:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/9631

## Proposed Changes
With this PR we are updating Electron to the latest version to keep it up-to-date.

## Testing Instructions
1. Run Studio locally `npm start`
2. Assert that there are no (new) errors/warning <br />Screenshot 2025-01-07 at 17 16 43
3. Open Studio
4. Select `Electron` in the app's menu and then `Test Render Failure (dev only)` <br />Screenshot 2025-01-07 at 17 28 05
5. Assert that you still see the same expected error (here were changes in code of this PR)<br />Screenshot 2025-01-07 at 17 29 27
6. Smoke test Studio


wojtekn (approved):
The change looks clear, and the Studio still works fine. I've checked dev build, too, and it installs and runs correctly.

We may upgrade `electron-forge` at some point, too, but we will need to review issues with version format change we spotted last time we tried.

wojtekn (changes requested):
Actually, it seems I can't import Jetpack backup to a Studio site anymore as I'm getting the error:

```
Error occurred in handler for 'importSite': TypeError: Cannot read properties of undefined (reading 'endsWith')
    at /Users/cyphelf/Sites/studio/.webpack/main/index.js:227050:58
    at Array.some (<anonymous>)
    at BackupHandlerFactory.isTarGz (/Users/cyphelf/Sites/studio/.webpack/main/index.js:227050:34)
    at BackupHandlerFactory.create (/Users/cyphelf/Sites/studio/.webpack/main/index.js:227034:23)
    at /Users/cyphelf/Sites/studio/.webpack/main/index.js:227583:81
    at Generator.next (<anonymous>)
    at fulfilled (/Users/cyphelf/Sites/studio/.webpack/main/index.js:227546:58)
```

nightnei:
@wojtekn 

> Actually, it seems I can't import Jetpack backup to a Studio site anymore as I'm getting the error:

Thank you, fixed

> We may upgrade electron-forge at some point, too, but we will need to review issues with version format change we spotted last time we tried.

I can take a look at it. Do we have discussion somewhere to read and get familiar with background and previous attempts.

wojtekn:
> Thank you, fixed

Thanks, I will run through the whole smoke test suite.

> I can take a look at it. Do we have discussion somewhere to read and get familiar with background and previous attempts.

Sounds great! See https://github.com/Automattic/studio/pull/201 for the context and feel free to open a separate issue for that.

src/menu.ts
```diff
@@ -57,8 +57,10 @@ function getAppMenu(
 		},
 		{
 			label: __( 'Test Render Failure (dev only)' ),
-			click: ( menuItem, browserWindow ) => {
-				browserWindow?.webContents.send( 'test-render-failure' );
+			click: () => {
+				withMainWindow( ( window ) => {
+					window.webContents.send( 'test-render-failure' );
+				} );
```
fredrikekelund, line 63: Out of curiosity, which change in Electron made it so we had to make this change?

src/hooks/use-import-export.tsx
```diff
@@ -101,9 +101,16 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
 			try {
 				await stopServer( selectedSite.id );
 
+				let filePath: string;
+				if ( file instanceof File ) {
+					filePath = getIpcApi().showFilePath( file );
```
fredrikekelund: The [Electron docs](https://www.electronjs.org/docs/latest/api/web-utils) led me to believe that we can access `webUtils.getPathForFile` directly in this file, but that's clearly not the case when I test it… 

How should we understand it when the Electron docs say `Process: Renderer`..? cc @Automattic/yolo

src/hooks/use-import-export.tsx
```diff
@@ -64,7 +64,7 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
 
 	const importFile = useCallback(
 		async (
-			file: BackupArchiveInfo,
+			file: File | BackupArchiveInfo,
```
fredrikekelund, line 67: Why did this work previously? 🤔 Because `File` and `BackupArchiveInfo` both have a `path` prop?

fredrikekelund (commented):
I don't have any decisive feedback on this yet. I'll take another look tomorrow, but @wojtekn might beat me to approving it.

Should we hold off with merging this until the 1.3.2 release to have some more time to test it..?

src/hooks/use-import-export.tsx
nightnei: Long story short - If it walks like a duck and quacks like a duck, then it probably is a duck 😂

Will try to explain step by step, maybe it will be easier to understand:
1. `BackupArchiveInfo` is `name` and `path` properties
2. Inside of `importFile` we also expect only `name` and `path`
3. So TS doesn't care which extra properties of the object we provide and what the object actualy is. Since anyway we use only these two properties.

src/hooks/use-import-export.tsx
nightnei: I think it's mistake in documentation. After a few days of working with Electron documentation and Issues I noticed that it's not rare case of mistakes in docs.
It should not be in Renderer for sure, since the intention of webUtils.getPathForFile is to move retrieving path logic from Renderer to Preload layer.
Screenshot 2025-01-09 at 16 19 48

src/menu.ts
nightnei: They replaced [BrowserWindow](https://www.electronjs.org/docs/latest/api/browser-window) with [BaseWindow](https://www.electronjs.org/docs/latest/api/base-window), so in this argument we don't have access to webContents anymore.

nightnei: @fredrikekelund https://github.com/electron/electron/pull/44953/files

wojtekn:
> I don't have any decisive feedback on this yet. I'll take another look tomorrow, but @wojtekn might beat me to approving it.
> Should we hold off with merging this until the 1.3.2 release to have some more time to test it..?

I think we should run the whole smoke test suite on Mac and Windows before we merge.

src/hooks/use-import-export.tsx
fredrikekelund: I just mean that we clearly passed a `File` instance to `importFile` previously, too (see below). It's curious that TS didn't bark at us for doing that 🤔

https://github.com/Automattic/studio/blob/51a4aa267ac0fe16ec0dcbfcaf9d6c65f9f90f5c/src/hooks/use-add-site.ts#L61-L64

nightnei: I see what you mean.
Yeah, as I mentioned before, it's about "duck typing".
File is just set of properties, so as soon as we provide all expected properties - it's enough for TS to be happy 😄
If you curious, I made [a quick example](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBMCu0QFsBiBLANgUxgXhgG8AoGMmCFbAUQA8oAnAQwAUGQAHbBqATwC4YAcgCMY8WKEAaUuUrIa9Zm07c+AJk2bBQrXumyyYJgsEAibLRMccAOij0zM8hXQAvbIJEAWZ+T5c5lCWUAD0NkzoYE6GMJhM0ACyIAAm6ABm6NgpggAiTMG2YCAA7gAUAJR+ZCXYAEYA1uhQAErY8VDoAG7YLAUAFuYxLkwMzLwAQvDp6dyCCbxgwDCV+AB8MGDYJTAAgmNMk9OzDGUADFWxEJjowJ4rFeub2zATmCB1ldUUjNgmgqs8Bstjs2kwUkw6jgAMq-ExfGCxYL0eYQRbLQEbMwAFX66AgMHxMCg-VwoDAwQpw3IdV4wQgqPRDyeJBcLnJ0Dg4EpsAITBKkVgOLxBKJJLJ3OwVNsEBszTKZjMFVsyCYHDKZWA-VGjyBcG1DFsWtGAGFUthdlBzhUKgBuWIuBjYKDwBhgZ47ACqUSgAA59uNNZKKXbYgBfGRh+3EDmwTI4fAemAYHBlADaZnJPLMAF0pDALFZkDZsPZHPnCMTeIEC8iwhEomYYGHQzHwJz0kt1ImyqMAOaCFPYXUbQhR4jETvAdRlePD+1TmcIJBoLDzoA) for you to play with it.

fredrikekelund: Gotcha 👍 Thanks for the clarification!

src/ipc-types.d.ts
```diff
@@ -65,6 +65,14 @@ type IpcApi = {
 	[ K in keyof typeof import('./ipc-handlers') ]: (
 		...args: Tail< Parameters< ( typeof import('./ipc-handlers') )[ K ] > >
 	) => ReturnType< ( typeof import('./ipc-handlers') )[ K ] >;
+} & {
+	/**
+	 * webUtils.getPathForFile is available only inside preload script, that's why this one function is exception and need to be defined here manually.
+	 *
+	 * The nonstandard `path` property of the Web `File` object was added in an early version of Electron as a convenience method for working with native files when doing everything in the renderer was more common.
+	 * However, it represents a deviation from the standard and poses a minor security risk as well, so beginning in Electron 32.0 it has been removed in favor of the [`webUtils.getPathForFile`](api/web-utils.md#webutilsgetpathforfilefile) method.
```
```suggestion
	 * `webUtils.getPathForFile` is available only inside preload script, that's why this one
	 * function is exception and need to be defined here manually.
	 *
	 * See https://www.electronjs.org/docs/latest/breaking-changes#planned-breaking-api-changes-320
	 * for more details.
```
Nit, but I would shorten this and refer directly to the Electron docs.

src/hooks/use-import-export.tsx
```diff
@@ -101,9 +101,16 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
 			try {
 				await stopServer( selectedSite.id );
 
+				let filePath: string;
+				if ( file instanceof File ) {
+					filePath = getIpcApi().showFilePath( file );
+				} else {
```
fredrikekelund: Maybe we should move this logic to `src/hooks/use-add-site.ts` instead? Seems reasonable that this function would consume a single format, and `src/hooks/use-add-site.ts` already calls the IPC API in some other places, so it wouldn't be exotic to also introduce a `getIpcApi().showFilePath` call

src/preload.ts
```diff
@@ -101,6 +101,7 @@ const api: IpcApi = {
 		ipcRenderer.invoke( 'getConnectedWpcomSites', localSiteId ),
 	addSyncOperation: ( id: string ) => ipcRenderer.invoke( 'addSyncOperation', id ),
 	clearSyncOperation: ( id: string ) => ipcRenderer.invoke( 'clearSyncOperation', id ),
+	showFilePath: webUtils.getPathForFile,
```
fredrikekelund:
```suggestion
	getPathForFile: webUtils.getPathForFile,
```
I don't think `showFilePath` is an ideal name. Something with "get" would be better, since the function is responsible for retrieving a string

fredrikekelund (approved):
I didn't encounter any issues while testing the following stuff:
- creating a new blank site
- creating a new site from a Studio export
- creating a new site from a Playground export
- importing a Studio export to an existing site
- importing a database into an existing site
- creating a demo site
- pulling a site
- pushing a site
- chatting with Assistant
- changing the PHP version
- deleting a site

I left a few remarks on the code, but still approving the PR based on my smoke testing on Mac. 

Per @wojtekn's latest comment, we should also run a similar test suite on Windows before merging.

src/hooks/use-import-export.tsx
nightnei: It's really reasonable idea, but it's used more than in `src/hooks/use-add-site.ts`. There are 6 calls of this function, and that's why I decided to keep it flexible with expecting both`File | { path: ... }`. Otherwise, it would be just copy-paste. WDYT about it?
Screenshot 2025-01-13 at 22 04 47

src/preload.ts
nightnei: I was thinking about it too, but decided to take this name, since it was somewhere in documentation, don't remember where exactly I saw it. It's hard to provide some arguments or explain my thoughts, but I will try 😅
This name is provided by creators of Electron, so maybe it will be kinda-ish `following style guides for naming`. I just want to believe, that they were thinking about something, when gave this name to "expose it to main world".
Totally not strong arguments, it's more about my feelings 😅 So I am ok to go with either name, just decided to explain my decision 🙂

Also, if we decided to use something else instead of `showFilePath ` - I think it's better to chose whatever, but not `getPathForFile `, to simplify `Command + F` across our codebase, avoid confusion and simplify reading code at a glance.

src/hooks/use-import-export.tsx
fredrikekelund: I think `src/hooks/use-add-site.ts` is the only place where we pass a `File` instance, though.

src/preload.ts
fredrikekelund: > if we decided to use something else instead of showFilePath  - I think it's better to chose whatever, but not getPathForFile , to simplify Command + F across our codebase, avoid confusion and simplify reading code at a glance

I actually think that reusing the `getPathForFile` name makes sense since we are simply exposing that API without any modifications.

wojtekn (approved):

src/hooks/use-import-export.tsx
nightnei: I double-checked - it's 2 files, not 3. Also `content-tab-import-export`

src/hooks/use-import-export.tsx
```diff
@@ -101,9 +101,16 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
 			try {
 				await stopServer( selectedSite.id );
 
+				let filePath: string;
+				if ( file instanceof File ) {
+					filePath = getIpcApi().getPathForFile( file );
+				} else {
+					filePath = file.path;
+				}
```
fredrikekelund:
```suggestion
				const filePath = file instanceof File ? getIpcApi().getPathForFile( file ) : file.path;
```
We could make this slightly more compact

wojtekn:
@nightnei I ran a smoke test suite on Windows for this build and haven't found issues.

PR #790, opened 2025-01-08
Update demo sites to open link in browser instead of copying the link

sejas:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10196

## Proposed Changes

- Replace link in demo sites to open the browser instead of copy the link.
- Add tests to check the link opens the browser with the correct link

It follows the same style as Sync links but without tooltip.

https://github.com/Automattic/studio/blob/81760938fd3b6dff83fc13ab1e5921565d369353/src/components/sync-connected-sites.tsx#L225-L235

## Testing Instructions

- Run `npm start``
- Click on Share tab
- Create a new demo site
- Observe the demo site link opens the browser instead of copying it
- Tests should pass

| Before | After |
|--------|--------|
|  Screenshot 2025-01-08 at 20 42 16 | Screenshot 2025-01-08 at 20 42 08 |

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-snapshots.tsx
+++ b/src/components/content-tab-snapshots.tsx
@@ -15,9 +15,9 @@ import { useSnapshots } from '../hooks/use-snapshots';
 import { useUpdateDemoSite } from '../hooks/use-update-demo-site';
 import { cx } from '../lib/cx';
 import { getIpcApi } from '../lib/get-ipc-api';
+import { ArrowIcon } from './arrow-icon';
 import { Badge } from './badge';
 import Button from './button';
-import { CopyTextButton } from './copy-text-button';
 import offlineIcon from './offline-icon';
 import ProgressBar from './progress-bar';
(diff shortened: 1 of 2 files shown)


fredrikekelund (approved):
LGTM 👍 Why did we add the tooltip in the Sync tab again..? There's less horizontal space in the Sync tab compared to the Share tab. Is that it?

sejas:
> LGTM 👍 Why did we add the tooltip in the Sync tab again..? There's less horizontal space in the Sync tab compared to the Share tab. Is that it?

Yes! That’s the reason. We have less space in the Sync tab, and the domains are determined by the user, which means they can be very long and might be truncated. In contrast, the subdomains generated in the Share tab are a combination of 2-3 words.

wojtekn:
Thanks for the fix, it improves the UX around demo sites.

PR #794, opened 2025-01-09
Add info about missed appdmg to docs

nightnei:
## Proposed Changes
I had the error `Error: Cannot find module 'appdmg'` locally. And it turned out that I didn't have `setuptools ` installed. I decided to add error message to docs, to simplify finding the solution in our docs. I think it will be useful in the future, if somebody forgets about it.

## Testing Instructions
Visual review

### DIFF
--- a/docs/code-contributions.md
+++ b/docs/code-contributions.md
@@ -12,11 +12,14 @@ Before you can build and run the app, you need to install the following dependen
 
 Many project contributors rely upon [`nvm`](https://github.com/nvm-sh/nvm) and [Homebrew](https://brew.sh) to manage Node.js and Python installations respectively.
 
-If you manage packages with Homebrew you can do the following:
-
+If you manage packages with Homebrew you can do the following:<br />
 ```bash
 brew install python3 python-setuptools
 ```
+>[!TIP]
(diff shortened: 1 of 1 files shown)


docs/code-contributions.md
```diff
@@ -18,6 +18,9 @@ If you manage packages with Homebrew you can do the following:
 brew install python3 python-setuptools
 ```
 
+> [!TIP]
+> If you encounter the error `Error: Cannot find module 'appdmg'`, ensure that `setuptools` is installed in your environment.
```
katinthehatsite: Should we link to this package on npm: https://www.npmjs.com/package/@cowasm/python-setuptools?activeTab=dependencies to be clear which `setuptools` we are referring to?

katinthehatsite: actually, ignore my comment, I see that it is mentioned more explicitly on the top of the document 👍

katinthehatsite (approved):

docs/code-contributions.md
wojtekn:
```suggestion
> If you encounter `Error: Cannot find module 'appdmg'` error, ensure that `python-setuptools` are installed in your environment according to the previous steps.
```

I suggest to move it to the end of 'Running the App' section, too, as this error would occur at that time.

nightnei: Good point, adjusted and moved to the end of 'Running the App'

PR #795, opened 2025-01-09
Studio: Update electron-forge to the latest version 7.6.0

nightnei:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10220

## Proposed Changes
[Here](https://github.com/Automattic/studio/pull/201) we reverted electron-forge to 7.3.1, but the actual issue was connected not to electron-forge, it was about `@electron/packager`, which we [pinned](https://github.com/Automattic/studio/commit/58fce2f9fe43b393a82b5ec7e5e36fd83d832120#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R123) to `18.1.3` version. So actually, there is no sense to keep pinned both - `electron-forge` and `@electron/packager`. As a result, since `@electron/packager` is anyway has pinned version, then we are safe to continue updating `electron-forge` to new versions, as long as it works well.

With this PR I propose to update electron-forge to the latest version and create a new issue to remove overrides for ["@electron/packager": "18.1.3"](https://github.com/Automattic/studio/commit/58fce2f9fe43b393a82b5ec7e5e36fd83d832120#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R124), when the package is fixed.

In the history of commits of this PR you can see that if we unpin override for @electron/packager then it has still the old error. And actually [the open issue](https://github.com/electron/packager/issues/1714) confirms that the issue still exists.
And I tested - the latest electron-forge works well.

## Testing Instructions
1. Run Studio (`npm start`)
2. Smoke test it
3. Run `npm run make`
4. Install Studio from `./out` directory
5. Smoke test it
6. Obvious, but also take a look that all CI points of this PR are green (The main important one is `Windows Dev Build`)


forge.config.ts
```diff
@@ -55,6 +55,7 @@ const config: ForgeConfig = {
 			? []
 			: [
 					new MakerDMG(
+						// @ts-expect-error - https://github.com/electron/forge/issues/3712
```
nightnei, line 58: There is bug in the MakerDMG - `appPath` is required by mistake, and it should be optional in the next release. So we will remove this `@ts-expect-error` when it's fixed.

wojtekn: The bug is closed. Does it mean it was not released yet?

wojtekn (approved):
The suggested approach looks great, thanks for linking to the issue in Electron repository.

The app builds and starts on my local without issues.

forge.config.ts
nightnei: They made it [here](https://github.com/electron/forge/pull/3761/files) optional, but looks like it wasn't in 7.6.0 release (according to [this](https://github.com/electron/forge/issues/3712#issuecomment-2508775834) comment)

PR #796, opened 2025-01-09
MobX proof-of-concept

fredrikekelund:
## Related issues

- Related to https://github.com/Automattic/dotcom-forge/issues/10209

## Why MobX?

We recently discussed (pesfPa-1jO-p2) the possibility of introducing some tool for managing global state in Studio. There's a consensus around the fact that our current React context-based approach is making life unnecessarily difficult. 

I took a stab at researching our options and found that [MobX](https://mobx.js.org/) strikes an excellent balance between simplicity and scalability.

Compared to [Redux](https://redux.js.org/), it seems to require much less boilerplate and mental overhead. With Redux, you always need to keep immutability in mind, and it's not uncommon that every piece of state also requires _at least_ one action to update it in one way or another, which adds to the amount of boilerplate you need to write. I find our implementation in Calypso woefully difficult to navigate—especially the network request parts. That said, Redux is still a familiar option, and there are ways to reduce complexity in our implementation. At this point, it wouldn't be my first choice for Studio, though.

I also looked at more minimalistic options like [jotai](https://www.npmjs.com/package/jotai), [valtio](https://www.npmjs.com/package/valtio), or [zustand](https://www.npmjs.com/package/zustand). Zustand initially looked quite appealing with its hooks-first approach, but when I tried to mock an actual store in Studio, it quickly turned cumbersome. The way that state is defined together with `set` "actions" can make for difficult-to-read code. MobX lets us structure our stores/states in whatever way we like. That might seem daunting initially, but I'm convinced it'll be a big plus in the long run.

I haven't worked with MobX before, but I find this [ten-minute official tutorial](https://mobx.js.org/getting-started) very enlightening and inspiring. 

If you think of JS decorators like `@observable`, `@action` etc when you hear MobX, it's also important to note that in MobX v6 (the latest version), the official recommendation is now **not to use those**. This helps make the code more approachable to newcomers.

### Are there any downsides?

There are two notable ones, as I see it:

1. Redux is used in both Calypso and Gutenberg, making it the established convention in A8C React apps. I don't think this means we must use it, but it's the main argument in favor of Redux.
2. There's a learning curve to MobX. I think we'll be reasonably comfortable after a day or so of familiarizing ourselves with it, though. I.e., it's not excessive, but it's worth considering.

## What's in this PR?

⚠️🚨 I don't intend for this PR to be merged. It's only meant as a proof-of-concept. ⚠️🚨

I recently worked with `ChatInputProvider` in https://github.com/Automattic/studio/pull/786. It's a very small React context that helps save and restore draft messages to Assistant when users switch between Studio sites.

This PR reimplements that file as an observable store. As you'll see, it's just a regular JS class paired with a `makeObservable` call in the constructor. To make the `ContentTabAssistant` component re-render when that state changes, we use the `observable` higher-order-component from `mobx-react-lite`.

It's all very simple and relatively easy to understand once you've groked some of the basic MobX concepts.


.eslintrc.json
```diff
@@ -38,6 +39,10 @@
 				},
 				"groups": [ "builtin", "external", "internal", "parent", "sibling", "index", "type" ]
 			}
-		]
+		],
+		"mobx/exhaustive-make-observable": "warn",
+		"mobx/unconditional-make-observable": "error",
+		"mobx/missing-make-observable": "error",
+		"mobx/missing-observer": "off"
```
fredrikekelund, line 46:
```suggestion
		"mobx/missing-observer": "warn"
```
In a real-life scenario, this rule should be enabled. I kept it off for this PR since it adds a lot of noise (it prompts us to wrap _every single component_ in an `observer` call)

nightnei: As for me, if we go with MobX, I would advocate avoiding any custom rule unless we have strong arguments to override the default behavior.

src/hooks/use-chat-input.tsx
```diff
@@ -1,46 +1,23 @@
...
-} );
-
-export const ChatInputProvider: React.FC< ChatInputProviderProps > = ( { children } ) => {
-	const inputBySite = useRef< Record< string, string > >( {} );
-
-	const getChatInput = useCallback( ( siteId: string ) => {
-		return inputBySite.current[ siteId ] ?? '';
-	}, [] );
+class ChatInputStore {
+	inputBySite: Record< string, string > = {};
```
nightnei, line 4: Thoughts out loud - we could make agreement to make such properties private and always create methods to get/set value. But I am not sure that it's good idea, necessary to think more :)

nightnei:
I am totally onboard with every word in the description.
Regarding code - store looks simple, clear and adorable 😍
Regarding Redux - the fact that it's used in Calypso is a really big argument to avoid mix of tools inside A8C, but I also find MobX much more appealing. On the next weekly we will discuss it to know opinions of other our teammates.

And thanks for your effort in researching and this PR 👏

src/hooks/use-chat-input.tsx
fredrikekelund: > we could make agreement to make such properties private and always create methods to get/set value

We could. I don't think it'd be my first choice, though, since TS already helps us enforce types effectively, and there's no need to use getters and setters to make MobX work. We can discuss this further, though.

nightnei: I have only one argument/though about it - if we have some extra method like `getChatInput` then it will be easier to find all places where it's used.
But on the other hand it's boilerplate and maybe even overengineering.

src/components/content-tab-assistant.tsx
```diff
@@ -347,7 +348,7 @@ const UnauthenticatedView = ( { onAuthenticate }: { onAuthenticate: () => void }
 	</ChatMessage>
 );
 
-export function ContentTabAssistant( { selectedSite }: ContentTabAssistantProps ) {
+export const ContentTabAssistant = observer( ( { selectedSite }: ContentTabAssistantProps ) => {
```
sejas, line 351: Something I don't like about MobX is that we need to wrap components with the `observer` function.
I think we may need to still use `mobx/missing-observer` ESLint rule in production.

sejas (commented):
@fredrikekelund , thanks for creating this proof of concept! I think it's an improvement to centralize all the state and the read/writes of AppData file.

About [MobX](https://www.npmjs.com/package/mobx), it seems a valid option. It would be interesting to compare it with other libraries like [Zustand](https://www.npmjs.com/package/zustand) when we migrate more "states".

src/components/content-tab-assistant.tsx
fredrikekelund: > Something I don't like about MobX is that we need to wrap components with the observer function

I understand your perspective, but I feel this might be more of a superficial concern. It doesn't make the code harder to read or reason about—it just looks ugly. If we adopt MobX, I'm sure we'll forget about this pretty soon.

> I think we may need to still use mobx/missing-observer ESLint rule in production.

I agree we should 👍 It will wrap **every** component in an `observer` call, though, so I would also be OK with not enabling it from day one.

wojtekn:
Thanks for working on PoC @fredrikekelund !

PR #804, opened 2025-01-14
Assistant: Fix scroll glitch on initial render

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/9281

## Proposed Changes

As described in https://github.com/Automattic/dotcom-forge/issues/9281, there are currently some glitches in how the viewport scrolls when the Assistant tab is opened. This PR fixes those by changing `endOfMessagesRef` into a ref on the topmost wrapper in the Assistant tab and by computing whether `AIClearHistoryReminder` should be displayed instantly instead of doing it in a `useEffect` callback.

## Testing Instructions

1. Have two or more Studio sites
2. Have an Assistant conversation where the last message is older than 2 hours
3. Switch back and forth between the Assistant and Settings tabs and ensure that the viewport scroll starts at the very bottom every time the Assistant tab is opened
4. Switch between two sites and ensure that the scroll behaves well in this scenario, too

## Pre-merge Checklist

### DIFF
--- a/src/components/ai-clear-history-reminder.tsx
+++ b/src/components/ai-clear-history-reminder.tsx
@@ -9,7 +9,7 @@ import Button from './button';
 function shouldShowReminder( lastMessage?: MessageType ) {
 	return (
 		lastMessage?.role === 'assistant' &&
-		Date.now() - lastMessage?.createdAt > CLEAR_HISTORY_REMINDER_TIME
+		Date.now() - ( lastMessage?.createdAt ?? 0 ) > CLEAR_HISTORY_REMINDER_TIME
 	);
 }
 
--- a/src/components/ai-clear-history-reminder.tsx
+++ b/src/components/ai-clear-history-reminder.tsx
@@ -6,6 +6,13 @@ import { Message as MessageType } from '../hooks/use-assistant';
 import { getIpcApi } from '../lib/get-ipc-api';
 import Button from './button';
 
+function shouldShowReminder( lastMessage?: MessageType ) {
+	return (
+		lastMessage?.role === 'assistant' &&
+		Date.now() - lastMessage?.createdAt > CLEAR_HISTORY_REMINDER_TIME
+	);
+}
+
 function AIClearHistoryReminder( {
 	lastMessage,
 	clearConversation,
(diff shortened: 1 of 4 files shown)


src/components/ai-clear-history-reminder.tsx
```diff
@@ -6,43 +6,36 @@ import { Message as MessageType } from '../hooks/use-assistant';
...
-			}, timeToRemind );
-		} else {
-			setShowReminder( true );
+			}, Date.now() - lastMessage.createdAt );
 		}
 
 		return () => {
-			clearTimeout( timeoutId.current );
+			clearTimeout( timeoutId );
 		};
```
fredrikekelund, line 39: There's no need to store `timeoutId` outside the hook callback, so a big part of this change is just simplifying the existing logic and moving `timeoutId` inside this closure.

wojtekn:
@fredrikekelund Assistant still lives its own life and scrolls up and down when I switch tabs. Is it caused by double rendering in dev mode? Should we test PRs related to the scrolling using proper build instead of running the app on local?

fredrikekelund:
@wojtekn, oh yeah, `StrictMode` could definitely be messing things up. Try applying the following diff and see if that fixes the problem: 

```diff
diff --git a/src/renderer.ts b/src/renderer.ts
index c496ade..32a275c 100644
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -128,6 +128,6 @@ getIpcApi()
                const rootEl = document.getElementById( 'root' );
                if ( rootEl ) {
                        const root = createRoot( rootEl );
-                       root.render( createElement( StrictMode, null, createElement( Root ) ) );
+                       root.render( createElement( Root ) );
                }
        } );

```

wojtekn:
@fredrikekelund applying the patch improves it, so it doesn't scroll to the top of the message. However, the glitch is still there:

https://github.com/user-attachments/assets/b63f28a7-3f83-43cf-b241-6a08d9874884

fredrikekelund:
> applying the patch improves it

Great 👍

> However, the glitch is still there

The movement that happens when you open the Assistant tab in your screencast is actually an animation—not the viewport scrolling 🙂

wojtekn:
> The movement that happens when you open the Assistant tab in your screencast is actually an animation—not the viewport scrolling 🙂

Not sure. The scrollbar appears, and the content moves until the view becomes static. 🤔

fredrikekelund:
I applied the following diff to confirm my theory. You can do the same to check, @wojtekn.

```diff
diff --git a/src/components/content-tab-assistant.tsx b/src/components/content-tab-assistant.tsx
index 74da799..d5cec49 100644
--- a/src/components/content-tab-assistant.tsx
+++ b/src/components/content-tab-assistant.tsx
@@ -240,24 +240,11 @@ const AuthenticatedView = memo(
                                                >
                                                        <AnimatePresence mode="wait">
                                                                { showThinking ? (
-                                                                       <motion.div
-                                                                               key="thinking"
-                                                                               initial="initial"
-                                                                               animate="animate"
-                                                                               exit="exit"
-                                                                               variants={ thinkingAnimation }
-                                                                               transition={ { duration: 0.3 } }
-                                                                       >
+                                                                       <div key="thinking">
                                                                                <MessageThinking />
-                                                                       </motion.div>
+                                                                       </div>
                                                                ) : (
-                                                                       <motion.div
-                                                                               key="content"
-                                                                               variants={ messageAnimation }
-                                                                               transition={ { duration: 0.3 } }
-                                                                               initial="initial"
-                                                                               animate="animate"
-                                                                       >
+                                                                       <div key="content">
                                                                                <MarkDownWithCode
                                                                                        message={ message }
                                                                                        siteId={ siteId }
@@ -265,7 +252,7 @@ const AuthenticatedView = memo(
                                                                                        content={ message.content }
                                                                                />
                                                                                { children }
-                                                                       </motion.div>
+                                                                       </div>
                                                                ) }
                                                        </AnimatePresence>
                                                </ChatMessage>
```

wojtekn:
Thanks for checking it more @fredrikekelund . It looks much better after removing those animations.

It seems we've added them under https://github.com/Automattic/dotcom-forge/issues/7679. Let's remove those if we can't fix them.

While testing that, I noticed small remaining issue and added https://github.com/Automattic/dotcom-forge/issues/10254 for that.

CC @sejas

fredrikekelund:
I created https://github.com/Automattic/dotcom-forge/issues/10256 for removing the animation on the most recent message when the tab is opened. I think it's a nice touch when you're chatting with the Assistant and a new message appears, but I agree it makes less sense on initial render.

wojtekn (approved):

PR #805, opened 2025-01-14
Bump demo site's size limit to 2GB from 250MB

ashfame:
Bump demo site's max size limit to 2GB from 250MB

Note: Requires 170040-ghe-Automattic/wpcom deployed first

## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/9190

## Proposed Changes

- Bump max size limit both in studio app and JN endpoint

## Testing Instructions

- Create a new site under studio app
- Create a dummy.zip file under `uploads` dir to inflate the backup size `dd if=/dev/urandom bs=1M count=1800 | zip -0 dummy.zip -`
- Create a demo site, which should succeed. Previously it failed with `The site exceeds the maximum size of 250MB.`

## Pre-merge Checklist


src/constants.ts
```diff
@@ -6,8 +6,8 @@ export const SCREENSHOT_WIDTH = 1040;
 export const SCREENSHOT_HEIGHT = 1248;
 export const LIMIT_OF_ZIP_SITES_PER_USER = 5;
 export const LIMIT_OF_PROMPTS_PER_USER = 200;
-export const DEMO_SITE_SIZE_LIMIT_MB = 250;
-export const DEMO_SITE_SIZE_LIMIT_BYTES = DEMO_SITE_SIZE_LIMIT_MB * 1024 * 1024; // 250MB
+export const DEMO_SITE_SIZE_LIMIT_GB = 2;
+export const DEMO_SITE_SIZE_LIMIT_BYTES = DEMO_SITE_SIZE_LIMIT_GB * 1024 * 1024 * 1024; // 2GB
 export const SYNC_PUSH_SIZE_LIMIT_GB = 2;
 export const SYNC_PUSH_SIZE_LIMIT_BYTES = SYNC_PUSH_SIZE_LIMIT_GB * 1024 * 1024 * 1024; // 2GB
```
sejas, line 12: In the past, we separated the limits for Demo sites and Sync Push. Now that we’ve decided both should share the same limit, I considered if it would make sense to consolidate them into a single variable again. However, keeping two distinct variables seems cleaner and provides greater flexibility for future adjustments.

sejas (approved):
Nice work! The changes are clear and the feature works as expected.

I've tested it along with 170040-ghe-Automattic/wpcom and I was able to create a Demo site of 1.5GB 🥳. It took 17 mins to upload it.

Screenshot 2025-01-15 at 13 29 10

I also tested trying to upload a bigger site and I got this alert:

alert

So we are getting a generic alert that we catch from archiveSite IPC call.
The messages updated in this PR are displayed when we got the CORS response from Nginx.

wojtekn:
> So we are getting a generic alert that we catch from archiveSite IPC call.
> The messages updated in this PR are displayed when we got the CORS response from Nginx.

Should we explicitly check for the size and not rely on the Node limit, which is accidentally equal to the current limit we want to use for demo sites and sync? In other case, the validation would stop working as soon as we fix https://github.com/Automattic/dotcom-forge/issues/8689.

src/constants.ts
ashfame: Agree! Consolidating only makes sense if it's guaranteed to not diverge in future.

ashfame:
@wojtekn Not sure I fully understand what's being discussed, but in https://github.com/Automattic/dotcom-forge/issues/9174 I proposed showing a user friendly message as opposed to the error itself & rely on our defined constants to error out as opposed to remote api call rejection/failure

sejas:
> Should we explicitly check for the size and not rely on the Node limit, which is accidentally equal to the current limit we want to use for demo sites and sync? In other case, the validation would stop working as soon as we fix https://github.com/Automattic/dotcom-forge/issues/8689.

@wojtekn , Agreed. We are already checking the file size, but what happens is that `const archiveContent = fs.readFileSync( archivePath );` throws an error before returning the file size.

Let's handle the custom alert and checking the file size on https://github.com/Automattic/dotcom-forge/issues/9174 so we can merge this PR.

ashfame:
Deployed on wpcom, so merging this PR.

PR #807, opened 2025-01-15
Add `.cursorrules` to `.gitignore`

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10255

## Proposed Changes

<del>Inspired by [this YouTube video](https://youtu.be/aG-utUqVrb0?si=NR5peM41oS7R3x4K), I created a basic `.cursorrules` file for the Studio repo. I can't say exactly how important it's been to improve my answers, but I think it's easy to see how giving the AI agent this basic context of the project would help produce better answers. </del>

<del>By adding it to the repo, we can collectively iterate on it, too.</del>

<ins>Per the discussion in this PR, we'll add `.cursorrules` to `.gitignore` instead, allowing team members to maintain individual versions</ins>

## Testing Instructions

To verify that Cursor respects these rules, you can add a statement like this:

> Always say "I love Studio" at the beginning of your output.

Then, initiate a new Composer session and verify that the AI agent indeed prints "I love Studio" at the beginning of the output.

## Pre-merge Checklist


wojtekn:
I'm not convinced we should add helper or config files for each tool or IDE developers could use, and I would prefer to keep only files required for running the app locally, for running builds or tests. 🤷

.cursorrules
```diff
@@ -0,0 +1,53 @@
+# WordPress.com Studio
+
+Every time you choose to apply a rule(s), explicitly state the rule(s) in the output. You can abbreviate the rule description to a single word or phrase.
```
sejas: What is the purpose of stating he rule(s) output?
I've seen in BigSky they set the personality, something like:
You are a great Software Engineer, creating maintainable code ....

sejas (approved):
It looks good to me!
I’m in favor of adding it to the repository and using it. We can continue refining these cursor rules as needed.

Another option could be adding .cursorrules to the .gitignore file, allowing each of us to use our own version.

wojtekn:
> Another option could be adding .cursorrules to the .gitignore file, allowing each of us to use our own version.

This looks like better approach, so we wouldn't store personal configs in the Studio repo.

fredrikekelund:
I'm OK with not committing this to the repo, even if I think our overall productivity would benefit from regularly sharing `.cursorrules` improvements. One way to resolve that would be to regularly share the changes we make to this file on Slack.

I've repurposed this PR to add `.cursorrules` to `.gitignore` instead.

bcotrim:
> I'm OK with not committing this to the repo, even if I think our overall productivity would benefit from regularly sharing `.cursorrules` improvements. One way to resolve that would be to regularly share the changes we make to this file on Slack.
> 
> I've repurposed this PR to add `.cursorrules` to `.gitignore` instead.

I also believe there would be a great benefit in sharing configurations like this. We have a good part of the team using Cursor and already have editor specific configurations (for VS Code), so this would seem to make sense.
For everyone that uses Cursor, we can collaborate, create and improve on a set of rules that lead to better results. For everyone that doesn't it has no impact.

fredrikekelund:
Let's proceed like this for now and we can keep the discussion going elsewhere 👍

PR #809, opened 2025-01-16
Mount symlinks when executing WP-cli commands

### LINKED ISSUE #20: Plugins and themes can't be symlinked
### What

I would love to be able to symlink plugins and themes. 

### Why

As a plugin developer, I often spin up new sites to reproduce specific scenarios, discover edge cases, etc and I would like to be able to symlink plugin(s) into my studio managed sites so that I can use a canonical version of the plugin(s) on my machine across different studio sites.

### How

I would like to be able to :

- navigate to the plugins directory of my studio site:
`cd ~/Studio/site-name/wp-content/plugins`

- symlink a plugin I have on my machine into the site: 
`ln -s ~/path/to/my/plugin`

- Activate the symlinked plugin in my studio site

----

Currently, I can symlink a plugin into the plugins directory and it shows properly symlinked on my machine, but the WP admin does not show the symlinked plugin as an available plugin to activate.

bgrgicak:
## Related issues

- Fixes #20 

## Proposed Changes

This PR mounts symlinked project folders to the Playground filesystem when executing WP-cli commands in Studio.

Before this PR WP-cli executions would mount the project into the Playground filesystem without mounting symlinked folders.
The lack of symlinked folders would lead to issues like a symlinked theme or plugin getting deactivated during a CLI run.

When WP-cli starts it loads WordPress, which loads the current theme and activated plugins, while loading it performs a check to see if the theme and plugins are available. In the case of symlinked themes and plugins they weren't available in the filesystem during a CLI run, so WordPress deactivated them. 

## Testing Instructions

- symlink a local theme to a site running in Studio `ln -s THEME_PATH SITE_PATH`
- start the site and activate the theme
- reload Studio with `cmd+r` 
- stop and start the site
- click on a different site and click back on the site with a symlinked theme
- confirm that the symlinked theme is still active

## Pre-merge Checklist

### DIFF
--- a/src/lib/symlink-manager.ts
+++ b/src/lib/symlink-manager.ts
@@ -34,18 +34,20 @@ export class SymlinkManager {
 	readonly projectPath: string;
 	private watcher?: AsyncIterable< FileChangeInfo< string > > | null;
 	private watcherAbortController?: AbortController | null;
+	readonly documentRoot: string;
 
 	/**
 	 * Create a new symlink manager.
 	 * @param php
 	 * @param projectPath
-	 *
+	 * @param documentRoot
 	 */
-	constructor( php: PHP, projectPath: string ) {
+	constructor( php: PHP, projectPath: string, documentRoot: string ) {
 		this.symlinks = new Map< string, string >();
 		this.mountedTargets = new Map< string, MountedTarget >();
 		this.php = php;
 		this.projectPath = projectPath;
+		this.documentRoot = documentRoot;
 	}
 
 	/**
@@ -150,11 +152,7 @@ export class SymlinkManager {
 			return;
 		}
 
-		if ( ! this.php.requestHandler ) {
-			throw new Error( 'Request handler is not set' );
-		}
-
-		const vfsPath = path.posix.join( this.php.requestHandler.documentRoot, filename );
+		const vfsPath = path.posix.join( this.documentRoot, filename );
 
 		// Double check to ensure the symlink exists
 		if ( ! this.php.fileExists( vfsPath ) ) {
--- a/src/lib/wpcli-versions.ts
+++ b/src/lib/wpcli-versions.ts
@@ -54,7 +54,12 @@ async function getLatestWPCliVersion() {
 }
 
 export const getWPCliVersionFromInstallation = async () => {
-	const { stdout } = await executeWPCli( '.', [ '--version' ] );
+	/**
+	 * This command is used to check the version of WP-CLI
+	 * which doesn't require mounting the project directory
+	 * so we are not passing the projectPath.
+	 */
+	const { stdout } = await executeWPCli( undefined, [ '--version' ] );
 	if ( stdout?.startsWith( 'WP-CLI ' ) ) {
 		const version = stdout.split( ' ' )[ 1 ];
 		if ( ! version ) {
--- a/src/lib/wpcli-versions.ts
+++ b/src/lib/wpcli-versions.ts
@@ -2,7 +2,7 @@ import fs from 'fs-extra';
 import semver from 'semver';
 import { downloadWpCli } from '../../vendor/wp-now/src/download';
 import { executeWPCli } from '../../vendor/wp-now/src/execute-wp-cli';
-import getWpCliPath from '../../vendor/wp-now/src/get-wp-cli-path';
+import getWpCliPath, { getWpCliFolderPath } from '../../vendor/wp-now/src/get-wp-cli-path';
 
 export async function updateLatestWPCliVersion() {
 	let shouldOverwrite = false;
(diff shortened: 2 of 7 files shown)


src/lib/symlink-manager.ts
```diff
@@ -150,11 +152,7 @@ export class SymlinkManager {
 			return;
 		}
 
-		if ( ! this.php.requestHandler ) {
-			throw new Error( 'Request handler is not set' );
-		}
-
-		const vfsPath = path.posix.join( this.php.requestHandler.documentRoot, filename );
+		const vfsPath = path.posix.join( this.documentRoot, filename );
```
bgrgicak, line 155: WP-cli doesn't use request handlers, so we need to enable the Symlink manager to work without assuming that a request handler will be available.

src/lib/wpcli-versions.ts
```diff
@@ -54,7 +54,12 @@ async function getLatestWPCliVersion() {
 }
 
 export const getWPCliVersionFromInstallation = async () => {
-	const { stdout } = await executeWPCli( '.', [ '--version' ] );
+	/**
+	 * This command is used to check the version of WP-CLI
+	 * which doesn't require mounting the project directory
+	 * so we are not passing the projectPath.
+	 */
+	const { stdout } = await executeWPCli( undefined, [ '--version' ] );
```
bgrgicak: We used to mount '.' here which can take a long time depending on the directory.
For example, locally it mounts the Studio project folder and this takes 5-6s on my machine. 
In CI it takes +20s.

Because we don't need to mount any directories to perform a version check I added a way to skip mounting. 
Passing explicitly `undefined` and documenting its behavior seemed like the cleanest way of doing it.

src/lib/wpcli-versions.ts
```diff
@@ -54,7 +54,8 @@ async function getLatestWPCliVersion() {
 }
 
 export const getWPCliVersionFromInstallation = async () => {
-	const { stdout } = await executeWPCli( '.', [ '--version' ] );
+	const { stdout } = await executeWPCli( getWpCliFolderPath(), [ '--version' ] );
```
bgrgicak, line 61: Using`.` caused performance issues because `.` can be a large folder. Locally it pointed to the project folder and took 5-6s to load and in CI it took +20s.

Using the CLI folded path seemed like a reasonable default.

vendor/wp-now/src/get-wp-cli-path.ts
```diff
@@ -3,11 +3,18 @@ import getWpCliTmpPath from './get-wp-cli-tmp-path';
 import { getServerFilesPath } from '../../../src/storage/paths';
 
 /**
- * The path for wp-cli phar file within the WP Now folder.
+ * The path for wp-cli folder within the WP Now folder.
```
bgrgicak:
```suggestion
 * The path to the wp-cli folder within the WP Now folder.
```

src/lib/wpcli-versions.ts
bgrgicak, line 61:
```suggestion
	/**
	 * Version checks don't require a project folder to be mounted for WP-cli
	 * to execute a version check. 
	 */
	const { stdout } = await executeWPCli( getWpCliFolderPath(), [ '--version' ] );
```

vendor/wp-now/src/execute-wp-cli.ts
```diff
@@ -6,19 +6,18 @@ import { DEFAULT_PHP_VERSION, DEFAULT_WORDPRESS_VERSION } from './constants';
...
-	MountHandler,
-	writeFiles,
-	setPhpIniEntries,
-	loadPHPRuntime,
-} from '@php-wasm/universal';
+import { PHP, MountHandler, writeFiles, setPhpIniEntries } from '@php-wasm/universal';
 import { readFileSync } from 'fs';
-
+import { startSymlinkManager } from './wp-now';
 const isWindows = process.platform === 'win32';
```
fredrikekelund:
```suggestion
import { startSymlinkManager } from './wp-now';

const isWindows = process.platform === 'win32';
```
Formatting nit. Let's add a blank line after the import statements.

vendor/wp-now/src/execute-wp-cli.ts
```diff
@@ -39,6 +38,7 @@ export async function executeWPCli(
 		options.documentRoot,
 		createNodeFsMountHandler( projectPath ) as unknown as MountHandler
 	);
+	await startSymlinkManager( php, projectPath, options.documentRoot );
```
fredrikekelund, line 42: Such a simple change 👍

src/lib/wpcli-versions.ts
fredrikekelund: Nice catch 👍

fredrikekelund (approved):
Wonderful work! Looks good and works as expected 👍

bgrgicak:
Thanks for the review @fredrikekelund!

sejas (commented):
Bero, thanks for fixing the root cause and adding some optimizations to get the wp-cli version.

I confirm now I can symlink themes and plugins and the symlinks persist ! 🥳 
I tested it with Atlas and Gutenberg.

Screenshot 2025-01-20 at 12 35 23

PR #810, opened 2025-01-16
Defer reading the archive content after our max size checks

ashfame:
- Fixes https://github.com/Automattic/dotcom-forge/issues/9174

## Proposed Changes

Previously, we were reading the file inside of `archiveSite()` which prevented us from getting to our user-friendly error-msg alert since reading a large file resulted in `RangeError [ERR_FS_FILE_TOO_LARGE]` error.

In this PR, I have changed `archiveSite()` to just stick to its responsibility of creating the archive file and not try to read it. For reading, new method on ipc-handler has been defined, which is only invoked once we have checked the archive size is within the max limits defined by us.

## Testing Instructions

- Try uploading a site more than 2GB and we get our alert instead of generic error.

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -81,14 +81,15 @@ export function useArchiveSite() {
 			try {
 				const { archivePath, archiveSizeInBytes } = await getIpcApi().archiveSite( siteId, 'zip' );
 				if ( archiveSizeInBytes > DEMO_SITE_SIZE_LIMIT_BYTES ) {
-					alert(
-						sprintf(
+					getIpcApi().showErrorMessageBox( {
+						title: __( 'Adding demo site failed' ),
+						message: sprintf(
 							__(
 								'The site exceeds the maximum size of %dGB. Please remove some files and try again.'
 							),
 							DEMO_SITE_SIZE_LIMIT_GB
-						)
-					);
+						),
+					} );
 					setUploadingSites( ( _uploadingSites ) => ( { ..._uploadingSites, [ siteId ]: false } ) );
 					getIpcApi().removeTemporalFile( archivePath );
 					return;
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -79,10 +79,7 @@ export function useArchiveSite() {
 			}
 
 			try {
-				const { archiveContent, archivePath, archiveSizeInBytes } = await getIpcApi().archiveSite(
-					siteId,
-					'zip'
-				);
+				const { archivePath, archiveSizeInBytes } = await getIpcApi().archiveSite( siteId, 'zip' );
 				if ( archiveSizeInBytes > DEMO_SITE_SIZE_LIMIT_BYTES ) {
 					alert(
 						sprintf(
@@ -97,9 +94,13 @@ export function useArchiveSite() {
 					return;
 				}
 
-				const file = new File( [ archiveContent ], 'loca-env-site-1.zip', {
-					type: 'application/zip',
-				} );
+				const file = new File(
+					[ await getIpcApi().getFileContent( archivePath ) ],
+					'loca-env-site-1.zip',
+					{
+						type: 'application/zip',
+					}
+				);
 
 				const formData = [ [ 'import', file ] ];
 				const wordpressVersion = await getIpcApi().getWpVersion( siteId );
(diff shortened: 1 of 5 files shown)


src/hooks/use-archive-site.ts
```diff
@@ -97,7 +94,7 @@ export function useArchiveSite() {
 					return;
 				}
 
-				const file = new File( [ archiveContent ], 'loca-env-site-1.zip', {
+				const file = new File( [ fs.readFileSync( archivePath ) ], 'loca-env-site-1.zip', {
```
sejas: I think we cannot use `fs` from a React hook or component. We'll need to always run `fs.readFileSync` from a NodeJS file.

ashfame: Yep, I realized the blunder when unit tests failed. It's a mystery how it even seemed to work when it was tested 🤣 This is why I removed the review request and put the PR back in draft mode. I just force pushed commits, so should be sorted now.

src/hooks/use-archive-site.ts
```diff
@@ -97,9 +94,13 @@ export function useArchiveSite() {
 					return;
 				}
 
-				const file = new File( [ archiveContent ], 'loca-env-site-1.zip', {
-					type: 'application/zip',
-				} );
+				const file = new File(
+					[ await getIpcApi().getFileContent( archivePath ) ],
+					'loca-env-site-1.zip',
```
ashfame, line 100: What's the story behind this file name here? Didn't try to fix what seems like a typo

ashfame:
Rebased again to resolve conflicts with updated `trunk`.

src/hooks/use-archive-site.ts
sejas: Yeah, it's a typo. Studio was called "Local Environment" before it had name, so that's the reason behind the archive name.
I thought we could use the site names here, but my intention was to display that we could use any random name here. Maybe we could update it to studio-site.zip

src/hooks/use-update-demo-site.tsx
```diff
@@ -35,10 +36,37 @@ export const DemoSiteUpdateProvider: React.FC< DemoSiteUpdateProviderProps > = (
...
+					);
+
+					setUpdatingSites( ( prev ) => {
+						const newSet = new Set( prev );
+						newSet.delete( localSite.id );
+						return newSet;
+					} );
+					getIpcApi().removeTemporalFile( archivePath );
+					return;
+				}
```
sejas: thanks for adding the alert also for update a demo site 👏

src/hooks/use-archive-site.ts
ashfame: I see. Let's leave renaming files in a separate PR with a broader context consideration 👍

src/hooks/use-update-demo-site.tsx
```diff
@@ -35,10 +36,37 @@ export const DemoSiteUpdateProvider: React.FC< DemoSiteUpdateProviderProps > = (
...
+
+				if ( archiveSizeInBytes > DEMO_SITE_SIZE_LIMIT_BYTES ) {
+					alert(
+						sprintf(
+							__(
+								'The site exceeds the maximum size of %dGB. Please remove some files and try again.'
+							),
+							DEMO_SITE_SIZE_LIMIT_GB
+						)
+					);
```
sejas: We could consider updating the code to use `showErrorMessageBox` or `showMessageBox` instead of `alert`.

Something like:

```suggestion
					getIpcApi().showErrorMessageBox( {
						title: __( 'Updating demo site failed' ),
						message: sprintf(
							__(
								'The site exceeds the maximum size of %dGB. Please remove some files and try again.'
							),
							DEMO_SITE_SIZE_LIMIT_GB
						),
					} );
```

Screenshot 2025-01-17 at 14 37 51

Same applies to the message displayed in use-archive-site.

sejas (approved):
Great work @ashfame.
I like the approach of deferring to read the archive contents until we submit it.
I've tested the happy path and I was able to create a new demo site and updated it successfully.

I also tested updating a demo site that exceeds the 2GB and I confirm I saw the correct alert:

Screenshot 2025-01-17 at 14 12 02

I left a suggestion to use `showErrorMessageBox` instead of `alert`.

Thanks for improving the tests! 🙌 
We could consider also adding the case where we display the alert.

src/hooks/use-update-demo-site.tsx
ashfame: Yep! I agree. I did wonder about it whether we choose what to use on the basis of context.

PR #811, opened 2025-01-16
Studio: Add tooltips to create and update buttons when size limit is reached

katinthehatsite:
## Related issues

Closes https://github.com/Automattic/dotcom-forge/issues/9191

## Proposed Changes

This PR adds a tooltip to `Update demo site` and `Add demo site` buttons when the site directory (specifically `wp-content` folder) is bigger than 2GB. The tooltip informs the user that updating the demo site or creating the demo site might take considerable amount of time and might fail:

Screenshot 2025-01-17 at 12 30 19 PM

## Testing Instructions

* Pull the changes from this site
* Run `npm start` (this is necessary as there is a new ipc-hander added)
* Create a new site in Studio
* Open the `uploads` directory in the Terminal with `cd /path/to/wp-content/uploads`
* Run `dd if=/dev/urandom bs=1M count=2560 | zip -0 dummy.zip -` to increase the size of the site and create a `dummy.zip` file
* Navigate to the `Share` tab for that site in Studio
* Hover over `Add demo site` button
* Confirm that you see the tooltip that informs you about the size limit

----
To check calculations:

```diff
+++ b/src/lib/calculate-directory-size.ts
@@ -4,6 +4,7 @@ import path from 'path';
 export function calculateDirectorySize( directoryPath: string ): Promise< number > {
        return new Promise( ( resolve, reject ) => {
                let totalSize = 0;
+               let filesProcessed = 0;
 
                async function calculateSize( dirPath: string ): Promise< void > {
                        try {
@@ -14,10 +15,12 @@ export function calculateDirectorySize( directoryPath: string ): Promise< number
                                                const filePath = path.join( dirPath, file.name );
                                                try {
                                                        if ( file.isDirectory() ) {
+                                                               console.log( `Processing directory: ${ filePath }` );
                                                                await calculateSize( filePath );
                                                        } else {
                                                                const stats = await fs.promises.stat( filePath );
                                                                totalSize += stats.size;
+                                                               filesProcessed++;
                                                        }
                                                } catch ( error ) {
                                                        console.warn( `Error processing ${ filePath }:`, error );
@@ -30,7 +33,14 @@ export function calculateDirectorySize( directoryPath: string ): Promise< number
                }
 
                calculateSize( directoryPath )
-                       .then( () => resolve( totalSize ) )
+                       .then( () => {
+                               console.log( '----------------------------------------' );
+                               console.log( `Directory size calculation complete for: ${ directoryPath }` );
+                               console.log( `Total size: ${ ( totalSize / 1024 / 1024 ).toFixed( 2 ) } MB` );
+                               console.log( `Total files processed: ${ filesProcessed }` );
+                               console.log( '----------------------------------------' );
+                               resolve( totalSize );
+                       } )
                        .catch( reject );
        } );
 }
 ```
* Apply the diff above
* Restart Studio with `npm start`
* Open the `Share` tab for a very large site
* Note down the values that are displayed in the terminal
* Open the terminal for the same site in Studio 
* Run `cd wp-content`
* Run `find . -type f | wc -l`
* Confirm the number of files corresponds to `Total files processed` 
* Run `du -sh` in the terminal
* Confirm the size to be corresponding to  `Total size` (approximately e.g 2.5 GB when you run `du -sh` and `2580.89 MB` in the terminal)

**Note**: It would be nice if we could check the performance for this on Windows build. 

## Pre-merge Checklist

### CODE REVIEW

File: src/lib/calculate-directory-size.ts [hunk 1]
```diff
@@ -0,0 +1,31 @@
+import fs from 'fs';
+import path from 'path';
+
+export function calculateDirectorySize( directoryPath: string ): Promise< number > {
+	return new Promise( ( resolve, reject ) => {
+		let totalSize = 0;
+
+		function calculateSize( dirPath: string ): Promise< void > {
+			return new Promise( ( resolveDir ) => {
+				fs.readdir( dirPath, { withFileTypes: true }, async ( err, files ) => {
```
wojtekn: If we follow that approach, we should review performance in detail, including Windows.

katinthehatsite: Yeah that's just the experiment for now, not something we might necessarily merge yet.  I am not completely convinced this is a good approach exactly because [I think it will affect performace. ](https://github.com/Automattic/dotcom-forge/issues/9191#issuecomment-2595913417)

I think displaying tooltip + error is a good suggestion [in this comment a](https://github.com/Automattic/dotcom-forge/issues/9191#issuecomment-2597764496)nd I think that's the one we should proceed with to avoid performing unnecessary complex calculations.

wojtekn: Thanks for the experiment @katinthehatsite .

I think we may still need to calculate the site size (not the archive size) for UI purposes:
- If we want to display a tooltip that shows a messaging if size is larger than X
- if we want to display the current site size somewhere in Studio UI

src/ipc-handlers.ts
```diff
@@ -1084,3 +1085,11 @@ export function addSyncOperation( event: IpcMainInvokeEvent, id: string ) {
 export function clearSyncOperation( event: IpcMainInvokeEvent, id: string ) {
 	ACTIVE_SYNC_OPERATIONS.delete( id );
 }
+
+export function getSiteSize( _event: IpcMainInvokeEvent, siteId: string ) {
+	const site = SiteServer.get( siteId );
+	if ( ! site ) {
+		throw new Error( 'Site not found.' );
+	}
+	return calculateDirectorySize( site.details.path );
```
sejas: Let's make sure we check the size of wp-content folder.
Something like:

```suggestion
	return calculateDirectorySize( nodePath.join( site.details.path, 'wp-content' ) );
```

src/lib/calculate-directory-size.ts
katinthehatsite: Actually yes, you are right. I misread the previous comment as I thought we would display the generic message without actually displaying the site size in the tooltip. We do still need the calculation in that case.

src/ipc-handlers.ts
katinthehatsite: Thanks, I made these changes in https://github.com/Automattic/studio/pull/811/commits/d3988d71ca82589670eaa83f0cdff5bcb278652a

src/components/content-tab-snapshots.tsx
```diff
@@ -178,6 +181,12 @@ function SnapshotRow( {
 		};
 	} else if ( snapshotCreationBlocked ) {
 		tooltipContent = { text: userBlockedMessage };
+	} else if ( isOverLimit ) {
+		tooltipContent = {
+			text: __(
+				'Your site exceeds 2GB in size. Updating this demo site may take considerable amount of time and could exceed the maximum allowed size for a demo site.'
```
katinthehatsite: I set 2GB here as we set 2GB as the limit just recently

wojtekn: It makes sense. If the directory size is lower than 2GB, the produced archive shouldn't exceed 2 GB in most cases. If it's greater than 2 GB, the produced archive may possibly exceed the limit.

wojtekn:
The approach looks good in general. We should fix the label placement so it is still visible when the sidebar is closed. Also, we should check if there is a way to calculate directory size that doesn't require iterating through directories.

src/components/content-tab-snapshots.tsx
```diff
@@ -187,6 +190,12 @@ function SnapshotRow( {
...
+			),
```
sejas: Could we use `DEMO_SITE_SIZE_LIMIT_GB` to display the size limit ?

Something like:

```suggestion
			text: sprintf(
				__(
					'Your site exceeds %dGB in size. Updating this demo site may take considerable amount of time and could exceed the maximum allowed size for a demo site.'
				),
				DEMO_SITE_SIZE_LIMIT_GB
			),
```

Same for `overLimitMessage` on creating a demo site.

src/hooks/use-site-size.ts
```diff
@@ -0,0 +1,21 @@
...
+			try {
+				const size = await getIpcApi().getSiteSize( siteId );
+				setIsOverLimit( size > DEMO_SITE_SIZE_LIMIT_BYTES );
+			} catch ( error ) {
+				console.error( 'Error checking site size:', error );
+			}
+		}
+		checkSiteSize();
+	}, [ siteId ] );
+
```
sejas, line 21: What do you think if we use a `useWindowListener( 'focus', async () => {` instead of a useEffect?
I wonder what happens if the Demo site tab is selected and then the user installs a bunch of plugins.

src/lib/calculate-directory-size.ts
```diff
@@ -0,0 +1,36 @@
...
+		async function calculateSize( dirPath: string ): Promise< void > {
+			try {
+				const files = await fs.promises.readdir( dirPath, { withFileTypes: true } );
+
+				await Promise.all(
+					files.map( async ( file ) => {
+						const filePath = path.join( dirPath, file.name );
+						try {
+							if ( file.isDirectory() ) {
+								await calculateSize( filePath );
```
sejas, line 17: I’ve explored different alternatives, but it appears we still need to iterate through the entire directory tree in order to calculate the wp-content size

src/ipc-handlers.ts
```diff
@@ -1084,3 +1085,11 @@ export function addSyncOperation( event: IpcMainInvokeEvent, id: string ) {
 export function clearSyncOperation( event: IpcMainInvokeEvent, id: string ) {
 	ACTIVE_SYNC_OPERATIONS.delete( id );
 }
+
+export function getSiteSize( _event: IpcMainInvokeEvent, siteId: string ) {
```
sejas: Could we be specific about getting the size of wp-content in the function name? For example getWpContentSize, getWpContentSiteSize or getWpContentDirectorySize.

```suggestion
export function getWpContentSize( _event: IpcMainInvokeEvent, siteId: string ) {
```

sejas (commented):
Looks great!

I've tested it and it worked as expected. It takes about  20 ms to calculate the size of a small site.

Screenshot 2025-01-20 at 17 59 35

As Wojtek mentioned, let's improve the tooltip when the sidebar is collapsed:
Screenshot 2025-01-20 at 18 01 29

I would like to test it on Windows before merging.

I've left a coupe of suggestions in the code.

src/components/content-tab-snapshots.tsx
katinthehatsite: Sounds good, I made changes for this in https://github.com/Automattic/studio/pull/811/commits/29d5ec228a03fbfe9bbea209221b70ce1d94470d

I also added separation between `2 GB` because some languages might use a different order for units and number.

src/lib/calculate-directory-size.ts
katinthehatsite: I did not spot another method either, so far.

src/ipc-handlers.ts
katinthehatsite: I improved the function naming in https://github.com/Automattic/studio/pull/811/commits/8babff9ba267cd5a5a74a233e53af339d062a9d4

katinthehatsite:
@wojtekn and @sejas I made some updates to the position of the tooltip and this is how it will look like with collapses sidebar:

RTL:


Screenshot 2025-01-21 at 11 12 05 AM

LTR:

Screenshot 2025-01-21 at 11 11 46 AM

One thing that I am wondering is whether we should display the tooltip at the bottom so that it does not cover the text because with the `Update demo site` button it looks like this. I find that it hides too much:

Screenshot 2025-01-21 at 2 15 27 PM

What do you think?

Finally, I noticed that we also have a tooltip on the the `Update demo site` button with the data for when the tooltip was last updated:

Screenshot 2025-01-21 at 11 22 22 AM

For now, I have disabled it for the case when the user reaches the limit and displayed the tooltip about the site limit instead. I think it makes sense to have this as an order of priority. Open for discussion though.

src/hooks/use-site-size.ts
katinthehatsite: I added useWindowListener in https://github.com/Automattic/studio/pull/811/commits/933a73d469a2059b87487467629f9798236bfefa
I also kept `useEffect` to make sure that the switches between sites update the tooltip. If I remove it, it then only updates on focus and not when switching between sites. Perhaps I am missing something - does it make sense?

sejas:
@katinthehatsite , thanks for applying the changes.

> One thing that I am wondering is whether we should display the tooltip at the bottom so that it does not cover the text because with the Update demo site button it looks like this. I find that it hides too much:

Yeah, let's display the tooltip at the bottom.

wojtekn (approved):
Works as expected.

src/hooks/use-site-size.ts
sejas: Makes sense. Thanks for testing all cases and finding out that 🙏

sejas (approved):
Looks great Kat! Thanks!
It would be great if we can display the tooltip at the bottom for the update button.

I tested it in different scenarios with the sidebar expanded and collapsed:

**Over the limit**
Screenshot 2025-01-21 at 23 09 07

**Over the limit - Sidebar Collapsed**
Screenshot 2025-01-21 at 23 09 17

**Error when trying to upload**
Screenshot 2025-01-21 at 23 10 18

**Under the limit, the update displays the last time updated**
Screenshot 2025-01-21 at 23 12 08


**Over the limit - Update**

Screenshot 2025-01-21 at 23 12 27

Screenshot 2025-01-21 at 23 12 40

katinthehatsite:
@sejas I have dropped the position for the `Update demo site` tooltip down like so:

Screenshot 2025-01-23 at 9 01 10 AM

For now, the `Add demo site` position remains as if. I am going to merge this and if you think we should also drop the `Add demo site` tooltip down, let me know and I can create a follow-up PR 👍

sejas:
It's ok to keep the add demo site position as it is . Thanks!

PR #812, opened 2025-01-16
Studio: Display tooltip with the demo site last update time

ivan-ottinger:
## Related issues

- Resolves https://github.com/Automattic/dotcom-forge/issues/10207.

## Proposed Changes

- add tooltip with the demo site last update time to the `Update demo site` button
- update placement of "offline" tooltips displayed for `Update demo site` and `Delete demo site` buttons

Markup on 2025-01-16 at 14:35:07

## Testing Instructions

2. While on an existing Studio site, head over to the Share tab and create a new demo site.
3. When you hover over the `Update demo site` button, a tooltip with information when the demo site was updated should show up.
4. If you click the button and confirm the update, the time displayed in the tooltip should reset.

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-snapshots.tsx
+++ b/src/components/content-tab-snapshots.tsx
@@ -9,6 +9,7 @@ import { useArchiveErrorMessages } from '../hooks/use-archive-error-messages';
 import { useArchiveSite } from '../hooks/use-archive-site';
 import { useAuth } from '../hooks/use-auth';
 import { useExpirationDate } from '../hooks/use-expiration-date';
+import { useFormatLocalizedTimestamps } from '../hooks/use-format-localized-timestamps';
 import { useOffline } from '../hooks/use-offline';
 import { useProgressTimer } from '../hooks/use-progress-timer';
 import { useSnapshots } from '../hooks/use-snapshots';
@@ -21,7 +22,7 @@ import Button from './button';
 import offlineIcon from './offline-icon';
 import ProgressBar from './progress-bar';
 import { ScreenshotDemoSite } from './screenshot-demo-site';
-import { Tooltip, TooltipProps } from './tooltip';
+import { Tooltip, TooltipProps, DynamicTooltip } from './tooltip';
 
 interface ContentTabSnapshotsProps {
 	selectedSite: SiteDetails;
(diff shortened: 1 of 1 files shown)


wojtekn (approved):
The code change looks correct. The tooltip works as expected.

We could consider moving the tooltip below the button as now it is displayed over the expiration text and introduces a feeling that it's not a tooltip but an expiration text being replaced. However, we will redesign the list as a part of the project so we can skip it.

Also, I'm wondering if we should use a language similar to the one in the Sync tab, e.g., `You pushed this site %s ago.` for consistency. However, the tooltip is now consistent with `Expires in %s`.

ivan-ottinger:
Thanks for the review and feedback, Wojtek!

> We could consider moving the tooltip below the button as now it is displayed over the expiration text and introduces a feeling that it's not a tooltip but an expiration text being replaced. However, we will redesign the list as a part of the project so we can skip it.

That's a good point. I have moved the tooltip to the bottom - since I expect this change will be available to users before we launch the redesigned list.

Markup on 2025-01-17 at 09:41:22

> Also, I'm wondering if we should use a language similar to the one in the Sync tab, e.g., `You pushed this site %s ago.` for consistency. However, the tooltip is now consistent with `Expires in %s`.

I was considering using similar language to the one we already use on the Sync tab, but apart it being more in line with the expiration time message as you mention, I also felt that having the tooltip in the passive form is more concise and easier to read.

src/components/content-tab-snapshots.tsx
```diff
@@ -68,6 +70,13 @@ function SnapshotRow( {
 	const deleteDemoSiteOfflineMessage = __(
 		'Deleting a demo site requires an internet connection.'
 	);
+	const getLastUpdateTimeText = () => {
+		if ( ! date ) {
+			return __( 'Never updated' );
```
sejas, line 75: I think this Never updated won’t ever be displayed.
No need to change it, it's just an observation.

sejas (commented):
Great work @ivan-ottinger , thanks for adding the tooltip.

I confirm it works as expected and the tooltip appears over the update button.
Screenshot 2025-01-17 at 11 45 53

src/components/content-tab-snapshots.tsx
ivan-ottinger: Thank you for your review and feedback, Antonio. Yes it should never display. I left it there as a fallback in case the date is not available for some reason. But now thinking about it further, maybe instead we should hide the tooltip in that case completely.

PR #814, opened 2025-01-17
Tweak how the terminal is launched on Windows

### LINKED ISSUE #359: WP CLI: environment variables not set
### Quick summary

I want to use WP CLI on Windows. When I open a shell by clicking Studio's 'Terminal' button, the environment variables needed to run WP CLI are not set correctly.

### Steps to reproduce

1. Start the Studio App on Windows
2. Start an existing WordPress project
3. Click on the 'Open in Terminal' button to open a shell
4. enter a command, e.g., `wp user list`

WP CLI should launch and execute the desired command

### What actually happened

because PATH is not set correctly, the `wp` command is not found. Also, in order to successfully execute` C:\Users\[username]\AppData\Local\studio\app-1.0.5\resources\bin\wp.bat`, the variable `STUDIO_APP_PATH` must be set.

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Platform

Windows

### Logs or notes

A possible workaround is to:
- set `STUDIO_APP_PATH` to `C:\Users\[username]\AppData\Local\studio\app-1.0.5\Studio.exe`
- set `PATH` to include `C:\Users\[username]\AppData\Local\studio\app-1.0.5\resources\bin`

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/studio/issues/359

## Proposed Changes

<del>Window distinguishes between user-level and system-level environment variables. Studio launches with the system-level `PATH` environment variable, which is then inherited by the command prompt when users click the `Open in… Terminal` button in Studio (this is just how Node's `child_process` module works).</del>

<del>This can be problematic for users because launching a new command prompt from the Start menu behaves differently. In that case, the user-level `PATH` environment variable is used. Starting a terminal with an unexpected `PATH` can obviously be troublesome since you may be missing the utilities you're used to.</del>

<del>This PR changes the logic for how the terminal is launched on Windows to fix this issue. The new logic depends on PowerShell. AFAIU, PowerShell [comes bundled](https://en.wikipedia.org/wiki/PowerShell) with all currently supported Windows versions, so this PR should hopefully work for essentially all of our users.</del>

<ins>After more testing, we concluded that we can't reliably reproduce this issue (see https://github.com/Automattic/studio/pull/814#issuecomment-2619381226). I repurposed this PR to focus on more cosmetic issues with how the terminal is launched on Windows</ins>

## Testing Instructions

1. Follow the instructions outlined in https://github.com/Automattic/studio/issues/359#issuecomment-2487958320 to install PHP and WP-CLI on Windows
2. Ensure that the user-level `PATH` envvar contains the WP-CLI path, but the system-level `PATH` does not (see [clarification](https://github.com/Automattic/studio/issues/359#issuecomment-2596106639))
3. Start Studio and have a Studio site
4. Click the `Open in… Terminal` button
5. Run `wp` to ensure that the default, user-level `PATH` envvar is loaded

## Pre-merge Checklist


src/ipc-handlers.ts
fredrikekelund, line 1: I recommend viewing this diff with the `Hide whitespace` option enabled on GitHub, since a large part of my changes here is just re-indentation.

src/ipc-handlers.ts
```diff
@@ -862,49 +862,71 @@ export async function getThumbnailData( _event: IpcMainInvokeEvent, id: string )
...
+
+	const exePath = app.getPath( 'exe' );
+	const appDirectory = app.getAppPath();
+	const appPath = ! app.isPackaged ? `${ exePath } ${ appDirectory }` : exePath;
+
+	let command: string;
+	if ( platform === 'win32' ) {
+		// Windows
+		const isPowerShellAvailable = await checkPowerShellAvailability();
+		const defaultShell = process.env.ComSpec || 'cmd.exe';
```
fredrikekelund, line 890: This mirrors Node's internal logic for locating the default Windows shell. See https://nodejs.org/api/child_process.html#default-windows-shell

wojtekn:
I tested it on my Windows machine and confirmed that with the build from this PR, `wp` works in the Terminal session opened from Studio, the same as in the Terminal session opened outside Studio. After removing the wp-cli path from my user environment variable, both stopped working.

However, everything still worked fine when I tested the same with 1.3.1-beta1. 🤔

fredrikekelund:
I can confirm that in 1.3.2, Studio picks up the user-level `PATH` environment variable, and it's inherited correctly by the terminal. I'm researching what the behavior was like in v1.0.5 (since that's when the issue was reported).

fredrikekelund:
I can't reproduce it in v1.0.5, either… I'm beginning to wonder if this was only reproducible while running the development version of Studio to begin with

wojtekn:
@fredrikekelund do you mean development build? Or the app executed via `npm start`? If the latter, I can confirm I've never started Studio on Windows for development purposes.

fredrikekelund:
I meant when the app is started with `npm start`. 

> If the latter, I can confirm I've never started Studio on Windows for development purposes.

Interesting. Thanks for sharing that, @wojtekn.

fredrikekelund:
Here's a quick **status update**. The tl;dr is that I've changed the scope of this PR to fix more cosmetic issues with how the terminal is launched on Windows.

- I can't reliably reproduce the issue that prompted this PR. AFAICT, it doesn't seem to affect built versions of Studio consistently. It's possible that the original problem was that we were modifying environment variables after Studio had launched, in which case they wouldn't get inherited by spawned shells.
- I chose to take a look at how GitHub desktop handles this problem (see [this directory](https://github.com/desktop/desktop/tree/development/app/src/lib/shells)) and ended up copying their solution for Windows.
  - Our original Windows solution is fundamentally very similar to theirs, but the benefit of theirs is that the title of the opened window is now `Command Prompt` instead of `C:\Windows\system32\cmd.exe` (which is the same as when you open the terminal from the Start menu).
  - I also modified the Linux solution to be more compact.

wojtekn:
I tested the new build:
1. Ensured that wp-cli is in my user env variables
2. restarted system
3. started Studio and confirmed `wp` works in Terminal opened from Studio

wojtekn (approved):

PR #825, opened 2025-01-21
Allow non-relative imports starting with `src/` or `vendor/`

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10292

## Proposed Changes

There's currently no way to use absolute import paths in Studio. This PR changes that by updating the Typescript, ESLint and Webpack configs to allow import paths starting with `src/` or `vendor/`. 

This enables us to use a similar approach as Calypso, where import paths can start with `calypso/`.

This was something I set up while working on https://github.com/Automattic/studio/pull/826. That PR introduces a new folder—`src/stores`—and I figured having a consistent import path for it would be nice.

## Testing Instructions

1. Restart your editor after running `npm install`
2. Pick a file and change the import paths to be absolute
3. Ensure that the editor doesn't complain and that `npm start` works as expected

## Pre-merge Checklist


wojtekn (approved):
The change looks correct, and works as expected. IDE allows for both type of paths.

Also, it shows where path could be shortened e.g.:

Screenshot 2025-01-21 at 16 54 42

Thanks for that improvement!

PR #826, opened 2025-01-21
Refactor `ChatProvider` into a Redux store

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10269

## Proposed Changes

> [!TIP]
> Feel free to ping me when reviewing this PR. I know the diff is large, and I'm happy to pair as needed.

We recently decided (see https://github.com/Automattic/dotcom-forge/issues/10209) to add [Redux Toolkit](https://redux-toolkit.js.org/) to the Studio codebase. As part of the decision process, we looked at a mock re-implementation of `ChatProvider` as a Redux store. Following up on that work, this PR implements an actual refactoring of `ChatProvider` as a Redux store.

I originally thought we could keep this PR smaller, but once I started pulling the threads, I realized that it didn't make sense to keep the `useAssistant`, `useAssistantApi`, and `useSendFeedback` hooks. So, that functionality now also lives in `src/stores/chat-slice.ts`.

Aside from better code organization and easier-to-reason-about primitives, the fact that we can [subscribe](https://github.com/Automattic/studio/pull/826#discussion_r1925553306) to state changes with Redux and persist state as needed is a big win. 

This Redux implementation should hopefully serve as a reference for future Redux implementations (fresh ones and refactorings).

## Testing Instructions

1. Before applying this PR, make sure that you have some existing Assistant conversations
2. Quit Studio before checking out the PR
3. Check out this PR
4. `npm install`
5. Start Studio
6. Ensure that your previous Assistant conversations load as expected
7. Do a full smoke test of the Assistant, including:
7.1 Switching between sites
7.2 Basic chatting
7.3 Running WP CLI commands
7.4 Sending feedback
7.5 Explicitly referencing previous messages ("what did I say in my previous message?")
7.6 Reload the main view to ensure your new messages are persisted
7.7 Log out, make sure your conversations disappear
7.8 Log back in, make sure your conversations are restored
7.9 Make sure you can see your remaining message count in the app settings

## Pre-merge Checklist


src/stores/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+listenerMiddleware.startListening( {
+	predicate( action, currentState, previousState ) {
+		return currentState.chat.messagesDict !== previousState.chat.messagesDict;
+	},
+	effect( action, listenerApi ) {
+		const state = listenerApi.getState() as RootState;
+		localStorage.setItem( CHAT_MESSAGES_STORE_KEY, JSON.stringify( state.chat.messagesDict ) );
+	},
+} );
```
fredrikekelund: I want to highlight this piece in particular because it illustrates how we could use Redux logic to more effectively update `appdata-v1.json`, too.

We are subscribing to state changes here, with the `predicate` function using a plain old `===` equality check (that works because Redux data is immutable). In the `effect` callback, we can access the entire state tree and do what we want with it 🙂

src/components/tests/content-tab-assistant.test.tsx
fredrikekelund, line 1: This isn't a completely exhaustive test suite, but it's pretty good, and the limited changes I had to make should hopefully be reassuring.

src/stores/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const store = configureStore( {
+	reducer: {
+		chat: chatReducer,
+	},
+	middleware: ( getDefaultMiddleware ) =>
+		getDefaultMiddleware().prepend( listenerMiddleware.middleware ),
+} );
+
+export default store;
```
nightnei: WDYT about avoiding default exports?
I liked such an approach in Tumblr, and IMO, there were reasonable arguments about simplifying looking for import names and consistency with imported names.

src/stores/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export type AppDispatch = typeof store.dispatch;
```
nightnei, line 50: We already started copy-pasting and importing it everywhere. WDYT about creating `useAppDispatch` and use it, instead of useDispatch? I think code will be less verbose and clear.

src/stores/index.ts
```diff
@@ -0,0 +1,41 @@
+import { configureStore, createListenerMiddleware } from '@reduxjs/toolkit';
+import { CHAT_ID_STORE_KEY, CHAT_MESSAGES_STORE_KEY } from 'src/constants';
+import chatReducer from 'src/stores/chat-slice';
+
+export type RootState = {
```
nightnei, line 6: I am thinking, what about creating `useAppSelector`, similar to what I mention here. Reaso and arguments for it are the same

https://github.com/Automattic/studio/pull/826/files#diff-a962836e7056c855240fcab71b4e4f6a56e8462e298844f0b928226d230cda83R41

src/stores/chat-slice.ts
```diff
@@ -0,0 +1,384 @@
...
+	},
+} );
+
+export const { updateFromTheme, setMessages, setChatInput, updateMessage, resetChatState } =
+	chatSlice.actions;
+
+export const { selectChatInput, selectMessages, selectChatApiId, selectIsLoading } =
+	chatSlice.selectors;
+
+export default chatSlice.reducer;
```
nightnei: I would avoid default exports, as mentioned it above

src/stores/index.ts
```diff
@@ -0,0 +1,41 @@
...
+	chat: ReturnType< typeof chatReducer >;
+};
+
+const listenerMiddleware = createListenerMiddleware< RootState >();
+
+listenerMiddleware.startListening( {
```
nightnei, line 13: Not critical, more just thinking out loud - should we unsubscribe during logout? 
I don't predict any issues with it and it's not about performance problems, since Studio is small. But maybe as an initial example, maybe would be nice to implement it.

src/stores/index.ts
```diff
@@ -0,0 +1,41 @@
...
+	predicate( action, currentState, previousState ) {
+		return currentState.chat.messagesDict !== previousState.chat.messagesDict;
+	},
+	effect( action, listenerApi ) {
+		const state = listenerApi.getState() as RootState;
+		localStorage.setItem( CHAT_MESSAGES_STORE_KEY, JSON.stringify( state.chat.messagesDict ) );
+	},
+} );
+
+listenerMiddleware.startListening( {
```
nightnei, line 27: Looking at a glance - it's dificult to understand what specific listener do. They look similar and when we have much more listeners - it will be difficult to look up specific one. Maybe let's add some small comment to every listener? WDYT?

src/stores/index.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const listenerMiddleware = createListenerMiddleware< RootState >();
+
+listenerMiddleware.startListening( {
+	predicate( action, currentState, previousState ) {
+		return currentState.chat.messagesDict !== previousState.chat.messagesDict;
+	},
+	effect( action, listenerApi ) {
+		const state = listenerApi.getState() as RootState;
```
nightnei: It's redundant, sicne you type middleware [here](https://github.com/Automattic/studio/pull/826/files#diff-a962836e7056c855240fcab71b4e4f6a56e8462e298844f0b928226d230cda83R9)
```suggestion
		const state = listenerApi.getState();
```

src/stores/index.ts
```diff
@@ -0,0 +1,41 @@
...
+		localStorage.setItem( CHAT_MESSAGES_STORE_KEY, JSON.stringify( state.chat.messagesDict ) );
+	},
+} );
+
+listenerMiddleware.startListening( {
+	predicate( action, currentState, previousState ) {
+		return currentState.chat.chatApiIdDict !== previousState.chat.chatApiIdDict;
+	},
+	effect( action, listenerApi ) {
+		const state = listenerApi.getState() as RootState;
```
nightnei:
```suggestion
		const state = listenerApi.getState();
```

src/stores/index.ts
```diff
@@ -0,0 +1,41 @@
+import { configureStore, createListenerMiddleware } from '@reduxjs/toolkit';
+import { CHAT_ID_STORE_KEY, CHAT_MESSAGES_STORE_KEY } from 'src/constants';
```
nightnei: WDYT about adding prefix for such constants as `LOCAL_STORAGE_` or `LOCAL_STORAGE_KEY_`. It would simpify reading code.
```suggestion
import { CHAT_ID_STORE_KEY, CHAT_MESSAGES_STORE_KEY } from 'src/constants';
```

src/stores/chat-slice.ts
```diff
@@ -0,0 +1,384 @@
...
+}
+
+const chatSlice = createSlice( {
+	name: 'chat',
+	initialState,
+	reducers: {
+		resetChatState: ( state ) => {
+			// Reassigning `initialState` to `state` doesn't work. Probably because of Immer.js.
+			Object.assign( state, initialState );
+		},
```
nightnei: Do you mean `state = initialState`? If so, it's because arguments don't work by reference, as in PHP with prefix `&`. So with assigning you just change the variable inside this function.

Anyay, I found more right approach https://stackoverflow.com/questions/59424523/reset-state-to-initial-with-redux-toolkit

```suggestion
		resetChatState: ( state ) => initialState,
```

src/stores/chat-slice.ts
```diff
@@ -0,0 +1,384 @@
...
+		selectChatApiId: ( state, instanceId: string ) => state.chatApiIdDict[ instanceId ],
+		selectIsLoading: ( state, instanceId: string ) => state.isLoadingDict[ instanceId ] ?? false,
+	},
+} );
+
+export const { updateFromTheme, setMessages, setChatInput, updateMessage, resetChatState } =
+	chatSlice.actions;
+
+export const { selectChatInput, selectMessages, selectChatApiId, selectIsLoading } =
+	chatSlice.selectors;
```
nightnei: WDYT about exporting `chatSelectors` and `chatActions`, or something similar, but don't spread them to keep them behind some isolated space. It will be easier to read in code `chatAction.setMessage` instead of `setMessage`, since it looks not obvious that it's action and specifically from `chat` reducer.

src/components/content-tab-assistant.tsx
```diff
@@ -513,10 +458,18 @@ export function ContentTabAssistant( { selectedSite }: ContentTabAssistantProps
 					<AIInput
 						ref={ inputRef }
 						disabled={ disabled }
-						input={ currentInput }
-						setInput={ setInput }
-						handleSend={ handleSend }
-						handleKeyDown={ handleKeyDown }
+						input={ chatInput }
+						setInput={ ( input ) => {
+							dispatch( setChatInput( { siteId: selectedSite.id, input } ) );
```
nightnei: I was thinking and:
1. I don't see benefits to store temporary value in global store. We can achieve the same with storing in inside this component's state.
2. Every keydown triggers a lot of redundant store checks, even every prodicate inside  `listenerMiddleware.startListening` is triggered.

Maybe I am missing something, but so far it looks like overkill.
WDYT?

UPDATE: I see, we store values for different chats... sad that we trigger store so many times during typing, but looks like it's optimal option. Unless, we consider saving it in localStorage, it's typical UX approach, to avoid losing data in case of some errors or reloads of window. But it's more as an improvement. So I think it's better to leave it as is.

src/stores/chat-slice.ts
```diff
@@ -0,0 +1,384 @@
...
+	},
+	extraReducers: ( builder ) => {
+		builder
+			.addCase( updateFromSiteThunk.pending, ( state, action ) => {
+				const { site } = action.meta.arg;
+
+				state.currentURL = `http://localhost:${ site.port }`;
+				state.phpVersion = site.phpVersion ?? DEFAULT_PHP_VERSION;
+				state.siteName = site.name;
+			} )
```
nightnei: I was looking at code and looks like there is no sense to create separate case for `pending`, we can keep everything in one place `fulfilled`. We have there access to `meta` too.

src/stores/chat-slice.ts
```diff
@@ -0,0 +1,384 @@
...
+	};
+}
+
+const chatSlice = createSlice( {
+	name: 'chat',
+	initialState,
+	reducers: {
+		resetChatState: ( state ) => {
+			// Reassigning `initialState` to `state` doesn't work. Probably because of Immer.js.
+			Object.assign( state, initialState );
```
nightnei: Also, I am thinking - initial state stores initial values of `messagesDict` and `chatApiIdDict`. I mean they were read and saved during initial rendering of Studio. I didn't get completely what is `chatApiIdDict` but I have assumption that we potentially can have bug here with resetting localSotage to data which were during the initial render of Studio. 
Just thoughts out loud, maybe I am wrong, since don't have the whole picture of how everything works yet.

nightnei: Hm, more thoughts - looks like you went another approach and we can remove this action:
https://github.com/Automattic/studio/pull/826/files#diff-d3d6432d7d681344c73a14e2e30339dbdd7fe784d4446baf42efaf59eed40b33R396-R398

nightnei (changes requested):
Tested, works well 👍
And teh code in general is amazing 😍
Redux-toolkit much much better than clear Redux.

nightnei (changes requested):
Tested, works well 👍
And teh code in general is amazing 😍
Redux-toolkit much much better than clear Redux.

src/components/content-tab-assistant.tsx
fredrikekelund: Storing this state in the Redux store helps maintain the bug fixes introduced in https://github.com/Automattic/studio/pull/788. Sounds like we're on the same page already, but to respond to your points:

> I don't see benefits to store temporary value in global store. We can achieve the same with storing in inside this component's state.

The component state is reset when navigating between tabs, and will soon (see https://github.com/Automattic/dotcom-forge/issues/10219) be reset when navigating between sites, too.

> Every keydown triggers a lot of redundant store checks, even every prodicate inside listenerMiddleware.startListening is triggered.

The middleware predicates are very cheap. Worth keeping in mind.

src/stores/chat-slice.ts
fredrikekelund: Even if it takes a very short time to resolve `updateFromSiteThunk`, it still technically makes sense to store data that's immediately available _before_ resolving the promise. If we moved these assignments, we'd also have to duplicate them in the `rejected` handler.

src/stores/chat-slice.ts
fredrikekelund: Yeah, `resetChatState` is really only used in unit tests (and that's also what it's intended for). I'll add a comment about that to clarify 👍

src/stores/chat-slice.ts
fredrikekelund: Returning `initialState` makes sense 👍 Less of a workaround

src/stores/index.ts
fredrikekelund: Good point, makes sense 👍

src/stores/index.ts
fredrikekelund: Makes sense 👍

src/stores/index.ts
fredrikekelund: I think this would be easier to accomplish if the auth data was also stored in Redux. Given that it's not necessary, I would defer this until later.

src/stores/index.ts
fredrikekelund: Sure, makes sense 👍

src/stores/index.ts
fredrikekelund: Good idea 👍

src/stores/index.ts
nightnei: Yeah, makes sense 👍

src/stores/chat-slice.ts
fredrikekelund: Good idea – I like it 👍 Will make even more sense once we have more stores.

fredrikekelund:
Thanks for the review, @nightnei 🙏 I've addressed all your feedback

src/stores/chat-slice.ts
nightnei: > If we moved these assignments, we'd also have to duplicate them in the rejected handler.

But we don't have rejected handler for this case.
And, do we use `currentURL ` etc w/o fulfilling of updateFromSiteThunk?

fredrikekelund: > But we don't have rejected handler for this case.

Then we'd need to add one. Duplicating the logic (or abstracting it) is not very appealing.

> And, do we use currentURL  etc w/o fulfilling of updateFromSiteThunk?

Only if a user manages to send a message before `updateFromSiteThunk` is resolved, which admittedly is very unlikely.

src/stores/chat-slice.ts
nightnei: The last one thing is I wanted gained to comeback this this specific message https://github.com/Automattic/studio/pull/826#discussion_r1932937208

It's cool that we added comment, makes sense, but theoretically, somebody can use it in code in the future and I have assumption that it can be buggy.
I don't completely understood `chatApiIdDict ` and why do we store a few chats in `messagesDict `, maybe it's the reason of my next concern, but I would like to discuss it:

For the initialState we retrieve data from localStorage during starting Studio:

```
const storedMessages = localStorage.getItem( CHAT_MESSAGES_STORE_KEY );
const storedChatIds = localStorage.getItem( CHAT_ID_STORE_KEY );
```

Then we store it inside `initialState` object. And these fields are never updated.
So, if somebody decided to use `resetChatState`, then it will errase all chat history to the initial state of Studio.

Maybe I wasn't enough clear, so I will provide the code, which I think will be selfexplanation of m concern:

```
const getInitialState = () => (

  const messagesDict = storedMessages ? JSON.parse( storedMessages ) : {};
  const chatApiIdDict = storedChatIds ? JSON.parse( storedChatIds ) : {};
  // !!! here messagesDict and chatApiIdDict should be somehow cleaned up to keep everything except the current chat. Not to the initial state which were during initial loading of Studio

  return {
	currentURL: '',
	pluginListDict: {},
    .......
	messagesDict,
	chatApiIdDict,
	.......
```

nightnei: And in general, as for me, it's very bad practice to store something like this action in code, only for test purposes.

I propose either to find some way to patch store in test to add this method or at least to add `__` prefix to this action and keep the comment of why it's here. In this case we will avoid using it in the future by mistake.

fredrikekelund: Point taken! 👍 Given that `resetChatState` is indeed meant for tests mainly, I think it's better not to include any logic that modifies `localStorage` in this file. I did change it so that we re-read the stored messages and chat API IDs every time the state resets, though.

fredrikekelund: > And in general, as for me, it's very bad practice to store something like this action in code, only for test purposes.

I get your point. Let me think on this and see if I can come up with a better approach

nightnei: I found this:

```
  // Inject a new reducer inside the test
store.replaceReducer({
    chat: (state = initialState, action) => {
      if (action.type === 'test/resetChatState') {
        return initialState; // Reset state manually
      }
      return chatReducer(state, action); // Otherwise, use original reducer
    },
  });
```

nightnei: Maybe we can use it for our case

package.json
```diff
@@ -108,6 +108,7 @@
 		"@php-wasm/node": "^0.9.44",
 		"@php-wasm/scopes": "^0.9.44",
 		"@php-wasm/universal": "^0.9.44",
+		"@reduxjs/toolkit": "^2.5.0",
```
sejas: A new [version 2.5.1](https://www.npmjs.com/package/@reduxjs/toolkit?activeTab=versions) was released 3 days ago. We could update it.
```suggestion
		"@reduxjs/toolkit": "^2.5.1",
```

src/stores/chat-slice.ts
fredrikekelund: I moved the reset state action to a test util file per your suggestion in this thread, @nightnei. Great suggestion 👍 Much cleaner this way. 

`testReducer` is also reusable now, which is an added benefit.

katinthehatsite:
So far the changes look good to me and everything persists when I switch between tabs, sites etc. I noticed one small regression compared to trunk. This is how you can reproduce it:

* Open assistant tab on a site
* Ask it to give some CLI commands 
* Run a CLI command from the code block
* Ensure that you see the outcome
* Switch between tabs e.g to the Settings tab
* Come back to the Assistant tab
* Observe that you do not see the outcome of the command that you ran while the outcome should persist:

https://github.com/user-attachments/assets/03083ebc-891c-424e-a4f4-f788456b9f8e

sejas (commented):
The Assistant behaviour looks good feature wise. The only issue is what @katinthehatsite mentioned about persisting the wp-cli output.
The rest looks good.

I took a quick look to the code, mostly to learn about react toolkit. Couldn't find any suggestion.
I've seen people creating a hook that returns the actions, like useChatActions to avoid repeating the `useAppDispatch` in every component but it's not a huge improvement.

nightnei (approved):
Looks marvelous 🙂👍

fredrikekelund:
Thanks for the reviews, everyone! 🙏 I've fixed the issues with persisting the output of WP CLI commands. 

> I've seen people creating a hook that returns the actions, like useChatActions to avoid repeating the useAppDispatch in every component 

I changed the way the thunk actions are exported from `chat-slice.ts` so they all live in a `chatThunks` object. It's not exactly the same as your proposal, but it matches `chatActions` and `chatSelectors` more closely.

nightnei:
> I changed the way the thunk actions are exported from chat-slice.ts so they all live in a chatThunks object. It's not exactly the same as your proposal, but it matches chatActions and chatSelectors more closely.

Very nice idea, looks a bit easier to work with it 👍

PR #830, opened 2025-01-22
Use platform-aware path separator in `WpressValidator`

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/9567

## Proposed Changes

`BackupHandlerWpress::listFiles` uses `path.join` when enumerating the files in a `.wpress` archive (All-in-One WP Migration format). This breaks `WpressValidator` on Windows because it assumes Unix style path separators in both `canHandle` and `parseBackupContents`.

This PR fixes the problem by replacing `'/'` with `path.sep` and by changing the unit tests to account for both Unix and Windows-style paths.

## Testing Instructions

1. Import a `.wpress` archive on Windows
2. Ensure that the import completes successfully and the site works

## Pre-merge Checklist

### DIFF
--- a/src/lib/import-export/import/validators/wpress-validator.ts
+++ b/src/lib/import-export/import/validators/wpress-validator.ts
@@ -10,7 +10,7 @@ export class WpressValidator extends EventEmitter implements Validator {
 		const optionalDirs = [ 'uploads', 'plugins', 'themes' ];
 		return (
 			requiredFiles.every( ( file ) => fileList.includes( file ) ) &&
-			fileList.some( ( file ) => optionalDirs.some( ( dir ) => file.startsWith( dir + '/' ) ) )
+			optionalDirs.some( ( dir ) => fileList.some( ( file ) => file.startsWith( dir + path.sep ) ) )
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


src/lib/import-export/import/validators/wpress-validator.ts
```diff
@@ -10,7 +10,7 @@ export class WpressValidator extends EventEmitter implements Validator {
 		const optionalDirs = [ 'uploads', 'plugins', 'themes' ];
 		return (
 			requiredFiles.every( ( file ) => fileList.includes( file ) ) &&
-			fileList.some( ( file ) => optionalDirs.some( ( dir ) => file.startsWith( dir + '/' ) ) )
+			optionalDirs.some( ( dir ) => fileList.some( ( file ) => file.startsWith( dir + path.sep ) ) )
```
fredrikekelund, line 13: The change from `'/'` to `path.sep` is the only functional change here. I flipped the order of `fileList.some` and `optionalDirs.some` because I find it easier to read this way.

wojtekn (approved):
The change looks clear. Importing .wpress fikle works fine on Windows now. Good catch!

PR #833, opened 2025-01-22
Fix: Topbar icons display padding in macOS fullscreen

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/9264

## Proposed Changes

- Added smooth transitions for window controls positioning when entering/exiting full-screen mode
- Implemented native full-screen state detection through Electron's IPC
- Added proper event handling for full-screen state changes in the main window

## Testing Instructions
1. Launch the application
2. Verify window topbar controls are properly positioned based on system language direction
3. Enter full-screen mode (⌘+Control+F or green window button)
   - Verify smooth transition of window topbar controls
   - Confirm correct padding in full-screen mode
4. Exit full-screen mode
   - Verify smooth transition back to windowed mode
   - Confirm window topbar controls return to correct position
5. Repeat tests with both LTR and RTL language settings

### Expected Behavior
- Window topbar controls should smoothly transition when entering/exiting full-screen
- Padding should adjust correctly based on window state and language direction
- No visual glitches or jumps during transitions
- Consistent behavior in both LTR and RTL

## Pre-merge Checklist

### DIFF
--- a/src/components/mac-titlebar.tsx
+++ b/src/components/mac-titlebar.tsx
@@ -1,3 +1,4 @@
+import { useFullscreen } from '../hooks/use-fullscreen';
 import { cx } from '../lib/cx';
 import { isWindowFrameRtl } from '../lib/is-window-frame-rtl';
 
@@ -8,15 +9,20 @@ export default function MacTitlebar( {
 	className?: string;
 	children?: React.ReactNode;
 } ) {
+	const isFullscreen = useFullscreen();
+	const isRtl = isWindowFrameRtl();
+
 	return (
 		<div
 			className={ cx(
-				// Leave space for window controls depending on which side they are on
-				// Take into account the "chrome" padding, the position of which depends on the language direction
-				isWindowFrameRtl() &&
+				'transition-[padding] duration-700 ease-in-out',
+				! isFullscreen &&
+					isRtl &&
 					'ltr:pr-window-controls-width-excl-chrome-mac ltr:pl-chrome rtl:pr-window-controls-width-mac rtl:-ml-chrome',
-				! isWindowFrameRtl() &&
+				! isFullscreen &&
+					! isRtl &&
 					'ltr:pl-window-controls-width-mac rtl:pl-window-controls-width-excl-chrome-mac rtl:pr-chrome',
+				isFullscreen && 'ltr:pl-4 rtl:pr-4',
 				className
 			) }
 		>
--- a/src/components/mac-titlebar.tsx
+++ b/src/components/mac-titlebar.tsx
@@ -15,7 +15,7 @@ export default function MacTitlebar( {
 	return (
 		<div
 			className={ cx(
-				'transition-[padding] duration-700 ease-in-out',
+				'transition-[padding] duration-500 ease-in-out',
 				! isFullscreen &&
 					isRtl &&
 					'ltr:pr-window-controls-width-excl-chrome-mac ltr:pl-chrome rtl:pr-window-controls-width-mac rtl:-ml-chrome',
--- a/src/hooks/tests/use-fullscreen.test.ts
+++ b/src/hooks/tests/use-fullscreen.test.ts
@@ -1,0 +1,78 @@
+import { renderHook } from '@testing-library/react';
+import { act } from 'react';
+import { getIpcApi } from '../../lib/get-ipc-api';
+import { useFullscreen } from '../use-fullscreen';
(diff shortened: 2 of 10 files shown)


katinthehatsite (commented):
Nice work! 

I like the effect it creates when the window becomes full-size 👍 

One thing that I find a bit strange is when you go from the full size window to the smaller window and for a couple  of seconds while the transition is happening, the buttons are overlapping the sidebar toggle. I could see this happening both with either buttons or key board shortcuts. I tried to capture it in the video below. Hopefully it illustrates well what I mean but let me know if I am not explaining this well. 

https://github.com/user-attachments/assets/f28a49a9-d301-4569-bc25-e4ecd4c0d80e

I am not sure if there is an easy solution for this though. Perhaps this is okay. 

Everything worked well on my end for rtl and ltr languages.

src/hooks/tests/use-fullscreen.test.ts
```diff
@@ -0,0 +1,78 @@
+import { renderHook } from '@testing-library/react';
+import { act } from 'react';
+import { getIpcApi } from '../../lib/get-ipc-api';
+import { useFullscreen } from '../use-fullscreen';
```
katinthehatsite, line 4: Nice, thanks for adding tests!

src/components/mac-titlebar.tsx
```diff
@@ -8,15 +9,20 @@ export default function MacTitlebar( {
...
+	const isFullscreen = useFullscreen();
+	const isRtl = isWindowFrameRtl();
+
 	return (
 		<div
 			className={ cx(
-				// Leave space for window controls depending on which side they are on
-				// Take into account the "chrome" padding, the position of which depends on the language direction
-				isWindowFrameRtl() &&
+				'transition-[padding] duration-700 ease-in-out',
```
fredrikekelund:
```suggestion
				'transition-[padding] duration-500 ease-in-out',
```
I would speed up this animation slightly. 500ms matches the sidebar transition

src/hooks/tests/use-fullscreen.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+	it( 'should initialize with false and update when isFullscreen resolves', async () => {
+		mockIpcApi.isFullscreen.mockResolvedValue( true );
+		const { result } = renderHook( () => useFullscreen() );
+
+		expect( result.current ).toBe( false );
+
+		await act( async () => {
+			await Promise.resolve();
+		} );
```
fredrikekelund:
```suggestion
		await waitFor( () => {
			expect( mockIpcApi.isFullscreen ).toHaveBeenCalledTimes( 1 );
		} );
```
Optional, but I found the current code a bit cryptic. I think this clarifies the intention.

src/hooks/tests/use-fullscreen.test.ts
```diff
@@ -0,0 +1,78 @@
...
+		let eventHandler: ( _: unknown, fullscreen: boolean ) => void = () => undefined;
+		( useIpcListener as jest.Mock ).mockImplementation( ( _channel, handler ) => {
+			eventHandler = handler;
+		} );
+
+		const { result } = renderHook( () => useFullscreen() );
+
+		await act( async () => {
+			await Promise.resolve();
+		} );
```
fredrikekelund: Same thing here about potentially changing.

fredrikekelund (approved):
LGTM 👍 I agree with @katinthehatsite that it would be nice if the transition ran earlier, but looking at when the `enter-full-screen` and `leave-full-screen` events are fired, that seems difficult to pull off. I also couldn't find any other viable methods for listening to earlier events about fullscreen state changes, so we may just have to accept that small nuisance.

bcotrim:
Thanks, @katinthehatsite and @fredrikekelund for the review!
I can also see that issue and agree it looks a bit strange.
The delay is indeed a limitation of the current implementation. It's caused by the IPC communication chain - when full-screen changes, it has to go through main process → IPC → renderer process → React state updates.
I tried to find alternative ways to get the full-screen state, like DOM events, but with no success.

I've addressed the other comments, if you both agree, I would propose we merge like this and accept this as a limitation for now.

src/hooks/tests/use-fullscreen.test.ts
bcotrim: Much cleaner approach, thank you!

fredrikekelund:
I say :shipit:

PR #838, opened 2025-01-23
Studio: Add connected site ID to the `create-new-site-from-zip` request

ivan-ottinger:
## Related issues

- Related to https://github.com/Automattic/dotcom-forge/issues/10243#issuecomment-2609523395.

## Proposed Changes

- if there's a WordPress.com production site connected to local Studio site, add its ID to the `create-new-site-from-zip request`

## Testing Instructions

1. Please follow the testing instructions in 170864-ghe-Automattic/wpcom.
2. Newly introduced test should pass (`npm test src/hooks/tests/use-archive-site.test.ts`).

## Pre-merge Checklist

### DIFF
--- a/src/hooks/tests/use-archive-site.test.ts
+++ b/src/hooks/tests/use-archive-site.test.ts
@@ -1,0 +1,92 @@
+import { renderHook, act } from '@testing-library/react';
+import { getIpcApi } from '../../lib/get-ipc-api';
+import { useSyncSites } from '../sync-sites';
+import { useArchiveSite } from '../use-archive-site';
+import { useAuth } from '../use-auth';
+import { useSiteDetails } from '../use-site-details';
+
+jest.mock( '../use-auth' );
+jest.mock( '../sync-sites' );
+jest.mock( '../use-site-details' );
+jest.mock( '../../lib/get-ipc-api' );
+
+describe( 'useArchiveSite', () => {
+	const LOCAL_SITE_ID = '1658e275-3e68-4aff-a016-2dbf9c5de3db';
+	const mockPost = jest.fn();
+
+	beforeEach( () => {
+		jest.clearAllMocks();
+
+		mockPost.mockImplementation( () => Promise.resolve( {} ) );
+
+		( useAuth as jest.Mock ).mockImplementation( () => ( {
+			client: { req: { post: mockPost } },
+		} ) );
+
+		( useSiteDetails as jest.Mock ).mockImplementation( () => ( {
+			uploadingSites: {},
+			setUploadingSites: jest.fn(),
+		} ) );
+
+		( getIpcApi as jest.Mock ).mockImplementation( () => ( {
+			archiveSite: () =>
+				Promise.resolve( {
+					archivePath: '/tmp/test.zip',
+					archiveSizeInBytes: 1000,
+				} ),
+			getFileContent: () => Promise.resolve( new Uint8Array() ),
+			removeTemporalFile: () => Promise.resolve(),
+			showErrorMessageBox: () => {},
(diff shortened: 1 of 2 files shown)


src/hooks/use-archive-site.ts
```diff
@@ -109,6 +111,11 @@ export function useArchiveSite() {
 					formData.push( [ 'wordpress_version', wordpressVersion ] );
 				}
 
+				const connectedProductionSite = connectedSites.find( ( site ) => ! site.isStaging );
+				if ( connectedProductionSite ) {
+					formData.push( [ 'connected_site_id', String( connectedProductionSite.id ) ] );
```
sejas: Why are we casting the Int to String?
I've seen that in 170864-ghe-Automattic/wpcom/files#diff-2fce9fc365683142c30cb822e249f39d53d53ca73e431da75987e2622349d43dR240 we are casting it back to int again. I wonder if it makes sense using just the default type Int in the first place.

```suggestion
					formData.push( [ 'connected_site_id', connectedProductionSite.id ] );
```

sejas (commented):
Ivan, the PR looks great. I confirm that the new entry is created in the production site.

Screenshot 2025-01-28 at 00 01 56


I left a comment about not changing the type of connected_site_id and use Int in the backend.
It would be great if we can add some tests around this behaviour.

src/hooks/use-archive-site.ts
ivan-ottinger: Good catch! I have fixed the issue in https://github.com/Automattic/studio/pull/838/commits/03e0ecbbb786ee962222becca8dcc3d35fd964d4.

ivan-ottinger:
Thank you for your testing and review, @sejas . 🙂 

> I left a comment about not changing the type of connected_site_id and use Int in the backend.
It would be great if we can add some tests around this behaviour.

Do you mean adding tests like checking whether the API request will include the `connected_site_id` if a connected site exists?

sejas:
> Automattic <> Titan - Monthly sync

Yep!, exactly.
Let's check if we add the production Site ID to the API call when we have a Production and Staging site connected.

ivan-ottinger:
> > Automattic <> Titan - Monthly sync
> 
> Yep!, exactly. Let's check if we add the production Site ID to the API call when we have a Production and Staging site connected.

Thanks for clarifying, @sejas. I have added two tests in https://github.com/Automattic/studio/pull/838/commits/cecc5c4c01a9ea747c2101998c80408ccde5d251 and https://github.com/Automattic/studio/pull/838/commits/33f8a3c1f4c83972ef0cea84294a30f71d489fd4:

- `should include connected_site_id in form data when a production WPCOM site is connected`
- `should not include connected_site_id in form data when no WPCOM site is connected`

→ `npm test src/hooks/tests/use-archive-site.test.ts`.

Feel free to let me know what you think. 🙂

src/hooks/tests/use-archive-site.test.ts
```diff
@@ -0,0 +1,92 @@
+import { renderHook, act } from '@testing-library/react';
```
sejas, line 1: We could add the instructions about how to execute this test.

```suggestion
// To run tests, execute `npm run test -- src/hooks/tests/use-archive-site.test.ts` from the root directory
import { renderHook, act } from '@testing-library/react';
```

sejas (approved):
Thanks for adding the tests 👍 
Looks great. Thanks for using the Site ID as integer.

src/hooks/tests/use-archive-site.test.ts
wojtekn: What is the reason to add instruction in code for this particular text?

wojtekn (approved):

PR #842, opened 2025-01-24
Split sync:pull into 2 steps

ashfame:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10036

## Proposed Changes

In this PR, we change the sync:pull operation to check the backup size before download to prompt user if they really want to continue in case of site being too large to push back again.

Screenshot 2025-01-24 at 22 55 35

"Cancel" halts the pull operation. "Continue" starts downloading the backup for restoration.

Also, added console log calls in a group:

Screenshot 2025-01-24 at 22 03 50

## Testing Instructions

- Restore a normal sized site (<2GB) and it would succeed exactly before the changes in this PR
- Modify `SYNC_PUSH_SIZE_LIMIT_BYTES` to be really small so that normal size backup is also over the max limit
- Try "pull" and you will see the prompt for Cancel/Continue

## Pre-merge Checklist


src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -90,6 +91,9 @@ export function useSyncPull( {
 			if ( ! client ) {
 				return;
 			}
+
+			console.groupCollapsed( 'Sync Pull' );
```
wojtekn, line 95: I'm not entirely convinced we need that debugging in the browser console, but thanks to the collapsed message group feature and clear messages, the whole chain looks okay.

src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -101,6 +105,7 @@ export function useSyncPull( {
 			} );
 
 			try {
+				console.log( 'Signal for backup' );
```
wojtekn: Would 'Initialize backup' be clearer?

src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -128,46 +137,108 @@ export function useSyncPull( {
...
-				status: pullStatesProgressInfo.importing,
-			} );
+				if ( fileSize > SYNC_PUSH_SIZE_LIMIT_BYTES ) {
+					console.log( 'File size exceeds limit, prompting user' );
+					const { response: userChoice } = await getIpcApi().showMessageBox( {
+						type: 'warning',
+						title: __( 'Large Backup File' ),
+						message: sprintf(
+							__(
+								'The backup file is over %s GB. You will not be able to push back the site. Continue?'
```
wojtekn: We could improve the message wording, e.g.:

```suggestion
								"Your site's backup exceeds %s GB. Pulling it will prevent you from pushing the site back. Do you want to continue?"
```

Could we also add a line break before the question and change title/message fields to message/details, similarly to 'Overwrite Production site'?

src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -128,46 +137,108 @@ export function useSyncPull( {
...
+							),
+							SYNC_PUSH_SIZE_LIMIT_BYTES
```
wojtekn: I believe it should use the GB constant.

```suggestion
							SYNC_PUSH_SIZE_LIMIT_GB
```

src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -128,46 +137,108 @@ export function useSyncPull( {
...
+				console.log( 'Backup file size:', { fileSize, limit: SYNC_PUSH_SIZE_LIMIT_BYTES } );
 
-			updatePullState( selectedSite.id, remoteSiteId, {
-				status: pullStatesProgressInfo.importing,
-			} );
+				if ( fileSize > SYNC_PUSH_SIZE_LIMIT_BYTES ) {
+					console.log( 'File size exceeds limit, prompting user' );
+					const { response: userChoice } = await getIpcApi().showMessageBox( {
+						type: 'warning',
+						title: __( 'Large Backup File' ),
```
wojtekn:
```suggestion
						title: __( "Large site's backup" ),
```

wojtekn (commented):
The flow looks good, and works as expected.

We could change 'Pulling changes' message as now it makes less sense that it made before. I would propose 'Initializing backup' instead.

wojtekn:
@ashfame I noticed one issue while testing. When I get a large backup dialog and I decide to cancel, I get the alert when I try to quit Studio:

Screenshot 2025-01-28 at 10 24 10

src/hooks/sync-sites/use-sync-pull.ts
ashfame: I found myself constantly adding these during development to understand the flow. I think they are quite useful to have in dev env. I would even go far as saying, we should have lots of these with different flows that happen. Having them all collapsed in their own group doesn't create clutter either. Perhaps being able to toggle them or if they only get logged during dev env, would be better.

src/hooks/sync-sites/use-sync-pull.ts
ashfame: I struggled with naming it, since the backup happens on remote and not locally. But happy to change it whatever you think is appropriate. Considering these are only log messages and not really exposed to user like error messages, I didn't worry too much about it.

How does "Initializing backup on remote" sound?

src/hooks/sync-sites/use-sync-pull.ts
wojtekn: Sure, we can work on a toggle in the future. For now, we should be good as soon as they go to the browser console, as the built binary shouldn't include the dev console at all.

src/hooks/sync-sites/use-sync-pull.ts
wojtekn: Sounds clear, thanks.

src/hooks/sync-sites/use-sync-pull.ts
ashfame: Re: line break, sure!

Re: copy, it's not the pulling that prevents it, so maybe we shouldn't emphasize it? How about this:

`Your site's backup exceeds %s GB. You will not be able to push the site back. Do you want to continue?`

src/hooks/sync-sites/use-sync-pull.ts
ashfame: Good catch! Didn't realize using a modified value of constant locally.

src/hooks/sync-sites/use-sync-pull.ts
wojtekn: I think the version I proposed is more precise. If the Studio site currently takes 50MB, and the user is going to pull a 2500 MB site, the pulling is the exact reason that will prevent them from pushing it back.

ashfame: Ah yes, makes sense. Thanks for explaining!

ashfame:
@wojtekn Regarding the message when you try to close the app, that's cool. I am seeing it for the first time. Looks like functioning as it should. Can you elaborate the issue?

wojtekn:
> @wojtekn Regarding the message when you try to close the app, that's cool. I am seeing it for the first time. Looks like functioning as it should. Can you elaborate the issue?

It should alert users when they try to close the app while the sync is running. In my case, the sync was not running, but I still got an alert:
1. Start a pull
2. See the large site notice
3. Click cancel (sync is not running anymore)
4. Close the Studio
5. Note that it displays the warning, but shouldn't

ashfame:
Ah I see, missed "sync was not running" part. Thanks for explaining!

ashfame:
@wojtekn Fixed the false "sync in progress" alarm. Ready for review again.

src/hooks/sync-sites/use-sync-pull.ts
wojtekn: Could you please adjust how we display the message, mentioned in my previous comment?

> Could we also add a line break before the question and change title/message fields to message/details, similarly to 'Overwrite Production site'?

src/hooks/sync-sites/use-sync-pull.ts
wojtekn: Thanks, looks good now.

wojtekn:
Thanks @ashfame - it doesn't show the dialog when I try to close the app now.

Could you please improve dialog message formatting and adjust the label mentioned in the comment above?

> We could change 'Pulling changes' message as now it makes less sense that it made before. I would propose 'Initializing backup' instead.

src/hooks/sync-sites/use-sync-pull.ts
ashfame: Sorry missed doing that, done now!

ashfame:
@wojtekn Sorry missed doing that, but done now.

wojtekn:
Thanks @ashfame , they look consistent now:

Screenshot 2025-01-30 at 12 11 58

Screenshot 2025-01-30 at 12 12 36

Could you please change the label too?

> We could change 'Pulling changes' message as now it makes less sense that it made before. I would propose 'Initializing backup' instead.

ashfame:
@wojtekn ~Sorry, I am not sure what label are you referring to?~ Got clarification in DM

katinthehatsite (approved):
The changes look good 👍

PR #844, opened 2025-01-27
Delete too verbose logging to reduce noise

wojtekn:
## Related issues

- N/A

## Proposed Changes

- I propose to delete too verbose logging to reduce noise. There is no point in logging the content of the loaded JSON file, as well as the data of the server being started.

## Testing Instructions

1. Start Studio
2. Confirm it still works and logs proper data when it starts, and when it starts the site

## Pre-merge Checklist


src/site-server.ts
```diff
@@ -98,7 +98,7 @@ export class SiteServer {
 			);
 		}
 
-		console.log( 'Starting server with options', sanitizeForLogging( options ) );
```
ivan-ottinger, line 101: Since we are removing `sanitizeForLogging` here, we can also remove its import (in both `src/site-server.ts` and `src/ipc-handlers.ts`).

ivan-ottinger (approved):
The proposed changes look great and I did not observe any regressions.

The server starts without any issues and the among of logs is reduced as expected. 👌🏼 

Just noticed that we can also remove `sanitizeForLogging` import from both files as it isn't needed there anymore. Please see the inline comment.

src/site-server.ts
wojtekn: Good catch, removed in 13bfd784edae44f103bdaa3c879be5faf8218397

ivan-ottinger: Thanks. 🙂

ivan-ottinger: It is also in `src/ipc-handlers.ts`.

wojtekn: Thanks: 98a0fa1aec9f4d3f31f16655601dbf93ff9d2b08. It's easy to miss now as we have 110 lint warnings. 🤷

ivan-ottinger: Cool! All is looking good. 👍🏼 I did not notice anything else. 🙂

PR #847, opened 2025-01-27
Mount the SQLite plugin to the Playground filesystem

bgrgicak:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/9235

## Proposed Changes

- Mount the SQLite Plugin to the Playground FS instead of adding them to the site's mu-plugins folder.
- Remove the SQLite plugin from mu-plugins if it was previously added.
- Automatically switch to MySQL if `wp-config.php` includes valid MySQL credentials.
- When importing a site only keep the `database` folder.
- Remove code that kept the SQLite plugin up to date in the site folder. 

### Mount SQLite plugin

- Start the production version of Studio
- Create a new site
-  Confirm that the site has the SQLite plugin inside the mu-plugins folder
- Close Studio
- Checkout this branch
- Open the dev version of Studio
- Start the site you previously created
- Confirm that the site works
- Confirm that the site doesn't have the SQLite plugin inside the mu-plugins folder
- Create a new site
- Confirm that it works

### MySQL switcher

- Change the site name in wp-admin (so that you know it's using the SQLite DB)
- Stop the site
- Add valid MySQL credentials to `wp-config.php`
- Start the site
- Confirm that it uses MySQL
- Comment out some of the credentials in wp-config.php` like `define( 'DB_HOST', '127.0.0.1' );`
- Restart the site
- Confirm that it loads the SQLite DB again

## Pre-merge Checklist


bgrgicak:
@wojtekn I realized that we can't use WP-cli in the terminal if we move the SQLite plugin to internals because it wouldn't have a valid SQLite connection. I assume this is a blocker for us. How would you like to proceed?

wojtekn:
> I realized that we can't use WP-cli in the terminal if we move the SQLite plugin to internals because it wouldn't have a valid SQLite connection. I assume this is a blocker for us. How would you like to proceed?

We can leave this one for now and come back to this later.

bgrgicak:
I will close this PR as we don't plan to work on it for now.

PR #835, opened 2025-01-23
Switch to yauzl for unzipping files without loading everything into memory

ashfame:
## Related issues

- Fixes #800

## Proposed Changes

- Remove `adm-zip` lib in favor of `yauzl` which is more mature & adheres to spec tightly
- Read zip file without loading it all into memory 

## Testing Instructions

- Import a 2GB+ back up file. I tested with a 8.93GB file.

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #850, opened 2025-01-28
Disable ESLint rule `import/no-named-as-default-member`

fredrikekelund:
## Related issues

–

## Proposed Changes

As I mentioned in https://github.com/Automattic/studio/pull/835#discussion_r1928772419, it's my fault that ESLint's `import/no-named-as-default-member` rule is enabled (happened in https://github.com/Automattic/studio/pull/825), and this is currently generating a lot of linter noise.

Aside from the noisy linter warnings, I don't see a good reason for enabling this rule in our case. I think it's overly strict (see [docs](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-named-as-default-member.md)). 

## Testing Instructions

Code review should suffice. Check the linter output if you want to, we should now have 31 warnings instead of 111.

## Pre-merge Checklist

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,9 +32,10 @@
 			{
 				"args": "after-used",
 				"argsIgnorePattern": "^_",
+				"caughtErrors": "none",
+				"ignoreRestSiblings": true,
 				"vars": "all",
-				"varsIgnorePattern": "^_",
-				"ignoreRestSiblings": true
+				"varsIgnorePattern": "^_"
 			}
 		],
 		"import/no-named-as-default-member": "off",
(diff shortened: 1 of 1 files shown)


wojtekn (approved):
Thanks for the cleanup, looks good.

Most of the remaining errors are related to try/catch that don't use the error object.

Would it make sense to disable those by setting `"caughtErrors": "none"` for `no-unused-vars`?

fredrikekelund:
> Would it make sense to disable those by setting `"caughtErrors": "none"` for `no-unused-vars`?

Good idea 👍 I pushed that change, too

PR #852, opened 2025-01-28
Include mu-plugins in Studio site import

### LINKED ISSUE #500: Consider importing mu-plugins directories
We currently ignore mu-plugins directories in our import workflow. We should consider properly importing them.

bgrgicak:
## Related issues

- Fixes #500

## Proposed Changes

- Don't delete the `mu-plugins/sqlite-integration-plugin` folder during import
- Include mu-plugins in Playground, Jetpack, Wpress, and Local imports
- Prevent import from failing if an empty folder wasn't extracted

## Testing Instructions

- Checkout this branch
- Optionally [download import files with included mu-plugin folders](https://drive.google.com/drive/folders/1lnPb5kq5cE5AIz2NEi5nHBew5FqA74-x?usp=drive_link) if you don't have them already
- Start Studio

### Jetpack Import

- Import the Jetpack import file
- Confirm that the site name is updated to Jetpack
- Confirm that the mu-plugins/sqlite-integration-plugin folder is present
- Confirm that a mu-plugins/jetpack.php file is present

### Playground Import

- Import the Playground import file
- Confirm that the site name is updated to Playground
- Confirm that the mu-plugins/sqlite-integration-plugin folder is present
- Confirm that a mu-plugins/playground.php file is present

### Wpress Import

- Import the Wpress import file
- Confirm that the site name is updated to Wpress
- Confirm that the mu-plugins/sqlite-integration-plugin folder is present
- Confirm that a mu-plugins/wpress.php file is present

### Local Import

- Import the Local import file
- Confirm that the site name is updated to Local
- Confirm that the mu-plugins/sqlite-integration-plugin folder is present
- Confirm that a mu-plugins/local.php file is present

## Pre-merge Checklist

### DIFF
--- a/src/lib/import-export/import/importers/importer.ts
+++ b/src/lib/import-export/import/importers/importer.ts
@@ -4,7 +4,7 @@ import fs, { createReadStream, createWriteStream } from 'fs';
 import fsPromises from 'fs/promises';
 import path from 'path';
 import { createInterface } from 'readline';
-import { lstat, move, ensureDir } from 'fs-extra';
+import { lstat, move } from 'fs-extra';
 import semver from 'semver';
 import { DEFAULT_PHP_VERSION } from '../../../../../vendor/wp-now/src/constants';
 import { SiteServer } from '../../../../site-server';
@@ -218,9 +218,18 @@ abstract class BaseBackupImporter extends BaseImporter {
 		const wpContentDestDir = path.join( rootPath, 'wp-content' );
 		for ( const files of Object.values( wpContent ) ) {
 			for ( const file of files ) {
-				const stats = await lstat( file );
-				// Skip if it's a directory
-				if ( stats.isDirectory() ) {
+				try {
+					const stats = await lstat( file );
+					// Skip if it's a directory
+					if ( stats.isDirectory() ) {
+						continue;
+					}
+				} catch {
+					/**
+					 * If the file does not exist, skip it.
+					 * This can happen if a empty directory is included in the backup
+					 * because the empty directory won't be included in the extraction.
+					 */
 					continue;
 				}
 				const relativePath = path.relative(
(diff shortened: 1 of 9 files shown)


src/lib/import-export/import/importers/importer.ts
```diff
@@ -215,9 +218,18 @@ abstract class BaseBackupImporter extends BaseImporter {
...
+					// Skip if it's a directory
+					if ( stats.isDirectory() ) {
+						continue;
+					}
+				} catch {
+					/**
+					 * If the file does not exist, skip it.
+					 * This can happen if a empty directory is included in the backup
+					 * because the empty directory won't be included in the extraction.
+					 */
```
bgrgicak, line 232: I discovered this issue in a Local export.
The `uploads/2025/01` folder was empty on the site and the `01` folder is missing from the extracted ZIP.
This caused `lstat` to fail because the directory didn't exist.

src/lib/import-export/import/types.ts
```diff
@@ -12,6 +12,7 @@ export interface WpContent {
 	uploads: string[];
 	plugins: string[];
 	themes: string[];
+	muPlugins?: string[];
```
bgrgicak, line 15: Should I make `muPlugins` mandatory?

wojtekn: Let's keep it optional, as none of the supported backup formats includes it by default.

wojtekn:
@bgrgicak the code change looks good and works as expected for the packages you provided.

However, it doesn't work for my Playground import file, which works fine in Studio 1.3.2.

Also, we should include `mu-plugins/` to export for completeness, so the site exported from Studio and imported again would have the same set of plugins. We can add a separate issue for that, though.

wojtekn:
> However, it doesn't work for my Playground import file, which works fine in Studio 1.3.2.

It seems it happens in the trunk, too. I've added https://github.com/Automattic/dotcom-forge/issues/10369

wojtekn (approved):

wojtekn:
> Also, we should include `mu-plugins/` to export for completeness, so the site exported from Studio and imported again would have the same set of plugins. We can add a separate issue for that, though.

I've opened https://github.com/Automattic/dotcom-forge/issues/10373 for that.

PR #857, opened 2025-01-29
chore: allow sending errors to Sentry from dev env

nightnei:
## Proposed Changes
I spent some time figuring out - why we have `development` logs in Sentry, but I don't see mine.
It's kinda edge case (since now I know 😄), but as for me, would be useful to allow sending errors by default, w/o the necessity to turn it on manually in code every time.

## Testing Instructions

Visual review should suffice

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,7 +41,7 @@ if ( ! isCLI() && ! process.env.IS_DEV_BUILD ) {
 	Sentry.init( {
 		dsn: 'https://97693275b2716fb95048c6d12f4318cf@o248881.ingest.sentry.io/4506612776501248',
 		debug: true,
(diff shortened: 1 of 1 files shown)


wojtekn:
I don't think we need to send development logs to Sentry. Can we use local debugging tools for that and leave Sentry for unexpected crashes that happen to users?

fredrikekelund:
I agree with @wojtekn. I think it's a sane default not to send errors from our development environments to Sentry.

nightnei:
> Can we use local debugging tools for that and leave Sentry for unexpected crashes that happen to users?
I am thinking, actually we need to log some issues from LDE only when we are working on improving our Sentry logs or something specific for Sentry configuration.
So yeah, I will close this PR.

PR #845, opened 2025-01-27
Mount mu-plugin polyfills to the Playground filesystem

bgrgicak:
## Related issues

- Related https://github.com/Automattic/dotcom-forge/issues/9235

To complete #9235 we also need to mount the SQLite plugin to the Playground FS instead of adding it to the site folder. 

## Proposed Changes

- Mount mu-plugin polyfills to the Playground FS instead of adding them to the site's mu-plugins folder.
- Remove mu-plugin polyfills from the site folder when the server starts.
- Add a CLI mode to `prepareWordPress`, so all Playground modes can be configured from one place to avoid code duplication.
- Add constants for internal imports.

## Testing Instructions

- Start the production version of Studio
- Create a new site
- Confirm that the site has polyfills inside the mu-plugins folder (e.g. 0-32bit-integer-warnings.php)
- Close Studio
- Checkout this branch
- Open the dev version of Studio
- Start the site you previously created
- Confirm that the site works
- Confirm that the site doesn't have polyfills inside the mu-plugins folder (e.g. 0-32bit-integer-warnings.php)
- Create a new site 
- Confirm that it works

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #861, opened 2025-01-30
Include mu-plugins in site exports

bgrgicak:
## Proposed Changes

- Include `mu-plugins` in Studio exports.
- Exclude the SQLite plugin located in mu-plugins from Studio exports.

## Testing Instructions

- Checkout this branch
- Create a site
- Add a mu-plugin to your site (it can be an empty PHP file)
- Export site
- Confirm the mu-plugin is included in the site export
- Confirm there is no SQLite plugin in mu-plugins
- (Optional) Import site into Studio
- Confirm the site works

## Pre-merge Checklist


src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -219,22 +221,32 @@ export class DefaultExporter extends EventEmitter implements Exporter {
 				uploads: [],
 				plugins: [],
 				themes: [],
+				muPlugins: [],
 			},
 		};
 
 		const siteFiles = await this.getSiteFiles();
+		const pathsToExclude = [ 'wp-content/mu-plugins/sqlite-database-integration' ];
```
fredrikekelund: Can we move this to the top level? Like a `readonly` prop on the class

src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -219,22 +221,32 @@ export class DefaultExporter extends EventEmitter implements Exporter {
...
 		siteFiles.forEach( ( file ) => {
 			const relativePath = path.relative( options.site.path, file );
 			const relativePathItems = relativePath.split( path.sep );
-			const [ wpContent, category ] = relativePathItems;
+			const [ wpContent, wpContentDirectory ] = relativePathItems;
 
+			if ( pathsToExclude.some( ( path ) => relativePath.startsWith( path ) ) ) {
+				return;
+			}
```
fredrikekelund: I feel like the `getSiteFiles` method might be a better fit for this logic. Thoughts?

fredrikekelund (changes requested):
The code and functionality here LGTM overall 👍 However, there's an issue where sites that still have the old mu-plugins (the ones we moved to Playground's internal folder in https://github.com/Automattic/studio/pull/845) will now fail to export if they aren't started before exporting.

Definitely an edge case, but I imagine it'd be relatively simple to address.

bgrgicak:
> However, there's an issue where sites that still have the old mu-plugins (the ones we moved to Playground's internal folder in https://github.com/Automattic/studio/pull/845) will now fail to export if they aren't started before exporting.

Good catch @fredrikekelund! I added these files to the list and also addressed your other comments. 
The outdated internal plugins should exist only until WP-CLI is executed for the first time or the site is started, but there is still a chance that the user will click export before the WP-CLI command runs (there is a command that runs every time a site screen is loaded in Studio to check the version).

fredrikekelund (approved):
I took the liberty of pushing a change that fixes an issue with the excluded mu-plugins on Windows. Aside from that, this LGTM 👍 Feel free to tweak my changes if needed, @bgrgicak.

> The outdated internal plugins should exist only until WP-CLI is executed for the first time or the site is started, but there is still a chance that the user will click export before the WP-CLI command runs

Yeah, this is definitely an edge case, but it _can_ happen, so I think implementing this kind of protection makes sense.

fredrikekelund:
@bgrgicak, please note that there's currently a code freeze until v1.3.3 is released (see pfHvTO-lI-p2), so please hold off with merging this for now.

PR #863, opened 2025-01-30
Add user identifier for Sentry

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10341

## Proposed Changes

Store a UUID in `appdata-v1.json` to identify users in Sentry. This will help us determine how many users are affected by the issues logged to Sentry.

The UUID is (like all UUIDs) random, meaning we don't log any PII to Sentry.

Per @nightnei's suggestion, I considered naming the `sentryUserId` key in the user config differently to indicate that it could be used for other reporting purposes in the future. However, I concluded that this is hypothetical, and it's better to optimize for the scenario at hand by using an explicit name.

## Testing Instructions

1. Apply the following diff

```diff
diff --git a/src/index.ts b/src/index.ts
index 741245a..838464b 100644
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,7 +41,7 @@ if ( ! isCLI() && ! process.env.IS_DEV_BUILD ) {
        Sentry.init( {
                dsn: 'https://97693275b2716fb95048c6d12f4318cf@o248881.ingest.sentry.io/4506612776501248',
                debug: true,
-               enabled: process.env.NODE_ENV !== 'development' && process.env.NODE_ENV !== 'test',
+               enabled: process.env.NODE_ENV !== 'test',
                release: `${ app.getVersion() ? app.getVersion() : COMMIT_HASH }-${ getPlatformName() }`,
        } );
 }
@@ -297,6 +297,8 @@ async function appBoot() {
                // Bump stat for unique weekly app launch, approximates weekly active users
                bumpAggregatedUniqueStat( 'local-environment-launch-uniques', process.platform, 'weekly' );
 
+               Sentry.captureException( new Error( 'App booted' ) );
+
                finishedInitialization = true;
        } );
```

2. Start the app
3. Find the line `Setting Sentry user ID: xxxxx…` in your console output and take note of your ID
4. Check Sentry issues from the development environment
5. Find the issue with the message `App booted`
6. Ensure that your Sentry user ID appears under `Tags > user` (feel free to ping me for directions)

## Pre-merge Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -34,7 +34,7 @@ import {
 } from './migrations/migrate-from-wp-now-folder';
 import { setupWPServerFiles, updateWPServerFiles } from './setup-wp-server-files';
 import { stopAllServersOnQuit } from './site-server';
-import { loadUserData } from './storage/user-data'; // eslint-disable-next-line import/order
+import { loadUserData, saveUserData } from './storage/user-data'; // eslint-disable-next-line import/order
 import { setupUpdates } from './updates';
 
 if ( ! isCLI() && ! process.env.IS_DEV_BUILD ) {
@@ -271,6 +271,16 @@ async function appBoot() {
 			await migrateFromWpNowFolder();
 		}
 
+		const userData = await loadUserData();
+
+		if ( ! userData.sentryUserId ) {
+			userData.sentryUserId = crypto.randomUUID();
+			await saveUserData( userData );
+		}
+
+		console.log( 'Setting Sentry user ID:', userData.sentryUserId );
+		Sentry.setUser( { id: userData.sentryUserId } );
+
 		createMainWindow();
 
 		// Handle CLI commands
@@ -278,7 +288,6 @@ async function appBoot() {
 		executeCLICommand();
 
 		// Bump stats for the first time the app runs - this is when no lastBumpStats are available
-		const userData = await loadUserData();
 		if ( ! userData.lastBumpStats ) {
 			bumpStat( 'studio-app-launch-first', process.platform );
 		}
--- a/src/__mocks__/@sentry/electron/main.ts
+++ b/src/__mocks__/@sentry/electron/main.ts
@@ -1,3 +1,4 @@
 export const init = jest.fn();
 export const addBreadcrumb = jest.fn();
 export const captureException = jest.fn();
+export const setUser = jest.fn();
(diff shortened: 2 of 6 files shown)


src/index.ts
```diff
@@ -271,14 +271,23 @@ async function appBoot() {
 			await migrateFromWpNowFolder();
 		}
 
+		const userData = await loadUserData();
+
+		if ( ! userData.sentryUserId ) {
+			userData.sentryUserId = crypto.randomUUID();
```
nightnei, line 277: WDYT about `studioInstanceId`?

```suggestion
			userData.studioInstanceId = crypto.randomUUID();
```

src/index.ts
```diff
@@ -271,14 +271,23 @@ async function appBoot() {
...
+			await saveUserData( userData );
+		}
+
+		console.log( 'Setting Sentry user ID:', userData.sentryUserId );
+		Sentry.setUser( { id: userData.sentryUserId } );
```
nightnei, line 282: Also, looking at Sentry UI - it seems `studio-instance-id` will be more clear at a glance:
```suggestion
		Sentry.setUser( { studio-instance-id: userData.sentryUserId } );
```

nightnei (changes requested):
In general LGTM, but I want to talk a little bit about naming in Studio and Sentry :)

src/index.ts
fredrikekelund: `Sentry.setUser` doesn't accept arbitrary keys. The `id` parameter is standardized. See [docs](https://docs.sentry.io/platforms/javascript/guides/electron/enriching-events/identify-user/)

src/index.ts
fredrikekelund: This is what I referred to in the PR description 😄

> Per @nightnei's suggestion, I considered naming the `sentryUserId` key in the user config differently to indicate that it could be used for other reporting purposes in the future. However, I concluded that this is hypothetical, and it's better to optimize for the scenario at hand by using an explicit name.

To elaborate, I think a name like `studioInstanceId` is more difficult to understand at a glance, and there's currently no plan to use it for other purposes than Sentry reporting.

To that end, I think it's better to cross that bridge when and if we get there.

nightnei (approved):

src/index.ts
nightnei: Oh, in this case I definitely ok with naming 👍

PR #866, opened 2025-01-31
Update localization docs

nightnei:
## Proposed Changes
Added info about AutoProxxy to "Import to GlotPress" section.

## Testing Instructions

Visual review is suffice

### DIFF
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -26,7 +26,7 @@ If you want to add support for another language you will need to add it to the
 #### Step 2: Import to GlotPress:
 
    1. Open [our project in GlotPress](https://translate.wordpress.com/projects/studio/).
-   2. Click the **Project actions** menu. (You should be proxied through AutoProxxy or similar way)
+   2. Click the **Project actions** menu. You must have admin permissions to view the menu.
    3. Click **Import Originals**.
    4. Import `out/pots/bundle-strings.pot` (auto-detecting the file format is fine).
 
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -26,7 +26,7 @@ If you want to add support for another language you will need to add it to the
 #### Step 2: Import to GlotPress:
 
    1. Open [our project in GlotPress](https://translate.wordpress.com/projects/studio/).
-   2. Click the **Project actions** menu. You must have admin permissions to view the menu.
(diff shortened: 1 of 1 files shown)


docs/localization.md
```diff
@@ -26,7 +26,7 @@ If you want to add support for another language you will need to add it to the
 #### Step 2: Import to GlotPress:
 
    1. Open [our project in GlotPress](https://translate.wordpress.com/projects/studio/).
-   2. Click the **Project actions** menu.
+   2. Click the **Project actions** menu. (You should be proxied through AutoProxxy or similar way)
```
katinthehatsite: Aren't these documents public? I am not completely convinced that we should mention AutoProxxy there.

sejas: What about mentioning the requirement and not the tools?

```suggestion
   2. Click the **Project actions** menu. You must have admin permissions to view the menu.
```

sejas (commented):
@nightnei , Thanks for adding more context. I think clarifying the need to be an admin is important for other to avoid the issue in the future.
I see the point of Kat about mentioning the exact tool.

I left a suggestion. Let me know what do you think.

docs/localization.md
nightnei: @sejas yeah, makes sense 👍

fredrikekelund: I think this is OK, but people could still get confused about what "admin permissions" means. We could tweak it to something like "You must be a proxied Automattician to view the menu" to make it more instantly understandable. However, I also assume a12s are typically proxied, making this documentation somewhat redundant…

nightnei: > However, I also assume a12s are typically proxied, making this documentation somewhat redundant…

Sometimes I forget to run AutoProxxy, since it's not critical for Studio development. And that was the case, so I could not understand where this menu is, since I had access but the menu is hidden.

nightnei: @sejas WDYT about `You must be a proxied Automattician to view the menu`?
I think it's indeed more direct, clear and we don't mention AutoProxxy

sejas (approved):

docs/localization.md
sejas: Sounds good to me. It's simple and direct 👍.

PR #856, opened 2025-01-29
Display error when id is missing from /me when authenticating

ashfame:
## Related issues

- Related to: https://github.com/Automattic/dotcom-forge/issues/10215

## Proposed Changes

-

## Testing Instructions

-

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #871, opened 2025-01-31
Validate `authToken` from user data file before returning it

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10215
- Fixes https://github.com/Automattic/dotcom-forge/issues/10258

## Proposed Changes

In https://github.com/Automattic/dotcom-forge/issues/10215, we observed that the `authToken` object in the `appdata-v1.json` file sometimes misses the `id` prop. Even if the TS type previously designated this prop as optional, that's not actually how the code works—we need the WordPress.com user ID for several features that interact with the WP.com APIs.

In this PR, I've addressed this problem by:

1. Adding [zod](https://zod.dev/) as a dependency (a validation library with incredibly good TS support that's successfully being used on Tumblr)
2. Adding a zod schema for the `authToken` object in the user data file. Most properties are _not_ optional.
3. Making `getToken` in `src/lib/oauth.ts` return null if the zod validation fails.

Moreover, I took the opportunity to move the `onOpenUrlCallback` function from `src/index.ts` to `src/lib/oauth.ts`. Given it deals exclusively with OAuth-related logic, it makes more sense for it to live there. I also inlined the logic from `setUpAuthCallbackHandler` into `onOpenUrlCallback`, reducing the number of events being passed around.

## Testing Instructions

1. Be logged out
2. Log in
3. Ensure that the Assistant works as expected
4. Ensure that connecting sites in the Sync tab works as expected
5. Ensure that the Share tab works as expected

## Pre-merge Checklist

### DIFF
--- a/src/components/auth-provider.tsx
+++ b/src/components/auth-provider.tsx
@@ -1,4 +1,5 @@
 import * as Sentry from '@sentry/electron/renderer';
+import { useI18n } from '@wordpress/react-i18n';
 import { createContext, useState, useEffect, useMemo, useCallback, ReactNode } from 'react';
 import WPCOM from 'wpcom';
 import { useI18nData } from '../hooks/use-i18n-data';
@@ -36,14 +37,19 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 	const [ client, setClient ] = useState< WPCOM | undefined >( undefined );
 	const [ user, setUser ] = useState< AuthContextType[ 'user' ] >( undefined );
 	const { locale } = useI18nData();
+	const { __ } = useI18n();
 
 	const authenticate = useCallback( () => getIpcApi().authenticate(), [] );
 
 	useIpcListener( 'auth-updated', ( _event, { token, error } ) => {
 		if ( error ) {
-			Sentry.captureException( error );
+			getIpcApi().showErrorMessageBox( {
+				title: __( 'Authentication error' ),
+				message: __( 'Please try again.' ),
+			} );
 			return;
 		}
+
 		setIsAuthenticated( true );
 		setClient( createWpcomClient( token.accessToken, locale ) );
 		setUser( {
--- a/src/components/auth-provider.tsx
+++ b/src/components/auth-provider.tsx
@@ -46,13 +46,11 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 		}
 		setIsAuthenticated( true );
 		setClient( createWpcomClient( token.accessToken, locale ) );
-		if ( token.id || token.email || token.displayName ) {
-			setUser( {
-				id: token.id,
-				email: token.email,
-				displayName: token.displayName || '',
-			} );
-		}
+		setUser( {
+			id: token.id,
+			email: token.email,
+			displayName: token.displayName || '',
+		} );
 	} );
 
 	const logout = useCallback( async () => {
--- a/src/components/auth-provider.tsx
+++ b/src/components/auth-provider.tsx
@@ -47,8 +47,8 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 		setClient( createWpcomClient( token.accessToken, locale ) );
 		if ( token.id || token.email || token.displayName ) {
 			setUser( {
-				id: token.id || null,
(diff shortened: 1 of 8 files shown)


src/tests/index.test.ts
fredrikekelund, line 1: I had to edit this file to accommodate my changes to `src/lib/oauth.ts`, and I took the opportunity to simplify some previously somewhat convoluted logic.

src/lib/oauth.ts
```diff
@@ -1,26 +1,28 @@
...
+} );
+
+export type StoredToken = z.infer< typeof authTokenSchema >;
 
 async function getToken(): Promise< StoredToken | null > {
 	try {
 		const userData = await loadUserData();
-		return userData[ TOKEN_KEY ] ?? null;
+		const parsed = authTokenSchema.parse( userData.authToken );
+		return parsed;
```
ashfame:
```suggestion
		return authTokenSchema.parse( userData.authToken );
```

src/lib/oauth.ts
```diff
@@ -31,7 +33,7 @@ async function storeToken( tokens: StoredToken ) {
 		const userData = await loadUserData();
 		await saveUserData( {
 			...userData,
-			[ TOKEN_KEY ]: tokens,
+			authToken: tokens,
```
ashfame: I believe it's a single token? If so, should rename it to singular? `tokens` to `token`

src/lib/oauth.ts
```diff
@@ -67,32 +66,37 @@ export async function isAuthenticated(): Promise< boolean > {
 	return !! token;
 }
 
-export async function handleAuthCallback( hash: string ): Promise< Error | StoredToken > {
+async function handleAuthCallback( hash: string ): Promise< StoredToken > {
```
ashfame, line 68: 👌

ashfame (approved):
Nice work! I noticed 3 tests that were skipping are also resolved now. 🎆

src/lib/oauth.ts
```diff
@@ -67,32 +66,37 @@ export async function isAuthenticated(): Promise< boolean > {
...
 	if ( isNaN( expiresIn ) || expiresIn === 0 || ! accessToken ) {
-		return new Error( 'Error while getting token' );
+		throw new Error( 'Error while getting token' );
 	}
-	let response: { ID?: number; email?: string; display_name?: string } = {};
+
 	try {
-		response = await new wpcom( accessToken ).req.get( '/me?fields=ID,email,display_name' );
+		const response = await new wpcom( accessToken ).req.get( '/me?fields=ID,email,display_name' );
+
```
ashfame: @fredrikekelund I guess here would be the new point to specifically log it to sentry if `ID` is ever returned empty from api endpoint unless whatever zod will throw would capture that info?

src/components/auth-provider.tsx
```diff
@@ -48,8 +48,8 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 		setClient( createWpcomClient( token.accessToken, locale ) );
 		if ( token.id || token.email || token.displayName ) {
```
ashfame: This `if` check is now redundant?

ashfame: And `if (error)` seems to be the best place to show an error message to the user?

src/lib/oauth.ts
fredrikekelund: We already have a `Sentry.captureException` call in the catch clause, and the nice thing about zod's `parse` function is that it includes a clear explanation of the validation error in the error **message**. Here's an example:
```
[
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "null",
    "path": [
      "accessToken"
    ],
    "message": "Expected string, received null"
  }
]
```

src/components/auth-provider.tsx
fredrikekelund: Makes sense 👍 I've added an error modal here, inspired by https://github.com/Automattic/studio/pull/856

fredrikekelund:
Thanks for the speedy and diligent review, @ashfame 🙏 I've addressed your feedback and added an "all-purpose" error modal to be displayed if something goes wrong in `handleAuthCallback`.

PR #874, opened 2025-01-31
Handle zip imports regression

ashfame:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10369

## Proposed Changes

- Show a specific error to the user about this issue, and suggesting them to re-package the zip archive
- Change `importBackup()` and `importSite()` to not return undefined and just throw

| Before | After |
|--------|--------|
| Screenshot 2025-02-03 at 19 52 33 | Screenshot 2025-02-03 at 19 51 26 |

## Testing Instructions

- Try importing a zip file which has absolute paths in it
- Before: Error would show as in the linked issue
- After: Specific error shown

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-import-export.tsx
+++ b/src/hooks/use-import-export.tsx
@@ -85,7 +85,7 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode 
 			} ) );
 
 			const handleImportError = async ( error: unknown ) => {
-				if ( error instanceof Error && error.message.includes( 'absolute path' ) ) {
+				if ( error instanceof Error && error.message.includes( 'Error: absolute path: /' ) ) {
 					await getIpcApi().showErrorMessageBox( {
 						title: __( 'Failed importing site' ),
 						message: __(
--- a/src/hooks/use-import-export.tsx
+++ b/src/hooks/use-import-export.tsx
@@ -89,7 +89,7 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode 
 					await getIpcApi().showErrorMessageBox( {
 						title: __( 'Failed importing site' ),
 						message: __(
-							'The ZIP archive is incorrect. Try to unpack and pack it again. If this problem persists, please contact support.'
+							'The ZIP archive is invalid. Try to unpack and pack it again. If this problem persists, please contact support.'
 						),
 					} );
 				} else {
--- a/src/hooks/use-import-export.tsx
+++ b/src/hooks/use-import-export.tsx
@@ -84,15 +84,24 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode 
 				},
 			} ) );
 
-			const handleImportError = async ( error?: unknown ) => {
-				await getIpcApi().showErrorMessageBox( {
-					title: __( 'Failed importing site' ),
-					message: __(
-						'An error occurred while importing the site. Verify the file is a valid Jetpack backup, Local, Playground, .wpress or .sql database file and try again. If this problem persists, please contact support.'
-					),
-					error,
-					showOpenLogs: true,
-				} );
+			const handleImportError = async ( error: unknown ) => {
+				if ( error instanceof Error && error.message.includes( 'absolute path' ) ) {
+					await getIpcApi().showErrorMessageBox( {
+						title: __( 'Failed importing site' ),
+						message: __(
+							'The ZIP archive is incorrect. Try to unpack and pack it again. If this problem persists, please contact support.'
+						),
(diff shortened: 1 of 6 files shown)


src/hooks/use-import-export.tsx
```diff
@@ -84,15 +84,24 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
...
-				await getIpcApi().showErrorMessageBox( {
-					title: __( 'Failed importing site' ),
-					message: __(
-						'An error occurred while importing the site. Verify the file is a valid Jetpack backup, Local, Playground, .wpress or .sql database file and try again. If this problem persists, please contact support.'
-					),
-					error,
-					showOpenLogs: true,
-				} );
+			const handleImportError = async ( error: unknown ) => {
+				if ( error instanceof Error && error.message.includes( 'absolute path' ) ) {
```
ashfame: Or is this better?

```suggestion
				if ( error instanceof Error && error.message.includes( 'Error: absolute path: /' ) ) {
```

src/hooks/use-import-export.tsx
```diff
@@ -84,15 +84,24 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
...
+					await getIpcApi().showErrorMessageBox( {
+						title: __( 'Failed importing site' ),
+						message: __(
+							'The ZIP archive is incorrect. Try to unpack and pack it again. If this problem persists, please contact support.'
```
fredrikekelund:
```suggestion
							'The ZIP archive is invalid. Try to unpack and pack it again. If this problem persists, please contact support.'
```
I think "invalid" makes more sense than "incorrect" in this context.

src/lib/import-export/import/import-manager.ts
```diff
@@ -50,22 +50,23 @@ export async function importBackup(
...
+		throw new Error( 'No suitable backup handler found for the provided backup file' );
+	}
+
+	const fileList = await backupHandler.listFiles( backupFile );
+	const importer = selectImporter( fileList, '', onEvent, options );
+
+	if ( ! importer ) {
+		throw new Error( 'No suitable importer found for the provided backup contents' );
+	}
+
```
fredrikekelund, line 65: Nice 👍

src/hooks/use-import-export.tsx
fredrikekelund: I think following your updated suggestion makes sense 👍 `"absolute path"` is relatively ambiguous.

It'd be nice if yauzl provided an error code, but it doesn't…

src/tests/ipc-handlers.test.ts
```diff
@@ -118,3 +123,109 @@ describe( 'isFullscreen', () => {
...
+
+		// Simulate an import event
+		onEvent( mockEventData );
+
+		expect( mockParentWindow.webContents.send ).toHaveBeenCalledWith(
+			'on-import',
+			mockEventData,
+			'test-site'
+		);
+	} );
```
fredrikekelund: I know we have to mock a lot of stuff in our tests because of Electron, but this test still stuck out to me as very mock-heavy for testing something so simple. I would almost advocate that we remove it… Thoughts, @ashfame?

fredrikekelund (approved):
LGTM 👍 Given our current circumstance, the PR is an improvement, it tests well and the code looks good.

However… If many other unzip implementations (including the built-in tools on macOS and Windows) don't have a problem with leading slashes, I would strongly argue that Studio shouldn't either. I understand the argument that the spec makes about security, but my response would simply be: "let's automatically fix those filenames instead of outright rejecting the ZIP file". @ashfame, do you know of a way to do this with yauzl, or should we consider looking into replacement libraries for upcoming versions of Studio?

src/hooks/use-import-export.tsx
ashfame: Yep, I changed as per my suggestion. This will have to do for now.

src/tests/ipc-handlers.test.ts
ashfame: Yep, removed it!

ashfame:
@fredrikekelund Yep, I agree in principle of not leaving users hanging. Please see [this comment](https://github.com/Automattic/dotcom-forge/issues/10369#issuecomment-2623726578) where I laid down my thoughts on approaching this. I believe we need a archive handler abstraction that contains all this logic. We can continue that discussion there and figure out future next steps and let this PR be merged.

fredrikekelund:
All good with the Sentry error, it's expected 👍

PR #839, opened 2025-01-23
Make unit tests pass on Windows

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/7611

## Proposed Changes

This PR makes all unit tests pass on Windows by mocking the `path` module and/or using it correctly. Unsurprisingly, all of the failing tests were related to the differences between file paths on Unix and Windows…

I've added `src/tests/utils/platform-test-suite.ts` that wraps a test suite in a `describe.each` call that iterates over Unix and Windows platforms (meaning we run a mock test suite for both platforms at all times). Having this wrapper should make it easier to implement future tests for code that uses the `path` module.

## Testing Instructions

Run `npm test` on Windows and make sure all tests pass

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #878, opened 2025-02-03
Use `path.posix.join` for relevant paths

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10389

## Proposed Changes

This PR fixes a regression from https://github.com/Automattic/studio/pull/845. In that PR, we moved all mu-plugins but the SQLite one to be mounted in Playground's internal mu-plugins directory to give back more control to users over the regular mu-plugins directory.

The problem is that we use `path.join` for the paths passed to `php.writeFile`. The `php.writeFile` seems to expect Unix-style paths, but `path.join` is platform-aware, meaning we get Windows-style paths on Windows.

This PR fixes that problem by using `path.posix.join` instead.

## Testing Instructions

Ensure that starting a server works as expected on Windows

## Pre-merge Checklist

### DIFF
--- a/vendor/wp-now/src/execute-wp-cli.ts
+++ b/vendor/wp-now/src/execute-wp-cli.ts
@@ -97,14 +97,14 @@ export async function executeWPCli(
 			$_SERVER['argv'][0] = '${ wpCliPath }';
 
 			require( '${ wpCliPath }' );`,
-		[ path.join( PLAYGROUND_INTERNAL_SHARED_FOLDER, 'ca-bundle.crt' ) ]:
+		[ path.posix.join( PLAYGROUND_INTERNAL_SHARED_FOLDER, 'ca-bundle.crt' ) ]:
 			rootCertificates.join( '\n' ),
 	};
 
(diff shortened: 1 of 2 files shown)


nightnei (approved):
Works well with MacOS and Windows 👍 Thanks

sejas (commented):
Thank you for the quick fix! ❤️

bgrgicak:
Thanks for fixing this @fredrikekelund TIL!

Should we add a linter rule to discourage the use of `path.join` in the future?

fredrikekelund:
> Should we add a linter rule to discourage the use of path.join in the future?

In other places (anything related to the computer's regular file system), the platform-awareness of `path.join` is exactly what we want. This case — where we want Unix-style paths on all platforms — is the more uncommon one.

It's worth noting that since https://github.com/Automattic/studio/pull/839, we're running a CI job with unit tests on Windows again. Hopefully, that should help catch similar issues in the future 🙂

PR #883, opened 2025-02-04
Validate `additionalData` passed to `second-instance` event handler

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/9703

## Proposed Changes

> [!NOTE]  
> https://github.com/Automattic/studio/pull/871 also adds Zod as a dependency. The schema in this PR is very small, but #871 illustrates a slightly more complex example, and I advocate that we should start using Zod even more.

`src/lib/cli.ts` is experimental code that supports executing `wp` CLI commands with the PHP-WASM-powered WP-CLI instance bundled in Studio. It only works when the `STUDIO_TERMINAL_WP_CLI` feature flag is enabled. Still, as seen in https://github.com/Automattic/dotcom-forge/issues/9703, the `second-instance` event handler in this file sometimes triggers an error when the `additionalData` argument doesn't match the expected schema. This event callback happens when the user attempts to launch Studio while a previous instance is already running.

I haven't managed to reproduce this error. However, the error stack trace in Sentry is clear, and apparently, there's no principal guarantee that `additionalData` will always match the `{ cliCommand: string }` schema. See [the Electron docs](https://www.electronjs.org/docs/latest/api/app#event-second-instance) for additional reference.

This PR fixes the problem by adding a Zod schema for `additionalData`.

## Testing Instructions

Code review should suffice

## Pre-merge Checklist


nightnei:
> https://github.com/Automattic/studio/pull/871 also adds Zod as a dependency. The schema in this PR is very small, but https://github.com/Automattic/studio/pull/871 illustrates a slightly more complex example, and I advocate that we should start using Zod even more.

+1 👍

nightnei (approved):
LGTM 👍
Also agree with using `zod` as much as possible 👍

PR #884, opened 2025-02-04
Add back `0-dns-functions.php` mu-plugin

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10394

## Proposed Changes

Fixes a regression from https://github.com/Automattic/studio/pull/845 where the `0-dns-functions.php` mu-plugin disappeared. This plugin was introduced in https://github.com/Automattic/dotcom-forge/issues/99 and fixed an issue where Jetpack throws a fatal error in Studio because PHP-WASM doesn't define the `dns_get_record` function and the `DNS_NS` constant.

Excluding this plugin might have been intentional by @bgrgicak since I see that https://github.com/WordPress/wordpress-playground/pull/1067 added polyfills for DNS functions directly in Playground. However, we still need the `DNS_NS` constant…

## Testing Instructions

1. Import a Jetpack backup (pulling a site in the Sync tab also works)
2. Ensure that you can access wp-admin of the site

## Pre-merge Checklist

### DIFF
--- a/vendor/wp-now/src/wp-now.ts
+++ b/vendor/wp-now/src/wp-now.ts
@@ -550,14 +550,9 @@ set_error_handler(function($severity, $message, $file, $line) {
 	php.writeFile(
 		path.join( PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER, '0-dns-functions.php' ),
 		`<?php
-		// Polyfill for DNS functions/features which are not currently supported by @php-wasm/node.
+		// Polyfill for DNS features which are not currently supported by @php-wasm/node.
 		// See https://github.com/WordPress/wordpress-playground/issues/1042
 		// These specific features are polyfilled so the Jetpack plugin loads correctly, but others should be added as needed.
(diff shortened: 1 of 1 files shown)


fredrikekelund:
> Excluding this plugin might have been intentional by @bgrgicak since I see that https://github.com/WordPress/wordpress-playground/pull/1067 added polyfills for DNS functions directly in Playground. However, we still need the `DNS_NS` constant…

I proceeded with excluding the `dns_get_record` polyfill since Playground's internal polyfill already did the trick. The mu-plugin now only defines the `DNS_NS` constant.

nightnei (approved):
LGTM and the issue is fixed 👍

bgrgicak (approved):
I'm not sure how we missed the DNS_NS constant in Playground. 
I can add it there, so we won't need this polyfill in the future.

sejas (approved):
Thanks for bringing back the constant. We can delete it in the future when it's added to Playground package.
I was able to reproduce the issue in 6.7.1 but not in other sites with old Wordpress versions.

PR #864, opened 2025-01-30
chore: improve breadcrumbs in Sentry with providing more info

nightnei:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10276

## Proposed Changes
Sometimes in Sentry breadcrumbs it's not possible to understand which exactly element was clicked, either due to no selectors on element or tailwind classes which are not unique so easy to mix up buttons. Example:

Screenshot 2025-01-30 at 17 12 08
Screenshot 2025-01-30 at 17 13 11

Check out what is the result of the proposed changes. I think, as a first option for the nearest release is a good start and maybe in the future we will improve it, in the next releases.

## Testing Instructions
1. Apply the next diff:
```
diff --git a/src/index.ts b/src/index.ts
index 4545504..9ca2251 100644
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,7 +41,7 @@ if ( ! isCLI() && ! process.env.IS_DEV_BUILD ) {
        Sentry.init( {
                dsn: 'https://97693275b2716fb95048c6d12f4318cf@o248881.ingest.sentry.io/4506612776501248',
                debug: true,
-               enabled: process.env.NODE_ENV !== 'development' && process.env.NODE_ENV !== 'test',
+               enabled: process.env.NODE_ENV !== 'test',
                release: `${ app.getVersion() ? app.getVersion() : COMMIT_HASH }-${ getPlatformName() }`,
        } );
 }
diff --git a/src/lib/shell-open-external-wrapper.ts b/src/lib/shell-open-external-wrapper.ts
index 53916ef..f4cd525 100644
--- a/src/lib/shell-open-external-wrapper.ts
+++ b/src/lib/shell-open-external-wrapper.ts
@@ -12,37 +12,5 @@ const shouldMute = ( error: Error ) => {
 // while only reporting specific errors to Sentry. Some common "application not found"
 // errors are muted to avoid unnecessary error reporting.
 export const shellOpenExternalWrapper = async ( url: string ) => {
-       try {
-               await shell.openExternal( url );
-       } catch ( error ) {
-               if ( error instanceof Error && ! shouldMute( error ) ) {
-                       Sentry.captureException( error );
-               }
-
-               let title = '';
-               let message = '';
-               if ( url.startsWith( 'vscode://file/' ) ) {
-                       title = __( 'Failed to open VS Code' );
-                       message = __(
-                               'Studio is unable to open VS Code. Please ensure it is functioning correctly.'
-                       );
[...]

PR #888, opened 2025-02-04
Update RELEASE-NOTES.txt for v.1.3.3

nightnei:
## Proposed Changes
Update RELEASE-NOTES.txt for v.1.3.3

## Testing Instructions
Visual review should suffice

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,28 @@
+1.3.3
+=====
+* Handle zip imports regression #874
+* Sentry: change release naming #868
+* Sentry: improve breadcrumbs with providing more info #864
+* Sentry: add user identifier #863
+* Show warning when user tries to pull a site that is larger than 2GB #842
+* Remove internal mu-plugins during prepareWordPress #860
+* Include mu-plugins in Studio site import #852
+* Introduced Redux with refactoring ChatProvider as the first store #826
+* Fixed setting environment variables for WP CLI #814
+* Mounted mu-plugin polyfills to the Playground filesystem #845
+* Handled errors of failed opening browser on Windows #840
+* Increase width of tooltips #841
+* Fixed importing tables and Jetpack backups over 2 GB #800
+* Increased limit of demo sites from 5 to 10 #836
+* Improved fullscreen view in macOS #833
+* Fixed cleaning up of archive files created for creating/updating demo sites #816
+* Fixed absolute imports in main process #837
+* Fixed importing of .wpress backups #830
+* Added tooltips to create and update Demo Sites buttons when size limit is reached #811
+* Added description for Advanced settings of creating a new site #828
+* Allowed non-relative imports starting with src/ or vendor/ #825
+* Skip Sentry initialization in E2E tests #813
+
 1.3.2
 =====
 * Upgraded Electron to 33.3.1 (#782)
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -12,13 +12,9 @@
 * Mounted mu-plugin polyfills to the Playground filesystem #845
 * Handled errors of failed opening browser on Windows #840
 * Increase width of tooltips #841
-* Fixed importing tables and Jetpack backups over 2 GB #800
-* Increased limit of demo sites from 5 to 10 #836
 * Improved fullscreen view in macOS #833
-* Fixed cleaning up of archive files created for creating/updating demo sites #816
 * Fixed absolute imports in main process #837
 * Fixed importing of .wpress backups #830
-* Added tooltips to create and update Demo Sites buttons when size limit is reached #811
 * Added description for Advanced settings of creating a new site #828
 * Allowed non-relative imports starting with src/ or vendor/ #825
 * Skip Sentry initialization in E2E tests #813
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,14 +1,14 @@
 1.3.3
 =====
-* Handle zip imports regression #874
-* Sentry: change release naming #868
-* Sentry: improve breadcrumbs with providing more info #864
-* Sentry: add user identifier #863
-* Show warning when user tries to pull a site that is larger than 2GB #842
-* Remove internal mu-plugins during prepareWordPress #860
-* Include mu-plugins in Studio site import #852
+* Handled zip imports regression #874
+* Changed Sentry release naming #868
+* Improved Sentry breadcrumbs by providing more info #864
+* Added user identifier in Sentry #863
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.txt
```diff
@@ -1,3 +1,24 @@
+1.3.3
+=====
+* Handle zip imports regression #874
+* Sentry: change release naming #868
+* Sentry: improve breadcrumbs with providing more info #864
+* Sentry: add user identifier #863
+* Show warning when user tries to pull a site that is larger than 2GB #842
+* Remove internal mu-plugins during prepareWordPress #860
+* Include mu-plugins in Studio site import #852
```
sejas: @nightnei , can we change these sentences to use past tense?

```suggestion
* Handled zip imports regression #874
* Changed Sentry release naming #868
* Improved Sentry breadcrumbs by providing more info #864
* Added user identifier in Sentry #863
* Added a warning when user tries to pull a site that is larger than 2GB #842
* Removed internal mu-plugins during prepareWordPress #860
* Included mu-plugins in Studio site import #852
```

sejas (approved):
@nightnei , the change log looks good.
Nit: I left a comment about using past tense in every item

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,24 @@
+1.3.3
+=====
+* Handle zip imports regression #874
+* Sentry: change release naming #868
+* Sentry: improve breadcrumbs with providing more info #864
+* Sentry: add user identifier #863
```
fredrikekelund: This isn't user-facing. My instinct would be not to include it. Thoughts, @sejas?

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,24 @@
...
+* Show warning when user tries to pull a site that is larger than 2GB #842
+* Remove internal mu-plugins during prepareWordPress #860
+* Include mu-plugins in Studio site import #852
+* Introduced Redux with refactoring ChatProvider as the first store #826
```
fredrikekelund: Also not user-facing

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,24 @@
...
+* Fixed setting environment variables for WP CLI #814
```
fredrikekelund:
```suggestion
* Cosmetic improvements to how the terminal is launched on Windows #814
```

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,24 @@
+1.3.3
+=====
+* Handle zip imports regression #874
+* Sentry: change release naming #868
+* Sentry: improve breadcrumbs with providing more info #864
+* Sentry: add user identifier #863
+* Show warning when user tries to pull a site that is larger than 2GB #842
+* Remove internal mu-plugins during prepareWordPress #860
```
fredrikekelund: Not user-facing. Mostly just fixed a bug from #845

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,24 @@
...
+* Include mu-plugins in Studio site import #852
+* Introduced Redux with refactoring ChatProvider as the first store #826
+* Fixed setting environment variables for WP CLI #814
+* Mounted mu-plugin polyfills to the Playground filesystem #845
+* Handled errors of failed opening browser on Windows #840
+* Increase width of tooltips #841
+* Improved fullscreen view in macOS #833
+* Fixed absolute imports in main process #837
```
fredrikekelund: Not user-facing

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,24 @@
...
+* Fixed importing of .wpress backups #830
+* Added description for Advanced settings of creating a new site #828
+* Allowed non-relative imports starting with src/ or vendor/ #825
+* Skip Sentry initialization in E2E tests #813
```
fredrikekelund: Not user-facing

RELEASE-NOTES.txt
nightnei: I was thinking about it, but noticed that Sentry notes were added before, so decide that such type of changes probably should be included.
@sejas WDYT, should we include or it's better to keep only user facing changes?

sejas: Yeah, Let's remove what's not relevant to users and are small changes.
We can join all the sentry entries in a single line. Like : `Improved Sentry reports #863, #864....`

RELEASE-NOTES.txt
sejas: Let's remove it 👌

RELEASE-NOTES.txt
sejas: Let's remove it, or add the issue number to the same sentry line.

RELEASE-NOTES.txt
nightnei: We had such notes in the previous releases
But I think it makes sense to avoid them, since they are quite small

RELEASE-NOTES.txt
nightnei: Only Sentry related points merged to 1 line, other removed

RELEASE-NOTES.txt
sejas: yeah, it's a small change that doesn't affect any functionality. We can remove it

nightnei:
@fredrikekelund addressed feedback

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,11 @@
+1.3.3
+=====
+* Handled zip imports regression #874
```
fredrikekelund:
```suggestion
* Fixed an issue where ZIP files larger than 2GB could not be imported #835 #874
```
#835 is the real kicker

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,11 @@
+1.3.3
+=====
```
fredrikekelund, line 2:
```suggestion
=====
* Added support for mu-plugins in site import #852
* Moved mu-plugin polyfills to an internal Playground directory #845
* Fixed an issue with window control placement in fullscreen view on macOS #833
```

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,11 @@
...
+* Handled zip imports regression #874
+* Cosmetic improvements to how the terminal is launched on Windows #814
+* Fixed importing of .wpress backups #830
```
fredrikekelund, line 8:
```suggestion
* Fixed an issue with .wpress imports on Windows #830
```

fredrikekelund (approved):
I added a few items that I think should be included and made a few tweaks. Other than that, this LGTM 👍

PR #890, opened 2025-02-04
Use path.posix.join for 0-dns-functions.php

nightnei:
## Related issues

- Fixes p1738684909079949/1738681657.150269-slack-C04GESRBWKW

## Proposed Changes

The php.writeFile expects Unix-style paths, but path.join is platform-aware, meaning we get Windows-style paths on Windows.
This PR fixes that problem by using `path.posix.join` instead.

## Testing Instructions
Try to create new website on Windows, it should be successful now (before had error).
Also visual review should suffice, since we use `path.posix.join` in all other places for `php.writeFile`


sejas (approved):
Good catch!

PR #897, opened 2025-02-05
Ensure that staging sites display a `Staging` sticker from the moment they are created

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10308

## Proposed Changes

As reported in https://github.com/Automattic/dotcom-forge/issues/10308, newly created staging sites display a `Production` sticker in the Sync tab. This happens because the `/sites` API endpoint returns the wrong value for the `is_wpcom_staging_site` prop until the site setup is completed. The problem was also made worse by the `reconcileConnectedSites` function not refreshing the value for `isStaging`.

This PR changes the logic in `transformSitesResponse` so we look at the `wpcom_staging_blog_ids` arrays of other sites to determine if the given site is a staging site. This data appears to be stable from the moment a new staging site is created. For good measure, I've also updated `reconcileConnectedSites` to refresh the `isStaging` value.

## Testing Instructions

1. Log in to WordPress.com in Studio
2. Connect a Studio site to a WordPress.com site
3. If the site already has a staging site, delete it
4. Go to `wordpress.com/sites` and create a new staging site for the site in question
5. Open Studio again and periodically switch between the `Share` and `Sync` tabs (this triggers a new fetch from the `/sites` API endpoint)
6. Ensure that when the staging site appears in the Sync tab, the sticker to the left says `Staging` (see screenshot below for clarification)

Screenshot 2025-02-05 at 15 49 06

## Pre-merge Checklist

### DIFF
--- a/src/hooks/sync-sites/use-listen-deep-link-connection.ts
+++ b/src/hooks/sync-sites/use-listen-deep-link-connection.ts
@@ -30,7 +30,8 @@ export function useListenDeepLinkConnection( {
 				}
 				const newConnectedSite = transformSingleSiteResponse(
 					newConnectedSiteResponse,
-					'already-connected'
+					'already-connected',
+					false
 				);
 				await connectSite( newConnectedSite, studioSiteId );
 				if ( selectedTab !== 'sync' ) {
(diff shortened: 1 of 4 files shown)


nightnei (approved):
LGTM and works well - confirms that even if site was marked as Production in Trunk, then after switching to this branch it's fixed ("stage" badge is rendered)

PR #898, opened 2025-02-05
User capabilities returned from `/sites` endpoint determine sync support

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10257

## Proposed Changes

To be able to push a site to WordPress.com, users need the `manage_options` capability. Currently, we only look for this capability after users have initiated the push. This PR modifies our network request to the `/sites` API endpoint to return user capability data and then looks at that data to determine sync support. Hopefully, this will resolve https://github.com/Automattic/dotcom-forge/issues/10257 (it's possible that a broken Jetpack connection might still mess things up, though).

Moreover, I took the opportunity to add a [zod](https://zod.dev/) schema for the `/sites` response in `src/hooks/use-fetch-wpcom-sites/index.tsx`. Sites are parsed one-by-one, meaning we don't reject the entire response if a single site doesn't conform. Errors are reported to Sentry.

## Testing Instructions

1. Be invited to a WordPress.com site where your user has a **lower role** than administrator
2. Open the `Connect site` modal in Studio
3. Ensure that you cannot connect the site in question and that the explanation says `Missing permissions`

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-fetch-wpcom-sites/index.tsx
+++ b/src/hooks/use-fetch-wpcom-sites/index.tsx
@@ -92,7 +92,7 @@ export function transformSingleSiteResponse(
 	};
 }
 
-function transformSiteResponse( sites: unknown[], connectedSiteIds: number[] ): SyncSite[] {
+export function transformSiteResponse( sites: unknown[], connectedSiteIds: number[] ): SyncSite[] {
 	return sites
 		.map( ( rawSite ) => {
 			try {
@@ -108,10 +108,10 @@ function transformSiteResponse( sites: unknown[], connectedSiteIds: number[] ): 
 				return null;
 			}
 		} )
-		.filter( ( site ) => site !== null );
+		.filter( ( site ): site is SyncSite => site !== null );
 }
 
-export type FetchSites = () => Promise< unknown[] >;
+export type FetchSites = () => Promise< SyncSite[] >;
 
 export const useFetchWpComSites = ( connectedSiteIdsOnlyForSelectedSite: number[] ) => {
 	const [ rawSyncSites, setRawSyncSites ] = useState< unknown[] >( [] );
@@ -186,7 +186,7 @@ export const useFetchWpComSites = ( connectedSiteIdsOnlyForSelectedSite: number[
 
 			setRawSyncSites( parsedResponse.sites );
 
-			return parsedResponse.sites;
+			return syncSites;
 		} catch ( error ) {
 			Sentry.captureException( error );
 			console.error( error );
--- a/src/hooks/use-fetch-wpcom-sites/index.tsx
+++ b/src/hooks/use-fetch-wpcom-sites/index.tsx
@@ -49,7 +49,7 @@ function isJetpackSite( site: SitesEndpointSite ): boolean {
 }
 
 function hasSupportedPlan( site: SitesEndpointSite ): boolean {
-	return !! site.plan && site.plan.features.active.includes( STUDIO_SYNC_FEATURE_NAME );
+	return site.plan.features.active.includes( STUDIO_SYNC_FEATURE_NAME );
 }
 
 function getSyncSupport( site: SitesEndpointSite, connectedSiteIds: number[] ): SyncSupport {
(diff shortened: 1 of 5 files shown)


src/hooks/use-fetch-wpcom-sites/index.tsx
```diff
@@ -1,40 +1,47 @@
...
+	is_wpcom_atomic: z.boolean(),
+	is_wpcom_staging_site: z.boolean(),
+	name: z.string(),
+	URL: z.string(),
+	jetpack: z.boolean().optional(),
+	is_deleted: z.boolean(),
+	options: z.object( {
+		created_at: z.string(),
+		wpcom_staging_blog_ids: z.array( z.number() ),
+	} ),
```
nightnei, line 20: Hm, is it intentional to be not optional now?
```suggestion
	} ).optional(),
```

src/hooks/use-fetch-wpcom-sites/index.tsx
```diff
@@ -1,40 +1,47 @@
...
+		features: z.object( {
+			active: z.array( z.string() ),
+			available: z.record( z.string(), z.array( z.string() ) ).optional(),
+		} ),
+		is_free: z.boolean(),
+		product_id: z.number(),
+		product_name_short: z.string(),
+		product_slug: z.string(),
+		user_is_owner: z.boolean(),
+	} ),
```
nightnei, line 35: Also, shoudn't it be optional?
```suggestion
	} ).optional(),
```

src/hooks/use-fetch-wpcom-sites/index.tsx
```diff
@@ -1,40 +1,47 @@
...
+		wpcom_staging_blog_ids: z.array( z.number() ),
+	} ),
+	capabilities: z.object( {
+		manage_options: z.boolean(),
+	} ),
+	plan: z.object( {
+		expired: z.boolean(),
+		features: z.object( {
+			active: z.array( z.string() ),
+			available: z.record( z.string(), z.array( z.string() ) ).optional(),
```
nightnei, line 28: It wasn't not optional, shoudl it be?
```suggestion
			available: z.record( z.string(), z.array( z.string() ) ),
```

src/hooks/use-fetch-wpcom-sites/index.tsx
```diff
@@ -82,25 +92,29 @@ export function transformSingleSiteResponse(
...
 
-		return acc;
-	}, [] );
+				return transformSingleSiteResponse( site, getSyncSupport( site, connectedSiteIds ) );
+			} catch ( error ) {
+				Sentry.captureException( error );
+				return null;
+			}
+		} )
+		.filter( ( site ): site is SyncSite => site !== null );
```
nightnei: `site is SyncSite` - personally for me looks not very nice and looks more wordy, having extra step with filtering. I dont' have strong preference, but take a look one more time, maybe `reduce` would be more simple way here

nightnei (approved):
Left a few comments, but in general LGTM and works well 👍

<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2025-02-07 at 12 45 05Screenshot 2025-02-07 at 12 45 38
<td>Screenshot 2025-02-07 at 12 49 38Screenshot 2025-02-07 at 12 49 55
</tr>
</table>

src/hooks/use-fetch-wpcom-sites/index.tsx
fredrikekelund: It is indeed 👍 I looked closer at the API, and `options` should always be defined for us

src/hooks/use-fetch-wpcom-sites/index.tsx
fredrikekelund: You're right that it was optional previously, but after looking at the API code, I determined it's not.

src/hooks/use-fetch-wpcom-sites/index.tsx
fredrikekelund: Yep, apparently it's sometimes undefined for certain sites.

src/hooks/use-fetch-wpcom-sites/index.tsx
fredrikekelund: I typically find chained `filter` and `map` calls easier to read, but I also don't have a strong preference. I had to make additional tweaks to accommodate my changes from https://github.com/Automattic/studio/pull/897, but I've reverted the first call to a `.reduce`

PR #900, opened 2025-02-06
Generate always available name for a new website

nightnei:
## Proposed Changes
When you run out of available names, there is very weird result: after clicking "Add site" button we immediately see the next:
Screenshot 2025-02-06 at 12 08 34
Screenshot 2025-02-06 at 12 09 25

With this PR, the generation of names is always successful (we add numeric postfix), so users don't need to do extra actions and spend time to figure out "what is going on" .

## Testing Instructions
Apply the next diff:

```
diff --git a/src/lib/generate-site-name.ts b/src/lib/generate-site-name.ts
index 3052a2f..ae3e6ad 100644
--- a/src/lib/generate-site-name.ts
+++ b/src/lib/generate-site-name.ts
@@ -4,24 +4,6 @@ export function generateSiteName( usedSiteNames: string[] ): string {
        const siteNames = [
                __( 'My Bold Website' ),
                __( 'My Bright Website' ),
-               __( 'My Blissful Website' ),
-               __( 'My Calm Website' ),
-               __( 'My Cool Website' ),
-               __( 'My Dreamy Website' ),
-               __( 'My Elite Website' ),
-               __( 'My Fresh Website' ),
-               __( 'My Glowing Website' ),
-               __( 'My Happy Website' ),
-               __( 'My Joyful Website' ),
-               __( 'My Noble Website' ),
-               __( 'My Pure Website' ),
-               __( 'My Peak Website' ),
-               __( 'My Prime Website' ),
-               __( 'My Serene Website' ),
-               __( 'My Shiny Website' ),
-               __( 'My Sparkly Website' ),
-               __( 'My Swift Website' ),
-               __( 'My True Website' ),
        ];
 
        const defaultName = __( 'My WordPress Website' );
```

1. Remove all websites
2. Assert that default website has name `My WordPress Website` and create it
3. "Add new" website and assert that it has random name between `My Bold Website` and `My Bright Website`
4. Add one more website and assert that it has the second name which left from previous step
5. Add new website and assert that we have numeric postfix and further generations increase postfix correctly for random name

### DIFF
--- a/src/lib/generate-site-name.ts
+++ b/src/lib/generate-site-name.ts
@@ -35,7 +35,7 @@ export function generateSiteName( usedSiteNames: string[] ): string {
 		return availableNames[ Math.floor( Math.random() * availableNames.length ) ];
 	}
 
-	let siteNumber = 1;
+	let siteNumber = 2;
 	while ( usedSiteNames.includes( `${ defaultName } ${ siteNumber }` ) ) {
 		siteNumber++;
 	}
--- a/src/lib/generate-site-name.ts
+++ b/src/lib/generate-site-name.ts
@@ -46,7 +46,7 @@ export const sanitizeFolderName = ( filename: string ) => {
 	const LATIN = 'a-z';
 	const CYRILLIC = 'а-яё';
 	const ARABIC = '\\u0600-\\u06FF';
-	const HEREW = '\\u0590-\\u05FF';
+	const HEBREW = '\\u0590-\\u05FF';
 	const CHINESE = '\\u4e00-\\u9fa5';
 	const JAPANESE_HIRAGANA = '\\u3040-\\u309F';
 	const JAPANESE_KATAKANA = '\\u30A0-\\u30FF';
--- a/src/lib/generate-site-name.ts
+++ b/src/lib/generate-site-name.ts
@@ -23,22 +23,30 @@ export function generateSiteName( usedSiteNames: string[] ): string {
 		__( 'My Swift Website' ),
 		__( 'My True Website' ),
 	];
-	let proposedName = __( 'My WordPress Website' );
-	let tryCount = 0;
 
-	while ( usedSiteNames.includes( proposedName ) && tryCount < siteNames.length ) {
-		tryCount++;
-		proposedName = siteNames[ Math.floor( Math.random() * siteNames.length ) ];
+	const defaultName = __( 'My WordPress Website' );
+	if ( ! usedSiteNames.includes( defaultName ) ) {
+		return defaultName;
 	}
 
-	return proposedName;
+	const availableNames = siteNames.filter( ( name ) => ! usedSiteNames.includes( name ) );
+	if ( availableNames.length > 0 ) {
+		return availableNames[ Math.floor( Math.random() * availableNames.length ) ];
+	}
+
+	const randomName = siteNames[ Math.floor( Math.random() * siteNames.length ) ];
+	let postfix = 1;
+	while ( usedSiteNames.includes( `${ randomName } ${ postfix }` ) ) {
+		postfix++;
+	}
+	return `${ randomName } ${ postfix }`;
 }
 
 export const sanitizeFolderName = ( filename: string ) => {
 	const LATIN = 'a-z';
 	const CYRILLIC = 'а-яё';
 	const ARABIC = '\\u0600-\\u06FF';
-	const HEBREW = '\\u0590-\\u05FF';
+	const HEREW = '\\u0590-\\u05FF';
 	const CHINESE = '\\u4e00-\\u9fa5';
 	const JAPANESE_HIRAGANA = '\\u3040-\\u309F';
 	const JAPANESE_KATAKANA = '\\u30A0-\\u30FF';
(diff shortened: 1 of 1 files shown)


src/lib/generate-site-name.ts
```diff
@@ -23,22 +23,30 @@ export function generateSiteName( usedSiteNames: string[] ): string {
...
+	}
+	return `${ randomName } ${ postfix }`;
 }
 
 export const sanitizeFolderName = ( filename: string ) => {
 	const LATIN = 'a-z';
 	const CYRILLIC = 'а-яё';
 	const ARABIC = '\\u0600-\\u06FF';
-	const HEBREW = '\\u0590-\\u05FF';
+	const HEREW = '\\u0590-\\u05FF';
```
sejas: I think this change is not intentional.

src/lib/generate-site-name.ts
```diff
@@ -23,22 +23,30 @@ export function generateSiteName( usedSiteNames: string[] ): string {
...
+	if ( availableNames.length > 0 ) {
+		return availableNames[ Math.floor( Math.random() * availableNames.length ) ];
+	}
+
+	const randomName = siteNames[ Math.floor( Math.random() * siteNames.length ) ];
+	let postfix = 1;
+	while ( usedSiteNames.includes( `${ randomName } ${ postfix }` ) ) {
+		postfix++;
+	}
+	return `${ randomName } ${ postfix }`;
```
sejas: Should we take into account RTL languages ? For those languages the postfix should be on the left.

sejas (commented):
Left a couple of suggestions.
I think adding a postfix is an improvement 💪 

In addition to this, we could increase our list of default sitenames by having a really long list of adjetives produced by an LLM and then use something like `sprintf( __( 'My %s Website' ), adjetive )`,

src/lib/generate-site-name.ts
wojtekn: Would it make sense to always use `My WordPress Website` for subsequent names with the number, so user who uses all random names will keep getting `My WordPress Website 1`, `My WordPress Website 2` and so on?

src/lib/generate-site-name.ts
```diff
@@ -23,22 +23,30 @@ export function generateSiteName( usedSiteNames: string[] ): string {
...
 	}
 
-	return proposedName;
+	const availableNames = siteNames.filter( ( name ) => ! usedSiteNames.includes( name ) );
+	if ( availableNames.length > 0 ) {
+		return availableNames[ Math.floor( Math.random() * availableNames.length ) ];
+	}
+
+	const randomName = siteNames[ Math.floor( Math.random() * siteNames.length ) ];
+	let postfix = 1;
```
wojtekn: Should it be `suffix` instead of `postfix`? Or just a `siteNumber` to ensure it makes sense for LTR and RTL?

src/lib/generate-site-name.ts
nightnei: Personally for me, it's one more good option. And I don't see any pros/cons, anyway it's more kinda "Temporary name", 99% that the user will change it to something connected with their website, unless it's testing site.
Only one good thing what I see with the current approach is that we have diversity of names 😅
So I would go with either proposed approach by me or remove all adjectives and keep only `My WordPress Website ${number}`. But I am ok with any approach, as long as we provide unique names.

nightnei: UPDATE, Oh, I guess you meant exactly to use *only* subsequent names of `My WordPress Website` and remove adjective options, right? If so, ignore my previous message and I think that it's good simple approach too, I don't have strong preference either to use it or keep proposed by me.

src/lib/generate-site-name.ts
nightnei: > Should it be suffix instead of postfix? Or just a siteNumber to ensure it makes sense for LTR and RTL?

If we make it as suffix, then it will be confusing for LTR environments, since we used to see number at the end (e.g. in MacOS, Windows).
So my opinion, we should change postfix with suffix only for RTL.
WDYT?
Screenshot 2025-02-10 at 11 49 22

src/lib/generate-site-name.ts
wojtekn: Yes, this is what I meant. So the order would look as follows:
- `My WordPress Website`
- One of `My RANDOM Website` names
- `My WordPress Website 1`
- `My WordPress Website 2`
This seems to be the least confusing to me.

src/lib/generate-site-name.ts
wojtekn: My comment referred only to the variable name. `postfix` is the name of the mail-sending agent. I think it should be `affix`, instead, or even better, just `siteNumber`.

src/lib/generate-site-name.ts
nightnei: Done

wojtekn (approved):
Works as expected. I'm just unsure if the first number should be 1 or 2, because in the end, "My WordPress Website 1" will be second site of that name. Not a big deal, though, as it's edge case.

fredrikekelund:
>  I'm just unsure if the first number should be 1 or 2, because in the end, "My WordPress Website 1" will be second site of that name

I haven't reviewed the code here, but the commonsensical approach would be to start the counter from 2, IMO.

PR #902, opened 2025-02-06
Attempt to fix osascript issue with launching Terminal

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10416

## Proposed Changes

There are numerous reports in Sentry about `osascript` failing to run the command used to launch a Terminal from Studio. It's not clear why this happens, but in this PR, I've taken a few steps to make the script safer:

1. Escape the site folder path by wrapping it in quotes (would help if there's whitespace in the path)
2. Use `;` instead of `&&` to separate statements in the init script (proceed even if the previous step failed)
3. Use STDIN to pipe the script to `osascript` instead of passing it as an argument to `osascript -e`. Why? Because the manpage for `osascript` mentions that the `-e` argument technically only accepts single-line scripts:

```
     -e statement
           Enter one line of a script.  If -e is given, osascript will not look for a filename in the argument list.  Multiple -e options
           may be given to build up a multi-line script.  Because most scripts use characters that are special to many shell programs (for
           example, AppleScript uses single and double quote marks, “(”, “)”, and “*”), the statement will have to be correctly quoted and
           escaped to get it past the shell intact.
```

## Testing Instructions

1. Ensure that the `Open in… Terminal` button still works as expected on macOS

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -908,15 +908,25 @@ export function openTerminalAtPath(
 			env: { ...process.env, ...env },
 		} );
 	} else if ( platform === 'darwin' ) {
-		const loadWpCliCommand = `clear && export PATH=\\"${ cliPath }\\":$PATH && export STUDIO_APP_PATH=\\"${ appPath }\\" &&`;
+		const initScriptSteps = [];
+
+		if ( wpCliEnabled ) {
+			initScriptSteps.push(
+				`export PATH=\\"${ cliPath }\\":$PATH`,
+				`export STUDIO_APP_PATH=\\"${ appPath }\\"`
+			);
+		}
+
+		const escapedPath = targetPath.replace( /"/g, '\\"' );
+		initScriptSteps.push( `cd \\"${ escapedPath }\\"`, 'clear' );
+
 		const osascript = `
 		tell application "Terminal"
-			if not application "Terminal" is running then launch
-			do script "${ wpCliEnabled ? loadWpCliCommand : '' } cd ${ targetPath } && clear"
+			do script "${ initScriptSteps.join( ' ; ' ) }"
 			activate
 		end tell`;
 
-		return promiseExec( `osascript -e '${ osascript }'` );
+		return promiseExec( `osascript << END '${ osascript }' END` );
 	} else if ( platform === 'linux' ) {
 		if ( wpCliEnabled ) {
 			return promiseExec(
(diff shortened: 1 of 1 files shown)


src/ipc-handlers.ts
```diff
@@ -908,15 +908,25 @@ export function openTerminalAtPath(
...
+
+		if ( wpCliEnabled ) {
+			initScriptSteps.push(
+				`export PATH=\\"${ cliPath }\\":$PATH`,
+				`export STUDIO_APP_PATH=\\"${ appPath }\\"`
+			);
+		}
+
+		const escapedPath = targetPath.replace( /"/g, '\\"' );
+		initScriptSteps.push( `cd \\"${ escapedPath }\\"`, 'clear' );
```
nightnei, line 912: Hm, why do we do cleaning at the end?

nightnei (changes requested):
It fails for me
Screenshot 2025-02-07 at 14 01 19

fredrikekelund:
> It fails for me

🤦‍♂️ Looks like I failed to format the multiline string properly. Should be fixed now

src/ipc-handlers.ts
wojtekn: We have a separate issue to remove that https://github.com/Automattic/dotcom-forge/issues/10116

wojtekn (approved):
It works well for me. However, it worked well, before, too.

fredrikekelund:
> It works well for me. However, it worked well, before, too.

Same here. I haven't been able to reproduce the issue. There's no guarantee that this will fix it, but escaping the `cd` path is definitely an improvement, and passing the script in STDIN could _maybe_ help.

src/ipc-handlers.ts
fredrikekelund: > why do we do cleaning at the end?

That's what I would expect as a user. Seeing setup commands in the terminal feels like exposing the app's inner workings.

NB that the `osascript` is only needed for running the experimental WP CLI code. If we only want to open the terminal in the site directory, we could follow [GitHub Desktop's implementation](https://github.com/desktop/desktop/blob/6ce9e221854c6a6f47c29286db540bec05ebe9c6/app/src/lib/shells/darwin.ts#L205)

nightnei (approved):
Tested, works well 👍

PR #903, opened 2025-02-06
fix dev versions

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10409

The reported issue is caused by an error while parsing the commit hash in the dev builds.
This PR suggest a change that makes dev builds follow a correct versioning convention by using `-dev${number}` format.
We currently already follow that convention on other builds, for example, `-beta1`, `-beta2`, etc...
By using the commit count since the latest release we maintain the ability to track which commit generated that version.

**Notes:**
- The approval of this PR requires changes in the `studio-app/updates` endpoint.
- After releasing a version (e.g., `1.3.3`), we should immediately update `package.json` to the next intended version (e.g., `1.3.4`). This ensures dev builds are correctly labeled with the version they're working towards (e.g., `1.3.4-dev1`) rather than appearing as development builds of the just-released version.

## Proposed Changes

- Changed development version format from -dev.HASH to -devN (where N is commit count since last release)
- Updated version counting to use latest release tag (excluding pre-release tags) instead of latest tag
- Added backward compatibility to handle both -dev.HASH and -devN formats
- Updated version handling in build scripts and runtime checks
- Modified artifact naming to use new version format

New dev version on macOS:
image

New dev version on Windows:
image

## Testing Instructions

1. Test the version generation:
```bash
node scripts/prepare-dev-build-version.mjs
```
2. Check the generated version in package.json:
```bash
cat package.json | grep version
```

3. Test the releases manifest generation:
```bash
IS_DEV_BUILD=true node scripts/generate-releases-manifest.mjs
```

4. Check the releases manifest
```bash
cat out/releases.json | jq
```

5. Test the fastlane distribution (in dry-run mode):
```bash
DRY_RUN=true BUILDKITE_BUILD_NUMBER=test-build bundle exec fastlane distribute_dev_build
```

## Pre-merge Checklist

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -113,6 +113,18 @@ def aws_upload_to_s3(bucket:, key:, file:, if_exists:, auto_prefix: false, dry_r
   )
 end
 
+def get_commit_count
+  # Get the most recent tag
+  latest_tag = `git describe --tags --abbrev=0`.strip rescue ''
+  
+  # Get commit count since the last tag
+  if latest_tag.empty?
+    `git rev-list --count HEAD`.strip.to_i
+  else
+    `git rev-list #{latest_tag}..HEAD --count`.strip.to_i
+  end
+end
+
 def distribute_builds(
   commit_hash: last_git_commit[:abbreviated_commit_hash],
   build_number: get_required_env('BUILDKITE_BUILD_NUMBER'),
@@ -121,7 +133,8 @@ def distribute_builds(
   UI.message("Running in #{DRY_RUN ? 'DRY RUN' : 'NORMAL'} mode")
   # If we are distributing a build without a tag, i.e. a development build, we also want to update the latest build reference for distribution.
   update_latest = release_tag.nil?
-  suffix = release_tag.nil? ? "v#{PACKAGE_VERSION}-#{commit_hash}" : release_tag
+  commit_count = get_commit_count
+  suffix = release_tag.nil? ? "#{PACKAGE_VERSION.split('-')[0]}-dev#{commit_count}" : release_tag
   filename_root = 'studio'
   bucket_folder = 'studio'
 
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -114,8 +114,9 @@ def aws_upload_to_s3(bucket:, key:, file:, if_exists:, auto_prefix: false, dry_r
 end
 
 def get_commit_count
-  # Get the most recent tag
-  latest_tag = `git describe --tags --abbrev=0`.strip rescue ''
+  # Get the most recent release tag (excluding pre-release tags)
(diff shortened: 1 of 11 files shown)


src/updates.ts
```diff
@@ -18,7 +18,9 @@ let timeout: NodeJS.Timeout | null = null;
 let showManualCheckDialogs = false;
 
 const shouldPoll =
-	process.env.NODE_ENV === 'production' && app.isPackaged && ! app.getVersion().includes( '-dev.' );
+	process.env.NODE_ENV === 'production' &&
+	app.isPackaged &&
+	! /(-dev\.|-dev\d+)/.test( app.getVersion() );
```
fredrikekelund: Not entirely necessary, but it might be nice to add an `isDevRelease` utility function for checking a version string. Could be used here and in `src/lib/sentry-release.ts`

fastlane/Fastfile
```diff
@@ -54,19 +54,43 @@ end
...
+    puts "\nAttempting to notarize DMG at: #{path}"
     notarize(
       bundle_id: APPLE_BUNDLE_IDENTIFIER,
       package: path,
       api_key_path: APPLE_API_KEY_PATH
     )
+    puts "Notarization completed for DMG"
   end
+  
+  puts "\n=== Notarization process completed ==="
```
fredrikekelund: Do we need these changes? Seems like it's all just logging calls

fredrikekelund (approved):
LGTM 👍 

To quickly recap why we need to do make this change: while the Mac installer will simply install whatever app our API returns, the Windows Squirrel installer compares the installed version with the new version provided by the API. The new bundle is only installed if the provided version is higher according to some NuGet logic (presumably covered by [these docs](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort)). Version numbers derived from git hashes cannot be reliably compared, whereas this approach can.

The one thing I take issue with (I've already discussed this with @bcotrim), is this part:

> After releasing a version (e.g., 1.3.3), we should immediately update package.json to the next intended version (e.g., 1.3.4). This ensures dev builds are correctly labeled with the version they're working towards (e.g., 1.3.4-dev1) rather than appearing as development builds of the just-released version.

Committing a new version number in `package.json` currently triggers a release build on CI. Changing that seems overly cumbersome for this purpose, so I suggest we accept the quirk that `v1.3.3-dev24` is **newer** than `v1.3.3`. This is in fact a "problem" we already have with our existing dev version format, so we can tackle that separately (if needed)

src/updates.ts
bcotrim: Makes a lot of sense, added! 👍

fastlane/Fastfile
bcotrim: Thanks for catching this. I was getting some errors and added those lines as debug.
I cleaned up now.

Issue #544, opened 2024-09-17
Database import from .sql file failes, but process keeps running in the background


countfak:
### Quick summary

Hi, I wanted to share my experience with database import from a .sql file.
I tried multiple times with different database sizes, last one was around 40MB.
The import goes on for a while, then the import failed message appears. But the import process is still running in the background.

### Steps to reproduce

Drop a large .sql file with the built-in import functionality of an existing site and start the import. It will likely fail, open up the Activity Monitor app and find the Studio process at the top with high CPU usage, meaning that the import is still running in the background. After a while the CPU usage of the process drops, and the import is actually finished.

### What you expected to happen

I wasn't expecting to get the error message if the import process was still running in the background.

### What actually happened

Got import failed message even if the import was still running, see above.

### Impact

One

### Available workarounds?

No but the app is still usable

### Platform

Mac Silicon

wojtekn:
Thanks for reporting the issue @countfak . We will check it.

Robertght:
Marking this as Triaged but let me know @wojtekn or @countfak if you have a large DB file I could use to test this.

countfak:
Just wanted to follow up on this and share my latest successful attempt.

What I tried and didn't work out:

- import sql file via the built-in import functionality: this causes the issue explained above, not even sure the import finishes and is done correctly, it just takes a lot of time
- import using all in one wp migration plugin with a .wpress file: this fails with error 400 bad request in the javascript console
- import via command line with wp cli sqlite (https://github.com/Automattic/wp-cli-sqlite-command): this takes a lot of time too, I didn't wait for the import to finish, probably same as above

What worked for me was:
- create new site in Studio
- export sql file from the website that I want to import into Studio
- replace the url in the database if needed
- convert sql to sqlite with this: https://github.com/mysql2sqlite/mysql2sqlite
- move the new database to /wp-content/database/.ht.sqlite (replace existing)

Please note that this only applies in case of large databases with a lot of entries (for reference mine was around 40MB), the built-in import works fine if I try with a smaller database.

wojtekn:
Thanks for sharing your workaround @countfak, it's useful. We will check how we can improve Studio Importer to support databases of any size.

katinthehatsite:
Hi @countfak and thank you for reporting the issue! In the upcoming Studio release, we have introduced the following:

* the database import process should be terminated if it fails and it won't be running in the background (this could be tested with `Activity Monitor` once the new version of the app is released)
[...]

PR #904, opened 2025-02-06
Set higher timeouts for import/export wp-cli commands

ashfame:
## Related issues

- https://github.com/Automattic/studio/issues/544

## Proposed Changes

- Default timeout of 5mins for waiting on wp-cli command to finish is too small for big imports. In this PR, I have added a higher timeout for import/export commands, which gets used by passing an arg `longRunning: true` to `executeWpCliCommand()`
- Show a custom error message when it does timeout (currently set to 24hrs)

Screenshot 2025-02-06 at 22 13 40

## Testing Instructions

- I tested it by lowering the default timeout to small like 5 secs and process a normal backup file, it fails as per the linked issue
- Then on this branch, keep the default timeout lowered and see it uses the higher timeout set for import/export and finishes up.

## Pre-merge Checklist


src/hooks/use-import-export.tsx
```diff
@@ -92,6 +92,13 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
 							'The ZIP archive is invalid. Try to unpack and pack it again. If this problem persists, please contact support.'
 						),
 					} );
+				} else if (
+					( error as Error ).message.includes( 'WP-CLI command was canceled (timed out)' )
+				) {
+					await getIpcApi().showErrorMessageBox( {
+						title: __( 'Failed importing site' ),
+						message: __( 'Import process timed out. Very large import?' ),
```
ashfame: What should this message say?

src/lib/wp-cli-process.ts
```diff
@@ -11,7 +11,8 @@ export type MessageName = 'execute';
 export type WpCliResult = ReturnType< typeof executeWPCli >;
 export type MessageCanceled = { error: Error; canceled: boolean };
 
-const DEFAULT_RESPONSE_TIMEOUT = 300 * 1000;
+const DEFAULT_RESPONSE_TIMEOUT = 5 * 60 * 1000; // 5min
+const IMPORT_EXPORT_RESPONSE_TIMEOUT = 24 * 60 * 60 * 1000; // 24hr
```
ashfame: What should be the timeout we set here?

src/hooks/use-import-export.tsx
sejas: Here is my suggestion: `The import process timed out after %d hours, which can occur when processing very large imports. If the issue persists, please contact support.')`

src/lib/wp-cli-process.ts
sejas: What if we start with fewer hours? I think 6 hours should be enough for most imports, if not maybe we could remove the timeout from imports. Usually 0 represents infinite in these context.
I wonder if these constants could live in a separate config file that the user can tune for their specific needs.

src/lib/wp-cli-process.ts
```diff
@@ -54,15 +55,25 @@ export default class WpCliProcess {
 
 	async execute(
 		args: string[],
-		{ phpVersion }: { phpVersion?: string } = {}
+		{
+			phpVersion,
+			longRunning = false,
+		}: {
+			phpVersion?: string;
+			longRunning?: boolean;
```
sejas: What do you think if we pass the exact timeout to wait as a parameter? That way we could be more precise with the different type of executions.
Instead of having a boolean longRunning, we could have an integer with the timeout ms. By default it would be DEFAULT_RESPONSE_TIMEOUT.

src/lib/import-export/import/importers/importer.ts
```diff
@@ -57,15 +57,15 @@ abstract class BaseImporter extends EventEmitter implements Importer {
 				const { stderr, exitCode } = await server.executeWpCliCommand(
 					`sqlite import ${ sqlTempFile } --require=/tmp/sqlite-command/command.php`,
 					// SQLite plugin requires PHP 8+
-					{ targetPhpVersion: DEFAULT_PHP_VERSION, skipPluginsAndThemes: true }
+					{ targetPhpVersion: DEFAULT_PHP_VERSION, skipPluginsAndThemes: true, longRunning: true }
 				);
 
 				if ( stderr ) {
-					console.error( `Warning during import of ${ sqlFile }:`, stderr );
+					console.error( `Error during import of ${ sqlFile }:`, stderr );
```
sejas, line 64: 👍

sejas (commented):
The changes look good. Increasing the import process will allow users to import websites with more content.

To establish reasonable limits, we might need to benchmark the amount of content imported on a typical computer and set a limit based on those metrics.

I’ve added a couple of suggestions, please let me know what you think.

src/lib/wp-cli-process.ts
ashfame: I agree that removing timeout from import/export makes sense. But since we don't have any way of tracking the progress of big imports while they run or no ability to halt the import without waiting further, let's go with 6hrs as the timeout for now.

src/lib/wp-cli-process.ts
ashfame: Yep, that's probably better. But I am thinking if we should just infer the timeout based on the command we are running. So import/export command has higher timeout on their own without having to explicitly call it that way.

src/hooks/use-import-export.tsx
```diff
@@ -92,6 +92,18 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
...
+				} else if (
+					( error as Error ).message.includes( 'WP-CLI command was canceled (timed out)' )
+				) {
+					await getIpcApi().showErrorMessageBox( {
+						title: __( 'Failed importing site' ),
+						message: sprintf(
+							__(
+								'The import process timed out after %d hours, which can occur when processing very large imports. If the issue persists, please contact support.'
+							),
+							6
```
ashfame: @sejas Importing the constant from `wp-cli-process.ts` was breaking the electron app in a weird way, complaining about webpack 5 and I couldn't figure out why. So, for now, I am hardcoding it and we should consider having a constants file like you mentioned.

src/lib/wp-cli-process.ts
ashfame: @sejas I removed the additional arg I added and now it uses the higher timeout when it detects a sqlite import/export command. Please re-review the PR.

src/lib/wp-cli-process.ts
```diff
@@ -62,7 +63,11 @@ export default class WpCliProcess {
 			args,
 			phpVersion,
 		} );
-		return await this.waitForResponse( message, messageId );
+		const timeout =
+			args[ 0 ] === 'sqlite' && ( args[ 1 ] === 'import' || args[ 1 ] === 'export' )
```
sejas: Another slightly more flexible alternative:

```suggestion
			args[ 0 ] === 'sqlite' && [ 'import', 'export' ].includes( args[ 1 ] )
```

src/hooks/use-import-export.tsx
sejas: We could move the constants into our existing `src/constants.ts` file to ensure that the same number of hours used in the code is also displayed in the alert.

sejas (approved):
@ashfame, Thanks for considering my suggestions. I like the new approach where wp-cli-process handles selecting the timeout based on the command.

I've tested it and confirmed that the timeout for SQLite export and import processes is 6 hours, while the rest of wp-cli commands still 5 minutes.

Screenshot 2025-02-17 at 15 35 46

I've left a couple of non-blocking comments.

ashfame:
@sejas Thanks for the review! I made both the changes. Will merge shortly 👍

PR #906, opened 2025-02-07
Log backup file size and extension when no suitable backup handler exception occurs

ashfame:
As per p1738839275131699-slack-C06DRMD6VPZ


nightnei (approved):
LGTM and works well 👍
Screenshot 2025-02-07 at 11 13 39

PR #919, opened 2025-02-11
Preview Sites: Display full date in the tooltip over relative site expiration time

ivan-ottinger:
## Related issues

- Resolves https://github.com/Automattic/dotcom-forge/issues/10423.

## Proposed Changes

- add full date format to the tooltip displayed over the preview site expiration time
- widen columns to add more space for the relative date values

| Before | After |
|--------|--------|
| Markup on 2025-02-12 at 14:03:34 | Markup on 2025-02-12 at 13:58:01 | 

## Testing Instructions

1. Check out the PR branch and build the app with `STUDIO_QUICK_DEPLOYS=true npm start`.
2. If you don't have any preview sites, create a couple of them.
3. When you hover over values in the `Expires` column, a tooltip with full expiration date should display (as can be seen in the screenshot above).
4. Columns should have more space than before.
5. You can test with different locales.

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-expiration-date.ts
+++ b/src/hooks/use-expiration-date.ts
@@ -2,15 +2,31 @@ import { useI18n } from '@wordpress/react-i18n';
 import { intervalToDuration, formatDuration, addDays, Duration, addHours } from 'date-fns';
 import { HOUR_MS, DAY_MS } from 'src/constants';
 import { useI18nData } from 'src/hooks/use-i18n-data';
-import { formatDistanceShort } from 'src/lib/date';
+import { formatDistance } from 'src/lib/date';
 import { SupportedLocale } from 'src/lib/locale';
 
-function formatStringDate( ms: number, locale: SupportedLocale ): string {
-	const formatter = new Intl.DateTimeFormat( locale, {
-		day: 'numeric',
-		month: 'long',
-		year: 'numeric',
-	} );
+function formatStringDate(
+	ms: number,
+	locale: SupportedLocale,
+	format: 'long' | 'short' = 'short'
+): string {
+	const options: Intl.DateTimeFormatOptions =
+		format === 'short'
+			? {
+					day: 'numeric',
+					month: 'long',
+					year: 'numeric',
+			  }
+			: {
+					day: '2-digit',
+					month: 'short',
+					year: 'numeric',
+					hour: '2-digit',
+					minute: '2-digit',
+					second: '2-digit',
+					hour12: false,
+			  };
+	const formatter = new Intl.DateTimeFormat( locale, options );
 	return formatter.format( new Date( ms ) );
 }
 
@@ -40,13 +56,14 @@ export function useExpirationDate( snapshotDate: number ) {
 		{
 			format,
 			delimiter: ', ',
-			locale: { formatDistance: formatDistanceShort },
+			locale: { formatDistance },
 		}
 	);
 
 	return {
 		isExpired,
 		countDown: isExpired ? __( 'Expired' ) : countDown,
+		fullDateString: formatStringDate( endDate.getTime(), locale, 'long' ),
 		dateString: formatStringDate( snapshotDate, locale ),
 	};
 }
--- a/src/hooks/use-expiration-date.ts
+++ b/src/hooks/use-expiration-date.ts
@@ -5,27 +5,29 @@ import { useI18nData } from 'src/hooks/use-i18n-data';
 import { formatDistance } from 'src/lib/date';
 import { SupportedLocale } from 'src/lib/locale';
 
+type FormatKey = 'short' | 'long';
+
 function formatStringDate(
 	ms: number,
 	locale: SupportedLocale,
-	format: 'long' | 'short' = 'short'
+	format: FormatKey = 'short'
 ): string {
-	const options: Intl.DateTimeFormatOptions =
-		format === 'short'
-			? {
-					day: 'numeric',
-					month: 'long',
-					year: 'numeric',
-			  }
-			: {
-					day: '2-digit',
-					month: 'short',
-					year: 'numeric',
-					hour: '2-digit',
-					minute: '2-digit',
-					hour12: false,
-			  };
-	const formatter = new Intl.DateTimeFormat( locale, options );
+	const options: Record< FormatKey, Intl.DateTimeFormatOptions > = {
+		short: {
+			day: 'numeric',
+			month: 'long',
+			year: 'numeric',
+		},
+		long: {
+			day: '2-digit',
+			month: 'short',
+			year: 'numeric',
+			hour: '2-digit',
+			minute: '2-digit',
+			hour12: false,
+		},
+	};
+	const formatter = new Intl.DateTimeFormat( locale, options[ format ] );
 	return formatter.format( new Date( ms ) );
 }
 
--- a/src/hooks/use-expiration-date.ts
+++ b/src/hooks/use-expiration-date.ts
@@ -23,7 +23,6 @@ function formatStringDate(
 					year: 'numeric',
 					hour: '2-digit',
 					minute: '2-digit',
-					second: '2-digit',
 					hour12: false,
 			  };
 	const formatter = new Intl.DateTimeFormat( locale, options );
--- a/src/hooks/use-expiration-date.ts
+++ b/src/hooks/use-expiration-date.ts
@@ -64,7 +64,7 @@ export function useExpirationDate( snapshotDate: number ) {
 	return {
 		isExpired,
 		countDown: isExpired ? __( 'Expired' ) : countDown,
-		fullDateString: formatStringDate( endDate.getTime(), locale, 'long' ),
+		expireDateString: formatStringDate( endDate.getTime(), locale, 'long' ),
 		dateString: formatStringDate( snapshotDate, locale ),
 	};
 }
--- a/src/hooks/use-expiration-date.ts
+++ b/src/hooks/use-expiration-date.ts
@@ -2,7 +2,7 @@ import { useI18n } from '@wordpress/react-i18n';
 import { intervalToDuration, formatDuration, addDays, Duration, addHours } from 'date-fns';
 import { HOUR_MS, DAY_MS } from 'src/constants';
 import { useI18nData } from 'src/hooks/use-i18n-data';
-import { formatDistance } from 'src/lib/date';
+import { formatDistanceShort } from 'src/lib/date';
 import { SupportedLocale } from 'src/lib/locale';
 
 function formatStringDate( ms: number, locale: SupportedLocale ): string {
@@ -40,7 +40,7 @@ export function useExpirationDate( snapshotDate: number ) {
 		{
 			format,
 			delimiter: ', ',
-			locale: { formatDistance },
+			locale: { formatDistance: formatDistanceShort },
 		}
 	);
 
--- a/src/hooks/use-format-localized-timestamps.tsx
+++ b/src/hooks/use-format-localized-timestamps.tsx
@@ -1,36 +1,32 @@
 import { Duration, intervalToDuration } from 'date-fns';
 import { useCallback } from 'react';
-import { formatDistance, formatDistanceShort, TimeFormat } from 'src/lib/date';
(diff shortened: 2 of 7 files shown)


src/lib/date.ts
```diff
@@ -24,3 +26,27 @@ export const formatDistance: FormatDistanceFn = ( token, count ) => {
...
+	switch ( token ) {
+		case 'xDays':
+			stringToFormat = _n( '%d d', '%d d', count );
+			break;
+		case 'xHours':
+			stringToFormat = _n( '%d h', '%d h', count );
+			break;
+		case 'xMinutes':
+			stringToFormat = _n( '%d min', '%d min', count );
+			break;
```
ivan-ottinger: I left pluralization on purpose here - in case some languages need it.

src/hooks/use-format-localized-timestamps.tsx
```diff
@@ -1,32 +1,36 @@
...
+				start: new Date( timestamp ),
+				end: new Date(),
+			} );
 
-		return formatTimeDistance( duration );
-	}, [] );
+			return formatTimeDistance( duration, format );
+		},
+		[]
+	);
```
fredrikekelund: Would it work to use `formatDistanceToNow` from `date-fns` here? See [docs](https://date-fns.org/v4.1.0/docs/formatDistanceToNow#). Seems like a potential simplification

bcotrim (commented):
The code looks good and works as expected—great job!
My only suggestion is about the format. Displaying only the most significant value (e.g., `6 days` or `20 hours` if it's under a day) in the UI and adding a tooltip for the full value could make it cleaner.

sejas:
I like @bcotrim's suggestion. I think displaying the most significant value makes sense, and the tooltip is also a good suggestion.
Maybe inside the tooltip we could display the final date. Something like `Feb 12 2025 10:47:33` or any similar format.

What do you think @ivan-ottinger ?

wojtekn:
The short format goes well in English, but I'm unsure if everyone will well understand it. Also, it may not be that clear in other languages. This looked clearer in previous UI:

Screenshot 2025-02-12 at 12 00 23

wojtekn:
What if we increased those columns' width, especially since they are too wide to fit heading text in some languages? And in worst case, added a deliberate line break after days, to avoid having "23" in one line and "hours" in another one?

ivan-ottinger:
> I like @bcotrim's suggestion. I think displaying the most significant value makes sense, and the tooltip is also a good suggestion.
> Maybe inside the tooltip we could display the final date. Something like Feb 12 2025 10:47:33 or any similar format.

> What do you think @ivan-ottinger ?

I have been experimenting with this in the past hour and it could be something like this for instance (I have also added word `in` there, e.g. `in 5 days`):

Markup on 2025-02-12 at 12:56:31

But we could also use a full date in the tooltip as you are suggesting. 🙂 

> What if we increased those columns' width, especially since they are too wide to fit heading text in some languages? And in worst case, added a deliberate line break after days, to avoid having "23" in one line and "hours" in another one?

This also sounds good to me. It is probably the simplest approach. Full date could still be displayed in the tooltip if we wanted.

wojtekn:
@ivan-ottinger yes, displaying the full date in the tooltip makes sense in every case. I still have concerns about using h, m, d. They don't go well with other languages, e.g., "2d 3h 5s" would be "2 dni 3 godz. 5 sek." in Polish.

ivan-ottinger:
> @ivan-ottinger yes, displaying the full date in the tooltip makes sense in every case. I still have concerns about using h, m, d. They don't go well with other languages, e.g., "2d 3h 5s" would be "2 dni 3 godz. 5 sek." in Polish.

In that case we could use the full format in the field, e.g. `4 days, 2 hours`, maybe also including `in` as well: `in 4 days, 2 hours` and make the column wider.

Then, we could also add the full date into the tooltip.

What do you think?

wojtekn:
@ivan-ottinger displaying "4 days, 2 hours" in line and the exact date in tooltip makes sense. The "in" part can be handled as a part of the table heading. For example, in English, "Expires" makes sense alone, but in Polish, I would translate it as "Expires in" (e.g., "Wygasa za").

ivan-ottinger:
> @ivan-ottinger displaying "4 days, 2 hours" in line and the exact date in tooltip makes sense. The "in" part can be handled as a part of the table heading. For example, in English, "Expires" makes sense alone, but in Polish, I would translate it as "Expires in" (e.g., "Wygasa za").

Right! Makes sense. 👍🏼 

When it comes to the exact date, would we like to use the format suggested by Antonio in his comment above (`Feb 12 2025 10:47:33`)? We could keep it as is or we could translate it as well.

ivan-ottinger:
What about formatting like this? 🙂 

English:
Markup on 2025-02-12 at 13:52:53

Polish:
Markup on 2025-02-12 at 13:51:03

src/lib/date.ts
ivan-ottinger: This is not relevant anymore. → Closing this thread.

src/hooks/use-format-localized-timestamps.tsx
ivan-ottinger: Thank you, Fredrik! Yeah, I think this is something we could consider in the future. I am not that familiar with `date-fns` yet, but looks like we could definitely make some improvements and simplifications in the way we calculate relative time in Studio.

Since we decided on a more straightforward solution to address the issue at hand, I did not explore `formatDistanceToNow` further today. I have created a dedicated task: https://github.com/Automattic/dotcom-forge/issues/10449 though.

src/hooks/use-expiration-date.ts
```diff
@@ -47,6 +63,7 @@ export function useExpirationDate( snapshotDate: number ) {
 	return {
 		isExpired,
 		countDown: isExpired ? __( 'Expired' ) : countDown,
+		fullDateString: formatStringDate( endDate.getTime(), locale, 'long' ),
```
sejas: Should we call it `expireDateString` or `fullExpireDateString` or `endDateString` ?

```suggestion
		expireDateString: formatStringDate( endDate.getTime(), locale, 'long' ),
```

src/hooks/use-expiration-date.ts
```diff
@@ -5,12 +5,28 @@ import { useI18nData } from 'src/hooks/use-i18n-data';
...
+			: {
+					day: '2-digit',
+					month: 'short',
+					year: 'numeric',
+					hour: '2-digit',
+					minute: '2-digit',
+					second: '2-digit',
+					hour12: false,
+			  };
+	const formatter = new Intl.DateTimeFormat( locale, options );
```
sejas: Could we define an object with the formats instead of using a ternary operator?

Something like:
```
type FormatKey = 'short' | 'long';
```

```suggestion
	const options: Record< FormatKey, Intl.DateTimeFormatOptions > = {
		short: {
			day: 'numeric',
			month: 'long',
			year: 'numeric',
		},
		long: {
			day: '2-digit',
			month: 'short',
			year: 'numeric',
			hour: '2-digit',
			minute: '2-digit',
			second: '2-digit',
			hour12: false,
		},
	};
	const formatter = new Intl.DateTimeFormat( locale, options[ format ] );
```

bcotrim (commented):
Maybe it's unrelated to this PR but the table behavior looks odd on smaller windows (notice the alignment of the `X hours..` and the alignment of the titles compared with the values)
image

src/hooks/use-expiration-date.ts
```diff
@@ -5,12 +5,28 @@ import { useI18nData } from 'src/hooks/use-i18n-data';
...
+					month: 'long',
+					year: 'numeric',
+			  }
+			: {
+					day: '2-digit',
+					month: 'short',
+					year: 'numeric',
+					hour: '2-digit',
+					minute: '2-digit',
+					second: '2-digit',
```
sejas: I think we can remove the seconds from the long date.

```suggestion
```

src/hooks/use-expiration-date.ts
```diff
@@ -5,12 +5,28 @@ import { useI18nData } from 'src/hooks/use-i18n-data';
...
-		day: 'numeric',
-		month: 'long',
-		year: 'numeric',
-	} );
+function formatStringDate(
+	ms: number,
+	locale: SupportedLocale,
+	format: 'long' | 'short' = 'short'
+): string {
+	const options: Intl.DateTimeFormatOptions =
```
sejas: Another alternative could be using the format function from `date-fns`. Something like: `format( ms, 'PPpp', { locale: dateFnsLocales[ locale ] } )`

`Intl.DateTimeFormatOptions` is also good!, and consistent with what we had before.

sejas (changes requested):
Making the columns wider is a good solution.
Would be possible to make them more flexible, even if it means two rows?

In my case, because the domain is longer, it moves the other two columns to the right and the action menu disappears out of window producing a scroll.

Screenshot 2025-02-12 at 14 06 10

As an alternative, I was thinking about truncate the dates 🤔 .

sejas:
> Maybe it's unrelated to this PR but the table behavior looks odd on smaller windows (notice the alignment of the X hours.. and the alignment of the titles compared with the values)

Good catch.
~~What if we hide the updated column when the sidebar is hidden?~~

wojtekn:
@sejas shouldn't the main content width match the window width of Studio that has sidebar closed?

src/hooks/use-expiration-date.ts
wojtekn: Is there anything against using the solution we followed in previous UI?

src/hooks/use-expiration-date.ts
ivan-ottinger: That's a great idea! Thanks! I have made the change in https://github.com/Automattic/studio/pull/919/commits/e8711e3e4098243d55cbf4957fcca74fd9d40f58.

src/hooks/use-expiration-date.ts
sejas: If you mean using `Intl.DateTimeFormatOptions`, there is nothing bad about using it, let's keep it.

sejas:
> @sejas shouldn't the main content width match the window width of Studio that has sidebar closed?

You are right!, Let's not focus on the collapsed version as we can fix it by keeping the same width of the main content.

src/hooks/use-expiration-date.ts
ivan-ottinger: That's a valid point. I named the string based on the existing `dateString` we are already returning by `useExpirationDate` when used in the current Demo sites UI.

But since we are going to remove that UI soon, I think it makes sense to improve the naming of `fullDateString`. 

`expireDateString` sounds the most clear to me.

ivan-ottinger: → Updated in https://github.com/Automattic/studio/pull/919/commits/c85f3d276bc55e8b6606ad0779532e98ba5599db.

ivan-ottinger:
Thank you everyone for your feedback! It has been helpful.

I have addressed your suggestions and adjusted the column widths a little bit (focusing only on the view when the sidebar is not collapsed).

I have tested the view with multiple different languages and different sizes of sites and haven't found other issues there, but feel free to take a look as well.

ivan-ottinger:
Ok, I noticed the issue with long URLs is still there when it is bit longer than what I tested with. Going to look into it further:

> In my case, because the domain is longer, it moves the other two columns to the right and the action menu disappears out of window producing a scroll.

ivan-ottinger:
> Ok, I noticed the issue with long URLs is still there when it is bit longer than what I tested with. Going to look into it further:

I see we had truncate styling already applied to the first column, but it never applied since the max-width was set to 100%. I have now made the change for a fix width of the first column as well and truncating will now do its job:

Markup on 2025-02-12 at 16:51:53

wojtekn (approved):

sejas (approved):
Looks good! The dates are displayed in a single line, and we've added a tooltip in the expires column to display the exact date.
We can improve the spacing between the columns to be consistent in a different PR.

Screenshot 2025-02-12 at 17 31 39

Screenshot 2025-02-12 at 17 37 04

PR #926, opened 2025-02-12
Use share as tab name for previews

sejas:
## Related issues

- Related to p58i-jjs-p2#comment-66306

## Proposed Changes

- I think using `Share` is easier for users to understand what's the tab for.
- It's also easier to translate to other languages like Spanish

## Testing Instructions

- Observe the new tab says Share and it contains the previews table.

**Before**

Screenshot 2025-02-12 at 14 32 00
Screenshot 2025-02-12 at 14 40 40

**After**
Screenshot 2025-02-12 at 14 31 52
Screenshot 2025-02-12 at 14 40 21

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-content-tabs.tsx
+++ b/src/hooks/use-content-tabs.tsx
@@ -37,7 +37,7 @@ function useTabs() {
 			tabs.push( {
 				order: 3,
 				name: 'previews',
-				title: __( 'Previews' ),
+				title: __( 'Share' ),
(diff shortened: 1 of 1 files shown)


sejas:
Let's stick with Previews so we avoid confusion with Sync tab, and receive feedback in CfT.
Slack p1739371912108429-slack-C04GESRBWKW

PR #932, opened 2025-02-13
Remove `common_plugins` YAML anchor in favor of explicit list

mokagio:
## Proposed Changes

Remove the `common_plugins` YAML anchor previously used to DRY the plugins list in favor of explicitly listing each plugin a step needs.

The shared node was a remnant of the time where we didn't have the infrastructure to define plugins as environment variables that the pipeline can read (e.g. `$CI_TOOLKIT_PLUGIN`). Back then, DRYing in that way simplified bumping a plugin's version. But now that the plugins definition can be DRYed in the `shared-pipeline-var` file, we no longer need the shared plugins node.

Notice that I haven't added `NVM_PLUGIN` to the Windows build steps because the build script they use from the current CI toolkit version runs the Node.js setup. That will change soon, with the `nvm` plugin being the recommended way to set up Node.js in Windows, too. At that point, I'll add `NVM_PLUGIN` to those steps, too.

## Testing Instructions

- If CI is green, then the change was successful

## Pre-merge Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,9 +1,3 @@
-# Nodes with values to reuse in the pipeline.
-common_params:
-  plugins: &common_plugins
-    - $CI_TOOLKIT_PLUGIN
-    - $NVM_PLUGIN
-
 # IMAGE_ID is an env var that only macOS agents need.
 # Defining it at the root level propagates it too all agents, which can seem unnecessary but is a the same time convenient and DRY.
 env:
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,8 +1,5 @@
-# Nodes with values to reuse in the pipeline.
-common_params:
-  plugins: &common_plugins
-    - $CI_TOOLKIT_PLUGIN
-    - $NVM_PLUGIN
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
```
mokagio, line 2: Added the schema while at it...

AliSoftware (approved):

wojtekn (approved):

PR #934, opened 2025-02-13
Fix layout shift caused by scrollbar

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10254

## Proposed Changes

- I propose explicitly defining `overflow` as `auto` and then setting the `scrollbar-gutter` to `stable` to ensure that the scrollbar appearing for main content doesn't cause a layout shift, which can now be observed in the Assistant tab.

Image

## Testing Instructions

1. Open the Assistant tab
2. Run prompt observing bottom UI and confirm it doesn't move to the left when scrollbar appears
3. Clear discussion and confirm bottom UI width doesn't change
4. Switch between two sites, one with empty Assistant tab and another one with discussion, and confirm there is no visible layout shift in the bottom UI area
5. Test other tabs for any regressions

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-assistant.tsx
+++ b/src/components/content-tab-assistant.tsx
@@ -410,7 +410,7 @@ export function ContentTabAssistant( { selectedSite }: ContentTabAssistantProps 
 	const disabled = isOffline || ! isAuthenticated || ! userCanSendMessage || hasFailedMessage;
 
 	return (
-		<div className="relative min-h-full flex flex-col bg-gray-50" ref={ wrapperRef }>
+		<div className="relative min-h-full flex flex-col" ref={ wrapperRef }>
 			<div
 				data-testid="assistant-chat"
 				className={ cx(
--- a/src/components/site-content-tabs.tsx
+++ b/src/components/site-content-tabs.tsx
@@ -45,12 +45,7 @@ export function SiteContentTabs() {
 					key={ selectedTab }
 				>
 					{ ( { name } ) => (
-						<div
-							className="h-full overflow-y-auto"
-							style={ {
-								scrollbarGutter: 'stable',
-							} }
-						>
+						<div className="h-full">
 							{ name === 'overview' && <ContentTabOverview selectedSite={ selectedSite } /> }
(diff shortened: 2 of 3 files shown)


ivan-ottinger (commented):
The code change looks good and I haven't noticed any regressions. 👍🏼 

However, I am unable to reproduce the issue in the first place so leaving the PR to test by somebody else as well.

fredrikekelund (approved):
Tests well

nightnei (approved):
What I don't like with this approach is that:
1) At a glance, there is a feeling that we have different left/right paddings
2) But actually we have "white gap"

I am not a fan of pre-reservation space for scrollbars, but I see your point. Both approaches are not perfect, so we can try to go with this one, proposed.

Tested both MacOS and Windows - works well, as expected:
Screenshot 2025-02-14 at 12 58 05

fredrikekelund:
I agree with @nightnei – I prefer the default behavior of not displaying scrollbars when the content doesn't scroll. But @wojtekn shared in https://github.com/Automattic/dotcom-forge/issues/10254 that we've already established this convention in the sidebar and the settings tab

wojtekn:
@nightnei @fredrikekelund showing the scrollbar in each case is not the goal here. For that purpose, we would use `overflow: scroll`. The goal is to reserve the space to avoid layout shifts that appear when the scrollbar appears and disappears.

I consider the white bar a bug here. It seems that a room for a scrollbar is reserved in the tab panel container, but the grey background is set on one child container. I am not sure how to approach it - I guess we should move `bg-gray-50` up to `TabPanel` or rework overflow support to handle it inside `<ContentTab` components. The first won't work as the tab panel content is part of the tab component, and className is inaccessible. 

Any thoughts?

sejas (approved):
I've tried it  in every tab and it looks great. 👍 
I didn't observe any shift and the background is consistent.

Screenshot 2025-02-17 at 14 16 19

Screenshot 2025-02-17 at 14 17 01

Screenshot 2025-02-17 at 14 17 50

PR #936, opened 2025-02-14
Add tunneling support

sejas:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/9783

## Proposed Changes

- It enables users to add `WP_SITEURL` and `WP_HOME` constants in wp-config.php
- It adds a mu-plugin to redirect from localhost:XXXX to WP_SITEURL if defined, in that case wp-admin autologin doesn't work.

Kudos to @wojtekn for suggesting this approach p1739469267935719-slack-C06DRMD6VPZ

## Testing Instructions

- Run `npm start`
- Create a site
- Test the new site and confirm it's working as expected.
- Edit wp-config.php to add `WP_SITEURL` and `WP_HOME` constants at the top of the file
- You can set `http://whatever.localhost:8881` or run `ngrok http 8881` and use the url returned by  ngrok. 8881 is the port set by Studio and you can see it in the Settings tab.
- Start Studio site
- Observe the wp-admin and open site links open the defined URL and the site works as expected

> [!NOTE]  
> The autologin doesn't work if you are using Ngrok, and we don't plan to fix it in this PR.

https://github.com/user-attachments/assets/f779c511-33db-4390-a1ed-114a28a205d9

## Pre-merge Checklist


vendor/wp-now/src/config.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-wrapper-object-types */
+/* eslint-disable import/order */
```
sejas, line 2: I'm disabling these eslint rules to reduce noise in this PR.

vendor/wp-now/src/wp-now.ts
```diff
@@ -433,10 +435,15 @@ async function initWordPress(
...
-	};
+	const wpConfigConsts = {};
+
+	if ( ! wpConfigConstants?.WP_SITEURL ) {
+		wpConfigConsts[ 'WP_SITEURL' ] = siteUrl;
+	}
+
+	if ( ! wpConfigConstants?.WP_HOME ) {
+		wpConfigConsts[ 'WP_HOME' ] = siteUrl;
+	}
```
sejas: We set these constants only if they are not defined in wp-config.php.
If they are defined, then we do nothing and regular WordPress will use them.

sejas: When tested https://github.com/Automattic/studio/pull/936/commits/e98835dfe9bbda9b30039eb4120c80d778295d5e it seems setting the siteUrl was needed. After testing it now, I can create new sites without saving the siteurl in constants.

vendor/wp-now/src/wp-now.ts
```diff
@@ -471,6 +459,36 @@ export function getThemeTemplate( projectPath: string ) {
...
+			$current_host = $_SERVER['HTTP_HOST'] ?? '';
+
+			if ( preg_match( '/^localhost:\\d+$/', $current_host ) ) {
+				$requested_uri = $_SERVER['REQUEST_URI'] ?? '/';
+				wp_redirect( rtrim( WP_SITEURL, '/' ) . $requested_uri, 302 );
+				exit;
+			}
+		});
+		`
+	);
```
sejas, line 490: On https://github.com/Automattic/studio/pull/936/commits/01e01202c28a772636c00f863b618fada0ebb888 I explored a solution having a function `geWpConfigConstants` that exposed wp-config.php constants to Studio to display the correct links in wp-admin and open site.

After discussing it with @wojtekn , we decided to simplify the logic and relay in WordPress default redirect. Due to this issue https://core.trac.wordpress.org/ticket/33821#comment:10 the redirect doesn't work out of the box and we need this mu plugin.

The wp-admin link redirects correctly to the ngrok url defined in WP_SITEURL, but it doesn't perform the autologin.

vendor/wp-now/src/wp-now.ts
```diff
@@ -433,10 +424,7 @@ async function initWordPress(
 		initializeDefaultDatabase = true;
 	}
 
-	const wpConfigConsts = {
-		WP_HOME: siteUrl,
-		WP_SITEURL: siteUrl,
-	};
```
sejas, line 439: Removing this constants can uncover some issues in sites where the option siteurl and home point to a different URL.
When starting those sites WordPress will redirect to the URL saved in the database which will lead to a "non-response" page.
We can fix this behaviour in a different PR where we update those variables in database when we create a site with an existing SQLite database that already contains a siteurl with a different port.

wojtekn: Added one issue I've uncovered: https://github.com/Automattic/studio/issues/939

vendor/wp-now/src/wp-now.ts
```diff
@@ -471,6 +459,36 @@ export function getThemeTemplate( projectPath: string ) {
 async function mountInternalMuPlugins( php: PHP ) {
 	php.mkdir( PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER );
 
+	php.writeFile(
+		path.posix.join( PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER, '0-https-for-reverse-proxy.php' ),
```
fredrikekelund, line 463: Good on remembering to use POSIX paths 👍

vendor/wp-now/src/wp-now.ts
fredrikekelund: To recap, is the issue here that `redirect_canonical` doesn't work properly when we request a URL that includes a port number?

vendor/wp-now/src/wp-now.ts
fredrikekelund: Where do we actually define `siteurl` and `home` now..? Are those options just implied in the installation process?

fredrikekelund (approved):
LGTM 👍 I tested this with Jurassic Tube and it worked great.

I left a couple of questions that would be good to have answers for, but they shouldn't block merging.

vendor/wp-now/src/wp-now.ts
fredrikekelund: > Removing this constants can uncover some issues in sites where the option siteurl and home point to a different URL.

Isn't that normal behavior for WordPress, though?

vendor/wp-now/src/wp-now.ts
wojtekn: @fredrikekelund correct. It seems it also doesn't handle non-ssl `http://localhost:8888/` redirect to `https://my-tunnel.com/` well. It would make sense to dedicate some time later to check what we could do to rely on the core `redirect_canonical`.

vendor/wp-now/src/wp-now.ts
wojtekn: > Where do we actually define siteurl and home now..? Are those options just implied in the installation process?

This is correct. Studio calls `wp-admin/install.php` to install WordPress, which internally guesses the correct site url and saves it as option in the database. It happens in https://github.com/WordPress/WordPress/blob/c5cc9402b42939fd6a1370a7a463748ac9a2f595/wp-admin/includes/upgrade.php#L93

wojtekn: > Isn't that normal behavior for WordPress, though?

It is, and it should be done that way in the first place. Removing that now can uncover some issues, though. With those constants being set by Studio in a brute force way, it was masking flows in which the siteurl/home ended up being incorrect. One of those is an issue I linked above.

wojtekn (approved):

PR #937, opened 2025-02-14
Preview Sites: Avoid removing snapshots when switching between users

ivan-ottinger:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/7516.

## Proposed Changes

- add `userId` to the snapshot object as the new preview site is created
- automatically add current `userId` to all snapshots that are missing when `appdata` is getting loaded
- add filter that will ensure only snapshots linked with the logged-in user get loaded on the Previews tab 
- remove `clearFloatingSnapshots` that is no longer needed

## Testing Instructions

To test the changes in this PR, we will need two WPCOM accounts.

1. While logged-in to one of your WPCOM account in Studio, review your current preview sites.
2. Create a new preview site.
3. Take a look into `appdata-v1.json`. The newly created preview site should have a new `userId` value that equals to your WPCOM user ID:

Markup on 2025-02-14 at 11:26:59

4. Log out from your WPCOM account in Studio and log in to your another WPCOM account.
5. When you review the existing preview sites in Studio, the preview site that you created in the step 2 **should not** load in the Preview tab of the related local site.
6. Create another preview site while still logged-in to your second WPCOM account.
7. Take a look into `appdata-v1.json`. The newly created preview site should have a new `userId` value - but this time it should equal to your second WPCOM user ID (that you are currently logged into).
8. Log out from your second WPCOM account in Studio and log in to your first WPCOM account.
9. When you review the existing preview sites in Studio, you should see the preview site that you have created in the step 2. The preview site that you created with your second account in the step 6 **should not** load in the Preview tab of the related local site.
10. If you manually remove `userId` value from any of your existing preview sites in the `appdata-v1.json`, the value should get added automatically on the next `appdata-v1.json` load, e.g. when you restart the app. The userId of a currently-logged-in user will be used.
11. You can try other cases / variations as well.

The snapshot sequence number should be calculated for each user and site separately.

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -80,8 +80,14 @@ export function useArchiveSite() {
 	const errorMessages = useArchiveErrorMessages();
 	const archiveSite = useCallback(
 		async ( siteId: string ) => {
-			if ( ! client || ! user ) {
+			if ( ! client || ! user?.id ) {
 				// No-op if logged out
+				getIpcApi().showErrorMessageBox( {
+					title: __( 'Failed to archive site' ),
+					message: sprintf(
+						__( 'There was an authentication error. Please reauthenticate and try again.' )
+					),
+				} );
 				return;
 			}
 
@@ -164,7 +170,7 @@ export function useArchiveSite() {
 							nextSequence
 						),
 						sequence: nextSequence,
-						userId: user.id ?? undefined,
+						userId: user.id,
 					} );
 				} catch ( error ) {
 					if ( isWpcomNetworkError( error ) ) {
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -69,7 +69,7 @@ export function useArchiveSite() {
 		userId: number | null
 	): number => {
 		const siteSnapshots = snapshots.filter(
-			( s ) => s.localSiteId === siteId && ( ! s.userId || s.userId === userId )
+			( s ) => s.localSiteId === siteId && s.userId === userId
 		);
 		const existingSequences = siteSnapshots
 			.map( ( s ) => s.sequence ?? 0 )
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -158,7 +158,7 @@ export function useArchiveSite() {
 							nextSequence
 						),
 						sequence: nextSequence,
-						userId: user.id?.toString(),
+						userId: user.id,
 					} );
 				} catch ( error ) {
 					if ( isWpcomNetworkError( error ) ) {
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -18,7 +18,7 @@ export function useArchiveSite() {
 		( localSiteId: string ) => uploadingSites[ localSiteId ] || false,
 		[ uploadingSites ]
 	);
-	const { client } = useAuth();
+	const { client, user } = useAuth();
 	const { __ } = useI18n();
 	const { connectedSites } = useSyncSites();
 
@@ -74,7 +74,7 @@ export function useArchiveSite() {
 	const errorMessages = useArchiveErrorMessages();
 	const archiveSite = useCallback(
 		async ( siteId: string ) => {
-			if ( ! client ) {
+			if ( ! client || ! user ) {
 				// No-op if logged out
 				return;
 			}
@@ -158,6 +158,7 @@ export function useArchiveSite() {
 							nextSequence
 						),
 						sequence: nextSequence,
+						userId: user.id?.toString(),
 					} );
 				} catch ( error ) {
 					if ( isWpcomNetworkError( error ) ) {
@@ -194,6 +195,7 @@ export function useArchiveSite() {
 			__,
 			addSnapshot,
 			client,
+			user,
 			errorMessages,
 			fetchSnapshotUsage,
 			setUploadingSites,
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -158,7 +158,7 @@ export function useArchiveSite() {
 							nextSequence
 						),
 						sequence: nextSequence,
-						userId: user.id,
+						userId: user.id ?? undefined,
 					} );
 				} catch ( error ) {
 					if ( isWpcomNetworkError( error ) ) {
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -63,8 +63,14 @@ export function useArchiveSite() {
 		};
 	}, [ client, snapshots, updateSnapshot ] );
 
-	const getNextSequenceNumber = ( siteId: string, snapshots: Array< Snapshot > ): number => {
-		const siteSnapshots = snapshots.filter( ( s ) => s.localSiteId === siteId );
+	const getNextSequenceNumber = (
+		siteId: string,
+		snapshots: Array< Snapshot >,
+		userId: number | null
+	): number => {
+		const siteSnapshots = snapshots.filter(
+			( s ) => s.localSiteId === siteId && ( ! s.userId || s.userId === userId )
+		);
 		const existingSequences = siteSnapshots
 			.map( ( s ) => s.sequence ?? 0 )
 			.filter( ( n ) => ! isNaN( n ) );
@@ -143,7 +149,7 @@ export function useArchiveSite() {
 						formData,
 					} );
 
-					const nextSequence = getNextSequenceNumber( siteId, snapshots );
+					const nextSequence = getNextSequenceNumber( siteId, snapshots, user.id );
 
 					addSnapshot( {
 						url: response.domain_name,
--- a/src/hooks/use-snapshots.tsx
+++ b/src/hooks/use-snapshots.tsx
@@ -27,7 +27,6 @@ interface SnapshotContextType {
 		displayAlert?: boolean
 	) => Promise< void >;
 	deleteAllSnapshots: ( snapshots: Pick< Snapshot, 'atomicSiteId' >[] ) => Promise< void >;
-	clearFloatingSnapshots: ( allSnapshots: Pick< Snapshot, 'atomicSiteId' >[] ) => void;
 	isLoading: boolean;
 	loadingDeletingAllSnapshots: boolean;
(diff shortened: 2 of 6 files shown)


src/hooks/use-archive-site.ts
```diff
@@ -158,6 +158,7 @@ export function useArchiveSite() {
 							nextSequence
 						),
 						sequence: nextSequence,
+						userId: user.id ?? undefined,
```
katinthehatsite: I am testing this now and if I am reading this correctly, if the userId is undefined for some reason, it seems that
any snapshot with an undefined userId will be visible to all users/ any user, right?

katinthehatsite: Actually, never mind, I just read this in the description:

> You can try other cases / variations as well. Generally, the expected behavior is that a preview site loads on the Previewe tab only if:
that site belongs to that user, or:
doesn't belong to any user

👍

katinthehatsite: I guess maybe the only bit with this approach that is a bit tricky is that the site count in the user settings:  I am seeing three sites (2 previously created without user ID and one associated with my user ID but the site count shows me just one site):

Screenshot 2025-02-14 at 12 39 13 PM

ivan-ottinger: Thank you, Kat! That is a good catch.

> I guess maybe the only bit with this approach that is a bit tricky is that the site count in the user settings: I am seeing three sites (2 previously created without user ID and one associated with my user ID but the site count shows me just one site):

Thank you, Kat! That is a good catch.

I think this is expected since before this change comes into production, those already-existing preview sites won't have `userId` assigned and the active preview sites number does not come from `appdata.json` file.

Based on what @sejas mention in his comment (https://github.com/Automattic/dotcom-forge/issues/7516#issuecomment-2656765992), I think we will just consider those sites will eventually expire:

> If snapshots don't have userId because were created before this PR, then I think it's ok to include them in the list of previews, as they'll eventually expire.

---

However, your comment made me realize that we also need to ensure the snapshot _sequence numbers_ are counted for each user separately. At the moment, the sequence number is shared among all snapshots regardless of the user logged-in. As a result it is causing the default names for sites generating in incorrect order.

I will address this in the next commit.

ivan-ottinger: > I will address this in the next commit.

I have updated the logic in https://github.com/Automattic/studio/pull/937/commits/ea79d1149305957d7354ffc8e582cf0e5f7a2531.

katinthehatsite: >  I think we will just consider those sites will eventually expire

Yeah, I guess that makes sense. I was thinking how we could address it otherwise and I think the most straightforward solution would be to let them expire and be removed 👍 

I will give it another test shortly.

ivan-ottinger: Thank you, Kat.

katinthehatsite:
The switching between accounts worked for me! I noticed one other small thing - if I start creating a site with the user account A and then switch to user account B, I would still see the loading state. I have not investigated in detail though, it might be a different issue. I can give it another test on Monday.

ivan-ottinger:
> The switching between accounts worked for me! I noticed one other small thing - if I start creating a site with the user account A and then switch to user account B, I would still see the loading state. I have not investigated in detail though, it might be a different issue. I can give it another test on Monday.

Great observation!

I can see why this is happening.

When the first user initiates the preview site creation and then quickly logs out and then logs in with another user before the preview site creation is finished, the progress bar will still display itself until the site is created.

This is because the `isUploading` gets its value from `isUploadingSiteId` callback that is tied with the local site ID instead of the snapshot being created.

I would say this edge-case is expected behavior because when the preview site is being created `isUploading` should return `true` for any user that is logged in at the moment - since we want to block creating new preview sites at that time. If the user switches their accounts so quickly before the preview site creation is finished, it should be clear that the process is still finishing - since it is tied to that same local site.

Let's also see what Antonio thinks about this as well.

katinthehatsite:
> I would say this edge-case is expected behavior because when the preview site is being created isUploading should return true for any user that is logged in at the moment - since we want to block creating new preview sites at that time. If the user switches their accounts so quickly before the preview site creation is finished, it should be clear that the process is still finishing - since it is tied to that same local site.

Yeah, makes sense. Also, I do not think the user would be switching when the process is running as this would not be a logical user behaviour so it would indeed be more of an edge case 👍

katinthehatsite (approved):

src/hooks/use-archive-site.ts
```diff
@@ -158,6 +164,7 @@ export function useArchiveSite() {
...
```
sejas: Let's make sure user.id always exists when archiving a site.

```suggestion
						userId: user.id,
```

src/hooks/use-archive-site.ts
```diff
@@ -74,7 +80,7 @@ export function useArchiveSite() {
 	const errorMessages = useArchiveErrorMessages();
 	const archiveSite = useCallback(
 		async ( siteId: string ) => {
-			if ( ! client ) {
+			if ( ! client || ! user ) {
```
sejas: Let's check if `user?.id` has a value and if not, we can display a generic error to the user, suggesting to re-authenticate.

```suggestion
			if ( ! client ) {
```

src/hooks/use-snapshots.tsx
```diff
@@ -193,12 +193,6 @@ export const SnapshotProvider: React.FC< { children: ReactNode } > = ( { childre
 		);
 	}, [] );
 
-	useEffect( () => {
-		if ( initiated && ! loadingServerSnapshots && allSnapshots ) {
-			clearFloatingSnapshots( allSnapshots );
```
sejas, line 198: Let's also remove the definition of `clearFloatingSnapshots` function which is not used anymore.

sejas (commented):
@ivan-ottinger , thanks for creating the PR. It works as expected and now I can see the preview sites for my user.
The expired sites are not automatically removed.
The sequence number takes into account the userId ✨ .


My only concern is something I suggested in the issue.
> If snapshots don't have userId because were created before this PR, then I think it's ok to include them in the list of previews, as they'll eventually expire.

After testing the PR, I think it would be better if we assign the current userId to old snapshots in appData when Studio starts. What do you think?

src/hooks/use-snapshots.tsx
ivan-ottinger: Yeah, that's a good point. I was considering it, but left it there in case it is needed elsewhere. But it is a good call to remove it. 🙂 Will do it tomorrow.

ivan-ottinger:
Thank you for your reviews Kat and Antonio!

> After testing the PR, I think it would be better if we assign the current userId to old snapshots in appData when Studio starts. What do you think?

Sounds good. 👍🏼  I will make the change tomorrow, alongside the other changes you have suggested. 🙂

src/hooks/use-snapshots.tsx
ivan-ottinger: I have removed the definition in https://github.com/Automattic/studio/pull/937/commits/773722c214c14645eac2cd6d037a00ea721259e2.

src/hooks/use-archive-site.ts
ivan-ottinger: Nice idea! Thanks! What do you think about a change like this? https://github.com/Automattic/studio/pull/937/commits/ff18ea9c03e3f82557004fdd20ebd5d2bdb5120d

src/hooks/use-archive-site.ts
ivan-ottinger: Related to https://github.com/Automattic/studio/pull/937#discussion_r1959609647.

ivan-ottinger:
> After testing the PR, I think it would be better if we assign the current userId to old snapshots in appData when Studio starts. What do you think?

I have made the change in https://github.com/Automattic/studio/pull/937/commits/5ff1f6f961024998e94802ffc8adc1f3565222ff and https://github.com/Automattic/studio/pull/937/commits/68855c3e707e44e617f48aaf125595642c4d362a. When the appdata loads, it will populate the `userId` value for all snapshots that are missing it.

I have also removed checks that accepted snapshots without `userId`: https://github.com/Automattic/studio/pull/937/commits/9b31ef9d9baf9ae2a650eafefc008d2ba72765e8.

The PR is now ready for another review round. 🙂

src/storage/user-data.ts
```diff
@@ -46,6 +46,19 @@ function populatePhpVersion( sites: SiteDetails[] ) {
...
+
+	if ( userId && data.snapshots ) {
+		data.snapshots = data.snapshots.map( ( snapshot ) => {
+			if ( ! snapshot.userId ) {
+				return { ...snapshot, userId };
+			}
+			return snapshot;
+		} );
+	}
+}
```
sejas, line 60: We are adding this function that assigns a userId to pre-existing snapshots so they appear to in the previews tab.
We can consider removing this code after a few weeks/months as those preview sites would expire.
Before this PR, if a user logged out the snapshots disappeared from Studio.

src/storage/user-data.ts
```diff
@@ -55,6 +68,8 @@ export async function loadUserData(): Promise< UserData > {
 		try {
 			const parsed = JSON.parse( asString );
 			const data = fromDiskFormat( parsed );
+
+			populateSnapshotUserIds( data );
```
sejas, line 72: Thanks for isolating the code into this function. ✨

sejas (approved):
@ivan-ottinger , great work!
I've tested it and I confirm that now the preview sites appear under the user who created it.
We've removed `clearFloatingSnapshots`, so no preview sites are removed from appdata unless the user delete the preview site or they expire and the user clears the row.
For former demo sites without userId assigned we run a function when Studio starts that populates that value with the current authenticated user.

src/storage/user-data.ts
ivan-ottinger: My pleasure! 🙂

src/storage/user-data.ts
ivan-ottinger: → I have created a new task so we don't forget about the removal: https://github.com/Automattic/dotcom-forge/issues/10484.

ivan-ottinger: Wojtek suggested that we could rename the function and add a little bit of context instead of leaving the task hanging, so I updated the task and created related PR: https://github.com/Automattic/studio/pull/954.

PR #944, opened 2025-02-14
Fix port replacement on creating new site from backup

### LINKED ISSUE #939: Search and replace doesn't work correctly during the import
### Quick summary

When the site is being imported from a previously exported file, the search and replace doesn't work correctly.

Instead of replacing URLs with new correct URLs, it replaces them with `localhost:undefined`.

The issue becomes more visible after https://github.com/Automattic/studio/pull/936

### Steps to reproduce

1. Export Studio site
2. Add a new site and choose the exported site's backup
3. Confirm that the previous site's URLs were replaced with `localhost:undefined`

Search and replace works correctly.

### What actually happened

Search and replace replaced URLs with `localhost:undefined`

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Platform

_No response_

### Logs or notes

The following debug code can be used to observe the issue:

```
diff --git a/src/lib/import-export/import/importers/importer.ts b/src/lib/import-export/import/importers/importer.ts
index b82ec5d..f7a0ff0 100644
--- a/src/lib/import-export/import/importers/importer.ts
+++ b/src/lib/import-export/import/importers/importer.ts
@@ -102,6 +102,7 @@ abstract class BaseImporter extends EventEmitter implements Importer {
                const oldUrlVariants = [ `http://${ urlWithoutProtocol }`, `https://${ urlWithoutProtocol }` ];
 
                for ( const urlToReplace of oldUrlVariants ) {
+                       console.info( `Replacing URL: ${ urlToReplace } -> ${ studioUrl }` );
                                `search-replace '${ urlToReplace }' '${ studioUrl }'`,
                                { skipPluginsAndThemes: true }
```

The possible quick and dirty fix is to ensure server is started before the import:

```
diff --git a/src/hooks/use-import-export.tsx b/src/hooks/use-import-export.tsx
index 3aeffca..aed179d 100644
--- a/src/hooks/use-import-export.tsx
+++ b/src/hooks/use-import-export.tsx
@@ -107,7 +107,7 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
                        };
 
                        try {
-                               await stopServer( selectedSite.id );
+                               await startServer( selectedSite.id );
 
                                const filePath = file instanceof File ? getIpcApi().getPathForFile( file ) : file.path;
 
```

However, the server is being stopped there for a reason. E.g. if it's already started, the WP Admin auto-login doesn't work correctly. We may need to see how we could get the correct port for the stopped server in the first place.

nightnei:
## Related issues

- Fixes https://github.com/Automattic/studio/issues/939

## Proposed Changes

If we create a new site with the help of backup archive, then it works under the hood as - we create a new website and then import backup to created website. During importing, we need to replace website URL with local URL (e.g. localhost:8001), but we set up port only during running a website, so it ends up with `localhost:undefined`.
With proposed changes in this PR we are chisong port for website during creation of it, instead of during first running.

## Testing Instructions
Apply the next diff:
```
diff --git a/src/lib/import-export/import/importers/importer.ts b/src/lib/import-export/import/importers/importer.ts
index b82ec5d..f7a0ff0 100644
--- a/src/lib/import-export/import/importers/importer.ts
+++ b/src/lib/import-export/import/importers/importer.ts
@@ -102,6 +102,7 @@ abstract class BaseImporter extends EventEmitter implements Importer {
                const oldUrlVariants = [ `http://${ urlWithoutProtocol }`, `https://${ urlWithoutProtocol }` ];
 
                for ( const urlToReplace of oldUrlVariants ) {
+                       console.info( `Replacing URL: ${ urlToReplace } -> ${ studioUrl }` );
                                `search-replace '${ urlToReplace }' '${ studioUrl }'`,
                                { skipPluginsAndThemes: true }

```

1. Open Studio
2. Create a new website with teh help of backup
3. Assert that in terminal logs (from the diff above) you see a correct port, instead of `undefined`

### DIFF
--- a/src/hooks/use-add-site.ts
+++ b/src/hooks/use-add-site.ts
@@ -67,6 +67,9 @@ export function useAddSite() {
 							isNewSite: true,
 						} );
 						clearImportState( newSite.id );
+					} else {
+						// when we import file we start the server automatically, so doesn't make sense to run it again, to avoid unexpected behaviour
+						await startServer( newSite.id );
 					}
 
 					getIpcApi().showNotification( {
--- a/src/hooks/use-add-site.ts
+++ b/src/hooks/use-add-site.ts
@@ -68,7 +68,7 @@ export function useAddSite() {
 						} );
 						clearImportState( newSite.id );
 					}
-					await startServer( newSite.id );
+
 					getIpcApi().showNotification( {
 						title: newSite.name,
 						body: __( 'Your new site is up and running' ),
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -218,7 +218,7 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 						id: tempSiteId,
 						name: siteName || path,
 						path,
-						port: 9999, // Set a temporary port
+						port: -1, // Set a temporary port
 						running: false,
 						isAddingSite: true,
 						phpVersion: '',
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -218,6 +218,7 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 						id: tempSiteId,
 						name: siteName || path,
 						path,
+						port: 9999, // Set a temporary port
 						running: false,
 						isAddingSite: true,
 						phpVersion: '',
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -51,6 +51,7 @@ import { loadUserData, saveUserData } from 'src/storage/user-data';
 import { DEFAULT_PHP_VERSION } from 'vendor/wp-now/src/constants';
 import type { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
 import type { WpCliResult } from 'src/lib/wp-cli-process';
+import { portFinder } from 'src/lib/port-finder';
 
 const TEMP_DIR = nodePath.join( app.getPath( 'temp' ), 'com.wordpress.studio' ) + nodePath.sep;
 if ( ! fs.existsSync( TEMP_DIR ) ) {
@@ -161,11 +162,14 @@ export async function createSite(
 		}
 	}
 
+	const port = await portFinder.getOpenPort();
+
 	const details = {
 		id: crypto.randomUUID(),
 		name: siteName || nodePath.basename( path ),
 		path,
 		adminPassword: createPassword(),
+		port,
 		running: false,
 		phpVersion: DEFAULT_PHP_VERSION,
 	} as const;
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -38,6 +38,7 @@ import { StoredToken } from 'src/lib/oauth';
 import * as oauthClient from 'src/lib/oauth';
 import { createPassword } from 'src/lib/passwords';
 import { phpGetThemeDetails } from 'src/lib/php-get-theme-details';
+import { portFinder } from 'src/lib/port-finder';
 import { shellOpenExternalWrapper } from 'src/lib/shell-open-external-wrapper';
 import { sortSites } from 'src/lib/sort-sites';
 import { installSqliteIntegration, keepSqliteIntegrationUpdated } from 'src/lib/sqlite-versions';
@@ -51,7 +52,6 @@ import { loadUserData, saveUserData } from 'src/storage/user-data';
 import { DEFAULT_PHP_VERSION } from 'vendor/wp-now/src/constants';
 import type { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
 import type { WpCliResult } from 'src/lib/wp-cli-process';
-import { portFinder } from 'src/lib/port-finder';
 
 const TEMP_DIR = nodePath.join( app.getPath( 'temp' ), 'com.wordpress.studio' ) + nodePath.sep;
 if ( ! fs.existsSync( TEMP_DIR ) ) {
--- a/src/components/tests/assistant-code-block.test.tsx
+++ b/src/components/tests/assistant-code-block.test.tsx
@@ -25,6 +25,7 @@ const selectedSite: SiteDetails = {
 	path: '/test-site',
 	phpVersion: '8.0',
 	adminPassword: btoa( 'test-password' ),
+	port: 9999,
 };
 
 ( useSiteDetails as jest.Mock ).mockReturnValue( {
--- a/src/components/tests/content-tab-import-export.test.tsx
+++ b/src/components/tests/content-tab-import-export.test.tsx
@@ -19,6 +19,7 @@ const selectedSite: SiteDetails = {
 	path: '/test-site',
 	phpVersion: '8.0',
 	adminPassword: btoa( 'test-password' ),
+	port: 9999,
 };
 
 beforeEach( () => {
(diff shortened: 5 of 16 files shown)


src/ipc-handlers.ts
```diff
@@ -161,11 +162,14 @@ export async function createSite(
 		}
 	}
 
+	const port = await portFinder.getOpenPort();
```
nightnei, line 165: We are running website immidiatelly after creating it, so it's tiotale safe to define the port during setting up website. And I would say, there are even benefit of such approach, since we configure everything for our site in advance and have complete data for it and I don't see any benefit in postponning choosing port for it.

src/hooks/use-add-site.ts
```diff
@@ -68,7 +68,7 @@ export function useAddSite() {
 						} );
 						clearImportState( newSite.id );
 					}
-					await startServer( newSite.id );
```
nightnei, line 71: It's redundant, since we are starting server in the previous step above https://github.com/Automattic/studio/blob/trunk/src/hooks/use-import-export.tsx#L134

src/hooks/use-site-details.tsx
```diff
@@ -218,6 +218,7 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 						id: tempSiteId,
 						name: siteName || path,
 						path,
+						port: 9999, // Set a temporary port
```
wojtekn: Why we need to set temporary port?

wojtekn (commented):
Works as expected.

src/hooks/use-site-details.tsx
sejas: It's used as a placeholder to comply with the types.
This site placeholder is added so that Studio can display the new site in the sidebar while it's being created. The site information will be replaced once the site is created when the call to  `const data = await getIpcApi().createSite(path, siteName);` resolves.

We could split `getIpcApi().createSite` into two functions: one that retrieves the siteId and new port, add the site to the state in the react, and then pass those variables as parameters to the createSite function.

If we decide to use the temporary port, we could use a more different port like `-1` that could help us discover any issues during site creation.

src/ipc-handlers.ts
sejas: Now that the port is assigned in site creation, should we remove tha code from `SiteServer.start` ? https://github.com/Automattic/studio/blob/00db122a4d81b5096cac5710fa7a2d146f2de36c/src/site-server.ts#L81-L82

~~Maybe we can also remove the code related to release port:~~ https://github.com/Automattic/studio/blob/00db122a4d81b5096cac5710fa7a2d146f2de36c/src/site-server.ts#L74

sejas (commented):
Looks great. I confirm I was able to import a new studio site from a backup and the site started correctly with a new port following the next sequence.

I left a comment suggesting to remove the portFinder code from the start and stop functions.

My reasoning is that now that we don't override the WordPress siteurl, if we start a site in a port different to what's saved in sitedetails, then it will redirect to that URL producing unexpected behaviours.

src/ipc-handlers.ts
nightnei: I was thinking about it, but:
1. What in 1 months the users starts their website again, but another tool uses already this port too. Then in this case, `portFinder.getOpenPort( this.details.port );` fixes the issue and will pick up the next available port. I thought that it way intentionally added. If not intentionally, then anyway we should keep it to cover such case :) 
2. Actually, reading the code, I have assumption that we have another bug, connected to the point above. And it's edge case, but I was going to emulate case from the previous point and check how Studio handles it, since I think that we should do some thing like `replacUrl` in DB if port is busy during starting site and if `await portFinder.getOpenPort( this.details.port ); ` selected another one port. So in this case we need to update DB and update `appdata-v1.json` (but this point is already hanled inside `start` function, so we need to updater just DB).

But this my assumption is connected, but another topic, so in this PR I want to fix described issues and later cover that edge case as a separate PR, to not mix up everything.

WDYT about my thoughts?

src/hooks/use-site-details.tsx
nightnei: @sejas Updated to `-1`, makes sense, good idea 👍

src/ipc-handlers.ts
nightnei: @sejas btw, I have also assumption that [this issue](https://github.com/Automattic/dotcom-forge/issues/7075) is related to what I am mentioning above.

I will investigate it, but off the top of my head, the [issue here](https://github.com/Automattic/dotcom-forge/issues/7075) is:
1. We created website and set some port for urls inside DB
2. Then we removed site and created a new one with the same DB
3. As a result, data in the DB has old port, but we assigned a new one in appdata-v1.json during creating a site

What I want to say is - maybe, during starting server, we should always take port from `const port = await portFinder.getOpenPort( this.details.port ); ` and make updated to DB? So that, I suppose, we will fix [this issue](https://github.com/Automattic/dotcom-forge/issues/7075) and also we will fix edge case mentioned by me in the previous message.

Thinking about it more, I believe that we should do complex fix, like I proposed above or something similar, to keep Studio more stable.

sejas: My concern is that in this PR we are adding another `portFinder.getOpenPort` call and now we are setting the port to the studio site in two places, and it's not clear to me which one actually is setting the final port. That's why I was suggesting to  address it in this PR.

I understand the concern about a port being already busy, but currently the WordPress site will redirect to that port producing an unexpected behaviour. So I don't see harm on removing that code.


We could create an issue to consider that edge case and "replacing" the url+port in the database and appdata. We don't need to try to fix that in this PR. Maybe the issue is an edge case that it's not worth implementing it and we could document it somewhere, or wait to see if any user reports it. There are other solutions, like displaying an alert if the port is busy.

nightnei: I see, ok, let's go this way. I will create issues and I removed this:

```
const port = await portFinder.getOpenPort( this.details.port ); 
portFinder.addUnavailablePort( this.details.port ); 
```

But I think it's nice to keep `portFinder.releasePort( this.details.port );`, however it's not critical. The reason is - if we remove a site (e.g. :8002) and creating a new site, then willt be created 8003, instead of 8002 (which is actually free, and there is no sense to skip it).
WDYT, should we remove `releasePort`? I don't have strong preference, both options are ok - I like removing code but also I like when ports are used sequentially instead of jumping, so there are pros/cons in both cases :)

nightnei: I checked my assumption and it indeed fails and we don't show any error in UI, just endless "Starting..." copy.
Created [an issue](https://github.com/Automattic/dotcom-forge/issues/10483).

sejas (approved):
Thanks for the changes! I tested it and worked as expected 👍 
Thanks for creating the other issue.

src/ipc-handlers.ts
sejas: > But I think it's nice to keep portFinder.releasePort( this.details.port );, however it's not critical. The reason is - if we remove a site (e.g. :8002) and creating a new site, then willt be created 8003, instead of 8002 (which is actually free, and there is no sense to skip it).
> WDYT, should we remove releasePort? I don't have strong preference, both options are ok - I like removing code but also I like when ports are used sequentially instead of jumping, so there are pros/cons in both cases :)

I didn't see the code or tested this case but I also think the port assignment should be sequential. Maybe we should release the port when the site is removed ? I thought that if it's not in appData then it could be free to use.

sejas: Ok!, It seems the releaseport is already in delete and not in stop as I thought. It's good to leave it 👌 

https://github.com/Automattic/studio/blob/9b31ef9d9baf9ae2a650eafefc008d2ba72765e8/src/site-server.ts#L74

PR #950, opened 2025-02-18
Install the SQLite database integration plugin from the A8C repository

bgrgicak:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10260

## Proposed Changes

- Update the SQLite database integration plugin [from the A8C fork](https://github.com/Automattic/sqlite-database-integration)

## Testing Instructions

- Run `npm run postinstall`
- Confirm it doesn't fail
- Open the `wp-files/sqlite-database-integration` folder inside your Studio project folder
- Find the readme.md and confirm the plugin version is 2.1.17-alpha
- Start a site and confirm wp-admin loads

## Pre-merge Checklist


bgrgicak:
TODO:

- [x] Decide what to do with the [SQLite plugin path inside WP-now](https://github.com/Automattic/studio/blob/13b522e845369f43add5a5ba0bf5ca889f8592c7/vendor/wp-now/src/constants.ts#L24)
- [ ] ~~[Resolve SQLite issue that prevents Studio from booting with the latest plugin version ](https://github.com/Automattic/sqlite-database-integration/issues/19)~~

bgrgicak:
> Resolve SQLite issue that prevents Studio from booting with the latest plugin version 

I'm worried that [even if we fix this issue, ](https://github.com/Automattic/sqlite-database-integration/issues/19) there might be more, so I'm thinking of creating a release based on the last commit before the SQLite driver was merged. 

This would ensure Studio gets a stable update.

bgrgicak:
> I'm worried that [even if we fix this issue, ](https://github.com/Automattic/sqlite-database-integration/issues/19) there might be more, so I'm thinking of creating a release based on the last commit before the SQLite driver was merged.

I released a [new SQLite database integration version](https://github.com/Automattic/sqlite-database-integration/releases/tag/v2.1.17-alpha) without AST features.

src/lib/sqlite-database-integration-release.ts
```diff
@@ -0,0 +1,7 @@
+export async function getLatestSQLiteDatabaseIntegrationRelease(): Promise< string > {
+	const response = await fetch(
+		'https://api.github.com/repos/automattic/sqlite-database-integration/releases/latest'
+	);
+	const data = await response.json();
+	return `https://github.com/Automattic/sqlite-database-integration/archive/refs/tags/${ data.tag_name }.zip`;
```
wojtekn: Would it be safer to simplify it and lock it to a particular release so we won't have a situation in which the SQLite plugin is released with a feature that is incompatible or not tested with Studio?

bgrgicak: That depends on how you want to approach plugin updates. 
A8C controls this repository, so we decide what's included in releases and can ensure they are compatible with Studio.

@JanJakes might have more context on what's our plan for releasing new plugin versions from the private repo.

bgrgicak: @wojtekn what would you prefer?

JanJakes: I don't know how Studio handles other dependencies, but generally, I think it's safer to pin them (Studio `1.3.5` downloads SQLite integration `2.1.17-alpha` for all users, etc.).

wojtekn: Yes, let's pin it to the exact version. Could we also clean up the logic that auto-updates it for existing Studio sites, so it only updates the SQLite plugin if it was updated for the updated Studio version?

bgrgicak: Done

bgrgicak:
@wojtekn what would be a good place to document how SQLite plugin updates are handled?

wojtekn:
I've merged the trunk to this feature branch and solved conflicts. The PR works well - my new site uses 2.1.17-alpha and the old sites get the plugin upgraded when started.

As the SQLite plugin is downloaded when Studio is being built and bundled inside wp-file (in `download-wp-server-files.ts`), we could remove the whole code that downloads the plugin on run-time and always copy the bundled plugin. 

@Automattic/yolo any thoughts?

bgrgicak:
@wojtekn do you mean [removing this code for downloading the SQLite plugin?](https://github.com/Automattic/studio/blob/2b1afd580f2e3de7f21006e6265d32a3b546a1c8/vendor/wp-now/src/download.ts#L181)

wojtekn:
@wojtekn do you mean [removing this code for downloading the SQLite plugin?](https://github.com/Automattic/studio/blob/2b1afd580f2e3de7f21006e6265d32a3b546a1c8/vendor/wp-now/src/download.ts#L181)

Correct, as far as I understand, the download should now happen only in `download-wp-server-files.ts`.

bgrgicak:
@wojtekn I removed the download code, could you please take another look?

wojtekn:
Works as expected, thanks for cleaning up the code.

wojtekn (approved):

src/lib/sqlite-database-integration-release.ts
```diff
@@ -0,0 +1,3 @@
+export const SQLITE_DATABASE_INTEGRATION_VERSION = 'v2.1.17-alpha';
+
+export const SQLITE_DATABASE_INTEGRATION_RELEASE_URL = `https://github.com/Automattic/sqlite-database-integration/archive/refs/tags/${ SQLITE_DATABASE_INTEGRATION_VERSION }.zip`;
```
sejas: We could consider moving these constants to `src/constants.ts`
Not a strong opinion.

sejas (approved):
Works as expected.
It installs the SQLite plugin from GitHub with version `SQLITE_DATABASE_INTEGRATION_VERSION`.
Sites start and stop correctly, I confirm I was able to save a page, and create a new site.
SQLite plugin won't be updated automatically from .org plugins repository anymore.
We'll need to create a new GitHub release of the plugin, and release a new Studio version updating the version constant.

Screenshot 2025-03-24 at 09 40 27

src/lib/sqlite-database-integration-release.ts
bgrgicak: Done [7948cc9](https://github.com/Automattic/studio/pull/950/commits/7948cc93589fdc328b3418d8fdccea78a37ca53e)

PR #952, opened 2025-02-19
Studio: Add notification when the preview site is created

katinthehatsite:
## Related issues

Closes https://github.com/Automattic/dotcom-forge/issues/10486

## Proposed Changes

This PR adds a notification when a new preview site has been successfully created:

Screenshot 2025-02-20 at 9 52 45 AM

## Testing Instructions

* Navigate to the `Previews` tab
* Click on `Create preview site`
* Once the process is finished and the loading bar is not present anymore, observe that there is a notification about new site being created

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -49,7 +49,6 @@ export function useArchiveSite() {
 								isLoading: false,
 							} );
 
-							// Show notification only when site is actually ready
 							await getIpcApi().showNotification( {
 								title: __( 'Preview created' ),
 								body: snapshot.localSiteName
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -32,6 +32,9 @@ export function useArchiveSite() {
 		if ( loadingSnapshots.length === 0 ) {
 			return;
 		}
+
+		let activeSnapshot: Snapshot | null = null;
+
 		const intervalId = setInterval( async () => {
 			for ( const snapshot of loadingSnapshots ) {
 				if ( snapshot.isLoading ) {
@@ -48,16 +51,7 @@ export function useArchiveSite() {
 								...snapshot,
 								isLoading: false,
 							} );
-
-							await getIpcApi().showNotification( {
-								title: __( 'Preview created' ),
-								body: snapshot.localSiteName
-									? sprintf(
-											__( "Preview site for '%s' has been created." ),
-											snapshot.localSiteName
-									  )
-									: __( 'Preview site has been created.' ),
-							} );
+							activeSnapshot = snapshot;
 						}
 					} catch ( error ) {
 						updateSnapshot( {
@@ -70,6 +64,18 @@ export function useArchiveSite() {
 		}, 3000 );
 		return () => {
 			clearInterval( intervalId );
+			if ( activeSnapshot ) {
+				console.log( 'Showing notification...' );
+				getIpcApi().showNotification( {
+					title: __( 'Preview created' ),
+					body: activeSnapshot.localSiteName
+						? sprintf(
+								__( "Preview site for '%s' has been created." ),
+								activeSnapshot.localSiteName
+						  )
+						: __( 'Preview site has been created.' ),
+				} );
+			}
 		};
 	}, [ __, client, snapshots, updateSnapshot ] );
 
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -65,7 +65,6 @@ export function useArchiveSite() {
 		return () => {
 			clearInterval( intervalId );
 			if ( activeSnapshot ) {
-				console.log( 'Showing notification...' );
 				getIpcApi().showNotification( {
 					title: __( 'Preview created' ),
 					body: activeSnapshot.localSiteName
--- a/src/hooks/use-archive-site.ts
+++ b/src/hooks/use-archive-site.ts
@@ -1,19 +1,19 @@
 import * as Sentry from '@sentry/electron/renderer';
 import { sprintf } from '@wordpress/i18n';
 import { useI18n } from '@wordpress/react-i18n';
-import { useCallback, useEffect, useMemo } from 'react';
+import { useCallback, useMemo } from 'react';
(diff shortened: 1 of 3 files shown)


src/hooks/use-archive-site.ts
```diff
@@ -171,6 +182,7 @@ export function useArchiveSite() {
 						),
 						sequence: nextSequence,
 						userId: user.id,
+						localSiteName: selectedSite?.name,
```
katinthehatsite: This is to handle the case when the user starts creating a new site for a local site A and then switches to the local site B, we need to ensure that in the notification, the user still sees that name of the site A and not site B. One way to simplify this would be to remove the mention of the local site and simply say something like: You preview site has been created.

src/hooks/use-archive-site.ts
```diff
@@ -48,6 +48,16 @@ export function useArchiveSite() {
 								...snapshot,
 								isLoading: false,
 							} );
+
+							await getIpcApi().showNotification( {
+								title: __( 'Preview created' ),
```
katinthehatsite: The title could also potentially be the name of the local site from which the notification is coming instead of `Preview created`

src/hooks/use-archive-site.ts
ivan-ottinger: I wonder if we couldn't use the existing `localSiteId` value that is already stored in the snapshot data to get to its name (instead of adding the new `localSiteName` property to the array).

src/hooks/use-archive-site.ts
ivan-ottinger: I like it without including the name, but maybe we could consider adding the `site` word in there as well:

```suggestion
								title: __( 'Preview site created' ),
```

ivan-ottinger (commented):
The notification loads nicely. 👌🏼🙂 I just noticed that it gets triggered twice in my tests:

Markup on 2025-02-19 at 12:00:38

I have also shared a couple inline comments we could consider below.

src/hooks/use-archive-site.ts
katinthehatsite: I thought about that too but I did not find a way that was more straightforward than saving it with snapshot. It resulted in more code at the end so I decided against it but I can check further.

katinthehatsite:
>  I just noticed that it gets triggered twice in my tests

Hmm, let me check 👀

katinthehatsite:
> I just noticed that it gets triggered twice in my tests

I fixed it in https://github.com/Automattic/studio/pull/952/commits/c5b7b083a6be30e3cb617a5f139826b3f307ad94#diff-c73dac73b6601c39fc7747afaaa6a9c8d31b5960efd0bf65a7fd96e7c21f9410 - I realized that the issue was related to React re-renders. It should work now 🙇

src/hooks/use-archive-site.ts
katinthehatsite: I passed the site name instead in the title: https://github.com/Automattic/studio/pull/952/commits/8228e476ab8d4d19485b36a9c516c7434ad2f2ae to not repeat "Preview site" in the title and in the notificaiton body.

src/hooks/use-archive-site.ts
```diff
@@ -48,6 +51,7 @@ export function useArchiveSite() {
 								...snapshot,
 								isLoading: false,
 							} );
+							activeSnapshot = snapshot;
```
sejas: Why we save the activeSnapshot variable here, instead of sending the notification in this line?

src/hooks/use-archive-site.ts
sejas: One downside is that we're storing duplicate information in appData.

sejas (commented):
I've tested it and confirmed the notification is sent correctly after creating a preview site.
I think we don't need the `activeSnapshot` variable.
Let's see if we an avoid saving the localSiteName in appData.
Maybe we can use `snapshot.name` and `snapshot.url` instead. Something like:

```
	title: snapshot.name,
	body: sprintf( __( "Preview site '%s' has been created." ), snapshot.url ),
} );
```

src/hooks/use-archive-site.ts
katinthehatsite: This is what I did initially but I found out that this results in sending the notification twice and is related to the issue Ivan mentioned with double notification.

I think this is what happens based on the `console.log` I tested with after adding it inside the loop: 

* When we set `isLoading: false`, React re-renders
* The effect cleanup runs (because dependencies changed)
* Then the effect runs again with new state with the new snapshots value
* This triggers the second notification

Does it make sense?

katinthehatsite:
> Maybe we can use snapshot.name and snapshot.url instead. Something like:

Sounds good, I updated this in  https://github.com/Automattic/studio/pull/952/commits/76b67a7978efab7615e78a6e7f12cfbd5d17bec9 to look like this:

Screenshot 2025-02-20 at 9 52 45 AM

src/hooks/use-archive-site.ts
katinthehatsite: I removed it and used the snapshot name and URL instead.

src/hooks/use-archive-site.ts
sejas: @katinthehatsite, please remove `activeSnapshot` and move the `getIpcApi().showNotification` call to this line. While debugging, I discovered that this hook is a regular hook and does not reside within the context. As a result, it displays the notification multiple times and makes several calls to the JN/status endpoint. Additionally, `useArchiveSite` is used in both `content-tab-previwes` and `create-preview-button`, causing the notification to execute twice.

Let's move this useEffect inside the `useSnapshots` context. I've tested it and it works as expected 👍

sejas (changes requested):
Kat, let's move the useEffect inside the useSnapshots context.
Let's make sure we identify the purpose of the hook correctly.

src/hooks/use-archive-site.ts
katinthehatsite: @sejas  I applied suggested changes, feel free to give it another test.

sejas (approved):
Works as expected. I see the notification sent once 👍 

Thanks for the changes!

Screenshot 2025-02-21 at 09 33 52

PR #955, opened 2025-02-19
Clean up archiver options handling

wojtekn:
## Related issues

- https://github.com/Automattic/dotcom-forge/issues/9931

## Proposed Changes

- I propose to clean up archiver options handling to set and pass them explicitly for each format.

## Testing Instructions

- Export the site and confirm that the correct .zip is produced
- Push site and confirm the process works fine
- Optionally, comment out `unlink` call in `removeTemporalFile` and replace it with `console.log` to confirm file sizes

```
index f84e155..3ef3e47 100644
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -603,7 +603,8 @@ export function removeTemporalFile( event: IpcMainInvokeEvent, path: string ) {
                throw new Error( 'The given path is not a temporal file' );
        }
        try {
-               fs.unlinkSync( path );
+               console.log( path );
+               //fs.unlinkSync( path );
        } catch ( error ) {
                if ( isErrnoException( error ) && error.code === 'ENOENT' ) {
                        // Silently ignore if the temporal file doesn't exist

```

## Pre-merge Checklist

### DIFF
--- a/src/lib/import-export/export/exporters/default-exporter.ts
+++ b/src/lib/import-export/export/exporters/default-exporter.ts
@@ -4,6 +4,7 @@ import fsPromises from 'fs/promises';
 import os from 'os';
 import path from 'path';
 import archiver from 'archiver';
+import { ARCHIVER_OPTIONS } from 'src/constants';
 import { ExportEvents } from 'src/lib/import-export/export/events';
 import {
 	exportDatabaseToFile,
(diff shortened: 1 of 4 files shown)


src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -120,10 +121,8 @@ export class DefaultExporter extends EventEmitter implements Exporter {
 	private createArchive(): archiver.Archiver {
 		this.emit( ExportEvents.BACKUP_CREATE_START );
 		const isZip = this.options.backupFile.endsWith( '.zip' );
-		return archiver( isZip ? 'zip' : 'tar', {
-			gzip: ! isZip,
-			gzipOptions: isZip ? undefined : { level: 9 },
-		} );
+		const format = isZip ? 'zip' : 'tar';
+		return archiver( format, ARCHIVER_OPTIONS[ format ] );
```
wojtekn, line 125: Before the change, if a user chose the `.zip` file, Studio passed gzip options in the form of `{ gzip: false, gzipOptions: undefined }`. It didn't do a harm as it used non-zero default compression for the ZIP file internally, but it's better to handle options in both cases explicitly.

nightnei (approved):
LGTM and works correctly 👍

PR #960, opened 2025-02-20
Preview Sites: Update site creation progress bar logic

ivan-ottinger:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10485.

## Proposed Changes

- introduce a new `progress` and `isSnapshotProcessing` states in the `useSnapshot` that will ensure the current progress is preserved during component remounts (e.g. during tab switching)
- move the progress bar logic to a separate `useProgress` hook 
- remove `useProgressTimer` that is no longer used

## Testing Instructions

2. Head over to the Previews tab and create a new preview site.
3. The progress bar should be progressing continuously even when switching to different tabs / sites.
4. Updating another site while a new preview site is created shouldn't have any effect on the progress bar.  

## Pre-merge Checklist


src/components/content-tab-previews.tsx
```diff
@@ -178,6 +178,10 @@ export function ContentTabPreviews( { selectedSite }: ContentTabPreviewsProps )
 	);
 	const isOffline = useOffline();
 
+	useEffect( () => {
+		setIsSnapshotProcessing( isUploading || isSnapshotLoading );
+	}, [ isUploading, isSnapshotLoading, setIsSnapshotProcessing ] );
```
sejas: Let's remove this useEffect in favor of changing the state more imperatively in useArchiveSite

ivan-ottinger: Thank you, Antonio! I have made the change in https://github.com/Automattic/studio/pull/960/commits/1201b045efa28e61d201442d1519fcb4d1a0901b.

src/hooks/use-archive-site.ts
```diff
@@ -48,20 +49,22 @@ export function useArchiveSite() {
 								...snapshot,
 								isLoading: false,
 							} );
+							setIsSnapshotProcessing( false );
```
ivan-ottinger: I have set the progress to finish here instead of `finally{}` because this is actually where the snapshot is marked as active.

If we put the logic into `finally{}`, the progress bar would get reset before the snapshot is ready.

src/hooks/use-archive-site.ts
```diff
@@ -48,20 +49,22 @@ export function useArchiveSite() {
...
 						}
 					} catch ( error ) {
 						updateSnapshot( {
 							...snapshot,
 							isLoading: false,
 						} );
+						setIsSnapshotProcessing( false );
```
ivan-ottinger: Related to the decision to move the `setIsSnapshotProcessing( false )` logic away from `finally{}`: https://github.com/Automattic/studio/pull/960/files#r1963724839.

src/hooks/use-snapshots.tsx
```diff
@@ -93,6 +100,8 @@ export const SnapshotProvider: React.FC< { children: ReactNode } > = ( { childre
 	const [ snapshotQuota, setSnapshotQuota ] = useState( LIMIT_OF_ZIP_SITES_PER_USER );
 	const [ isLoadingSnapshotUsage, setIsLoadingSnapshotUsage ] = useState( false );
 	const [ snapshotCreationBlocked, setSnapshotCreationBlocked ] = useState( false );
+	const [ isSnapshotProcessing, setIsSnapshotProcessing ] = useState( false );
+	const progress = useProgress( { isProcessing: isSnapshotProcessing } );
```
sejas: What if we move the state inside the useProgress?
It could expose `startProgress` , `stopProgress` and `isActive`. Or maybe an object progress that contains all of that { start, stop, isActive, number }.

That way we could return. a single variable:
```
-			isSnapshotProcessing,
- 			setIsSnapshotProcessing,
			processingProgress, // this variable will contain start, stop and isActive
```

src/hooks/use-progress.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+interface UseProgressOptions {
+	isProcessing: boolean;
+	initialProgress?: number;
+	maxProgress?: number;
+	interval?: number;
+	step?: number;
+}
+
+export function useProgress( {
```
sejas, line 10: ha!, I didn't remember we had `useProgressTimer`. Could we re-use it or adapt it?

sejas (commented):
Looks good and works as expected.
I left a couple of suggestions.

src/hooks/use-snapshots.tsx
ivan-ottinger: I have made the refactor in https://github.com/Automattic/studio/pull/960/commits/8aa4e7284bf02ccb1a174e036c29e096170c3030, but it made things a bit more complicated with the progress for site deletion.

For now I am starting / stopping the progress there with `useEffect` as otherwise it looks like we would need to lift the deletion progress to the context as well.

But at the same time, the deletion is so quick that it may not be worth keeping track of the progress of each site that is being deleted separately (for the matter of switching tabs).

What do you think about the currently proposed code? 🤔

src/hooks/use-progress.tsx
ivan-ottinger: The logic is quite similar, but not the same. But since we are not using the old version anymore, maybe we could just rename the new one to use the same name (and file name)? 🙂

sejas: makes sense, let's leave it as it is. I like we are removing the useProgressTimer hook in favor of yours.

ivan-ottinger: Thank you, Antonio.

ivan-ottinger:
I have now rebased the PR based on the recent changes merged to `trunk` in https://github.com/Automattic/studio/commit/914eade0134e45b0c863b29cea4c1d5189755f90.

src/hooks/use-progress.tsx
```diff
@@ -0,0 +1,51 @@
...
+		return () => clearInterval( intervalId );
+	}, [ isProgressing, initialProgress, maxProgress, interval, step ] );
+
+	const startProgress = () => {
+		setIsProgressing( true );
+	};
+
+	const stopProgress = () => {
+		setIsProgressing( false );
+	};
```
sejas: Let's add useCallbacks

src/hooks/use-snapshots.tsx
```diff
@@ -39,6 +40,9 @@ interface SnapshotContextType {
 	isLoadingSnapshotUsage: boolean;
 	initiated: boolean;
 	snapshotCreationBlocked: boolean;
+	progress: number;
+	startProgress: () => void;
+	stopProgress: () => void;
```
sejas: What do you think about being specific about creationProgress:
```suggestion
	progressValue: number;
	startCreationProgress: () => void;
	stopCreationProgress: () => void;
```

src/hooks/use-snapshots.tsx
```diff
@@ -79,6 +83,9 @@ export const SnapshotContext = createContext< SnapshotContextType >( {
 	isLoadingSnapshotUsage: false,
 	initiated: false,
 	snapshotCreationBlocked: false,
+	progress: 10,
+	startProgress: () => undefined,
+	stopProgress: () => undefined,
```
sejas:
```suggestion
	progressValue: 10,
	startCreationProgress: () => undefined,
	stopCreationProgress: () => undefined,
```

src/modules/preview-site/components/preview-site-row.tsx
```diff
@@ -35,6 +36,19 @@ export function PreviewSiteRow( {
...
+		stopProgress: stopDeletionProgress,
+	} = useProgress();
+
+	useEffect( () => {
+		if ( isDeleting ) {
+			startDeletionProgress();
+		} else {
+			stopDeletionProgress();
+		}
+	}, [ isDeleting, startDeletionProgress, stopDeletionProgress ] );
```
sejas: We are calling startDeletion multiple times. I suggest refactoring to its own component.
UPDATE: Actually those were executing multiple times because the functions are not "useCallbacks".

Screenshot 2025-02-21 at 10 55 32

sejas: I've created a separate component https://github.com/Automattic/studio/pull/960/commits/6492a251c30c708c68d22b9aba06747abdc11f1e

src/hooks/use-progress.tsx
sejas: Done on https://github.com/Automattic/studio/pull/960/commits/2a9cde1d7d34ab1e03e42c08d740d9a6af343ffb

src/hooks/use-snapshots.tsx
sejas: Done https://github.com/Automattic/studio/pull/960/commits/02e0a4d67430e4cec523ccfda28a36907d1bf228

src/hooks/use-snapshots.tsx
sejas: Done https://github.com/Automattic/studio/pull/960/commits/02e0a4d67430e4cec523ccfda28a36907d1bf228

sejas (approved):
Beautiful creation and deletion progress bars. Now more consistent than ever. 🥳

[create-delete-progress-bar.webm](https://github.com/user-attachments/assets/a5818e80-f86c-474a-bbdd-3df5cad19144)

ivan-ottinger:
Just confirming the progress bar is working well after the latest changes. 👌🏼🙂

PR #961, opened 2025-02-20
More permissive zod schema when fetching sites

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10511

## Proposed Changes

The Sentry error shows that more than one key didn't pass the schema validation, but it only shows the name of the first one, `options` (the rest is truncated). This PR makes the `options` and `plan` objects optional. So, I'm assuming that `plan` could also be undefined.

I initially suspected this might happen for third-party Jetpack-connected site, but I couldn't reproduce it for sites like that, or for any other type of site.

## Testing Instructions

Code review should suffice

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-fetch-wpcom-sites/index.tsx
+++ b/src/hooks/use-fetch-wpcom-sites/index.tsx
@@ -14,25 +14,29 @@ export const sitesEndpointSiteSchema = z.object( {
 	URL: z.string(),
 	jetpack: z.boolean().optional(),
 	is_deleted: z.boolean(),
-	options: z.object( {
-		created_at: z.string(),
-		wpcom_staging_blog_ids: z.array( z.number() ),
-	} ),
+	options: z
+		.object( {
+			created_at: z.string(),
+			wpcom_staging_blog_ids: z.array( z.number() ),
+		} )
+		.optional(),
 	capabilities: z.object( {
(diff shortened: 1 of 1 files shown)


nightnei (approved):
LGTM 👍

wojtekn:
Shouldn't we gracefully handle records that are not valid according to schema and skip such records from UI, or display them as unsupported?

wojtekn (approved):

fredrikekelund:
> Shouldn't we gracefully handle records that are not valid according to schema and skip such records from UI, or display them as unsupported?

The changes in this PR make it so the sites that caused this particular exception are displayed as unsupported in the connection modal. I take your point that any other schema validation errors would make it so those sites are not displayed in the modal.

I'm not sure how we would tackle that in the most practical way, though… There would have to be more minimal schema that even those sites should conform (name and URL, maybe) to if we want to display those sites in the UI.

wojtekn:
> I'm not sure how we would tackle that in the most practical way, though… There would have to be more minimal schema that even those sites should conform (name and URL, maybe) to if we want to display those sites in the UI.

Should it fail gracefully, e.g., skip the record and not show or log the error when the record doesn't match the schema?

PR #962, opened 2025-02-20
Update urls in DB if site is being created from existing folder

nightnei:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/7075

## Proposed Changes
The initial issue was about broken URLs for images, but actually we have much bigger problems (Antonio was right [here](https://github.com/Automattic/dotcom-forge/issues/7075#issuecomment-2662715171)) - in 100% cases the website is completely broken. But this PR indeed fixes both - images and broken wp-admin.

Screenshot 2025-02-20 at 17 28 29

## Testing Instructions
1. We need to create a website with some big port, it will simplify testing
  1.1. Create a few websites (e.g. three)
  1.2 Open wp admin for the last one (let's assume that it has port 8883) and create a new post with adding some image there
  1.3 Remove the last one, but keep a folder on your laptop
  1.4 Remove other websites, together with folders, we don't need them anymore
2. So now we have folder with WP site, which saved port 8883 in URLs inside your DB (you can confirm it by opening `/wp-content/databases/.ht.sqlite` and check for example `wp-options` -> `home` || `siteurl`)
3. Create a new website, but provide this fodler as `Local path` in the creation dialog
4. Click on "WP Admin" link
5. Assert that wp admin looks good in browser, there are no error in console, connected to not available favicon, files etc
6. Assert that the image is available from 1.2 step
7. Additionally, you can check your DB (`wp_options`, `wp_users` and `wp_posts`)

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -42,6 +42,7 @@ import { portFinder } from 'src/lib/port-finder';
 import { shellOpenExternalWrapper } from 'src/lib/shell-open-external-wrapper';
 import { sortSites } from 'src/lib/sort-sites';
 import { installSqliteIntegration, keepSqliteIntegrationUpdated } from 'src/lib/sqlite-versions';
+import { updateSiteUrlToLocal } from 'src/lib/updateSiteUrlToLocal';
 import * as windowsHelpers from 'src/lib/windows-helpers';
 import { getLogsFilePath, writeLogToFile, type LogLevel } from 'src/logging';
 import { getMainWindow } from 'src/main-window';
@@ -190,6 +191,10 @@ export async function createSite(
 		if ( ! ( await pathExists( nodePath.join( path, 'wp-config.php' ) ) ) ) {
 			await installSqliteIntegration( path );
 		}
+	}
+
+	if ( await pathExists( nodePath.join( path, 'wp-content/database/.ht.sqlite' ) ) ) {
+		await updateSiteUrlToLocal( details.id );
 	}
 
 	const parentWindow = BrowserWindow.fromWebContents( event.sender );
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -190,11 +190,9 @@ export async function createSite(
 
 		if ( ! ( await pathExists( nodePath.join( path, 'wp-config.php' ) ) ) ) {
 			await installSqliteIntegration( path );
+		} else {
+			await updateSiteUrlToLocal( details.id );
 		}
-	}
-
-	if ( await pathExists( nodePath.join( path, 'wp-content/database/.ht.sqlite' ) ) ) {
-		await updateSiteUrlToLocal( details.id );
 	}
 
 	const parentWindow = BrowserWindow.fromWebContents( event.sender );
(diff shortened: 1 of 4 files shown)


src/ipc-types.d.ts
```diff
@@ -34,7 +34,6 @@ interface StoppedSiteDetails {
 interface StartedSiteDetails extends StoppedSiteDetails {
 	running: true;
 
-	port: number;
```
nightnei, line 37: Missed removing it in [this my PR](https://github.com/Automattic/studio/pull/944), so decided to remove in this PR

src/ipc-handlers.ts
```diff
@@ -189,6 +190,8 @@ export async function createSite(
 
 		if ( ! ( await pathExists( nodePath.join( path, 'wp-config.php' ) ) ) ) {
 			await installSqliteIntegration( path );
+		} else {
+			await updateSiteUrlToLocal( details.id );
```
nightnei, line 194: I was thinking about different options and places to do it, but this one looks as the most correct one, since existence of `wp-config.php` tells us - was the site already initialized or not.

src/lib/import-export/import/importers/importer.ts
```diff
@@ -72,53 +72,14 @@ abstract class BaseImporter extends EventEmitter implements Importer {
...
-		await this.replaceSiteUrl( siteId );
+		await updateSiteUrlToLocal( siteId );
 		this.emit( ImportEvents.IMPORT_DATABASE_COMPLETE );
 	}
 
 	protected async prepareSqlFile( _tmpPath: string ): Promise< void > {
 		// This method can be overridden by subclasses to prepare the SQL file before import.
 	}
 
-	protected async replaceSiteUrl( siteId: string ) {
```
nightnei, line 83: Moved it outside as a separate function since now it's used not only for import flow, so it's better to move it outside of import/export modules.

src/lib/updateSiteUrlToLocal.ts
```diff
@@ -0,0 +1,38 @@
...
+
+	const studioUrl = `http://localhost:${ server.details.port }`;
+	const oldUrl = currentSiteUrl.trim();
+	const urlWithoutProtocol = oldUrl.replace( /^https?:\/\//, '' );
+
+	const oldUrlVariants = [ `http://${ urlWithoutProtocol }`, `https://${ urlWithoutProtocol }` ];
+
+	for ( const urlToReplace of oldUrlVariants ) {
+		const { stderr, exitCode } = await server.executeWpCliCommand(
+			`search-replace '${ urlToReplace }' '${ studioUrl }' --skip-columns=guid`,
```
nightnei, line 26: The whole function is totally the same as it was before, so not necessary to review it, but two things:
1. Renamed to `updateSiteUrlToLocal`.
2. I added `--skip-columns=guid` to skip the `guid` column of posts, which is considered, AFAIU, as a best practice in WP world. I know that in this `Testing instructions` and similar cases it won't be a problem, but I believe that it's worth adding, since `guid` is just id, despite that fact that they look like urls.

src/ipc-handlers.ts
wojtekn: Looks like a good approach to fix the URL when the site is created.

src/lib/updateSiteUrlToLocal.ts
wojtekn: Thanks for summarizing the changes.

wojtekn (approved):
Works as expected.

PR #963, opened 2025-02-21
Suppress dns_get_record warnings

bgrgicak:
## Proposed Changes

- Suppress dns_get_record warnings with a internal mu-plugin 

## Testing Instructions

- Checkout this branch
- Create a site or start an existing site
- Add this code to the site, for example to a `functions.php` file
```
dns_get_record("php.net");
trigger_error('test', E_USER_WARNING);
```
- Reload the site
- Confirm there is no warning about using `dns_get_record`. This is what you would get before this PR.
```
 Warning: dns_get_record(): dns_get_record() always returns an empty array in PHP.wasm. in /var/www/html/wp-content/themes/twentytwentythree/functions.php on line 2
```
- Confirm that you see the _test_ warning

## Pre-merge Checklist


vendor/wp-now/src/wp-now.ts
```diff
@@ -623,6 +623,25 @@ set_error_handler(function($severity, $message, $file, $line) {
...
+	 * so we suppress the warning to avoid it being displayed to users.
+	 */
+	php.writeFile(
+		path.posix.join(
+			PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER,
+			'0-suppress-dns-get-record-warnings.php'
+		),
+		`<?php
+		set_error_handler(function($severity, $message, $file, $line) {
+			if ($severity === E_WARNING && strpos($message, "dns_get_record()") === 0) {
```
sejas: What do you think if we include the complete message to precisely target the warning triggered by Playground?

```suggestion
			if ($severity === E_WARNING && strpos($message, "dns_get_record(): dns_get_record() always returns an empty array in PHP.wasm.") === 0) {
```

sejas (approved):
It works as expected ✨ . I don't see the error anymore.
I suggest adding the specific error message to avoid hiding other possible errors related to dns_get_record.

PR #966, opened 2025-02-21
Studio: Sites with old SQLite database integration are not updated #

katinthehatsite:
## Related issues

Closes https://github.com/Automattic/dotcom-forge/issues/8638
Related to https://github.com/Automattic/dotcom-forge/issues/10305

## Proposed Changes

This PR ensures that if an older site has `wp-content/mu-plugins/sqlite-database-integration-main` installed on a site, the plugin will be updated to `wp-content/mu-plugins/sqlite-database-integration` on app restart. 

## Testing Instructions

Firstly, you would need to prepare a site to reproduce the issue:

* Create a Studio site
* Open it in Finder and rename  `mu-plugins/sqlite-database-integration` to `mu-plugins/sqlite-database-integration-main`
* Next, change the following line in `db.php`:

```
$sqlite_plugin_implementation_folder_path = realpath( __DIR__ . '/mu-plugins/sqlite-database-integration-main' );
```
* Now, restart the app and check the `mu-plugins` folder
* Observe that it does not update the `mu-plugins/sqlite-database-integration-main` to `mu-plugins/sqlite-database-integration`
* Pull the change from this branch
* Restart the app in the terminal where you started it with `rs`
* Check the `mu-plugins` folder
* Observe that the old integration was updated to `mu-plugins/sqlite-database-integration` and `main` was removed at the end

## Pre-merge Checklist


src/lib/sqlite-versions.ts
```diff
@@ -137,7 +137,7 @@ export async function keepSqliteIntegrationUpdated( sitePath: string ) {
 	const sqlitePath = path.join( sitePath, 'wp-content', 'mu-plugins', SQLITE_FILENAME );
 	const hasWpConfig = await fs.pathExists( path.join( sitePath, 'wp-config.php' ) );
 	const sqliteInstalled = await isSqlLiteInstalled( sqlitePath );
-	const sqliteOutdated = sqliteInstalled && ( await isSqliteInstallationOutdated( sqlitePath ) );
```
katinthehatsite: This current logic is problematic because is checks if the sqlite is installed at `export const SQLITE_FILENAME = 'sqlite-database-integration'` file path. When it does not find it (which it won't because the directory has `main` at the end), it evaluates to `false` so the update fails on the condition `sqliteOutdated` below: `if ( ( ! sqliteInstalled && ! hasWpConfig ) || sqliteOutdated )`. By removing the check, we are only checking if SqLite is outdated resulting in `true` and the installation of the update gets executed correctly.

katinthehatsite:
FYI, I will fix the unit test later if we agree on the approach!

bcotrim (commented):
I think removing the `sqliteInstalled` check from the `sqliteOutdated` calculation might make the `hasWpConfig` protection ineffective. Since `isSqliteInstallationOutdated()` returns `true` when no version is found (i.e., when SQLite isn't installed), wouldn't this cause SQLite to be installed on MySQL sites (with wp-config.php) when it shouldn't?

I might be missing something, what do you think?

katinthehatsite:
Thanks for the feedback @bcotrim - I thought this might be a problem so I have another possible solution and that is to keep the check for `sqliteInstalled` as a part of `sqliteOutdated` and instead solve the issue as part of `sqliteInstalled` check where we also check for the legacy path. I made some changes here: https://github.com/Automattic/studio/pull/966/files 

Let me know what you think about this approach 🙇

bcotrim (approved):
Thanks for the clarification and adjustments!
I tested the changes and they work as expected.
The code looks good, and I think this more conservative approach is better.
LGTM 👍

wojtekn:
@katinthehatsite it's great that we have a simple solution than the initial one.

It fixed the issue for my older sites, and now, all of them have the plugin in the correct location, and all can be exported.

PR #971, opened 2025-02-24
Run Lint and Unit Tests on mac agent

wojtekn:
## Related issues

- N/A

## Proposed Changes

- For some reason, the default agent stopped working today. I'm trying to run Lint and Unit Tests on the mac agent to fix recent build issues. For Lint, it doesn't matter where it runs, and Unit Tests should run on mac as there is separate build step for Windows Unit Tests.

## Testing Instructions

- N/A

## Pre-merge Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -8,6 +8,8 @@ env:
 
 steps:
   - label: Lint
+    agents:
+      queue: mac
     key: lint
     command: |
       .buildkite/commands/install-node-dependencies.sh
@@ -19,6 +21,8 @@ steps:
           context: Lint
 
   - label: Unit Tests
+    agents:
+      queue: mac
     key: unit_tests
     command: |
       .buildkite/commands/install-node-dependencies.sh


AliSoftware (approved):
I'm gonna approve to unblock and because with my limited knowledge of the whole context of this project, it makes sense to me to run the Lint and Unit Tests steps for this on a Mac queue given we also have their counterpart running on the `windows` queue (`Windows Unit Tests`).

Cc @mokagio who knows the context better though, because:
 - I wonder if the goal had always been to run those tests on Linux (`default` agent / EC2 instance) instead of Mac?
 - If the intent had always been to run them on Mac, how come it worked before today? Did it actually run on Mac agents by some magic I'm missing (I checked ou IaC and the `pipeline upload` steps don't specify a fallback `agent: { queue: mac }` for that Buildkite pipeline, so that's not it… and I also looked at the git history of the `pipeline.yml` file and didn't see `agent: { queue: mac }` neither being defined on those steps before like this PR adds, nor at the top level of the `pipeline.yml` to act as a default… so I don't see how they'd have magically run on Mac agents before given that… am I missing something?
 - Or, they always ran on the `default` queue (EC2 instances) all this time (be it intentional or not)… but somehow, `save_cache` worked on those default instances (and had the necessary credentials to be able to read and write into the `a8c-ci-cache` S3 bucket)… until it suddenly stopped working recently for some reason…?!

Anyway, I'm still puzzled on how it worked before, or if it worked as intended (built on Linux), why did it suddenly start breaking… if you have any additional insights about this @mokagio I'm interested!

PR #975, opened 2025-02-25
Move CI lint step back to default (Linux) queue

mokagio:
---

## Related issues

See internal discussion at p1740390251662719-slack-CC7L49W13

## Proposed Changes

Move the lint CI step back from the `mac` queue to the `default` queue. 

It was switched to the `mac` queue in https://github.com/Automattic/studio/pull/971 because of an issues with connecting to AWS in the `default` queue (Linux agents).

The AWS issue is yet to be addressed, but we have [a CI toolkit PR](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/162) that introduces a brute workaround: skip fetching the cache from AWS if you can't authenticate with AWS.

This way, at least we won't spam the `mac` queue, which is more expensive and should be used to run steps that legitimately need macOS.

The main reason for this is to have a working example of the changes in the plugin, https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/162, I am okay with closing this if you'd rather have faster CI and stay on the `mac` queue. (Although I'd say that given we have other steps that run in considerably slower than the lint one, even if lint is not as fast as possible the time to complete feedback won't be affected, so... 🤷‍♂️ ⚖️ )

_We ought to measure it, but don't I think the $ difference would be that much._

## Testing Instructions

Refer to [green CI](https://buildkite.com/automattic/studio/builds/4496#01953b27-d490-412c-ae79-32d9cc2362ea):

image

For reference, the lint step here ran ~45 seconds slower than [this random build](https://buildkite.com/automattic/studio/builds/4447#0195283c-c13f-4614-8383-c10f541f4604) from a few days ago before the AWS issue.

## Pre-merge Checklist

### DIFF
--- a/.buildkite/shared-pipeline-vars
+++ b/.buildkite/shared-pipeline-vars
@@ -6,7 +6,7 @@
 # The ~> modifier is not currently used, but we check for it just in case
 XCODE_VERSION=$(sed -E -n 's/^(~> )?(.*)/xcode-\2/p' .xcode-version)
 CI_TOOLKIT_VERSION='3.10.0'
-NVM_PLUGIN_VERSION='0.3.0'
+NVM_PLUGIN_VERSION='0.4.0'
 
 export IMAGE_ID="$XCODE_VERSION"
 export CI_TOOLKIT_PLUGIN="automattic/a8c-ci-toolkit#$CI_TOOLKIT_VERSION"
--- a/.buildkite/shared-pipeline-vars
+++ b/.buildkite/shared-pipeline-vars
@@ -5,7 +5,7 @@
 
 # The ~> modifier is not currently used, but we check for it just in case
 XCODE_VERSION=$(sed -E -n 's/^(~> )?(.*)/xcode-\2/p' .xcode-version)
-CI_TOOLKIT_VERSION='3.10.0'
+CI_TOOLKIT_VERSION='mokagio/fail-early-on-aws-missing-credentials'
 NVM_PLUGIN_VERSION='0.4.0'
 
 export IMAGE_ID="$XCODE_VERSION"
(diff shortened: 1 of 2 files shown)


.buildkite/shared-pipeline-vars
```diff
@@ -5,8 +5,8 @@
 
 # The ~> modifier is not currently used, but we check for it just in case
 XCODE_VERSION=$(sed -E -n 's/^(~> )?(.*)/xcode-\2/p' .xcode-version)
-CI_TOOLKIT_VERSION='3.10.0'
-NVM_PLUGIN_VERSION='0.3.0'
+CI_TOOLKIT_VERSION='mokagio/fail-early-on-aws-missing-credentials'
```
wojtekn, line 8: Would it be safer to wait for the CI Toolkit release?

wojtekn (approved):
@mokagio if the cost difference is negligible, I would consider leaving it on the faster queue. However, it's just 15-30 seconds so it's also negligible in the overall build time.

.buildkite/shared-pipeline-vars
mokagio: Yes. It would be best. In other PRs, e.g. https://github.com/Automattic/pocket-casts-android/pull/3653/files, I left a note about it, but I forgot here.

mokagio:
Closing because we are reasonably confident the issue the workaround addressed was an AWS hiccup. See also paaHJt-82Y-p2.

This does not mean we may not move the lint step back to the `default` but the context of this PR is off.

I have another CI-related PR to open soon that I'll use to cherry pick https://github.com/Automattic/studio/pull/975/commits/e2b531ede55f63c6524c4ab343dc85c8b28d5d15

PR #972, opened 2025-02-24
Improve site name suggestions to ensure both name and path uniqueness

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10503

## Proposed Changes

- Modified site name generation to ensure both name and path uniqueness
- Simplified code by using sites directly from useAddSite
- Improved type safety with proper SiteDetails type usage

## Testing Instructions

1. Happy path
- Create a new site using the default name
- Create another site.
Expected results: Site name is suggested without conflicting name or path

2. Previous error case
- Create a new site using the default name
- Rename the site in "Settings"
- Create a new site
Expected results: Site name is suggested without conflicting name or path

3. Suggested name limit
- Create a new site using the default name
- Create more sites until all suggestions are used (you can comment some names [here](https://github.com/Automattic/studio/blob/trunk/src/lib/generate-site-name.ts#L4) to speed up the process)
- Create another site.
Expected results: Site name is suggested, using the default name plus an incremental value, without conflicting name or path

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #976, opened 2025-02-25
Improve site path uniqueness validation using normalized comparisons

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10503
- Follow up from https://github.com/Automattic/studio/pull/972#discussion_r1969226935

## Proposed Changes

- Uses `generateProposedSitePath` to check if the path is available for a new site
- Updated `generateSiteName` to be async to support proper path validation
- Enhanced reliability of site path uniqueness validation

## Testing Instructions

- Create multiple WordPress instances with similar names
- Verify that path uniqueness is correctly validated across platforms
- Test on case-sensitive and case-insensitive filesystems if possible
- Ensure site creation works properly with the updated async flow

## Pre-merge Checklist

### DIFF
--- a/src/components/onboarding.tsx
+++ b/src/components/onboarding.tsx
@@ -86,6 +86,7 @@ export default function Onboarding() {
 			setDoesPathContainWordPress( isWordPress );
 		};
 		run();
+		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [] );
 
 	const handleSubmit = useCallback(
--- a/src/components/onboarding.tsx
+++ b/src/components/onboarding.tsx
@@ -75,16 +75,18 @@ export default function Onboarding() {
 		},
 	} );
 
-	const run = async () => {
-		const siteName = await generateSiteName( [] );
-		const { path, name, isWordPress } = await getIpcApi().generateProposedSitePath( siteName );
-		setSiteName( name );
-		setProposedSitePath( path );
-		setSitePath( '' );
-		setError( '' );
-		setDoesPathContainWordPress( isWordPress );
-	};
-	run();
+	useEffect( () => {
+		const run = async () => {
+			const siteName = await generateSiteName( [] );
+			const { path, name, isWordPress } = await getIpcApi().generateProposedSitePath( siteName );
+			setSiteName( name );
+			setProposedSitePath( path );
+			setSitePath( '' );
+			setError( '' );
+			setDoesPathContainWordPress( isWordPress );
+		};
+		run();
+	}, [] );
 
 	const handleSubmit = useCallback(
 		async ( event: FormEvent ) => {
--- a/src/components/onboarding.tsx
+++ b/src/components/onboarding.tsx
@@ -75,20 +75,16 @@ export default function Onboarding() {
 		},
 	} );
 
-	useEffect( () => {
-		const run = async () => {
-			const { path, name, isWordPress } = await getIpcApi().generateProposedSitePath(
-				generateSiteName( [] )
-			);
-			setSiteName( name );
-			setProposedSitePath( path );
-			setSitePath( '' );
-			setError( '' );
-			setDoesPathContainWordPress( isWordPress );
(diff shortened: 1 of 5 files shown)


bcotrim:
@Automattic/yolo 
We could potential just change the `isPathEnding` function to something like

```typescript
function isPathEndingWith(path: string, suffix: string): boolean {
  // Normalize paths by standardizing separators and case
  const normalize = (p: string) => p.toLowerCase().replace(/[\\\/]+/g, '/');
  return normalize(path).endsWith(normalize(suffix));
}
```

I opted for consistency and reliability, that this solution doesn't offer, but at the cost of some more calls to the IPC.
Since this operation only happens when creating a site, I don't think performance will be an issue, but let me know what you think.

wojtekn (approved):
I tested it on Windows and Mac and it works fine.

src/lib/generate-site-name.ts
```diff
@@ -26,22 +27,30 @@ export function generateSiteName( usedSites: SiteDetails[] ): string {
...
-	const isPathUnique = ( name: string ): boolean => {
+	const isPathUnique = async ( name: string ): Promise< boolean > => {
 		const sanitizedPath = sanitizeFolderName( name );
-		return ! usedSites.some( ( site ) => site.path.toLowerCase().endsWith( '/' + sanitizedPath ) );
+		for ( const site of usedSites ) {
+			if ( await getIpcApi().isPathEndingWith( site.path, `/${ sanitizedPath }` ) ) {
+				return false;
+			}
+		}
+		return true;
```
fredrikekelund:
```tsx
		const { isEmpty } = await getIpcApi().generateProposedSitePath( name );
		return isEmpty;
```
Could we do something like this instead? It seems like we're kind of duplicating logic now between `generateProposedSitePath` and `isPathEndingWith` now, and this solution seems to work well in my testing.

This would also allow us to remove `isPathEndingWith` altogether, simplifying this PR.

src/components/onboarding.tsx
```diff
@@ -75,20 +75,16 @@ export default function Onboarding() {
...
+	const run = async () => {
+		const siteName = await generateSiteName( [] );
+		const { path, name, isWordPress } = await getIpcApi().generateProposedSitePath( siteName );
+		setSiteName( name );
+		setProposedSitePath( path );
+		setSitePath( '' );
+		setError( '' );
+		setDoesPathContainWordPress( isWordPress );
+	};
+	run();
```
fredrikekelund: I don't understand this change. Why should we move this logic outside the `useEffect` callback? 🤔

fredrikekelund (commented):
Making `generateSiteName` async so that we can properly validate the proposed path seems to work well, but I think we should consider my proposal of dropping `isPathEndingWith` in favor of `generateProposedSitePath`

src/components/onboarding.tsx
bcotrim: We shouldn't 😅 
I rolled back the change.

src/components/onboarding.tsx
```diff
@@ -77,17 +77,15 @@ export default function Onboarding() {
...
+			const siteName = await generateSiteName( [] );
+			const { path, name, isWordPress } = await getIpcApi().generateProposedSitePath( siteName );
 			setSiteName( name );
 			setProposedSitePath( path );
 			setSitePath( '' );
 			setError( '' );
 			setDoesPathContainWordPress( isWordPress );
 		};
 		run();
-		// eslint-disable-next-line react-hooks/exhaustive-deps
```
fredrikekelund: Let's bring this line back to signal that we are intentionally excluding those effect dependecies.

bcotrim:
@fredrikekelund thanks for the suggestion, I agree that's an improvement!
I tested the changes on mac and windows and it seems to be working as expected.

fredrikekelund (approved):
LGTM and works as expected 👍

PR #977, opened 2025-02-25
Studio: Add failed state when the preview site fails to update

katinthehatsite:
## Related issues

Closes https://github.com/Automattic/dotcom-forge/issues/10384

## Proposed Changes

This PR adds a failed state for when an update to a preview site fails due to an error. Currently it displays the `Updated` state / wording even when update fails due to an error:

Screenshot 2025-02-25 at 1 28 28 PM

## Testing Instructions

* Make sure you have at least one preview site created
* Apply the following diff to simulate an error during update process:

```diff --git a/src/hooks/use-update-demo-site.tsx b/src/hooks/use-update-demo-site.tsx
index f55a9d1a..11d5f61e 100644
--- a/src/hooks/use-update-demo-site.tsx
+++ b/src/hooks/use-update-demo-site.tsx
@@ -83,6 +83,9 @@ export const DemoSiteUpdateProvider: React.FC< DemoSiteUpdateProviderProps > = (
                                        formData.push( [ 'wordpress_version', wordpressVersion ] );
                                }
 
+                               // Temporary: Simulate a failure for testing
+                               throw new Error( 'Artificial error for testing' );
+
                                const response = await client.req.post( {
                                        path: '/jurassic-ninja/update-site-from-zip',
                                        apiNamespace: 'wpcom/v2',
```
* Click on the `Update` button by a preview site in the preview actions menu
* Observe that when the error is triggered, you can see the failed state highlighted on the screenshot above
* Observe that the failed state gets removed after either 1 minute or when you switch tabs and come back
* Observe that the update time gets set back to what it was before the update was triggered 

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-update-demo-site.tsx
+++ b/src/hooks/use-update-demo-site.tsx
@@ -37,7 +37,6 @@ export const DemoSiteUpdateProvider: React.FC< DemoSiteUpdateProviderProps > = (
 			setUpdatingSites( ( prev ) => new Set( prev ).add( snapshot.atomicSiteId ) );
 
 			let archivePath = '';
-
 			try {
 				const { archivePath: tempArchivePath, archiveSizeInBytes } = await getIpcApi().archiveSite(
 					localSite.id,
--- a/src/hooks/use-update-demo-site.tsx
+++ b/src/hooks/use-update-demo-site.tsx
@@ -37,6 +37,12 @@ export const DemoSiteUpdateProvider: React.FC< DemoSiteUpdateProviderProps > = (
 				return;
 			}
 			setUpdatingSites( ( prev ) => new Set( prev ).add( snapshot.atomicSiteId ) );
+			// Clear any previous error state when starting a new update
+			setErrorSites( ( prev ) => {
+				const next = new Set( prev );
+				next.delete( snapshot.atomicSiteId );
+				return next;
+			} );
 
 			let archivePath = '';
 			try {
--- a/src/hooks/use-update-demo-site.tsx
+++ b/src/hooks/use-update-demo-site.tsx
@@ -2,7 +2,11 @@ import * as Sentry from '@sentry/electron/renderer';
 import { sprintf } from '@wordpress/i18n';
 import { useI18n } from '@wordpress/react-i18n';
 import { useCallback, useState, createContext, useContext, useMemo, ReactNode } from 'react';
-import { DEMO_SITE_SIZE_LIMIT_BYTES, DEMO_SITE_SIZE_LIMIT_GB } from 'src/constants';
+import {
+	DEMO_SITE_SIZE_LIMIT_BYTES,
+	DEMO_SITE_SIZE_LIMIT_GB,
+	UPDATED_MESSAGE_DURATION_MS,
+} from 'src/constants';
 import { useAuth } from 'src/hooks/use-auth';
 import { useSnapshots } from 'src/hooks/use-snapshots';
 import { getIpcApi } from 'src/lib/get-ipc-api';
@@ -37,12 +41,6 @@ export const DemoSiteUpdateProvider: React.FC< DemoSiteUpdateProviderProps > = (
 				return;
 			}
 			setUpdatingSites( ( prev ) => new Set( prev ).add( snapshot.atomicSiteId ) );
-			// Clear any previous error state when starting a new update
-			setErrorSites( ( prev ) => {
-				const next = new Set( prev );
-				next.delete( snapshot.atomicSiteId );
-				return next;
-			} );
 
 			let archivePath = '';
 			try {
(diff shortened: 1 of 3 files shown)


src/hooks/use-update-demo-site.tsx
```diff
@@ -96,6 +99,7 @@ export const DemoSiteUpdateProvider: React.FC< DemoSiteUpdateProviderProps > = (
 				} );
 				return response;
 			} catch ( error ) {
+				setErrorSites( ( prev ) => new Set( prev ).add( snapshot.atomicSiteId ) );
```
sejas, line 110: I believe we only add sites to errorSites and never remove them. Is that correct?

src/modules/preview-site/components/preview-site-row.tsx
```diff
@@ -35,31 +35,40 @@ export function PreviewSiteRow( {
 	const { url, date, isDeleting } = snapshot;
 	const { countDown, expireDateString, isExpired } = useExpirationDate( date );
 	const { fetchSnapshotUsage, removeSnapshot } = useSnapshots();
-	const { isDemoSiteUpdating } = useUpdateDemoSite();
+	const { isDemoSiteUpdating, hasDemoSiteError } = useUpdateDemoSite();
 	const isPreviewSiteUpdating = isDemoSiteUpdating( snapshot.atomicSiteId );
+	const hasError = hasDemoSiteError( snapshot.atomicSiteId );
 	const { formatRelativeTime } = useFormatLocalizedTimestamps();
 	const [ showUpdatedMessage, setShowUpdatedMessage ] = useState( false );
+	const [ showFailedMessage, setShowFailedMessage ] = useState( false );
```
sejas: Would be possible to use only one state to represent the error message? For example removing `hasDemoSiteError` and using just `hasDemoSiteError` ?

sejas (commented):
The implementation works as expected, although it seems we have two states representing the same thing.

src/hooks/use-update-demo-site.tsx
katinthehatsite: I think the state should be cleared by itself when the component unmounts. And the UI should be reset after 1 minute timeout: 

```
		const timeoutId = setTimeout( () => {
			setShowUpdatedMessage( false );
			setShowFailedMessage( false );
		}, UPDATED_MESSAGE_DURATION_MS );
```


In any case, I added some clean up when we attemp another update: https://github.com/Automattic/studio/pull/977/commits/69c4b08f0e6ecfc1bd493fdf35a377d288d91d76k
Let me know what you think.

src/modules/preview-site/components/preview-site-row.tsx
katinthehatsite: I made some changes in https://github.com/Automattic/studio/pull/977/commits/3f8cd264ddb460c9b5e6bf59a3bf2ad5b1481f62
Let me know what you think!

src/hooks/use-update-demo-site.tsx
```diff
@@ -96,6 +103,14 @@ export const DemoSiteUpdateProvider: React.FC< DemoSiteUpdateProviderProps > = (
...
+				setTimeout( () => {
+					setErrorSites( ( prev ) => {
+						const next = new Set( prev );
+						next.delete( snapshot.atomicSiteId );
+						return next;
+					} );
+				}, UPDATED_MESSAGE_DURATION_MS );
```
sejas: Let's avoid removing the error automatically and instead clear it when the user retries the udpate on line 37.


```suggestion
```

src/hooks/use-update-demo-site.tsx
```diff
@@ -25,6 +31,7 @@ export const DemoSiteUpdateProvider: React.FC< DemoSiteUpdateProviderProps > = (
 	const { client } = useAuth();
 	const { __ } = useI18n();
 	const [ updatingSites, setUpdatingSites ] = useState< Set< number > >( new Set() );
+	const [ errorSites, setErrorSites ] = useState< Set< number > >( new Set() );
 	const { updateSnapshot } = useSnapshots();
```
sejas, line 32: I think it's better to clear the error when the user re-tries and not after one minute.

```suggestion
			if ( errorSites.has( snapshot.atomicSiteId ) ) {
				setErrorSites( ( prev ) => {
					const next = new Set( prev );
					next.delete( snapshot.atomicSiteId );
					return next;
				} );
			}
```

sejas (approved):
@katinthehatsite , everything works as expected and I'm able to see the errors.
Thanks for removing the "duplicated" state and handling the errors just in one place.

I left a suggestion about when removing the errors.

Otherwise it works as expected and I confirm I see the alert and error message:
Screenshot 2025-02-27 at 12 48 03

Screenshot 2025-02-27 at 12 48 05

src/hooks/use-update-demo-site.tsx
katinthehatsite: Thanks, I made changes in https://github.com/Automattic/studio/pull/977/commits/e57be3afb0d7f6f56e19847f2c6e46788c5d75be

PR #978, opened 2025-02-25
Assistant: Ensure the last message does not animate on tab switch

ivan-ottinger:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10256.

## Proposed Changes

- ensure the last AI Assistant message is animated only when it arrives, not when a tab is switched

## Testing Instructions

1. Build the app with `npm install && npm run make` and test the changes on the build app outputted in the `/out` directory. Context: https://github.com/Automattic/studio/pull/978#issuecomment-2690157542.
2. Head over to the Assistant tab and try to reproduce the issue reported in https://github.com/Automattic/dotcom-forge/issues/10256. It should not be possible:
  - when a new message arrives, it should still animate
  - when you switch to a different tab and back, the last message should not animate 

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-assistant.tsx
+++ b/src/components/content-tab-assistant.tsx
@@ -120,7 +120,6 @@ const AuthenticatedView = memo(
 		submitPrompt,
 		wrapperRef,
 	}: AuthenticatedViewProps ) => {
-		const isInitialRenderRef = useRef( true );
 		const lastMessageRef = useRef< HTMLDivElement >( null );
 		const [ showThinking, setShowThinking ] = useState( isAssistantThinking );
 		const lastMessage = useMemo(
@@ -134,6 +133,7 @@ const AuthenticatedView = memo(
 			messages[ messages.length - 1 ]?.role === 'assistant' ? messages.slice( 0, -1 ) : messages;
 		const showLastMessage = lastMessage?.role === 'assistant';
 		const previousMessagesLength = useRef( messages.length );
+		const isInitialRenderRef = useRef( true );
 
 		// This effect may run twice when the component is mounted, which makes the viewport scroll
 		// to the wrong position. This happens because the app runs in React strict mode, meaning
--- a/src/components/content-tab-assistant.tsx
+++ b/src/components/content-tab-assistant.tsx
@@ -123,7 +123,6 @@ const AuthenticatedView = memo(
 		const isInitialRenderRef = useRef( true );
 		const lastMessageRef = useRef< HTMLDivElement >( null );
 		const [ showThinking, setShowThinking ] = useState( isAssistantThinking );
-
 		const lastMessage = useMemo(
 			() =>
 				showThinking
--- a/src/components/content-tab-assistant.tsx
+++ b/src/components/content-tab-assistant.tsx
@@ -123,7 +123,6 @@ const AuthenticatedView = memo(
 		const isInitialRenderRef = useRef( true );
 		const lastMessageRef = useRef< HTMLDivElement >( null );
 		const [ showThinking, setShowThinking ] = useState( isAssistantThinking );
-		const [ shouldAnimate, setShouldAnimate ] = useState( ! isInitialRenderRef.current );
 
 		const lastMessage = useMemo(
 			() =>
@@ -144,10 +143,6 @@ const AuthenticatedView = memo(
 		useEffect( () => {
 			if ( ! messages.length ) {
 				return;
-			}
-
-			if ( ! isInitialRenderRef.current ) {
-				setShouldAnimate( true );
 			}
 
 			let timer: NodeJS.Timeout;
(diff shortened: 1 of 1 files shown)


src/components/content-tab-assistant.tsx
```diff
@@ -120,8 +120,11 @@ const AuthenticatedView = memo(
 		submitPrompt,
 		wrapperRef,
 	}: AuthenticatedViewProps ) => {
+		const isInitialRenderRef = useRef( true );
```
ivan-ottinger: We are just moving the definition higher up from the line line `136`.

src/components/content-tab-assistant.tsx
```diff
@@ -203,12 +209,12 @@ const AuthenticatedView = memo(
...
 			( { message, children }: { message: MessageType; children: React.ReactNode } ) => {
 				const thinkingAnimation = {
-					initial: { opacity: 0, y: 20 },
+					initial: shouldAnimate ? { opacity: 0, y: 20 } : { opacity: 1, y: 0 },
 					animate: { opacity: 1, y: 0 },
 					exit: { opacity: 0, y: -20 },
 				};
 				const messageAnimation = {
-					initial: { opacity: 0, y: 20 },
+					initial: shouldAnimate ? { opacity: 0, y: 20 } : { opacity: 1, y: 0 },
```
sejas: Let's keep these values static and change the initial in the component prop. Line 247.

```suggestion
					initial: { opacity: 0, y: 20 } ,
```

src/components/content-tab-assistant.tsx
```diff
@@ -239,6 +246,7 @@ const AuthenticatedView = memo(
 									transition={ { duration: 0.3 } }
 									initial="initial"
 									animate="animate"
+									onAnimationComplete={ () => setShouldAnimate( false ) }
```
sejas:
```suggestion
```

src/components/content-tab-assistant.tsx
```diff
@@ -239,6 +246,7 @@ const AuthenticatedView = memo(
 									transition={ { duration: 0.3 } }
 									initial="initial"
```
sejas: Let's add here the condition if it should animate or not.
I think this is the only change that is required to fix the issue.

```suggestion
									initial={ ! isInitialRenderRef.current ? 'initial' : 'animate' }
```

src/components/content-tab-assistant.tsx
```diff
@@ -120,8 +120,11 @@ const AuthenticatedView = memo(
 		submitPrompt,
 		wrapperRef,
 	}: AuthenticatedViewProps ) => {
+		const isInitialRenderRef = useRef( true );
 		const lastMessageRef = useRef< HTMLDivElement >( null );
 		const [ showThinking, setShowThinking ] = useState( isAssistantThinking );
+		const [ shouldAnimate, setShouldAnimate ] = useState( ! isInitialRenderRef.current );
```
sejas: I don't think this state is necessary. We can infer the shouldAnimate from `! isInitialRenderRef.current`

sejas (changes requested):
Let's try fixing the issue without adding a new state: `const [ shouldAnimate, setShouldAnimate ] = useState( ! isInitialRenderRef.current );`

ivan-ottinger:
Thank you very much, @sejas! That is much cleaner solution. 🙂

src/components/content-tab-assistant.tsx
ivan-ottinger: Very nice! Thanks!

I have just used

```
initial={ isInitialRenderRef.current ? 'animate' : 'initial' }` 
```

instead of

```
initial={ ! isInitialRenderRef.current ? 'initial' : 'animate' }
```

wojtekn:
It works as expected. There is no distracting scroll animation on the tab change.

I'm unsure if the testing instructions are correct, though. It needs to be tested using a proper build and not the app running in dev mode.

ivan-ottinger:
Thanks for the review and testing, Wojtek.

> I'm unsure if the testing instructions are correct, though. It needs to be tested using a proper build and not the app running in dev mode.

I was not aware of this requirement. Do I understand correctly that all PRs should be tested on their builds (`npm install && npm run make`)?

---

I have updated the description of the PR and tested the change with the normal build. All is working well.

wojtekn:
> I was not aware of this requirement. Do I understand correctly that all PRs should be tested on their builds (npm install && npm run make)?

There is no such requirement. When I test this PR in dev mode, the content is still scrolled due to double rendering so it's not easy to validate that the animation is gone. There is no double render when using the proper build, so it's easier to validate the fix. See [comments](https://github.com/Automattic/studio/pull/804#issuecomment-2591896173).

ivan-ottinger:
> There is no such requirement. When I test this PR in dev mode, the content is still scrolled due to double rendering so it's not easy to validate that the animation is gone. There is no double render when using the proper build, so it's easier to validate the fix. See https://github.com/Automattic/studio/pull/804#issuecomment-2591896173.

Thanks for sharing the details. That makes sense. 👍🏼

sejas (approved):
Works as expected. I confirm the bottom-up animation in the text and thinking ellipsis appear only when receiving a new message and not when changing tabs or sites. 👍

PR #979, opened 2025-02-25
wip - multi worker poc

ashfame:
## Related issues

- Fixes #

## Proposed Changes

-

## Testing Instructions

-

## Pre-merge Checklist


brandonpayton:
@ashfame I spent a little time reading over these changes and haven't noticed anything specific to call out in the code.

IIRC, you mentioned earlier that performance seemed about the same as a single worker. If we can confirm that multiple processes are indeed handling requests, I am wondering a few things:
- Could this have anything to do with the express server?
  - How does its HTTP version compare with the faster reference points (like the Pressable example [here](https://github.com/Automattic/dotcom-forge/issues/8820#issuecomment-2665328977))
  - Is it handling connections differently than Pressable (WP Cloud)? For example, are HTTP connections being reused or kept open or just per-request?
- Does profiling the main process show us anything?
- Would message passing between multiple [workers](https://nodejs.org/api/worker_threads.html)(instead of processes) be any faster? Intuitively, working between multiple threads in one process seems like it should be faster than requiring the OS to coordinate multiple processes.
NOTE: If we switch to multiple workers, would we learn more from profiling because we'd be profiling all activity for a given server?

wojtekn:
@ashfame, will you be picking this one up soon, or should we close it?

ashfame:
@wojtekn This is very actively being worked on by @brandonpayton https://github.com/Automattic/wordpress-playground-private/pull/77

This POC PR can however be closed, since it would very likely not receive any updates. The support would land in PG first and then trickle down to Studio.

wojtekn:
Thanks for confirming that @ashfame .

PR #983, opened 2025-02-26
Add MC stats for import / export to track usage

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/9027

## Proposed Changes

- Add typed constants for stats groups and metrics
- Replace string literals with typed constants in bump-stats functions
- Update tests to use the new typed constants
- Add stats tracking for import and export operations (success/failure)
- Added tracking for import file types
- Added tracking for export content types

## Testing Instructions

- Run the test suite to verify that the bump-stats tests pass with the new typed implementation
- Verify that stats are properly tracked for import and export operations:
  - Run the app in development mode and check the console logs for "Would have bumped stat: studio-app-import=success" when importing a site
  - Run the app in development mode and check the console logs for bumped stats when exporting a site. For example
  - Verify that failure stats are logged when operations fail
 
- Check that app launch stats continue to work as expected:
  - Run the app in development mode and verify console logs show "Would have bumped stat: studio-app-launch-total=darwin" (or your platform) on startup
  - Verify weekly unique stats are logged with "Would have bumped stat: local-environment-launch-uniques=darwin"
- Confirm that the TypeScript type safety prevents using invalid stat groups or metrics

### Example Testing logs

 ```bash
  Would have bumped stat: studio-app-export=success
  Would have bumped stat: studio-app-export-content=database
  Would have bumped stat: studio-app-export-content=uploads
  Would have bumped stat: studio-app-export-content=plugins
  Would have bumped stat: studio-app-export-content=themes
  Would have bumped stat: studio-app-export-content=mu-plugins
```

```bash
Would have bumped stat: studio-app-import=success
Would have bumped stat: studio-app-import-type=application/zip
```

## Pre-merge Checklist


src/ipc-handlers.ts
```diff
@@ -676,8 +682,35 @@ export async function exportSite(
...
+				if ( options.includes.plugins ) {
+					bumpStat( STATS_GROUP.STUDIO_EXPORT_CONTENT, STATS_METRIC.PLUGINS );
+				}
+				if ( options.includes.themes ) {
+					bumpStat( STATS_GROUP.STUDIO_EXPORT_CONTENT, STATS_METRIC.THEMES );
+				}
+				if ( options.includes.muPlugins ) {
+					bumpStat( STATS_GROUP.STUDIO_EXPORT_CONTENT, STATS_METRIC.MU_PLUGINS );
+				}
+			}
```
sejas: I don't think we need this granularity in the export.
We have only two buttons in the UI.
One to export the whole site and another to export the database.
What if we expose the bumpStat to ipc-handlers and call it onClick on the UI?

https://github.com/Automattic/studio/blob/edf000ca48d2ff4f8a98e564fb32dc6c5c5988c2/src/components/content-tab-import-export.tsx#L77-L94

src/ipc-handlers.ts
```diff
@@ -119,11 +120,16 @@ export async function importSite(
 			sendIpcEventToRendererWithWindow( parentWindow, 'on-import', data, id );
 		};
 		const result = await importBackup( backupFile, site.details, onEvent, defaultImporterOptions );
+
+		bumpStat( STATS_GROUP.STUDIO_IMPORT, STATS_METRIC.SUCCESS );
+		bumpStat( STATS_GROUP.STUDIO_IMPORT_TYPE, backupFile.type );
```
sejas: I believe backupFile.type will indicate whether the file is a zip, SQL, or tar.gz format. However, it might be more interesting to differentiate between jetpack, playground, local, etc.

sejas (commented):
Code looks good.
I suggest tracking clicks on the export buttons to make the underlying logic easier to understand. Export entire site vs Export database.

src/lib/bump-stats/index.ts
fredrikekelund, line 1: Heads up that I just landed https://github.com/Automattic/studio/pull/988 which modified this file in its previous location. Should be an easy conflict to resolve

bcotrim:
@sejas @fredrikekelund Thanks for the reviews!
I've made some changes to address the comments, please let me know what you think!

src/hooks/use-import-export.tsx
```diff
@@ -303,6 +304,7 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
 
 	const exportFullSite = useCallback(
 		async ( selectedSite: SiteDetails ): Promise< string | undefined > => {
+			getIpcApi().ipcBumpStat( STATS_GROUP.STUDIO_EXPORT, STATS_METRIC.FULL_SITE );
```
fredrikekelund: Could we submit the bump stat from `exportSite` in `src/ipc-handlers.ts` instead? Aside from making it so we don't have to expose another IPC method, there's also an issue with submitting the bump stat here because it happens before the actual export operation. It's probably not very common for users to click the export button and then click `Cancel` in the `Save as…` dialog, but if they do, we still submit a bump stat right now.

src/lib/import-export/import/import-manager.ts
```diff
@@ -85,7 +85,8 @@ export async function importBackup(
 		removeBackupListeners = handleEvents( backupHandler, onEvent, BackupExtractEvents );
 		removeImportListeners = handleEvents( importer, onEvent, ImporterEvents );
 		await backupHandler.extractFiles( backupFile, extractionDirectory );
-		return await importer.import( site.path, site.id );
+		const result = await importer.import( site.path, site.id );
+		return result;
```
fredrikekelund: This looks like a leftover, we could probably revert it 🙂

src/ipc-handlers.ts
```diff
@@ -119,11 +121,16 @@ export async function importSite(
 			sendIpcEventToRendererWithWindow( parentWindow, 'on-import', data, id );
 		};
 		const result = await importBackup( backupFile, site.details, onEvent, defaultImporterOptions );
+
+		bumpStat( STATS_GROUP.STUDIO_IMPORT, STATS_METRIC.SUCCESS );
+		bumpStat( STATS_GROUP.STUDIO_IMPORT, result.importerType || 'unknown' );
```
fredrikekelund: Do we intentionally use the same group name for both these calls?

src/lib/bump-stats/types.ts
fredrikekelund, line 1: Adding explicit types for bump groups and metrics like this is great 👍 I would argue enums are a great fit here, though. 

This seems like a relatively simple change aside from that we have to fix the `importSite` function in ipc-handlers, where we're currently passing a dynamic value as the second argument to `bumpStat`. Maybe we could add a `bumpStat` prop to the `ImporterResult` type to fix this… 

Consider this a recommendation, not a hard blocker.

fredrikekelund (commented):
Using typed arguments for `bumpStat` functions is great, and the code mostly works well, but I left a few we should adress before merging.

src/hooks/use-import-export.tsx
bcotrim: Tracking the event in ipc handler was my initial approach, changed later by @sejas suggestion.
My suggestion would be to actually use both. On the UI side we can track the user click and action, on the ipc handler we track the success/failure of the request.
What do you think @sejas @fredrikekelund ?

src/lib/import-export/import/import-manager.ts
bcotrim: Thanks for catching that!

src/ipc-handlers.ts
bcotrim: I did.
I wasn't sure about the best approach for this, should we have everything regarding import in the same group and track `success`, `failure` and `type`, or have a distinct group for each, like `import_result` and `import_type`. What do you think it's the best?

src/lib/bump-stats/types.ts
bcotrim: Done, let me know how it looks!

src/hooks/use-import-export.tsx
fredrikekelund: Hmm, tracking both operations seems like overkill to me. If we want to understand which errors users encounter (if any), Sentry is a better tool than bump stats. 

I think your original implementation is clearer, even if I also agree with @sejas that we [don't need that much granularity](https://github.com/Automattic/studio/pull/983/files/83f1a6a4f389f7a3f2aa7a5fa8aac645ce80d490#r1973261873) when submitting the bump stat. It's fine to distinguish only between full site exports and database exports, but we could do that in the `exportSite` function by comparing the full shape of `options.includes`

src/lib/bump-stats/types.ts
```diff
@@ -0,0 +1,29 @@
+export enum STATS_GROUP {
```
fredrikekelund:
```suggestion
export enum StatsGroup {
```
The nice thing about enums is that we get types and values kind of rolled into one. No need to export a separate `StatsGroup` type.

src/lib/bump-stats/types.ts
```diff
@@ -0,0 +1,29 @@
...
+	FULL_SITE = 'full-site',
+	DATABASE_ONLY = 'database-only',
+	// Importer types
+	JETPACK_IMPORTER = 'JetpackImporter',
+	LOCAL_IMPORTER = 'LocalImporter',
+	SQL_IMPORTER = 'SQLImporter',
+	PLAYGROUND_IMPORTER = 'PlaygroundImporter',
+	WPRESS_IMPORTER = 'WpressImporter',
+	UNKNOWN = 'unknown',
+}
```
fredrikekelund, line 35:
```suggestion
	DARWIN = 'darwin',
	LINUX = 'linux',
	WINDOWS = 'win32',
}

export function getPlatform(): StatsMetric {
	switch ( process.platform ) {
		case 'darwin':
			return STATS_METRIC.DARWIN;
		case 'linux':
			return STATS_METRIC.LINUX;
		case 'win32':
			return STATS_METRIC.WINDOWS;
		default:
			throw new Error( `Unsupported platform: ${ process.platform }` );
	}
}
```

Something like this would allow us not to export a separate type for `StatsMetric`. Kind of a nit, but would be nice to address, IMO.

src/lib/import-export/import/importers/importer.ts
```diff
@@ -114,6 +115,7 @@ abstract class BaseBackupImporter extends BaseImporter {
 				wpContentDirectory: this.backup.wpContentDirectory,
 				wpConfig: this.backup.wpConfig,
 				meta: this.meta,
+				importerType: this.constructor.name,
```
fredrikekelund, line 118:
```suggestion
				importerType: StatsMetric.SQL_IMPORTER,
```
Defining the value by referencing the enum directly seems more understandable.

src/hooks/use-import-export.tsx
bcotrim: I moved all the logic to the IPC now.
The following stats should be bumped now.
- **Import** - `studio-app-import` group
  - importer type if the request is successful
  - failure if the request is unsuccessful 
- **export** - `studio-app-export`
   - `full-site` or `database-only` if successful and depending on the options
   - failure if the request is unsuccessful
   
I believe this will allow us to have an idea of usage, while still providing some insights into the number of failed requests. 

@sejas @fredrikekelund please let me know what you think about this approach.

src/lib/import-export/import/importers/importer.ts
bcotrim: I find it strange to reference stats props in the importer class.
What do you think about using a similar approach to the one we implemented for `getPlatform` ( check `getImporterMetric` )?

src/ipc-handlers.ts
```diff
@@ -676,8 +682,25 @@ export async function exportSite(
...
+			bumpStat(
+				StatsGroup.STUDIO_EXPORT,
+				options.includes.database &&
+					! options.includes.uploads &&
+					! options.includes.plugins &&
+					! options.includes.themes &&
+					! options.includes.muPlugins
+					? StatsMetric.DATABASE_ONLY
+					: StatsMetric.FULL_SITE
+			);
```
```suggestion
			const isDatabaseOnly =
				options.includes.database &&
				! options.includes.uploads &&
				! options.includes.plugins &&
				! options.includes.themes &&
				! options.includes.muPlugins;

			bumpStat(
				StatsGroup.STUDIO_EXPORT,
				isDatabaseOnly ? StatsMetric.DATABASE_ONLY : StatsMetric.FULL_SITE
			);
```
Nit, but this would read better if we assigned the condition to a variable.

src/lib/bump-stats/types.ts
```diff
@@ -0,0 +1,58 @@
...
+	switch ( importer ) {
+		case 'JetpackImporter':
+			return StatsMetric.JETPACK_IMPORTER;
+		case 'LocalImporter':
+			return StatsMetric.LOCAL_IMPORTER;
+		case 'SQLImporter':
+			return StatsMetric.SQL_IMPORTER;
+		case 'PlaygroundImporter':
+			return StatsMetric.PLAYGROUND_IMPORTER;
+		case 'WpressImporter':
```
fredrikekelund:
```suggestion
		case JetpackImporter.name:
			return StatsMetric.JETPACK_IMPORTER;
		case LocalImporter.name:
			return StatsMetric.LOCAL_IMPORTER;
		case SQLImporter.name:
			return StatsMetric.SQL_IMPORTER;
		case PlaygroundImporter.name:
			return StatsMetric.PLAYGROUND_IMPORTER;
		case WpressImporter.name:
```
Optional nit, but this would help clarify what we're referencing.

src/lib/import-export/import/importers/importer.ts
fredrikekelund: I like the new approach 👍

fredrikekelund (approved):
Nice! Looks good and works as expected 👍

src/lib/bump-stats/types.ts
bcotrim: Nice, that's perfect! 👍

bcotrim:
@sejas since you created the original issue and made some suggestion for different approaches I would love to have your feedback. Could I ask you for another review? 🙇

src/lib/bump-stats/index.ts
```diff
@@ -2,8 +2,7 @@ import * as Sentry from '@sentry/electron/main';
 import { isSameDay, isSameMonth, isSameWeek } from 'date-fns';
 import fetch from 'node-fetch';
 import { loadUserData, saveUserData } from 'src/storage/user-data';
-
-export type AggregateInterval = 'daily' | 'weekly' | 'monthly';
+import { AggregateInterval, StatsGroup, StatsMetric } from './types';
```
sejas: Should we use an absolute path?
```suggestion
import { AggregateInterval, StatsGroup, StatsMetric } from 'src/lib/bump-stats/types';
```

src/lib/bump-stats/types.ts
```diff
@@ -0,0 +1,66 @@
...
+export function getPlatformMetric( platform: typeof process.platform ): StatsMetric {
+	switch ( platform ) {
+		case 'darwin':
+			return StatsMetric.DARWIN;
+		case 'linux':
+			return StatsMetric.LINUX;
+		case 'win32':
+			return StatsMetric.WINDOWS;
+		default:
+			throw new Error( `Unsupported platform: ${ process.platform }` );
```
sejas: Not sure if this Error could crash the app in some systems. It's not very likely but maybe some unix systems return a different platform string. Maybe returning 'unknown' is safer?

src/ipc-handlers.ts
```diff
@@ -676,8 +682,25 @@ export async function exportSite(
 			const parentWindow = BrowserWindow.fromWebContents( event.sender );
 			sendIpcEventToRendererWithWindow( parentWindow, 'on-export', data, siteId );
 		};
-		return await exportBackup( options, onEvent );
+
+		const result = await exportBackup( options, onEvent );
+
+		if ( result ) {
```
sejas, line 688: Why do we need this `result` check? Or if we leave it should we consider the else as a failure?

sejas (approved):
@bcotrim , looks great! thanks for making the changes.
I left a couple of comments, like removing the error in `getPlatformMetric` and the `result` check  in export site.

I tested it by running npm start and observing the the bumpStat logs:

```
Would have bumped stat: studio-app-export=full-site
Would have bumped stat: studio-app-export=database-only
Would have bumped stat: studio-app-import=SQLImporter
Would have bumped stat: studio-app-import=JetpackImporter
Would have bumped stat: studio-app-import=PlaygroundImporter
```

PR #948, opened 2025-02-17
Update directory structure documentation

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10432

## Proposed Changes

- I propose to add more details on the Studio directory structure.

## Testing Instructions

- Review if changes are clear and correct: [code-contributions.md](https://github.com/Automattic/studio/blob/df01677f89b72644ab96f8a49e5c81a89b6ed183/docs/code-contributions.md#project-structure)

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #986, opened 2025-02-27
Create redux slice to list WordPress versions

sejas:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10528

## Proposed Changes

- Add a WordPress slice to get the list of versions with shortnames

## Testing Instructions

- Apply this diff. We could decide to load the information only once when the app starts.

```diff
diff --git a/src/components/content-tab-settings.tsx b/src/components/content-tab-settings.tsx
index 803b84d7..4bdc6d69 100644
--- a/src/components/content-tab-settings.tsx
+++ b/src/components/content-tab-settings.tsx
@@ -1,11 +1,16 @@
 import { useI18n } from '@wordpress/react-i18n';
-import { PropsWithChildren } from 'react';
+import { PropsWithChildren, useEffect } from 'react';
 import { CopyTextButton } from 'src/components/copy-text-button';
 import DeleteSite from 'src/components/delete-site';
 import EditPhpVersion from 'src/components/edit-php-version';
 import EditSite from 'src/components/edit-site';
 import { useGetWpVersion } from 'src/hooks/use-get-wp-version';
 import { decodePassword } from 'src/lib/passwords';
+import { useAppDispatch, useRootSelector } from 'src/stores';
+import {
+	wordpressVersionsThunks,
+	wordpressVersionsSelectors,
+} from 'src/stores/wordpress-versions-slice';
 
 interface ContentTabSettingsProps {
 	selectedSite: SiteDetails;
@@ -29,6 +34,12 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
 	const storedPassword = decodePassword( selectedSite.adminPassword ?? '' );
 	const password = storedPassword === '' ? 'password' : storedPassword;
 	const wpVersion = useGetWpVersion( selectedSite );
+	const wordpressVersions = useRootSelector( wordpressVersionsSelectors.selectWordPressVersions );
+	const dispatch = useAppDispatch();
+	useEffect( () => {
+		dispatch( wordpressVersionsThunks.fetchWordPressVersions() );
+	}, [ dispatch ] );
+	console.log( wordpressVersions );
 	return (
 		<div className="p-8">
 			<table className="mb-2 m-w-full" cellPadding={ 0 } cellSpacing={ 0 }>

```
- Run `npm start`
- Go to the settings tab
- Open the chrome developer tools
- Observe the list of versions in the console

Screenshot 2025-02-27 at 16 23 25

## Pre-merge Checklist

### DIFF
--- a/src/stores/tests/wordpress-versions-slice.test.ts
+++ b/src/stores/tests/wordpress-versions-slice.test.ts
@@ -1,0 +1,182 @@
+import * as Sentry from '@sentry/electron/renderer';
+import { store } from 'src/stores';
+import { testActions, testReducer } from 'src/stores/tests/utils/test-reducer';
+import {
+	fetchWordPressVersions,
+	wordpressVersionsSelectors,
+} from 'src/stores/wordpress-versions-slice';
+
+// Mock fetch
+global.fetch = jest.fn();
+
+// Mock Sentry
+jest.mock( '@sentry/electron/renderer', () => ( {
+	captureException: jest.fn(),
+} ) );
+
+store.replaceReducer( testReducer );
+
+describe( 'wordpress-versions-slice', () => {
+	beforeEach( () => {
+		jest.clearAllMocks();
+		store.dispatch( testActions.resetState() );
+	} );
+
+	describe( 'fetchWordPressVersions', () => {
+		it( 'should update versions when API call is successful', async () => {
+			// Mock successful API response
+			( global.fetch as jest.Mock ).mockResolvedValueOnce( {
+				ok: true,
+				json: jest.fn().mockResolvedValueOnce( {
+					offers: [
+						{ version: '6.4.0', response: 'autoupdate' },
+						{ version: '6.5.0-beta1', response: 'autoupdate' },
+						{ version: '6.3.0', response: 'upgrade' }, // Should be filtered out
+					],
+				} ),
+			} );
+
+			const result = await store.dispatch( fetchWordPressVersions() );
+
+			expect( result.type ).toBe( 'wordpressVersions/fetchWordPressVersions/fulfilled' );
+			expect( result.payload ).toEqual( [
+				{ version: '6.4.0', isBeta: false },
+				{ version: '6.5.0-beta1', isBeta: true },
+			] );
+
+			const state = store.getState();
+			const versions = wordpressVersionsSelectors.selectWordPressVersions( state );
+
+			expect( versions ).toHaveLength( 2 );
+			expect( versions[ 0 ] ).toEqual( { version: '6.4.0', isBeta: false } );
+			expect( versions[ 1 ] ).toEqual( { version: '6.5.0-beta1', isBeta: true } );
+			expect( state.wordpressVersions.status ).toBe( 'succeeded' );
+			expect( state.wordpressVersions.error ).toBeNull();
+		} );
+
+		it( 'should handle API response with no autoupdate offers', async () => {
+			// Mock API response with no autoupdate offers
+			( global.fetch as jest.Mock ).mockResolvedValueOnce( {
+				ok: true,
+				json: jest.fn().mockResolvedValueOnce( {
+					offers: [
+						{ version: '6.3.0', response: 'upgrade' },
+						{ version: '6.2.0', response: 'upgrade' },
+					],
+				} ),
+			} );
+
+			const result = await store.dispatch( fetchWordPressVersions() );
+
+			expect( result.type ).toBe( 'wordpressVersions/fetchWordPressVersions/fulfilled' );
+			expect( result.payload ).toEqual( [] );
+
+			const state = store.getState();
+			const versions = wordpressVersionsSelectors.selectWordPressVersions( state );
+
+			expect( versions ).toHaveLength( 0 );
+			expect( state.wordpressVersions.status ).toBe( 'succeeded' );
+		} );
+
+		it( 'should handle non-OK API response', async () => {
+			// Mock failed API response
+			( global.fetch as jest.Mock ).mockResolvedValueOnce( {
+				ok: false,
+			} );
+
+			const result = await store.dispatch( fetchWordPressVersions() );
+
+			expect( result.type ).toBe( 'wordpressVersions/fetchWordPressVersions/rejected' );
+			expect( result.payload ).toEqual( undefined );
+
+			const state = store.getState();
+			expect( state.wordpressVersions.versions ).toHaveLength( 0 );
+			expect( state.wordpressVersions.status ).toBe( 'failed' );
+			expect( state.wordpressVersions.error ).toBe( 'Failed to fetch WordPress versions' );
+		} );
+
+		it( 'should handle API fetch error', async () => {
+			// Mock fetch throwing an error
+			( global.fetch as jest.Mock ).mockRejectedValueOnce( new Error( 'Network error' ) );
+
+			const result = await store.dispatch( fetchWordPressVersions() );
+
+			expect( result.type ).toBe( 'wordpressVersions/fetchWordPressVersions/rejected' );
+
+			const state = store.getState();
+			expect( state.wordpressVersions.versions ).toHaveLength( 0 );
+			expect( state.wordpressVersions.status ).toBe( 'failed' );
+			expect( state.wordpressVersions.error ).toBe( 'Network error' );
+		} );
+
+		it( 'should handle schema validation error', async () => {
+			// Mock API response with invalid schema
+			( global.fetch as jest.Mock ).mockResolvedValueOnce( {
+				ok: true,
+				json: jest.fn().mockResolvedValueOnce( {
+					// Missing 'offers' field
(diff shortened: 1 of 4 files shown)


src/stores/wordpress-versions-slice.ts
```diff
@@ -0,0 +1,82 @@
...
+	versions: [],
+	status: 'idle',
+	error: null,
+};
+
+export const fetchWordPressVersions = createAsyncThunk(
+	'wordpressVersions/fetchWordPressVersions',
+	async () => {
+		const response = await fetch(
+			`https://api.wordpress.org/core/version-check/1.7/?channel=beta&version=${ MINIMUM_WORDPRESS_VERSION }`
```
wojtekn: Would it make sense to extract code responsible for the API call to `src/api` according to the [Project Structure](https://github.com/Automattic/studio/blob/trunk/docs/code-contributions.md#project-structure) doc?

src/stores/wordpress-versions-slice.ts
```diff
@@ -0,0 +1,82 @@
...
+		);
+		if ( ! response.ok ) {
+			throw new Error( 'Failed to fetch WordPress versions' );
+		}
+		const data: WordPressApiResponse = await response.json();
```
fredrikekelund: Let's add a zod schema here 🙂

src/stores/wordpress-versions-slice.ts
fredrikekelund: We could, but my two cents is that for such a simple network request, we would make things unnecessarily complex by moving it to a utility function. 

As a general comment on the `src/api` folder, I think we'll find that requesting WordPress.com APIs from hooks or Redux slices will be much easier because we can access the WPCOM client there (not yet in Redux stores, but we will be able to once `AuthProvider` has been refactored to a Redux store).

wojtekn: In such a case, would we store all API calls along with Redux stores and skip using `src/api` completely?

fredrikekelund (approved):
The code works as expected 👍 I already mentioned that we should [add a zod schema](https://github.com/Automattic/studio/pull/986/files#r1973429779) for the API response, but I'll still approve now to unblock us.

On more of a UX note, I think Playground's approach of not displaying patch versions in the WordPress version selector is good. Maybe you've already planned for that here, too, but I wanted to call it out nonetheless.

In Playground, the list is hard-coded, but we could deduce it from the API response by stripping out the patch version number.

Screenshot 2025-02-27 at 13 05 40

src/stores/wordpress-versions-slice.ts
fredrikekelund: That might actually be a good idea… This didn't strike me when I reviewed https://github.com/Automattic/studio/pull/948, but I think it'll be a recurring challenge. I also think there'll be very little API communication that's not related to front-end state.

src/stores/wordpress-versions-slice.ts
sejas: Done! https://github.com/Automattic/studio/pull/986/commits/a2304c4e6cb3f06face918fe3716578e348bc933

sejas:
> The code works as expected 👍 I already mentioned that we should [add a zod schema](https://github.com/Automattic/studio/pull/986/files#r1973429779) for the API response, but I'll still approve now to unblock us.
> 
> On more of a UX note, I think Playground's approach of not displaying patch versions in the WordPress version selector is good. Maybe you've already planned for that here, too, but I wanted to call it out nonetheless.
> 
> In Playground, the list is hard-coded, but we could deduce it from the API response by stripping out the patch version number.
> 
> Screenshot 2025-02-27 at 13 05 40

Great catch with the Zod schema. 👌

I also appreciated your suggestion about hiding the patch version from the "name" selector. I've added that in https://github.com/Automattic/studio/pull/986/commits/026e4854bd4f0b742bfb9dc1a61f2b98d3982f64.

For betas and RCs, I decided to show the full version. We can refine this later if needed.

@fredrikekelund, could you please take another look? Thank you!

src/stores/tests/wordpress-versions-slice.test.ts
```diff
@@ -0,0 +1,219 @@
...
+import { testActions, testReducer } from 'src/stores/tests/utils/test-reducer';
+import {
+	fetchWordPressVersions,
+	wordpressVersionsSelectors,
+} from 'src/stores/wordpress-versions-slice';
+
+// Mock fetch
+global.fetch = jest.fn();
+
+// Mock Sentry
```
katinthehatsite: Not sure if we need these comments as it seems to be pretty self-explanatory. Not a strong opinion

sejas: Removed! https://github.com/Automattic/studio/pull/986/commits/3409452d1d23665c7b751f29d11238c028433245

src/stores/wordpress-versions-slice.ts
```diff
@@ -0,0 +1,104 @@
+import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
+import * as Sentry from '@sentry/electron/renderer';
+import { z, ZodError } from 'zod';
+
+const MINIMUM_WORDPRESS_VERSION = '5.9.9';
```
katinthehatsite, line 5: nice, we can use this for now and reconsider later once we tested more versions and determined compatibility 👍

src/stores/wordpress-versions-slice.ts
sejas: I like the idea of keeping the API calls inside the thunks 👍

src/stores/wordpress-versions-slice.ts
```diff
@@ -0,0 +1,104 @@
...
+			const data = wordPressApiResponseSchema.parse( rawData );
+
+			const offers = data.offers
+				.filter( ( offer ) => offer.response === 'autoupdate' )
+				.map( ( offer ) => ( {
+					version: offer.version,
+					isBeta: offer.version.includes( 'beta' ) || offer.version.includes( 'RC' ),
+					name: extractVersionName( offer.version ),
+				} ) );
+			return offers;
```
fredrikekelund:
```suggestion
			return data.offers
				.filter( ( offer ) => offer.response === 'autoupdate' )
				.map( ( offer ) => ( {
					version: offer.version,
					isBeta: offer.version.includes( 'beta' ) || offer.version.includes( 'RC' ),
					name: extractVersionName( offer.version ),
				} ) );
```
Nitpick, but no need for the variable assignment

src/stores/wordpress-versions-slice.ts
```diff
@@ -0,0 +1,104 @@
...
+	versions: [],
+	status: 'idle',
+	error: null,
+};
+
+const extractVersionName = ( version: string ): string => {
+	if ( version.includes( 'beta' ) || version.includes( 'RC' ) ) {
+		return version;
+	}
+	const match = version.match( /^(\d+\.\d+)/ );
```
katinthehatsite, line 17: what happens with `name` when the version is something like `6.7.2` and `6.7.1`? would two different versions in that case end up with the same name?

src/stores/wordpress-versions-slice.ts
```diff
@@ -0,0 +1,104 @@
...
+	versions: WordPressVersion[];
+	status: 'idle' | 'loading' | 'succeeded' | 'failed';
+	error: string | null;
+}
+
+const initialState: WordPressVersionsState = {
+	versions: [],
+	status: 'idle',
+	error: null,
+};
```
fredrikekelund: Optional nit, but I'd move this part closer to the `createSlice` call. I don't think there's a universal standard, but that's what I did for the chat slice, and I think it makes sense.

src/stores/tests/wordpress-versions-slice.test.ts
fredrikekelund, line 1: Such a joy to test Redux reducers 👍

fredrikekelund (approved):
LGTM

src/stores/wordpress-versions-slice.ts
fredrikekelund: Good catch. This needs to be resolved somehow

katinthehatsite (approved):
I am approving the PR as the changes look good. Nice work!

There is mostly one concern that I have about versions name matching.

src/stores/wordpress-versions-slice.ts
sejas: Moved! https://github.com/Automattic/studio/pull/986/commits/0b10974366ad94151e60734cf93204b26122ca4e

src/stores/wordpress-versions-slice.ts
sejas: Removed https://github.com/Automattic/studio/pull/986/commits/6a1a4f04c997501d91e57d5c3ffd74f381727dff

fredrikekelund:
I created https://github.com/Automattic/studio/pull/1003 to address the Sentry issue

PR #967, opened 2025-02-21
Studio: Fix text formatting on Preview and Sync tabs when the user is logged out

ivan-ottinger:
## Related issues

- fixes https://github.com/Automattic/dotcom-forge/issues/10516
- related discussion: p1740153536076449/1740142244.756879-slack-C06DRMD6VPZ

## Proposed Changes

- fix text formatting on the `Sync` and `Previews` tab pages when the user is logged out:

| Before | After |
|--------|--------|
| Markup on 2025-02-21 at 17:01:07 | Markup on 2025-02-21 at 17:00:34 |
| Markup on 2025-02-21 at 17:01:20 | Markup on 2025-02-21 at 17:00:46 | 

## Testing Instructions

1. Check out the PR branch and build the app with `npm start`.
2. Ensure you are not logged in to your WPCOM account in Studio.
3. Head over to the `Sync` and `Previews` tab when there are no synced sites connected / no preview sites created.
4. The formatting of the bottom text should be correct and the clicking the link should work correctly as well. 

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #987, opened 2025-02-27
Sync: Remove unexpected horizontal scrollbar on the tab when the user is logged out

ivan-ottinger:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10522.

## Proposed Changes

- use `max-width` instead of `width` for the "New to WordPress.com?" section on the Sync tab

## Testing Instructions

1. Try to reproduce the https://github.com/Automattic/dotcom-forge/issues/10522 issue.
2. Check out the PR branch and build the app with `npm start`.
3. Try to reproduce the issue again. It should not be possible.

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-sync.tsx
+++ b/src/components/content-tab-sync.tsx
@@ -78,7 +78,7 @@ function NoAuthSyncTab() {
 					</Button>
 				</Tooltip>
 			</div>
-			<div className="mt-3 max-w-[40ch] text-a8c-gray-70 a8c-body">
+			<div className="mt-3 text-a8c-gray-70 a8c-body">
 				<Tooltip
 					disabled={ ! isOffline }
 					icon={ offlineIcon }
--- a/src/components/content-tab-sync.tsx
+++ b/src/components/content-tab-sync.tsx
@@ -78,7 +78,7 @@ function NoAuthSyncTab() {
 					</Button>
 				</Tooltip>
 			</div>
-			<div className="mt-3 w-[40ch] text-a8c-gray-70 a8c-body">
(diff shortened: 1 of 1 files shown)


src/components/content-tab-sync.tsx
```diff
@@ -78,7 +78,7 @@ function NoAuthSyncTab() {
 					</Button>
 				</Tooltip>
 			</div>
-			<div className="mt-3 w-[40ch] text-a8c-gray-70 a8c-body">
+			<div className="mt-3 max-w-[40ch] text-a8c-gray-70 a8c-body">
```
fredrikekelund:
```suggestion
			<div className="mt-3 text-a8c-gray-70 a8c-body">
```
Setting an explicit width on a text element where the width is already constrained by the parent seems odd to me. 

I think the better solution here is to drop this declaration.

katinthehatsite (approved):
The fix worked for me so I am approving the change although we might want to consider Fredrik's suggestion.

src/components/content-tab-sync.tsx
ivan-ottinger: That's a great catch, Fredrik!

I couldn't find any justification for keeping the width there and the formatting looks good without it. Even when testing with different locales and screen sizes.

It couple be some leftover from the past. I could see some changes to the formatting when testing with the old code that was already replaced in https://github.com/Automattic/studio/pull/967.

Anyways, applying your suggestion.

ivan-ottinger:
Thank you for your review and testing, Kat and Fredrik!

PR #992, opened 2025-02-28
Fix replacing site url in DB after pulling site

nightnei:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10428

## Proposed Changes
When user pull or puh a site to/from Studio - all links should reflect the actual domain, all of them should be `localhost:<port>` if it was pulling to Studio or actual domain on dotcom if it was pushing.
This PR handles improving pulling process. We had it before, but I found not covered cases, so this PR fixes them
Regarding pushing - we don't handle it at all. I was thinking about it end ended up that it's better to do it in VaultPress side (details - p1740755830254999-slack-C07Q6R5HW3T).

Examples of issues covered in this PR:
Screenshot 2025-02-28 at 15 05 15
Screenshot 2025-02-28 at 15 07 47

## Testing Instructions
1. Create a site in wp.com
2. Create a post with video and add poster to this video
3. Install Elementor plugin
4. Create a post with the help of Elementor, add video and poster to it
5. Open Studio -> Create a new site -> Pull site from wp.com, created in the 1-st step of this instructions
6. Open local site
7. Assert that posters mentioned on screenshots in "Proposed Changes" have the correct localhost domain

### DIFF
--- a/src/lib/updateSiteUrlToLocal.ts
+++ b/src/lib/updateSiteUrlToLocal.ts
@@ -19,7 +19,16 @@ export const updateSiteUrlToLocal = async ( siteId: string ) => {
 	const oldUrl = currentSiteUrl.trim();
 	const urlWithoutProtocol = oldUrl.replace( /^https?:\/\//, '' );
 
-	const oldUrlVariants = [ `http://${ urlWithoutProtocol }`, `https://${ urlWithoutProtocol }` ];
+	const oldUrlVariants = [
+		`http://${ urlWithoutProtocol }`,
+		`https://${ urlWithoutProtocol }`,
+		// e.g. "posterUrl" for videos uses encoded URLs
+		`http%3A%2F%2F${ urlWithoutProtocol }`,
+		`https%3A%2F%2F${ urlWithoutProtocol }`,
+		// e.g. Elementor plugin uses escaped URLs
+		String.raw`http:\/\/${ urlWithoutProtocol }`,
+		String.raw`https:\/\/${ urlWithoutProtocol }`,
+	];
 
 	for ( const urlToReplace of oldUrlVariants ) {
 		const { stderr, exitCode } = await server.executeWpCliCommand(


src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -95,8 +95,6 @@ export function useSyncPull( {
 				return;
 			}
 
-			console.groupCollapsed( 'Sync Pull' );
```
wojtekn: If you want to clean up the logging, could you consider opening a separate PR for that?

src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -174,28 +169,25 @@ export function useSyncPull( {
 						cancelId: 1,
 					} );
 
-					if ( userChoice === 1 ) {
-						console.log( 'User cancelled pull operation' );
+					const isCanceledByUser = userChoice === 1;
+					if ( isCanceledByUser ) {
```
wojtekn: Do we need this variable if it's used only once?

src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -291,15 +277,14 @@ export function useSyncPull( {
 					pullStatesProgressInfo[ frontendStatus ] || pullStatesProgressInfo.failed;
 				const downloadUrl = hasBackupCompleted ? response.download_url : null;
 
-				if ( hasBackupCompleted && downloadUrl ) {
+				if ( downloadUrl ) {
```
wojtekn: What's the context of this removal?

src/lib/updateSiteUrlToLocal.ts
```diff
@@ -19,7 +19,16 @@ export const updateSiteUrlToLocal = async ( siteId: string ) => {
 	const oldUrl = currentSiteUrl.trim();
 	const urlWithoutProtocol = oldUrl.replace( /^https?:\/\//, '' );
 
-	const oldUrlVariants = [ `http://${ urlWithoutProtocol }`, `https://${ urlWithoutProtocol }` ];
+	const oldUrlVariants = [
```
wojtekn, line 22: Looks correct, nice catch.

wojtekn:
@nightnei, it worked as expected for me - IMG had a local URL after the pull.

However, I couldn't make the video load from wp-admin, and it always had the videopress URL. How could I test this part?

src/hooks/sync-sites/use-sync-pull.ts
nightnei: The above line has `const downloadUrl = hasBackupCompleted ? response.download_url : null;` - `downloadUrl ` is available only if `hasBackupCompleted ` is `true`. So it's duplication of the condition.

src/hooks/sync-sites/use-sync-pull.ts
nightnei: I just wanted to avoid "magic" number `1`, since at a glance I thought that `1` means kinda `true`, as a confirmation to continue pulling, but actually it's canceling.

So to keep readability, we need either to add comment about `1` or create readable and simple variable. And the good variable name is much better than any comment.

nightnei:
>However, I couldn't make the video load from wp-admin, and it always had the videopress URL. How could I test this part?

@wojtekn hm, maybe it's the same case as my assumption here? p1741020425670679/1740755830.254999-slack-C07Q6R5HW3T

Could you please check and ensure that you are adding a newly uploaded image as a poster, instead of some image in your library. If it's not the case, then I will think more to try to understand why it works for you differently. Or, as an option, could you please share that website with me (nei.css@gmail.com) with Admin access and I will take a look.

nightnei:
@wojtekn I revert changes for console.logs etc, I will create separate PR for them, but anyway could you please take a look at my answers regarding your feedback.
UPDATE: Hovewer up to you, if you prefer to comment these lines in that separate PR then I totally don't mind :)

wojtekn:
> Could you please check and ensure that you are adding a newly uploaded image as a poster, instead of some image in your library.

Could you show a screenshot that illustrates what is a poster in Elementor?

nightnei:
> Could you show a screenshot that illustrates what is a poster in Elementor?

@wojtekn  I decided to record video, to make it more simple for you and show in real time what exactly I did:

https://github.com/user-attachments/assets/1641223a-b989-46df-a622-3374d0eeaa24

wojtekn:
@nightnei I've reproduced the case with the video's poster URL not being replaced and confirmed that PR fixes the issue.

wojtekn (approved):

PR #993, opened 2025-02-28
Update tooltips in the new Previews tab

ndiego:
## Proposed Changes

This PR contains a few minute updates to the tooltips used on the new Previews tab. 

- Remove the tooltip for the site URL. It provides no new information beyond the link that's already visible. 
- Add a date/time tooltip for the Updated value. This provides users with the exact date/time their preview was last updated.

## Testing Instructions

| Preview URL - Before |  Preview URL - After |
|-|-|
|image|image|

| Updated - Before |  Updated - After |
|-|-|
|image|image|

## Pre-merge Checklist

### DIFF
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -220,7 +220,7 @@ const SyncConnectedSitesSection = ( {
 									<Badge className="bg-a8c-green-5 text-a8c-green-80">{ __( 'Production' ) }</Badge>
 								) }
 							</div>
-							<Tooltip 
+							<Tooltip
 								text={ sprintf(
 									/* translators: %s: The URL of the connected site */
 									__( 'Open %s' ),
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -220,8 +220,14 @@ const SyncConnectedSitesSection = ( {
 									<Badge className="bg-a8c-green-5 text-a8c-green-80">{ __( 'Production' ) }</Badge>
 								) }
 							</div>
-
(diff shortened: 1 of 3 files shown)


matt-west:
Nice change. Thanks @ndiego!

ivan-ottinger:
Thanks for the PR, Nick! I like the tooltip on the values in the _Updated_ column

> Remove the tooltip for the site URL. It provides no new information beyond the link that's already visible.

We may want to display the tooltip at least in cases when there's not enough space for the whole URL to display:

Markup on 2025-03-03 at 09:41:19

At the moment we are also displaying permanent tooltips on sites in the Sync tab.

wojtekn:
> We may want to display the tooltip at least in cases when there's not enough space for the whole URL to display:

It would make sense to bring those back and display the URL, and this is one of the reasons. Maybe we should make the tooltip consistent with other tooltips and change it to say "Open https://.."?

ndiego:
> We may want to display the tooltip at least in cases when there's not enough space for the whole URL to display:

Oh, great point. I didn't think of that. 

> It would make sense to bring those back and display the URL, and this is one of the reasons. Maybe we should make the tooltip consistent with other tooltips and change it to say "Open https://.."?

I like this as well! I can go ahead and make this change. Sound good?

wojtekn:
@ndiego sounds great!

ndiego:
Tooltips have been updated. Let me know if any additional changes need to be made @wojtekn. Thanks!

wojtekn (approved):

wojtekn:
@ndiego looks great. Tooltips are consistent now. 🥳

fredrikekelund (approved):
Approving to try and trigger CI

PR #994, opened 2025-03-02
STU-22: Add custom local domain support for WordPress sites

### LINKED ISSUE #215: Feature Request: Router mode request (localhost and site domain)
### What

Need router mode for prettier link like in local flywheel:

- Site Domains Router Mode
     The default mode that Local runs. In this mode, Local will start a “router” that listens on port 80 and directs traffic to the individual sites that Local manages. The domain for each site will look something like example.local

- localhost Router Mode

    This fallback mode doesn’t start the Local router and instead, each site is reached directly using a dedicated port on the computer. Using this mode means that the domain will look something like localhost:10005

### Why

This will helps developer to choose between both mode.

### How

_No response_

youknowriad:
## Related issues

- Fixes #215

## Summary            …

- Implemented custom domain support for WordPress sites using a proxy server
- Added UI for setting custom domains when creating sites
- Integrated with system hosts file for local domain resolution
- The user chooses a custom domain on site creation (editing an existing site is harder because it would require updates to the WP database).

## Implementation details

- Created a proxy server that listens on port 80 and forwards requests to the appropriate local site. This means that if 80 is already used, custom domains won't work (there's no way around this)
- Added domain management in hosts file to resolve custom domains to localhost
- Updated site details to store and display custom domain information
- Modified URL display in settings to show custom domain when enabled
- Added validation for custom domain format

## Test plan

- Create a new site with custom domain enabled
- Verify the domain is added to hosts file (requires admin permission)
- Visit the site using the custom domain (without port)
- Verify admin and frontend URLs work properly
- Delete the site and verify domain is removed from hosts file

## Pre-merge Checklist

**Important note** This PR has been created for a huge part using AI. My main purpose with this PR was to learn both Studio's code base and learn AI while implementing a useful feature.

Screenshot 2025-03-02 at 2 20 18 PM

### DIFF
--- a/src/components/content-tab-settings.tsx
+++ b/src/components/content-tab-settings.tsx
@@ -56,26 +56,22 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) 
 						</div>
 					</SettingsRow>
 					<SettingsRow label={ __( 'Local URL' ) }>
-						<div className="flex">
-							<CopyTextButton
-								text={ `http://${ domain }` }
-								label={ `${ domain }, ${ __( 'Copy site url to clipboard' ) }` }
-								copyConfirmation={ __( 'Copied!' ) }
-							>
-								{ domain }
-							</CopyTextButton>
-						</div>
+						<CopyTextButton
+							text={ `http://${ domain }` }
+							label={ `${ domain }, ${ __( 'Copy site url to clipboard' ) }` }
+							copyConfirmation={ __( 'Copied!' ) }
+						>
+							{ domain }
+						</CopyTextButton>
 					</SettingsRow>
 					<SettingsRow label={ __( 'Local path' ) }>
-						<div className="flex">
-							<CopyTextButton
-								text={ selectedSite.path }
-								label={ __( 'Copy local path to clipboard' ) }
-								copyConfirmation={ __( 'Copied!' ) }
-							>
-								<span className="line-clamp-1 break-all">{ selectedSite.path }</span>
-							</CopyTextButton>
-						</div>
+						<CopyTextButton
+							text={ selectedSite.path }
+							label={ __( 'Copy local path to clipboard' ) }
+							copyConfirmation={ __( 'Copied!' ) }
+						>
+							<span className="line-clamp-1 break-all">{ selectedSite.path }</span>
+						</CopyTextButton>
 					</SettingsRow>
 					<SettingsRow label={ __( 'WP Version' ) }>{ wpVersion }</SettingsRow>
 					<SettingsRow label={ __( 'PHP Version' ) }>
--- a/src/components/content-tab-settings.tsx
+++ b/src/components/content-tab-settings.tsx
@@ -32,10 +32,9 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) 
 	const storedPassword = decodePassword( selectedSite.adminPassword ?? '' );
 	const password = storedPassword === '' ? 'password' : storedPassword;
 	const [ wpVersion, refreshWpVersion ] = useGetWpVersion( selectedSite );
-	const domain =
-		selectedSite.useCustomDomain && selectedSite.customDomain
-			? `${ selectedSite.customDomain }`
-			: `localhost:${ selectedSite.port }`;
+	const domain = selectedSite.customDomain
+		? `${ selectedSite.customDomain }`
+		: `localhost:${ selectedSite.port }`;
 	return (
 		<div className="p-8">
 			<table className="mb-2 m-w-full" cellPadding={ 0 } cellSpacing={ 0 }>
--- a/src/components/content-tab-settings.tsx
+++ b/src/components/content-tab-settings.tsx
@@ -32,6 +32,10 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) 
 	const storedPassword = decodePassword( selectedSite.adminPassword ?? '' );
 	const password = storedPassword === '' ? 'password' : storedPassword;
 	const [ wpVersion, refreshWpVersion ] = useGetWpVersion( selectedSite );
+	const domain =
+		selectedSite.useCustomDomain && selectedSite.customDomain
+			? `${ selectedSite.customDomain }`
+			: `localhost:${ selectedSite.port }`;
 	return (
 		<div className="p-8">
 			<table className="mb-2 m-w-full" cellPadding={ 0 } cellSpacing={ 0 }>
@@ -53,13 +57,15 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) 
 						</div>
 					</SettingsRow>
 					<SettingsRow label={ __( 'Local URL' ) }>
-						<CopyTextButton
-							text={ `http://localhost:${ selectedSite.port }` }
-							label={ `localhost:${ selectedSite.port }, ${ __( 'Copy site url to clipboard' ) }` }
-							copyConfirmation={ __( 'Copied!' ) }
-						>
-							{ `localhost:${ selectedSite.port }` }
-						</CopyTextButton>
+						<div className="flex">
+							<CopyTextButton
+								text={ `http://${ domain }` }
+								label={ `${ domain }, ${ __( 'Copy site url to clipboard' ) }` }
+								copyConfirmation={ __( 'Copied!' ) }
+							>
+								{ domain }
+							</CopyTextButton>
+						</div>
 					</SettingsRow>
 					<SettingsRow label={ __( 'Local path' ) }>
 						<div className="flex">
@@ -105,13 +111,11 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) 
 					</SettingsRow>
 					<SettingsRow label={ __( 'Admin URL' ) }>
 						<CopyTextButton
-							text={ `http://localhost:${ selectedSite.port }/wp-admin` }
-							label={ `localhost:${ selectedSite.port }/wp-admin, ${ __(
-								'Copy wp-admin url to clipboard'
-							) }` }
+							text={ `http://${ domain }/wp-admin` }
+							label={ `${ domain }/wp-admin, ${ __( 'Copy wp-admin url to clipboard' ) }` }
 							copyConfirmation={ __( 'Copied!' ) }
 						>
-							{ `localhost:${ selectedSite.port }/wp-admin` }
+							{ `${ domain }/wp-admin` }
 						</CopyTextButton>
 					</SettingsRow>
 				</tbody>
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -12,6 +12,7 @@ import { ACCEPTED_IMPORT_FILE_TYPES } from 'src/constants';
 import { useDocsLink } from 'src/hooks/use-docs-link';
 import { useFeatureFlags } from 'src/hooks/use-feature-flags';
 import { cx } from 'src/lib/cx';
+import { generateCustomDomainFromSiteName } from 'src/lib/generate-custom-domain';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { useAppDispatch, useRootSelector } from 'src/stores';
 import {
@@ -85,20 +86,6 @@ interface SiteFormWithoutVersionsProps extends SiteFormBaseProps {
 }
 
 type SiteFormProps = SiteFormWithVersionsProps | SiteFormWithoutVersionsProps;
-
-/**
- * Generates a suitable domain name from site name
- */
-const generateDomainFromSiteName = ( siteName: string ): string => {
-	// Convert the site name to lowercase and replace spaces with hyphens
-	const domainBase = siteName
-		.toLowerCase()
-		.replace( /[^a-z0-9-]/g, '-' ) // Replace non-alphanumeric chars with hyphens
-		.replace( /-+/g, '-' ) // Replace multiple hyphens with single hyphen
-		.replace( /^-|-$/g, '' ); // Remove hyphens from start and end
-
-	return `${ domainBase }.wp.local`;
-};
 
 const SiteFormError = ( { error, tipMessage = '', className = '' }: SiteFormErrorProps ) => {
 	return (
@@ -318,6 +305,7 @@ export const SiteForm = ( {
 	} else {
 		chevronIcon = chevronRight;
 	}
+	const generatedDomainName = generateCustomDomainFromSiteName( siteName );
 
 	return (
 		<form className={ className } onSubmit={ onSubmit }>
@@ -332,10 +320,7 @@ export const SiteForm = ( {
 							type="checkbox"
 							id="use-custom-domain"
 							checked={ useCustomDomain }
-							onChange={ ( e ) => {
-								setUseCustomDomain( e.target.checked );
-								setCustomDomain( generateDomainFromSiteName( siteName ) );
-							} }
+							onChange={ ( e ) => setUseCustomDomain( e.target.checked ) }
 						/>
 						<label htmlFor="use-custom-domain">{ __( 'Use custom domain' ) }</label>
 						<span className="text-a8c-gray-50 text-xs ml-2">
@@ -353,7 +338,7 @@ export const SiteForm = ( {
 							id="custom-domain"
 							value={ customDomain ?? '' }
 							onChange={ setCustomDomain }
-							placeholder="my-site.wp.local"
+							placeholder={ generatedDomainName }
 						/>
 						{ customDomainError && (
 							<div className="text-red-500 text-xs">{ customDomainError }</div>
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -314,47 +314,7 @@ export const SiteForm = ( {
 					<span className="font-semibold">{ __( 'Site name' ) }</span>
 					<TextControlComponent onChange={ setSiteName } value={ siteName }></TextControlComponent>
 				</label>
-				{ setUseCustomDomain && setCustomDomain && (
-					<div className="flex items-center gap-2 mb-6">
-						<input
-							type="checkbox"
-							id="use-custom-domain"
-							checked={ useCustomDomain }
-							onChange={ ( e ) => setUseCustomDomain( e.target.checked ) }
-						/>
-						<label htmlFor="use-custom-domain">{ __( 'Use custom domain' ) }</label>
-						<span className="text-a8c-gray-50 text-xs ml-2">
-							{ __( 'This will require your password to authorize changes to the hosts file.' ) }
-						</span>
-					</div>
-				) }
 
-				{ useCustomDomain && setCustomDomain && (
-					<div className="flex flex-col gap-2 mb-6">
-						<label htmlFor="custom-domain" className="font-semibold">
-							{ __( 'Domain name' ) }
-						</label>
-						<TextControlComponent
-							id="custom-domain"
-							value={ customDomain ?? '' }
-							onChange={ setCustomDomain }
-							placeholder={ generatedDomainName }
-						/>
-						{ customDomainError && (
-							<div className="text-red-500 text-xs">{ customDomainError }</div>
-						) }
-						<div className="text-a8c-gray-50 text-xs">
-							{ __(
-								'You will be prompted for administrator permission to modify your system hosts file.'
-							) }
-						</div>
-						<div className="text-a8c-gray-50 text-xs mt-1">
-							{ __(
-								'Note: You will need administrator privileges to enable direct domain access.'
-							) }
-						</div>
-					</div>
-				) }
 				{ setFileForImport && (
 					<>
 						<div className="flex flex-col gap-1.5 leading-4 mb-6">
@@ -414,7 +374,7 @@ export const SiteForm = ( {
 								</div>
 								<div
 									className={ cx(
-										'transition-all duration-500 ease-in-out overflow-hidden',
+										'transition-all duration-500 ease-in-out overflow-hidden flex flex-col gap-2',
 										isAdvancedSettingsVisible ? 'max-h-96 opacity-100' : 'max-h-0 opacity-0'
 									) }
 								>
@@ -482,6 +442,62 @@ export const SiteForm = ( {
 											</div>
 										) }
 									</div>
+
+									{ setUseCustomDomain && setCustomDomain && (
+										<div
+											className={ cx(
+												'flex items-center gap-2',
+												isAdvancedSettingsVisible ? 'py-2' : 'p-2',
+												! isAdvancedSettingsVisible && 'hidden'
+											) }
+										>
+											<input
+												type="checkbox"
+												id="use-custom-domain"
+												checked={ useCustomDomain }
+												onChange={ ( e ) => setUseCustomDomain( e.target.checked ) }
+											/>
+											<label htmlFor="use-custom-domain">{ __( 'Use custom domain' ) }</label>
+											<span className="text-a8c-gray-50 text-xs ml-2">
+												{ __(
+													'This will require your password to authorize changes to the hosts file.'
+												) }
+											</span>
+										</div>
+									) }
+
+									{ useCustomDomain && setCustomDomain && (
+										<div
+											className={ cx(
+												'flex flex-col gap-2',
+												isAdvancedSettingsVisible ? 'py-2' : 'p-2',
+												! isAdvancedSettingsVisible && 'hidden'
+											) }
+										>
+											<label htmlFor="custom-domain" className="font-semibold">
+												{ __( 'Domain name' ) }
+											</label>
+											<TextControlComponent
+												id="custom-domain"
+												value={ customDomain ?? '' }
+												onChange={ setCustomDomain }
+												placeholder={ generatedDomainName }
+											/>
+											{ customDomainError && (
+												<div className="text-red-500 text-xs">{ customDomainError }</div>
+											) }
+											<div className="text-a8c-gray-50 text-xs">
+												{ __(
+													'You will be prompted for administrator permission to modify your system hosts file.'
+												) }
+											</div>
+											<div className="text-a8c-gray-50 text-xs mt-1">
+												{ __(
+													'Note: You will need administrator privileges to enable direct domain access.'
+												) }
+											</div>
+										</div>
+									) }
 								</div>
 							</>
 						) }
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -365,10 +365,12 @@ export const SiteForm = ( {
 											{ __( 'Advanced settings' ) }
 										</div>
 									</Button>
-									{ error && (
+									{ ( error || customDomainError ) && (
 										<span className="text-red-500 text-[13px] leading-[16px] ml-2 flex items-center">
 											<Icon icon={ warning } size={ 16 } className="mr-1 fill-red-500" />
-											{ __( '1 error found' ) }
+											{ error && customDomainError
+												? __( '2 errors found' )
+												: __( '1 error found' ) }
 										</span>
 									) }
 								</div>
@@ -483,9 +485,7 @@ export const SiteForm = ( {
 												onChange={ setCustomDomain }
 												placeholder={ generatedDomainName }
 											/>
-											{ customDomainError && (
-												<div className="text-red-500 text-xs">{ customDomainError }</div>
-											) }
+											{ customDomainError && <SiteFormError error={ customDomainError } /> }
 											<div className="text-a8c-gray-50 text-xs">
 												{ __(
 													'You will be prompted for administrator permission to modify your system hosts file.'
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -58,6 +58,11 @@ interface SiteFormBaseProps {
 	setFileForImport?: ( file: File | null ) => void;
 	onFileSelected?: ( file: File ) => void;
 	fileError?: string;
+	useCustomDomain?: boolean;
+	setUseCustomDomain?: ( use: boolean ) => void;
+	customDomain?: string;
+	setCustomDomain?: ( domain: string ) => void;
+	customDomainError?: string;
 }
 
 // TODO: Add all this props to the SiteForm component once we have the versions selects in edit site page.
@@ -80,6 +85,20 @@ interface SiteFormWithoutVersionsProps extends SiteFormBaseProps {
 }
 
 type SiteFormProps = SiteFormWithVersionsProps | SiteFormWithoutVersionsProps;
+
+/**
+ * Generates a suitable domain name from site name
+ */
+const generateDomainFromSiteName = ( siteName: string ): string => {
+	// Convert the site name to lowercase and replace spaces with hyphens
+	const domainBase = siteName
+		.toLowerCase()
+		.replace( /[^a-z0-9-]/g, '-' ) // Replace non-alphanumeric chars with hyphens
+		.replace( /-+/g, '-' ) // Replace multiple hyphens with single hyphen
+		.replace( /^-|-$/g, '' ); // Remove hyphens from start and end
+
+	return `${ domainBase }.wp.local`;
+};
 
 const SiteFormError = ( { error, tipMessage = '', className = '' }: SiteFormErrorProps ) => {
 	return (
@@ -263,6 +282,11 @@ export const SiteForm = ( {
 	onFileSelected,
 	fileError,
 	allowVersionsChange = false,
+	useCustomDomain,
+	setUseCustomDomain,
+	customDomain,
+	setCustomDomain,
+	customDomainError,
 }: SiteFormProps ) => {
 	const { __, isRTL } = useI18n();
 	const getDocsLink = useDocsLink();
@@ -302,6 +326,47 @@ export const SiteForm = ( {
 					<span className="font-semibold">{ __( 'Site name' ) }</span>
 					<TextControlComponent onChange={ setSiteName } value={ siteName }></TextControlComponent>
 				</label>
+				{ setUseCustomDomain && setCustomDomain && (
+					<div className="flex items-center gap-2 mb-6">
+						<input
+							type="checkbox"
+							id="use-custom-domain"
+							checked={ useCustomDomain }
+							onChange={ ( e ) => {
+								setUseCustomDomain( e.target.checked );
+								setCustomDomain( generateDomainFromSiteName( siteName ) );
+							} }
+						/>
+						<label htmlFor="use-custom-domain">{ __( 'Use custom domain' ) }</label>
+					</div>
+				) }
+
+				{ useCustomDomain && setCustomDomain && (
+					<div className="flex flex-col gap-2 mb-6">
+						<label htmlFor="custom-domain" className="font-semibold">
+							{ __( 'Domain name' ) }
+						</label>
+						<TextControlComponent
+							id="custom-domain"
+							value={ customDomain ?? '' }
+							onChange={ setCustomDomain }
+							placeholder="my-site.wp.local"
+						/>
+						{ customDomainError && (
+							<div className="text-red-500 text-xs">{ customDomainError }</div>
+						) }
+						<div className="text-a8c-gray-50 text-xs">
+							{ __(
+								'You will be prompted for administrator permission to modify your system hosts file.'
+							) }
+						</div>
+						<div className="text-a8c-gray-50 text-xs mt-1">
+							{ __(
+								'Note: You will need administrator privileges to enable direct domain access.'
+							) }
+						</div>
+					</div>
+				) }
 				{ setFileForImport && (
 					<>
 						<div className="flex flex-col gap-1.5 leading-4 mb-6">
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -443,16 +443,9 @@ export const SiteForm = ( {
 												</div>
 											</div>
 										) }
-									</div>
 
-									{ setUseCustomDomain && setCustomDomain && (
-										<div
-											className={ cx(
-												isAdvancedSettingsVisible ? 'py-2' : 'p-2',
-												! isAdvancedSettingsVisible && 'hidden'
-											) }
-										>
-											<div className="flex items-center gap-2">
+										{ setUseCustomDomain && setCustomDomain && (
+											<div className="flex items-center gap-2 mt-4">
 												<input
 													type="checkbox"
 													id="use-custom-domain"
@@ -461,33 +454,28 @@ export const SiteForm = ( {
 												/>
 												<label htmlFor="use-custom-domain">{ __( 'Use custom domain' ) }</label>
 											</div>
-											<div className="text-a8c-gray-50 text-xs pt-2">
-												{ __(
-													'This will require your password to authorize changes to the hosts file.'
-												) }
-											</div>
-										</div>
-									) }
+										) }
 
-									{ useCustomDomain && setCustomDomain && (
-										<div
-											className={ cx(
-												'flex flex-col gap-2',
-												isAdvancedSettingsVisible ? 'py-2' : 'p-2',
-												! isAdvancedSettingsVisible && 'hidden'
-											) }
-										>
-											<label htmlFor="custom-domain" className="font-semibold">
-												{ __( 'Domain name' ) }
-											</label>
-											<TextControlComponent
-												id="custom-domain"
-												value={ customDomain !== null ? customDomain : generatedDomainName }
-												onChange={ setCustomDomain }
-											/>
-											{ customDomainError && <SiteFormError error={ customDomainError } /> }
-										</div>
-									) }
+										{ useCustomDomain && setCustomDomain && (
+											<div className="flex flex-col gap-2 mt-4">
+												<label htmlFor="custom-domain" className="font-semibold">
+													{ __( 'Domain name' ) }
+												</label>
+												<TextControlComponent
+													id="custom-domain"
+													value={ customDomain !== null ? customDomain : generatedDomainName }
+													onChange={ setCustomDomain }
+												/>
+												{ customDomainError && <SiteFormError error={ customDomainError } /> }
+											</div>
+										) }
+
+										{ setUseCustomDomain && setCustomDomain && (
+											<div className="text-a8c-gray-50 text-xs mt-2">
+												{ __( 'Your system password will be required to set up the domain.' ) }
+											</div>
+										) }
+									</div>
 								</div>
 							</>
 						) }
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -338,6 +338,9 @@ export const SiteForm = ( {
 							} }
 						/>
 						<label htmlFor="use-custom-domain">{ __( 'Use custom domain' ) }</label>
+						<span className="text-a8c-gray-50 text-xs ml-2">
+							{ __( 'This will require your password to authorize changes to the hosts file.' ) }
+						</span>
 					</div>
 				) }
 
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -61,7 +61,7 @@ interface SiteFormBaseProps {
 	fileError?: string;
 	useCustomDomain?: boolean;
 	setUseCustomDomain?: ( use: boolean ) => void;
-	customDomain?: string;
+	customDomain?: string | null;
 	setCustomDomain?: ( domain: string ) => void;
 	customDomainError?: string;
 }
@@ -271,7 +271,7 @@ export const SiteForm = ( {
 	allowVersionsChange = false,
 	useCustomDomain,
 	setUseCustomDomain,
-	customDomain,
+	customDomain = null,
 	setCustomDomain,
 	customDomainError,
 }: SiteFormProps ) => {
@@ -482,9 +482,8 @@ export const SiteForm = ( {
 											</label>
 											<TextControlComponent
 												id="custom-domain"
-												value={ customDomain ?? '' }
+												value={ customDomain !== null ? customDomain : generatedDomainName }
 												onChange={ setCustomDomain }
-												placeholder={ generatedDomainName }
 											/>
 											{ customDomainError && <SiteFormError error={ customDomainError } /> }
 										</div>
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -448,23 +448,24 @@ export const SiteForm = ( {
 									{ setUseCustomDomain && setCustomDomain && (
 										<div
 											className={ cx(
-												'flex items-center gap-2',
 												isAdvancedSettingsVisible ? 'py-2' : 'p-2',
 												! isAdvancedSettingsVisible && 'hidden'
 											) }
 										>
-											<input
-												type="checkbox"
-												id="use-custom-domain"
-												checked={ useCustomDomain }
-												onChange={ ( e ) => setUseCustomDomain( e.target.checked ) }
-											/>
-											<label htmlFor="use-custom-domain">{ __( 'Use custom domain' ) }</label>
-											<span className="text-a8c-gray-50 text-xs ml-2">
+											<div className="flex items-center gap-2">
+												<input
+													type="checkbox"
+													id="use-custom-domain"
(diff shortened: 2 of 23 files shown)


src/lib/proxy-server.ts
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * Attempts to start the proxy server with elevated privileges
+ */
+async function startProxyWithElevation(): Promise< boolean > {
+	const currentPlatform = platform();
+
+	try {
+		console.log( 'Attempting to start proxy server with elevated privileges' );
+
+		if ( currentPlatform === 'win32' ) {
```
youknowriad: Apparently in Windows, we need to run the proxy with admin rights. The AI proposed using `netsh` to forward to a custom port. I need someone to help me test Windows.

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,189 @@
...
+			} catch ( error ) {
+				console.error( 'Failed to set up port forwarding:', error );
+				// Even though port forwarding failed, we still have a working proxy on 8880
+				return proxyStarted;
+			}
+		} else {
+			// For Mac/Linux: Use sudo to run a command that starts a proxy directly on port 80
+			// This is a more complex solution that would require a proper script
+			// For now, we'll just return false and let the app fall back to non-privileged ports
+			console.log( 'Elevated privileges for Mac/Linux not fully implemented yet' );
```
youknowriad: In MacOS I didn't really need elevated privileges to run the proxy in my tests. But worth testing more.

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,123 @@
+import fs from 'fs';
+import { platform } from 'os';
+import { promisify } from 'util';
+import sudo from 'sudo-prompt';
```
mcsf, line 7: Should it be [`@vscode/sudo-prompt`](https://www.npmjs.com/package/@vscode/sudo-prompt)? The [former](https://www.npmjs.com/package/sudo-prompt) is reported deprecated.

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,123 @@
...
+			} );
+		} else {
+			// For Linux, we also need sudo
+			const tempPath = '/tmp/wp-studio-hosts';
+			await writeFile( tempPath, content );
+			// @ts-expect-error promisify doesn't seem typed properly.
+			await sudoExec( `cat ${ tempPath } > ${ hostsPath }`, {
+				name: 'WordPress Studio',
+			} );
+		}
```
mcsf: Aren't these two branches identical?

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,123 @@
...
+
+		// Check if the domain is already in the hosts file
+		const domainRegex = new RegExp( `127\\.0\\.0\\.1\\s+${ domain.replace( /\./g, '\\.' ) }`, 'i' );
+		if ( domainRegex.test( hostsContent ) ) {
+			return; // Domain already exists
+		}
+
+		// Add the domain entry with comment showing the port for reference
+		const newEntry = `\n# Added by WordPress Studio (Port: ${ port })\n127.0.0.1\t${ domain }`;
+		const newContent = hostsContent + newEntry;
```
mcsf: I wonder if we could group all Studio entries together rather than just append them. This will sound like such a picky detail, but some people (I included) use their hosts files as blocklists for the whole Internet (ads, tracking, etc.), which means these files can get quite big. In this scenario, I would be a nice detail if the Studio entries could just be placed wherever the first Studio comment is found. Then a hosts file might look like:

```
# The usual system stuff
127.0.0.1        localhost
255.255.255.255  broadcasthost
::1              localhost
fe80::1%lo0      localhost

# A dev's custom definitions
127.0.0.1 mysandbox.local

# Studio's entries
# BEGIN WordPress.com Studio
127.0.0.1 blablabla.bla
# END WordPress.com Studio

# The long list of blocks
0.0.0.0 baddomain1
0.0.0.0 baddomain2
0.0.0.0 baddomain...
0.0.0.0 baddomain2000

File: src/site-server.ts [hunk 6]
```diff
@@ -86,7 +98,16 @@ export class SiteServer {
 			siteTitle: this.details.name,
 			php: this.details.phpVersion,
 		} );
-		const absoluteUrl = `http://localhost:${ this.details.port }`;
+		// Determine the URL to use - either custom domain or localhost with port
+		let absoluteUrl;
+		if ( this.details.useCustomDomain && this.details.customDomain ) {
+			absoluteUrl = `http://${ this.details.customDomain }`;
```
mcsf, line 118: Maybe it's worth calling `addDomainToHosts` here, in case the hosts definitions were modified?

src/index.ts
```diff
@@ -305,7 +320,7 @@ async function appBoot() {
 		const clickedButtonIndex = dialog.showMessageBoxSync( {
 			message: __( 'Sync in progress' ),
 			detail: __(
-				'There’s a sync operation in progress. Quitting the app will abort that operation. Are you sure you want to quit?'
+				"There's a sync operation in progress. Quitting the app will abort that operation. Are you sure you want to quit?"
```
ellatrix: Accident?

src/lib/hosts-file.ts
youknowriad: maybe, but this one is already used in Studio before this PR. So seems better addressed separately.

src/lib/hosts-file.ts
youknowriad: Looks like it

src/lib/hosts-file.ts
youknowriad: It is possible but we'd lose the "port" comment.

src/index.ts
youknowriad: yes, I guess the AI changed this randomly :)

src/lib/hosts-file.ts
mcsf: I know in abstract that Windows has come a long way with WSL, but I have no idea if that means that it supports calling `cat` just like that. Now I wonder if the reason why there were two identical branches was that the model had the (correct?) instinct to treat write operations differently on Windows but in the end it didn't interpolate the right command invocation for it. Maybe...? Needs extra testing.

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,177 @@
...
+	} catch ( error ) {
+		console.error( 'Error writing hosts file:', error );
+		throw error;
+	}
+};
+
+/**
+ * Adds a domain to the hosts file pointing to 127.0.0.1
+ */
+export const addDomainToHosts = async ( domain: string, port: number ): Promise< void > => {
```
youknowriad, line 80: We would probably benefit from unit testing this file.

src/lib/hosts-file.ts
andrewserong: Just a quick comment that this stood out to me, too, I have a very long hosts file as well and the grouping would be helpful.

To retain the port comment, could it be something like:

```
# Studio's entries
# BEGIN WordPress.com Studio
127.0.0.1 blablabla.bla # (Port 1234)
127.0.0.1 blablabla2.bla # (Port 1235)
# END WordPress.com Studio
```

andrewserong: Oh, too quick, I see @mcsf already implement that in https://github.com/Automattic/studio/pull/994/commits/e44084905f8e99ac9b89e6eb83f2a32c9d3a7197 — nice!

src/components/site-form.tsx
```diff
@@ -254,6 +278,50 @@ export const SiteForm = ( {
...
 				</label>
+				{ setUseCustomDomain && setCustomDomain && (
+					<div className="flex items-center gap-2 mb-6">
+						<input
+							type="checkbox"
+							id="use-custom-domain"
+							checked={ useCustomDomain }
+							onChange={ ( e ) => {
+								setUseCustomDomain( e.target.checked );
+								setCustomDomain( generateDomainFromSiteName( siteName ) );
```
jsnajdr: This will overwrite the domain name I entered into the `custom-domain` text field when I check or uncheck the checkbox. Probably not wanted, it's a data loss.

src/components/site-form.tsx
```diff
@@ -254,6 +278,50 @@ export const SiteForm = ( {
...
+				{ useCustomDomain && setCustomDomain && (
+					<div className="flex flex-col gap-2 mb-6">
+						<label htmlFor="custom-domain" className="font-semibold">
+							{ __( 'Domain name' ) }
+						</label>
+						<TextControlComponent
+							id="custom-domain"
+							value={ customDomain ?? '' }
+							onChange={ setCustomDomain }
+							placeholder="my-site.wp.local"
```
jsnajdr: Should the placeholder show the value generated by `generateDomainFromSiteName`? We always have a better suggestion for the name than the generic `my-site`.

src/components/site-form.tsx
```diff
@@ -254,6 +278,50 @@ export const SiteForm = ( {
...
+						/>
+						{ customDomainError && (
+							<div className="text-red-500 text-xs">{ customDomainError }</div>
+						) }
```
jsnajdr: We usually use some kind of a `Notice` component for error messages like this.

src/components/site-form.tsx
```diff
@@ -212,6 +226,11 @@ export const SiteForm = ( {
 	setFileForImport,
 	onFileSelected,
 	fileError,
+	useCustomDomain,
+	setUseCustomDomain,
+	customDomain,
+	setCustomDomain,
+	customDomainError,
```
fredrikekelund: This is not the fault of this PR, but we should really move this state inside `SiteForm` and pass it back to the parent in the `onSubmit` handler. Too much prop drilling now with no clear benefit.

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,177 @@
...
+	}
+};
+
+/**
+ * Writes content to the hosts file (requires elevated permissions)
+ */
+export const writeHostsFile = async ( content: string ): Promise< void > => {
+	const hostsPath = getHostsFilePath();
+	try {
+		const tempPath = '/tmp/wp-studio-hosts';
```
fredrikekelund:
```suggestion
		const tempPath = path.join( tmpdir(), 'wp-studio-hosts' );
```
`tmpdir` coming from the `os` module

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,177 @@
...
+		await writeFile( tempPath, content );
+		// @ts-expect-error promisify doesn't seem typed properly.
+		await sudoExec( `cat ${ tempPath } > ${ hostsPath }`, {
```
fredrikekelund: As @mcsf also pointed out https://github.com/Automattic/studio/pull/994#discussion_r1977898256, I don't believe `cat` will work on Windows (it doesn't work in cmd). `type ${ tempPath } > ${ hostsPath }` should work

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Adds a domain to the hosts file pointing to 127.0.0.1
+ */
+export const addDomainToHosts = async ( domain: string, port: number ): Promise< void > => {
+	try {
+		const hostsContent = await readHostsFile();
+
+		// Check if the domain is already in the hosts file
+		const domainRegex = new RegExp(
+			`^127\\.0\\.0\\.1\\s+${ domain.replace( /\./g, '\\.' ) }`,
```
fredrikekelund: The line does not need to begin with the IP address, it just can't be preceded by a comment delineator `#`

src/lib/proxy-server.ts
fredrikekelund, line 1: A couple of comments on this file:
- In my quick testing, `startDomainProxy` seemed to work fine on Windows (i.e. I did not need the elevated privileges). This requires additional testing, but it would greatly simplify the implementation, and we should strive towards not having to execute that separate command.
- We should explicitly handle cases where a domain config exists but the server isn't started. Doesn't need to be super fancy, but we should report an error in the HTTP response that says something like `Server not started`

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,190 @@
...
+			}
+
+			console.log( `Proxying request for ${ host } to port ${ port }` );
+
+			// Forward the request with the original host preserved
+			proxy.web( req, res, {
+				target: `http://localhost:${ port }`,
+				xfwd: true, // Pass along x-forwarded headers
+			} );
+		} );
```
```suggestion
		proxyServer = http.createServer( async ( req, res ) => {
			const host = req.headers.host?.split( ':' )[ 0 ]; // Remove port if present

			// Look up the port directly from user data
			if ( ! host ) {
				console.log( 'No host header found' );
				res.writeHead( 404 );
				res.end( 'No host header found' );
				return;
			}

			const port = await getPortForDomain( host );
			if ( ! port ) {
				console.log( `Domain not found: ${ host }` );
				res.writeHead( 404 );
				res.end( `Domain not found: ${ host }` );
				return;
			}

			// Forward the request with the original host preserved
			proxy.web( req, res, {
				target: `http://localhost:${ port }`,
				xfwd: true, // Pass along x-forwarded headers
			} );
		} );
```
Let's remove the console.log's here

src/storage/user-data.ts
```diff
@@ -119,29 +119,43 @@ export async function saveUserData( data: UserData ): Promise< void > {
...
+				// Add each property explicitly instead.
+				const persistedSiteDetails: PersistedUserData[ 'sites' ][ number ] = {
+					id,
+					name,
+					path,
+					adminPassword,
+					port,
+					phpVersion,
+					customDomain,
+					useCustomDomain,
```
fredrikekelund: Do we really need to store a boolean for `useCustomDomain` in the user config? Is it not enough to store `customDomain`?

src/lib/hosts-file.ts
fredrikekelund: Indeed 👍

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,177 @@
...
+			'i'
+		);
+		if ( domainRegex.test( hostsContent ) ) {
+			return; // Domain already exists
+		}
```
fredrikekelund: What do we do when this happens..? Display some kind of error in the app?

fredrikekelund (commented):
This seems to work relatively well in my light testing, which is great 👍 Still, this is a big feature, and we will need to test it thoroughly before landing. 

A couple of initial high-level thoughts:
- There are several visual issues in the `Add site` modal right now. We should get some designer input from @matt-west or someone else on how to implement this.
- Users should be able to edit the domain after a site is added. This is already a big PR, so we don't necessarily need to implement that here, but it should go into the same release as this feature.

src/ipc-handlers.ts
```diff
@@ -389,6 +394,19 @@ export async function startServer(
...
+	// Handle custom domain if necessary
+	if ( server.details.useCustomDomain && server.details.customDomain ) {
+		try {
+			await addDomainToHosts( server.details.customDomain, server.details.port );
+			console.log(
+				`Domain ${ server.details.customDomain } added to hosts file for port ${ server.details.port }`
+			);
+		} catch ( error ) {
+			console.error( 'Failed to setup custom domain:', error );
+			// Continue even if custom domain setup fails - the site will still work with localhost:PORT
```
fredrikekelund: We no longer show the site port in the UI, though. And wouldn't WordPress redirect to the custom domain..?

src/lib/hosts-file.ts
fredrikekelund: To clarify, I was thinking of cases when the domain is pointing to a different IP. Edge case, perhaps, but it ties into other error handling questions like https://github.com/Automattic/studio/pull/994#discussion_r1979277381

src/components/site-form.tsx
youknowriad: Entirely agree, I noticed it too :)

src/components/site-form.tsx
```diff
@@ -254,6 +278,50 @@ export const SiteForm = ( {
...
+						<input
+							type="checkbox"
+							id="use-custom-domain"
+							checked={ useCustomDomain }
+							onChange={ ( e ) => {
+								setUseCustomDomain( e.target.checked );
+								setCustomDomain( generateDomainFromSiteName( siteName ) );
+							} }
+						/>
+						<label htmlFor="use-custom-domain">{ __( 'Use custom domain' ) }</label>
```
wojtekn: Should we follow the current design from Figma, which assumes we place the 'Local domain' field in the 'Advance settings' block?

src/components/site-form.tsx
```diff
@@ -254,6 +278,50 @@ export const SiteForm = ( {
...
+
+				{ useCustomDomain && setCustomDomain && (
+					<div className="flex flex-col gap-2 mb-6">
+						<label htmlFor="custom-domain" className="font-semibold">
+							{ __( 'Domain name' ) }
+						</label>
+						<TextControlComponent
+							id="custom-domain"
+							value={ customDomain ?? '' }
+							onChange={ setCustomDomain }
```
wojtekn: Could we regenerate the proposed domain when the user changes the name, similar to what happens for the directory name field?

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,177 @@
...
+			return; // Domain doesn't exist
+		}
+
+		// All WordPress Studio entries will be grouped in a block
+		const beginMarker = '# BEGIN WordPress Studio';
+		const endMarker = '# END WordPress Studio';
+
+		// Find the Studio block, if any. For performance, avoid greedy
+		// searches; there should only be a small block anyway.
+		const blockRegex = new RegExp( `(${ beginMarker })[\\s\\S]*?(${ endMarker })` );
```
wojtekn: I'm noting that it conflicts with entries added by the SwitchHosts app - the Studio entries are getting replaced when anything new in SwitchHosts is enabled.

However, it seems to be an issue caused by SwitchHosts as this app doesn't use the start and end blocks to reserve a place for the app but edits anything after the "# --- SWITCHHOSTS_CONTENT_START ---" tag.

To fix that, I copied the SwitchHosts tag to the end of the file, so it's now below the Studio tag.

src/lib/hosts-file.ts
wojtekn: It asks for permission each time I add a new site. Is there a way to save it somehow, at least for the current Studio session?

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,177 @@
...
+
+		if ( blockMatch ) {
+			// Extract existing entries between markers
+			const existingBlock = blockMatch[ 0 ];
+			const entries = existingBlock
+				.split( '\n' )
+				.filter( ( line ) => line !== beginMarker && line !== endMarker && line.trim() !== '' );
+
+			// Add new entry to entries
+			entries.push( domainEntry );
```
wojtekn: It seems to be adding duplicated entries each time I start the site:

```
# BEGIN WordPress Studio
127.0.0.1	moja-witryna-wordpress.wp.local # Port: 8891
127.0.0.1	moja-witryna-wordpress.wp.local # Port: 8891
127.0.0.1	moja-witryna-wordpress.wp.local # Port: 8891
# END WordPress Studio
```

src/components/site-form.tsx
```diff
@@ -35,6 +35,20 @@ interface SiteFormErrorProps {
 	className?: string;
 }
 
+/**
+ * Generates a suitable domain name from site name
+ */
+const generateDomainFromSiteName = ( siteName: string ): string => {
+	// Convert the site name to lowercase and replace spaces with hyphens
+	const domainBase = siteName
```
wojtekn: The current implementation doesn't work for languages that use characters that don't transliterate to a-z ASCII chars. Could we try reusing a logic from the existing `sanitizeFolderName()` function, which currently supports all supported languages?

wojtekn (commented):
The basic happy path works fine on Mac. I left some comments related to small issues and different cases.

src/components/site-form.tsx
youknowriad: I don't think we should be doing that if the user already set a custom domain but might be good to do if the user didn't actually set the custom domain explicitly.

src/lib/hosts-file.ts
mcsf: @wojtekn: thanks for the report. I wonder how common this will be (whether devs are using SwitchHosts or something equivalent, as opposed to manual manipulation). Maybe I misjudged how many devs will appreciate having the rules grouped in a block.

I wouldn't want this addition — which already introduces complexity — to also break some dev flows. So I'm happy to revert. What do you think?

wojtekn: SwitchHosts was recommended in our dosc, so more people are probably using that. This is how I learned about this tool and started using it.

Your idea to add the group with clear start and end tags is excellent. The problem I noted has a workaround (one-time manual edit of the host file) and is rather a bug coming from SwitchHosts using the file without start and end tags.

src/components/site-form.tsx
wojtekn: Correct, this is how site name and directory name relation currently works - directory name is autogenerated when user edits the site name, and stops when user picks the directory manually.

src/components/site-form.tsx
youknowriad: Didn't use a notice but I've used whatever they were using before.

src/storage/user-data.ts
youknowriad: I'm hesitant, I believe that we can get away with just "customDomain" but a boolean gives a sense of clarity. We absolutely need both as state values in the form.

src/lib/hosts-file.ts
youknowriad: this is solved in [4c8c336](https://github.com/Automattic/studio/pull/994/commits/4c8c336011a846ad9e28eb60f1c64a9cbca6e98d)

src/lib/hosts-file.ts
youknowriad: If I'm reading this properly https://www.npmjs.com/package/sudo-prompt#concurrency it's probably not possible.

That said, I don't think it's a problem, I think it was more visible when we had the bug where the hosts file was modified on each "start". But right now, it's only when you start the server for the first time for a given site. So a real user is very unlikely to face the issue on a regular basis.

src/lib/proxy-server.ts
youknowriad: Right now! I'm thinking of removing the elevated privileges entirely. In both our testing it didn't seem necessary. It's easy to add back if we find that we need it. Thoughts?

youknowriad: I've removed proxy elevation in 0513843 We can reconsider when we face issues.

mcsf:
Pushed some unit tests for the `hosts-file` module. [9bf1b56](https://github.com/Automattic/studio/pull/994/commits/9bf1b560f23e5e6b0d7bab8420d30835558ee317) and [b3c4c7b](https://github.com/Automattic/studio/pull/994/commits/b3c4c7b22174ef3b98fdaa26630be78171e4cc4f) show different possible approaches. Read the their commit messages for details. Feel free to push a revert for the latter if the former is preferred.

mcsf:
I pushed a revert, so we keep the first approach. Decisions, not options. :)

youknowriad:
Thanks @mcsf 

I tried adding e2e test myself for the feature, there's a good and a bad news:

 - The good news is that the infrastructure in place allows writing e2e tests pretty easily. Thanks @wojtekn and friends.
 - The bad news is that the current PR can't be e2e tested because it relies on a system prompt (sudo)

---

Overall, I think the PR is a good state, I would love more testing specifically. cc @wojtekn @fredrikekelund

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,127 @@
...
+	try {
+		return await startDomainProxy();
+	} catch ( error ) {
+		console.error( 'Failed to start proxy server:', error );
+		return false;
+	}
+}
+
+/**
+ * Attempts to start the proxy server directly (without privilege elevation)
```
```suggestion
 * Starts the proxy server for sites with custom domains
```

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,127 @@
...
+ */
+async function startDomainProxy(): Promise< boolean > {
+	try {
+		// Create proxy with additional options to preserve host header
+		const proxy = httpProxy.createProxyServer();
+
+		proxy.on( 'error', ( err, req, res ) => {
+			console.error( 'Proxy error:', err );
```
fredrikekelund: This makes for a lot of console noise. I say let's just remove it.

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,175 @@
...
+import { promisify } from 'util';
+import sudo from 'sudo-prompt';
+
+const readFile = promisify( fs.readFile );
+const writeFile = promisify( fs.writeFile );
+const sudoExec = promisify( sudo.exec );
+
+// Host file paths for different operating systems
+const HOST_FILES: Record< string, string > = {
+	win32: 'C:\\Windows\\System32\\drivers\\etc\\hosts',
```
fredrikekelund:
```suggestion
	win32: path.resolve( process.env.SystemRoot ?? 'C:\\Windows', 'System32\\drivers\\etc\\hosts' ),
```
Nit, but this seems appropriate.

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,175 @@
...
+
+		const newContent = updateStudioBlock( hostsContent, ( entries ) => {
+			const pattern = createHostsEntryPattern( domain );
+
+			// No changes if domain already present
+			if ( entries.some( ( entry ) => entry.match( pattern ) ) ) {
+				return entries;
+			}
+
+			return [ ...entries, `127.0.0.1 ${ domain } # Port ${ port } (WordPress Studio)` ];
```
fredrikekelund:
```suggestion
			return [ ...entries, `127.0.0.1 ${ domain } # Port ${ port }` ];
```
The entry is already inside a `BEGIN WordPress Studio` block. Do we really need to mention Studio on each line, too?

src/lib/hosts-file.ts
fredrikekelund, line 1: Nice improvements to this file – it's looking pretty good now 👍

src/hooks/use-add-site.ts
```diff
@@ -23,6 +27,20 @@ export function useAddSite() {
 		[ sites ]
 	);
 
+	const handleCustomDomainChange = useCallback(
+		( value: string ) => {
+			setCustomDomain( value );
+			// Validate custom domain if enabled
+			const domainPattern = /^[a-zA-Z0-9][a-zA-Z0-9-]{0,61}[a-zA-Z0-9](?:\.[a-zA-Z]{2,})+$/;
```
fredrikekelund:
```suggestion
			const domainPattern = /^[a-z0-9][a-z0-9-]{0,61}[a-z0-9](?:\.[a-z]{2,})+$/i;
```
A few comments on the validation:
- We could shorten it by adding the case insensitive flag, no?
- It would be appropriate to validate that the total length is less than 253 characters.
- Every domain label can technically contain hyphens. This feels like nitpicking, but we might as well give it a shot.

fredrikekelund (commented):
This is shaping up nicely overall and continues to work well in my testing on Mac and Windows 👍 A few questions/comments:

- Was there a specific reason for choosing `*.wp.local` as the default custom domain pattern?
- The custom domain name input automatically generates a suggestion, which is a nice touch. However, it should be an actual value and not a placeholder. It's not clear right now that the value will be used if I don't enter something in the text box.
- I don't see the need for storing the `useCustomDomain` boolean in app-data. Let's remove it.
- The Add site modal needs design work.
- When I request a custom domain in the browser while the site server is stopped, I get a blank page with the message `Proxy error:`. Let's improve that and instead display something like `This Studio site is currently stopped.`
- The error message modal when the user doesn't grant admin permissions is OK, but it'd be nice to improve it so we don't display the part about "Please verify your site's local path directory contains the standard WordPress installation files…"
- Let's add some unit tests for `src/lib/hosts-file.ts`. AI is great at that, so it shouldn't be a huge lift.

youknowriad:
> Let's add some unit tests for src/lib/hosts-file.ts. AI is great at that, so it shouldn't be a huge lift.

This is done already, what unit tests you think are missing?

fredrikekelund:
I don't see any unit tests targeting that file specifically. Am I missing something?

youknowriad:
What kind of design work is needed? Is there anything that you don't like or you just want a validation from @matt-west or someone else?

youknowriad:
@fredrikekelund Oh you're right, looks like my rebase kind of reverted @mcsf's tests.

youknowriad:
I restored the tests :)

src/hooks/use-add-site.ts
youknowriad: > Every domain label can technically contain hyphens. This feels like nitpicking, but we might as well give it a shot.

I don't understand this part of the comment.

fredrikekelund:
> design

We should get a review from @matt-west or someone else and incorporate their feedback. 

The layout for the `Use custom domain` checkbox and the accompanying description looks broken right now, so it's also a question of fixing that. I'm fine with waiting for a full review before making changes there, though.

A few other thoughts:
- The `Domain name` label/heading should always be visible, IMO. The checkbox should be rendered below it.
- We could probably roll the two "admin privileges required" notices into one.

Screenshot 2025-03-07 at 11 24 55

src/lib/hosts-file.ts
wojtekn: Correct, now it asks for permissions only when I add a new site, and I choose to use a custom domain. It doesn't ask for permission when I start or stop the site afterward. It seems acceptable, and we could improve it in the future if possible.

src/lib/hosts-file.ts
wojtekn: Works well now, thanks.

src/hooks/use-add-site.ts
fredrikekelund: For clarity, in the domain name `hello.co.uk`, there are three **labels**: `hello`, `co`, and `uk`.

What I meant by my comment was that the regex currently only allows for hyphens in the first label, while it's technically OK to add them in any label. So, `my-domain.local-network` is technically a valid domain.

I know this is slightly nitpicky, but I think the regex needs relatively little adjustment to make this work.

wojtekn:
> The Domain name label/heading should always be visible, IMO. The checkbox should be rendered below it.
> We could probably roll the two "admin privileges required" notices into one.

Thanks for moving new fields to the Advanced settings section @youknowriad .  I agree with @fredrikekelund comment about the design review and the proposed improvements, though. I think it will be great after those small finished touches.

> Was there a specific reason for choosing *.wp.local as the default custom domain pattern?

I guess the reason might be that it's in line with `*.wp.build` links. It looks nice as it's consistent.

> The custom domain name input automatically generates a suggestion, which is a nice touch. However, it should be an actual value and not a placeholder. It's not clear right now that the value will be used if I don't enter something in the text box.

Agreed. As a user, I'm happy that Studio generated the domain for me, but I would like to be able to edit it instead of filling out the domain name from scratch.

I noticed two minor issues.

1. One is an effect of [my proposal](https://github.com/Automattic/studio/pull/994#discussion_r1979352529) to fix how we generate domain for languages that can't be transliterated to a-z characters. Although domain is not broken anymore and matches the site name, I couldn't make such a domain resolve locally. 

   I was able to fix it by updating the domain in Studio JSON and `/etc/hosts` manually to punycode-encoded version - changed `マイ-wordpress-サイト.wp.local` to `xn---wordpress--nz4jla1vqk3l.wp.local`. The site starts now, and resolves correctly. The only quirk is that it shows punycoded version in Studio UI, so we might need to store decoded version and encode it only when we a) start the site b) write to the hosts file.

2. The second one is that I see `ECONNREFUSED` exception logged to the console when I start sites with custom domain. It doesn't seem to affect Studio, though.

youknowriad:
> The error message modal when the user doesn't grant admin permissions is OK, but it'd be nice to improve it so we don't display the part about "Please verify your site's local path directory contains the standard WordPress installation files…"

This one is tricky, I can remove that part of the message but it will be removed for the other cases where startServer fails. That said, in my reading of the startServer code, the other cases have already dedicated messages so it might be ok to remove that part of the error entirely (but your call)

fredrikekelund:
Could we send a different message to `showErrorMessageBox` depending on the exception type/message and keep the old one for all exceptions but the ones stemming from `sudoExec`..? See [source](https://github.com/Automattic/studio/blob/ae765a90b290ad0ceb5b01ec348e3c77183bcc46/src/hooks/use-site-details.tsx#L286-L293).

I view this as just a nice-to-have and we don't want to jump through _too many_ hoops to fix it. Still, not granting admin privileges will probably be the most common error when starting a server after we land this PR, so having some specific error handling logic for it would be warranted.

youknowriad:
> Could we send a different message to showErrorMessageBox depending on the exception type/message and keep the old one for all exceptions but the ones stemming from sudoExec..? See [source](https://github.com/Automattic/studio/blob/ae765a90b290ad0ceb5b01ec348e3c77183bcc46/src/hooks/use-site-details.tsx#L286-L293).

I don't see any other way other than checking the string or introducing a format/classes for errors. The latter is better but I think it warrants its own PR instead.

youknowriad:
@wojtekn What is the domain you're using to test the transliterated domains? In my testing these are rejected during validation at the moment (which could be a good way to start)

fredrikekelund:
> I don't see any other way other than checking the string or introducing a format/classes for errors. The latter is better but I think it warrants its own PR instead.

Gotcha. I agree with all of that. Let's look into it separately 👍

wojtekn:
> What is the domain you're using to test the transliterated domains? In my testing these are rejected during validation at the moment (which could be a good way to start)

It was Japanese with the "マイ WordPress サイト" site name auto-generated. I pulled changes, but I don't see an auto-generated domain for any language.

src/hooks/use-add-site.ts
youknowriad: I think the regex allows that now.

youknowriad:
> The second one is that I see ECONNREFUSED exception logged to the console when I start sites with custom domain. It doesn't seem to affect Studio, though.

I'm not seeing that at the moment but my guess is that it happens maybe because the apps is trying to generate a "preview" while the site is not ready yet. (just a guess).

youknowriad:
Thanks both of you for your valuable reviews. I think I've addressed all the points. Let me know if we're missing something.

wojtekn:
Works great now! Thanks for adding test for the domain generation.

> I'm not seeing that at the moment but my guess is that it happens maybe because the apps is trying to generate a "preview" while the site is not ready yet. (just a guess).

I can't reproduce it now, too.

src/hooks/use-add-site.ts
```diff
@@ -31,6 +35,25 @@ export function useAddSite() {
 		[ sites ]
 	);
 
+	const handleCustomDomainChange = useCallback(
+		( value: string | null ) => {
+			setCustomDomain( value );
+			// Validate custom domain if enabled
+			const domainPattern =
+				/^[\p{L}\p{N}][\p{L}\p{N}-]{0,61}[\p{L}\p{N}](?:\.[\p{L}\p{N}-]{2,})+$/u;
```
fredrikekelund:
```suggestion
				/^[a-z0-9][a-z0-9-]{0,61}[a-z0-9](\.[a-z0-9][a-z0-9-]{0,61}[a-z0-9])+$/i;
```
I don't think it's correct to allow any unicode characters in the domain name.

I believe my suggestion catches the relevant syntax criteria: two or more labels that can only contain ASCII letters (lower- or uppercase), digits, and hyphens (but not start or end with a hyphen).

src/components/content-tab-settings.tsx
```diff
@@ -53,13 +56,15 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
...
 					</SettingsRow>
 					<SettingsRow label={ __( 'Local URL' ) }>
-						<CopyTextButton
-							text={ `http://localhost:${ selectedSite.port }` }
-							label={ `localhost:${ selectedSite.port }, ${ __( 'Copy site url to clipboard' ) }` }
-							copyConfirmation={ __( 'Copied!' ) }
-						>
-							{ `localhost:${ selectedSite.port }` }
-						</CopyTextButton>
+						<div className="flex">
```
fredrikekelund: Do we really need this div wrapper?

src/hooks/use-add-site.ts
wojtekn: We need to allow for more than ASCII letters, digits, and hyphens at this step, to support other supported languages. The domain is now encoded using `domainToASCII` when we start proxy or write to the hosts file.

fredrikekelund: Or maybe I'm misunderstanding how we intend to support internationalized domain names..?

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,138 @@
...
+			if ( ! site ) {
+				console.log( `Domain not found: ${ host }` );
+				res.writeHead( 404 );
+				res.end( `Domain not found: ${ host }` );
+				return;
+			}
+
+			if ( ! site.running ) {
+				res.writeHead( 404 );
+				res.end( `The Studio site is currently stopped: ${ host }` );
```
fredrikekelund: Would it perhaps be more useful to output the site name instead of the site host here? That's how users can find the site in the Studio sidebar.

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,138 @@
...
+					console.error( `Error starting proxy server on port 80:`, err );
+					reject( err );
+				} );
+		} );
+
+		return true;
+	} catch ( error ) {
+		console.error( `Failed to start proxy server directly:`, error );
+		return false;
+	}
```
fredrikekelund: Let's report to Sentry here.

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,138 @@
...
+			// Forward the request with the original host preserved
+			proxy.web( req, res, {
+				target: `http://localhost:${ site.port }`,
+				xfwd: true, // Pass along x-forwarded headers
+			} );
+		} );
+
+		await new Promise< void >( ( resolve, reject ) => {
+			proxyServer!
+				.listen( 80, () => {
```
fredrikekelund, line 85:
```suggestion
			proxyServer?
				.listen( 80, () => {
```
Nit, but `?.` is native JS whereas `!.` is TS, but they do the same thing.

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Starts the proxy server for sites with custom domains
+ */
+async function startDomainProxy(): Promise< boolean > {
+	try {
+		// Create proxy with additional options to preserve host header
+		const proxy = httpProxy.createProxyServer();
+
+		proxy.on( 'error', ( err, req, res ) => {
+			if ( res && 'writeHead' in res ) {
```
```suggestion
			if ( res instanceof http.ServerResponse ) {
```
Nit, but this seems to clarify the intention better.

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,178 @@
...
+
+			return [ ...entries, `127.0.0.1 ${ encodedDomain } # Port ${ port }` ];
+		} );
+
+		if ( newContent !== hostsContent ) {
+			await writeHostsFile( newContent );
+		}
+	} catch ( error ) {
+		console.error( `Error adding domain ${ domain } to hosts file:`, error );
+		throw error;
```
fredrikekelund, line 102: Let's report to Sentry here, too. Most errors will be related to permissions, but I think we can mute those in Sentry.

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,178 @@
...
+			entries.filter( ( entry ) => ! entry.match( pattern ) )
+		);
+
+		// Only write if content changed
+		if ( newContent !== hostsContent ) {
+			await writeHostsFile( newContent );
+		}
+	} catch ( error ) {
+		console.error( `Error removing domain ${ domain } from hosts file:`, error );
+		throw error;
```
fredrikekelund, line 126: And let's report to Sentry here, too.

fredrikekelund (commented):
Looking good, @youknowriad 👍 Thanks for addressing all those comments. Let's make sure we get a design review for this before landing, but the core functionality looks mostly solid to me now (I just need to understand the logic around internationalized domain names)

src/hooks/use-add-site.ts
fredrikekelund: Thanks for the clarification, @wojtekn 👍

fredrikekelund:
```suggestion
				/^[\p{L}\p{N}][\p{L}\p{N}-]{0,61}[\p{L}\p{N}](?:\.[\p{L}\p{N}][\p{L}\p{N}-]{0,61}[\p{L}\p{N}])+$/u;
```
Then let's do something like this to account for my previous comment about hyphens in all labels.

youknowriad: @fredrikekelund This is already accounted for in my tests, can you give me an example of domain that don't pass validation?

src/lib/proxy-server.ts
youknowriad: This is not exactly the same, it would create a TS error.

src/lib/hosts-file.ts
mcsf: I think we can remove it, yes, but I'll leave that up to Riad.

Also,

> Let's add some unit tests for src/lib/hosts-file.ts. AI is great at that, so it shouldn't be a huge lift.

You'd think so, but two very different systems both got it spectacularly wrong and cost me tons of time. Had to reset and do it myself. In hindsight I think they got confused by the use of `promisify` in `hosts-file`. :)

matt-west:
Thanks @youknowriad. A lot of users are going to love this feature!

I tested this locally and it worked well, although starting the site after entering my system password was quite slow.

What’s stopping us from creating a custom domain for every new site that’s created in Studio? Local takes that approach. I feel like the custom domain field should be directly under `Site name`. That would also allow us to eliminate the checkbox.

---

A few things I noticed testing this.

1. The `Advanced settings` dropdown and `Local path` field are misaligned.

Screenshot 2025-03-07 at 17 05 55

2. Let’s display the password note on the `Domain name` field instead and update the copy to:

> Your system password will be required to set up the domain.

3. There’s no way to edit the domain after I’ve created the site. Perhaps that’s coming in a follow-up PR?

4. Was the custom domain option added to the onboarding screen too? (I don’t have an easy way to test that.)

youknowriad:
> What’s stopping us from creating a custom domain for every new site that’s created in Studio? Local takes that approach. I feel like the custom domain field should be directly under Site name. That would also allow us to eliminate the checkbox.

Local has a "router mode" that is applied globally that you can disable. Also switching it back and forth means all your existing sites would stop working for now. Personally I prefer the control site per site but we can always have it as a global mode later. The reason we don't do it automatically though is that it depends on admin rights and not all the users have these.


> The Advanced settings dropdown and Local path field are misaligned.

This was a result of a rebase, it's now fixed

> Let’s display the password note on the Domain name field instead and update the copy to:

This is addressed too

> There’s no way to edit the domain after I’ve created the site. Perhaps that’s coming in a follow-up PR?

A potential follow-up indeed, but it's not that simple because it requires updating the WP database for existing sites.

> Was the custom domain option added to the onboarding screen too? (I don’t have an easy way to test that.)

Yes

src/hooks/use-add-site.ts
fredrikekelund: I see now that hyphens are already allowed in all labels, but we still need to adjust the regex to account for the fact that no label can start or end with hyphens. For example, `hey.-hey` passes the current regex, which is incorrect.

youknowriad: Should be addressed in d6a42be69d3bb66a6b206733a272191fc42042a0

fredrikekelund:
> The Advanced settings dropdown and Local path field are misaligned.

That looks like a regression unrelated to this PR. I've created https://github.com/Automattic/dotcom-forge/issues/10582 to track it. It still looks better with @youknowriad's latest changes – at least the local path and domain fields are consistently aligned now.

src/lib/hosts-file.ts
```diff
@@ -0,0 +1,181 @@
...
+	} catch ( error ) {
+		console.error( 'Error writing hosts file:', error );
+		throw error;
+	}
+};
+
+/**
+ * Create a regular expression matching the hosts entry for a given domain:
+ *
+ * 	127.0.0.1 foo.wp.cloud # Port 8000 (WordPress Studio)
```
fredrikekelund:
```suggestion
 * 	127.0.0.1 foo.wp.cloud # Port 8000
```
Nit, but let's adjust this now that we no longer add `(WordPress Studio)` to each entry

src/lib/proxy-server.ts
fredrikekelund: My bad! I see the difference now 👍

youknowriad:
Can I get an approval here? What do you think is left?

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,140 @@
...
+					console.error( `Error starting proxy server on port 80:`, err );
+					reject( err );
+				} );
+		} );
+
+		return true;
+	} catch ( error ) {
+		Sentry.captureException( error );
+		console.error( `Failed to start proxy server directly:`, error );
+		return false;
```
fredrikekelund, line 113:
```suggestion
		if ( error instanceof Error && error.message.includes( 'EADDRINUSE' ) ) {
			dialog.showMessageBox( { // imported from the `electron` module
				type: 'error',
				message: __( 'Failed to start custom domain proxy server' ),
				detail: __(
					'Another server is already running on port 80. For custom domains to work, please stop that server and then restart Studio.'
				),
				buttons: [ __( 'OK' ) ],
			} );

			return false;
		}

		Sentry.captureException( error );
		console.error( `Failed to start proxy server directly:`, error );
		return false;
```
While testing, I ran into a case where port 80 was already in use, which led to a pretty weird user experience (the site thumbnail in Studio was a 404 page from my other app). We should look for `EADDRINUSE` errors here and display an error modal.

My suggestion is far from perfect, but it's a start. Feel free to add or improve.

src/lib/proxy-server.ts
```diff
@@ -0,0 +1,140 @@
...
+export async function startProxyServer(): Promise< boolean > {
+	if ( isProxyRunning ) return true;
+
+	try {
+		return await startDomainProxy();
+	} catch ( error ) {
+		console.error( 'Failed to start proxy server:', error );
+		return false;
+	}
+}
```
fredrikekelund: Do we really need this wrapper function? `startDomainProxy` seems to already do everything we need except return early if `isProxyRunning` is true.

src/index.ts
```diff
@@ -197,6 +198,20 @@ async function appBoot() {
...
+			if ( proxyStarted ) {
+				console.log( 'Custom domain proxy server started successfully' );
+			} else {
+				console.warn(
+					'Failed to start custom domain proxy server - custom domains will require port numbers'
+				);
+			}
+		} catch ( error ) {
+			console.error( 'Error starting proxy server:', error );
+		}
```
fredrikekelund: To accommodate the `EADDRINUSE` error modal, I recommend moving this logic below the `createMainWindow` call. Otherwise, the error modal will display before we create the main window, which looks odd.

fredrikekelund (commented):
I ran another round of tests on Mac and Windows – it's looking good and works as expected. I did however encounter one more error handling scenario that I think we should handle before landing this PR (`EADDRINUSE` errors on port 80).

I believe we should be good to land this PR after addressing that 👍

src/lib/proxy-server.ts
youknowriad: Maybe we need a way to display error notifications in the UI in a more unified way. I guess that would do it for now.

fredrikekelund: We typically use `showErrorMessageBox` from the IPC API for that, displaying an error modal like this one (modals for errors, notifications for successes). I agree it doesn't look very pretty in this case, but acknowledging the error is still better UX, IMO. I created https://github.com/Automattic/dotcom-forge/issues/10583 for iterating on the error message.

fredrikekelund (approved):
Nice work, @youknowriad! We should be good to land this now 👍

Let's add some additional rows to our smoke testing sheet for this feature, @wojtekn. That gives us another chance to catch potential errors before the next release.

wojtekn:
> Let's add some additional rows to our smoke testing sheet for this feature, @wojtekn. That gives us another chance to catch potential errors before the next release.

I've added a test case for local domain.

PR #1001, opened 2025-03-04
fix dev version on release

bcotrim:
## Related issues

- I noticed the build was failing on the release, for example: https://buildkite.com/automattic/studio/builds/4620#019560dd-3ab4-4bdd-9c7b-935cc6c4f21e

## Proposed Changes

- Fix condition in scripts/prepare-dev-build-version.mjs to properly handle when commit count is 0

## Testing Instructions

- Hard code the return value for `getCommitCount` to 0 (or any other value)
- Run the prepare dev version command 
```bash
node ./scripts/prepare-dev-build-version.mjs
```
- The version in package.json should be like `-dev0`

-

## Pre-merge Checklist

### DIFF
--- a/scripts/prepare-dev-build-version.mjs
+++ b/scripts/prepare-dev-build-version.mjs
@@ -11,7 +11,7 @@ const __dirname = path.dirname( fileURLToPath( import.meta.url ) );
 const latestTag = getLatestTag();
 const commitCount = getCommitCount( latestTag );
 
-if ( ! commitCount ) {
+if ( ! commitCount && commitCount !== 0 ) {
 	// Are you trying to dev on the build scripts outside of CI?
 	// You will need to define the GITHUB_SHA or BUILDKITE_COMMIT environment
 	// variable before running build scripts. e.g.


fredrikekelund (approved):
Nice 👍

fredrikekelund:
@mokagio, the `Distribute Dev Builds` CI job failed because of an error in our code, but I raised a question about whether it even makes sense to run that job for release builds. It seems we [intended](https://github.com/Automattic/studio/blob/20125541b75f2e98e6b5a78e78515d94e9ff0909/.buildkite/pipeline.yml#L181) not to do that, but release commits are tagged **after** they're merged, so that config isn't going to work. Any good ideas about mitigating that, or should we just keep running dev build jobs for release commits, too?

mokagio:
Hey @fredrikekelund, thanks for the ping

> I raised a question about whether it even makes sense to run that job [distribute dev builds] for release builds

With release builds, I think we mean builds originating from a tag?

In that case, I can see we only run the distribute release job. [Example](https://buildkite.com/automattic/studio/builds/4621/canvas):

image

So, the way I'm reading the rest of the comment:

>  release commits are tagged after they're merged, so that config [do only run on `trunk` build without a tag] isn't going to work.

is that we are looking for a way to no distribute dev builds on commits that _will become_ release builds once tagged. Is that right?

For other readers' reference, here's the config:

https://github.com/Automattic/studio/blob/20125541b75f2e98e6b5a78e78515d94e9ff0909/.buildkite/pipeline.yml#L181

I think the problem is that we don't know in advance which commits will be promoted to release builds, right?

We _could_ address this with [git-flow](https://danielkummer.github.io/git-flow-cheatsheet/) like setup with `development` and `production` branches, but I don't think that's advisable as it unnecessarily slows down development with the added process around merging `development` into `production` etc.

An alternative would be to require a manual input to ship dev builds. That would prevent shipping a dev build when there's a release one coming. But, this too would slow down development because now folks would have to go and click the button.

So in the end my two cents are that it's okay to ship a redundant, so to speak, dev build. It's relatively cheap and keeps our development process smooth.

Of course, I'm coming at it as an infra engineer, so I might be missing something as far the dev team likes to operate and ship. If so, please explain, thanks.

If we decide to go down the ship-redundant-build-anyway path, I suggest updating the pipeline condition to something like

```diff
-    if: build.branch == 'trunk' && build.tag !~ /^v[0-9]+/
+    # We ship a dev build for every build on trunk, including those that we might eventually tag as release build
+    if: build.branch == 'trunk'
```

fredrikekelund:
Thanks for an exhaustive and clarifying answer, @mokagio 👍 Given what you shared, I agree it's best to keep the config as it is now.

wojtekn (approved):

PR #1002, opened 2025-03-04
Clean up console logs from sync pulling

nightnei:
## Proposed Changes
Remove dev tool (console.log) from production.
I was working around Pulling process, so cleaned it up. There are no pros in keeping them on production, only cons.

## Testing Instructions
Visual review should suffice

### DIFF
--- a/src/hooks/sync-sites/use-sync-pull.ts
+++ b/src/hooks/sync-sites/use-sync-pull.ts
@@ -95,8 +95,6 @@ export function useSyncPull( {
 				return;
 			}
 
-			console.groupCollapsed( 'Sync Pull' );
-
 			const remoteSiteId = connectedSite.id;
 			updatePullState( selectedSite.id, remoteSiteId, {
 				backupId: null,
@@ -108,7 +106,7 @@ export function useSyncPull( {
 			} );
 
 			try {
-				console.log( 'Initializing backup on remote' );
+				// Initializing backup on remote
 				const response = await client.req.post< { success: boolean; backup_id: string } >( {
 					path: `/sites/${ remoteSiteId }/studio-app/sync/backup`,
 					apiNamespace: 'wpcom/v2',
@@ -124,7 +122,6 @@ export function useSyncPull( {
 				}
 			} catch ( error ) {
 				console.error( 'Pull request failed:', error );
-				console.groupEnd();
 
 				Sentry.captureException( error );
 				updatePullState( selectedSite.id, remoteSiteId, {
@@ -144,7 +141,7 @@ export function useSyncPull( {
 		try {
 			return await getIpcApi().checkSyncBackupSize( downloadUrl );
 		} catch ( error ) {
-			console.log( 'Failed to check backup file size', error );
+			console.error( 'Failed to check backup file size', error );
 			Sentry.captureException( error );
 			throw new Error( 'Failed to check backup file size' );
 		}
@@ -156,10 +153,8 @@ export function useSyncPull( {
 
 			try {
 				const fileSize = await checkBackupFileSize( downloadUrl );
-				console.log( 'Backup file size:', { fileSize, limit: SYNC_PUSH_SIZE_LIMIT_BYTES } );
 
 				if ( fileSize > SYNC_PUSH_SIZE_LIMIT_BYTES ) {
-					console.log( 'File size exceeds limit, prompting user' );
 					const { response: userChoice } = await getIpcApi().showMessageBox( {
 						type: 'warning',
 						message: __( "Large site's backup" ),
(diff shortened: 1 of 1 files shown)


src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -174,28 +169,25 @@ export function useSyncPull( {
 						cancelId: 1,
 					} );
 
-					if ( userChoice === 1 ) {
-						console.log( 'User cancelled pull operation' );
+					const isCanceledByUser = userChoice === 1;
+					if ( isCanceledByUser ) {
```
nightnei: *Wojtek*:
Do we need this variable if it's used only once?

*Vova*:
I just wanted to avoid "magic" number `1`, since at a glance I thought that `1` means kinda `true`, as a confirmation to continue pulling, but actually it's canceling.

So to keep readability, we need either to add comment about `1` or create readable and simple variable. And the good variable name is much better than any comment.

src/hooks/sync-sites/use-sync-pull.ts
```diff
@@ -291,15 +277,14 @@ export function useSyncPull( {
 					pullStatesProgressInfo[ frontendStatus ] || pullStatesProgressInfo.failed;
 				const downloadUrl = hasBackupCompleted ? response.download_url : null;
 
-				if ( hasBackupCompleted && downloadUrl ) {
+				if ( downloadUrl ) {
```
nightnei, line 281: *Wojtek*:
What's the context of this removal?

*Vova*:
The above line has `const downloadUrl = hasBackupCompleted ? response.download_url : null;` - `downloadUrl ` is available only if `hasBackupCompleted ` is `true`. So it's duplication of the condition.

src/hooks/sync-sites/use-sync-pull.ts
wojtekn: What if we renamed `userChoice` to `userCanceled`? Wouldn't it be clear enough without introducing a variable for this simple condition?

src/hooks/sync-sites/use-sync-pull.ts
wojtekn: Thanks for explanation, looks good.

src/hooks/sync-sites/use-sync-pull.ts
nightnei: Oh, my bad, I just realized that I took a look at the code not thoroughly enough, sorry.

We can go with either `userCanceled` or [this way](https://github.com/Automattic/studio/pull/1002/commits/d54e9c4664b2dc4dedd98f6636d2e5c2258ffb7b). I am ok with both options. What do you prefer?

wojtekn: The way with named constant looks great!

ashfame:
> There are no pros in keeping them on production, only cons.

I added them as they were quite helpful during development and while there may not be a reason for them to be in production, because we really don't have any means of controlling logging verbosity or on the basis of environment yet, I chose to keep them. Wouldn't adding support for just logging in development be better than removing them altogether? In fact, I would argue we should have tons of such logging for all operations. Very helpful to just look at logs and see what happened when a UI interaction misbehaves.

nightnei:
@ashfame In general I see your point and it makes sense, but there is a few critical points, why I believe that console.logs should be used only locally and should not be even committed to repo

> I added them as they were quite helpful during development and while there may not be a reason for them to be in production, because we really don't have any means of controlling logging verbosity or on the basis of environment yet, I chose to keep them.

Just saying:
1. In renderer process, excessive console.log calls affect **frame rate** and UI rendering speed and even can block it for a while.
2. console.log can increase memory usage, especially in long-running desktop app like Electron - too many logs can cause memory leaks

So, it's the main points for this PR, there are no pros to keep it in prod, only cons.

> Wouldn't adding support for just logging in development be better than removing them altogether?
1. It's an option, but, In this case we should at least ensure that tree-shaking works correctly and we don't add useless lines of code to production build.
2. But more important point is - while working on a task around Pulling process - these logs just distracted me. In rare cases, I think, it's ok to use console.error, since it is shown rarely when some exception happens, but such console.logs are persistent and flood console. I am on board with keeping console.error, but in case of console.log - every issue is unique and if somebody needs to debug specific flow, then this developer can add only necessary logs for them during debugging and remove before committing as not necessary anymore, since the issue is fixed.
3. Also, if we keep console.log inside some part of codebase, it will show other devs that it's ok and they could start adding and keep them everywhere. So my point is - let's keep console in codebase only in cases when we very want to highlight some exception for our teammates.

Example - there is totally no sense in [console.log( 'Starting import process' );](https://github.com/Automattic/studio/pull/1002/files#diff-600258d9240b0fc908f6652559009061c334bd0bf2c6c326437507c7ec62919aL198), but there are benefits to print [console.error( 'Failed to check backup file size', error );](https://github.com/Automattic/studio/pull/1002/files#diff-600258d9240b0fc908f6652559009061c334bd0bf2c6c326437507c7ec62919aR144).

> Very helpful to just look at logs and see what happened when a UI interaction misbehaves.

It should not misbehave 😄 But if it is, then the developer will see errors in console and will debug the issue with adding  temporary and necessary for them console.logs.

To sum-up, my proposition is the next:
1. By default don't commit console.log to github repo
2. In some edge cases it's ok to add console.log, but necessary to have some really strong reason for it and what is the most important - before adding them, necessary to ensure that they are wrapped with something like `process.env.NODE_ENV === "development"` and ensure that tree-shaking works well for it

WDYT?

wojtekn (approved):

PR #1003, opened 2025-03-04
Fix WP.org API zod schema

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10564

## Proposed Changes

https://api.wordpress.org/core/version-check/1.7/?channel=beta&version now returns the string `development` in an `offers[].response` key. This PR changes the zod schema to be more permissive to avoid those validation errors. It also makes the parsing less sensitive by discarding individual array items that don't conform to the schema instead of discarding the entire response.

## Testing Instructions

Code review should suffice

## Pre-merge Checklist

### DIFF
--- a/src/stores/wordpress-versions-slice.ts
+++ b/src/stores/wordpress-versions-slice.ts
@@ -6,7 +6,7 @@ const MINIMUM_WORDPRESS_VERSION = '5.9.9';
 
 const wordPressOfferSchema = z.object( {
 	version: z.string(),
-	response: z.enum( [ 'autoupdate', 'upgrade' ] ),
+	response: z.enum( [ 'autoupdate', 'development', 'upgrade' ] ),
 } );
 
 const wordPressApiResponseSchema = z.object( {
--- a/src/stores/wordpress-versions-slice.ts
+++ b/src/stores/wordpress-versions-slice.ts
@@ -6,7 +6,7 @@ const MINIMUM_WORDPRESS_VERSION = '5.9.9';
 
 const wordPressOfferSchema = z.object( {
 	version: z.string(),
-	response: z.enum( [ 'autoupdate', 'development', 'upgrade' ] ),
+	response: z.string(),
 } );
 
 const wordPressApiResponseSchema = z.object( {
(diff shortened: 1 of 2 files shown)


bcotrim (approved):
LGTM 👍
My only suggestion would be to add some unit tests for this logic.

wojtekn (approved):
Looks good 👍

fredrikekelund:
I spent some extra time thinking about schema validation approaches, because I don't want using zod to become too much of a burden, and this PR definitely jumps through some extra hoops.

A good way to think about it, IMO, is that we can apply graceful schema parsing like this when we don't have full control over the API endpoint. In this case, we don't, so I think it's a reasonable measure. It would not make sense for WordPress.com API endpoints with a normal degree of complexity.

src/stores/wordpress-versions-slice.ts
```diff
@@ -6,11 +6,13 @@ const MINIMUM_WORDPRESS_VERSION = '5.9.9';
 
 const wordPressOfferSchema = z.object( {
 	version: z.string(),
-	response: z.enum( [ 'autoupdate', 'upgrade' ] ),
+	response: z.string(),
```
sejas, line 9: ❤️

src/stores/wordpress-versions-slice.ts
```diff
@@ -6,11 +6,13 @@ const MINIMUM_WORDPRESS_VERSION = '5.9.9';
...
 } );
 
+type WordPressOffer = z.infer< typeof wordPressOfferSchema >;
+
 const wordPressApiResponseSchema = z.object( {
-	offers: z.array( wordPressOfferSchema ),
+	offers: z.array( z.any() ),
```
sejas, line 15: Ha! That's a neat trick. Allowing flexibility at the top level while discarding elements that don't adhere to the `wordPressOfferSchema` without rejecting the whole array 📝 👍

sejas (commented):
I tried to find the source code of the API `https://api.wordpress.org/core/version-check/1.7/` so we could implement more secure types, but couldn't find it. Thanks for adding more flexibility to the types.

PR #998, opened 2025-03-03
Use CI toolkit plugin v5 and nvm plugin v0.4.0

mokagio:
## Related issues

- Internal ref pdnsEh-24x-p2

## Proposed Changes

- Use nvm CI plugin v0.4.0 which has clearer logs
- Use CI toolkit plugin v5.0.0 which comes with more automation for Windows and allows us to remove custom scripts

## Testing Instructions

Refer to green Windows CI builds

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? – N.A.

PR #1007, opened 2025-03-05
Use `env` node for all `IS_DEV_BUILD` CI assignments

mokagio:
## Related issues

- Builds on to of #998

## Proposed Changes

Always use `env: IS_DEV_BUILD: true` in CI to set the `IS_DEV_BUILD` env var.

In https://github.com/Automattic/studio/pull/998#discussion_r1978739613 I got confirmation that it's okay to use this approach for the env var setting.

Given it's already in use, because it would not work otherwise, in the Windows steps, this commit updates all the other steps so we have a consistent setup.

## Testing Instructions

-

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? – N.A.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -54,7 +54,6 @@ steps:
     command: |
       # call with bash explicitly because we run this on Windows, too
       bash .buildkite/commands/install-node-dependencies.sh
-      export IS_DEV_BUILD=true
       echo '--- :package: Package app for testing'
       npm run package
       echo '--- :playwright: Run End To End Tests'
@@ -67,6 +66,7 @@ steps:
     env:
       # See https://playwright.dev/docs/ci#debugging-browser-launches
       DEBUG: "pw:browser"
+      IS_DEV_BUILD: true
     matrix:
       - mac
       #- windows
@@ -86,8 +86,6 @@ steps:
           .buildkite/commands/install-node-dependencies.sh
 
           node ./scripts/prepare-dev-build-version.mjs
-
-          export IS_DEV_BUILD=true
 
           echo "--- :node: Building Binary"
           npm run make:macos-{{matrix}}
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -67,6 +66,7 @@ steps:
     env:
       # See https://playwright.dev/docs/ci#debugging-browser-launches
       DEBUG: "pw:browser"
+      IS_DEV_BUILD: true
```
mokagio, line 69: Mmm... in the E2E on the `mac` queue, `IS_DEV_BUILD` is `true` in the environment tab, but the build process mustn't be reading it as such, because the Sentry log check reads "production"

_See bottom of the list here:_
image

image

.buildkite/pipeline.yml
```diff
@@ -113,6 +111,8 @@ steps:
 
           echo "--- 📃 Notarizing Binary"
           bundle exec fastlane notarize_binary
+        env:
+          IS_DEV_BUILD: true
```
mokagio, line 115: Okay in universal

image

image

mokagio: Okay in x64

image

image

mokagio: Okay in arm64
image

image

.buildkite/pipeline.yml
mokagio: Given the packaging steps below correctly send the env var to the Mac VM, my current guess is that the issue here has to do with how the E2E tests are scripted.

Will look into it...

mokagio: Odd... If I run with `IS_DEV_BUILD=true` on my local, I get sentry development as expected.

image

Also had to `export DEBUG=pw:browser`" to see the sentry logs.

cc @wojtekn any ideas?

wojtekn: The variable is read in the Electron app via `process.env.IS_DEV_BUILD` to decide whether Sentry should be initialized or not. It seems that when the environment variable is set in `env:` section, it's not available in the Studio process launched by E2E.

If we want to remove it from the command section, what if we set it explicitly in the `package.json` where we define `e2e` script? We do the same for other environment variables with other scripts there.

wojtekn:
@mokagio, will you be picking this one up soon, or should we close it?

mokagio:
Closing this as _very_ stale.

Thanks for the ping @wojtekn

PR #1018, opened 2025-03-06
Remove hello wp-now default index php that should be virtualized

sejas:
## Related issues

- Related https://github.com/Automattic/dotcom-forge/issues/10569

## Proposed Changes

- Remove unnecessary default wp-now index.php that was virtualized in the past
- It doesn't fix the issue of `rotatePHPRuntime` , but at least doesn't modify user's file.

## Testing Instructions

- Test start and stop sites
- Observe Studio sites work as expected

## Pre-merge Checklist

### DIFF
--- a/vendor/wp-now/src/wp-now.ts
+++ b/vendor/wp-now/src/wp-now.ts
@@ -135,7 +135,7 @@ export default async function startWPNow(
 			await prepareWordPress( php, options );
 			return runtimeId;
 		},
-		maxRequests: 400,
+		maxRequests: 10,
 	} );
 
 	return {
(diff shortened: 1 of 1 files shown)


vendor/wp-now/src/wp-now.ts
```diff
@@ -135,7 +135,7 @@ export default async function startWPNow(
 			await prepareWordPress( php, options );
 			return runtimeId;
 		},
-		maxRequests: 400,
+		maxRequests: 10,
```
wojtekn, line 138: Is this change here only for test purposes?

wojtekn (commented):
> It doesn't fix the issue of rotatePHPRuntime , but at least doesn't modify user's file.

What issue we are left with?

vendor/wp-now/src/wp-now.ts
sejas: yeah, this shouldn't go here 😅 . Good catch.

sejas:
> What issue we are left with?

The runtime is crashing, and the website is unresponsive.

sejas:
Closed in favor of https://github.com/Automattic/studio/pull/1019

PR #1026, opened 2025-03-07
Fix background site form file selectors 

sejas:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10579

## Proposed Changes

- Let's use a default style for all TextControl disabled components
- Avoid keyboard focus on a "fake" text input with tabIndex instead of disabled

We could consider removing the text controls in favor of a regular Div with custom styles, but then it might diverge in the future after updating `@wordpress/components`.

## Testing Instructions

- Run `npm start``
- Press `cmd+n`
- Observe the input files have a white background
- Use the tab to change the focus between the elements
- Observe the "fake" text input is not focused

https://github.com/user-attachments/assets/2d999c71-2b75-4dcd-9d20-fe1eff9f0679

Screenshot 2025-03-07 at 12 25 21
Screenshot 2025-03-07 at 12 24 41

## Pre-merge Checklist

### DIFF
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -138,7 +138,7 @@ function FormPathInputComponent( {
 			>
 				<TextControlComponent
 					aria-hidden="true"
-					disabled={ true }
+					tabIndex={ -1 }
 					className="[&_.components-text-control\_\_input]:bg-transparent [&_.components-text-control\_\_input]:border-none [&_input]:pointer-events-none w-full"
 					value={ value }
 					// eslint-disable-next-line @typescript-eslint/no-empty-function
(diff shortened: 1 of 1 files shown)


katinthehatsite (commented):
Looks good to me. The grey border is fixed 👍 

I noticed two things:

* We still need to fix the spacing for the error
* I am seeing that we have the following e.g. two betas:

Screenshot 2025-03-07 at 2 13 04 PM

I guess we can check the betas as a part of different issue.

bcotrim (approved):
Tested and worked as described
LGTM 👍

sejas:
> We still need to fix the spacing for the error

See https://github.com/Automattic/studio/pull/1028

> I am seeing that we have the following e.g. two betas:

Yeah, we can create an issue to improve the conciliation of WP versions, since that behaviour is not introduced in this PR.
Users can upgrade or downgrade their versions in many ways outside of Studio, so what I do is to "inject" the version in the dropdown if the current version is not in the list.

sejas:
> I am seeing that we have the following e.g. two betas:

@katinthehatsite , I think it's ok to have different betas as probably they are using different values in `wp-includes/version.php`. 
An example is if/when we support Nightly version, that version name will be the same, but the "git hash" suffix will be different every day.

PR #1034, opened 2025-03-08
STU-19: Add SSL Support

### LINKED ISSUE #191: Feature Request: SSL
### What

We need SSL capabilities in Studio on Windows because we deal with sites that require SSL for processing credit cards and other sensitive information. Without it, this work cannot be done in Studio.

### Why

Without SSL some forms and websites will not work. This is by design to safeguard people's personal and financial data.

### How

If you download and try a Studio competitor like LocalWP you can see that they make it as easy as flipping a switch. 
i7 SNAG- 5-30-2024 3 22 23 AM

youknowriad:
## Related issues

- Fixes #191
- Builds on top of #994

## Summary

This PR adds SSL/HTTPS support to WordPress Studio, allowing users to create sites with custom domains and secure connections. Key features include:

  - Added HTTPS to WordPress Studio sites
  - Implemented certificate management system with auto-generation and trust helpers
  - Updated the proxy server to handle HTTPS

 ## Details

  - Certificate authority generation and site certificate issuance
  - Automatic certificate trust helpers with OS-specific instructions
  - UI updates to site creation form for enabling HTTPS
  - Menu option to help users trust certificates

  These changes make WordPress Studio sites more realistic for local development by supporting custom domains with proper HTTPS, closely mimicking production environments.

## Testing instructions

 - Create a new site
 - Check "use custom domaine" and "use https" (they go hand in hand)
 - On newer MacOS versions, it seems it's not possible to install the root certificate automatically, so you can open the "Help > Trust certificate" dialog which provides instructions on how to do so.
 - If you don't install the certificate, you'll see the warning in the browser when trying to access the site.

I didn't test Windows yet.

The PR is largely unpolished at the moment but I believe it's a decent start, so we can start looking at polishing the UI and the implementation.


andreilupu:
Yaaaay 😃  I can't wait for this feature to land in production.

As far as I can see, this feature prefers to use only one Certificate for all the sites, opposed to LocalWP which creates one for each site. This is already better.

Trying to help somehow, I've built this branch and started a new WP instance with HTTPS checked. ~On my end, macOS, the site is not accessible(`ERR_CONNECTION_CLOSED`) even if I trust the certificate. After I find how to debug PHP/nginx logs, I'll be back with more details.~ <- scratch this, it was a port conflict with LocalWp.

Some feedback about the `Your connection is not private` warning when the certificate is not trusted yet. Some people might not know how to deal with it, and the small notice on the site creation is not enough. Maybe we can put another alert box in the Overview tab to inform the user that they haven't trusted the certificate yet?
I'm pretty sure this error can be matched when the Theme preview thumbnail is updated.

youknowriad:
> As far as I can see, this feature prefers to use only one Certificate for all the sites, opposed to LocalWP which creates one for each site. This is already better.

This might work similarly to Local (although I don't know Local very well). You have to trust what is called the "root" certificate but we do generate a certificate for each site internally, but they will just work when you trust the "root" certificate.

matt-west (commented):
Thanks @youknowriad!

The dialog with the instructions for trusting a certificate disappears once you click the "Open Certificate File" button. If you didn’t read the instructions fully—which many won’t—you’ll now have to go back and open the dialog again to find out what to do next.

Screenshot 2025-03-10 at 11 37 05

Perhaps we could show these instructions in the site creation modal once you click the link instead?

Let’s add a row to the Settings tab that shows if SSL is enabled. If it is, we should display a Trust link that shows a modal with the instructions for trusting the certificate.

Screenshot 2025-03-10 at 11 25 13

It would be great to have an option to enable SSL for existing sites too. I assume that will require the ability to add custom domains to existing sites first though.

youknowriad:
> Let’s add a row to the Settings tab that shows if SSL is enabled. If it is, we should display a Trust link that shows a modal with the instructions for trusting the certificate.

I'll add a row to just show the value but no link at the moment. We need a generic solution to be able to change the url of the WP site. We should create a dedicate issue for that. It applies to both changing the domain and when you trust/untrust a site.

youknowriad:
> Perhaps we could show these instructions in the site creation modal once you click the link instead?

Do you mean show the instructions inline? I can do that, I'm wondering if it's too much information inline though.

matt-west:
@youknowriad Yes. We can keep the existing link, then show the instructions inline once it’s clicked.

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,395 @@
...
+import { getUserDataCertificatesPath } from 'src/storage/paths';
+import { shellOpenExternalWrapper } from './shell-open-external-wrapper';
+
+// Certificate configuration
+const CA_NAME = 'WordPress Studio CA';
+const CA_CERT_VALIDITY_DAYS = 3650; // 10 years
+const SITE_CERT_VALIDITY_DAYS = 825; // a little over 2 years
+const CERT_DIRECTORY = getUserDataCertificatesPath();
+const CA_CERT_PATH = path.join( CERT_DIRECTORY, 'studio-ca.crt' );
+const CA_KEY_PATH = path.join( CERT_DIRECTORY, 'studio-ca.key' );
```
mcsf, line 68: Let's avoid collisions between the root CA and the domain-specific ones. With the current naming schema, we have to actively guard against users choosing the domain "studio-ca". For example:

```js
const rootCA = path.join( CERT_DIRECTORY, 'studio-ca.crt' )
const domainCA = path.join( CERT_DIRECTORY, `domain-${ domain }.crt` )
```

IMO it's cleaner with a new directory:

```js
const rootCA = path.join( CERT_DIRECTORY, 'root.crt' )
const domainCA = path.join( CERT_DIRECTORY, 'domains', `${ domain }.crt` )
```

youknowriad:
https://github.com/user-attachments/assets/ffa678eb-5b88-4d68-87f6-9163902d5555

Is this what you had in mind @matt-west ? I find it a bit weird that you have to click and scroll multiple times. What if instead we keep the dialog but ensure it stays open when you open the certificate.

Let me know what you prefer.

matt-west:
@youknowriad Yes, that’s what I thinking. I agree that it feels a bit to have all these extra section appear though.

If we can keep the dialog open, that’s a good compromise.

src/lib/certificate-manager.ts
youknowriad: Good call.

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,375 @@
...
+ */
+async function trustRootCA( showResult = false ): Promise< void > {
+	try {
+		const platform = process.platform;
+
+		if ( platform === 'darwin' ) {
+			// macOS - Use sudo to add to system keychain
+			await new Promise< void >( ( resolve, reject ) => {
+				sudo.exec(
+					`security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain "${ CA_CERT_PATH }"`,
```
mcsf: TLS is way out of my comfort zone, so please bear with me.

Why add the CA as a trusted certificate to the system-wide keychain? Wouldn't it be enough (and more conservative) to add it to the current user's (`~/Library/Keychains/login.keychain-db`)? If so, I don't think you need `sudo`.

youknowriad: In my test it was necessary to install the certificate system wide. But could use more testing.

mcsf:
Something that I think is worth trying in the certificate manager:

1. Generate a _Root CA_, self-sign it, and trust it system-wide
2. Using the Root CA, generate an _Intermediate CA_ restricted to only issue certificates in the `*.wp.local` domain
3. **Delete** the Root CA's private key
4. Ensure the Intermediate CA's files are only accessible to the user (`chmod -go $f`)
5. From that moment on, only use the Intermediate CA to issue/manage local sites' certificates

First I thought of simply setting `chmod 000` on the Root CA, but now I think it's worth trying this more drastic approach. It seems to me that, theoretically, this would guarantee that any unwanted access to these keys could only be used to impersonate Studio sites, nothing else.

I believe the procedure goes like this:

* Issue a Root CA, as we already do
* Generate a private key for the Intermediate CA
* Create a Certificate Signing Request (CSR) for the Intermediate CA:
  * Set up a configuration including the Subject Alternative Name (SAN) extension for `*.wp.local`
  * Generate a CSR for that configuration 
* Sign the Intermediate CA's **CSR** with the Root CA
  * In OpenSSL, the new CA can be verified with `openssl x509 -in intermediateCA.crt -noout -text`, which should reveal the correct SAN
* Hopefully, then, delete the Root CA's private key

matt-west:
Testing this again, it feels like there are way too many pop-ups if I try to trust the certificate while creating the site. I get lost in all the different dialogs and windows that appear.

Screenshot 2025-03-14 at 11 31 30


I don’t think we should introduce the flow for trusting a certificate as part of the site creation flow. Instead this link should open the browser with a help doc containing the instructions. I can easily keep that alongside Studio while I set up the site, whereas the current dialog prevents me from interacting with the Studio UI.

Screenshot 2025-03-14 at 11 13 39

We then need a link on the site settings tab to open the certificate file. We can include a brief explanation of why the certificate needs to be trusted and a link to the help doc too.

image

Is this possible @youknowriad?

youknowriad:
A link on the site settings tab is possible but only after this other PR lands #1064 :)

matt-west:
@youknowriad Why? That PR allows you to change the custom domain, but my suggestion is just to display the "Trust certificate" link + copy for sites that already have SSL enabled. We would hide it for sites that don’t have SSL enabled.

We can follow up in another PR to add the ability to enable SSL on an existing site.

fredrikekelund:
@mcsf, your [proposal](https://github.com/Automattic/studio/pull/1034#issuecomment-2724236891) seems quite sound, but two questions popped up for me:

1. Do we need the intermediate certificate? Can't we put the same domain restrictions on the root certificate?
2. If we make it so `*.wp.local` are the only domains that support HTTPS, we'd need to update the custom domains feature accordingly. I think this is a reasonable limitation from a security perspective, but we need to consider this from a UI standpoint, @youknowriad.

youknowriad:
@matt-west because that PR introduces the code that updates the domain name when you edit site settings which is required when the url moves from http to https.

youknowriad:
>  That PR allows you to change the custom domain, but my suggestion is just to display the "Trust certificate" link + copy for sites that already have SSL enabled. We would hide it for sites that don’t have SSL enabled.

So you're saying we should keep the "toggle" on the creation modal, I thought you wanted to remove it from there. Ok, yes, this change we can do here, without waiting for the other PR :) Sorry for the confusion.

youknowriad:
@matt-west So is your suggestion to create a page on https://developer.wordpress.com/docs/developer-tools/studio/ for this help? I think that makes sense but how do we "release that"? Should I just add a hidden page for now and we add it to the menu after release or something?

youknowriad:
I created this draft page, we can fill its content later https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/

fredrikekelund:
I've mostly wrapped my head around this now. A few high-level thoughts:

1. This approach of trusting only a root certificate is more convenient but inherently less secure. It's a big potential attack vector and we should apply restrictions along the lines of what @mcsf suggested if we proceed with this solution.
2. If we do indeed proceed with @mcsf's suggestion, only `*.wp.local` domains will be supported for HTTPS. This is definitely a limitation, but I think it's smart to lean into this differentiator. The upside is that we can default to HTTPS for new custom domain sites and make the setup process pain-free, which is very cool. We just have to embrace the `*.wp.local` name limitation for domains.
4. Newer macOS versions don't seem to support automatically trusting certificates. This [limitation](https://localwp.com/help-docs/getting-started/managing-local-sites-ssl-certificate-in-macos/) has also been identified by Local. Considering this, I think it's better not to implement this feature at all for Mac. Instead, we could focus on having a UI indicator to determine if the certificate is correctly trusted. 
5. @matt-west, I think adding a new settings tab or a standalone wizard modal for configuring the root certificate would be very cool. It's more work than pointing to an online doc, but also a nicer experience. We'd need a button for opening the certificate file, some instructions for how to add it to the system keychain (and trust it), and lastly a polling indicator that automatically confirms when the user has completed the whole operation.
6. I suggest ignoring Linux for now. This PR is complex enough as it is.

youknowriad:
> If we do indeed proceed with @mcsf's suggestion, only *.wp.local domains will be supported for HTTPS. This is definitely a limitation, but I think it's smart to lean into this differentiator. The upside is that we can default to HTTPS for new custom domain sites and make the setup process pain-free, which is very cool. We just have to embrace the *.wp.local name limitation for domains.

I also agree that it's better to have this limitation than having to trust a certificate for each site. We can also accept all `.local` domains it's quite common.

matt-west:
> So you're saying we should keep the "toggle" on the creation modal

@youknowriad Yep, that's correct. We keep the toggle, but just update the link in the note to open the help doc instead of the dialog.

> So is your suggestion to create a page on https://developer.wordpress.com/docs/developer-tools/studio/ for this help? I think that makes sense but how do we "release that"?

Yep. I’m not familiar with the release process for help docs, but @Automattic/yolo may be able to help as they’ve done a bunch for Studio.

> Newer macOS versions don't seem to support automatically trusting certificates. This [limitation](https://localwp.com/help-docs/getting-started/managing-local-sites-ssl-certificate-in-macos/) has also been identified by Local. Considering this, I think it's better not to implement this feature at all for Mac. Instead, we could focus on having a UI indicator to determine if the certificate is correctly trusted.

I agree. If it’s unlikely to work we shouldn’t present the option to automatically trust certificates.

> @matt-west, I think adding a new settings tab or a standalone wizard modal for configuring the root certificate would be very cool. It's more work than pointing to an online doc, but also a nicer experience. We'd need a button for opening the certificate file, some instructions for how to add it to the system keychain (and trust it), and lastly a polling indicator that automatically confirms when the user has completed the whole operation.

That would be ideal. I think we can leave that for a future iteration though. The help doc + links in the settings tab are okay for the V1 of this feature.

mcsf:
> @mcsf, your [proposal](https://github.com/Automattic/studio/pull/1034#issuecomment-2724236891) seems quite sound, but two questions popped up for me:
> 
> 1. Do we need the intermediate certificate? Can't we put the same domain restrictions on the root certificate?

As best as I could tell, yes, unfortunately. I think the crux is to understand the relationship between certificates and certificate authorities.
* My understanding of TLS is that a **certificate** can be configured in several ways, including setting restrictions like the domains for which it is valid. But an **authority** is just an authority and isn't defined by anything other than the chain of trust: it needs to be backed by another authority (via a certificate) or be a root authority. 
* We need a root CA (because none of the existing "real" CAs will back our authority or certificates), and — as far as I could tell — definitionally this new root CA will act as an unquestioned source of trust. So the next best thing we could do is use our root CA to sign a certificate — which itself **can** be restricted via the SAN extension — to back an intermediate CA of our making. By virtue of the certificate linking the two authorities, our intermediate CA can then only sign for `*.wp.local` domains.

> 2. If we make it so `*.wp.local` are the only domains that support HTTPS, we'd need to update the custom domains feature accordingly. I think this is a reasonable limitation from a security perspective

Right, I didn't consider that currently users could opt out of those subdomains. I do think it's worth the compromise, but I'll let you all decide.

* * *

Of course, this is all predicated on my interpretation being correct and my plan actually working. :)

fredrikekelund:
AFAICT, the X509 name constraints extensions ([explainer](https://www.sysadmins.lv/blog-en/x509-name-constraints-certificate-extension-all-you-should-know.aspx)) should technically allow us to restrict our root certificate in this way without an intermediate certificate.

The only problem is that node-forge doesn't support this extension (see https://github.com/digitalbazaar/forge/issues/958). 

`subjectAltName` (SAN) does seem to be supported by node-forge, though.

mcsf:
> AFAICT, the X509 name constraints extensions ([explainer](https://www.sysadmins.lv/blog-en/x509-name-constraints-certificate-extension-all-you-should-know.aspx)) should technically allow us to restrict our root certificate in this way without an intermediate certificate.
> 
> The only problem is that node-forge doesn't support this extension (see [digitalbazaar/forge#958](https://github.com/digitalbazaar/forge/issues/958)).

@fredrikekelund: good find! This is what I was initially hoping to see but couldn't find.

> `subjectAltName` (SAN) does seem to be supported by node-forge, though.

Interesting. The issue you mentioned has a suggestion perhaps worth trying.

@youknowriad: am I right to assume node-forge was chosen because it's a native JS implementation, and thus is most likely to be portable across platforms? [As they themselves warn](https://www.npmjs.com/package/node-forge#security-considerations), it's not as battle-tested as something like OpenSSL. I'm not trying to derail the PR by suggesting we switch direction this late; just curious.

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,333 @@
...
+	fs.writeFileSync( CA_KEY_PATH, keyPem );
+
+	// Trust the CA certificate
+	await trustRootCA();
+
+	return { cert: certPem, key: keyPem };
+}
+
+export async function openCertificate() {
+	shellOpenExternalWrapper( `file://${ CA_CERT_PATH }` );
```
```suggestion
	shell.showItemInFolder( CA_CERT_PATH );
```
Quick note that we probably want to use this method when showing the root certificate to users.

youknowriad:
Ok I've finally managed to get the name constraint added properly, I still need to add validation for the domain creation but I left it as is for now to easy testing with "invalid domains".

Let me know what you think :)

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,391 @@
...
+}
+
+export async function openCertificate() {
+	shell.showItemInFolder( CA_CERT_PATH );
+}
+
+/**
+ * Trust the root CA certificate in the system trust store
+ */
+async function trustRootCA( showResult = false ): Promise< void > {
```
youknowriad: We can in theory remove this function but since it works for a large part of our users (windows probably and older macOS) I wonder if we should keep it or not. Would love some testing on Windows.

fredrikekelund:
Off-topic, but just FYI, @youknowriad: you don't have to rebase branches on `trunk`. Merging the latest changes from `trunk` works fine and makes it easier to pull the latest changes.

youknowriad:
@fredrikekelund haha :) I do prefer rebasing, maybe it's just a habit of 8 years of Gutenberg. But yeah I know I can merge and only fix the conflicts once. Maybe I should start getting used to it.

src/lib/certificate-manager.ts
fredrikekelund: Given the high degree of complexity here, I would advise that we maintain a single installation flow for Mac (i.e., remove it from this function).

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,391 @@
...
+								title: 'Certificate Installed',
+								message: 'Root certificate has been installed successfully.',
+								detail: 'Please restart your browsers for the changes to take effect.',
+								buttons: [ 'OK' ],
+							} );
+						}
+						resolve();
+					}
+				} );
+			} );
```
fredrikekelund: Re-iterating the comment from my previous review: I advise skipping this for now to reduce complexity. There's no official Linux support yet, and testing this change on Mac and Windows is complex enough as it is.

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,391 @@
+import * as crypto from 'crypto';
+import { dialog, shell } from 'electron';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as Sentry from '@sentry/electron/main';
+import * as forge from 'node-forge';
+import sudo from 'sudo-prompt';
```
mcsf: @youknowriad: this needs to be migrated to `@vscode/sudo-prompt` (#1065).

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,391 @@
...
+						if ( error ) {
+							console.error( 'Error adding certificate to system trust store:', error );
+							if ( showResult ) {
+								dialog.showMessageBox( {
+									type: 'error',
+									title: 'Certificate Trust Failed',
+									message: 'Failed to install certificate automatically.',
+									detail:
+										'Try the manual installation process or restart the application with administrator privileges.',
+									buttons: [ 'OK' ],
```
mcsf: These and other message boxes should be translatable, right?

mcsf:
Unless it's only happening to me, you're probably aware that Studio can fail when attempting to obtain a thumbnail for SSL-enables sites, resulting in a blank thumbnail and messages in the console:

Screenshot 2025-03-19 at 16 20 15

youknowriad:
> Unless it's only happening to me, you're probably aware that Studio can fail when attempting to obtain a thumbnail for SSL-enables sites, resulting in a blank thumbnail and messages in the console:

It only happens if you didn't "trust" the CA certificate right?

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,290 @@
...
+			cRLSign: true,
+		},
+		{
+			name: 'extKeyUsage',
+			serverAuth: true,
+			clientAuth: true,
+		},
+		// Add properly encoded nameConstraints extension
+		// Using "wp.local" instead of ".wp.local" for better OS compatibility
+		// while still enforcing the domain constraint
```
fredrikekelund: Could you elaborate on which OS compatibility issues `.wp.local` might cause?

youknowriad:
This should be ready for more reviews.

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,290 @@
...
+		nameConstraintsComponents
+	);
+
+	// Create the extension object
+	return {
+		id: '2.5.29.30', // nameConstraints OID
+		critical: true, // This extension should be marked critical
+		value: nameConstraintsValue,
+	};
+}
```
fredrikekelund: This is very impressive and seems to work as expected in my testing (generating certificates for `*.wp.local` domains works, but not for other domains) 👍 Ideally, someone who is familiar with cryptography would review this code before we land the PR to verify that there are no glaring omissions. 

If we can't find someone else to do that, then I'll do my best to look closer at the RFC, but I have no previous familiarity.

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,290 @@
...
+
+	// Convert to PEM format
+	const certPem = forge.pki.certificateToPem( cert );
+	const keyPem = forge.pki.privateKeyToPem( keys.privateKey );
+
+	// Save to files
+	fs.writeFileSync( CA_CERT_PATH, certPem );
+	fs.writeFileSync( CA_KEY_PATH, keyPem );
+
+	// Trust the CA certificate
```
fredrikekelund: I don't think the number of comments here is entirely justified. `// Set extensions`, for example, is pretty self-explanatory. I'll leave it up to you which ones to remove exactly, but we could definitely have fewer.

src/components/site-form.tsx
fredrikekelund, line 1: We should add some logic here that limits which custom domain names can be used when `enableSSL` is enabled. I.e., enforce that users follow the `*.wp.local` pattern.

src/components/site-form.tsx
```diff
@@ -475,9 +487,15 @@ export const SiteForm = ( {
...
+											<div className="flex items-center gap-2 mt-4">
+												<input
+													type="checkbox"
+													id="enable-ssl"
+													checked={ enableSSL }
+													onChange={ ( e ) => setEnableSSL( e.target.checked ) }
+												/>
+												<label htmlFor="enable-ssl">{ __( 'Enable SSL' ) }</label>
 											</div>
 										) }
```
fredrikekelund: It would be really cool to enable this by default if the root certificate is trusted on the system. Do you think that's a good fit for this PR, @youknowriad, or should we pursue that separately?

fredrikekelund:
Reviewing now 👍

fredrikekelund:
I ran into an issue adding the root certificate to my keychain. This is easily resolved by adding it to the `login` keychain instead of `Local Items`, but making a note of it for documentation purposes later.

Screenshot 2025-03-20 at 13 26 41

src/ipc-types.d.ts
```diff
@@ -21,7 +21,10 @@ interface StoppedSiteDetails {
 	phpVersion: string;
 	wpVersion?: string;
 	customDomain?: string;
+	enableSSL?: boolean;
```
```suggestion
	enableHttps?: boolean;
```
I know "SSL" is often used interchangeably with "HTTPS", but I don't believe we actually use any SSL here. Instead of changing to `enableTLS`, `enableHttps` seems to convey the meaning more effectively.

src/ipc-handlers.ts
```diff
@@ -54,6 +54,10 @@ import { SiteServer, createSiteWorkingDirectory } from 'src/site-server';
 import { DEFAULT_SITE_PATH, getResourcesPath, getSiteThumbnailPath } from 'src/storage/paths';
 import { loadUserData, saveUserData } from 'src/storage/user-data';
 import { DEFAULT_PHP_VERSION } from 'vendor/wp-now/src/constants';
+import {
+	generateSiteCertificate,
+	openCertificate as openCertificateDialog,
```
fredrikekelund, line 60:
```suggestion
	openCertificate,
```
Seems like a leftover

src/lib/proxy-server.ts
```diff
@@ -35,93 +49,167 @@ async function getSiteByHost( domain: string ): Promise< SiteDetails | null > {
...
+		target: `http://localhost:${ site.port }`,
+		xfwd: true, // Pass along x-forwarded headers
+		headers,
+	} );
+}
+
+export async function checkPortInWindows( port: number ): Promise< boolean > {
+	// On Windows, node doesn't throw an error if port is busy, so we use the net module to explicitly check
+	// if it's possible to establish a TCP connection to that port (meaning it's busy)
+	if ( process.platform === 'win32' ) {
```
```suggestion
// On Windows, node doesn't throw an error if port is busy, so we use the net module to explicitly check
// if it's possible to establish a TCP connection to that port (meaning it's busy)
export async function checkPortOnWindows( port: number ): Promise< boolean > {
	if ( process.platform !== 'win32' ) {
		return true;
	}
```
Nit, but the comment placement and the function name could be tweaked. Seems logical to return early instead of having a big `if` wrapper, too.

src/lib/proxy-server.ts
```diff
@@ -35,93 +49,167 @@ async function getSiteByHost( domain: string ): Promise< SiteDetails | null > {
...
+							);
+							cb( new Error( `No certificates available for ${ servername }` ) );
+							return;
+						}
+
+						const ctx = require( 'tls' ).createSecureContext( {
+							key: site.tlsKey,
+							cert: site.tlsCert,
+							version: 'TLSv1.2',
+						} );
```
```suggestion
						const ctx = tls.createSecureContext( {
							key: site.tlsKey,
							cert: site.tlsCert,
							minVersion: 'TLSv1.2',
						} );
```
Let's import `tls` at the top of the file instead. `version` also doesn't seem to be supported by the node TLS module. I'm guessing `minVersion` is what we want.

mcsf:
> > Unless it's only happening to me, you're probably aware that Studio can fail when attempting to obtain a thumbnail for SSL-enables sites, resulting in a blank thumbnail and messages in the console:
> 
> It only happens if you didn't "trust" the CA certificate right?

Added at the browser level only. But I did delete all my certificates to see if `trustRootCA` would work at all.

src/lib/proxy-server.ts
```diff
@@ -35,93 +49,167 @@ async function getSiteByHost( domain: string ): Promise< SiteDetails | null > {
...
-				} );
-		} );
+		// Start HTTPS server if not already running
+		if ( ! isHttpsProxyRunning ) {
+			await checkPortInWindows( 443 );
+			const defaultOptions = {
+				SNICallback: async (
+					servername: string,
+					cb: ( err: Error | null, ctx?: SecureContext ) => void
+				) => {
```
```suggestion
			const defaultOptions: https.ServerOptions = {
				SNICallback: async ( servername, cb ) => {
```
Better type definition

src/lib/proxy-server.ts
```diff
@@ -134,42 +222,65 @@ export async function startProxyServer(): Promise< boolean > {
...
+						console.log( 'HTTPS Proxy server stopped' );
+						resolve();
+					} );
+				} )
+			);
+		} else {
+			isHttpsProxyRunning = false;
+		}
+
+		Promise.all( promises ).then( () => resolve() );
```
```suggestion
		await Promise.all( promises );
```
Optional nit, but if we made the function async, we could do this instead.

src/lib/proxy-server.ts
```diff
@@ -35,93 +49,167 @@ async function getSiteByHost( domain: string ): Promise< SiteDetails | null > {
...
 /**
- * Starts the proxy server for sites with custom domains
+ * Common handler for both HTTP and HTTPS requests
  */
-export async function startProxyServer(): Promise< boolean > {
-	if ( isProxyRunning ) return true;
+async function handleProxyRequest(
+	req: http.IncomingMessage,
+	res: http.ServerResponse,
+	isHttps = false
```
```suggestion
	isHttps: boolean
```
I don't really see a point in assigning a default value here.

youknowriad:
@mcsf Per @fredrikekelund's suggestion I removed the handling of MacOS on `trustRootCA` since it only works in older MacOS versions.

fredrikekelund (commented):
This is looking pretty great overall. Very nice work, @youknowriad 👍 I still need to do some more testing and go through the changes some more. Figured I'd share the comments I left so far so it's more digestible for you.

youknowriad:
The add/edit/onboarding/form are unnecessary complex and convoluted. All these components should ideally be refactored and simplified. There's duplication and unclear data flows at the moment. But it's not specific to this PR, so I'm not going to touch these at the moment.

src/components/site-form.tsx
```diff
@@ -475,9 +487,15 @@ export const SiteForm = ( {
...
-												{ __( 'Your system password will be required to set up the domain.' ) }
+										{ useCustomDomain && setEnableSSL && (
+											<div className="flex items-center gap-2 mt-4">
+												<input
+													type="checkbox"
+													id="enable-ssl"
+													checked={ enableSSL }
+													onChange={ ( e ) => setEnableSSL( e.target.checked ) }
+												/>
+												<label htmlFor="enable-ssl">{ __( 'Enable SSL' ) }</label>
```
fredrikekelund:
```suggestion
												<label htmlFor="enable-ssl">{ __( 'Enable HTTPS' ) }</label>
```
HTTPS conveys the meaning more effectively and is technically more accurate.

src/lib/certificate-manager.ts
youknowriad: This was a suggestion that I found by asking the AI. and confirmed by looking online a bit 

https://serverfault.com/questions/975838/x509-certificate-with-excluded-subtree-violation-issue

Basically the only difference in theory between both is that we also allow wp.local but other than that it should be the same and "safer"

src/hooks/use-site-details.tsx
```diff
@@ -194,6 +195,7 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 			siteName?: string,
 			wpVersion?: string,
 			customDomain?: string,
+			enableSSL?: boolean,
 			callback?: ( site: SiteDetails | void ) => Promise< void >
```
fredrikekelund:
```suggestion
			callback?: ( site: SiteDetails ) => Promise< void >
```
We only ever call the callback if we have a `SiteDetails` object. Let's take this opportunity to update the type signature. This will allow us to also simplify the callback in `src/hooks/use-add-site.ts`. 

I agree with your previous point that the data flows here are convoluted for no good reason and we should do more work on that. This is very low-hanging fruit, though.

src/lib/certificate-manager.ts
youknowriad: I have to admit that I wouldn't have been able to do this myself manually. I've used AI extensively here :). I did compare it to an example I've found online that was using IP addresses.

src/hooks/use-site-details.tsx
```diff
@@ -22,6 +22,7 @@ interface SiteDetailsContext {
 		siteName?: string,
 		wpVersion?: string,
 		customDomain?: string,
+		enableSSL?: boolean,
 		callback?: ( site: SiteDetails | void ) => Promise< void >
```
fredrikekelund:
```suggestion
		callback?: ( site: SiteDetails ) => Promise< void >
```
Per my other suggestion of updating the callback signature.

src/lib/certificate-manager.ts
youknowriad: Agreed :)

src/components/site-form.tsx
youknowriad: I'm thinking of adding a global rule to limit all the custom domains to `*.wp.local` WDYT?

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,290 @@
+import * as crypto from 'crypto';
+import { shell } from 'electron';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as Sentry from '@sentry/electron/main';
+import sudo from '@vscode/sudo-prompt';
+import * as forge from 'node-forge';
+import { getUserDataCertificatesPath } from 'src/storage/paths';
```
```suggestion
import { shell } from 'electron';
import crypto from 'node:crypto';
import fs from 'node:fs';
import path from 'node:path';
import Sentry from '@sentry/electron/main';
import sudo from '@vscode/sudo-prompt';
import forge from 'node-forge';
import { getUserDataCertificatesPath } from 'src/storage/paths';
```
For consistency with other imports.

src/ipc-types.d.ts
youknowriad: If you look at the commit history, I went back and forth :P. I'm not strongly opinionated

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,290 @@
...
+
+	// Generate a key pair
+	const keys = forge.pki.rsa.generateKeyPair( 2048 );
+
+	// Create a certificate
+	const cert = forge.pki.createCertificate();
+
+	// Set basic certificate fields
+	cert.publicKey = keys.publicKey;
+	cert.serialNumber = '01' + crypto.randomBytes( 19 ).toString( 'hex' ); // 40 hex characters
```
fredrikekelund: Doesn't look like a big deal either way, but I'm curious why we start with `'01'` here?

src/components/site-form.tsx
youknowriad: I had it as HTTP initially but I switched to be consistent with Local, users might already be familiar with that. WDYT?

src/components/site-form.tsx
youknowriad: Happy to try that. Let's keep it separate

src/components/content-tab-settings.tsx
```diff
@@ -63,13 +65,43 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
...
+						<div>
+							<span>{ selectedSite.enableSSL ? __( 'Enabled' ) : __( 'Disabled' ) }</span>{ ' ' }
+							{ selectedSite.enableSSL && (
+								<Button variant="link" onClick={ () => getIpcApi().openCertificate() }>
+									{ __( 'Trust Certificate' ) }
+								</Button>
+							) }
+						</div>
+						{ selectedSite.enableSSL && (
+							<div className="mt-1">
```
fredrikekelund:
```suggestion
							<div className="mt-1 max-w-96">
```
Makes for a better and more readable look

src/components/site-form.tsx
fredrikekelund, line 525: IMO, we should add an explainer here similar to the custom domain checkbox. Maybe we only need it for Mac if the Windows automation works… Something like:

> You need to manually add the Studio certificate authority to your keychain and trust it. For security reasons, we only allow `*.wp.build` to use HTTPS.

fredrikekelund:
> Unless it's only happening to me, you're probably aware that Studio can fail when attempting to obtain a thumbnail for SSL-enables sites, resulting in a blank thumbnail and messages in the console:

I suggest editing `src/screenshot-window.ts` like this to make the Electron browser window that takes the screenshot accept unsafe HTTPS certificates. This worked in my testing:

```tsx
	const newSession = session.fromPartition( crypto.randomUUID() );

	// Accept unsafe HTTPS certificates
	newSession.setCertificateVerifyProc( ( request, callback ) => {
		callback( 0 );
	} );

	const window = new BrowserWindow( {
		height: SCREENSHOT_HEIGHT,
		width: SCREENSHOT_WIDTH,
		show: false,
		webPreferences: { session: newSession },
	} );
```

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -278,6 +282,18 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
...
+
+								{ useCustomDomain && (
+									<div className="flex items-center gap-2 mt-4">
+										<input
+											type="checkbox"
+											id="enable-ssl"
+											checked={ enableSSL }
+											onChange={ ( e ) => setEnableSSL( e.target.checked ) }
+										/>
+										<label htmlFor="enable-ssl">{ __( 'Enable SSL' ) }</label>
```
fredrikekelund:
```suggestion
										<label htmlFor="enable-ssl">{ __( 'Enable HTTPS' ) }</label>
```
Per my other suggestion of updating the label

fredrikekelund (commented):
Another batch of comments 📤

Tested and confirm that editing the HTTPS option works for existing sites. Impressive 👍

I've yet to test Windows.

src/components/site-form.tsx
fredrikekelund: Given that custom domain support is already released, this would be a regression for anyone who's already configured a different custom domain.

I sympathize with your logic, but I think it's better only to enforce this when HTTPS is enabled.

src/components/site-form.tsx
fredrikekelund: I think HTTPS conveys the meaning better here. We can consider this from an SEO perspective when writing from the blog and make sure to mention "SSL" in the blog post.

fredrikekelund (commented):
Windows worked like a charm in my testing 👍 Very smooth when it's all automated.

The functionality here looks solid to me and tests well. Let's work through the comments on the PR and I'll approve after that.

src/lib/certificate-manager.ts
fredrikekelund: I don't see how that Serverfault link advises against using leading dot notation. It seems like a case where the nameConstraints extension was misconfigured by including an IP address in the excluded subtree. 

Only allowing subdomains of `wp.local` seems quite logical to me.

mcsf:
I tested this some more and I can confirm that:

* After manually adding the CA certificate to my system (macOS using Keychain Access):
  * A new site with custom domain under `wp.local` and with SSL is correctly accepted by the browser
  * A new site with a custom domain elsewhere (`foo.wp`) and with SSL is correctly rejected by the browser
  * Thumbnails are rendered in Studio for SSL domains under `wp.local`

However:
* When the domain is not under `wp.local`, the Overview screen is rendered as if it is still loading, with flashing placeholders (see screenshot below).
* There is nothing either preventing or explaining to the user what happens when they choose a non-wp.local domain with SSL enabled.
* Site initialisation is feeling a lot slower to me (and on a new machine). I can't guarantee it's directly tied to this PR, or its original base (custom domains), or just a coincidence.

Screenshot 2025-03-20 at 15 46 04

src/components/site-form.tsx
youknowriad: For me it's not really a problem, existing custom domains will continue to work.

My main worry with allowing all domains and only a subset with useHTTPS is that the UI to explain this to the users could be a bit confusing.

fredrikekelund: @matt-west @wojtekn, any input here?

src/ipc-handlers.ts
youknowriad: This is not possible because we expose a function with the same name.

fredrikekelund: Ah. Then let's call it `openCertificateInFolder` or something. `openCertificateDialog` is not accurate.

src/lib/certificate-manager.ts
youknowriad: No particular reason, the AI had initially generated '01' as a static number and I wanted to make it random. But keeping 01 was not necessary.

src/components/site-form.tsx
matt-west: Could we limit to `*.local` instead? I'm okay with adding a limit, but that feels a bit more flexible. Unless we really want the more branded `*.wp.local` domains.

youknowriad:
> When the domain is not under wp.local, the Overview screen is rendered as if it is still loading, with flashing placeholders (see screenshot below).

I can't reproduce this.

> There is nothing either preventing or explaining to the user what happens when they choose a non-wp.local domain with SSL enabled.

Yes, this is pending a decision above, either limit the creation of such domains or add a message.

> Site initialisation is feeling a lot slower to me (and on a new machine). I can't guarantee it's directly tied to this PR, or its original base (custom domains), or just a coincidence.

I've noticed this before in other PRs too.

---

Seems like all the comments are addressed. The last blocker is a decision on what to do here https://github.com/Automattic/studio/pull/1034/files/cbbabe32b15459d1e8a7ac0a72e59f58c6eb46ad#diff-e2be6df548f75063abd8db0f241c35626f7af6a1c6a10c429935de5f902dbbc6

src/components/site-form.tsx
youknowriad: It is definitely possible to limit to `*.local` I'll defer to you all for this.

fredrikekelund:
> I can't reproduce this.

Was that before making the change to `src/screenshot-window.ts`?

src/components/site-form.tsx
fredrikekelund: I can't think of any legitimate security concerns with using `*.local`, and you're probably right that users would appreciate the flexibility. Let's go with that, then. 

@youknowriad, we still need to validate the custom domain input field when `useHttps` is enabled.

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,259 @@
...
+	];
+	cert.setExtensions( extensions );
+
+	// Self-sign the certificate
+	cert.sign( keys.privateKey, forge.md.sha256.create() );
+
+	const certPem = forge.pki.certificateToPem( cert );
+	const keyPem = forge.pki.privateKeyToPem( keys.privateKey );
+	fs.writeFileSync( CA_CERT_PATH, certPem );
+	fs.writeFileSync( CA_KEY_PATH, keyPem );
```
fredrikekelund, line 138: Could we change the file permissions so only the owner can read the file? We should double-check the implications of this, but it seems sensible from a security standpoint.

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,259 @@
...
+	cert.validity.notAfter = new Date( now.getTime() );
+	cert.validity.notAfter.setDate( now.getDate() + CA_CERT_VALIDITY_DAYS );
+	const attrs = [
+		{ name: 'commonName', value: CA_NAME },
+		{ name: 'countryName', value: 'US' },
+		{ name: 'organizationName', value: 'WordPress Studio' },
+		{ name: 'organizationalUnitName', value: 'Development' },
+	];
+	cert.setSubject( attrs );
+	cert.setIssuer( attrs );
```
fredrikekelund: Studio is installed on a per-user basis. I'm thinking that we should maybe include the current user name somewhere here to make it easier for users to distinguish between certificates issued by different Studio installations on the same machine

mcsf:
> > I can't reproduce this.
> 
> Was that before making the change to `src/screenshot-window.ts`?

You're right, the issue is gone. ✅

src/components/site-form.tsx
wojtekn: It seems a good approach to limit those to '*.local`.

wojtekn:
This is a significant change! 🥳  I haven't looked into code much yet, but I reviewed the discussion and tried to approach it as a general Studio user:

1. Start Studio
2. Click 'Add site', enable custom domain and HTTPS
3. Provide the password to let Studio add a hosts entry
4. Click a screenshot to open the site

The site is started, opens in the browser, and displays a certificate error. I was not sure how to proceed further.

As I read the thread above and found some clues, I navigated to Settings and located "HTTP Enabled Trust Certificate. I opened a `.crt`file and was asked to add the certificate to a keychain. I approved, restarted the browser, and still couldn't make it work. Then, I tested Local's flow and discovered that I needed to locate my certificate in the keychain and mark it as trusted. I restarted the browser, and it works fine now.

Questions related to UX:
1. I can't see any easy transition from the "Add a site" step to the "Trust Certificate". The site starts and doesn't work in the browser. Should we automatically open Settings and highlight the part with the certificate when the user adds or starts a site configured to use HTTPS and the certificate is not trusted yet? Or display dialog asking user to do so?
2. Should we add a 'Learn more' link near the 'Enable HTTPS' field, what could also improve UX issue from the previous point?
3. I noticed that the Trust Certificate link and all instructions are hidden for sites with HTTPS disabled, which makes sense. Should we hide it also for those with HTTPS enabled when the certificate is already trusted?

fredrikekelund:
> I can't see any easy transition from the "Add a site" step to the "Trust Certificate". The site starts and doesn't work in the browser. Should we automatically open Settings and highlight the part with the certificate when the user adds or starts a site configured to use HTTPS and the certificate is not trusted yet? Or display dialog asking user to do so?

Displaying a dialog seems like a good approach. Trusting the certificate is required, and dialogs can't be missed. This could be a separate PR, because we only want to open that dialog when the root certificate isn't already trusted, and we need to figure out the best way for determining that.

> Should we add a 'Learn more' link near the 'Enable HTTPS' field, what could also improve UX issue from the previous point?

I agree we should do this. I previously [suggested](https://github.com/Automattic/studio/pull/1034#discussion_r2005691642) adding an explainer text below that checkbox. 

> I noticed that the Trust Certificate link and all instructions are hidden for sites with HTTPS disabled, which makes sense. Should we hide it also for those with HTTPS enabled when the certificate is already trusted?

Seems like a good idea 👍 This could also be a follow-up PR together with the dialog that opens after adding or editing a site.

src/lib/certificate-manager.ts
youknowriad: For me, ideally this is solved by moving all Studio files and config into `~`, so I wonder if it's a stop gap solution.

youknowriad:
Since this PR is getting big and to track the improvements we can make later, I opened this [follow-up issue](https://github.com/Automattic/studio/issues/1109) to address the different points that require detection of whether the root certificate is trusted.

src/lib/certificate-manager.ts
fredrikekelund: While I don't see how moving the config files (and certs) would make a difference here, I also just realized that the MacOS and Windows keychains are already scoped to the current user, which makes this a non-issue.

src/components/site-form.tsx
youknowriad: > @youknowriad, we still need to validate the custom domain input field when useHttps is enabled.

My opinion is that we should validate it every time, not just when useHttps is enabled. And for existing sites not using .local, it's not a problem, they can continue using it.

src/components/site-form.tsx
youknowriad: I've added a message to the forms. I agree it's better to know before hand. cc @matt-west

src/components/site-form.tsx
fredrikekelund: `*.local` is more flexible than `*.wp.local`. I'm OK with enforcing that pattern for non-HTTPS sites, too.

src/components/site-form.tsx
```diff
@@ -480,9 +492,34 @@ export const SiteForm = ( {
...
+												<label htmlFor="enable-https">{ __( 'Enable HTTPS' ) }</label>
+											</div>
+										) }
+
+										{ useCustomDomain && setEnableHttps && (
 											<div className="text-a8c-gray-50 text-xs mt-2">
-												{ __( 'Your system password will be required to set up the domain.' ) }
+												{ __(
+													'You need to manually add the Studio certificate authority to your keychain and trust it.'
+												) }{ ' ' }
```
fredrikekelund:
```suggestion
												{ __(
													'You need to manually add the Studio root certificate authority to your keychain and trust it to enable HTTPS.'
												) }{ ' ' }
```
"Root certificate authority" is what it's called in the MacOS Keychain Access app. Adding `to enable HTTPS` to the end also clarifies that this instruction is only relevant if the checkbox is ticked.

Aside from tweaking the wording, let's also hide this on Windows.

src/components/site-form.tsx
```diff
@@ -480,9 +492,34 @@ export const SiteForm = ( {
...
+												<Button
+													variant="link"
+													onClick={ () => {
+														getIpcApi().openURL(
+															'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/'
+														);
+													} }
+												>
+													{ __( 'Learn how' ) }
+													<span aria-label={ __( '(opens in a new tab)' ) }>&#8599;</span>
```
fredrikekelund:
```suggestion
													<span aria-label={ __( '(opens in a web browser)' ) }>&#8599;</span>
```
I see we already have a few instances of this description in the code. "In a new tab" doesn't really make sense in a desktop app, so I would advise updating all instances of this string.

src/components/content-tab-settings.tsx
```diff
@@ -63,13 +65,43 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
...
+								<Button
+									variant="link"
+									onClick={ () => {
+										getIpcApi().openURL(
+											'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/'
+										);
+									} }
+								>
+									{ __( 'Learn how' ) }
+									<span aria-label={ __( '(opens in a new tab)' ) }>&#8599;</span>
```
fredrikekelund:
```suggestion
									<span aria-label={ __( '(opens in a web browser)' ) }>&#8599;</span>
```

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,261 @@
...
+	cert.serialNumber = crypto.randomBytes( 20 ).toString( 'hex' );
+	const now = new Date();
+	cert.validity.notBefore = now;
+	cert.validity.notAfter = new Date( now.getTime() );
+	cert.validity.notAfter.setDate( now.getDate() + CA_CERT_VALIDITY_DAYS );
+	const attrs = [
+		{ name: 'commonName', value: CA_NAME },
+		{ name: 'countryName', value: 'US' },
+		{ name: 'organizationName', value: 'WordPress Studio' },
+		{ name: 'organizationalUnitName', value: 'Development' },
```
```suggestion
```
AFAICT, this field is optional. I would argue we should just skip it.

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,261 @@
...
+		cert.serialNumber = crypto.randomBytes( 20 ).toString( 'hex' );
+		const now = new Date();
+		cert.validity.notBefore = now;
+		cert.validity.notAfter = new Date( now.getTime() );
+		cert.validity.notAfter.setDate( now.getDate() + SITE_CERT_VALIDITY_DAYS );
+		const attrs = [
+			{ name: 'commonName', value: domain },
+			{ name: 'countryName', value: 'US' },
+			{ name: 'organizationName', value: 'WordPress Studio' },
+			{ name: 'organizationalUnitName', value: 'Development Sites' },
```
```suggestion
```
Again, optional field.

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -277,6 +281,37 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
...
+										<Button
+											variant="link"
+											onClick={ () => {
+												getIpcApi().openURL(
+													'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/'
+												);
+											} }
+										>
+											{ __( 'Learn how' ) }
+											<span aria-label={ __( '(opens in a new tab)' ) }>&#8599;</span>
```
fredrikekelund:
```suggestion
											<span aria-label={ __( '(opens in a web browser)' ) }>&#8599;</span>
```

src/lib/certificate-manager.ts
```diff
@@ -0,0 +1,261 @@
...
+	const attrs = [
+		{ name: 'commonName', value: CA_NAME },
+		{ name: 'countryName', value: 'US' },
+		{ name: 'organizationName', value: 'WordPress Studio' },
+		{ name: 'organizationalUnitName', value: 'Development' },
+	];
+	cert.setSubject( attrs );
+	cert.setIssuer( attrs );
+
+	const extensions = [
```
```suggestion
	cert.setExtensions( [
```
Total nit, but this'd be nice.

fredrikekelund, line 111:
```suggestion
	cert.setExtensions( [
```
Total nit, but this'd be nice.

fredrikekelund (approved):
Great work, @youknowriad 👍 I left a couple of final comments, but they are all minor concerns.

youknowriad:
Thanks all for the help landing this. I'm very excited about this one. The project is not done yet, there's a few related issues, you can follow on Linear.

PR #1036, opened 2025-03-10
Cleanup tooltips across the app

wojtekn:
## Related issues

- Fixes p1741367797915539-slack-C06DRMD6VPZ

## Proposed Changes

I propose to clean up tooltips:
- Remove "Open URL" tooltips on links that already show URL and have an arrow that indicates it opens a link
- Remove "Open URL" on buttons which already have descriptive texts
- Remove "Open Settings" on links that is descriptive
- Fix format date issue

## Testing Instructions

- Test if tooltips are removed and elements still work correctly.

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-overview.tsx
+++ b/src/components/content-tab-overview.tsx
@@ -1,5 +1,5 @@
 import * as Sentry from '@sentry/electron/renderer';
-import { __, sprintf } from '@wordpress/i18n';
+import { __ } from '@wordpress/i18n';
 import {
 	archive,
 	code,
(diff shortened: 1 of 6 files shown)


matt-west (approved):
Looks much better. Thanks @wojtekn

ivan-ottinger (approved):
The changes look good and the related tooltips are no longer displaying. 👍🏼 

The only thing I would like to mention is the case when the site URL is longer and the window is narrower, users then won't see it in full:

Markup on 2025-03-10 at 12:04:43

In that case they can either make the window wider or click the link.

---

Date format fix looks good as well.

PR #1041, opened 2025-03-10
Auto-start site when user opens it in header or overview

wojtekn:
## Related issues

- Related to https://github.com/Automattic/dotcom-forge/issues/9139

## Proposed Changes

- I propose to automatically start the site when a user opens it in the header using 'WP admin' or 'Open site' links or screenshot in the Overview tab.

## Testing Instructions

1. Ensure site is stopped
2. Click 'WP admin', 'Open site' and screenshot
3. Confirm each action starts and site and opens it

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-overview.tsx
+++ b/src/components/content-tab-overview.tsx
@@ -19,6 +19,7 @@ import { ArrowIcon } from 'src/components/arrow-icon';
 import { ButtonsSection, ButtonsSectionProps } from 'src/components/buttons-section';
 import { useCheckInstalledApps } from 'src/hooks/use-check-installed-apps';
 import { useFeatureFlags } from 'src/hooks/use-feature-flags';
+import { useSiteDetails } from 'src/hooks/use-site-details';
 import { useThemeDetails } from 'src/hooks/use-theme-details';
 import { isMac } from 'src/lib/app-globals';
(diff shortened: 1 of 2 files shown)


ivan-ottinger (commented):
The proposed changes look good and work as expected. Clicking on the header links or site preview screenshot first starts the site. 👌🏼 

One thing I would like to share is that with the currently proposed approach users might not notice that the site is starting. Instead of waiting for it to open, they might be motivated to click on those links again (multiple times) - as it may not be obvious the click went through correctly.

Clicking the link again before the site started opens it, but lands on the `This site can’t be reached` page.

I think that if we would like to autostart the site for the user, we could consider making it clearer that their click "went through" and is actually starting the site.

An alternative solution that comes to my mind could be disabling the links momentarily until the site is started and the site / WP Admin is getting opened.

What do you think? 🤔

wojtekn:
Good catch @ivan-ottinger. I've experimented with loading indicators, but it looked overwhelming. I decided to disable buttons when the server is starting.

ivan-ottinger (approved):
Nice! The changes look good and work as intended.

All three buttons get disabled as soon as the server is starting. Once the server is running, the site or WP Admin gets opened right away.

I did not notice any regressions. Clicking the links when the site is already running works as before.

PR #1044, opened 2025-03-10
Add stats for PHP and WP versions

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10534

## Proposed Changes

- Added tracking for WordPress and PHP versions used in sites
- Created new changeWordPressVersion IPC handler to centralize WordPress version changes
- Added utility functions to convert WordPress and PHP versions to metrics for stats tracking
- Added version tracking when creating sites, updating PHP versions, and changing WordPress versions
- Added tests for the new functionality

## Testing Instructions

- Create a new site and verify that WordPress and PHP version stats are tracked
- Change the PHP version of an existing site and verify that stats are tracked
- Change the WordPress version of an existing site and verify that stats are tracked
- Run the test suite to verify that all tests pass

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -21,7 +21,13 @@ import { ARCHIVER_OPTIONS, MAIN_MIN_WIDTH, SIDEBAR_WIDTH } from 'src/constants';
 import { sendIpcEventToRendererWithWindow } from 'src/ipc-utils';
 import { ACTIVE_SYNC_OPERATIONS } from 'src/lib/active-sync-operations';
 import { bumpStat } from 'src/lib/bump-stats';
-import { getImporterMetric, StatsGroup, StatsMetric } from 'src/lib/bump-stats/types';
+import {
+	getImporterMetric,
+	StatsGroup,
+	StatsMetric,
+	getWordPressVersionMetric,
+	getPHPVersionMetric,
+} from 'src/lib/bump-stats/types';
 import { calculateDirectorySize } from 'src/lib/calculate-directory-size';
 import { download } from 'src/lib/download';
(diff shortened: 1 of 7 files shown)


fredrikekelund:
Drive-by observation is that it's not entirely clear to me how this data will be useful to us 🤔 Do we have an idea about that?

src/ipc-handlers.ts
```diff
@@ -205,6 +212,13 @@ export async function createSite(
 		}
 	}
 
+	if ( details.phpVersion ) {
+		bumpStat( StatsGroup.STUDIO_SITE_VERSIONS, getPHPVersionMetric( details.phpVersion ) );
+	}
+	if ( wpVersion ) {
+		bumpStat( StatsGroup.STUDIO_SITE_VERSIONS, getWordPressVersionMetric( wpVersion ) );
+	}
```
fredrikekelund, line 220: If we are going to track this data, wouldn't it make sense to also track how many sites opt out of selecting a specific WP or PHP version?

wojtekn:
I'm unsure what we need this data for.

By looking into PT, I see we wanted to track this metric:

> Preview sites: Track the WordPress and PHP versions being used for preview sites.

This would be about adding one line on the API endpoint side.

katinthehatsite:
> I'm unsure what we need this data for.

I initially thought that this might be useful to determine if the feature is used / being adopted. If we do not need this data for version switching within Studio, we can skip tracking these stats and instead proceed with the original plan and focus on preview sites.

bcotrim:
Closing this PR as we decided not to implement this change.

PR #1046, opened 2025-03-10
Keep existing theme and plugin directories if they are not part of the import.

### LINKED ISSUE #1045: Import from backup removes plugins/themes directories
### Quick summary

I set up a local site with themes and plugins and then tried to import from a Jetpack backup that only contained the database and media. After the import was completed, both the `plugins` and `themes` directories had been removed.

I have tested this with various configurations of backs ups and the result is the same. If the resulting backup doesn't contain the `plugins` and/or `themes` directories, then they are deleted in the local site.

### Steps to reproduce

1. Create a new site in Studio
2. Install plugins and themes or just keep the default ones.
3. Create a Jetpack backup and download a `.tar` with only the Database and Media options selected (see screenshot).
4. Import the backup via the Import/Export screen.
5. Confirm that the `plugins` and `themes` directories have been deleted

Image

I would expect that with the exception of adding media, no other changes to the filesystem should take place.

### What actually happened

The `plugins` and `themes` directories were removed.

### Impact

All

### Available workarounds?

No but the app is still usable

### Platform

_No response_

### Logs or notes

_No response_

ryanwelcher:
This PR attempts to address the issue as described in #1045. 

## Related issues

- Fixes #1045 

## Proposed Changes

During the import process, the `wp-content` directory is deleted. This PR checks to see if the incoming import contains plugins or themes and if not, they are added to list of content to keep.

## Testing Instructions
Follow the testing instructions in #1045 and confirm that the directories are not deleted.

## Pre-merge Checklist


ryanwelcher:
I've updated the PR to also include the `fonts`,`uploads`, and `languages` directories

src/lib/import-export/import/importers/importer.ts
```diff
@@ -150,8 +150,18 @@ abstract class BaseBackupImporter extends BaseImporter {
 				/^database(\/|\\)?.*/, // Match database dir and all contents
 				/^db\.php$/, // Exact match for db.php
 				/^index\.php$/, // Exact match for index.php
+				/^languages(\/|\\)?.*/, // Match languages dir and all contents
 			];
 
+			// If these directories aren't in the backup, they'll be deleted. Let's not do that.
```
sejas: I find this comment a bit confusing.  I would suggest something like:

```suggestion
			// Directories to preserve if they are not included in the backup.
```

src/lib/import-export/import/importers/importer.ts
```diff
@@ -150,8 +150,18 @@ abstract class BaseBackupImporter extends BaseImporter {
...
+			const maybeKeepWpContentDirectories = [ 'plugins', 'themes', 'fonts', 'uploads' ];
+
+			for ( const directory of maybeKeepWpContentDirectories ) {
+				if ( this.backup.wpContent[ directory as keyof WpContent ]?.length === 0 ) {
```
sejas: Let's move the type definition to the constant instead of forcing the type later.

```suggestion
			const maybeKeepWpContentDirectories: ( keyof WpContent )[] = [
				'plugins',
				'themes',
				'fonts',
				'uploads',
			];

			for ( const directory of maybeKeepWpContentDirectories ) {
				if ( this.backup.wpContent[ directory ]?.length === 0 ) {
```

sejas (approved):
It works as expected, and now partial imports retain the original folders for plugins, themes, fonts, and uploads if they are not included in the backup.

The only thing I've noticed is that, in order to identify Jetpack backups, we need to include the database in the backup. Importing only themes or plugins will result in an error. There is no need to address this in this PR.



Screenshot 2025-03-20 at 11 16 21

PR #1048, opened 2025-03-11
Account for custom domains in `updateSiteUrlToLocal`

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10588

## Proposed Changes

Account for custom domains in the `updateSiteUrlToLocal` function, which is run during site imports, so that old site URL strings are replaced with the correct new site URL string.

## Testing Instructions

1. Have a Studio site with at least one post containing an image
2. Export the site
3. Create another Studio site using a custom domain
4. Import the first site to the second site
5. Ensure that the second site loads correctly after the import finishes and that the image in the post loads correctly, too

## Pre-merge Checklist

### DIFF
--- a/src/lib/updateSiteUrlToLocal.ts
+++ b/src/lib/updateSiteUrlToLocal.ts
@@ -1,3 +1,4 @@
+import { getSiteUrl } from 'src/lib/get-site-url';
 import { SiteServer } from 'src/site-server';
 
 export const updateSiteUrlToLocal = async ( siteId: string ) => {
@@ -15,9 +16,7 @@ export const updateSiteUrlToLocal = async ( siteId: string ) => {
 		return;
 	}
 
-	const studioUrl = server.details.customDomain
-		? `http://${ server.details.customDomain }`
-		: `http://localhost:${ server.details.port }`;
+	const studioUrl = getSiteUrl( server.details );
 	const oldUrl = currentSiteUrl.trim();
 	const urlWithoutProtocol = oldUrl.replace( /^https?:\/\//, '' );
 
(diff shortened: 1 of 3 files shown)


src/lib/updateSiteUrlToLocal.ts
```diff
@@ -15,7 +15,9 @@ export const updateSiteUrlToLocal = async ( siteId: string ) => {
 		return;
 	}
 
-	const studioUrl = `http://localhost:${ server.details.port }`;
+	const studioUrl = server.details.customDomain
```
wojtekn: Would it make sense to extract it to its own object method?

fredrikekelund: Maybe… If we could add something like a `getUrl()` method to `server.details`, it'd be a no-brainer. But `server.details` is a typed `SiteDetails` object (exactly the same shape as in the config file), so there's no clear-cut opportunity to start adding methods to that object. 

We could also add a helper function like `getSiteUrl( site: SiteDetails ) => string`, but given how short and intuitive the current logic is and the inherent overhead in adding a new module, I'm not convinced we gain that much from it.

fredrikekelund: Adding a method to the `server` object (a `SiteServer` instance) could work, but there are other places in the code where we need to run an equivalent check and we typically have access to a `SiteDetails` object rather than a `SiteServer` instance. If we can't use the new method/function everywhere, it also doesn't make sense, IMO.

fredrikekelund: OK, I changed my mind 😅 I realized that we'll have HTTPS support soon enough, which will need to be considered here, too. I'll add a helper function

wojtekn: Thanks for exploring, those, looks great! 🐥

wojtekn (approved):
Works as expected. The site I imported has custom domain used in post images.

PR #1051, opened 2025-03-11
Studio: Fix Sync tab sites list misalignment

ivan-ottinger:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10605.

## Proposed Changes

- fix Sync sites list alignment

| Before | After |
|--------|--------|
| Markup on 2025-03-11 at 12:17:38 | Markup on 2025-03-11 at 12:25:10 | 

## Testing Instructions

2. Head over to the Sync tab and connect to a WPCOM site that also has a staging site.
3. Review the alignment of both sites on the list.
4. Switch through different locales and window sizes. The alignment should be correct.

## Pre-merge Checklist

### DIFF
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -213,7 +213,7 @@ const SyncConnectedSitesSection = ( {
 							key={ connectedSite.id }
 							className="flex items-center gap-2 min-h-14 border-b border-a8c-gray-0 px-8"
 						>
-							<div className="flex items-left min-w-20 me-6 shrink-0">
+							<div className="flex items-left min-w-[100px] me-4 shrink-0">
 								{ connectedSite.isStaging ? (
 									<Badge>{ __( 'Staging' ) }</Badge>
 								) : (
@@ -222,7 +222,7 @@ const SyncConnectedSitesSection = ( {
 							</div>
 							<Button
 								variant="link"
-								className="!text-a8c-gray-70 hover:!text-a8c-blueberry max-w-[100%]"
+								className="!text-a8c-gray-70 hover:!text-a8c-blueberry max-w-full overflow-hidden"
 								onClick={ () => {
 									getIpcApi().openURL( connectedSite.url );
 								} }


src/components/sync-connected-sites.tsx
```diff
@@ -213,7 +213,7 @@ const SyncConnectedSitesSection = ( {
 							key={ connectedSite.id }
 							className="flex items-center gap-2 min-h-14 border-b border-a8c-gray-0 px-8"
 						>
-							<div className="flex items-left min-w-20 me-6 shrink-0">
+							<div className="flex items-left min-w-[100px] me-4 shrink-0">
```
fredrikekelund, line 216:
```suggestion
							<div className="flex items-left min-w-28 me-4 shrink-0">
```
I think we should try to stick to Tailwind presets when possible. In this case, this suggestion seems to work fine.

src/components/sync-connected-sites.tsx
```diff
@@ -222,7 +222,7 @@ const SyncConnectedSitesSection = ( {
 							</div>
 							<Button
 								variant="link"
-								className="!text-a8c-gray-70 hover:!text-a8c-blueberry max-w-[100%]"
+								className="!text-a8c-gray-70 hover:!text-a8c-blueberry max-w-full overflow-hidden"
```
fredrikekelund, line 225: 👍

fredrikekelund (approved):
Works fine, even if I think we should switch to a spacing preset instead of a fixed px value 👍 Will approve straight away to unblock

fredrikekelund:
I came to think that we should try a different type of layout for this… Experimenting with some stuff right now. I may open a separate PR shortly

ivan-ottinger:
Thanks for the review and feedback, Fredrik.

> I came to think that we should try a different type of layout for this… Experimenting with some stuff right now. I may open a separate PR shortly

Alright, I will leave this PR open for the time being.

fredrikekelund:
Cool, thanks! I opened https://github.com/Automattic/studio/pull/1063 for using a grid layout instead. The benefit being that the width of the leftmost column is now controlled dynamically by the length of the translation.

ivan-ottinger:
> Cool, thanks! I opened #1063 for using a grid layout instead. The benefit being that the width of the leftmost column is now controlled dynamically by the length of the translation.

Nice. Yeah that was a drawback of this solution so I was setting it after testing with all locales. Btw, I tried grid solution when working on the PR, but then dropped it as I saw we have been using flex more.

Once your PR is merged, I will close this one. 🙂

wojtekn (approved):
Works well, approving, but agreed we can try with grid layout if possible.

fredrikekelund:
Thanks, @ivan-ottinger! I've merged https://github.com/Automattic/studio/pull/1063, so closing this one now

PR #1059, opened 2025-03-12
Add tooltip for version mismatch on preview sites

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10543

## Proposed Changes

- Added version mismatch detection for PHP and WordPress versions in the preview site creation flows
- Added a tooltip message that informs users when their site's versions differ from the preview site's supported versions and which versions will be installed

image

## Testing Instructions

1. Test with different PHP versions:
   - Create a site with PHP 7.4 (unsupported)
   - Create a site with PHP 8.2 (supported)
   - Verify the tooltip message appears only for unsupported versions

2. Test with different WordPress versions:
   - Create a site with WordPress 6.3 (older version)
   - Create a site with WordPress 6.7.2 (latest at the time of this PR)
   - Create a site with WordPress 6.8-beta2 (beta version at the time of this PR)
   - Verify the tooltip message appears only for versions below the latest stable

3. Test combined version mismatches:
   - Create a site with both unsupported PHP and WordPress versions

## Pre-merge Checklist


bcotrim:
@Automattic/yolo I'm not sure about the best approach for the translations. At the moment I don't think the existing message will be clear to translate, but I also think the versions are an important context to provide to our users.
Maybe replacing the "and" with a space or a comma will make it easier, but I'll wait for your feedback on it.
Thank you! 🙇

src/modules/preview-site/components/create-preview-button.tsx
```diff
@@ -58,6 +76,23 @@ export function CreatePreviewButton( { onClick, selectedSite }: CreatePreviewBut
...
+		const versionChanges = [];
+		if ( ! isPhpVersionDefault ) {
+			versionChanges.push( sprintf( __( 'PHP %s' ), DEFAULT_PHP_VERSION ) );
+		}
+		if ( isWpVersionBelowDefault ) {
+			versionChanges.push( sprintf( __( 'WordPress %s' ), coercedLatestWpVersion ) );
+		}
+		versionMismatchMessages.push(
+			sprintf(
+				__( 'Your site is using an unsupported version. The preview site will use %s.' ),
```
katinthehatsite: What do you think about specifying e.g. unsupported PHP version or unsupported WP version in the message? 

Additionally, I am wondering if we could skip the exact output of the versions and say something more generic like `The preview site will use a newer version instead` as suggested in https://github.com/Automattic/dotcom-forge/issues/10543#issuecomment-2717041285

bcotrim: I started with that message initially, but this was the approach I found to have a clean way of displaying both PHP and WordPress in the same message.
I'm not sure if we can say newer version because, for PHP, that is incorrect. The preview site will use 8.2 even if the site is running a newer version.

What do you think about this more generic message:
`Your site is using unsupported versions and will use supported WordPress/PHP versions instead.`

wojtekn: What about `Your site is running versions not supported by preview sites and will automatically switch to supported WordPress and PHP versions.`?

bcotrim: I have updated the tooltip message to `Your site is running versions not supported by preview sites and will automatically switch to supported WordPress and PHP versions.`
Thank you both for the feedback

src/modules/preview-site/components/create-preview-button.tsx
```diff
@@ -57,6 +75,9 @@ export function CreatePreviewButton( { onClick, selectedSite }: CreatePreviewBut
 		),
 		DEMO_SITE_SIZE_LIMIT_GB
 	);
+	const versionMismatchMessage = __(
+		'Your site is running versions not supported by preview sites and will automatically switch to supported WordPress and PHP versions.'
```
katinthehatsite:
```suggestion
		'Your Studio site is running versions not supported by preview sites. The preview site will automatically switch to the supported WordPress and PHP versions.'
```

katinthehatsite: I added this small suggestion as I think when you read the tooltip, it sounds a bit like the Studio site will switch to supported versions and not the preview site.

katinthehatsite (approved):
The message looks good to me 👍 Thanks for all the adjustments. 

Not sure if we want to have another opinion though. Maybe @wojtekn wants to chime in

src/modules/preview-site/components/create-preview-button.tsx
```diff
@@ -24,10 +29,23 @@ export function CreatePreviewButton( { onClick, selectedSite }: CreatePreviewBut
...
 
 	const isCurrentSiteArchiving = archivingSiteId === selectedSite.id;
 	const isOtherSiteArchiving = isAnySiteArchiving && ! isCurrentSiteArchiving;
 
+	// Jurrasic Ninja will use the default version for PHP and latest or above for WordPress.
+	const latestWpVersion = wpVersions.find( ( version ) => version.isBeta === false )?.value;
+	const isPhpVersionDefault = selectedSite.phpVersion === DEFAULT_PHP_VERSION;
+	const coercedWpVersion = semver.coerce( wpVersion );
+	const coercedLatestWpVersion = semver.coerce( latestWpVersion );
+	const isWpVersionBelowDefault =
```
wojtekn: Could we extract the logic around `isPhpVersionDefault` and `isWpVersionBelowDefault` to a separate function, e.g., `shouldShowTooltip`, and add some tests covering possible cases?

src/modules/preview-site/components/create-preview-button.tsx
```diff
@@ -74,6 +95,8 @@ export function CreatePreviewButton( { onClick, selectedSite }: CreatePreviewBut
 		tooltipContent = { text: errorMessages.rest_site_creation_blocked };
 	} else if ( isOverLimit ) {
 		tooltipContent = { text: overLimitMessage };
+	} else if ( ! isPhpVersionDefault || isWpVersionBelowDefault ) {
```
nightnei: Not critical, but the word "Below" looks weird a bit 😅
In the context of code you most likelly wanted to menthion the below in the 44-47 lines, but I think it's redundant and a bit confusing.
```suggestion
	} else if ( ! isPhpVersionDefault || isWpVersionDefault ) {
```

nightnei: Hm, also, is it correct condition `isWpVersionBelowDefault`? Shouldn't it be `! isWpVersionBelowDefault`?

src/modules/preview-site/components/create-preview-button.tsx
```diff
@@ -24,10 +29,23 @@ export function CreatePreviewButton( { onClick, selectedSite }: CreatePreviewBut
...
+	// Jurrasic Ninja will use the default version for PHP and latest or above for WordPress.
+	const latestWpVersion = wpVersions.find( ( version ) => version.isBeta === false )?.value;
+	const isPhpVersionDefault = selectedSite.phpVersion === DEFAULT_PHP_VERSION;
+	const coercedWpVersion = semver.coerce( wpVersion );
+	const coercedLatestWpVersion = semver.coerce( latestWpVersion );
+	const isWpVersionBelowDefault =
+		latestWpVersion &&
+		coercedWpVersion &&
+		coercedLatestWpVersion &&
+		semver.compare( coercedWpVersion, coercedLatestWpVersion ) < 0;
```
nightnei: Not critical, but I think we can move it to separate variable, to isolate all this condition behind 1 variable
```
const isLatestCoercedWpVersion = coercedWpVersion &&
		coercedLatestWpVersion &&
		semver.compare( coercedWpVersion, coercedLatestWpVersion ) < 0
```

nightnei: Also, I am not in the context enough, so I supposed that it can be confusing for other developers to understand at a glance - what is the difference between checking `latestWpVersion` vs `semver.coerce`. Maybe adding a comment would be useful and simplify understand the code.

nightnei (changes requested):
In general LGTM but left a few comments for discussion

src/modules/preview-site/components/create-preview-button.tsx
nightnei: Yeah, good points 👍

Also, in this context, let's consider these two my comments [1](https://github.com/Automattic/studio/pull/1059#discussion_r1993671061) and [2](https://github.com/Automattic/studio/pull/1059#discussion_r1993674275).

bcotrim:
Thanks for the feedback @wojtekn @nightnei 
I did the following changes:
- New function to handle version comparison
- Unit tests for that function
- Improve variable names and added comments for better code readability

Please let me know what you think.

src/modules/preview-site/components/create-preview-button.tsx
```diff
@@ -24,10 +28,21 @@ export function CreatePreviewButton( { onClick, selectedSite }: CreatePreviewBut
...
 	const isOffline = useOffline();
 	const errorMessages = useArchiveErrorMessages();
+	const [ wpVersion ] = useGetWpVersion( selectedSite );
+	const wpVersions = useRootSelector( wordpressVersionsSelectors.selectWordPressVersions );
 
 	const isCurrentSiteArchiving = archivingSiteId === selectedSite.id;
 	const isOtherSiteArchiving = isAnySiteArchiving && ! isCurrentSiteArchiving;
 
+	const latestWpVersion = wpVersions.find( ( version ) => version.isBeta === false )?.value;
+	const { phpVersionMismatch, wpVersionMismatch } = compareVersions( {
```
wojtekn: Any reason for `compareVersions` not being `hasVersionMismatch`?

wojtekn (approved):
@bcotrim looks good. Let's use rename `utils` to `lib` as this is the directory we should use for such libraries. See [Project Structure](https://github.com/Automattic/studio/blob/trunk/docs/code-contributions.md#project-structure) docs.

nightnei (approved):
Thanks for addressing feedback, LGTM 👍

PR #1060, opened 2025-03-12
Studio: Update wording when PHP or WP versions are changed

katinthehatsite:
## Related issues

Closes: https://github.com/Automattic/dotcom-forge/issues/10602

## Proposed Changes

I propose to update the wording to `Saving and restarting...` (if the site was running) for when we change WP or PHP versions:

Screenshot 2025-03-14 at 2 13 23 PM

If the site was not running, the wording will remain as `Saving...`.
## Testing Instructions

* Ensure you have at least one Studio site created
* Navigate to the `Settings` tab
* Change WP or both WP and PHP versions
* Observe the loading indicator that says `Updating versions...`

## Pre-merge Checklist


bcotrim (commented):
It's minor, but you can notice the text of the button changing, even if we didn't change any of the versions.
I think we can either change the `setIsEditingSite( true )` only when `hasPhpVersionChanged` or `hasWpVersionChanged`, or create a new state to control the label based on those conditions.

https://github.com/user-attachments/assets/9e1dafbb-8998-4d66-a073-a020d0f42fe8

wojtekn:
'Restarting server' seemed to be better to me at first, but this label doesn't make sense when a user updates the WP version for a site that is not running.

Thus, 'Updating versions' seems correct.

@matt-west any thoughts on that?

matt-west:
Do we need to keep this modal open while we're restarting the server?

The start/stop button in the header and the icon alongside the site in the sidebar both indicate that the site is restarting.

Keeping the modal open while we wait for the site to restart makes the app feel slow. If we just close the modal and rely on the indicators elsewhere in the app to communicate that the site is restarting we give the perception that Studio is a lot faster.

This would also mean we don’t have to worry about updating this button text.

katinthehatsite:
Thanks for the suggestion @matt-west  - let me experiment a bit to see what it could look like.

katinthehatsite:
I tested a bit more and I think we should stick with the loading state in the modal. My reasoning for this is:

* If the user decides to re-open the modal again, it would need a loading state regardless and blocking them from re-opening the modal does not make sense.
* I did not find the UI indicative enough to know what exactly was happening. It was taking a while and it felt like nothing was happening. I think having loading state in the modal was more responsive for the user as they are provided clear feedback about what is happening.

katinthehatsite:
Thanks for testing @bcotrim - I tried to reproduce on my end but I can not seem to catch the glitch. I added the video below. What am I doing differently from your steps? (tried with both started and stopped sites)

https://github.com/user-attachments/assets/83896a6c-bc29-4bb2-8e7e-ca41cf9716c3

Actually never mind, I was not clicking `Save`, I see the issue now 👀

katinthehatsite:
@matt-west @wojtekn @bcotrim I am thinking that perhaps we should go back to something like `Updating` or `Saving` instead of having `Updating versions`. Essentially, the user can update:

* only site name
* only PHP or WP versions
* any combination of these

So we would either need to come up with different combinations e.g. what if they update site name and PHP version - then, `updating versions` will not work. Perhaps we should go for a more general word that covers every case.

bcotrim:
@katinthehatsite I like `Updating`, in my opinion it fits all the fields the user can change in that layout.

wojtekn:
What if we go simply for now, leave the modal, and say, either 'Saving' or 'Saving and restarting ...' depending on the action user took?

matt-west:
> I did not find the UI indicative enough to know what exactly was happening. It was taking a while and it felt like nothing was happening. I think having loading state in the modal was more responsive for the user as they are provided clear feedback about what is happening.

That's interesting @katinthehatsite. Do you know why the button doesn’t say "Starting..." even though we're restarting the server?

I would expect that their state is updated even though the modal is open.

Screenshot 2025-03-13 at 14 49 49

nightnei:
> If the user decides to re-open the modal again, it would need a loading state regardless and blocking them from re-opening the modal does not make sense.
> I did not find the UI indicative enough to know what exactly was happening. It was taking a while and it felt like nothing was happening. I think having loading state in the modal was more responsive for the user as they are provided clear feedback about what is happening.

@katinthehatsite @matt-west  Wouldn't be better to proceed with:
1. As recommended Matt - close the popup immediately
2. Disable "Edit" button while server is starting (theoretically it should be always disabled when we are starting the server)
3. We can add tooltip with copy e.g. "The server is starting..."

But I suppose that it can take much more effort which is not enough reasonable ATM. So I would go with @wojtekn 's suggestion: "What if we go simply for now, leave the modal, and say, either 'Saving' or 'Saving and restarting ...' depending on the action user took?"

katinthehatsite:
> That's interesting @katinthehatsite. Do you know why the button doesn’t say "Starting..." even though we're restarting the server?

I think it happens because we need to download WP version first if it is not installed before we can start the site. When changing PHP versions, it won't be the case because these are bundled. We can probably figure something out around it but regardless, I am finding that we could have a bit more prominent loading in the UI if we are to ditch the loading state in the modal.

I will give it a try though to see how it looks.

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -216,7 +216,7 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
 										! siteName.trim()
 								) }
 							>
-								{ isEditingSite ? __( 'Saving…' ) : __( 'Save' ) }
+								{ isEditingSite ? __( 'Updating…' ) : __( 'Save' ) }
 							</Button>
```
katinthehatsite, line 232: I ended up going with a simple `Updating` to not overcomplicate things.

wojtekn: I don't think changing that to 'Updating' fixes anything over 'Saving'.

If we go the way with short label, the CTA should be consistent with the text, e.g.,
- Create -> Creating
- Add -> Adding
- Save -> Saving
- Update -> Updating

Now we have 'Save' that changes to 'Updating'. 

Can we try with 'Saving and restarting'? The user knows what they are saving, so the first part of the label would just confirm that Studio is saving the form. The second part would inform it's restarting, explaining why it takes some time.

katinthehatsite: > Saving and restarting

That is an option, perhaps we could make a difference in wording for cases when the server needed to be restarted and not needed to be restarted.

wojtekn: @katinthehatsite yes, this is what I proposed in one of previous comments: https://github.com/Automattic/studio/pull/1060#issuecomment-2721493624

katinthehatsite: Oh I see, I read that message as we can pick either this or that and not as either this or that depending on the situation. That makes more sense now 👍 Apologies for misunderstanding.

katinthehatsite: Made updates in https://github.com/Automattic/studio/pull/1060/commits/fff3c88da54af9a5fe0de326fbafdd63bf920ce5

wojtekn: No worries, thanks for updating that.

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -73,12 +74,13 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
...
-		try {
-			const running = selectedSite.running;
 
-			const hasWpVersionChanged = selectedWpVersion !== currentWpVersion;
-			const hasPhpVersionChanged = selectedPhpVersion !== selectedSite.phpVersion;
-			const needsRestart = running && ( hasWpVersionChanged || hasPhpVersionChanged );
+		const hasWpVersionChanged = selectedWpVersion !== currentWpVersion;
+		const hasPhpVersionChanged = selectedPhpVersion !== selectedSite.phpVersion;
+		const needsRestart = selectedSite.running && ( hasWpVersionChanged || hasPhpVersionChanged );
+		setNeedsRestart( needsRestart );
```
nightnei, line 81: Nit, but I would propose this way, as, IMO, more simpler to read and understand at a glance
Also we don't trigger setting state every time, even when it's definitely not necessary
```suggestion
		if ( selectedSite.running ) {
			const hasWpVersionChanged = selectedWpVersion !== currentWpVersion;
			const hasPhpVersionChanged = selectedPhpVersion !== selectedSite.phpVersion;
			
			setNeedsRestart( hasWpVersionChanged || hasPhpVersionChanged );
		}
```

nightnei (approved):
LGTM but left one comment, take a look please

wojtekn (approved):
It works well now.

There is still a case in which it takes some time and Studio only shows "Saving" - when user changes WordPress version for site that is not running. I guess we can live with that for now, as displaying 'Saving and restarting' in such a case could be confusing.

katinthehatsite:
> There is still a case in which it takes some time and Studio only shows "Saving" - when user changes WordPress version for site that is not running. I guess we can live with that for now, as displaying 'Saving and restarting' in such a case could be confusing.

Yeah, I think we can wait for some feedback from users to see if someone mentions that the process looks slow and then re-iterate. I briefly played with the code to see if we can speed up the downloading of WP version by using the a similar approach to site creation instead of executing CLI but it was more messy than what we have now. I think we can wait and iterate either from technical perspective or from the UI.

PR #1062, opened 2025-03-12
Improve formatting of the latest version and add it to edit site form

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10601

## Proposed Changes

- Removes hardcoded "Latest" WordPress version option
- Implements dynamic "Latest" version based on most recent non-beta version
- Updates WordPress versions selector to use the new dynamic latest version approach

Example on Add Site form:
image

## Testing Instructions

- For Onboarding, Add Site and Edit Site workflows check that the wordpress version selector has the options according to the changes mentioned above.

## Pre-merge Checklist

### DIFF
--- a/src/stores/wordpress-versions-slice.ts
+++ b/src/stores/wordpress-versions-slice.ts
@@ -111,14 +111,21 @@ const wordpressVersionsSlice = createSlice( {
 	},
 	selectors: {
 		selectWordPressVersions: ( state ) => state.versions,
-		selectWordPressVersionsWithLatest: ( state ) => [
-			{
-				isBeta: false,
-				label: __( 'Latest' ),
-				value: 'latest',
-			},
-			...state.versions,
-		],
+		selectWordPressVersionsWithLatest: ( state ) => {
+			const latestNonBeta = state.versions.find( ( version ) => ! version.isBeta );
+			const otherVersions = state.versions.filter( ( version ) => version !== latestNonBeta );
+			if ( ! latestNonBeta ) {
+				return otherVersions;
+			}
+			return [
+				{
+					isBeta: false,
+					label: `${ latestNonBeta.label } (${ __( 'latest' ) })`,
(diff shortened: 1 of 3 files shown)


bcotrim:
Added PR as draft for now until we have feedback on the design (see https://github.com/Automattic/dotcom-forge/issues/10601)

src/stores/wordpress-versions-slice.ts
```diff
@@ -111,14 +111,21 @@ const wordpressVersionsSlice = createSlice( {
...
-			{
-				isBeta: false,
-				label: __( 'Latest' ),
-				value: 'latest',
-			},
-			...state.versions,
-		],
+		selectWordPressVersionsWithLatest: ( state ) => {
+			const latestNonBeta = state.versions.find( ( version ) => ! version.isBeta );
+			const otherVersions = state.versions.filter( ( version ) => version !== latestNonBeta );
```
katinthehatsite: Perhaps we could adjust the code a bit to pass through the versions once instead of iterating through the array twice? What do you think?

katinthehatsite (approved):
I have one suggestion about possible code improvement, otherwise the changes look good to me 👍

src/stores/wordpress-versions-slice.ts
bcotrim: We are not iterating the array twice. 
`find()` stops as soon as it finds a value that matches the conditions.

We could use `reduce()` to make everything in the same iteration, but given the expected number of versions the performance gains vs code readability is not worth it, in my opinion.

What do you think?

bcotrim: We could move the condition for `! latestNonBeta` to before the `filter()` to avoid that iteration if no version is found.

katinthehatsite: Alright, let's keep as is 👍  I think we are good to merge

PR #1064, opened 2025-03-13
Custom Domain: Allow updating domains for existing sites.

### LINKED ISSUE #1053: Feature Request: Allow modifying the custom domain of a given site
### What

We recently introduce the possibility to assign a custom local domain to a given site created in Studio.
That said, it is not possible right now to edit the domain for an existing site.

In the same way, we're introducing SSL support #1034 on site creation but not edition.

### Why

A small technical limitation is preventing that, it seems common sense to remove this limitation. 

### How

If we want to allow editing these features, we would have to trigger a configuration change in WordPress and update the database maybe by running a WP CLI command

```
wp search-replace 'http://example.local' 'https://example2.local'
```

I believe this logic might already be present in Studio for push/pull support, so let's try to use the same logic/code.

youknowriad:
## Related issues

- Fixes #1053 
- Fixes STU-72-linear-issue

## Summary

  - Adds ability to add, modify, or remove custom domains on existing sites
  - Ensures consistent domain validation and handling across site creation and editing
  - Ensures we update the hosts file and the database content when the domain is updated.

## Testing instructions

 - Create a site (once with a custom domain, and once without)
 - Try updating domains for running and stopped sites


youknowriad:
Can I get some reviews here?

src/lib/hosts-file.ts
```diff
@@ -127,6 +127,48 @@ export const removeDomainFromHosts = async ( domain: string ): Promise< void > =
 	}
 };
 
+export const replaceDomainInHosts = async (
```
bcotrim: I would rename this function as it handles all possible domain change scenarios.

And to keep the format consistent, move the replacing logic to its own `replaceDomainInHosts`

src/lib/updateSiteUrlToLocal.ts
```diff
@@ -12,12 +12,15 @@ export const updateSiteUrlToLocal = async ( siteId: string ) => {
 	} );
 
 	if ( ! currentSiteUrl ) {
-		console.error( 'Failed to fetch site URL after import' );
+		console.error( 'Failed to fetch site URL' );
```
bcotrim, line 15: Could we rename this file to follow kebab-case format as the other files in the project?

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -45,6 +46,11 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
 		( selectedSite?.phpVersion as AllowedPHPVersion ) ?? DEFAULT_PHP_VERSION
 	);
 	const [ selectedWpVersion, setSelectedWpVersion ] = useState( currentWpVersion );
+	const [ useCustomDomain, setUseCustomDomain ] = useState( Boolean( selectedSite?.customDomain ) );
```
bcotrim, line 49: It seems a lot of this logic is very similar to the one we added to `site-form`
I believe there's an opportunity here to create a shared component/hook, to avoid repetition.

bcotrim (commented):
Great feature!
Added some comments, but overall I think this looks great!

Something I noticed during my tests:
- Having an invalid domain disables the submit, even when "Use custom domain" isn't checked. I recorded on "Edit Site" but the same behavior happens on "Add Site":


https://github.com/user-attachments/assets/6970cb47-2f42-4374-898c-8ba299d2705c

src/modules/site-settings/edit-site-details.tsx
youknowriad: Entirely agree but this is not new. The way the form works today is not great and would deserve a refactor but it's not a small change.

src/lib/hosts-file.ts
youknowriad: Not sure I understand this comment fully but I renamed the function.

youknowriad:
I fixed the validation bug.

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -214,8 +272,11 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
 										! selectedSite ||
 										( selectedSite?.name === siteName &&
 											selectedSite?.phpVersion === selectedPhpVersion &&
-											currentWpVersion === selectedWpVersion ) ||
-										! siteName.trim()
+											currentWpVersion === selectedWpVersion &&
```
bcotrim: The disabled state logic is a bit long and could be made more readable by extracting it into meaningful variables:

```typescript
const isFormUnchanged = selectedSite && 
    selectedSite.name === siteName &&
    selectedSite.phpVersion === selectedPhpVersion &&
    currentWpVersion === selectedWpVersion &&
    Boolean(selectedSite.customDomain) === useCustomDomain &&
    (!useCustomDomain || selectedSite.customDomain === customDomain);

const hasValidationErrors = !selectedSite || 
    !siteName.trim() || 
    (useCustomDomain && customDomainError);

disabled={isEditingSite || isFormUnchanged || hasValidationErrors}
```

This would make the code more self-documenting and easier to maintain. What do you think?

bcotrim (approved):
Thanks for making the changes.
I added a small comment, but it shouldn't be blocking.

I noticed the custom domain error message get cut when we display multiple errors.
This should be fixed with [this change](https://github.com/Automattic/studio/pull/1078/files#diff-e2be6df548f75063abd8db0f241c35626f7af6a1c6a10c429935de5f902dbbc6R373) but something to re-test later.

image

src/modules/site-settings/edit-site-details.tsx
youknowriad: Sounds good to me :)

wojtekn:
> Having an invalid domain disables the submit, even when "Use custom domain" isn't checked. I recorded on "Edit Site" but the same behavior happens on "Add Site":

@bcotrim I tested it when I rebased https://github.com/Automattic/studio/pull/1078 and noticed it is fixed in Add Site, but it still happens in Edit Site. Are you also reproduce it there?

bcotrim:
> > Having an invalid domain disables the submit, even when "Use custom domain" isn't checked. I recorded on "Edit Site" but the same behavior happens on "Add Site":
> 
> @bcotrim I tested it when I rebased #1078 and noticed it is fixed in Add Site, but it still happens in Edit Site. Are you also reproduce it there?

I don't see it happening on "Add Site" or "Edit Site".
Note that if you don't have any change compared the existing settings the button is still disabled due to the form being unchanged, not the validation error.


https://github.com/user-attachments/assets/39ed88de-b887-49f6-93da-46e01d47e3c9

wojtekn:
> Note that if you don't have any change compared the existing settings the button is still disabled due to the form being unchanged, not the validation error.

Oh, it was that. Thanks for clarifying. When I edit the site name, then enable a custom domain and enter an invalid one, Save button toggles correctly when I keep disabling and enabling a custom domain.

PR #1065, opened 2025-03-13
Replace deprecated sudo-prompt with @vscode/sudo-prompt

mcsf:
## My Summary
* `sudo-prompt` is deprecated, as initially mentioned in #994 
* This week I am starting on a new laptop, and now Studio simply fails due to a surprising error in the `sudo.exec` call:
```
Command failed: ./applet
/bin/sh: ./applet: Bad CPU type in executable
```
* My suspicion is that the binary embedded in the deprecated module was still compatible with an Apple M1 Pro chip but not with an M4 Pro. But I haven't spent more than a couple of minutes on this. Switching packages fixes the problem.

## Claude's Summary
* Replace the deprecated `sudo-prompt` package with its maintained fork `@vscode/sudo-prompt`
* Update all imports and test mocks to use the new package
* The APIs are identical, so this is a straightforward replacement
* No functional changes required

## Test plan
* Ran the hosts-file unit tests to confirm the replacement works correctly
* Verified the replacement package has the same API and functionality

🤖 Generated with [Claude Code](https://claude.ai/code)


youknowriad:
I have an M2 and I can't reproduce this issue in trunk. But this PR works for me and is the right thing to do anyway (move from the deprecated package)

youknowriad (approved):

PR #1066, opened 2025-03-13
Adjust sorting of WordPress versions dropdown

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10621
- Fixes https://github.com/Automattic/dotcom-forge/issues/10601

## Proposed Changes

- Adjust the order or the wordpress versions, to preserve the semver order, instead of listing `latest` on top
- Ensure the latest version is selected by default on add site modal
- Improve version ordering logic when handling custom/non-listed WordPress versions
  - Use `semver.coerce()` to properly compare versions that aren't in the official version list
  - Ensure custom versions are inserted in the correct position based on semver comparison

| Onboarding - Default | Onboarding - Dropdown | Add Site - Default | Add Site - Dropdown | Edit Site - Default | Edit Site - Dropdown |
|--------|--------|--------|--------|--------|--------|
| image | image | image | image | image | image | 

## Testing Instructions

- Start the onboarding flow (delete all sites)
- Confirm the default WordPress version selected value is the latest version
- Confirm the order from the WordPress version dropdown is correct
- Adding a site should result in a site with the selected WordPress version
- Try for multiple versions and without selecting a version from the dropdown to ensure the values are correctly set
- Repeat the process for Add Site workflow
- Check Edit site in Settings
- Confirm the site WordPress version is the version selected in the Edit site modal
- Confirm the order of the WordPress version dropdown

Additional Testing for Custom Versions:
- Install a WordPress version that isn't in the default version list (e.g. 5.8)
- Verify that the custom version appears in the correct position in the version dropdown based on semver ordering
- Test with both older and newer versions to ensure proper ordering in all cases

## Pre-merge Checklist

### DIFF
--- a/src/components/onboarding.tsx
+++ b/src/components/onboarding.tsx
@@ -87,16 +87,15 @@ export default function Onboarding() {
 		},
 	} );
 
-	const wpVersions = useRootSelector(
-		wordpressVersionsSelectors.selectWordPressVersionsWithLatest
+	const latestStableVersion = useRootSelector(
+		wordpressVersionsSelectors.selectLatestStableVersion
 	);
 
 	useEffect( () => {
-		const latestVersion = wpVersions.find( ( version ) => ! version.isBeta );
-		if ( latestVersion ) {
-			setWpVersion( latestVersion.value );
+		if ( latestStableVersion ) {
+			setWpVersion( latestStableVersion.value );
 		}
-	}, [ wpVersions, setWpVersion ] );
+	}, [ latestStableVersion, setWpVersion ] );
 
 	useEffect( () => {
 		const run = async () => {
--- a/src/components/onboarding.tsx
+++ b/src/components/onboarding.tsx
@@ -11,6 +11,8 @@ import { useAddSite } from 'src/hooks/use-add-site';
 import { useDragAndDropFile } from 'src/hooks/use-drag-and-drop-file';
 import { generateSiteName } from 'src/lib/generate-site-name';
 import { getIpcApi } from 'src/lib/get-ipc-api';
+import { useRootSelector } from 'src/stores';
+import { wordpressVersionsSelectors } from 'src/stores/wordpress-versions-slice';
 
 const GradientBox = () => {
 	const { __ } = useI18n();
@@ -84,6 +86,17 @@ export default function Onboarding() {
 			}
 		},
 	} );
+
+	const wpVersions = useRootSelector(
+		wordpressVersionsSelectors.selectWordPressVersionsWithLatest
+	);
+
+	useEffect( () => {
+		const latestVersion = wpVersions.find( ( version ) => ! version.isBeta );
+		if ( latestVersion ) {
+			setWpVersion( latestVersion.value );
+		}
+	}, [ wpVersions, setWpVersion ] );
 
 	useEffect( () => {
 		const run = async () => {
--- a/src/modules/site-settings/edit-site-details.tsx
+++ b/src/modules/site-settings/edit-site-details.tsx
@@ -1,6 +1,7 @@
 import { SelectControl } from '@wordpress/components';
 import { useI18n } from '@wordpress/react-i18n';
 import { FormEvent, useCallback, useState } from 'react';
+import semver from 'semver';
 import stripAnsi from 'strip-ansi';
 import Button from 'src/components/button';
 import { ErrorInformation } from 'src/components/error-information';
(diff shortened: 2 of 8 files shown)


src/stores/wordpress-versions-slice.ts
```diff
@@ -112,19 +112,17 @@ const wordpressVersionsSlice = createSlice( {
...
-			const otherVersions = state.versions.filter( ( version ) => version !== latestNonBeta );
-			return [
-				{
-					isBeta: false,
-					label: `${ latestNonBeta.label } (${ __( 'latest' ) })`,
-					value: latestNonBeta.value,
-				},
-				...otherVersions,
-			];
+			let foundLatest = false;
```
fredrikekelund:
```suggestion
			let foundLatestStable = false;
```
Nit, but this would clarify what we're looking for.

src/components/onboarding.tsx
```diff
@@ -85,6 +87,17 @@ export default function Onboarding() {
...
+	const wpVersions = useRootSelector(
+		wordpressVersionsSelectors.selectWordPressVersionsWithLatest
+	);
+
+	useEffect( () => {
+		const latestVersion = wpVersions.find( ( version ) => ! version.isBeta );
+		if ( latestVersion ) {
+			setWpVersion( latestVersion.value );
+		}
+	}, [ wpVersions, setWpVersion ] );
```
fredrikekelund: Given that we're repeating this logic, I would consider adding a `selectLatestStableWordPressVersion` selector to `src/stores/wordpress-versions-slice.ts`.

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -52,8 +53,25 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
 		label: version.label,
 		value: version.value,
 	} ) );
+
 	if ( ! wordpressVersionOptions.some( ( version ) => version.value === currentWpVersion ) ) {
-		wordpressVersionOptions.push( { label: currentWpVersion, value: currentWpVersion } );
+		const customVersion = { label: currentWpVersion, value: currentWpVersion };
```
fredrikekelund: Total nit, but I would move this below the `findIndex` call

katinthehatsite:
In terms of functionality, everything looks good for me ❤️ I am taking a look at the code now

src/components/onboarding.tsx
katinthehatsite: Nice, I agree. If we are repeating some code, let's make sure to unify it

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -52,8 +53,25 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
...
+		// find the index of the version that is less than the current version
+		const insertIndex = wordpressVersionOptions.findIndex( ( compareVersion ) => {
+			const currentVer = semver.coerce( currentWpVersion );
+			const compareVer = semver.coerce( compareVersion.value );
+			if ( ! currentVer || ! compareVer ) {
+				return false;
+			}
+			return semver.lt( compareVer, currentVer );
+		} );
+		// if the current version is less than all the versions in the list, add it to the end
```
katinthehatsite: We can probably clean those up. Although, not a strong preference considering that there are a lof of things happening

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -52,8 +53,25 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
 		label: version.label,
 		value: version.value,
 	} ) );
+
 	if ( ! wordpressVersionOptions.some( ( version ) => version.value === currentWpVersion ) ) {
-		wordpressVersionOptions.push( { label: currentWpVersion, value: currentWpVersion } );
+		const customVersion = { label: currentWpVersion, value: currentWpVersion };
+		// find the index of the version that is less than the current version
+		const insertIndex = wordpressVersionOptions.findIndex( ( compareVersion ) => {
```
nightnei: What about this, to avoid comment and make variable more obvious to "what it is", i stead of "what it should do":
```suggestion
		const indexOfPrevVersion = wordpressVersionOptions.findIndex( ( compareVersion ) => {
```

src/modules/site-settings/edit-site-details.tsx
nightnei: +1
Together with removing this `// otherwise, insert it at the found index`.
And I think that my suggestion to rename `insertIndex ` to `indexOfPrevVersion` will make removing comments more reasonable.

src/modules/site-settings/edit-site-details.tsx
nightnei: Also, let's extract this insertion to a separate function. Ar a result the component will be smaller and easy to read.

```
if ( ! wordpressVersionOptions.some( ( version ) => version.value === currentWpVersion ) ) {
	addWpVersionToList( currentWpVersion, wordpressVersionOptions );
}
```

nightnei (changes requested):

bcotrim:
Thank you all for the feedback, while moving the function to lib folder and doing the tests, I noticed we weren't handling the case for a version newer than all the versions. Probably an edge case, but I think we shouldn't assume that.
I tried to be mindful of variables names and comments, but please let me know if you have any suggestions or notice any issues.

katinthehatsite (approved):
The changes look good to me 👍 thanks for incorporating the proposed adjustments!

src/modules/site-settings/lib/wordpress-versions.ts
```diff
@@ -0,0 +1,30 @@
...
+	const currentVer = semver.coerce( version );
+
+	// Being extra cautious here, if the version is not valid, we add it to the end of the list.
+	if ( ! currentVer ) {
+		options.push( customVersion );
+		return;
+	}
+
+	// Find the index to insert the new version according to the semver version, newest first.
+	const indexToInsert = options.findIndex( ( compareVersion ) => {
```
nightnei: WDYT about this my comment?
https://github.com/Automattic/studio/pull/1066#discussion_r1995834327

nightnei (approved):
LGTM 👍
Left 1 last comment

ivan-ottinger (approved):
The changes look good and test well. 👌🏼 Including when experimenting with WP version that is not on the default version list (tried with `5.7.12`).

PR #1068, opened 2025-03-14
Improve error message when port 80 is already in use

ivan-ottinger:
## Related issues

- Resolves https://github.com/Automattic/dotcom-forge/issues/10583. 

## Proposed Changes

- delay error message and update its copy

## Testing Instructions

1. Start a WordPress site with custom domain through the Local app: https://localwp.com/
2. Check out the PR branch and build the app with `npm start`. 
3. The error message should load with updated copy only after complete Studio app interface load.

## Pre-merge Checklist

### DIFF
--- a/src/lib/proxy-server.ts
+++ b/src/lib/proxy-server.ts
@@ -4,6 +4,7 @@ import { domainToASCII } from 'node:url';
 import * as Sentry from '@sentry/electron/main';
 import { __ } from '@wordpress/i18n';
 import httpProxy from 'http-proxy';
+import { getMainWindow } from 'src/main-window';
 import { SiteServer } from 'src/site-server';
 import { loadUserData } from 'src/storage/user-data';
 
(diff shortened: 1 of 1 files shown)


src/lib/proxy-server.ts
```diff
@@ -97,14 +97,16 @@ export async function startProxyServer(): Promise< boolean > {
...
+			setTimeout( () => {
+				dialog.showMessageBox( {
+					type: 'error',
+					message: __( 'Custom domain set up failed' ),
+					detail: __(
+						'Studio needs to use port 80, but it’s in use by another app. Close any local development apps and restart Studio.'
+					),
+					buttons: [ __( 'OK' ) ],
+				} );
+			}, 2000 );
```
```suggestion
			const mainWindow = await getMainWindow();
			dialog.showMessageBox( mainWindow, {
				type: 'error',
				message: __( 'Custom domain set up failed' ),
				detail: __(
					'Studio needs to use port 80, but it’s in use by another app. Close any local development apps and restart Studio.'
				),
				buttons: [ __( 'OK' ) ],
			} );
```
I realized that passing a window instance to `dialog.showMessageBox` unblocks the rendering. I think this approach is preferable since a timeout is inherently arbitrary.

src/lib/proxy-server.ts
```diff
@@ -97,14 +97,16 @@ export async function startProxyServer(): Promise< boolean > {
...
-					'Another server is already running on port 80. For custom domains to work, please stop that server and then restart Studio.'
-				),
-				buttons: [ __( 'OK' ) ],
-			} );
+			setTimeout( () => {
+				dialog.showMessageBox( {
+					type: 'error',
+					message: __( 'Custom domain set up failed' ),
+					detail: __(
+						'Studio needs to use port 80, but it’s in use by another app. Close any local development apps and restart Studio.'
```
```suggestion
						'Studio needs to use port 80, but it’s already in use by another app. Close any local development apps and restart Studio.'
```
Not a native English speaker, but this sounds better, IMO.

fredrikekelund (changes requested):
Copy changes look good, but I realized there's an approach that doesn't require us to use a timeout.

src/lib/proxy-server.ts
ivan-ottinger: That sounds good to me. 👍🏼

src/lib/proxy-server.ts
ivan-ottinger: That's very nice. Thanks for your review and feedback, Fredrik. I have applied your suggestion in https://github.com/Automattic/studio/pull/1068/commits/da29638d61ae362b2d4bcd4513c77ac5df61f120.

wojtekn (approved):

fredrikekelund (approved):
Nice 👍

PR #1072, opened 2025-03-14
Consider supporting trunk/nightly WordPress version

bcotrim:
## Related issues

- Fixes STU-78

## Proposed Changes

- Updated getWordPressVersionUrl to return `nightly` build URL for development versions
- Enhanced WordPress versions fetching to include both stable and development versions
- Added test coverage

| Add Site | Site Settings | Edit Site |
|--------|--------|--------|
| image | image | image | 

## Testing Instructions

- Using the "Add site" or "Onboarding" workflows, confirm that the wordpress version dropdown includes a `nightly` build.
- Start a new site with a `nightly` version and confirm it was installed successfully
- Go to the Settings tab and confirm the version is shown correctly (here it should display the full version, instead of `nightly`)
.- Open "Edit site"
- Confirm the `nightly` version is selected
- Try changing the version and confirm the installation is successful
- Add a new site with a non `nightly` version
- Go to the Settings tab and "Edit Site"
- Confirm `nightly` version option is available in the WordPress version dropdown
- Change the version to the `nightly` and save to confirm the update is successful

## Pre-merge Checklist


src/stores/wordpress-versions-slice.ts
```diff
@@ -25,40 +35,58 @@ export const fetchWordPressVersions = createAsyncThunk(
...
+			const allOffers = [ ...developmentOffers, ...stableOffers ];
+
+			return allOffers.map( ( { version, shortName } ) => {
 				const isBeta = version.includes( 'beta' ) || version.includes( 'RC' );
+				const isDevelopment = version.match( /^\d+\.\d+-[a-zA-Z0-9]+-\d+$/ ) !== null;
 				const occurrences = shortNameOccurrences.get( shortName ) || 0;
 				return {
 					isBeta,
-					label: occurrences > 1 || isBeta ? version : shortName,
+					isNightly: isDevelopment,
```
sejas: Why do we need this `isNightly` boolean?

sejas (commented):
Good first approach. I see the new `nightly` option in the dropdown.

Screenshot 2025-03-17 at 21 10 24

I see two possible options here. The premise is that users might want to update their sites to nightly version every day:
1. We could hardcode the nightly option in the dropdown, and display the user the numeric option. That way they could select nightly again.
2. We can fetch the development versions and match only the exact number version returned by the backend. That way users will see nightly the same they the site was created/updated , and they'll be able to update the following days.

src/stores/wordpress-versions-slice.ts
bcotrim: We probably don't, thanks for catching that.

bcotrim:
> I see two possible options here. The premise is that users might want to update their sites to nightly version every day:
  We could hardcode the nightly option in the dropdown, and display the user the numeric option. That way they could select nightly again.
   We can fetch the development versions and match only the exact number version returned by the backend. That way users will see nightly the same they the site was created/updated , and they'll be able to update the following days.

Thanks for the feedback @sejas 
I believe my approach is in line with your concern.
This is implementation I was going for:
- We fetch the developments versions and display `nightly` on the dropdown since it's easier to read and understand.
- Whenever a user installs a `nightly` version they will see the "real" version in the site settings
- If they edit the site details they will see `nightly` if their version is the latest `nightly` version and the "real" version number if there's a newer `nightly` build

Is this what you had in mind or did I misunderstood your initial approach?

sejas:
> If they edit the site details they will see nightly if their version is the latest nightly version and the "real" version number if there's a newer nightly build
> Is this what you had in mind or did I misunderstood your initial approach?

Yep, that's what I had in mind. 👍 
In the future we could consider enabling autoupdate to don't bother of changing the nightly version every day.

src/lib/version-utils.ts
```diff
@@ -7,3 +7,13 @@
 export function isDevRelease( version: string ): boolean {
 	return /-dev\..*|-dev\d+/.test( version );
 }
+
+/**
+ * Checks if a version string represents a WordPress development version
+ * Matches patterns like "6.8-beta2-59979"
```
sejas, line 13: Are we sure the nightly version follows this pattern even when there are no betas?
I think we shouldn't mix functions that check Studio versions and WordPress versions.

src/stores/wordpress-versions-slice.ts
```diff
@@ -16,52 +17,98 @@ const wordPressApiResponseSchema = z.object( {
...
+			const developmentOffers = processWordPressOffers(
+				developmentData.offers,
+				true,
+				shortNameOccurrences
+			).slice( 0, 1 );
+
+			const allOffers = [ ...developmentOffers, ...stableOffers ];
+
+			return allOffers.map( ( { version, shortName } ) => ( {
+				isBeta: version.includes( 'beta' ) || version.includes( 'RC' ),
```
sejas: We can consider re-using this boolean that is also used in `generateVersionLabel`.

src/stores/wordpress-versions-slice.ts
```diff
@@ -16,52 +17,98 @@ const wordPressApiResponseSchema = z.object( {
...
 	return match ? match[ 1 ] : version;
 };
 
+async function fetchWordPressApiData( channel: 'beta' | 'development', version?: string ) {
+	const baseUrl = 'https://api.wordpress.org/core/version-check/1.7/';
+	const params = new URLSearchParams( { channel } );
+
+	if ( version ) {
+		params.append( 'version', version );
+	}
```
sejas: The answer of the development API doesn't change even if we pass the version.

src/stores/wordpress-versions-slice.ts
```diff
@@ -16,52 +17,98 @@ const wordPressApiResponseSchema = z.object( {
...
-				};
-			} );
+			const stableOffers = processWordPressOffers( stableData.offers, false, shortNameOccurrences );
+			const developmentOffers = processWordPressOffers(
+				developmentData.offers,
+				true,
+				shortNameOccurrences
+			).slice( 0, 1 );
+
+			const allOffers = [ ...developmentOffers, ...stableOffers ];
```
sejas: `...developmentOffers` will be the nightly version or undefined in case the backend doesn't return a response `development`. If undefined, it won't appear in the array.

sejas (approved):
It works as expected. I tested the scenarios where we created a new site and update it using the nightly version.

**Creating nightly version**

https://github.com/user-attachments/assets/ebd4dcf7-5f57-4507-909d-6146c304bb32

**Creating beta and updating to nightly version**

https://github.com/user-attachments/assets/1efc2ec3-dffc-46af-b8e8-fbe4ffb9f15a


The nightly versions are saved in server files:

server-files

src/lib/get-wordpress-version-url.ts
```diff
@@ -1,7 +1,12 @@
+import { isWordPressDevVersion } from 'src/lib/version-utils';
 import { DEFAULT_WORDPRESS_VERSION } from 'vendor/wp-now/src/constants';
 import { isValidWordPressVersion } from 'vendor/wp-now/src/wp-playground-wordpress/is-valid-wordpress-version';
 
 export function getWordPressVersionUrl( version = DEFAULT_WORDPRESS_VERSION ) {
+	if ( isWordPressDevVersion( version ) ) {
+		return 'https://wordpress.org/nightly-builds/wordpress-latest.zip';
+	}
```
sejas: I'm a bit worried that when selecting a beta version we actually download a nightly version. This is not the case right now, but it could happen if the API returns a `6.8-beta3-600046`.

It would be better if we identify a `nightly` with that version name.

src/lib/version-utils.ts
bcotrim: Nice one!
I updated the check to allow for more scenarios, let me know what you think.
Also moved all WordPress version functions to a new lib file.

sejas: @bcotrim , I don't see this change. Did you already push the commit?
BTW, the image of Edit Site  in the description doesn't belong to the modal.

bcotrim: Sorry I sent the comment before I wanted to.
It should be good to review now.

sejas (approved):
Works as described 👍 
I tested it again for new and existing sites.

PR #1078, opened 2025-03-17
Update translation handling experience

wojtekn:
## Related issues
- Fixes STU-79-linear-issue

## Proposed Changes

I propose to improve the translation UX:
- by merging two phrases that differ with a trailing dot
- by adding support for plurals

Also, while testing that, I noticed small bug which makes bottom part of the form invisible when there are two validation errors displayed.

## Testing Instructions

1. Open a 'Add site' form
2. Trigger cases with one and two validation errors
3. Confirm validation works as expected
4. Confirm the custom domain info text is visible

## Pre-merge Checklist

### DIFF
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -267,8 +267,7 @@ export const SiteForm = ( {
 	const dispatch = useAppDispatch();
 	const isOffline = useOffline();
 
-	const errorCount = error && customDomainError ? 2 : error || customDomainError ? 1 : 0;
-	const hasErrors = errorCount > 0;
+	const errorCount = [ error, customDomainError ].filter( Boolean ).length;
 
 	const offlineMessage = __( 'Changing WordPress version requires an internet connection.' );
 	const wpVersions = useRootSelector(
@@ -358,7 +357,7 @@ export const SiteForm = ( {
 											{ __( 'Advanced settings' ) }
 										</div>
 									</Button>
-									{ hasErrors && (
+									{ errorCount > 0 && (
 										<span className="text-red-500 text-[13px] leading-[16px] ml-2 flex items-center">
 											<Icon icon={ warning } size={ 16 } className="mr-1 fill-red-500" />
 											{ sprintf(
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -372,7 +372,7 @@ export const SiteForm = ( {
 								<div
 									className={ cx(
 										'transition-all duration-500 ease-in-out overflow-hidden flex flex-col gap-2',
-										isAdvancedSettingsVisible ? 'max-h-96 opacity-100' : 'max-h-0 opacity-0'
+										isAdvancedSettingsVisible ? 'max-h-100 opacity-100' : 'max-h-0 opacity-0'
 									) }
 								>
 									<div className={ cx( 'flex flex-col gap-1.5 leading-4 py-2' ) }>
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -370,8 +370,8 @@ export const SiteForm = ( {
 								</div>
 								<div
 									className={ cx(
-										'transition-all duration-500 ease-in-out overflow-hidden flex flex-col gap-2',
-										isAdvancedSettingsVisible ? 'max-h-[450px] opacity-100' : 'max-h-0 opacity-0'
+										'transition-all duration-500 ease-in-out overflow-hidden flex flex-col gap-2 interpolate-size-allow-keywords',
+										isAdvancedSettingsVisible ? 'h-auto opacity-100' : 'h-0 opacity-0'
 									) }
 								>
 									<div className={ cx( 'flex flex-col gap-1.5 leading-4 py-2' ) }>
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -1,6 +1,6 @@
 import { Icon, SelectControl } from '@wordpress/components';
 import { createInterpolateElement } from '@wordpress/element';
-import { __ } from '@wordpress/i18n';
+import { __, sprintf, _n } from '@wordpress/i18n';
 import { tip, warning, trash, chevronRight, chevronDown, chevronLeft } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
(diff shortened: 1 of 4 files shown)


src/components/site-form.tsx
```diff
@@ -266,6 +266,10 @@ export const SiteForm = ( {
 	const getDocsLink = useDocsLink();
 	const dispatch = useAppDispatch();
 	const isOffline = useOffline();
+
+	const errorCount = error && customDomainError ? 2 : error || customDomainError ? 1 : 0;
```
wojtekn: Nested ternary are not great, but this one seems easy to ready. Any thoughts?

src/components/site-form.tsx
```diff
@@ -353,19 +357,21 @@ export const SiteForm = ( {
...
+											) }
 										</span>
 									) }
 								</div>
 								<div
 									className={ cx(
 										'transition-all duration-500 ease-in-out overflow-hidden flex flex-col gap-2',
-										isAdvancedSettingsVisible ? 'max-h-96 opacity-100' : 'max-h-0 opacity-0'
+										isAdvancedSettingsVisible ? 'max-h-100 opacity-100' : 'max-h-0 opacity-0'
 									) }
```
wojtekn, line 375: Do we ever need max height in this case?

src/components/site-form.tsx
nightnei:
```suggestion	
	const errors = [error, customDomainError].filter(item => item);
```

src/components/site-form.tsx
```diff
@@ -266,6 +266,10 @@ export const SiteForm = ( {
 	const getDocsLink = useDocsLink();
 	const dispatch = useAppDispatch();
 	const isOffline = useOffline();
+
+	const errorCount = error && customDomainError ? 2 : error || customDomainError ? 1 : 0;
+	const hasErrors = errorCount > 0;
```
nightnei:
```suggestion
```

src/components/site-form.tsx
```diff
@@ -353,19 +357,21 @@ export const SiteForm = ( {
 											{ __( 'Advanced settings' ) }
 										</div>
 									</Button>
-									{ ( error || customDomainError ) && (
+									{ hasErrors && (
```
nightnei:
```suggestion
									{ errors.length && (
```

src/components/site-form.tsx
```diff
@@ -353,19 +357,21 @@ export const SiteForm = ( {
...
 										<span className="text-red-500 text-[13px] leading-[16px] ml-2 flex items-center">
 											<Icon icon={ warning } size={ 16 } className="mr-1 fill-red-500" />
-											{ error && customDomainError
-												? __( '2 errors found' )
-												: __( '1 error found' ) }
+											{ sprintf(
+												/* translators: %d: number of errors found */
+												_n( '%d error found', '%d errors found', errorCount ),
```
nightnei, line 365:
```suggestion
												_n( '%d error found', '%d errors found', errors.length ),
```

nightnei (changes requested):
In general - nice improvement 👍
But WDYT about my proposition? As for me, it's even more simple and clear at a glance.

src/components/site-form.tsx
nightnei: Honestly, I have no idea why do we need two `max-height` + `opacity`, if we can simplify it to `display` 'none | block'

nightnei: UPDATE: oh, I see, it's due to animation.
but regarding max-height - we can replace it with just height.

wojtekn: It seems my fix broke the animation as `max-h-100` is not a valid Tailwind property. Auto wouldn't work, too, as it wouldn't be able to animate between absolute 0px and relative max.

I used an arbitrary max height that is higher than highest expected height and it works nice now.

src/components/site-form.tsx
wojtekn: Thanks! I slightly modified the proposes approach, but now it's much clearer.

src/components/site-form.tsx
sejas: If we want to use h-auto instead of an arbitrary heigh, we can use [interpolate-size](https://developer.mozilla.org/en-US/docs/Web/CSS/interpolate-size). I'll create a PR pointing to this branch to show an example.

sejas: Here is the PR using `interpolate-size: allow-keywords`:  https://github.com/Automattic/studio/pull/1098

src/components/site-form.tsx
```diff
@@ -266,6 +266,9 @@ export const SiteForm = ( {
 	const getDocsLink = useDocsLink();
 	const dispatch = useAppDispatch();
 	const isOffline = useOffline();
+
+	const errorCount = [ error, customDomainError ].filter( Boolean ).length;
```
sejas: ✨

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -101,7 +101,7 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
 				} catch ( wpError ) {
 					console.error( 'Error changing WordPress version:', wpError );
 					const errorMessage = stripAnsi( ( wpError as Error )?.message );
-					setIsChangeWpError( __( 'Error changing WordPress version.' ) );
+					setIsChangeWpError( __( 'Error changing WordPress version' ) );
```
sejas: We finish in a dot in path error. I guess it's expected as it has two sentences.

Screenshot 2025-03-20 at 09 58 21

sejas (approved):
Code looks great. I confirm the errors found now is dynamic and works with plural. The advance settings animation is still working.

Screenshot 2025-03-20 at 10 00 03

Screenshot 2025-03-20 at 09 58 11

nightnei (approved):
LGTM 👍

src/modules/site-settings/edit-site-details.tsx
wojtekn: Thanks for sharing the context. I tested it again and noticed we display this error two times: in dialog and as a form validation error. Technically, it's not a form validation error, but a form submission error. I've removed the form validation error, so the dialog is the only place that reports the error.

PR #1081, opened 2025-03-17
Improvements to site thumbnail generation

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10560

## Proposed Changes

- Despite my best efforts, I couldn't come up with a better solution for the Windows thumbnail bug than simply increasing the hard-coded timeout in `waitForCapture`. 
- I also changed the logic so a new thumbnail is not generated (and the old one is removed) when the home page responds with an HTTP error status code. This can happen if port 80 is already busy, for example. 

**QUESTION:** I'm open to suggestions for which behavior to implement when the site home page returns an error. I think generating a thumbnail makes for weird behavior, but maybe it's better to leave the old thumbnail in place, I'm not sure… 

## Testing Instructions

1. Ping me on Slack to get a site backup for reproducing this error
2. On Windows, create a Studio site from that backup
3. Stop and start the site a few times
4. Ensure that the site thumbnail appears correctly with each restart
5. Quit Studio
6. Have Local installed and start a site with a custom domain 
7. Start Studio
8. Start the Studio site
9. Ensure that the site thumbnail displays `No preview available`

## Pre-merge Checklist

### DIFF
--- a/src/screenshot-window.ts
+++ b/src/screenshot-window.ts
@@ -30,7 +30,7 @@ export function createScreenshotWindow( captureUrl: string ) {
 		await window.loadURL( captureUrl );
 		await responseStatusCodePromise;
 		await window.webContents.insertCSS( `
-			body {
+			body, html {
 				overflow: hidden;
 				height: 100vh;
 			}
@@ -39,7 +39,7 @@ export function createScreenshotWindow( captureUrl: string ) {
 			}
 		` );
 
-		await new Promise( ( resolve ) => setTimeout( resolve, 500 ) );
+		await new Promise( ( resolve ) => setTimeout( resolve, 2000 ) );
 		return window.webContents.capturePage();
 	};
 
--- a/src/screenshot-window.ts
+++ b/src/screenshot-window.ts
@@ -12,14 +12,23 @@ export function createScreenshotWindow( captureUrl: string ) {
 		webPreferences: { session: newSession },
 	} );
 
-	const finishedLoading = new Promise< void >( ( resolve ) => {
-		window.webContents.on( 'did-finish-load', () => resolve() );
+	const responseStatusCodePromise = new Promise< void >( ( resolve, reject ) => {
+		newSession.webRequest.onCompleted( ( details ) => {
+			if ( details.resourceType !== 'mainFrame' ) {
+				return;
+			}
+
+			if ( details.statusCode < 200 || details.statusCode >= 400 ) {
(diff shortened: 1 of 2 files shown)


src/screenshot-window.ts
```diff
@@ -12,16 +12,25 @@ export function createScreenshotWindow( captureUrl: string ) {
 		webPreferences: { session: newSession },
 	} );
 
-	const finishedLoading = new Promise< void >( ( resolve ) => {
-		window.webContents.on( 'did-finish-load', () => resolve() );
-	} );
```
fredrikekelund, line 17: `window.loadURL` returns a promise that resolves when the `did-finish-load` event has fired, so I took the opportunity to simplify this logic.

src/screenshot-window.ts
```diff
@@ -12,16 +12,25 @@ export function createScreenshotWindow( captureUrl: string ) {
...
+	const responseStatusCodePromise = new Promise< void >( ( resolve, reject ) => {
+		newSession.webRequest.onCompleted( ( details ) => {
```
fredrikekelund, line 16: Adding a URL filter ([docs](https://www.electronjs.org/docs/latest/api/web-request)) did not work well in my testing. Likely because of potential redirects.

src/site-server.ts
```diff
@@ -199,7 +199,10 @@ export class SiteServer {
 			.mkdir( outDir, { recursive: true } )
 			.then( waitForCapture )
 			.then( ( image ) => fs.promises.writeFile( outPath, image.toPNG() ) )
-			.catch( Sentry.captureException )
+			.catch( async ( error ) => {
+				Sentry.captureException( error );
+				await fs.promises.unlink( outPath );
```
fredrikekelund, line 204: Deleting the thumbnail path makes the UI look like this:

Screenshot 2025-03-17 at 17 00 02

wojtekn: This is expected.

wojtekn (approved):
Works well on Windows now. MacOS still works fine.

src/site-server.ts
fredrikekelund: I know it's intentional for the thumbnail to look this way when there's no image. My point was that deleting the thumbnail when the homepage returns an errant HTTP status is new behavior, so I wanted to clarify what the UI will look like in those cases.

wojtekn: Thanks for clarifying that, makes sense.

PR #1082, opened 2025-03-17
Fix Sync "Connect a WordPress.com site" modal tabbing and styling

ivan-ottinger:
## Related issues

Fixes https://github.com/Automattic/dotcom-forge/issues/10580.

## Proposed Changes

- add outline to site items, allowing to select each one by pressing Spacebar

Markup on 2025-03-17 at 18:02:58

- fix outline of sites' address

| Before | After |
|--------|--------|
| Markup on 2025-03-18 at 11:36:16 | Markup on 2025-03-18 at 11:37:10 | 

- add outline of `Create a new WordPress.com site` button in the modal footer

Markup on 2025-03-17 at 18:06:36

- fix alignment of site's URL

| Before | After |
|--------|--------|
| Markup on 2025-03-18 at 11:30:27 | Markup on 2025-03-17 at 17:42:06 | 

- fix site address truncating (when they are too long)

## Testing Instructions

1. Check out the branch PR and build the app with `npm start`.
2. Head over to the Sync tab of a local site that does not have any WPCOM site connected yet.
3. Tab through the app elements until you reach the `Connect site` button. Open it with pressing Spacebar.
4. Tab through the elements in the modal. It should be possible to select each site item, but also its URL.
5. If you press Spacebar while the site URL is selected, it should open in a new browser tab / window.
6. If you press Spacebar while the whole site item is selected, it should select it.
7. It should be possible to tab through (and open) the `Enable hosting site features` and `Upgrade plan` links as well (on particular sites).
8. When you tab to the bottom of the list of sites and hit Tab more, it should be possible to select the `Create a new WordPress.com site` and pressing Spacebar should open it in the new browser tab / window. 
9. You may manually make a site name and site URL longer and observe it gets truncated correctly.
10. You may also test the tabbing using VoiceOver. It should work correctly.
11. The site list should work well in RTL locale as well.

## Pre-merge Checklist

### DIFF
--- a/src/components/sync-sites-modal-selector.tsx
+++ b/src/components/sync-sites-modal-selector.tsx
@@ -213,7 +213,7 @@ function SiteItem( {
 				onClick();
 			} }
 		>
-			<div className="flex flex-col gap-0.5 min-w-0">
+			<div className="flex flex-col gap-0.5 min-w-0 items-start">
 				<div className={ cx( 'a8c-body truncate', ! isSyncable && 'text-a8c-gray-30' ) }>
 					{ site.name }
 				</div>
--- a/src/components/sync-sites-modal-selector.tsx
+++ b/src/components/sync-sites-modal-selector.tsx
@@ -156,7 +156,7 @@ function ListSites( {
 	const sortedSites = getSortedSites( syncSites );
 
 	return (
-		<div className="flex flex-col overflow-y-auto h-full">
+		<div className="flex flex-col overflow-y-auto h-full pt-px">
 			{ sortedSites.map( ( site ) => (
 				<SiteItem
 					key={ site.id }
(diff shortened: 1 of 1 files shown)


bcotrim (approved):
LGTM 👍 

A minor suggestion would be to move the focus to the "Connect" button after selecting a site, so the user wouldn't have to navigate through all the options.

wojtekn:
It fixes the issue - user can use keyboard to navigate through the Sync tab now.

> If you press Spacebar while the site URL is selected, it should open in a new browser tab / window.

According to [keyboard nav guidelines](https://webaim.org/techniques/keyboard/), links should only be activated using Enter. However, it seems in Studio, we already allow to open using Enter or Spacebar, so we can leave it as is.

> If you press Spacebar while the whole site item is selected, it should select it.

Should we treat the row as a button, not a checkbox, and allow selecting it using Enter in addition to Spacebar? Another thing is that the row gets selected after the user navigates to a link.

ivan-ottinger:
> Should we treat the row as a button, not a checkbox, and allow selecting it using Enter in addition to Spacebar? Another thing is that the row gets selected after the user navigates to a link.

That's a good catch! I have enabled `Enter` usage in https://github.com/Automattic/studio/pull/1082/commits/222fcfdd9effccc93125913846626d76480a184e. It also makes sure the row does not get selected when the link is opened.

ivan-ottinger:
> LGTM 👍
> 
> A minor suggestion would be to move the focus to the "Connect" button after selecting a site, so the user wouldn't have to navigate through all the options.

That's a great idea. While we discussed considering a solution with state variable, turns out we should be good with one `useEffect`: https://github.com/Automattic/studio/pull/1082/commits/3b7412d152cea3104dc164184c984253b4c52888.

Feel free to let me know what you think. 🙂

wojtekn (approved):
Feature-wise, works as expected.

bcotrim:
> That's a great idea. While we discussed considering a solution with state variable, turns out we should be good with one useEffect: https://github.com/Automattic/studio/commit/3b7412d152cea3104dc164184c984253b4c52888.
Feel free to let me know what you think. 🙂

Looks great! Thanks for following up. 👍

PR #1086, opened 2025-03-18
ESLint rule to disallow "floating" promises

fredrikekelund:
## Related issues

- Fixes STU-336

## Proposed Changes

- Enable the `@typescript-eslint/no-floating-promises` ESLint rule ([docs](https://typescript-eslint.io/rules/no-floating-promises/)) and fix the places ESLint yells at. So far, I've only fixed a subset of errors. I'm looking for alignment on the team before proceeding.

This rule generates the following error:

> Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator

## Testing Instructions

CI should pass

## Pre-merge Checklist


src/about-menu/open-about-menu.ts
```diff
@@ -27,11 +27,9 @@ export function openAboutWindow() {
 		},
 	} );
 
-	aboutWindow.loadFile( aboutPath );
-
 	// Open external links in the default browser
 	aboutWindow.webContents.setWindowOpenHandler( ( { url } ) => {
-		shellOpenExternalWrapper( url );
+		void shellOpenExternalWrapper( url );
```
fredrikekelund, line 32: `void` is a way of intentionally signaling that no error handling is required for this promise. Most of the linter errors in our existing codebase can be handled this way, and it's the part I expect to be most controversial.

nightnei:
In general, it's a great idea and a way to improve our codebase, but I have a concern. Let's imagine a case:

```
const qqq = async () => {
  try {
  
  } catch () {
  	// we handle errors here, so should we force handling errors inside `eee()`. Ofc, we can use `void` inside `eee`, but it looks wordy, however on the other hand it's pretty straightforward.... 
  	}
};
const eee = async () => {
	await qqq();
}
```

It's mostly thought out loud 😅 Hm, maybe we indeed could try to go this way with ESLint and adding `void` when it's necessary.

fredrikekelund:
@nightnei, yeah, I know… Having to prefix promises we know won't ever throw an error with `void` is pretty verbose. 

I paused my work on this the other day because I had a slight feeling that 95% of promises would be prefixed with `void`, which seemed very verbose. 

I still think it's fundamentally smart to enforce some kind of error handling for promises. I guess the risk is that we just get into a habit of prefixing everything with `void`, even promises that should really have a `.catch` handler…

src/ipc-types.d.ts
fredrikekelund, line 1: The changes in this file allowed me to make some functions in `src/ipc-handlers.ts` synchronous. When called from the front-end using `getIpcApi()`, they still return promises, but when called directly from the back-end, this change means they no longer require `await`, `void`, etc.

src/components/content-tab-overview.tsx
```diff
@@ -100,23 +109,29 @@ function CustomizeSection( {
 		{
 			label: __( 'Customizer' ),
 			icon: edit,
-			onClick: () => getIpcApi().openSiteURL( selectedSite.id, '/wp-admin/customize.php' ),
+			onClick: () => {
+				void getIpcApi().openSiteURL( selectedSite.id, '/wp-admin/customize.php' );
```
sejas: Would it be equivalent in all these cases of onClick to make an async/await call?
I don't see any errors in my linter indicating that there is an unhandled error for this call.

```suggestion
			onClick: async () => {
				await getIpcApi().openSiteURL( selectedSite.id, '/wp-admin/customize.php' );
```

sejas (commented):
I see the problem we are trying to solve, but I fear we'll normalize the use of `void`.
I would expect that the usage of `void` should be limited to very few scenarios, but currently is everywhere.
I guess we would need to refactor our code and wrap most of our IPC functions with a try/catch to "display an alert".
For instance, `getIpcApi().copyText` or `getIpcApi().openURL` should already catch the error and maybe display an alert to the user if something fails, instead of having to add void or a try/catch every time we use those functions.

UPDATE:
Actually, that's the current pattern for some functions like  [`getIpcApi().openURL`](https://github.com/Automattic/studio/blob/0d43d12ab2042b1db5ce1f6e1a2289265a45d3fc/src/ipc-handlers.ts#L771), which doesn't return any error 🤔 : https://github.com/Automattic/studio/blob/0d43d12ab2042b1db5ce1f6e1a2289265a45d3fc/src/lib/shell-open-external-wrapper.ts#L15-L49

fredrikekelund:
Thanks for the review, @sejas 👍 I wish there was even smarter linting for this so we wouldn't have to do anything for promises that don't ever throw errors. That's not the case right now, though…

> I see the problem we are trying to solve, but I fear we'll normalize the use of void.

The risk I see is that we prefix promises with `void` even when they should have a `catch` handler. In this PR, I used `void` only for promises with internal error handling or with a very low risk of propagating errors. 

> I guess we would need to refactor our code and wrap most of our IPC functions with a try/catch to "display an alert".
> For instance, getIpcApi().copyText or getIpcApi().openURL should already catch the error and maybe display an alert to the user if something fails, instead of having to add void or a try/catch every time we use those functions.

`void` isn't really very verbose, IMO, even if it does add some noise. Again, I think the rule should be to use `void` when calling async methods with internal error handling or with a very low risk of propagating errors.

src/components/content-tab-overview.tsx
fredrikekelund: Yep, this would also satisfy the linter. If you think that's preferable here, do you mind elaborating on why and when you think `void` is suitable compared to `async`/`await`?

sejas: I used "void" in the past when defining the return type while declaring a function. It's the first time I've seen its usage when calling a function, so I'm not accustomed to it.
Using `async/await` feels more natural to me, as we are not overriding the return type of the function.

fredrikekelund: Using it as an operator is definitely esoteric, even if it's [perfectly valid](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void). I'm guessing the ESLint authors chose `void` for this purpose since it doesn't change the functionality of the code, and it's relatively succinct. 

> Using async/await feels more natural to me

If we land this PR, I think we'll need to get accustomed to defaulting to `void` when we don't care about the return value or executing things in sequence. `async/await` will work the same way in many cases, but not always. 

> as we are not overriding the return type of the function

None of the `void` operators in this PR change the return type of any of the functions.

sejas (approved):
TIL about using `void` to discard the returned "promise" and using async functions in sync scopes.

I vote in favor of @fredrikekelund PR 👌

fredrikekelund:
Cool! I'll fix the merge conflicts and land this PR today or tomorrow 👍

fredrikekelund:
I ended up making a bunch of changes to `src/ipc-types.d.ts` and `src/preload.ts` to reduce the burden when working with components and other app code (and also to make the diff in this PR smaller).

Here's a summary of the changes I made:

- Functions in `src/ipc-handlers.ts` no longer have to be async. Let's make them sync if they don't contain any `await` expressions (as we would with any other code).
- I've removed explicit type definitions from the function arguments in `src/preload.ts`. Typescript intuits these automatically, ensuring correctness. 
- I've tweaked the `IpcApi` type in `src/ipc-types.d.ts` to use a `NonInvokeHandlers` list that specifies which IPC functions are called with `ipcRenderer.send`. Using `ipcRenderer.send` means that the IPC function doesn't return a promise and we don't have to think about the `@typescript-eslint/no-floating-promises` ESLint rule. If you're adding a new IPC function and don't care about the return value, use this approach.

PR #1067, opened 2025-03-13
Fix: Prevent server process from hanging during stop operation

bcotrim:
## Related issues

- Fixes #985

## Proposed Changes

- Added isResolved flag to prevent multiple promise resolutions in the #killProcess method
- Added force kill with SIGKILL if graceful shutdown fails
- Improved process cleanup by clearing timeouts and handling process state properly

## Testing Instructions

Follow the steps mentioned in the issue
- Setup default WordPress site.
- Install Remote Data Blocks Plugin.
- Setup third party data ingestion via Google Sheets (https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/tutorials/google-sheets.md).
- When the site eventually fails (This site can’t be reached - localhost refused to connect.). Attempt to stop the site.
- The site stops after the timeout timer runs out, instead of hanging.
- Confirm that there are no pending processes for that site

More details on reproducing the issue (thanks @ppolo99): p1741865890679379/1741864717.468099-slack-C04GESRBWKW

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1087, opened 2025-03-18
Error handling improvements in `phpGetThemeDetails`

fredrikekelund:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10646
- Fixes https://github.com/Automattic/dotcom-forge/issues/7291

## Proposed Changes

- Stop sending exceptions thrown in `phpGetThemeDetails` to Sentry. These exceptions indicate more general errors with the PHP-WASM environments and don't tell us anything useful.
- Parse the PHP output with zod to ensure we conform to the expected schema.

> [!IMPORTANT]  
> This PR should also fix some of the Sentry issues connected with https://github.com/Automattic/dotcom-forge/issues/7291. The underlying cause of timeout errors varies depending on the message name (`start-server`, `stop-server`, or `run-php`). `run-php` indicates an error in `phpGetThemeDetails` and it'll be fixed by this PR (by simply not reporting those errors to Sentry anymore). 
>
> Not reporting to Sentry may seem like a non-solution, but I believe it's viable because the fact that PHP execution fails is not always within our control (it could be due to PHP-WASM problems, or issues with the WP installation, etc). We also recently improved the logic for killing unresponsive processes in https://github.com/Automattic/studio/pull/1067.
> 
> https://github.com/Automattic/dotcom-forge/issues/7291 has grouped several timeout errors, and the ones related to `start-server` messages will not be affected by this PR. I will create new GitHub issues for tracking those.

## Testing Instructions

1. Have a Studio site
2. Open the `appdata-v1.json` file in your editor
3. Go to wp-admin and change the theme
4. Ensure that the `themeDetails` value for your site was updated correctly in `appdata-v1.json`

## Pre-merge Checklist

### DIFF
--- a/src/lib/php-get-theme-details.ts
+++ b/src/lib/php-get-theme-details.ts
@@ -1,5 +1,14 @@
-import * as Sentry from '@sentry/electron/main';
+import { z } from 'zod';
 import SiteServerProcess from 'src/lib/site-server-process';
+
+export const themeDetailsSchema = z.object( {
+	name: z.string().catch( '' ),
+	path: z.string(),
+	slug: z.string(),
+	isBlockTheme: z.boolean(),
+	supportsWidgets: z.boolean(),
+	supportsMenus: z.boolean(),
+} );
 
 export async function phpGetThemeDetails(
 	server: SiteServerProcess
@@ -8,13 +17,11 @@ export async function phpGetThemeDetails(
 		throw Error( 'PHP is not instantiated' );
 	}
 
-	let themeDetails = null;
 	const themeDetailsPhp = `<?php
 	require_once('${ server.php.documentRoot }/wp-load.php');
 	$theme = wp_get_theme();
 	echo json_encode([
 		'name' => $theme->get('Name'),
-		'uri' => $theme->get('ThemeURI'),
 		'path' => $theme->get_stylesheet_directory(),
 		'slug' => $theme->get_stylesheet(),
 		'isBlockTheme' => $theme->is_block_theme(),
@@ -22,16 +29,14 @@ export async function phpGetThemeDetails(
 		'supportsMenus' => get_registered_nav_menus() || current_theme_supports('menus'),
 	]);
 	`;
+
 	try {
-		themeDetails = await server.runPhp( {
+		const themeDetailsRaw = await server.runPhp( {
 			code: themeDetailsPhp,
 		} );
-		themeDetails = JSON.parse( themeDetails );
+		const themeDetailsParsed = JSON.parse( themeDetailsRaw );
+		return themeDetailsSchema.parse( themeDetailsParsed );
 	} catch ( error ) {
-		Sentry.captureException( error, {
-			extra: { themeDetails },
-		} );
-		themeDetails = null;
+		return undefined;
 	}
-	return themeDetails;
 }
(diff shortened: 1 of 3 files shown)


src/lib/php-get-theme-details.ts
```diff
@@ -1,37 +1,42 @@
-import * as Sentry from '@sentry/electron/main';
+import { z } from 'zod';
 import SiteServerProcess from 'src/lib/site-server-process';
 
+export const themeDetailsSchema = z.object( {
+	name: z.string().catch( '' ),
```
sejas, line 5: TIL!

src/lib/php-get-theme-details.ts
```diff
@@ -1,37 +1,42 @@
-import * as Sentry from '@sentry/electron/main';
+import { z } from 'zod';
 import SiteServerProcess from 'src/lib/site-server-process';
 
+export const themeDetailsSchema = z.object( {
```
sejas: I don't think we need to export this schema.

```suggestion
const themeDetailsSchema = z.object( {
```

sejas (approved):
I confirm that the theme details are still working, and after changing the theme, it is saved in appData.
Great work! I'm a big fan of reducing Sentry noise.

Screenshot 2025-03-18 at 14 30 04

fredrikekelund:
Thanks for the review, @sejas! Just a quick note that I updated the PR description after realizing that this PR also fixes https://github.com/Automattic/dotcom-forge/issues/7291 (at least in part). No new testing is needed, but I wanted to bring it to your attention

src/lib/php-get-theme-details.ts
fredrikekelund: Thanks! That was a leftover from me playing around with where to put this schema

sejas (approved):

fredrikekelund:
> the ones related to start-server messages will not be affected by this PR. I will create new GitHub issues for tracking those.

After looking closer at [STUDIO-9J](https://a8c.sentry.io/issues/5321107625/) and [STUDIO-9Q](https://a8c.sentry.io/issues/5324533862/) (the issues related to creating or starting sites), I realized that they seem to only affect older releases (circa 1.2.0). So instead of opening issues to fix them, I marked the Sentry issues as resolved. We can tackle them if it turns out later they also affect new releases.

src/lib/php-get-theme-details.ts
fredrikekelund: I included this because it wasn't 100% clear to me if `WP_Theme::get( 'Name' )` ([docs](https://developer.wordpress.org/reference/classes/wp_theme/get/)) is guaranteed to return a string. It seems theoretically possible that it would also return a boolean or array. Defaulting to an empty string in those cases seemed reasonable.

src/lib/php-get-theme-details.ts
```diff
@@ -1,37 +1,42 @@
...
 		throw Error( 'PHP is not instantiated' );
 	}
 
-	let themeDetails = null;
 	const themeDetailsPhp = `<?php
 	require_once('${ server.php.documentRoot }/wp-load.php');
 	$theme = wp_get_theme();
 	echo json_encode([
 		'name' => $theme->get('Name'),
-		'uri' => $theme->get('ThemeURI'),
```
fredrikekelund, line 17: In case it wasn't already clear, we don't actually use this data. `uri` is not included in the `StartedSiteDetails[ 'themeDetails' ]` type definition.

PR #1088, opened 2025-03-18
Fix symlink issues by using useHostFilesystem

sejas:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10566

## Proposed Changes

- I suggest using `useHostFilesystem` instead of mounting each symlink directory.

## Testing Instructions

1. Have a Studio site where Yoast SEO is not already installed
2. Download [Yoast SEO](https://wordpress.org/plugins/wordpress-seo) and extract it to `~/wordpress-seo`
3. In your site's `wp-content/plugins` directory, run `ln -s ../../../../wordpress-seo` (you'll need at least four levels of `../`, but the symlink also needs to resolve correctly)
4. Start the Studio site
5. Navigate to `/wp-admin/plugins.php`
6. Observe the Yoast SEO plugin there
7. Activate it
8. Observe the site is running correctly as always

## Pre-merge Checklist


fredrikekelund:
This fixed the problem for me on Mac, but it doesn't work at all on Windows (I can't start sites—even without symlinks). I get the following error on Windows:

```
 Error: EPERM: operation not permitted, lstat 'C:\Config.Msi'
```

fredrikekelund:
Looking at the `useHostFilesystem` [source](https://github.com/WordPress/wordpress-playground/blob/18524a4d63500db562c3dbd37ce771eb3afc26ce/packages/php-wasm/node/src/lib/use-host-filesystem.ts#L12), it seems possible that it would work on Windows, but we'd need to start by capturing errors thrown from `lstatSync`.

wojtekn:
Nice, simplifying symlink handling that way looks promising.

sejas:
I'm closing this PR to find the real issue in Playground/Emscripten. I'll work on it later this week and It will take more time than this solution, but it should be a more conservative solution that fixes the root cause.

bgrgicak:
@sejas using `useHostFilesystem` is the correct way of adding symlink support, and it should fix both Symlinks with relative target paths and nested symlinks like in the Jetpack plugin build.

We need this because of the way Emscripten's file system works. 
Playground uses NODEFS as the main filesystem in the Node version of Playground. 
Internally, in Emscripten, this creates a MEMFS (in-memory file system) root node and mounts the[ document root from Studio as a NODEFS node inside the filesystem.](https://github.com/Automattic/studio/blob/2ff18953f79c3eda7ce958b573d12cb129753052/vendor/wp-now/src/wp-now.ts#L238)

When Emscripten wants to access a symlink, it searches for the node in its filesystem.
In the case of an absolute symlink, it would point to the root node (`/`), which for Emscripten is a MEMFS node and not the underlying OS node.
Because the MEMFS node doesn't contain the rest of the symlink path, it fails, but if we run `useHostFilesystem` it will mount the directories from the OS root folder into the Emscripen's FS root folder, and the path will be available.

bgrgicak:
@sejas I will reopen this and add a few small changes to restart the discussion about using `useHostFilesystem`.

bgrgicak:
`useHostFilesystem` doesn't support Windows, I will take a look at that in Playground and ensure symlinks work in Windows.

sejas:
@bgrgicak , thanks for opening the PR, and continuing the discussion.
My biggest concern is that we'll expose the whole filesystem to every WordPress site, and it could be a security risk. Plugins reading secrets, or removing important files. Do you think it could be a real risk?

bgrgicak:
>  My biggest concern is that we'll expose the whole filesystem to every WordPress site, and it could be a security risk. Plugins reading secrets, or removing important files. Do you think it could be a real risk?

Playground/Emscripten can access only what we mount, similarly to Docker. This means that we either need to give it wide access like this PR suggests or find the exact paths that need to be mounted like the `SymlinkManager` does.

Below is a breakdown of these approaches. 

TL;DR: My conclusion is that the added risk of running Playground with `useHostFilesystem` is minimal because PHP inherits permissions from the Node process and can do only as much as the Node process and Studio can.

## SymlinkManager

This ensures PHP has access only to a subset of directories and limits the security risks to these directories. 

While the exposed surface is limited, the user still isn't in direct control, and could unintentionally mount a directory they aren't aware of. For example, the Jetpack plugin contains internal symlinks that the user might not be aware of.

If we want to enable the SymlinkManager to work for all use cases, WP-now should follow symlinks and also mount nested symlinked folders and symlinked folders that use relative paths.

I'm sure that with enough attempts, we could make it work, but we would need to keep iterating on the SymlinkManager until we get it right.

## useHostFilesystem

Using this approach, we simply enable PHP to access everything on the OS filesystem, so there can't be any edge cases that would break Studio in the future. 

While exposing the OS filesystem sounds scary, it's the same as installing Apache on your machine and using it to run WordPress. 

By exposing the OS filesystem to PHP, we give PHP only the permissions that the process has, so in the case of Studio, it's the user permissions.

For example, this PHP script works the same with Apache on a host machine as it does with Playground that has `useHostFilesystem`. PHP is only able to list the `/usr` directory, but they can't make any changes.

```php
<?php
$root_path = '/usr';

print_r(scandir($root_path)); 
// WORKS: lists the folder contents

file_put_contents($root_path . '/test.txt', 'test');
// FAILS: doesn't have write permissions

print_r(file_get_contents($root_path . '/test.txt'));
// FAILS: file doesn't exist

unlink($root_path . '/test.txt');
// FAILS: file doesn't exist
?>
```

Without `useHostFilesystem` Playground fails to list files in the directory.

## Potential alternative approaches

One alternative could be to use the existing symlink manager to detect if there are any symlinks in the project folder, and if there are, run `useHostFilesystem` to ensure Playground has OS filesystem access.

Or if we want to give control to the user, we could have a toggle in site settings that enables symlinks.

fredrikekelund:
> If we want to enable the SymlinkManager to work for all use cases, WP-now should follow symlinks and also mount nested symlinked folders and symlinked folders that use relative paths.

This is what `SymlinkManager` already does. Do you mean that we would need to also modify Playground code to make this work fully?

> My conclusion is that the added risk of running Playground with useHostFilesystem is minimal because PHP inherits permissions from the Node process and can do only as much as the Node process and Studio can.

Malicious code could easily do a lot of damage without root privileges (as @sejas also mentioned). It could, for example, read config files from the home directory or delete or modify any directory the user can access.

A local installation is not a typical WordPress installation environment, and the risk for these types of attacks seems relatively low. However, it's still a reality that we'd be trading a higher-security solution for a lower-security solution by exposing the entire file system to WordPress.

If it's not feasible to make symlinks work with the current approach, then I think working symlinks are more important than the security benefits of the current approach, but it's not clear to me exactly why and how we think `SymlinkManager` doesn't work (or couldn't work fully).

bgrgicak:
Thanks for the feedback, I will spend a bit more time finding ways to mount symlink paths on demand and see if we can get that to work instead of exposing the entire host FS.

PR #1089, opened 2025-03-18
Start proxy during site start and improve error wording

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10644

## Proposed Changes

A "Custom domain set up failed" error is displayed immediately after user starts Studio, even if the user doesn't use a custom domain. I propose to improve Proxy error wording to be more informative about what happened:

Screenshot 2025-03-18 at 16 19 17

Also, I delay starting the proxy until the user tries to start a site that uses custom domains, so users who have port 80 busy and start Studio won't see error immediately.

## Testing Instructions

1. Start a local web server app, e.g., Local
2. Start Studio
3. Confirm there is no error about custom domains displayed
4. Start a site that doesn't use custom domains
5. Confirm there is no error about custom domains displayed
4. Start a site that uses custom domains
7. Confirm that the error message is displayed and that the new wording is clear
8. Close Studio and other web server that uses 80 port
9. Repeat above steps and confirm there is no error message and that site is accessible using custom domain 

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -287,7 +287,9 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 			} catch ( error ) {
 				if (
 					error instanceof Error &&
-					error.message.includes( 'Studio failed to start the proxy server' )
+					error.message.includes(
+						'Studio failed to start the proxy server because port 80 is already in use'
+					)
 				) {
 					getIpcApi().showErrorMessageBox( {
 						title: __( 'Studio failed to initialize custom domains' ),
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -285,14 +285,27 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 			try {
 				updatedSite = await getIpcApi().startServer( id );
 			} catch ( error ) {
-				getIpcApi().showErrorMessageBox( {
-					title: __( 'Failed to start the site server' ),
-					message: __(
-						"Please verify your site's local path directory contains the standard WordPress installation files and try again. If this problem persists, please contact support."
-					),
-					error,
-					showOpenLogs: true,
-				} );
+				if (
+					error instanceof Error &&
+					error.message.includes( 'Studio failed to start the proxy server' )
+				) {
+					getIpcApi().showErrorMessageBox( {
+						title: __( 'Studio failed to initialize custom domains' ),
+						message: __(
+							'Studio needs to use port 80, but it’s already in use by another app. Close any local development apps and restart Studio.'
+						),
+						showOpenLogs: false,
+					} );
+				} else {
+					getIpcApi().showErrorMessageBox( {
+						title: __( 'Failed to start the site server' ),
+						message: __(
+							"Please verify your site's local path directory contains the standard WordPress installation files and try again. If this problem persists, please contact support."
+						),
+						error,
+						showOpenLogs: true,
+					} );
+				}
 				await getIpcApi().stopServer( id );
 			}
 
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -288,7 +288,7 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 				if (
 					error instanceof Error &&
 					error.message.includes(
-						'Studio failed to start the proxy server because port 80 is already in use'
+						'Studio failed to start the proxy server because port 80 is already in use.'
 					)
 				) {
 					getIpcApi().showErrorMessageBox( {
@@ -297,6 +297,17 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 							'Studio needs to use port 80 to enable custom domains, but it’s already in use by another app. Close any local development apps and restart Studio.'
 						),
 						showOpenLogs: false,
+					} );
+				} else if (
+					error instanceof Error &&
+					error.message.includes( 'Studio failed to start the proxy server.' )
+				) {
+					getIpcApi().showErrorMessageBox( {
+						title: __( 'Studio failed to initialize custom domains' ),
+						message: __(
+							'Please restart Studio and try again. If this problem persists, please contact support.'
+						),
+						showOpenLogs: true,
 					} );
 				} else {
 					getIpcApi().showErrorMessageBox( {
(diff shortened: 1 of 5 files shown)


src/lib/proxy-server.ts
```diff
@@ -89,25 +87,15 @@ export async function startProxyServer(): Promise< boolean > {
 					resolve();
 				} )
 				.on( 'error', ( err ) => {
-					console.log( err );
-					console.error( `Error starting proxy server on port 80:`, err );
+					console.error( `Error starting proxy server on port 80` );
```
wojtekn, line 117: Let's clean up the error logging, we are logging this error too many times.

src/lib/proxy-server.ts
```diff
@@ -89,25 +87,15 @@ export async function startProxyServer(): Promise< boolean > {
...
-			dialog.showMessageBox( mainWindow, {
-				type: 'error',
-				message: __( 'Custom domain set up failed' ),
-				detail: __(
-					'Studio needs to use port 80, but it’s already in use by another app. Close any local development apps and restart Studio.'
-				),
-				buttons: [ __( 'OK' ) ],
-			} );
-			return false;
+			throw new Error( 'Studio failed to start the proxy server' );
```
wojtekn: Let's move the part responsible for dialog to the renderer.

bcotrim (approved):
Code LGTM, and a much better user experience 👍 
Just FYI we have this related PR: https://github.com/Automattic/studio/pull/1090 there might be some code merging to do, and we should retest, depending on what is merged first.

My only suggestion would be to include unit tests for this scenario.

src/hooks/use-site-details.tsx
```diff
@@ -285,14 +285,27 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
-					showOpenLogs: true,
-				} );
+				if (
+					error instanceof Error &&
+					error.message.includes( 'Studio failed to start the proxy server' )
+				) {
+					getIpcApi().showErrorMessageBox( {
+						title: __( 'Studio failed to initialize custom domains' ),
+						message: __(
+							'Studio needs to use port 80, but it’s already in use by another app. Close any local development apps and restart Studio.'
```
sejas: Should we mention that the port 80 is only necessary for custom domains?

src/ipc-handlers.ts
```diff
@@ -405,6 +406,19 @@ export async function startServer(
...
 		);
+
+		try {
+			const proxyStarted = await startProxyServer();
+			if ( ! proxyStarted ) {
+				console.warn(
+					'Failed to start custom domain proxy server - custom domains will require port numbers'
+				);
+			}
+			console.log( 'Custom domain proxy server started successfully' );
```
sejas: No need to change anything. Just an observation. I noticed we display `Custom domain proxy server started successfully` each time we start a server with custom domain.
I think it's innocuous as we also display the `Domain example.wp.local added to hosts file for port 8900`

sejas (approved):
It works as expected. I confirm that the alert appears if port 80 is already in use. I also confirm that sites without a custom domain function as expected without any errors, even if port 80 is occupied. After freeing port 80, I was able to start two Studio sites with custom domains.

Great work!

Screenshot 2025-03-19 at 19 37 25


Here are the logs when running two sites with custom domains:

```
Server stopped
Stopped watching for file changes
SQLite folder already exists. Skipping download.
Domain example-2.wp.local added to hosts file for port 8901
Proxy server started on port 80
Custom domain proxy server started successfully
Starting server for 'example 2'
directory: /Users/macbookpro/Studio/example-2
mode: wordpress
php: 8.2
wp: latest
WordPress latest folder already exists. Skipping download.
SQLite folder already exists. Skipping download.
Server running at http://example-2.wp.local
Server started for 'example 2'
Domain example.wp.local added to hosts file for port 8900
Custom domain proxy server started successfully
Starting server for 'example'
directory: /Users/macbookpro/Studio/example
mode: wordpress
php: 8.2
wp: latest
WordPress latest folder already exists. Skipping download.
SQLite folder already exists. Skipping download.
Server running at http://example.wp.local
Server started for 'example' 
```

src/hooks/use-site-details.tsx
wojtekn: The title says it failed to initialize custom domains, but it makes sense to clarify the message. I've added that in https://github.com/Automattic/studio/pull/1089/commits/c82c9028335521850b766e5467dcf6620877fe08

src/ipc-handlers.ts
wojtekn: I removed redundant logging in https://github.com/Automattic/studio/pull/1089/commits/67bdbcfe656cb99949fd15cf9485a9446b7f0e5e and moved the hosts log message to the proper place in https://github.com/Automattic/studio/pull/1089/commits/f29028e26cc859fb28661ade2602a934030b1498

src/lib/proxy-server.ts
fredrikekelund: It looks a bit odd that we throw a new exception for known error types but just `return false` for unknown error types. The known error types indicate that port 80 is busy, but other errors also indicate that Studio failed to start the proxy server.

src/hooks/use-site-details.tsx
```diff
@@ -285,14 +285,27 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
-					title: __( 'Failed to start the site server' ),
-					message: __(
-						"Please verify your site's local path directory contains the standard WordPress installation files and try again. If this problem persists, please contact support."
-					),
-					error,
-					showOpenLogs: true,
-				} );
+				if (
+					error instanceof Error &&
+					error.message.includes( 'Studio failed to start the proxy server' )
```
fredrikekelund: If this error only indicates that port 80 is busy, I would argue we should use a more specific error message.

wojtekn: I've improved wording in e06d0622f27fa25b89763b09faf8ee43b379b019

src/lib/proxy-server.ts
wojtekn: I've updated that logic to handle error cases only through exceptions and cleaned up the logic around `false`.

src/hooks/use-site-details.tsx
wojtekn: Actually, as those are used internally to mitigate IPC handler that flattens our exceptions, we could change descriptive messages to codes e.g. `PROXY_ERROR_80` or prepend messages with the prefix e.g. `PROXY_ERROR_80: 'Studio failed to [...]`. Thoughts?

fredrikekelund: I agree we could use a message like `PROXY_ERROR_80`, since this is only used internally 👍

src/lib/proxy-server.ts
fredrikekelund: Nice improvement 👍

PR #1090, opened 2025-03-18
Add EADDRINUSE error on Windows when port 80 is already in use

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10626

## Proposed Changes

- Added Windows-specific port availability check before starting the proxy server
- Implemented a port tester using createConnection from node:net to verify if port 80 is available
- Added timeout handling (1 second) for the port check, I opted to be safe and assume that if the host is unreachable we won't be able to start a new server on that port, so an error is shown to the user.

**Note**: I couldn't understand why Windows doesn't catch the error when we create the server.
That would be the ideal scenario, but this solution should achieve the same result at a marginal cost, of doing an extra request to assert port availability.

## Testing Instructions

1. Port already in use
- Test on a Windows machine where port 80 is already in use by another application (e.g., Local with a custom domain site)
- The application should properly detect that the port is in use and handle the error

2. Port available
- Test on a Windows machine where port 80 is available
- The proxy server should start successfully
- Create and start a site with a custom domain and validate it works

3. macOS
- Test both scenarios above in macOS and confirm the same behavior

4. Tester timeout
- Adjust to code to simulate timeout scenarios. Examples:
    - unreachable host - by doing `const tester = createConnection( { port: 80, host: '240.0.0.0' }, () => { ...`
    - smaller tester timeout - `tester.setTimeout( 1, () => {`
- Both tests should result in the warning.

## Pre-merge Checklist

### DIFF
--- a/src/lib/proxy-server.ts
+++ b/src/lib/proxy-server.ts
@@ -83,24 +83,31 @@ export async function startProxyServer(): Promise< boolean > {
 		} );
 
 		// On Windows, we need to check if the port is in use before we can listen on it
-		await new Promise< void >( ( resolve, reject ) => {
-			const tester = createConnection( { port: 80 }, () => {
-				// If we can connect, port is in use
-				tester.end();
-				const error = new Error( 'Port 80 is in use' ) as NodeJS.ErrnoException;
-				error.code = 'EADDRINUSE';
-				reject( error );
-			} );
+		if ( process.platform === 'win32' ) {
+			await new Promise< void >( ( resolve, reject ) => {
+				const tester = createConnection( { port: 80 }, () => {
+					// If we can connect, port is in use
+					tester.end();
+					const error = new Error( 'Port 80 is in use' ) as NodeJS.ErrnoException;
+					error.code = 'EADDRINUSE';
+					reject( error );
+				} );
 
-			tester.on( 'error', ( err: NodeJS.ErrnoException ) => {
-				if ( err.code === 'ECONNREFUSED' ) {
-					// Port is available
-					resolve();
-				} else {
-					reject( err );
-				}
+				tester.setTimeout( 1000, () => {
+					tester.destroy();
+					reject( new Error( 'Port check timed out' ) );
+				} );
+
+				tester.on( 'error', ( err: NodeJS.ErrnoException ) => {
+					if ( err.code === 'ECONNREFUSED' ) {
+						// Port is available
+						resolve();
+					} else {
+						reject( err );
+					}
+				} );
 			} );
-		} );
+		}
 
 		await new Promise< void >( ( resolve, reject ) => {
 			proxyServer!
--- a/src/lib/proxy-server.ts
+++ b/src/lib/proxy-server.ts
@@ -1,5 +1,6 @@
 import { dialog } from 'electron';
 import http from 'http';
+import { createConnection } from 'node:net';
 import { domainToASCII } from 'node:url';
 import * as Sentry from '@sentry/electron/main';
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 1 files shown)


src/lib/proxy-server.ts
```diff
@@ -81,6 +82,33 @@ export async function startProxyServer(): Promise< boolean > {
...
+				} );
+
+				tester.setTimeout( 1000, () => {
+					tester.destroy();
+					reject( new Error( 'Port check timed out' ) );
+				} );
+
+				tester.on( 'error', ( err: NodeJS.ErrnoException ) => {
+					if ( err.code === 'ECONNREFUSED' ) {
+						// Port is available
```
```suggestion
				tester.on( 'error', ( err ) => {
					if ( isErrnoException( err ) && err.code === 'ECONNREFUSED' ) {
```
Better to adhere to the actual type of `err`

src/lib/proxy-server.ts
```diff
@@ -81,6 +82,33 @@ export async function startProxyServer(): Promise< boolean > {
...
+					// If we can connect, port is in use
+					tester.end();
+					const error = new Error( 'Port 80 is in use' ) as NodeJS.ErrnoException;
+					error.code = 'EADDRINUSE';
+					reject( error );
+				} );
+
+				tester.setTimeout( 1000, () => {
+					tester.destroy();
+					reject( new Error( 'Port check timed out' ) );
```
fredrikekelund: Are we sure that a timeout indicates that the port is busy?

src/lib/proxy-server.ts
```diff
@@ -81,6 +82,33 @@ export async function startProxyServer(): Promise< boolean > {
...
 
+		// On Windows, we need to check if the port is in use before we can listen on it
+		if ( process.platform === 'win32' ) {
+			await new Promise< void >( ( resolve, reject ) => {
+				const tester = createConnection( { port: 80 }, () => {
+					// If we can connect, port is in use
+					tester.end();
+					const error = new Error( 'Port 80 is in use' ) as NodeJS.ErrnoException;
+					error.code = 'EADDRINUSE';
+					reject( error );
```
```suggestion
					tester.end();
					reject( new Error( 'EADDRINUSE' ) );
```
`as` statements always irk me, even in relatively harmless cases like this. This is ultimately subjective, but I would advise changing this code per my suggestion and updating the catch clause to something like this:

```ts
	} catch ( error ) {
		if (
			( isErrnoException( error ) && error.code === 'EADDRINUSE' ) ||
			( error instanceof Error && error.message === 'EADDRINUSE' )
		) {
```

src/lib/proxy-server.ts
```diff
@@ -81,6 +82,33 @@ export async function startProxyServer(): Promise< boolean > {
 			} );
 		} );
 
+		// On Windows, we need to check if the port is in use before we can listen on it
```
```suggestion
		// On Windows, node doesn't throw an error if port 80 is busy, so we use the net module to explicitly check
		// if it's possible to establish a TCP connection to that port (meaning it's busy)
```
I think it's worth clarifying why we do this

src/lib/proxy-server.ts
```diff
@@ -81,6 +82,33 @@ export async function startProxyServer(): Promise< boolean > {
...
+			await new Promise< void >( ( resolve, reject ) => {
+				const tester = createConnection( { port: 80 }, () => {
+					// If we can connect, port is in use
+					tester.end();
+					const error = new Error( 'Port 80 is in use' ) as NodeJS.ErrnoException;
+					error.code = 'EADDRINUSE';
+					reject( error );
+				} );
+
+				tester.setTimeout( 1000, () => {
```
fredrikekelund, line 96: Are we sure we want to use `tester.setTimeout` here and not just `setTimeout`?

fredrikekelund (approved):
Fundamentally, this looks like a good solution and it works as expected 👍 I left a couple of comments, but they are all relatively small. Approving now to unblock

src/lib/proxy-server.ts
```diff
@@ -97,7 +125,10 @@ export async function startProxyServer(): Promise< boolean > {
 
 		return true;
 	} catch ( error ) {
-		if ( error instanceof Error && 'code' in error && error.code === 'EADDRINUSE' ) {
+		if (
+			( error instanceof Error && isErrnoException( error ) && error.code === 'EADDRINUSE' ) ||
```
```suggestion
			( isErrnoException( error ) && error.code === 'EADDRINUSE' ) ||
```
Sorry for the nitpicking, but `isErrnoException` already checks if `value instanceof Error`

bcotrim:
Thanks for all the feedback @fredrikekelund, no need to apologize for nitpicking, all feedback is appreciated! 👍 
Regarding the timeout, I think this approach is what we want.
My reasoning for throwing the error on timeout is that, if we can't get a response from the host (localhost) we most likely won't be able to create our proxy server. I think it's best to be safe, than it is to have a silent error.
What do you think?

I added a couple of tests to the PR description (smaller timeout and unreachable host), that I used to validate the timeout behavior, please let me know if something doesn't look right.

fredrikekelund:
> My reasoning for throwing the error on timeout is that, if we can't get a response from the host (localhost) we most likely won't be able to create our proxy server

Having some kind of timeout seems reasonable, I just don't know if a timeout indicates the port is busy or not… I paused to read up on the [Node net docs](https://nodejs.org/docs/latest/api/net.html), which didn't really answer that particular question. My interpretation is that `timeout` events will probably only happen if we managed to establish the connection to begin with. 

Anyway, rejecting on timeout seems fine. Let's stick with that approach.

PR #1101, opened 2025-03-20
Studio: Add base for what's new modal

katinthehatsite:
## Related issues

STU-226-linear-issue

## Proposed Changes

- TBD

## Testing Instructions

```diff --git a/src/components/main-sidebar.tsx b/src/components/main-sidebar.tsx
index 656b7a6f..d7069363 100644
--- a/src/components/main-sidebar.tsx
+++ b/src/components/main-sidebar.tsx
@@ -4,6 +4,7 @@ import { RunningSites } from 'src/components/running-sites';
 import SiteMenu from 'src/components/site-menu';
 import { isMac } from 'src/lib/app-globals';
 import { cx } from 'src/lib/cx';
+import { WhatsNewButton } from 'src/modules/whats-new-modal/whats-new-steps';
 
 interface MainSidebarProps {
        className?: string;
@@ -33,6 +34,7 @@ export default function MainSidebar( { className }: MainSidebarProps ) {
                                        <RunningSites />
                                        <div className={ cx( isMac() ? 'mx-5' : 'mx-4' ) }>
                                                <AddSite className="min-w-[168px] w-full mb-4" />
+                                               <WhatsNewButton className="min-w-[168px] w-full" />
                                        </div>
                                </div>
                        </div>
```

## Pre-merge Checklist


src/modules/whats-new-modal/whats-new-steps.tsx
```diff
@@ -0,0 +1,54 @@
...
+export function WhatsNewButton( { className }: { className?: string } ) {
+	const { __ } = useI18n();
+	const [ isModalOpen, setIsModalOpen ] = useState( false );
+
+	// Example steps for the modal
+	const steps = [
+		{
+			title: 'New Feature Heading',
+			description: 'Little paragraph sharing details of the new feature.',
+			links: [
```
ivan-ottinger, line 16: If I am understanding correctly the idea behind having the links as an array of objects is to be able to handle a situation when we would like to output not just one `Learn more` link on a particular item, but maybe share some other link as well, right?

src/modules/whats-new-modal/whats-new-steps.tsx
```diff
@@ -0,0 +1,54 @@
...
+import Button from 'src/components/button';
+import IllustrationSvg from './illustration-svg';
+import WhatsNewModal from './whats-new-modal';
+
+export function WhatsNewButton( { className }: { className?: string } ) {
+	const { __ } = useI18n();
+	const [ isModalOpen, setIsModalOpen ] = useState( false );
+
+	// Example steps for the modal
+	const steps = [
```
ivan-ottinger, line 12: I think we could consider moving the steps into a separate file in the `whats-new-modal` module.

Maybe we could have another directory inside `whats-new-modal`  that would hold all the items content - including the image resources. Just to separate logic of the modal from its contents.

src/modules/whats-new-modal/whats-new-content.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+interface WhatsNewContentProps {
+	title: string;
+	description: React.ReactNode;
+	links?: Link[];
+	illustration?: string;
+	illustrationComponent?: React.ComponentType;
+	className?: string;
+	width?: string | number;
+	height?: string | number;
```
ivan-ottinger, line 18: Generally (once we have the design), I think we could aim for keeping all the content items the same size.

Looking at the existing WordPress.com "What's New" modal, there's enough space provided in case the translations are longer in different locales:

Markup on 2025-03-21 at 11:51:42

ivan-ottinger (commented):
Nice work! I understand this is still a work-in-progress, so just wanted to share some initial thoughts below. 🙂

src/modules/whats-new-modal/whats-new-steps.tsx
katinthehatsite: Yeah possibly. I need to cleanup some things, the code is not nice yet. I mostly wanted to get some idea of how it would look. I am not sure we would need other links in there, what do you think?

src/modules/whats-new-modal/whats-new-steps.tsx
katinthehatsite: if we want to keep the content with illustrations, I think in that case, it makes sense to set them into a different directory 👍

src/modules/whats-new-modal/whats-new-steps.tsx
ivan-ottinger: Yes, I think based on the discussion in [STU-229](https://linear.app/a8c/issue/STU-229/decide-on-the-content-for-the-first-whats-new-section-items#comment-efee42c6) we can keep just one link in place.<br><br>We could still extend the logic in the future if we decide to display more links per item.

katinthehatsite:
Closing this as we changed the direction

PR #1102, opened 2025-03-20
Refactor WordPress version slice to use RTK queries

sejas:
## Related issues

- Fixes STU-222 STU-222-linear-issue

## Proposed Changes

- Re fetch WordPress versions every time we open the edit site modal

## Testing Instructions

**Edit site**

- Turn off your WI-FI and internet connection
- Run `npm start`
- Click on Settings
- Turn on your WI-FI and internet connection
- Click on Edit site
- Observe the dropdown has all the WordPress versions

https://github.com/user-attachments/assets/c4185ebd-73e2-443d-a150-cc2266b48a2e

Also test "**Add site**" and "**Onboarding flows**"

## Pre-merge Checklist


bcotrim (commented):
The implementation works well and solves the issue.
However, I think we could use [Redux Toolkit Queries](https://redux-toolkit.js.org/rtk-query/usage/queries) to build something a bit more robust.
While not using that approach is not a blocker, and the current solution is perfectly fine for the existing behavior, implementing something like this could serve as a baseline for future queries we need to add.

I was looking at this solution while working on this other similar issue: STU-197-linear-issue

What do you think?
If you agree this approach is worth looking into, I volunteer to add an initial implementation on my fix for that issue, that you could later adjust for this one as well.

fredrikekelund:
> Redux Toolkit Queries 

I'm instinctively hesitant to making this some kind of baseline because it seems like it could add a fair bit of overhead in our code. I haven't looked very closely at RTK Query, though, so maybe I'm wrong.

bcotrim:
> The implementation works well and solves the issue. However, I think we could use [Redux Toolkit Queries](https://redux-toolkit.js.org/rtk-query/usage/queries) to build something a bit more robust. While not using that approach is not a blocker, and the current solution is perfectly fine for the existing behavior, implementing something like this could serve as a baseline for future queries we need to add.
> 
> I was looking at this solution while working on this other similar issue: STU-197-linear-issue
> 
> What do you think? If you agree this approach is worth looking into, I volunteer to add an initial implementation on my fix for that issue, that you could later adjust for this one as well.

You can check the example here: https://github.com/Automattic/studio/pull/1106.

fredrikekelund:
Having reviewed https://github.com/Automattic/studio/pull/1106, I'm more positive towards using RTK Query 👍

It could probably be a good fit for this code, too, though it would make sense to work on other non-Redux logic before refactoring this, IMO.

sejas:
@fredrikekelund , @bcotrim , I've refactored this PR to use RTK Query. Let me know what you think. Open to feedback.

src/components/site-form.tsx
```diff
@@ -271,23 +267,15 @@ export const SiteForm = ( {
...
-	);
-	const wpVersionsStatus = useRootSelector( ( state ) => state.wordpressVersions.status );
 
-	useEffect( () => {
-		if ( wpVersionsStatus === 'idle' ) {
-			dispatch( wordpressVersionsThunks.fetchWordPressVersions() );
-		}
-	}, [ wpVersionsStatus, dispatch ] );
+	// Use RTK Query instead of the old slice
+	const { data: wpVersions = [] } = useGetWordPressVersionsQuery( undefined );
```
bcotrim: Nit: I don't think we need the `undefined` here.

src/stores/wordpress-versions-api.ts
```diff
@@ -0,0 +1,153 @@
...
+			! isWordPressBetaVersion( version.version ) && ! isWordPressDevVersion( version.version )
+	);
+}
+
+export const wordpressVersionsApi = createApi( {
+	reducerPath: 'wordpressVersionsApi',
+	baseQuery: fetchBaseQuery( { baseUrl: 'https://api.wordpress.org' } ),
+	endpoints: ( builder ) => ( {
+		getWordPressVersions: builder.query< WordPressVersion[], void >( {
+			query: () => ( {
```
bcotrim: Have you considered using `queryFn` instead?
That way we could do `Promise.all` with both endpoints, instead of dealing with the devResponse in the transform function.
What do you think?

src/stores/wordpress-versions-api.ts
```diff
@@ -0,0 +1,153 @@
...
+						Sentry.captureException( error );
+					}
+					throw error;
+				}
+			},
+		} ),
+	} ),
+} );
+
+export const { useGetWordPressVersionsQuery } = wordpressVersionsApi;
```
bcotrim: I would suggest using a similar wrapper to what we added in the wpcom-api, with the isOffline check, that way the query automatically runs if the connection status changes.
What do you think?

bcotrim (commented):
Works as described, nice improvement!
Added some small code suggestions.

src/stores/wordpress-versions-api.ts
fredrikekelund: +1 to this. We should not be fetching data in the `transformResponse` function, or caching won't work as expected.

sejas: Yep!, makes sense. The change was easier than I thought. Done https://github.com/Automattic/studio/pull/1102/commits/88fc5876f6150bbb0996d80d1cd3ba8cc8e7c722
Long live to RTK queries.

src/stores/wordpress-versions-api.ts
sejas: Done https://github.com/Automattic/studio/pull/1102/commits/79622b6163d5e286b84b9ecaa8b0bce2a814b9ad and https://github.com/Automattic/studio/pull/1102/commits/7b4cc2e7ee20ec2fdc5396afa3c30604974e2716
I've created a reusable hook, so it's possible to reuse it in any RTK query.

src/components/site-form.tsx
sejas: Done https://github.com/Automattic/studio/pull/1102/commits/5f75bd93cc98378ac8ba7fde5259e079d217623c

src/stores/wordpress-versions-api.ts
```diff
@@ -0,0 +1,167 @@
...
+	reducerPath: 'wordpressVersionsApi',
+	baseQuery: fetchBaseQuery(),
+	endpoints: ( builder ) => ( {
+		getWordPressVersions: builder.query< WordPressVersion[], void >( {
+			queryFn: async () => {
+				try {
+					const [ stableData, developmentData ] = await Promise.all( [
+						fetchWordPressApiData( 'beta', MINIMUM_WORDPRESS_VERSION ),
+						fetchWordPressApiData( 'development' ),
+					] );
```
fredrikekelund:
```suggestion
				try {
					// eslint-disable-next-line no-var
					var [ stableData, developmentData ] = await Promise.all( [
						fetchWordPressApiData( 'beta', MINIMUM_WORDPRESS_VERSION ),
						fetchWordPressApiData( 'development' ),
					] );
				} catch ( error ) {
					if ( error instanceof z.ZodError ) {
						Sentry.captureException( error );
					}
					throw error;
				}
```
Nit, but it'd be nice to reduce the indentation level in this function by only wrapping the fetch code in `try catch`. Doesn't have to be exactly like my suggestion, but something like this.

src/stores/wordpress-versions-api.ts
```diff
@@ -0,0 +1,167 @@
...
+	return match ? match[ 1 ] : version;
+};
+
+export interface WordPressVersion {
+	isBeta: boolean;
+	isDevelopment: boolean;
+	isLatest: boolean;
+	label: string;
+	value: string;
+}
```
fredrikekelund: Nitpicky, but I would move `WordPressOffer`, `WordPressApiOffer`, `ProcessedOffer`, and `WordPressVersion` closer to where they're used.

src/stores/tests/utils/with-offline-check.ts
fredrikekelund, line 1: Looks like this file ended up in the wrong location inside the `tests` folder

src/stores/wordpress-versions-api.ts
```diff
@@ -0,0 +1,167 @@
...
+): string {
+	if ( isWordPressDevVersion( version ) ) {
+		return 'nightly';
+	}
+	if ( isLatest ) {
+		return sprintf( __( '%s (latest)' ), version );
+	}
+	if ( occurrences > 1 || isWordPressBetaVersion( version ) ) {
+		return version;
+	}
```
fredrikekelund: This isn't exactly an issue, it's more that I haven't really understood the logic around `shortNameOccurences` and `occurrences`. Maybe it deserves a comment for clarification?

fredrikekelund (approved):
LGTM 👍

The RTK Query docs aren't as good as react-query, and the API doesn't feel entirely as intuitive either. I still think it's a good idea to use it, it's just a shame that it's not as easy to pick up 🫤

bcotrim (approved):
Changes look good! Approved 👍

src/stores/tests/utils/with-offline-check.ts
sejas: Good catch! 🫶 
I didn't realize utils was inside the tests folder. I've created a new utils folder inside stores https://github.com/Automattic/studio/pull/1102/commits/15069f27440d38952dbbcfd2ef98abaab5300730. I prefer that "hook" lives inside the stores folder, and it's not a "slice" or "api".

src/stores/wordpress-versions-api.ts
sejas: Thanks for the suggestion. Done! https://github.com/Automattic/studio/pull/1102/commits/66384622791bdb6f151f025d1029b4ab0a952f41
`processWordPressOffers` is using zod `safeParse` so I didn't include it into the try/catch.

src/stores/wordpress-versions-api.ts
sejas: Done https://github.com/Automattic/studio/pull/1102/commits/17afa2dc885750ad9e2e478ac0d9e8ab248f8b8a

src/stores/wordpress-versions-api.ts
sejas: Thanks for pointing it out. I've added some comments https://github.com/Automattic/studio/pull/1102/commits/78e3e6615b76812026a0df1d2df0219db42f4769
and renamed the parameter for consistency https://github.com/Automattic/studio/pull/1102/commits/f0efa981a196265b5d31d058e19bed5496a122b0

Let me know if it's not clear enough.

fredrikekelund: Thanks, that clarifies the logic 👍

PR #1106, opened 2025-03-20
Prevent excessive refetching of Assistant welcome messages

bcotrim:
## Related issues

- Fixes STU-197-linear-issue

## Proposed Changes

- Refactored welcome messages functionality to use [Redux Toolkit Queries](https://redux-toolkit.js.org/rtk-query/usage/queries) through wpComApi instead of a custom React context
- Added wpComApi reducer and middleware to the Redux store
- Updated auth-provider to manage WPCOM client
- Removed the use-welcome-messages hook and its provider
- Updated tests to accommodate the new implementation

**Note**: This implementation of RTK Query was a bit more complex since we need to work around the wpcom client. But it should be a one and done deal. Adding a new endpoint should be as easy as adding a new value on the `endpoints` objects, in `wpcom-api.ts`
Looking forward to your feedback

## Testing Instructions

- Launch the application and verify that welcome messages and example prompts are still displayed correctly in the Assistant Tab
- Verify that welcome messages are still fetched and displayed when:
    - The application starts
    - The user authenticates
    - The application comes back online after being offline
- Navigating away and back on the Assistant Tab should not trigger a new network request for welcome messages
- Check that all existing functionality related to welcome messages works as expected
- Run the test suite to ensure all tests pass with the new implementation

## Pre-merge Checklist


src/stores/api/wpcom-api.ts
```diff
@@ -0,0 +1,52 @@
...
+		return { data: response };
+	} catch ( error ) {
+		return {
+			error: {
+				status: 500,
+				data: error,
+			} as FetchBaseQueryError,
+		};
+	}
+};
```
fredrikekelund:
```suggestion
const wpcomBaseQuery: BaseQueryFn<
	{ path: string; apiNamespace?: string },
	unknown,
	FetchBaseQueryError
> = async ( args ) => {
	if ( ! wpcomClient?.req ) {
		const error: FetchBaseQueryError = {
			status: 401,
			data: 'Not authenticated',
		};

		return { error };
	}

	try {
		const response = await wpcomClient.req.get( args );
		return { data: response };
	} catch ( error ) {
		const fetchQueryError: FetchBaseQueryError = {
			status: 500,
			data: error,
		};

		return { error: fetchQueryError };
	}
};
```
A couple of thoughts about the base query function:
1. Let's use type definitions for the error objects instead of `as`
2. Let's use a type definition for the entire function instead of for just the args
3. Subjective nit is also that I think `wpcomClient` reads better than `wpComClient`

src/stores/api/wpcom-api.ts
```diff
@@ -0,0 +1,52 @@
...
+export const wpComApi = createApi( {
+	reducerPath: 'wpComApi',
+	baseQuery: wpComBaseQuery,
+	endpoints: ( builder ) => ( {
+		getWelcomeMessages: builder.query< WelcomeMessageResponse, void >( {
+			query: () => ( {
+				path: '/studio-app/ai-assistant/welcome',
+				apiNamespace: 'wpcom/v2',
+			} ),
+		} ),
```
fredrikekelund: I'm not sure where to do that, but we should provide a zod schema to validate the response. I guess we could either use `queryFn` or `transformResponse`

src/stores/api/wpcom-api.ts
```diff
@@ -0,0 +1,52 @@
...
+			error: {
+				status: 500,
+				data: error,
+			} as FetchBaseQueryError,
+		};
+	}
+};
+
+export const wpComApi = createApi( {
+	reducerPath: 'wpComApi',
```
fredrikekelund:
```suggestion
export const wpcomApi = createApi( {
	reducerPath: 'wpcomApi',
```
Same subjective nit 😄

src/stores/index.ts
```diff
@@ -45,12 +47,13 @@ listenerMiddleware.startListening( {
...
 	chat: chatReducer,
 	wordpressVersions: wordpressVersionsReducer,
+	[ wpComApi.reducerPath ]: wpComApi.reducer,
 } );
 
 export const store = configureStore( {
 	reducer: rootReducer,
 	middleware: ( getDefaultMiddleware ) =>
-		getDefaultMiddleware().prepend( listenerMiddleware.middleware ),
+		getDefaultMiddleware().concat( wpComApi.middleware ).prepend( listenerMiddleware.middleware ),
```
fredrikekelund:
```suggestion
		getDefaultMiddleware().prepend( listenerMiddleware.middleware, wpComApi.middleware ),
```
Is the order important here?

src/stores/index.ts
```diff
@@ -2,6 +2,7 @@ import { combineReducers, configureStore, createListenerMiddleware } from '@redu
 import { useDispatch, useSelector } from 'react-redux';
 import { LOCAL_STORAGE_CHAT_API_IDS_KEY, LOCAL_STORAGE_CHAT_MESSAGES_KEY } from 'src/constants';
 import { reducer as chatReducer } from 'src/stores/chat-slice';
+import { wpComApi } from './api/wpcom-api';
```
fredrikekelund:
```suggestion
import { wpComApi } from 'src/stores/api/wpcom-api';
```

src/stores/api/wpcom-api.ts
fredrikekelund:
```suggestion
	name: 'wpComApi',
```
`reducerPath` defaults to `name`, but `name` has no default.

src/stores/index.ts
```diff
@@ -10,6 +11,7 @@ import {
 export type RootState = {
 	chat: ReturnType< typeof chatReducer >;
 	wordpressVersions: ReturnType< typeof wordpressVersionsReducer >;
+	[ wpComApi.reducerPath ]: ReturnType< typeof wpComApi.reducer >;
```
fredrikekelund: Let's be consistent with the other reducers. I think we are mostly creating complexity by making this value dynamic instead of static, but if you feel strongly about it, then let's also change it for the other reducers.

src/components/auth-provider.tsx
```diff
@@ -51,9 +52,11 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 		}
 
 		const { token } = payload;
+		const newClient = createWpcomClient( token.accessToken, locale );
 
 		setIsAuthenticated( true );
-		setClient( createWpcomClient( token.accessToken, locale ) );
+		setClient( newClient );
+		setWpcomClient( newClient );
```
fredrikekelund, line 59: Nice. This will be very useful 👍

src/components/content-tab-assistant.tsx
```diff
@@ -345,11 +345,13 @@ export function ContentTabAssistant( { selectedSite }: ContentTabAssistantProps
 		chatSelectors.selectIsLoading( state, instanceId )
 	);
 	const { userCanSendMessage } = usePromptUsage();
-	const { messages: welcomeMessages, examplePrompts } = useWelcomeMessages();
 	const isOffline = useOffline();
 	const { __ } = useI18n();
 	const lastMessage = messages.length === 0 ? undefined : messages[ messages.length - 1 ];
 	const hasFailedMessage = messages.some( ( msg ) => msg.failedMessage );
+	const { data, isLoading } = useGetWelcomeMessagesQuery( undefined, {
+		skip: ! client || isOffline,
```
fredrikekelund: What would happen if we didn't disable the query when `client` is undefined? 🤔 Having to always pass the `skip` option for this purpose seems slightly tedious.

src/stores/api/wpcom-api.ts
fredrikekelund, line 1: I'd argue that we could store this file in `src/stores/wpcom-api.ts` instead of in a subdirectory.

src/components/content-tab-assistant.tsx
```diff
@@ -454,7 +457,7 @@ export function ContentTabAssistant( { selectedSite }: ContentTabAssistantProps
 				<div className="w-full flex flex-col items-center">
 					<AIInput
 						ref={ inputRef }
-						disabled={ disabled }
+						disabled={ disabled || isLoading }
```
fredrikekelund: Do we really need to disable the input while welcome messages are loading? 🤔

fredrikekelund (commented):
I was slightly hesitant to using RTK Query, thinking it would add overhead, but this is quite compact and I can see it being useful 👍

Functionality-wise, this is looking good. There's a difference to `trunk` in that welcome messages are now fetched when the Assistant tab is opened instead of when the application launches. It's a minuscule request that should resolve quickly, so I don't think the new behavior is a problem. Still, I wanted to mention it.

src/stores/api/wpcom-api.ts
bcotrim: I tried this and got a TS error. My implementation followed their example in the [docs](https://redux-toolkit.js.org/rtk-query/overview#create-an-api-slice)

fredrikekelund: My bad! Think I confused the `createApi` signature with `createSlice`

bcotrim:
@fredrikekelund thanks for the feedback!
I believe I addressed everything, but let me know if you see something missing or have additional feedback.

> Assistant tab is opened instead of when the application launches.

I added this intentionally, since it made sense for me for for request to live in that context.
Do you think it would be better for it to be done when the app starts?

src/components/auth-provider.tsx
sejas, line 59: Would it be possible and make sense to use only one client set?
I've noticed some duplication in the code when we set or unset it.

```ts
setClient( undefined );
setWpcomClient( undefined );
```

src/stores/wpcom-api.ts
```diff
@@ -0,0 +1,70 @@
...
+			keepUnusedDataFor: 60 * 60,
+		} ),
+	} ),
+} );
+
+function withConnectionChecks< TArg, TResult >(
+	useQueryHook: ( arg: TArg, options?: { skip?: boolean } ) => TResult
+): ( arg: TArg, options?: { skip?: boolean } ) => TResult {
+	return ( arg, options = {} ) => {
+		const isOffline = useOffline();
```
fredrikekelund, line 66: I was initially concerned about where the state stored inside this hook would live, but I realize now that it's actually fine if it's ephemeral. `navigator.onLine` is our source of truth.

src/stores/wpcom-api.ts
```diff
@@ -0,0 +1,70 @@
...
+	endpoints: ( builder ) => ( {
+		getWelcomeMessages: builder.query< WelcomeMessageResponse, void >( {
+			query: () => ( {
+				path: '/studio-app/ai-assistant/welcome',
+				apiNamespace: 'wpcom/v2',
+			} ),
+			transformResponse: ( response: unknown ) => {
+				return welcomeMessageSchema.parse( response );
+			},
+			keepUnusedDataFor: 60 * 60,
```
sejas, line 57: The cache will be 3600 seconds → 1 Hour.
I don't see the first loading state when refreshing the data

sejas (approved):
It works as described. I tested the Assistant both when logged in and logged out. I confirm that I see the welcome message request only once.

https://github.com/user-attachments/assets/4ab1b34a-f796-456e-82f2-7be7e8f5b2a4


https://github.com/user-attachments/assets/595a045c-9c3b-44b1-bfdf-45fb38593116

src/stores/wpcom-api.ts
```diff
@@ -0,0 +1,70 @@
...
+		} ),
+	} ),
+} );
+
+function withConnectionChecks< TArg, TResult >(
+	useQueryHook: ( arg: TArg, options?: { skip?: boolean } ) => TResult
+): ( arg: TArg, options?: { skip?: boolean } ) => TResult {
```
fredrikekelund:
```suggestion
function withConnectionChecks< TResult, TArg >(
	useQueryHook: TypedUseQuery< TResult, TArg, typeof wpcomBaseQuery >
): TypedUseQuery< TResult, TArg, typeof wpcomBaseQuery > {
```
Took me a frustratingly long time to think of the right type definition here, but I think this would be a more complete definition.

src/stores/wpcom-api.ts
```diff
@@ -0,0 +1,70 @@
...
+	wpcomClient = client;
+};
+
+const wpcomBaseQuery: BaseQueryFn<
+	{ path: string; apiNamespace?: string },
+	unknown,
+	FetchBaseQueryError
+> = async ( args ) => {
+	try {
+		const response = await wpcomClient!.req.get( args );
```
fredrikekelund, line 27:
```suggestion
> = async ( args, api ) => {
	const state = api.getState();
	const wpcomClient = state.auth.wpcomClient;

	try {
		const response = await wpcomClient!.req.get( args );
```
We don't have to make this change in this PR, but I realize that we should store `wpcomClient` in the Redux state to make it accessible in other slices/thunks, too.

src/components/auth-provider.tsx
fredrikekelund: We can't really access React state in RTK Query fetch functions (or Redux actions/thunks, for that matter).

We could flip it around if we stored the client as Redux state, though. So React would fetch the client from there instead of keeping a local reference. 

Storing the WPCOM client in Redux state proper is a larger task though, and (as I also mentioned elsewhere), it's something I think we could do in a follow-up PR.

src/stores/wpcom-api.ts
```diff
@@ -0,0 +1,70 @@
...
+	baseQuery: wpcomBaseQuery,
+	tagTypes: [ 'WelcomeMessages' ],
+	endpoints: ( builder ) => ( {
+		getWelcomeMessages: builder.query< WelcomeMessageResponse, void >( {
+			query: () => ( {
+				path: '/studio-app/ai-assistant/welcome',
+				apiNamespace: 'wpcom/v2',
+			} ),
+			transformResponse: ( response: unknown ) => {
+				return welcomeMessageSchema.parse( response );
```
fredrikekelund: Report to Sentry when this fails, maybe?

fredrikekelund (approved):
LGTM 👍 It'd be nice to fix the type definitions for `withConnectionChecks` before landing, but most of my other comments were suggestions or potential follow-up items.

fredrikekelund:
> Do you think it would be better for it to be done when the app starts?

I think the new behavior in this PR makes sense 👍

src/stores/wpcom-api.ts
fredrikekelund: The state is ephemeral, but the `useOffline` hook also makes it so the query automatically refetches when connectivity is restored. Very nice 👍

src/stores/wpcom-api.ts
bcotrim: Thank you for improving it, it is indeed a more complete definition! 👍

src/stores/wpcom-api.ts
bcotrim: I started to think about that, but I believe the general best practice is to not store complex objects in the Redux store.
I think there are some other alternatives we could explore, like middlewares, but it seems complex enough to warrant its own PR.

PR #1108, opened 2025-03-21
Scaffold CLI code

fredrikekelund:
## Related issues

- STU-218-linear-issue

## Proposed Changes

- Add basic dummy CLI code in `./cli`. I also included a `PreviewCreateCommand` (taken from https://github.com/Automattic/studio/pull/1055). It doesn't truly work, but it's meant to serve as a starting point for implementing the CLI commands once we get there.
- Added Webpack config for CLI code. 

> [!NOTE]
> We previously discussed (see 1eOYwX-CenpiJf5oiujB0a4T3zYiclZ09i07huE5wfYI-gdoc) whether we should bundle the CLI code. The CLI will run in a node runtime, and there's fundamentally no need to bundle JS files for node applications. However, our build pipeline doesn't include a `node_modules` folder in its output, and if we wanted the CLI to function like a regular node app with `require` calls looking up files from disk, we would need to include a `node_modules` folder specifically for the CLI in the build output. 
> 
> I don't think we need to rule out this option, but it's not immediately clear that it's a good approach. There's a risk that it would result in a (potentially significantly) larger installer bundle due to unnecessary code duplication (the Electron JS code is already bundled). So, for now, Webpack bundles all CLI code into a single file.

- Added `package.json` commands for building and watching the CLI code
- Updated the Jest config to run tests from `./cli`

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js`
3. Ensure that the file prints an overview of the CLI commands available
4. Run `npm run test cli/commands/preview`
5. Ensure that Jest can run the test successfully

## Pre-merge Checklist

### DIFF
--- a/cli/commands/base.ts
+++ b/cli/commands/base.ts
@@ -8,6 +8,8 @@ export abstract class BaseCommand {
 		this.outputFormat = outputFormat;
 	}
 
+	abstract run(): Promise< boolean >;
+
 	protected reportProgress( status: keyof typeof this.STATUSES ) {
 		if ( this.outputFormat === 'json' ) {
 			console.log( JSON.stringify( { status } ) );
--- a/cli/commands/base.ts
+++ b/cli/commands/base.ts
@@ -1,0 +1,26 @@
+export type OutputFormat = undefined | 'json';
+
+export abstract class BaseCommand {
+	protected abstract readonly STATUSES: Record< string, string >;
+	protected readonly outputFormat: OutputFormat;
+
+	constructor( outputFormat: OutputFormat ) {
+		this.outputFormat = outputFormat;
+	}
+
+	protected reportProgress( status: keyof typeof this.STATUSES ) {
+		if ( this.outputFormat === 'json' ) {
+			console.log( JSON.stringify( { status } ) );
+		} else {
+			console.log( status );
+		}
+	}
+
+	protected reportError( error: string ) {
+		if ( this.outputFormat === 'json' ) {
+			console.error( JSON.stringify( { error } ) );
+		} else {
+			console.error( error );
+		}
+	}
+}
--- a/cli/commands/preview/create.ts
+++ b/cli/commands/preview/create.ts
@@ -2,6 +2,7 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 import archiver from 'archiver';
+import { Command } from 'commander/typings';
 import fetch from 'node-fetch';
 import { BaseCommand, OutputFormat } from 'cli/commands/base';
 
@@ -20,6 +21,18 @@ export class PreviewCreateCommand extends BaseCommand {
 		super( outputFormat );
 		this.folder = folder;
 		this.archivePath = path.join( os.tmpdir(), `${ this.folder }.zip` );
+	}
+
+	static register( program: Command ) {
+		program
+			.command( 'go [folder]' )
+			.description(
+				'Start a new WordPress environment in the specified folder (defaults to current directory)'
+			)
+			.action( async ( folder: string = process.cwd(), options: { outputFormat?: 'json' } ) => {
+				const previewCreate = new PreviewCreateCommand( folder, options.outputFormat );
+				await previewCreate.run();
+			} );
 	}
 
 	async run() {
--- a/cli/commands/preview/create.ts
+++ b/cli/commands/preview/create.ts
@@ -1,0 +1,73 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import archiver from 'archiver';
+import fetch from 'node-fetch';
+import { BaseCommand, OutputFormat } from 'cli/commands/base';
+
+// This is DUMMY code for now. It's only meant as a reference for the actual implementation.
+export class PreviewCreateCommand extends BaseCommand {
+	private folder: string;
+	private archivePath: string;
+
+	protected readonly STATUSES = {
+		ARCHIVE_CREATED: 'ARCHIVE_CREATED',
+		ARCHIVE_UPLOADED: 'ARCHIVE_UPLOADED',
+		ARCHIVE_DELETED: 'ARCHIVE_DELETED',
+	};
+
+	constructor( folder: string, outputFormat: OutputFormat ) {
+		super( outputFormat );
+		this.folder = folder;
+		this.archivePath = path.join( os.tmpdir(), `${ this.folder }.zip` );
+	}
+
+	async run() {
+		await this.archiveFolder();
+		await this.uploadArchive();
+		await this.cleanup();
+		return true;
+	}
+
+	async archiveFolder() {
+		return new Promise( ( resolve, reject ) => {
+			const output = fs.createWriteStream( this.archivePath );
+
+			const archive = archiver( 'zip', {
+				zlib: { level: 9 },
+			} );
+
+			output.on( 'close', () => {
+				this.reportProgress( this.STATUSES.ARCHIVE_CREATED );
+				resolve( archive );
+			} );
+
+			archive.on( 'error', ( err: Error ) => {
+				this.reportError( err.message );
+				reject( err );
+			} );
+
+			archive.pipe( output );
+			archive.directory( `${ this.folder }/wp-content`, 'wp-content' );
+			archive.file( `${ this.folder }/wp-config.php`, { name: 'wp-config.php' } );
+
+			archive.finalize();
+		} );
+	}
+
+	async uploadArchive() {
+		const response = await fetch(
+			'https://public-api.wordpress.com/rest/v1.1/jurassic-ninja/create-new-site-from-zip',
+			{
+				method: 'POST',
+				body: fs.createReadStream( this.archivePath ),
+			}
+		);
+		this.reportProgress( this.STATUSES.ARCHIVE_UPLOADED );
+		return response.json();
+	}
+
+	async cleanup() {
+		fs.unlinkSync( this.archivePath );
+	}
+}
--- a/cli/commands/preview/create.ts
+++ b/cli/commands/preview/create.ts
@@ -4,23 +4,28 @@ import path from 'path';
 import archiver from 'archiver';
 import { Command } from 'commander/typings';
 import fetch from 'node-fetch';
-import { BaseCommand, OutputFormat } from 'cli/commands/base';
+import { Logger, OutputFormat } from 'cli/logger';
+
+enum Status {
+	ARCHIVE_CREATED = 'ARCHIVE_CREATED',
+	ARCHIVE_UPLOADED = 'ARCHIVE_UPLOADED',
+	ARCHIVE_DELETED = 'ARCHIVE_DELETED',
+}
+
+interface CommandInterface {
+	run(): Promise< boolean >;
+}
 
 // This is DUMMY code for now. It's only meant as a reference for the actual implementation.
-export class PreviewCreateCommand extends BaseCommand {
+export class PreviewCreateCommand implements CommandInterface {
 	private folder: string;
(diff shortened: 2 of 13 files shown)


cli/commands/preview/tests/create.test.ts
```diff
@@ -0,0 +1,80 @@
...
+describe( 'PreviewCreateCommand', () => {
+	let command: PreviewCreateCommand;
+	const mockFolder = '/test/folder';
+	const mockArchivePath = path.join( os.tmpdir(), `${ mockFolder }.zip` );
+
+	beforeEach( () => {
+		jest.clearAllMocks();
+		nock.cleanAll();
+		command = new PreviewCreateCommand( mockFolder, 'json' );
+	} );
```
nightnei: Works well, I am switching to code review, just initially wanted to propose mocking console.log on early stages, to keep terminal always clean:
```suggestion
	beforeEach( () => {
		jest.clearAllMocks();
		nock.cleanAll();
		command = new PreviewCreateCommand( mockFolder, 'json' );
		jest.spyOn( console, 'log' ).mockImplementation( () => {} );
	} );

	afterEach( () => {
		jest.spyOn( console, 'log' ).mockRestore();
	} );
```

fredrikekelund: This is testing dummy code. The point is mostly just to ensure that Jest tests work in the `./cli` directory.

nightnei: Makes sense 👍

nightnei (approved):
Cool, clear structure of folders and, in general, awesome scaffolding PR  👍

webpack.cli.config.ts
```diff
@@ -0,0 +1,31 @@
...
+const config: Configuration = {
+	target: 'node',
+	entry: './cli/index.ts',
+	output: {
+		filename: '[name].js',
+		path: path.join( __dirname, 'dist', 'cli' ),
+		libraryTarget: 'commonjs2',
+		chunkFormat: 'commonjs',
+	},
+	mode: process.env.NODE_ENV === 'production' ? 'production' : 'development',
```
bcotrim, line 14: I think webpack mode is used to set the `NODE_ENV` value, is that correct?
If that is true, isn't this creating a circular dependency?
Could we just set the default value, and override it in our `package.json` scripts?

cli/index.ts
```diff
@@ -0,0 +1,29 @@
+import { Command } from 'commander';
+import { PreviewCreateCommand } from 'cli/commands/preview/create';
+import { version } from '../package.json';
+
+const program = new Command();
+
+program.name( 'studio' ).description( 'WordPress.com Studio CLI' ).version( version );
```
bcotrim: Nit: Shouldn't it be "CLI for Studio by WordPress.com"?

cli/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+program
```
bcotrim, line 7: I'm worried about how well this will scale.

I think we could improve the command registration by using static properties.
We could add a static abstract `commands: Command[]` property to the `BaseCommand` class that would need to be implemented by each command class. This property would contain the command definitions (command, description, options, and action), making it easier to register commands in `index.ts` using something like `loadCommand(PreviewCreateCommand.commands)`.

I think this would scale better since each set of commands would live in its own class, instead of all commands in index.ts, making it easier to understand and grow. The abstract property would also ensure that every command class properly defines its commands.

Let me know what you think.

bcotrim (commented):
Nice work on the scaffold, overall I think this is a solid foundation for our CLI.
I added some suggestions, let me know what you think

sejas (approved):
It works as expected. I was able to run tests, create a build version `npm run cli:build` and then execute the dummy cli.
Great Start!

Screenshot 2025-03-21 at 19 28 00

Screenshot 2025-03-21 at 19 29 56

I also confirm the main `npm run test` also include the tests in CLI folder.

cli/commands/base.ts
```diff
@@ -0,0 +1,26 @@
+export type OutputFormat = undefined | 'json';
+
+export abstract class BaseCommand {
```
youknowriad: I have to admin that I'm not a big fan of classes and that I prefer typed structures (interface). But regardless, it's a style choice, I would love to understand more the choices made here:

- Why is this a class? 
- What if a command is just a function with a given interface and "reportProgress", "reportError" two helper function in a "lib/logger" or something like that?

Mostly curious to understand the vision here.

cli/commands/preview/create.ts
```diff
@@ -0,0 +1,73 @@
...
+import archiver from 'archiver';
+import fetch from 'node-fetch';
+import { BaseCommand, OutputFormat } from 'cli/commands/base';
+
+// This is DUMMY code for now. It's only meant as a reference for the actual implementation.
+export class PreviewCreateCommand extends BaseCommand {
+	private folder: string;
+	private archivePath: string;
+
+	protected readonly STATUSES = {
```
youknowriad: Why do all command needs "statuses"?

cli/commands/preview/create.ts
```diff
@@ -0,0 +1,73 @@
...
+import os from 'os';
+import path from 'path';
+import archiver from 'archiver';
+import fetch from 'node-fetch';
+import { BaseCommand, OutputFormat } from 'cli/commands/base';
+
+// This is DUMMY code for now. It's only meant as a reference for the actual implementation.
+export class PreviewCreateCommand extends BaseCommand {
+	private folder: string;
+	private archivePath: string;
```
youknowriad: For me a command is a short lived function, so I would like to understand more why you think it needs "state"?

cli/commands/preview/create.ts
```diff
@@ -0,0 +1,73 @@
...
+		ARCHIVE_DELETED: 'ARCHIVE_DELETED',
+	};
+
+	constructor( folder: string, outputFormat: OutputFormat ) {
+		super( outputFormat );
+		this.folder = folder;
+		this.archivePath = path.join( os.tmpdir(), `${ this.folder }.zip` );
+	}
+
+	async run() {
```
youknowriad: Is this something that you think should be an "abstract" in the parent class?

cli/commands/preview/tests/create.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+			const result = await command.run();
+
+			expect( result ).toBe( true );
+			expect( fs.createWriteStream ).toHaveBeenCalledWith( mockArchivePath );
+			expect( fs.unlinkSync ).toHaveBeenCalledWith( mockArchivePath );
+		} );
+	} );
+
+	describe( 'archiveFolder', () => {
```
youknowriad: I feel ideally we should be testing private functions as these are implementation details that can change, the important big being the "run" function.

cli/index.ts
```diff
@@ -0,0 +1,29 @@
...
+		( value: string ) => {
+			if ( value !== 'json' ) {
+				throw new Error( 'The only custom output format supported is "json"' );
+			}
+			return value;
+		}
+	)
+	.action( async ( folder: string = process.cwd(), options: { outputFormat?: 'json' } ) => {
+		const previewCreate = new PreviewCreateCommand( folder, options.outputFormat );
+		await previewCreate.run();
```
youknowriad: In my opinion it would be simpler to just do `await previewCommand( ...args )`. Instantiating the class and running it seems useless (at least at the moment).

webpack.cli.config.ts
```diff
@@ -0,0 +1,31 @@
...
+			cli: path.join( __dirname, 'cli' ),
+		},
+	},
+	optimization: {
+		minimize: false,
+		splitChunks: false,
+	},
+};
+
+export default config;
```
youknowriad, line 31: Discussed previously but I'm not sure we need webpack for the CLI tool. running typescript directly should be enough at the moment.

youknowriad (commented):
Great start, I think we could start simpler and avoid over-engineering from the start.

cli/commands/preview/create.ts
fredrikekelund: My ambition here was to add a baseline for printing well-structured standard output to be consumed by Studio. Having a statically defined list of known statuses seems like a good part of that strategy.

cli/commands/preview/create.ts
fredrikekelund: The meaning of "short-lived" is maybe relative, but I'm not so sure that'll be the most common case. Creating/updating preview sites, sync operations, import/export are all commands that can take several minutes to complete.

bcotrim:
I ran into an issue:
After running the testing instructions in this PR, running `npm run lint` stops working. Checking with `--debug` it looks like it is getting stuck on the `dist/cli/index.js` file.
I think could fix it by changing eslint config to ignore the cli dist folder.

cli/commands/preview/create.ts
youknowriad: > Creating/updating preview sites, sync operations, import/export are all commands that can take several minutes to complete.

True but, does it mean it needs local state in the "class". For me a class makes sense when something external is orchestrating it. Like using it now, keeping it in state and maybe use it later on some other completely unrelated flow. 

In the case of commands, it feels like it's just a single "run" function with start and end. The functions themselves can use variables to hold state and all but I don't know what the state attached to a class is for.

webpack.cli.config.ts
fredrikekelund: I discussed this in the PR description. Transpiling is enough to make it so node can run the code, but our build pipeline does not currently include a `node_modules` folder anywhere in the output, and we need that for the `require` algorithm to function normally. 

Bundling is an easy solution to this problem. The other option (that I see) is figuring out how to include a deduplicated/tree-shaken `node_modules` folder in the app output without increasing the installer size too much. 

I'm open to a different solution here down the line, but this seemed like a pragmatic approach for now.

youknowriad: 👍 You convinced me. Agreed it seems pragmatic, we can adapt later if needed.

webpack.cli.config.ts
fredrikekelund: From some quick research, it looks like you're right that this config option controls what value is assumed for `process.env.NODE_ENV` in the "compiled" code. 

Could you elaborate on how this would introduce a circular dependency problem?

cli/index.ts
fredrikekelund: I'll update it to `Studio by WordPress.com CLI` 👍

cli/index.ts
fredrikekelund: Great idea 👍 I updated the PR according to your suggestion

cli/commands/base.ts
fredrikekelund: > Why is this a class?

There are no hard blockers pushing us in the class direction. It does seemingly make it easier to keep track of `outputFormat` state and to enforce a structure for reporting statuses.

We could probably do this without too much hassle with functions, too. I'll give that a try tomorrow and see where we end up.

bcotrim (approved):
Changes look good, all code runs as described
LGTM 👍

fredrikekelund:
I've updated `commands/preview/create.ts` to structure its internal logic in functions instead of a class. Instead of using a base class to hold the logger functionality, there's now a standalone `Logger` class. Per @bcotrim's suggestion, individual commands now expose a `registerCommand` function (which is actually the only exported member).

Thanks for the reviews, everyone – I think this ended up in a good place 👍 Feel free to comment on the updated logic, but in the interest of moving forward, I will merge this as soon as CI passes.

youknowriad:
This looks good to me. I noticed the separate `package.json` which is probably the right path forward but it raises the question about whether "commander" should be a dependency there and whether we should introduce workspaces later.

PR #1111, opened 2025-03-21
studio cli boostrap files

bcotrim:
## Related issues

- Fixes STU-248-linear-issue

## Proposed Changes

- Added CLI bootstrap scripts for both Windows and Mac

### Prerequisites
1. Build and install the Studio app (we need this so the scripts exist in bin):
   ```bash
   npm run make
   ```

### Testing Steps

 1. First, create a test script at:
     -  `/Applications/Studio.app/Contents/Resources/cli-test.js` (Mac)
     - `C:\Users\<username>\AppData\Local\Programs\studio\app-1.3.8\resources` (Windows)
   ```javascript
   console.log('CLI Test');
   console.log('Arguments:', process.argv.slice(2));
   ```
2. Then modify the CLI scripts to point to the test file:

   **Mac** - Edit `/Applications/Studio.app/Contents/Resources/bin/studio-cli.sh`:
   ```bash
   # Change the CLI line to:
   CLI="$CONTENTS/Resources/cli-test.js"
   ```

   **Windows** - Edit `.\resources\bin\studio-cli.bat`:
   ```batch
   # Change the call line to:
   call "%~dp0..\cli-test.js" %*
   ```

 3. Then test with:
   ```bash
   # Mac - No arguments
   /Applications/Studio.app/Contents/Resources/bin/studio-cli.sh
   # Expected output:
   # CLI Test
   # Arguments: []

   # Mac - With arguments
   /Applications/Studio.app/Contents/Resources/bin/studio-cli.sh --test arg1 arg2
   # Expected output:
   # CLI Test
   # Arguments: [ '--test', 'arg1', 'arg2' ]

   # Windows - Similar tests using studio-cli.bat
   ```

## Pre-merge Checklist

### DIFF
--- a/bin/studio-cli.sh
+++ b/bin/studio-cli.sh
@@ -5,7 +5,7 @@ function realpath() {
   /usr/bin/perl -e "use Cwd;print Cwd::abs_path(@ARGV[0])" "$0";
 }
 
-CONTENTS="$(command dirname "$(command dirname "$(command dirname "$(command dirname "$(realpath "$0")")")")")"
+CONTENTS="$(command dirname "$(command dirname "$(command dirname "$(realpath "$0")")")")"
 BINARY_NAME="$(TERM=dumb command ls "$CONTENTS/MacOS/")"
 ELECTRON="$CONTENTS/MacOS/$BINARY_NAME"
 CLI="$CONTENTS/Resources/app/cli/index.js"
--- a/bin/studio-cli.sh
+++ b/bin/studio-cli.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# The least terrible way to resolve a symlink to its real path.
+function realpath() {
+  /usr/bin/perl -e "use Cwd;print Cwd::abs_path(@ARGV[0])" "$0";
+}
+
+CONTENTS="$(command dirname "$(command dirname "$(command dirname "$(command dirname "$(realpath "$0")")")")")"
+BINARY_NAME="$(TERM=dumb command ls "$CONTENTS/MacOS/")"
+ELECTRON="$CONTENTS/MacOS/$BINARY_NAME"
+CLI="$CONTENTS/Resources/app/cli/index.js"
+
+ELECTRON_RUN_AS_NODE=1 "$ELECTRON" "$CLI" "$@"
+
+exit $? 
--- a/bin/studio-cli.bat
+++ b/bin/studio-cli.bat
@@ -2,6 +2,6 @@
 setlocal
 
 set ELECTRON_RUN_AS_NODE=1
-call "%~dp0..\..\..\Studio.exe" "%~dp0..\cli\index.js" %*
+call "%~dp0..\..\Studio.exe" "%~dp0..\cli\index.js" %*
 
 endlocal 
--- a/bin/studio-cli.bat
+++ b/bin/studio-cli.bat
@@ -1,0 +1,7 @@
+@echo off
+setlocal
+
+set ELECTRON_RUN_AS_NODE=1
+call "%~dp0..\..\..\Studio.exe" "%~dp0..\cli\index.js" %*
+
+endlocal 


bcotrim:
@fredrikekelund @nightnei 
Note: Waiting for https://github.com/Automattic/studio/pull/1108 to be merged, so I can adjust the cli entry points.
I will update the testing instructions after I do, making them easier to execute.
In the meantime, you can validate the changes using these test instructions.
Any feedback is welcome.

bin/studio-cli.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# The least terrible way to resolve a symlink to its real path.
+function realpath() {
+  /usr/bin/perl -e "use Cwd;print Cwd::abs_path(@ARGV[0])" "$0";
+}
+
+CONTENTS="$(command dirname "$(command dirname "$(command dirname "$(realpath "$0")")")")"
+BINARY_NAME="$(TERM=dumb command ls "$CONTENTS/MacOS/")"
+ELECTRON="$CONTENTS/MacOS/$BINARY_NAME"
+CLI="$CONTENTS/Resources/app/cli/index.js"
```
fredrikekelund, line 11: Noting that I'm not sure yet what the path will be in the app output and if there are any hard blockers we need to take into account. This seems like a good start 👍

bin/studio-cli.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# The least terrible way to resolve a symlink to its real path.
+function realpath() {
+  /usr/bin/perl -e "use Cwd;print Cwd::abs_path(@ARGV[0])" "$0";
+}
+
+CONTENTS="$(command dirname "$(command dirname "$(command dirname "$(realpath "$0")")")")"
```
fredrikekelund, line 8: This does not look like "the least terrible way" to resolve three parent directories above the real path 😆 Never mind, though. Sticking with that works for GitHub Desktop seems like a way to avoid problems.

bin/studio-cli.sh
```diff
@@ -0,0 +1,15 @@
...
+BINARY_NAME="$(TERM=dumb command ls "$CONTENTS/MacOS/")"
```
fredrikekelund, line 9: I guess resolving the app name dynamically doesn't hurt. I just don't know/can't imagine when/how it would change…

fredrikekelund (approved):
Looking good and works well on MacOS 👍 I haven't tested on Windows yet, but approving now to unblock.

bin/studio-cli.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# The least terrible way to resolve a symlink to its real path.
+function realpath() {
+  /usr/bin/perl -e "use Cwd;print Cwd::abs_path(@ARGV[0])" "$0";
+}
```
mcsf, line 6: Since we're only targeting macOS, why implement `realpath`? It's not POSIX but it's available on macOS, per its BSD lineage:

```sh
which -a realpath
man 1 realpath
```

bin/studio-cli.sh
mcsf: To my knowledge there aren't any shell built-ins that would conflict with `dirname` (especially with `sh` as the shebang, which should make this more portable) so why wrap it with `command`? That is, I expected this line to simply read `$(dirname "$(dirname ...))`.

bin/studio-cli.sh
mcsf: This feels a bit dangerous. I take it that the goal is path expansion?

I was also surprised by the use of `command` (see my other comment on that — my fancy Fish shell has a `ls` built-in, but any interpreter emulating `/bin/sh` shouldn't) and `TERM=dumb`, which sounds like a different way of achieving `ls --color=never` (correct me if I'm wrong).

Regardless, I think something like `find` could be more predictable and would remove some of the steps:

```diff
-BINARY_NAME="$(TERM=dumb command ls "$CONTENTS_DIR/MacOS/")"
-ELECTRON="$CONTENTS_DIR/MacOS/$BINARY_NAME"
+ELECTRON=$(find "$CONTENTS_DIR/MacOS" -type f -maxdepth 1)
```

* * *

That said, do we ever expect the binary to be named anything other than `Studio`? If not, let's constrain the domain to our advantage. Ultimately the logic could be little more than `[ -x "$CONTENTS_DIR/MacOS/Studio" ] || exit`. Forget the first part of my comment, I think this should be the way to go:

```sh
CONTENTS_DIR=$(dirname "$(dirname "$(dirname "$(realpath "$1")")")")
ELECTRON="$CONTENTS_DIR/MacOS/Studio"
if ! [ -x "$ELECTRON" ]
then
	echo >&2 "Fatal: 'Studio' executable not found"
	exit 1
fi
```

bin/studio-cli.sh
```diff
@@ -0,0 +1,15 @@
...
+
+CONTENTS="$(command dirname "$(command dirname "$(command dirname "$(realpath "$0")")")")"
+BINARY_NAME="$(TERM=dumb command ls "$CONTENTS/MacOS/")"
+ELECTRON="$CONTENTS/MacOS/$BINARY_NAME"
+CLI="$CONTENTS/Resources/app/cli/index.js"
+
+ELECTRON_RUN_AS_NODE=1 "$ELECTRON" "$CLI" "$@"
+
+exit $?
```
mcsf, line 15: Nit-pick: this exit command is redundant, as scripts implicitly return the exit code of their last invocation. But I would go further here and propose using `exec` to _replace_ the shell with the Electron program rather than spawn it as a child process. 

Unless there are specific reasons to keep the shell running, I'd say `exec` is not only more economical but can help minimise surprises around job control [based on my experience with wp-now](https://github.com/WordPress/playground-tools/issues/220).

```sh
ELECTRON_RUN_AS_NODE=1 exec "$ELECTRON" "$CLI" "$@"
```

bin/studio-cli.sh
fredrikekelund: Your last suggestion here seems very commonsensical to me, @mcsf. I can't imagine when or why we would ever change the executable name, but since this script is packaged with each Studio release, we could theoretically just change the name when needed. 

I'll implement this in an upcoming PR together with tweaks based on your feedback in https://github.com/Automattic/studio/pull/1132.

bin/studio-cli.sh
fredrikekelund: Makes sense to me 👍

bin/studio-cli.sh
fredrikekelund: Interesting. Using `exec` seems like the way to go 👍 Good catch.

We expect the CLI will run more and more long-running commands, so if not in the first iteration, I'm sure someone would eventually have run into job control issues similar to what you described.

Issue #420, opened 2024-08-04
Feature Request: Access and edit the database


mikeott:
### What

It would be great if we could access the database in a browser, similar to PHPmyAdmin and alike.

### Why

From a developers perspective, not having convenient access to the database to make quick edits or confirmations is a deal-breaker.

babakfp:
WordPress Studio is better than alternatives like XAMPP, WampServer and Laragon. People and myself face many issues and annoying things using those apps. Having the DB view/add/edit/delete feature will eliminate the need for those other tools for so many people.

ramonjd:
Related: https://github.com/Automattic/studio/issues/1033

babakfp:
It just crossed my mind, could we just use whatever drizzle uses, [Drizzle Studio](https://orm.drizzle.team/drizzle-studio/overview)?

I thought it's free and open-source, but it seems like it's a paid product https://github.com/drizzle-team/drizzle-studio-npm?tab=readme-ov-file#pricing.

ramonjd:
I think Drizzle ORM is free. It would be interesting to test its SQLite integration:

https://orm.drizzle.team/docs/get-started-sqlite

I was testing with better-sqlite3 but I'd also like to give libSQL a crack.

wojtekn:
It's almost there: https://github.com/Automattic/studio/pull/2834 🥳

wojtekn:
Closing - phpMyAdmin is included in 1.7.7.

PR #1112, opened 2025-03-22
Add database management features with better-sqlite3 integration

ramonjd:
A hack to implement a simple table editor using https://github.com/WiseLibs/better-sqlite3

See https://github.com/Automattic/studio/issues/420

## TODO

- [ ] state management
	- use a single state for the table data, combine related state.
	- use reducers for complex state updates.
	- extract modal logic
	- create custom hooks for table fetching/table pagination
	- error state management
- [ ] How to accommodate very large tables such as `wp_posts`? Multi windows? Toggle column visibility? 
- [ ] cache table results ??
- [ ] Query and form validation
- [ ] combine loading states
- [ ] use memo for expensive operations
- [ ] Add drop row actions (batch and single)
- [ ] Shift + click to select multiple rows?
- [ ] Add custom query field
- [ ] Sorting and column selection view
- [ ]  Use Dataviews to render tables with sorting, column selection, pagination ??

https://github.com/user-attachments/assets/885377d3-15fd-41bc-84b5-c8819a30a977

## Related issues

- Fixes #

## Proposed Changes

-

## Testing Instructions

-

## Pre-merge Checklist


src/lib/database/database-manager.ts
```diff
@@ -0,0 +1,100 @@
+import Database from 'better-sqlite3';
+
+export default class DatabaseManager {
```
ramonjd: Not really required, but leaving as an example of abstraction during dev. I'll excise it, or reduce its scope later.

ramonjd:
Going to park this and explore an alternative way to allow users to edit the database, namely, a bundled (already-written) application in Studio that runs in a seperate node server process. Users will be able to open it in a browser.

PR #1114, opened 2025-03-24
Validate custom domain name availability

fredrikekelund:
## Related issues

- Fixes STU-253-linear-issue

## Proposed Changes

- Validate custom domain name availability when adding or editing a site

## Testing Instructions

1. Have a site with a custom domain name
2. Attempt to add a new site with the same custom domain name
3. Ensure that there's a validation error preventing you from adding the new site
4. Edit an existing site without a custom domain name
5. Attempt to use the custom domain name from the first site
6. Ensure that there's a validation error preventing you from submitting the form
7. Attempt to edit the site with the custom domain
8. Ensure that the editing modal doesn't open with a validation error

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-add-site.ts
+++ b/src/hooks/use-add-site.ts
@@ -1,9 +1,9 @@
 import * as Sentry from '@sentry/electron/renderer';
 import { useI18n } from '@wordpress/react-i18n';
-import { useCallback, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useImportExport } from 'src/hooks/use-import-export';
 import { useSiteDetails } from 'src/hooks/use-site-details';
-import { generateCustomDomainFromSiteName, validateDomainName } from 'src/lib/domains';
+import { generateCustomDomainFromSiteName, getDomainNameValidationError } from 'src/lib/domains';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import {
 	DEFAULT_PHP_VERSION,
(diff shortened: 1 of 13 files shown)


src/hooks/use-add-site.ts
```diff
@@ -28,8 +28,20 @@ export function useAddSite() {
 	const [ useCustomDomain, setUseCustomDomain ] = useState( false );
 	const [ customDomain, setCustomDomain ] = useState< string | null >( null );
 	const [ customDomainError, setCustomDomainError ] = useState( '' );
+	const [ existingDomainNames, setExistingDomainNames ] = useState< string[] >( [] );
 	const [ enableHttps, setEnableHttps ] = useState( false );
 
+	useEffect( () => {
```
bcotrim: This is only loading the domains once.
So if you create a site with a custom domain and, without restarting the app, try to create another one using the same domain, it will allow it.

src/lib/domains.ts
```diff
@@ -10,26 +10,34 @@ export const generateCustomDomainFromSiteName = ( siteName: string ): string =>
 	return `${ domainBase }.wp.local`;
 };
 
-export const validateDomainName = (
+export const getDomainNameValidationError = (
```
wojtekn, line 13: Nice rename and validation improvements.

bcotrim (changes requested):
Nice improvement! 👍 

I don't like how we need to fetch the userData for each validation. It's probably outside of the scope of this issue, but now that we have Redux, I think it would be great if we could make our userData sync with a Redux store, fixing all issues similar to this one.

I added a comment with an issue I've found. Besides that I think the code is good to go, if we want to tackle the Redux change in the future.

wojtekn (approved):
Works as expected. I get correct validation error in both Add site and Edit site forms.

src/hooks/use-add-site.ts
fredrikekelund: Good catch, thanks 👍

fredrikekelund:
> it would be great if we could make our userData sync with a Redux store

Agreed 👍 We could probably remove a good chunk of the IPC methods if we did this and reduce the overall complexity. But yeah, it's definitely outside the scope of this PR

bcotrim (approved):
Tested and confirmed the issue no longer happens, thanks for making the change!
LGTM 👍

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -54,7 +54,23 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
...
+			.then( ( domains ) => {
+				const domainsWithoutSelectedSite = domains.filter(
+					( domain ) => domain !== selectedSite?.customDomain
+				);
+				setExistingDomainNames( domainsWithoutSelectedSite );
+			} )
+			.catch( () => {
+				// Do nothing
+			} );
+	}, [ selectedSite?.customDomain ] );
```
youknowriad, line 72: It would be cool to extract this code into a `useAvailableCustomDomains()` hook to improve readability.

PR #1116, opened 2025-03-24
Site updates: Regenerate domain certificate when the domain is changed

youknowriad:
Related STU-19-linear-issue

## Proposed Changes

This PR fixes a couple small bugs and makes a small related refactoring:

- Bug 1: When changing the custom domain name for a given site with HTTPs enabled, the certificate is regenerated properly for that site.
- Bug 2: Avoid lingering certificates when sites are removed.
- Refactoring: The logic to updates hosts file and certificates was sometimes implemented in ipc-handlers and sometimes in the server class, this moves all the logic to the server class.

## Testing Instructions

- Create a site
- Run the server
- Change custom domain
- Enable/disable HTTPs
- Rename the custom domain

On each step, the site should function properly (assuming the root certificate is trusted)
- Also removing the site, removes the certificate from the file system.


ivan-ottinger (approved):
The proposed changes look good and work as expected. 👌🏼 

- changing and renaming domain works correctly
- disabling HTTPS or removing a site removes the related certificate from the file system
- I did not observe any regressions

PR #1117, opened 2025-03-24
SSL Support: Hide all HTTPS help messages when the certificate is trusted

### LINKED ISSUE #1109: Feature Request: Handle already trusted Studio certificate.
### What

In Automattic/studio#1034 we're introducing SSL support in Studio. But we can make the experience better by detecting whether the Studio CA certificate is trusted or not.

- [X] function to detect whenever the CA certificate is trusted or not.
- [X] Hide all the links to the docs and certificate when it's already trusted.
- [X] Enable SSL by default when you choose a custom domain if the certificate is already trusted.
- [ ] Display a dialog after site creation / edition when the CA certificate is not trusted

### Why

Improve the experience

### How

*No response*

youknowriad:
## Related issues

- Fixes #1109
- Fixes STU-255-linear-issue

## Proposed Changes

- Implements a function to check whether the CA certificate is trusted.
- Hide all the SSL help texts when the CA certificate is trusted.
- Auto enable https when the custom domain is enabled and the root certificate is trusted

## Testing Instructions

- Create an SSL site
- If your CA certificate is not trusted, you should see help text on the creation form and on the settings tab

CleanShot 2025-05-01 at 14 59 59@2x

- If your CA certificate is trusted, you should not see either of these.

### DIFF
--- a/src/components/content-tab-settings.tsx
+++ b/src/components/content-tab-settings.tsx
@@ -4,6 +4,7 @@ import { useI18n } from '@wordpress/react-i18n';
 import { PropsWithChildren } from 'react';
 import { CopyTextButton } from 'src/components/copy-text-button';
 import DeleteSite from 'src/components/delete-site';
+import { useCertificateTrust } from 'src/hooks/use-certificate-trust';
 import { useGetWpVersion } from 'src/hooks/use-get-wp-version';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { decodePassword } from 'src/lib/passwords';
@@ -26,6 +27,7 @@ function SettingsRow( { children, label }: PropsWithChildren< { label: string } 
 
 export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) {
 	const { __ } = useI18n();
+	const isCertificateTrusted = useCertificateTrust();
 	const username = 'admin';
 	// Empty strings account for legacy sites lacking a stored password.
 	const storedPassword = decodePassword( selectedSite.adminPassword ?? '' );
@@ -75,13 +77,13 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) 
 					<SettingsRow label={ __( 'HTTPS' ) }>
 						<div>
 							<span>{ selectedSite.enableHttps ? __( 'Enabled' ) : __( 'Disabled' ) }</span>{ ' ' }
-							{ selectedSite.enableHttps && (
+							{ ! isCertificateTrusted && selectedSite.enableHttps && (
 								<Button variant="link" onClick={ () => getIpcApi().openCertificate() }>
 									{ __( 'Trust Certificate' ) }
 								</Button>
 							) }
 						</div>
-						{ selectedSite.enableHttps && (
+						{ ! isCertificateTrusted && selectedSite.enableHttps && (
 							<div className="mt-1 max-w-96">
 								<span className="text-a8c-gray-50 mt-1">
 									{ __(
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -13,7 +13,6 @@ import { ACCEPTED_IMPORT_FILE_TYPES } from 'src/constants';
 import { useCertificateTrust } from 'src/hooks/use-certificate-trust';
 import { useI18nData } from 'src/hooks/use-i18n-data';
 import { useOffline } from 'src/hooks/use-offline';
-import { isWindows } from 'src/lib/app-globals';
 import { cx } from 'src/lib/cx';
 import { generateCustomDomainFromSiteName } from 'src/lib/domains';
 import { getIpcApi } from 'src/lib/get-ipc-api';
@@ -468,27 +467,24 @@ export const SiteForm = ( {
 											</div>
 										) }
 
-										{ ! isCertificateTrusted &&
-											! isWindows() &&
-											useCustomDomain &&
-											setEnableHttps && (
-												<div className="text-a8c-gray-50 text-xs mt-2">
-													{ __(
-														'You need to manually add the Studio root certificate authority to your keychain and trust it to enable HTTPS.'
-													) }{ ' ' }
-													<Button
-														variant="link"
-														onClick={ () => {
-															getIpcApi().openURL(
-																'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/'
-															);
-														} }
-													>
-														{ __( 'Learn how' ) }
-														<span aria-label={ __( '(opens in a web browser)' ) }>&#8599;</span>
-													</Button>
-												</div>
-											) }
+										{ ! isCertificateTrusted && useCustomDomain && setEnableHttps && (
+											<div className="text-a8c-gray-50 text-xs mt-2">
+												{ __(
+													'You need to manually add the Studio root certificate authority to your keychain and trust it to enable HTTPS.'
+												) }{ ' ' }
+												<Button
+													variant="link"
+													onClick={ () => {
+														getIpcApi().openURL(
+															'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/'
+														);
+													} }
+												>
+													{ __( 'Learn how' ) }
+													<span aria-label={ __( '(opens in a web browser)' ) }>&#8599;</span>
+												</Button>
+											</div>
+										) }
 									</div>
 								</div>
 							</>
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -4,16 +4,18 @@ import { __, sprintf, _n } from '@wordpress/i18n';
 import { tip, warning, trash, chevronRight, chevronDown, chevronLeft } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
 import { FormEvent, useRef, useState } from 'react';
+import { useDocsLink } from 'src/hooks/use-docs-link';
 import Button from 'src/components/button';
 import FolderIcon from 'src/components/folder-icon';
 import TextControlComponent from 'src/components/text-control';
 import { WPVersionSelector } from 'src/components/wp-version-selector';
 import { ACCEPTED_IMPORT_FILE_TYPES } from 'src/constants';
+import { useCertificateTrust } from 'src/hooks/use-certificate-trust';
 import { useI18nData } from 'src/hooks/use-i18n-data';
+import { useOffline } from 'src/hooks/use-offline';
 import { isWindows } from 'src/lib/app-globals';
 import { cx } from 'src/lib/cx';
 import { generateCustomDomainFromSiteName } from 'src/lib/domains';
-import { getDocsLink } from 'src/lib/get-docs-link';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import {
 	DEFAULT_WORDPRESS_VERSION,
@@ -265,12 +267,21 @@ export const SiteForm = ( {
 }: SiteFormProps ) => {
 	const { __, isRTL } = useI18n();
 	const { locale } = useI18nData();
+	const getDocsLink = useDocsLink();
+	const isOffline = useOffline();
+	const isCertificateTrusted = useCertificateTrust();
 
 	const shouldShowCustomDomainError = useCustomDomain && customDomainError;
 	const errorCount = [ error, shouldShowCustomDomainError ].filter( Boolean ).length;
 
 	const [ isAdvancedSettingsVisible, setAdvancedSettingsVisible ] = useState( false );
-
+	/*
+	useEffect( () => {
+		if ( useCustomDomain && enableHttps && ! isCertificateTrusted && ! isWindows() ) {
+			getIpcApi().openCertificate();
+		}
+	}, [ useCustomDomain, enableHttps, isCertificateTrusted ] );
+*/
 	const handleAdvancedSettingsClick = () => {
 		setAdvancedSettingsVisible( ! isAdvancedSettingsVisible );
 	};
@@ -463,24 +474,27 @@ export const SiteForm = ( {
 											</div>
 										) }
 
-										{ ! isWindows() && useCustomDomain && setEnableHttps && (
-											<div className="text-a8c-gray-50 text-xs mt-2">
-												{ __(
-													'You need to manually add the Studio root certificate authority to your keychain and trust it to enable HTTPS.'
-												) }{ ' ' }
-												<Button
-													variant="link"
-													onClick={ () => {
-														getIpcApi().openURL(
-															'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/'
-														);
-													} }
-												>
-													{ __( 'Learn how' ) }
-													<span aria-label={ __( '(opens in a web browser)' ) }>&#8599;</span>
-												</Button>
-											</div>
-										) }
+										{ ! isCertificateTrusted &&
+											! isWindows() &&
+											useCustomDomain &&
+											setEnableHttps && (
+												<div className="text-a8c-gray-50 text-xs mt-2">
+													{ __(
+														'You need to manually add the Studio root certificate authority to your keychain and trust it to enable HTTPS.'
+													) }{ ' ' }
+													<Button
+														variant="link"
+														onClick={ () => {
+															getIpcApi().openURL(
+																'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/'
+															);
+														} }
+													>
+														{ __( 'Learn how' ) }
+														<span aria-label={ __( '(opens in a web browser)' ) }>&#8599;</span>
+													</Button>
+												</div>
+											) }
 									</div>
 								</div>
 							</>
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -3,7 +3,7 @@ import { createInterpolateElement } from '@wordpress/element';
 import { __, sprintf, _n } from '@wordpress/i18n';
 import { tip, warning, trash, chevronRight, chevronDown, chevronLeft } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
-import { FormEvent, useRef, useState } from 'react';
+import { FormEvent, useRef, useState, useEffect } from 'react';
 import Button from 'src/components/button';
 import FolderIcon from 'src/components/folder-icon';
 import TextControlComponent from 'src/components/text-control';
@@ -266,6 +266,13 @@ export const SiteForm = ( {
 	const { __, isRTL } = useI18n();
 	const { locale } = useI18nData();
 	const isCertificateTrusted = useCertificateTrust();
+
+	// If the custom domain is enabled and the root certificate is trusted, enable HTTPS
+	useEffect( () => {
+		if ( useCustomDomain && isCertificateTrusted && setEnableHttps ) {
+			setEnableHttps( true );
+		}
+	}, [ useCustomDomain, isCertificateTrusted, setEnableHttps ] );
 
 	const shouldShowCustomDomainError = useCustomDomain && customDomainError;
 	const errorCount = [ error, shouldShowCustomDomainError ].filter( Boolean ).length;
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -275,13 +275,7 @@ export const SiteForm = ( {
 	const errorCount = [ error, shouldShowCustomDomainError ].filter( Boolean ).length;
 
 	const [ isAdvancedSettingsVisible, setAdvancedSettingsVisible ] = useState( false );
-	/*
-	useEffect( () => {
-		if ( useCustomDomain && enableHttps && ! isCertificateTrusted && ! isWindows() ) {
-			getIpcApi().openCertificate();
-		}
-	}, [ useCustomDomain, enableHttps, isCertificateTrusted ] );
-*/
+
 	const handleAdvancedSettingsClick = () => {
 		setAdvancedSettingsVisible( ! isAdvancedSettingsVisible );
 	};
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -11,7 +11,6 @@ import { WPVersionSelector } from 'src/components/wp-version-selector';
 import { ACCEPTED_IMPORT_FILE_TYPES } from 'src/constants';
 import { useCertificateTrust } from 'src/hooks/use-certificate-trust';
 import { useI18nData } from 'src/hooks/use-i18n-data';
-import { useOffline } from 'src/hooks/use-offline';
 import { cx } from 'src/lib/cx';
 import { generateCustomDomainFromSiteName } from 'src/lib/domains';
 import { getDocsLink } from 'src/lib/get-docs-link';
@@ -266,7 +265,6 @@ export const SiteForm = ( {
 }: SiteFormProps ) => {
 	const { __, isRTL } = useI18n();
 	const { locale } = useI18nData();
-	const isOffline = useOffline();
 	const isCertificateTrusted = useCertificateTrust();
 
 	const shouldShowCustomDomainError = useCustomDomain && customDomainError;
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -4,7 +4,6 @@ import { __, sprintf, _n } from '@wordpress/i18n';
 import { tip, warning, trash, chevronRight, chevronDown, chevronLeft } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
 import { FormEvent, useRef, useState } from 'react';
-import { useDocsLink } from 'src/hooks/use-docs-link';
 import Button from 'src/components/button';
 import FolderIcon from 'src/components/folder-icon';
 import TextControlComponent from 'src/components/text-control';
@@ -15,6 +14,7 @@ import { useI18nData } from 'src/hooks/use-i18n-data';
 import { useOffline } from 'src/hooks/use-offline';
 import { cx } from 'src/lib/cx';
 import { generateCustomDomainFromSiteName } from 'src/lib/domains';
+import { getDocsLink } from 'src/lib/get-docs-link';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import {
 	DEFAULT_WORDPRESS_VERSION,
@@ -266,7 +266,6 @@ export const SiteForm = ( {
 }: SiteFormProps ) => {
 	const { __, isRTL } = useI18n();
 	const { locale } = useI18nData();
-	const getDocsLink = useDocsLink();
 	const isOffline = useOffline();
 	const isCertificateTrusted = useCertificateTrust();
 
--- a/src/hooks/use-certificate-trust.ts
+++ b/src/hooks/use-certificate-trust.ts
@@ -1,0 +1,33 @@
+import { useEffect, useState } from 'react';
+import { getIpcApi } from 'src/lib/get-ipc-api';
+
+/**
+ * Custom hook that checks if the Studio CA certificate is trusted on the system
+ * @returns A boolean indicating if the certificate is trusted
+ */
+export function useCertificateTrust(): boolean {
+	const [ isTrusted, setIsTrusted ] = useState< boolean >( false );
+
+	useEffect( () => {
+		let isMounted = true;
+
+		const checkCertificateTrust = async () => {
+			try {
+				const trusted = await getIpcApi().isCATrusted();
+				if ( isMounted ) {
+					setIsTrusted( trusted );
+				}
+			} catch ( error ) {
+				console.error( 'Error checking certificate trust status:', error );
+			}
+		};
+
+		checkCertificateTrust();
+
+		return () => {
+			isMounted = false;
+		};
+	}, [] );
+
+	return isTrusted;
+}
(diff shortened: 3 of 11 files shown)


youknowriad:
I didn't not a add a dialog because the existing help text in the form is enough IMO.

src/components/site-form.tsx
```diff
@@ -290,7 +292,13 @@ export const SiteForm = ( {
...
 
 	const [ isAdvancedSettingsVisible, setAdvancedSettingsVisible ] = useState( false );
-
+	/*
+	useEffect( () => {
+		if ( useCustomDomain && enableHttps && ! isCertificateTrusted && ! isWindows() ) {
+			getIpcApi().openCertificate();
+		}
+	}, [ useCustomDomain, enableHttps, isCertificateTrusted ] );
+*/
```
bgrgicak: Should we remove this?
If not, let's add a comment to explain why it's needed.

bgrgicak (commented):
It works, but the component didn't update until I switched to another tab and back.
It's not a big deal, but it would be nice if we would periodically check in the background, or add a button so the user can check manually.

src/lib/certificate-manager.ts
```diff
@@ -148,11 +152,51 @@ export async function openCertificate() {
...
+	if ( platform === 'win32' ) {
+		try {
+			const { stdout } = await execFilePromise( 'certutil', [ '-store', 'ROOT', CA_NAME ] );
+			return stdout.includes( CA_NAME );
+		} catch ( error ) {
+			return false;
+		}
+	} else if ( platform === 'darwin' ) {
+		try {
+			const { stdout } = await execFilePromise( 'security', [
```
wojtekn: I had a certificate that I had already trusted, so I marked it as untrusted in Key Chain to test it. The UI didn't display the trust helpers. When I removed the certificate from the Key Chain completely, the UI worked fine.

This command seems to check if the certificate exists. Should we test if it is trusted using the `security verify-cert` command?

This flow could be common:
1. User clicks "Trust Certificate"
2. User opens certificate so it's added to Key Chain

At this point, certificate exists, but is not trusted, so UI text disappears.

src/components/site-form.tsx
youknowriad: Yes, sorry leftover :)

src/components/site-form.tsx
```diff
@@ -505,24 +513,27 @@ export const SiteForm = ( {
...
+													<Button
+														variant="link"
+														onClick={ () => {
+															getIpcApi().openURL(
+																'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/'
+															);
+														} }
+													>
+														{ __( 'Learn how' ) }
+														<span aria-label={ __( '(opens in a web browser)' ) }>&#8599;</span>
```
sejas: There should be a space between Learn how and the arrow. Same would apply to the Learn how in edit site overview. I've observed that we are not using the arrow to express external link in other "learn more" links, in my opinion the arrow make sense though.

Screenshot 2025-03-25 at 11 38 58

Figma design ( RToz6tIuQ7nlZrikBte4GU-fi-6924_4203 ):
Screenshot 2025-03-25 at 11 44 47

sejas (commented):
I tested it on macOS, and it works almost as expected. Great work!
It would be great, as Bero mentioned, to "refresh" the state when going back to Studio. You can use `useWindowListener('focus', yourFunction);` to trigger a call when the user focuses back to Studio.
As Wojtek said, "trusted/enabled" with the current implementation means "installed" in the keychain but not necessarily trusted by the user.

I was able to install and trust a certificate and after changing tabs I confirm the message was gone. 👍 .

**New site**

| No CA | CA Installed |
|--------|--------|
| add-site-show | add-site-hide |

**Edit site**

| No CA | CA Installed |
|--------|--------|
| edit-site-show | edit-site-hide |





While testing this PR I noticed some edge cases that are worth mentioning in case they are not already tracked in issues:
1. When editing a site > unchecking the HTTPS and saving does nothing.
2. While saving the edit site modal, every checkbox and input is "disabled" except "Enable HTTPS". Allowing the user to change the checkbox state when the saving is in process.
3. The documentation (p2gHKz-wsF-p2) mentions I should click on `Open the Help → Trust Certificate menu option in WordPress Studio` in order to trust the certificate. But actually that button was in Site settings.
4. When moving from custom domain to default domain, the links in my site point to the old URL, breaking the styles.

**Trust certificate**


https://github.com/user-attachments/assets/a186cac0-114a-408b-9b30-354e8c522f76



**Uncheck Enable HTTPS**

https://github.com/user-attachments/assets/64a7c315-a461-4f46-bed6-ba55b767d236

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -90,8 +92,8 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
 		setCustomDomain( selectedSite.customDomain ?? null );
 		setCustomDomainError( '' );
 		setIsChangeWpError( '' );
-		setEnableHttps( selectedSite.enableHttps ?? false );
-	}, [ currentWpVersion, selectedSite ] );
+		setEnableHttps( selectedSite.enableHttps ?? isCertificateTrusted );
```
fredrikekelund: This change makes it so the HTTPS checkbox is enabled by default when I open the `Edit site` form for a site with a custom domain without HTTPS.

Not sure if that's intentional, but it seems like unexpected behavior to me.

src/lib/certificate-manager.ts
fredrikekelund: +1. This is key.

src/lib/certificate-manager.ts
```diff
@@ -148,11 +152,51 @@ export async function openCertificate() {
...
+ */
+export async function isRootCATrusted(): Promise< boolean > {
+	if ( ! fs.existsSync( CA_CERT_PATH ) ) {
+		return false;
+	}
+
+	const platform = process.platform;
+	if ( platform === 'win32' ) {
+		try {
+			const { stdout } = await execFilePromise( 'certutil', [ '-store', 'ROOT', CA_NAME ] );
```
fredrikekelund: Similar to @wojtekn's comment about the `security` command on macOS, I would use something like `certutil -verify ${ CERT_PATH }` here and look for the following output (or parts of it):

```
Verified Issuance Policies: All
Verified Application Policies:
    1.3.6.1.5.5.7.3.1 Server Authentication
    1.3.6.1.5.5.7.3.2 Client Authentication
Cert is a CA certificate
Cannot check leaf certificate revocation status
CertUtil: -verify command completed successfully.
```

When the certificate is on the keychain but not fully trusted, I get the following output:

```
Verified Issuance Policies: All
Verified Application Policies: None
Cert is a CA certificate
Cannot check leaf certificate revocation status
CertUtil: -verify command completed successfully.
```

And when it's not on the keychain, I get the following output:

```
Verifies against UNTRUSTED root
Cert is a CA certificate
Cannot check leaf certificate revocation status
CertUtil: -verify command completed successfully.
```

src/lib/certificate-manager.ts
```diff
@@ -148,11 +152,51 @@ export async function openCertificate() {
...
+ * Checks if the root CA certificate is already trusted by the system
+ * @returns A promise that resolves to true if the certificate is trusted, false otherwise
+ */
+export async function isRootCATrusted(): Promise< boolean > {
+	if ( ! fs.existsSync( CA_CERT_PATH ) ) {
+		return false;
+	}
+
+	const platform = process.platform;
+	if ( platform === 'win32' ) {
```
```suggestion
	if ( process.platform === 'win32' ) {
```
Nit, but I don't see why we assign this value to a variable.

youknowriad: I see that you like having the minimum number of lines of code :)

src/lib/certificate-manager.ts
youknowriad: Completely right, I don't know how I missed that in my tests :)

youknowriad: It should be fixed at least for MacOS

src/lib/certificate-manager.ts
youknowriad: I'm going to need your help here. I did make changes for windows, but I'm not sure relying on the output string is good enough (apparently the output might differ), I'm currently relying on the exit code only but I'm also not sure whether that's sufficient, I'd appreciate some testing :)

src/lib/certificate-manager.ts
wojtekn: It works like a charm - it changes state as soon as I update it in Keychain Access and Focus Studio. Thanks for fixing that!

src/lib/certificate-manager.ts
fredrikekelund: > apparently the output might differ

I understand if you have a general concern regarding the output structure, but I'm still curious if you verified this assumption in testing or in documentation?

> exit code

The exit code for `certutil -verify` doesn't seem to change depending on the certificate trust status in my testing.

I looked closely at the [PowerShell PKI module](https://learn.microsoft.com/en-us/powershell/module/pki/?view=windowsserver2025-ps), which, in some ways, seems like a more scripting-friendly option. However, I was not able to get the correct output when the certificate is on the keychain, but not properly trusted (open the `certmgr` program > Trusted Root Certification Authorities > Certificates > WordPress Studio CA > Properties > Purposes > Disable all purposes). This might seem like an edge case, but I don't have full confidence in the PowerShell approach since it doesn't report this correctly.

Screenshot 2025-03-26 at 10 44 11

fredrikekelund: I would love to suggest that we employ a different strategy: request a local HTTPS URL and check for HTTPS errors. This would work cross-platform and seems more bulletproof, but it also requires the proxy server to be started (we recently delayed proxy server upstart in https://github.com/Automattic/studio/pull/1089).

I'm not sure this is better (or that it would work, for that matter), but I wanted to raise it nonetheless. Have you considered this option, @youknowriad?

youknowriad: I didn't not consider it, I'm happy to try it but it would mean reverting the lazy start for the proxy.

fredrikekelund: Yeah. It's either that or we don't report the certificate trust status until at least one server is started.

fredrikekelund: It would be interesting to at least see if that approach works. The cross-platform nature is appealing, and checking a network response for errors is seemingly more of an E2E solution.

youknowriad: Ok, so I've explored this a bit more and there's a major downside to the HTTPs solution, it requires a url/domain, so it's not something that we can use on the site creation/edition form properly. It would mostly be valuable for the ContentTabSettings component.

Potentially we could rely use a temporary domain that is always available but it would require modifying /etc/hosts on load as well.

I'm starting to think that it has too much complexity for little gain.

sejas: @youknowriad , have we considered using [\*.localhost domains](https://datatracker.ietf.org/doc/html/rfc6761#section-6.3) (e.g. [http://whatever.localhost](http://whatever.localhost))? They should work out of the box without modifying /etc/hosts if that's the concern. We would still need to create the certificates, though.<br>I think having a default domain assigned for new Studio sites would be awesome.

youknowriad: That's a big change that I believe should be considered separately if needed:

 - our current domain validation doesn't allow localhost
 - we would have to regenerate the root certificate for existing users because the existing one can't sign localhost domains. (So users will have to trust the certificate again)

fredrikekelund: Right… I guess in a perfect world, we'd have a Studio domain that's not tied to a site, which we could use for HTTPS testing. Seems like a big undertaking, though. Seems pragmatic to follow your original approach (and maybe accept some shortcomings).

I'm not sure if the PKI PowerShell module or `certutil` is the better approach, but I'll share my findings from yesterday when playing around with PowerShell:

The `Test-Certificate` cmdlet ([docs](https://learn.microsoft.com/en-us/powershell/module/pki/test-certificate?view=windowsserver2025-ps)) checks if the certificate was properly installed. The only gotcha is that we can't pass the certificate FS path to that command – we need a SHA1 hash that Windows uses internally.

The `Import-Certificate` cmdlet ([docs](https://learn.microsoft.com/en-us/powershell/module/pki/import-certificate?view=windowsserver2025-ps)) prints this hash, so if we decided to use that command when installing the root certificate, we could theoretically save the hash in the Studio config file. The other option is using `Get-ChildItem` ([docs](https://learn.microsoft.com/en-gb/powershell/module/microsoft.powershell.management/get-childitem?view=powershell-7.5)), like this:

```
Get-ChildItem -Path Cert:\CurrentUser\Root | Where-Object { $_.Subject -like "*WordPress Studio*" }
```

We parse the hash from the output and then do:

```
Test-Certificate -Cert Cert:\CurrentUser\Root\{{ HASH }} -AllowUntrustedRoot
```

`certutil` can distinguish between whether the certificate is just on the keychain or fully trusted, and it's easy to provide the certificate to check (just provide the file path, as you already know). I haven't personally tested it enough to know if the output is unreliable. Going back to my [previous question](https://github.com/Automattic/studio/pull/1117#discussion_r2013793843), is this something you've experienced, @youknowriad? If so, then the PKI PowerShell module might be a good option because the output is quite terse.

youknowriad: No, no I didn't experience the difference in output myself, it just felt fragile to me that's it. Like a tool can decide to change the output at any moment (and I wonder if it already changed over time).

fredrikekelund: I see. The [Microsoft docs](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/certutil) for `certutil` do say this:

> Certutil isn't recommended to be used in any production code and doesn't provide any guarantees of live site support or application compatibilities. It's a tool utilized by developers and IT administrators to view certificate content information on devices.

fredrikekelund: Our use of `certutil` wouldn't be mission-critical, but there's clearly an inherent fragility. 

Feel free to explore the PowerShell PKI module option to see if that works. Parsing the output should definitely be easier.

wojtekn: I would use an output to avoid additional complexity. If we checked for `UNTRUSTED` and got an unexpected output, in the worst case scenario users would not see a message asking them to trust the certificate.

fredrikekelund: If we stick with `certutil`, we should parse the `Verified Application Policies:` part of the output, too. I'm not sure if both the `Server Authentication` and `Client Authentication` policies are needed, but at least one of them is.

src/lib/certificate-manager.ts
```diff
@@ -148,11 +152,61 @@ export async function openCertificate() {
...
+	} else if ( process.platform === 'darwin' ) {
+		try {
+			await execFilePromise( 'security', [
+				'verify-cert',
+				'-c',
+				CA_CERT_PATH,
+				'-p',
+				'ssl',
+				'-s',
+				CA_NAME,
```
mcsf: Looking at the manual (`man security`), option `-s` is deprecated in favour of `-n`:

```
            -s sslHost      Specify SSL host name for the ssl policy. (This option is deprecated;
                            use -n instead.)

            -n name         Specify a name to be verified, e.g. the SSL host name for the ssl
                            policy, or RFC822 email address for the smime policy. For backward
                            compatibility, if the -n option is provided without an argument, it will
                            be interpreted as equivalent to -N.
```

mcsf: That said, `CA_NAME` isn't a host name, it's the string used as certicates' `commonName`, so I'm not sure what this option adds to the command.

src/hooks/use-certificate-trust.ts
```diff
@@ -0,0 +1,34 @@
...
+
+	const checkCertificateTrust = useCallback( () => {
+		getIpcApi()
+			.isCATrusted()
+			.then( ( trusted ) => {
+				if ( isMounted.current ) {
+					setIsTrusted( trusted );
+				}
+			} );
+	}, [ setIsTrusted ] );
```
mcsf: Not super important, but there are a bunch of different identifiers coexisting:

- use certificate trust
- check certificate trust
- is CA trusted
- is Root CA trusted

With just a little refactoring we could probably just have `useIsRootCATrusted` and `isRootCATrusted`. Especially if in `ipc-handlers.ts` we imported the whole module to avoid collisions:

```ts
export async function isRootCATrusted( _event: IpcMainInvokeEvent ): Promise< boolean > {
	return certificateManager.isRootCATrusted();
}
```

src/lib/certificate-manager.ts
mcsf: The `-n` flag is apt for the domain-specific certificates:

```sh
security verify-cert -c domains/foo.wp.local.crt -p ssl -s foo.wp.local # ok
security verify-cert -c domains/foo.wp.local.crt -p ssl -s example.org  # nope
```

but it doesn't work with the CA:

```sh
security verify-cert -c studio-ca.crt -p ssl -s example.org # no errors
                                                            # but this is misleading!
```

mcsf (approved):
Can't vouch for the Windows side, but LGTM on macOS!

src/lib/certificate-manager.ts
youknowriad: So I've updated the code here to check for the `Verified Application Policies` but my confidence with my updates is not 100%. I was not able to test despite spending a lot of time trying to setup a windows environment.

That said, even if it fails, the risks are small to the users. (hide/show a message)

src/lib/certificate-manager.ts
```diff
@@ -148,11 +152,67 @@ export async function openCertificate() {
...
+		return false;
+	}
+
+	if ( process.platform === 'win32' ) {
+		try {
+			// Execute certutil with more specific validation
+			const { stdout } = await execFilePromise( 'certutil', [
+				'-verify',
+				'-urlfetch',
+				'-purpose',
```
gavande1: I am getting the unknown option `-purpose` error on Windows 11 Pro. It needs some refactoring.

gavande1:
FYI, I am taking over this task. p1745233995626849-slack-C06DRMD6VPZ

src/hooks/use-certificate-trust.ts
```diff
@@ -0,0 +1,42 @@
...
+	useEffect( () => {
+		isMounted.current = true;
+		checkCertificateTrust();
+		return () => {
+			isMounted.current = false;
+		};
+	}, [ checkCertificateTrust ] );
+
+	return isTrusted;
+}
```
gcsecsey:
```suggestion
export function useCertificateTrust(): boolean {
	const [ isTrusted, setIsTrusted ] = useState< boolean >( false );

	const checkCertificateTrust = useCallback( () => {
		// If the certificate is already trusted, don't check it again
		if ( isTrusted ) {
			return;
		}

		getIpcApi()
			.isCATrusted()
			.then( ( trusted ) => {
				setIsTrusted( trusted );
			} )
			.catch( ( error ) => {
				console.error( 'Failed to check certificate trust:', error );
			} );
	}, [ setIsTrusted, isTrusted ] );

	useWindowListener( 'focus', checkCertificateTrust );

	useEffect( () => {
		checkCertificateTrust();
	}, [ checkCertificateTrust ] );

	return isTrusted;
}
```

Condsider this a suggestion, but after testing this a bit, I don't think using a ref here prevents any extra executions. Was it a different purpose this was added for?

gcsecsey (approved):
Thanks for driving this forward @gavande1! 🙌 

This is looking good, and works well. When trusting the certificate, the change is also immediately reflected in the Studio app.

macOS:
| Certificate | Add new site form | Settings tab |
|--------|--------|--------|
| Not trusted | CleanShot 2025-05-02 at 14 10 02@2x | CleanShot 2025-05-02 at 14 08 58@2x |
| Trusted | CleanShot 2025-05-02 at 14 11 51@2x | CleanShot 2025-05-02 at 14 11 43@2x | 

On Windows, I had the certificate already installed as trusted, so I first removed it by running `certutil -delstore ROOT "WordPress Studio CA"` in an elevated Command Prompt:
CleanShot 2025-05-02 at 14 39 31@2x

| Certificate | Add new site form | Settings tab |
|--------|--------|--------|
| Not trusted | CleanShot 2025-05-02 at 14 40 05@2x | CleanShot 2025-05-02 at 14 39 50@2x | 
| Trusted | CleanShot 2025-05-02 at 14 43 05@2x | CleanShot 2025-05-02 at 14 42 54@2x |

fredrikekelund:
Taking a look at this now.

src/lib/certificate-manager.ts
```diff
@@ -148,11 +152,65 @@ export async function openCertificate() {
...
+			// Only consider the certificate trusted if it has the Server Authentication policy.
+			return hasValidPolicies;
+		} catch ( error ) {
+			return false;
+		}
+	} else if ( process.platform === 'darwin' ) {
+		try {
+			await execFilePromise( 'security', [
+				'verify-cert',
+				'-c',
```
```suggestion
				'-r',
```
The `-r` option is used to pass a path to a root cert file. I think its main feature is that it allows us to specify _both_ `-r` and `-c`. We don't do that here, but I still see no reason not to use `-r`.

src/lib/certificate-manager.ts
fredrikekelund: Let's remove the `-s CA_NAME` options. As @mcsf mentioned, it doesn't make a difference.

src/ipc-handlers.ts
```diff
@@ -55,12 +54,16 @@ import { getLogsFilePath, writeLogToFile, type LogLevel } from 'src/logging';
...
 import { SiteServer, createSiteWorkingDirectory } from 'src/site-server';
 import { DEFAULT_SITE_PATH, getResourcesPath, getSiteThumbnailPath } from 'src/storage/paths';
 import { loadUserData, saveUserData } from 'src/storage/user-data';
 import { DEFAULT_PHP_VERSION, DEFAULT_WORDPRESS_VERSION } from 'vendor/wp-now/src/constants';
-import { SupportedEditor } from './modules/user-settings/lib/editor';
-import { SupportedTerminal } from './modules/user-settings/lib/terminal';
+import {
+	openCertificate as openCertificateDialog,
+	isRootCATrusted,
+} from './lib/certificate-manager';
```
```suggestion
} from 'src/lib/certificate-manager';
```
Truly a nitpick, but we strive to always use absolute import paths.

src/modules/site-settings/edit-site-details.tsx
fredrikekelund: Unresolving this conversation because I still don't understand this change.

src/ipc-handlers.ts
```diff
@@ -1235,7 +1238,11 @@ export function openCertificate( _event: IpcMainInvokeEvent ) {
 	return openCertificateDialog();
 }
 
-export function getFileContent( event: IpcMainInvokeEvent, filePath: string ) {
+export async function isCATrusted( _event: IpcMainInvokeEvent ): Promise< boolean > {
```
```suggestion
export async function isCATrusted(): Promise< boolean > {
```
No need to include the `_event` argument anymore (after https://github.com/Automattic/studio/pull/1086 landed) if we don't use it.

src/hooks/use-certificate-trust.ts
fredrikekelund: Theoretically, the component could very well have unmounted between the `checkCertificateTrust` call and when it resolves, and updating the state of an unmounted component will generate a warning.

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -155,7 +156,7 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
 				phpVersion: selectedPhpVersion,
 				isWpAutoUpdating: selectedWpVersion === DEFAULT_WORDPRESS_VERSION,
 				customDomain: usedCustomDomain,
-				enableHttps: !! usedCustomDomain && enableHttps,
+				enableHttps: !! usedCustomDomain && ( isCertificateTrusted || enableHttps ),
```
fredrikekelund:
```suggestion
				enableHttps: !! usedCustomDomain && enableHttps,
```
Why would we not respect the state of the `Enable HTTPS` checkbox when saving the site?

src/components/content-tab-settings.tsx
```diff
@@ -75,13 +77,13 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
 					<SettingsRow label={ __( 'HTTPS' ) }>
 						<div>
 							<span>{ selectedSite.enableHttps ? __( 'Enabled' ) : __( 'Disabled' ) }</span>{ ' ' }
-							{ selectedSite.enableHttps && (
+							{ ! isCertificateTrusted && selectedSite.enableHttps && (
 								<Button variant="link" onClick={ () => getIpcApi().openCertificate() }>
```
fredrikekelund, line 81: I realize now that it would make more sense if this button triggered the `trustRootCA` function from `src/lib/certificate-manager.ts` on Windows, since the CA installation process is automated on Windows.

src/lib/certificate-manager.ts
```diff
@@ -148,11 +152,65 @@ export async function openCertificate() {
...
+	if ( ! fs.existsSync( CA_CERT_PATH ) ) {
+		return false;
+	}
+
+	if ( process.platform === 'win32' ) {
+		try {
+			// Execute certutil with more specific validation
+			const { stdout } = await execFilePromise( 'certutil', [
+				'-verify',
+				'-urlfetch',
```
```suggestion
```
Let's remove this option. The `certutil` help docs say this about the `-verify` option:

> Retrieve and verify AIA Certs and CDP CRLs

The way I understand it, it means that `certutil` will potentially check the validity of the trust chain by downloading data from Microsoft servers. Since our CA is generated locally, it's not trusted by any global CAs, and we can skip this procedure altogether.

fredrikekelund (commented):
Overall, this works as expected on both macOS in Windows in my testing.

There are a few tweaks worth implementing in `src/lib/certificate-manager.ts`, as far as I'm concerned. There's also `src/modules/site-settings/edit-site-details.tsx`, where I don't understand why we use `isCertificateTrusted` the way we do. I'll approve once we've settled the questions about that file 👍

src/hooks/use-certificate-trust.ts
gavande1: I am considering refactoring `useCertificateTrust` using Redux to avoid managing local states. I will create an additional PR to allow everyone to review it easily.

src/lib/certificate-manager.ts
gavande1: It makes to check for the root certificate explicitly.

src/ipc-handlers.ts
gavande1: Fixed.

src/ipc-handlers.ts
gavande1: Thanks for the reminder. Removed it.

src/modules/site-settings/edit-site-details.tsx
gavande1: I totally agree about that change. We should always honour the saved state.

src/components/content-tab-settings.tsx
gavande1: This is an excellent suggestion. I will create a follow-up PR to install the certificate on Windows instead of opening the CA folder when clicking on the trust certificate.

src/lib/certificate-manager.ts
gavande1: Removed `-urlfetch` parameter.

src/components/content-tab-settings.tsx
gavande1: I created follow up PR: https://github.com/Automattic/studio/pull/1327

src/hooks/use-certificate-trust.ts
gavande1: I created follow up PR: https://github.com/Automattic/studio/pull/1325

fredrikekelund (approved):
LGTM 👍

fredrikekelund:
@gavande1, one more quick thing: [the issue specs](https://linear.app/a8c/issue/STU-255) suggest enabling HTTPS by default when the custom domain option is disabled. I think this makes a lot of sense when creating new sites, and I suggest we implement it in this PR.

gavande1:
@fredrikekelund I added functionality of auto enabling https when the custom domain is enabled and the root certificate is trusted while creating a new site. The changes were added in 0169e08c383e1c6c471d9d1001623e52f2406214.

fredrikekelund:
@gavande1, friendly tip, you don't need to rebase the PR branch on trunk to resolve conflicts – a simple merge PR works perfectly fine. The benefit of that approach is that the original commit history is preserved and it's easier for reviewers to pull the latest changes.

src/components/site-form.tsx
```diff
@@ -267,6 +267,13 @@ export const SiteForm = ( {
 	const { locale } = useI18nData();
 	const isCertificateTrusted = useCertificateTrust();
 
+	// If the custom domain is enabled and the root certificate is trusted, enable HTTPS
+	useEffect( () => {
+		if ( useCustomDomain && isCertificateTrusted && setEnableHttps ) {
+			setEnableHttps( true );
+		}
+	}, [ useCustomDomain, isCertificateTrusted, setEnableHttps ] );
```
fredrikekelund, line 275: Definitely beyond the scope of this PR, but as we've identified before (https://github.com/Automattic/studio/pull/994#discussion_r1979067939), `SiteForm` should really be refactored to handle state internally and then pass an entire object containing the form data back to the parent when submitted.

fredrikekelund (approved):
Latest changes work well 👍

src/components/site-form.tsx
gavande1: Created issue STU-457 to get back on refactoring.

youknowriad:
Thanks for pushing this one to the end. Highly appreciated folks.

PR #1118, opened 2025-03-24
Studio: Add notification about Release notes to new version modal

katinthehatsite:
## Related issues

Closes STU-223

## Proposed Changes

This PR adds a notification to Studio for the case when there is a new update that is already downloaded and the user needs to restart the app. This notification has a link to the release notes. 

## Testing Instructions

* Apply the following diff:

```diff --git a/src/updates.ts b/src/updates.ts
index 5c74d1f2..fc03a4dc 100644
--- a/src/updates.ts
+++ b/src/updates.ts
@@ -256,3 +256,15 @@ function showReleaseNotesNotification() {
 
        activeNotification.show();
 }
+
+if ( process.env.NODE_ENV === 'development' ) {
+       setTimeout( () => {
+               updaterState = 'waiting-for-restart';
+               showReleaseNotesNotification();
+
+               // Wait before showing dialog
+               setTimeout( () => {
+                       showUpdateReadyToInstallNotice();
+               }, 2000 );
+       }, 5000 );
+}
```
* Restart the app
* Observe that once the app is restarted, you will see the following notification:

Screenshot 2025-03-24 at 1 18 37 PM

* Click on the notification and confirm that you are brought to the release notes
* After about 2 seconds, you will see the dialog window to either restart the app now or later

## Pre-merge Checklist

### DIFF
--- a/src/updates.ts
+++ b/src/updates.ts
@@ -94,6 +94,12 @@ export function setupUpdates() {
 	autoUpdater.on( 'update-downloaded', async () => {
 		updaterState = 'waiting-for-restart';
 		console.log( 'Update has been downloaded' );
+
+		showReleaseNotesNotification();
+
+		// Wait a moment before showing the dialog
+		await new Promise( ( resolve ) => setTimeout( resolve, 2000 ) );
+
 		await showUpdateReadyToInstallNotice();
 	} );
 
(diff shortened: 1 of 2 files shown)


src/updates.ts
```diff
@@ -90,6 +94,12 @@ export function setupUpdates() {
 	autoUpdater.on( 'update-downloaded', async () => {
 		updaterState = 'waiting-for-restart';
 		console.log( 'Update has been downloaded' );
+
+		showReleaseNotesNotification();
+
+		// Wait a moment before showing the dialog
+		await new Promise( ( resolve ) => setTimeout( resolve, 2000 ) );
```
katinthehatsite, line 101: This looks strange but I realised that it does not work well otherwise. This is because the window focus immediately moves from notification to the dialog window. Without the timeout, the result will look like this: you will still hear the notification and it will in the notification center but you won't see it on the screen with that quick window focus switch.

wojtekn:
Why do we need a notification if Studio shows a dialog when a new version is available?

katinthehatsite:
@wojtekn  - you can see a bit more discussion here: STU-223

Mainly because the initial idea was to add a link to releases on the dialog but Electron does not support rich text in native dialogs so we were looking for the way around it. But it seems that we are going to hold with this for now and forego this task so I am going to close this issue.

PR #1119, opened 2025-03-24
Performance metrics: Bootstrap the test

youknowriad:
## Related issues

Fixes DOTCOM-668-linear-issue

## Proposed Changes

-  This PR introduces a new performance metrics testing framework that allows us to measure and track key performance indicators in Studio:

  - Creates a foundation for performance testing with Playwright
  - Implements measurements for site creation and startup times
  - Generates performance reports with median values
  - Adds documentation on how to run and interpret performance tests

  The metrics framework will help us identify performance regressions and track improvements over time, ensuring a better user experience.

On follow-up PRs, I'll be working on a CI job to compare and output results. I'll also work on pushing the metrics to codevitals.run to show a project dashboard.

## Testing Instructions

you can try the README instructions locally.

### DIFF
--- a/metrics/performance-reporter.ts
+++ b/metrics/performance-reporter.ts
@@ -1,0 +1,67 @@
+import { writeFileSync, existsSync, mkdirSync } from 'fs';
+import path from 'path';
+import type { Reporter, FullResult, TestCase, TestResult } from '@playwright/test/reporter';
+
+export type PerformanceResults = Record< string, number >;
+
+class PerformanceReporter implements Reporter {
+	private results: Record< string, PerformanceResults >;
+
+	constructor() {
+		this.results = {};
+	}
+
+	onTestEnd( test: TestCase, result: TestResult ): void {
+		for ( const attachment of result.attachments ) {
+			if ( attachment.name !== 'results' ) {
+				continue;
+			}
+
+			if ( ! attachment.body ) {
+				throw new Error( 'Empty results attachment' );
+			}
+
+			const testSuite = path.basename( test.location.file, '.test.js' );
+			const resultsId = testSuite;
+			const resultsPath = process.env.ARTIFACTS_PATH as string;
+			const resultsBody = attachment.body.toString();
+			const results = JSON.parse( resultsBody );
+
+			if ( ! existsSync( resultsPath ) ) {
+				mkdirSync( resultsPath, { recursive: true } );
+			}
+			writeFileSync(
+				path.join( resultsPath, `${ resultsId }-results.json` ),
+				JSON.stringify( results, null, 2 )
+			);
+
+			this.results[ testSuite ] = results;
+		}
+	}
+
+	onEnd( result: FullResult ) {
+		if ( result.status !== 'passed' ) {
+			return;
+		}
+
+		if ( process.env.CI ) {
+			return;
+		}
+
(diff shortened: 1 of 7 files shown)


metrics/utils.ts
```diff
@@ -0,0 +1,11 @@
+export function median( array ) {
```
sejas, line 1: We could consider including the standard deviation in the future if we want to have a more complete picture of certain metrics. For example, during my tests, the site's startup time ranged from 3 to 4 seconds, with the median being the lowest at 3 seconds.

metrics/tests/site-startup.test.ts
```diff
@@ -0,0 +1,86 @@
...
+		await session.launch();
+
+		// Complete onboarding before tests
+		const onboarding = new Onboarding( session.mainWindow );
+		await expect( onboarding.heading ).toBeVisible();
+		await onboarding.continueButton.click();
+	} );
+
+	// eslint-disable-next-line no-empty-pattern
+	test.afterAll( async ( {}, testInfo ) => {
```
sejas, line 23: I tried using a parameter with an underscore `_`, the linter was happy, but I got this error when running the metrics: `First argument must use the object destructuring pattern: _`. So we need this eslint exception 👍

metrics/tests/site-startup.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+			// Log performance data for this run
+			console.log( `Run ${ i + 1 }/5: Restart took ${ duration }ms` );
+			results.siteStartup.push( duration );
+
+			// Wait a moment before next cycle
+			await session.mainWindow.waitForTimeout( 100 );
+		}
+
+		// Delete the site after test
```
sejas, line 76: If we measure `siteCreation` in this test, then we could consider measuring siteDeletion as well.

metrics/performance-reporter.ts
```diff
@@ -0,0 +1,67 @@
...
+			const resultsPath = process.env.ARTIFACTS_PATH as string;
+			const resultsBody = attachment.body.toString();
+			const results = JSON.parse( resultsBody );
+
+			if ( ! existsSync( resultsPath ) ) {
+				mkdirSync( resultsPath, { recursive: true } );
+			}
+			writeFileSync(
+				path.join( resultsPath, `${ resultsId }-results.json` ),
+				JSON.stringify( results, null, 2 )
```
sejas: This is similar to directly using `resultsBody`.

sejas (approved):
It looks great! Thank you for adding a way to measure the app's performance. We can incorporate more metrics in the future, perhaps those directly related to Playground too, such as accessing wp-admin, etc.

I confirm that after running `npm run package && npm run test:metrics`, I saw the metrics in the console and a new report file was created at `artifacts/site-startup.test.ts-results.json`.

Screenshot 2025-03-25 at 10 33 24

metrics/tests/site-startup.test.ts
youknowriad: Good call, I'm not going to add it now cause I have a few tooling follow-ups but once the tooling is in place, let's add all the metrics that we need.

youknowriad: Good idea, let's get the tooling in place and figure out the right metrics later.

PR #1120, opened 2025-03-24
Add mu-plugin to prevent updates on selected WordPress versions

bcotrim:
## Related issues

- Fixes STU-113

## Proposed Changes

- Implemented a feature to disable WordPress auto-updates for sites not running the latest stable version
- Updated site update workflow to manage the auto-update disabling plugin
- Added a warning icon with tooltip in the WordPress version selection UIs
- Auto-updates are controlled via a `mu-plugin`

| Version updates warning | New version select UI | Auto-updates disabled | Auto-updates enabled |
| image | image | Screenshot 2025-03-25 at 19 54 29 | Screenshot 2025-03-25 at 19 56 21 | 

## Testing Instructions

- Create a new site with a WordPress version that is not the latest stable version
- Verify that a warning icon appears next to the WordPress version field with a tooltip indicating that auto-updates will be disabled
- Check that the mu-plugin for disabling auto-updates is added to the site
- Change a site's WordPress version from non-latest to latest and verify that the mu-plugin is removed
- Change a site's WordPress version from latest to non-latest and verify that the mu-plugin is added
- Verify that sites with the latest WordPress version don't have the auto-update disabling plugin

## Pre-merge Checklist


wojtekn:
@bcotrim I tested it, and it worked as expected.

The UI is not 100% clear yet; I think we may improve it. What if we grouped dropdown options, with the first one preselected and the only one that doesn't show a tooltip?

Screenshot 2025-03-26 at 17 32 05

Then, we could adjust the tooltip wording, for example:

> WordPress Core automatic updates will be disabled for this site.

wp-resources/mu-plugins/studio-disable-auto-updates.php
```diff
@@ -0,0 +1,42 @@
+<?php
```
wojtekn: Could we keep it as a mu-plugin mounted by Playground and not added to the user's site filesystem?

Also, is there a reason to disable all updates? As we add that plugin depending on the WordPress version selected, it would make sense to disable only auto updates of the WP core, letting users still benefit from automatic updates of plugins or themes.

Could we achieve this by setting the file with the constant? Or just defining constant for Playground environment?

```
define( 'WP_AUTO_UPDATE_CORE', true );
```

sejas: Here is were we usually set mu-plugins.
https://github.com/Automattic/studio/blob/f0efa981a196265b5d31d058e19bed5496a122b0/vendor/wp-now/src/wp-now.ts#L457-L468

We have some code that already sets that constant in. We should check the current behaviour:
https://github.com/Automattic/studio/blob/f0efa981a196265b5d31d058e19bed5496a122b0/vendor/wp-now/src/wp-now.ts#L427-L429

src/ipc-handlers.ts
```diff
@@ -208,6 +213,11 @@ export async function createSite(
 		}
 	}
 
+	// If not using the latest WordPress version, copy our mu-plugin to disable auto-updates
+	if ( wpVersion?.version && ! wpVersion.isLatest ) {
+		await server.addMuPlugin( STUDIO_DISABLE_WP_AUTO_UPDATES_PLUGIN );
```
sejas: As I mentioned in the issue, this might not be an effective check. Today I can create a site with the latest version, and tomorrow a new WP version may be published, making my version no longer the latest.

sejas (commented):
One possible solution could be checking if the user set the version with the dropdown , and calling `defineWpConfigConsts` when the site starts.

https://github.com/Automattic/studio/blob/f0efa981a196265b5d31d058e19bed5496a122b0/vendor/wp-now/src/wp-now.ts#L427-L430

src/ipc-handlers.ts
wojtekn: I proposed an approach for that above. If the user keeps the latest (6.7.2 from the Always Latest group), they will have auto updates. If the user selects any other named version, like 6.7.2 below, auto-updates will be disabled.

As you pointed out, we have a code that handles that, which was likely broken at some point.

sejas: The code we have seems to be inherited from wp-now CLI parameters, so it is not really used in Studio.<br>Having groups makes sense to me. "Always latest" could lead to confusion, as the user needs to manually update major versions, and minor updates can take time depending on whether the site is started and accessed. We could say something like "Auto updates enabled." Maybe moving nightly to that group also makes sense.

> For development sites, the default value of `WP_AUTO_UPDATE_CORE` is `true`. For other sites, the default value of `WP_AUTO_UPDATE_CORE` is `minor`. ([ref](https://developer.wordpress.org/advanced-administration/upgrade/upgrading/#configuration-via-wp-config-php))

wojtekn: > We could say something like "Auto updates enabled." Maybe moving nightly to that group also makes sense.

We could say "Auto Updated" to make it shorter, still clear, and consistent with other groups.

bcotrim:
Thank you for the feedback and review @wojtekn @sejas 
I think I've made all suggested changes:
- `mu-plugin` is now added in wp-now, similar to existing ones
- the wordpress version select dropdown now has a new style, grouping latest, beta and stable versions
- wpVersion is now persisted in `userData` so we can control the update auto-updates `mu-plugin`

Please let me know if you have any additional feedback or notice any errors.

vendor/wp-now/src/wp-now.ts
```diff
@@ -602,6 +602,20 @@ async function mountInternalMuPlugins( php: PHP ) {
...
 		`
 	);
+
+	// Only add the auto-updates disabling plugin for non-latest WordPress versions
+	if ( options?.wordPressVersion && options.wordPressVersion !== 'latest' ) {
+		php.writeFile(
+			path.posix.join( PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER, '0-disable-auto-updates.php' ),
+			`<?php
+			// Disable auto-updates
+			add_filter( 'auto_update_core', '__return_false' );
```
sejas: I guess this is not really necessary with the other three filters (dev, major and minor).
In fact, [`auto_udpate_core`](https://developer.wordpress.org/reference/hooks/auto_update_type/#description) is not working for me at all. I tried using only this hook in beta and old stable versions and is not making any effect. I think we can remove it.

sejas (changes requested):
@bcotrim, thanks for applying the changes. I like the grouped versions. What do you think about adding "(latest)" next to the  6.7.2  label as we had before?

For new sites it works as expected and it adds the option correctly, but I noticed some issues for existing sites:
- The info icon showing that auto-updates are disabled is misleading for existing sites. All existing sites have updates enabled. Maybe the best solution is to pin the version for existing sites with a version lower than 6.7.2
- We shouldn't change the behavior for the latest version. In this case, existing sites with version 6.7.2 are identified as "stable" in the dropdown, while I believe they should be identified as "Auto update." in the dropdown.
- When a site has a different version, that version doesn't appear in the dropdown anymore. For example, before this PR we included the old nightly version of the site in the dropdown.


https://github.com/user-attachments/assets/15c39ee8-4ebd-4e5c-987b-ee7885445f9e


https://github.com/user-attachments/assets/0f2f6511-b03b-49e4-8ed1-2d3d2c395905


> [!WARNING]  
> Creating a new site with nightly version throws an error:


https://github.com/user-attachments/assets/b84f7669-1b6a-4ec3-8cbc-872101708076

src/components/site-form.tsx
```diff
@@ -275,6 +275,12 @@ export const SiteForm = ( {
 	const offlineMessage = __( 'Changing WordPress version requires an internet connection.' );
 
 	const { data: wpVersions = [] } = useGetWordPressVersions();
+	const latestStableVersion = wpVersions.find( ( version ) => version.value === 'latest' );
+
+	const betaVersions = wpVersions.filter( ( version ) => version.isBeta || version.isDevelopment );
+	const stableVersions = wpVersions.filter(
+		( version ) => version !== latestStableVersion && ! version.isBeta && ! version.isDevelopment
+	);
```
sejas, line 283: It would be great if we could centralize these calculations in one place.

sejas (commented):
It works as expected.
I see some code repetition. I think we could simplify it by moving the version calculations inside a hook or directly to the WordPress versions API.

I found that we don't display the "(latest)" label when creating a site. We show it when editing a site though.

Screenshot 2025-04-09 at 11 17 14

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -70,25 +73,55 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
...
-	const wordpressVersionOptions = wordpressVersions.map( ( version ) => ( {
-		label: version.label,
-		value: version.value,
-	} ) );
+	const latestVersion = wordpressVersions.find( ( version ) => version.value === 'latest' );
+	const shouldAutoUpdate = ! selectedSite?.wpVersion || selectedSite.wpVersion === 'latest';
+	const autoUpdatedVersions = latestVersion
+		? [
+				{
+					label: `${ latestVersion.label } (latest)`,
```
sejas, line 83: Let's translate it:

```suggestion
					label: sprintf( __( '%s (latest)' ), latestVersion.label ),
```

sejas:
Just a heads-up that @nightnei  worked in another PR https://github.com/Automattic/studio/pull/1189

nightnei:
@bcotrim  Oh, I think now I will have difficulties with resolving conflicts with my branch 😅
What is the purpose for this force-push?

Update, yes ☹️
Screenshot 2025-04-15 at 16 18 49

bcotrim:
Changes were implemented with https://github.com/Automattic/studio/pull/1189
Closing.

PR #1123, opened 2025-03-25
Performance metrics: Add a CI job to compare the performance of PR branches with trunk

youknowriad:
## Related issues

Fixes DOTCOM-670-linear-issue

## Changes proposed in this Pull Request:

**Big Picture** My goal is to bring metrics tracking and performance dashboard to the tools Studio use daily to avoid performance regressions and work on performance improvements similarly to the [Gutenberg project](https://www.codevitals.run/project/gutenberg) for instance.

Implementing this require three complementary steps:

 1- First step is to have some test runner (e2e test here) that we can launch to compute the metrics.
 2- Second step is to be able to use the same test runner in two separate branches as part of the same CI job. The reason for this is that CI runners capacities are unstable and we can't rely on the absolute numbers. We have to compute relative numbers and compare each commit with a given "base commit".
 3- Third step is to send these values (base branch and commit) to the dashboard (codevitals) to log it and produce the graphs.

The first step is being implemented in https://github.com/Automattic/studio/pull/1119
The current PR implements the second step. 

So it adds a "compare-perf" tool that was copied and adapted from the Woo and Gutenberg repository. It's almost a generic tool. the only specific things at the moment is the `config.js` file and some specifics about switching apps to use for the e2e test depending on the branch we're running. 

The PR also adds a "metrics" workflow to CI. The workflow runs for each PR and compare its metrics with trunk. And it also runs on trunk for each commit and compares the commit with a "fixed" point of comparison (this is going to be used later to track the number on a graph like [here](http://codevitals.run)) 

### How to test the changes in this Pull Request:

You can run the perf tool locally like so (give random commits or tags to compare, the first commit must include the perf tests though as it's going to be used to run the tests)

```
cd tools/compare-perf
pnpm test perf 1485688ecf2e4746306879d938f5099493ab3d4b  trunk
```

you'll get something close to the following results:

Screenshot 2025-03-25 at 1 19 30 PM

Also, we can check the results of the "metrics" workflow on the PR as well.

### DIFF
--- a/metrics/README.md
+++ b/metrics/README.md
@@ -7,10 +7,10 @@ This directory contains tools for measuring and tracking performance metrics in 
 To run the performance tests:
 
 ```bash
-// Package the application first
+# Package the application first
 npm run package
 
-// Run the performance tests
+# Run the performance tests
 npm run test:metrics
 ```
 
@@ -23,7 +23,21 @@ This will:
 
 The performance tests simulate key user workflows and measure the time they take to complete. Currently, we measure:
 
-- Site creation and startup time: How long it takes to create a new WordPress site and have it running
+- **siteCreation**: How long it takes to create a new WordPress site and have it running
+- **siteStartup**: How long it takes to restart an existing site
+
+## Comparing Performance Between Commits
+
+You can compare performance metrics between different commits or branches:
+
+```bash
+npm run test:metrics:compare <commit1> <commit2>
+```
+
+This tool is useful for:
+- Testing performance impact of code changes
+- Identifying performance regressions
+- Benchmarking improvements in new features
 
 ## Understanding the Results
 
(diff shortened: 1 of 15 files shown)


metrics/README.md
```diff
@@ -23,7 +23,21 @@ This will:
...
-- Site creation and startup time: How long it takes to create a new WordPress site and have it running
+- **siteCreation**: How long it takes to create a new WordPress site and have it running
+- **siteStartup**: How long it takes to restart an existing site
+
+## Comparing Performance Between Commits
+
+You can compare performance metrics between different commits or branches:
+
+```bash
+npm run test:metrics:compare <commit1> <commit2>
```
WunderBart: Looks like this script is missing?

WunderBart (approved):
Left just one comment. Apart from that it looks good! 🚀 

_Accepting despite the CI job failing as the tests should start passing once the changes from this PR are in `trunk`._

metrics/README.md
youknowriad: Right I removed it. I thought it was useless. I'll update the docs

PR #1125, opened 2025-03-25
Add support for IDE installed in user's home directory

### LINKED ISSUE #21: Feature Request: Support PhpStorm installed from JetBrains Toolbox app
### What

Currently the PhpStorm app when installed by JetBrains Toolbox does not get set to installed via the logic in https://github.com/Automattic/studio/blob/trunk/src/hooks/use-check-installed-apps.tsx

### Why

JetBrains Toolbox does not keep the PhpStorm app in the `/Applications/` folder, it's actually stored in the user's directory like for mine that may be something like `/Users/sc0ttkclark/Applications/`.

### How

It may be possible to check for apps in both directories without having to make this JetBrains Toolbox specific.

The logic in https://github.com/Automattic/studio/blob/trunk/src/lib/is-installed.ts could be updated to have the function `isInstalled` check if the `process.platform != 'win32'` and in that case it would check if the `fs.existsSync( appPaths[ key ] )` OR `fs.existsSync( userPathPrefix + appPaths[ key ] )`.

Something like that could perhaps be done.

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/studio/issues/21 (STU-2)

## Proposed Changes

Currently, Studio only checks for VS Code and PHPStorm in the system-wide `/Applications` directory. This change adds support for detecting these apps in the user's `~/Applications` directory as well.

## Testing Instructions

1. Install VS Code or PHPStorm in your user's Applications folder (`~/Applications`)
2. Verify that Studio detects the app in the user's Applications folder
3. Verify that Studio still detects apps in the system-wide `/Applications` folder
4. Verify that the detection still works on Windows

## Pre-merge Checklist

### DIFF
--- a/src/lib/is-installed.ts
+++ b/src/lib/is-installed.ts
@@ -2,22 +2,32 @@ import { app } from 'electron';
 import fs from 'fs';
 import path from 'path';
 
-let appPaths: Record< keyof InstalledApps, string >;
+let appPaths: Record< keyof InstalledApps, string[] >;
 
 if ( process.platform === 'darwin' ) {
+	// Get both system and user Applications directories
+	const systemApplications = '/Applications';
+	const userApplications = path.join( app.getPath( 'home' ), 'Applications' );
+
 	appPaths = {
-		vscode: '/Applications/Visual Studio Code.app',
-		phpstorm: '/Applications/PhpStorm.app',
+		vscode: [
+			path.join( systemApplications, 'Visual Studio Code.app' ),
+			path.join( userApplications, 'Visual Studio Code.app' ),
+		],
+		phpstorm: [
+			path.join( systemApplications, 'PhpStorm.app' ),
+			path.join( userApplications, 'PhpStorm.app' ),
+		],
 	};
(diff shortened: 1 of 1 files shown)


ivan-ottinger (approved):
The changes look good and work as expected. 👌🏼 

I have tested the PR with PHPStorm on macOS and Studio picked it up correctly in both locations (user's `~/Applications` and system-wide `/Applications`).

sejas (approved):
Thanks for the change. I confirm I was able to open VSCode and PHP Storm in MacOS. 
This PR doesn't change the way we open the app using deep links:  👍 

https://github.com/Automattic/studio/blob/17e5b6ef6157feb78db66c0a04c36200b6cf5da2/src/ipc-handlers.ts#L1097-L1104

wojtekn:
Thanks for reviewing and testing @ivan-ottinger and @sejas .

> This PR doesn't change the way we open the app using deep links: 👍

Correct, it only changes the part that checks if apps are installed. The rest is unchanged and relies on deep links.

PR #1130, opened 2025-03-25
Implement studio go for creating preview sites

bcotrim:
## Related issues

- Fixes STU-233

## Proposed Changes

- Implemented `go` command to create a preview site in WordPress.com
- Added unit tests for the created workflow

**Note**: As discussed on Slack, the way we fetch the user auth token is hard coded and will be adjusted in the a follow-up PR.

## Testing Instructions

In the project root folder:
- `npm run cli:build`
- `node dist/cli/main.js go [folder]`

Some suggested test scenarios:
1. Folder is not a WordPress installation folder:

image

2. Folder is a WordPress installation but you don't have WordPress.com authentication in Studio app:

image

3. Folder is a WordPress installation

image

5. Empty folder argument, navigating to a WordPress installation folder and running the cli dist file, should try to create a preview site for that folder

image

6. Running the command when you reach the 10 preview sites limit

image

## Pre-merge Checklist


cli/commands/preview/create.ts
```diff
@@ -1,89 +1,82 @@
...
+	if ( ! isSiteReady ) {
+		logger.reportError( LoggerAction.READY, 'Failed to create preview site' );
+		return;
+	}
+	cleanup( archivePath );
+	logger.reportSuccess(
+		LoggerAction.READY,
+		`Preview site available at: https://${ uploadResponse.site_url }`
+	);
 }
```
```suggestion
class CommandError extends Error {
	action: LoggerAction;

	constructor( action: LoggerAction, message: string ) {
		super( message );
		this.action = action;
	}
}

async function runCommand( siteFolder: string, outputFormat?: OutputFormat ): Promise< void > {
	const archivePath = path.join(
		os.tmpdir(),
		`${ path.basename( siteFolder ) }-${ Date.now() }.zip`
	);
	const logger = new Logger< LoggerAction >( outputFormat );

	try {
		logger.reportStart( LoggerAction.VALIDATE, 'Validating...' );
		validateSiteFolder( siteFolder );

		const token = await getAuthToken();
		logger.reportSuccess( LoggerAction.VALIDATE, 'Validation successful' );

		logger.reportStart( LoggerAction.ARCHIVE, 'Creating archive...' );
		await createArchive( siteFolder, archivePath );
		logger.reportSuccess( LoggerAction.ARCHIVE, 'Archive created' );

		logger.reportStart( LoggerAction.UPLOAD, 'Uploading archive...' );
		const uploadResponse = await uploadArchive( archivePath, token );
		if ( ! uploadResponse.site_url || ! uploadResponse.site_id ) {
			throw new CommandError( LoggerAction.UPLOAD, 'Failed to upload archive' );
		}
		logger.reportSuccess( LoggerAction.UPLOAD, 'Archive uploaded' );

		logger.reportStart( LoggerAction.READY, 'Creating preview site...' );
		const isSiteReady = await waitForSiteReady( uploadResponse.site_id, token );
		if ( ! isSiteReady ) {
			throw new CommandError( LoggerAction.READY, 'Failed to create preview site' );
		}
		cleanup( archivePath );
		logger.reportSuccess(
			LoggerAction.READY,
			`Preview site available at: https://${ uploadResponse.site_url }`
		);
	} catch ( error ) {
		if ( error instanceof CommandError ) {
			logger.reportError( error.action, error.message );
		}
	}
}
```
This is probably subjective, but given this `async / await` ES6+ world we live in, I would love to embrace classical JS error handling with exceptions that actually throw instead of potentially returning errors.

Implementing custom error classes (like `CommandError` in my suggestion) and using them across the different CLI modules would help make this more convenient.

cli/commands/preview/lib/validation.ts
```diff
@@ -0,0 +1,29 @@
+import fs from 'fs';
+import path from 'path';
+
+function isWordPressDirectory( projectPath: string ): boolean {
+	return (
+		fs.existsSync( path.join( projectPath, 'wp-content' ) ) &&
+		fs.existsSync( path.join( projectPath, 'wp-includes' ) ) &&
+		fs.existsSync( path.join( projectPath, 'wp-load.php' ) )
+	);
+}
```
fredrikekelund: This is an interesting example of logic that would ideally be shared across Studio and the CLI, but that is small enough to not warrant implementing its own CLI command. 

I would actually consider replacing this with:

```ts
import { isWordPressDirectory } from 'src/lib/fs-utils';
```

Webpack/TS can tree shake the code just fine, and we should take advantage of the fact that the CLI code lives right next to the Studio code while we can.

cli/commands/preview/lib/auth.ts
```diff
@@ -0,0 +1,25 @@
...
+		'appdata-v1.json'
+	);
+
+	if ( ! fs.existsSync( appDataPath ) ) {
+		return null;
+	}
+
+	try {
+		const userData = JSON.parse( fs.readFileSync( appDataPath, 'utf8' ) );
+		return userData.authToken?.accessToken || null;
```
fredrikekelund: Let's add a zod schema to validate the `authToken.accessToken` part of the config file.

cli/commands/preview/lib/archive.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export async function createArchive(
+	siteFolder: string,
+	archivePath: string
+): Promise< archiver.Archiver > {
+	return new Promise( ( resolve, reject ) => {
+		const output = fs.createWriteStream( archivePath );
+		const archive = archiver( 'zip', {
+			zlib: { level: ZIP_COMPRESSION_LEVEL },
+		} );
```
fredrikekelund, line 16: Not something we have to look into now, but there was a recent request to follow symlinks when creating preview site archives. This makes a lot of sense to me, and if there's an easy way to solve it with `archiver`, we might as well solve it now.

fredrikekelund: For reference, here's the relevant issue STU-309

cli/commands/preview/lib/api.ts
```diff
@@ -0,0 +1,87 @@
...
+			},
+		],
+	];
+
+	try {
+		const response = await wpcom.req.post< CreateSiteResponse >( {
+			path: '/jurassic-ninja/create-new-site-from-zip',
+			apiNamespace: 'wpcom/v2',
+			formData,
+		} );
```
fredrikekelund: I'd add a zod schema here.

cli/commands/preview/lib/api.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+async function checkSiteStatus( siteId: number, token: string ): Promise< boolean > {
+	const wpcom = new WPCOM( token );
+
+	try {
+		const response = await wpcom.req.get< StatusResponse >( '/jurassic-ninja/status', {
+			apiNamespace: 'wpcom/v2',
+			site_id: siteId,
+		} );
```
fredrikekelund: I'd add a zod schema here.

cli/commands/preview/lib/api.ts
```diff
@@ -0,0 +1,87 @@
...
+	domain_name: string;
+	atomic_site_id: number;
+	is_deleted: string;
+}
+
+export enum SnapshotStatus {
+	Pending = '0',
+	Processing = '1',
+	Active = '2',
+}
```
fredrikekelund: I like the enum usage 👍

fredrikekelund (commented):
I have yet to test this, but the code is looking good overall and the structure is easy to follow 👍

My biggest comment is that I'd suggest following a "classical" error handling pattern where exceptions are thrown instead of returned. Implementing a custom error class that extends `Error` would make this easier because we could pass a logger action as a prop on the error.

cli/package.json
```diff
@@ -7,4 +7,4 @@
 	"description": "Studio by WordPress.com CLI",
 	"license": "GPL-2.0-or-later",
 	"main": "index.js"
-}
+}
```
fredrikekelund, line 10: Do you use some tooling that strips out newlines from the end of files, @bcotrim?

fredrikekelund (commented):
I've tested this and it's working great overall 👍 I pushed two small commits to fix a deprecation warning from an outdated descendant dependency and to fix how the `--output-format` option is parsed.

The only issue I encountered is that if I pass the `--output-format=json` option, the command exits early with the following error:

```
/Users/fredrik/Code/studio/dist/cli/main.js:43174
            throw new Error('Cannot report success for an action that is not currently in progress');
                  ^

Error: Cannot report success for an action that is not currently in progress
    at Logger.reportSuccess (/Users/fredrik/Code/studio/dist/cli/main.js:43174:19)
    at /Users/fredrik/Code/studio/dist/cli/main.js:42846:16
    at Generator.next (<anonymous>)
    at fulfilled (/Users/fredrik/Code/studio/dist/cli/main.js:42805:58)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)

Node.js v22.9.0
```

fredrikekelund:
One more thing: we should write the newly created preview site to the Studio appdata file.

cli/package.json
bcotrim: Nothing besides Cursor 🤔

cli/logger.ts
```diff
@@ -1,25 +1,82 @@
...
+		}
+
+		this.spinner.succeed( message );
+		this.currentAction = null;
+	}
+
+	public reportError( error: LoggerError< T > ) {
+		if ( this.currentAction !== error.action ) {
+			throw new Error( 'Cannot report error for an action that is not currently in progress' );
+		}
```
fredrikekelund: Do we really need this? It seems like removing it would greatly reduce the number of times we pass around `LoggerAction` references.

cli/commands/preview/lib/appdata.ts
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	try {
+		const fileContent = fs.readFileSync( appDataPath, 'utf8' );
+		const userData = JSON.parse( fileContent );
+		const result = UserDataSchema.safeParse( userData );
+
+		if ( ! result.success ) {
+			throw new LoggerError( `Invalid appdata format. Please run the Studio app again.`, action );
+		}
```
fredrikekelund:
```suggestion
		const result = UserDataSchema.parse( userData );
```
Since we already have an extensive `catch` clause here, it seems like we could use the `zod.parse` method.

cli/commands/preview/lib/appdata.ts
```diff
@@ -0,0 +1,132 @@
...
+export async function saveAppdata( userData: UserData, action: string ): Promise< void > {
+	const appDataPath = getAppdataPath();
+
+	try {
+		// Check all required properties
+		if ( ! userData.version ) {
+			userData.version = 1;
+		}
+
+		// Create a deep copy to avoid reference issues
```
fredrikekelund: Did you encounter any issues like this?

cli/commands/preview/lib/appdata.ts
```diff
@@ -0,0 +1,132 @@
...
+		const dataToSave = JSON.parse( JSON.stringify( userData ) );
+		const fileContent = JSON.stringify( dataToSave, null, 2 ) + '\n';
+
+		// Write the file
```
fredrikekelund:
```suggestion
```

cli/commands/preview/lib/appdata.ts
```diff
@@ -0,0 +1,132 @@
...
+	userId: z.number().optional(),
+} );
+
+const UserDataSchema = z.object( {
+	version: z.number().optional(),
+	sites: z.array( z.any() ).optional(),
+	snapshots: z.array( SnapshotSchema ).optional(),
+	authToken: z.any().optional(),
+	locale: z.string().optional(),
+} );
```
fredrikekelund: I applaud this initiative, but this schema is missing a couple of top-level items (`sentryUserId` and `lastSeenVersion`, for example).

The best-case scenario is that we have a shared schema for the user-config file between Studio and the CLI. However, to avoid blowing the scope of this PR, I suggest keeping a schema that validates only the `snapshots` array in the user config. All other properties should be left untouched between reading and writing the file.

cli/commands/preview/lib/auth.ts
```diff
@@ -0,0 +1,82 @@
...
+		const fileContent = fs.readFileSync( appDataPath, 'utf8' );
+		const userData = JSON.parse( fileContent );
+
+		// Validate the userData against our schema
+		const result = AuthTokenSchema.safeParse( userData );
+
+		if ( ! result.success ) {
+			// Format the error in a more user-friendly way
+			throw new LoggerError(
+				`Authentication data is invalid. Please run the Studio app and authenticate again.`,
```
fredrikekelund:
```suggestion
				`Authentication data is invalid. Please run the Studio app and log in to WordPress.com again.`,
```
Aligns better with how we label that action in the Studio UI.

cli/commands/preview/lib/auth.ts
```diff
@@ -0,0 +1,82 @@
...
+	if ( ! fs.existsSync( appDataPath ) ) {
+		throw new LoggerError(
+			'Authentication required. Please run the Studio app and authenticate first.',
+			action
+		);
+	}
+
+	try {
+		const fileContent = fs.readFileSync( appDataPath, 'utf8' );
+		const userData = JSON.parse( fileContent );
```
fredrikekelund: Would it make sense for this function to also use `readAppdata`?

cli/commands/preview/lib/auth.ts
```diff
@@ -0,0 +1,82 @@
...
+				`Authentication data is invalid. Please run the Studio app and authenticate again.`,
+				action
+			);
+		}
+
+		const { authToken } = result.data;
+
+		if ( ! authToken || ! authToken.accessToken ) {
+			throw new LoggerError(
+				'Authentication required. Please run the Studio app and authenticate first.',
```
fredrikekelund:
```suggestion
				'Authentication required. Please run the Studio app and log in to WordPress.com first.',
```

cli/commands/preview/lib/api.ts
```diff
@@ -0,0 +1,127 @@
...
+			formData,
+		} );
+
+		// Validate the response against our schema
+		const result = CreateSiteResponseSchema.safeParse( rawResponse );
+
+		if ( ! result.success ) {
+			throw new LoggerError( 'Invalid API response', action );
+		}
+
```
fredrikekelund:
```suggestion
		const result = CreateSiteResponseSchema.parse( rawResponse );
```
Again, we have an extensive `catch` clause already, so I don't see the need for using `safeParse`

cli/commands/preview/lib/api.ts
```diff
@@ -0,0 +1,127 @@
...
+			apiNamespace: 'wpcom/v2',
+			site_id: siteId,
+		} );
+
+		const result = StatusResponseSchema.safeParse( rawResponse );
+
+		if ( ! result.success ) {
+			return false;
+		}
+
```
fredrikekelund:
```suggestion
		const result = StatusResponseSchema.parse( rawResponse );
```

cli/commands/preview/lib/api.ts
```diff
@@ -0,0 +1,127 @@
...
+			fs.createReadStream( archivePath ),
+			{
+				filename: 'local-env-site-1.zip',
+				contentType: 'application/zip',
+			},
+		],
+	];
+
+	try {
+		const rawResponse = await wpcom.req.post< CreateSiteResponse >( {
```
fredrikekelund:
```suggestion
		const rawResponse = await wpcom.req.post( {
```
The default response type is `unknown`, which is actually accurate. We can't guarantee anything about the data type until it's parsed by zod.

cli/commands/preview/lib/api.ts
```diff
@@ -0,0 +1,127 @@
...
+		const errorMessage = error instanceof Error ? error.message : 'Unknown error occurred';
+		throw new LoggerError( `Failed to upload archive: ${ errorMessage }`, action );
+	}
+}
+
+async function checkSiteStatus( siteId: number, token: string ): Promise< boolean > {
+	const wpcom = new WPCOM( token );
+
+	try {
+		const rawResponse = await wpcom.req.get< StatusResponse >( '/jurassic-ninja/status', {
```
fredrikekelund:
```suggestion
		const rawResponse = await wpcom.req.get( '/jurassic-ninja/status', {
```
Again, no guarantees until we've parsed the data with zod.

cli/commands/preview/lib/api.ts
```diff
@@ -0,0 +1,127 @@
...
+		errorMap: () => ( { message: 'Invalid site status' } ),
+	} ),
+	domain_name: z.string(),
+	atomic_site_id: z.number().int().positive(),
+	is_deleted: z.string(),
+} );
+
+export type CreateSiteResponse = z.infer< typeof CreateSiteResponseSchema >;
+export type StatusResponse = z.infer< typeof StatusResponseSchema >;
+
```
fredrikekelund:
```suggestion
```
Let's remove these per my other comments about not assuming the response types without validating.

fredrikekelund (commented):
Shaping up nicely 👍

I left a number of comments on the code. Functionality-wise, I'm still getting errors when passing the `--output-format=json` option.

cli/commands/preview/tests/api.test.ts
```diff
@@ -0,0 +1,233 @@
...
+			await expect( uploadArchive( mockArchivePath, mockToken, mockAction ) ).rejects.toThrow(
+				LoggerError
+			);
+			await expect( uploadArchive( mockArchivePath, mockToken, mockAction ) ).rejects.toMatchObject(
+				{
+					message: 'Invalid API response',
+					action: mockAction,
+				}
+			);
+		} );
```
fredrikekelund: Seems like these two cases are testing more or less the same thing.

cli/commands/preview/tests/create.test.ts
```diff
@@ -1,82 +1,325 @@
...
+		registerCommand( program );
+
 		await program.parseAsync( [ 'node', 'test', 'go', mockFolder ] );
 
-		expect( mockLogger.reportError ).toHaveBeenCalledWith( mockError.message );
+		expect( mockLogger.reportError ).toHaveBeenCalled();
+		expect( mockErrorData ).toHaveProperty( 'message', errorMessage );
+		expect( mockErrorData ).toHaveProperty( 'action', 'validate' );
+		expect( createArchive ).not.toHaveBeenCalled();
 	} );
```
fredrikekelund: Not wrong per se, but this strikes me as an excessive set of test cases. I assume this test code was written by AI, and it should be easy enough to modify with AI, too. Still, I think we can reduce this to two or three test cases and it would give us the same assurances in practice.

cli/commands/preview/lib/validation.ts
```diff
@@ -0,0 +1,24 @@
...
+function hasWpContentDirectory( projectPath: string ): boolean {
+	return fs.existsSync( path.join( projectPath, 'wp-content' ) );
+}
+
+export function validateSiteFolder( siteFolder: string, action: string ): true {
+	if ( ! fs.existsSync( siteFolder ) ) {
+		throw new LoggerError( `Folder not found: ${ siteFolder }`, action );
+	}
+
+	if ( ! isWordPressDirectory( siteFolder ) && ! hasWpContentDirectory( siteFolder ) ) {
```
fredrikekelund: This is never going to be true, since `isWordPressDirectory` checks if the folder contains a `wp-content` folder.

cli/commands/preview/lib/appdata.ts
bcotrim: I did, but I re-tested and I didn't find the same issue, so I think it's safe to remove now.

fredrikekelund (approved):

PR #1131, opened 2025-03-26
Fix metrics job errors

WunderBart:
Checking CI errors from https://github.com/Automattic/studio/pull/1123

This PR:
- Fixes npm install error caused by GH API rate limiting when downloading SQLiteCommand release.
- Fixes test error due to insufficient total timeout.
- Fixes errors due to results file name mismatch.
- Uses `tests.step` instead of logs in tests for clarity.
- Add color to CI logs for clarity.
- Uses `npm ci` instead of `npm install` for stability.
- Saves time by limiting browser install to chromium only since the rest is not needed.
- Adds a few more small tweaks :)

cc: @youknowriad

### DIFF
--- a/src/lib/sqlite-command-release.ts
+++ b/src/lib/sqlite-command-release.ts
@@ -10,7 +10,17 @@ export async function getLatestSQLiteCommandRelease(): Promise< GithubRelease > 
 	const url = 'https://api.github.com/repos/automattic/wp-cli-sqlite-command/releases/latest';
 	console.log( '[sqlite-command] Fetching from:', url );
 
-	const response = await fetch( url );
+	const headers: HeadersInit = {
+		Accept: 'application/vnd.github.v3+json',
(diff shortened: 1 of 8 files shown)


src/lib/sqlite-command-release.ts
```diff
@@ -7,8 +7,36 @@ export interface GithubRelease {
...
+	}
+
+	const data = await response.json();
+	console.log( '[sqlite-command] Release data:', {
+		tag_name: data.tag_name,
+		assets: data.assets?.length,
+		download_url: data.assets?.[ 0 ]?.browser_download_url,
+	} );
+
+	return data;
```
youknowriad: I think this code is also used in the production application, should we remove the console logs there to avoid the unnecessary noise?

youknowriad (approved):
Great, let's bring this back into the original PR.

PR #1132, opened 2025-03-26
Implement CLI installer for macOS

fredrikekelund:
## Related issues

- Fixes STU-230

## Proposed Changes

- Add a `src/lib/install-cli.ts` module that creates a `studio` symlink in `/usr/local/bin` targeting `bin/studio-cli.sh` (implemented in https://github.com/Automattic/studio/issues/1111)
- Modified `forge-config.ts` to build the main CLI JS file and to include it in the Electron app resources
- Modified `bin/studio-cli.sh` to execute the correct JS file
- Added an `Install CLI…` entry to the main application menu (only displayed on macOS and not in production builds)

## Testing Instructions

1. Apply the following diff:

```diff
diff --git a/src/menu.ts b/src/menu.ts
index 48fe7a15..33aaaa5c 100644
--- a/src/menu.ts
+++ b/src/menu.ts
@@ -94,7 +94,7 @@ function getAppMenu(
                                                sendIpcEventToRenderer( 'user-settings' );
                                        },
                                },
-                               ...( process.platform === 'darwin' && process.env.NODE_ENV === 'development'
+                               ...( process.platform === 'darwin'
                                        ? [
                                                        {
                                                                label: __( 'Install CLI…' ),
```

2. Build the app by running `npm run make`
3. Once the build finishes, open `out/make/Studio-[version].dmg`
4. Install Studio from the DMG file
5. Open Studio
6. Open the application menu in the top left corner of the screen (the one labelled **Studio**) and click `Install CLI…`
7. Enter your password when prompted
8. Ensure that the installation finishes successfully
9. Open a terminal window
10. Run `studio`
11. Ensure that you see the following output:

```
Usage: studio [options] [command]

Studio by WordPress.com CLI

Options:
  -V, --version             output the version number
  --output-format [format]  Specify a non-standard output format
  -h, --help                display help for command

Commands:
  go [folder]               Start a new WordPress environment in the specified folder (defaults to current directory)
  help [command]            display help for command
```

## Pre-merge Checklist

### DIFF
--- a/bin/install-studio-cli.sh
+++ b/bin/install-studio-cli.sh
@@ -1,15 +1,18 @@
 #!/bin/sh
 
+# This script is used to install the Studio CLI on macOS. It creates a symlink at CLI_SYMLINK_SOURCE
+# (e.g. /usr/local/bin/studio) pointing to the packaged Studio CLI JS file at CLI_SYMLINK_TARGET.
+
 # Exit if any command fails
 set -e
 
-if [ -z "$INSTALLED_CLI_PATH" ] || [ -z "$PACKAGED_PATH" ]; then
-	echo "Error: INSTALLED_CLI_PATH and PACKAGED_PATH environment variables must be set"
+if [ -z "$CLI_SYMLINK_SOURCE" ] || [ -z "$CLI_SYMLINK_TARGET" ]; then
+	echo "Error: CLI_SYMLINK_SOURCE and CLI_SYMLINK_TARGET environment variables must be set"
 	exit 1
 fi
 
-DIRECTORY_PATH=$(dirname "$INSTALLED_CLI_PATH")
+DIRECTORY_PATH=$(dirname "$CLI_SYMLINK_SOURCE")
 
-rm -f "$INSTALLED_CLI_PATH"
+rm -f "$CLI_SYMLINK_SOURCE"
 mkdir -p "$DIRECTORY_PATH"
-ln -s "$PACKAGED_PATH" "$INSTALLED_CLI_PATH"
+ln -s "$CLI_SYMLINK_TARGET" "$CLI_SYMLINK_SOURCE"
--- a/bin/install-studio-cli.sh
+++ b/bin/install-studio-cli.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# Exit if any command fails
+set -e
+
+if [ -z "$INSTALLED_CLI_PATH" ] || [ -z "$PACKAGED_PATH" ]; then
+	echo "Error: INSTALLED_CLI_PATH and PACKAGED_PATH environment variables must be set"
+	exit 1
+fi
+
+DIRECTORY_PATH=$(dirname "$INSTALLED_CLI_PATH")
+
+rm -f "$INSTALLED_CLI_PATH"
+mkdir -p "$DIRECTORY_PATH"
+ln -s "$PACKAGED_PATH" "$INSTALLED_CLI_PATH"
--- a/src/lib/install-cli.ts
+++ b/src/lib/install-cli.ts
@@ -1,16 +1,15 @@
-import { app, dialog } from 'electron';
+import { dialog } from 'electron';
 import { mkdir, readlink, symlink, unlink } from 'node:fs/promises';
 import path from 'node:path';
 import { __ } from '@wordpress/i18n';
 import { sudoExec } from 'src/lib/sudo-exec';
 import { getMainWindow } from 'src/main-window';
+import { getResourcesPath } from 'src/storage/paths';
 import packageJson from '../../package.json';
 
 const installedCLIPath = '/usr/local/bin/studio';
 
-const binPath = app.isPackaged
-	? path.resolve( app.getAppPath(), '../bin' )
-	: path.resolve( app.getAppPath(), 'bin' );
+const binPath = path.join( getResourcesPath(), 'bin' );
 const packagedPath = path.join( binPath, 'studio-cli.sh' );
 const installScriptPath = path.join( binPath, 'install-studio-cli.sh' );
 
--- a/src/lib/install-cli.ts
+++ b/src/lib/install-cli.ts
@@ -13,7 +13,7 @@ const binPath = path.join( getResourcesPath(), 'bin' );
 const packagedPath = path.join( binPath, 'studio-cli.sh' );
 const installScriptPath = path.join( binPath, 'install-studio-cli.sh' );
 
-export async function installCLIAction() {
+export async function installCLIWithConfirmation() {
 	try {
 		await installCLI();
 		const mainWindow = await getMainWindow();
--- a/src/lib/install-cli.ts
+++ b/src/lib/install-cli.ts
@@ -1,0 +1,75 @@
+import { app, dialog } from 'electron';
+import { mkdir, readlink, symlink, unlink } from 'node:fs/promises';
+import path from 'node:path';
+import { __ } from '@wordpress/i18n';
+import { sudoExec } from 'src/lib/sudo-exec';
+import { getMainWindow } from 'src/main-window';
+import packageJson from '../../package.json';
+
+const installedCLIPath = '/usr/local/bin/studio';
+
+const binPath = app.isPackaged
+	? path.resolve( app.getAppPath(), '../bin' )
+	: path.resolve( app.getAppPath(), 'bin' );
+const packagedPath = path.join( binPath, 'studio-cli.sh' );
+const installScriptPath = path.join( binPath, 'install-studio-cli.sh' );
+
+export async function installCLIAction() {
+	try {
+		await installCLI();
+		const mainWindow = await getMainWindow();
+		await dialog.showMessageBox( mainWindow, {
+			type: 'info',
+			title: __( 'CLI Installed' ),
+			message: __( 'The CLI has been installed successfully.' ),
+		} );
+	} catch ( error ) {
+		console.error( 'Failed to install CLI', error );
+
+		const mainWindow = await getMainWindow();
+		await dialog.showMessageBox( mainWindow, {
+			type: 'error',
+			title: __( 'Failed to install CLI' ),
+			message: __(
+				'Please try again and ensure you grant Studio admin permissions when prompted.'
+			),
+		} );
+	}
+}
+
+// Install the command line tool on macOS.
+async function installCLI(): Promise< void > {
+	if ( process.platform !== 'darwin' ) {
+		return;
+	}
+
+	const installedPath = await getResolvedInstallPath();
+
+	if ( installedPath === packagedPath ) {
+		return;
+	}
+
+	try {
+		const directoryPath = path.dirname( installedCLIPath );
+
+		await unlink( installedCLIPath );
+		await mkdir( directoryPath, { recursive: true } );
+		await symlink( packagedPath, installedCLIPath );
+	} catch ( e ) {
+		await sudoExec( `/bin/sh "${ installScriptPath }"`, {
+			name: packageJson.productName,
+			env: {
+				INSTALLED_CLI_PATH: installedCLIPath,
+				PACKAGED_PATH: packagedPath,
+			},
+		} );
+	}
+}
+
+async function getResolvedInstallPath(): Promise< string | null > {
+	try {
+		return await readlink( installedCLIPath );
+	} catch {
+		return null;
+	}
+}
--- a/src/menu.ts
+++ b/src/menu.ts
@@ -4,11 +4,11 @@ import { openAboutWindow } from 'src/about-menu/open-about-menu';
 import { BUG_REPORT_URL, FEATURE_REQUEST_URL } from 'src/constants';
 import { getDocsLink } from 'src/hooks/use-docs-link';
 import { sendIpcEventToRenderer } from 'src/ipc-utils';
-import { installCLIOnMacOSWithConfirmation } from 'src/lib/install-cli-macos';
 import { getUserLocaleWithFallback } from 'src/lib/locale-node';
 import { shellOpenExternalWrapper } from 'src/lib/shell-open-external-wrapper';
 import { promptWindowsSpeedUpSites } from 'src/lib/windows-helpers';
 import { getMainWindow } from 'src/main-window';
+import { installCLIOnMacOSWithConfirmation } from 'src/modules/cli/lib/install-macos';
 import { isUpdateReadyToInstall, manualCheckForUpdates } from 'src/updates';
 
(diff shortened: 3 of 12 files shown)


src/storage/paths.ts
```diff
@@ -26,7 +26,7 @@ export function getSiteThumbnailPath( siteId: string ): string {
 
 export function getResourcesPath(): string {
 	if ( process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test' ) {
-		return process.cwd();
+		return app.getAppPath();
```
fredrikekelund, line 29: This is only tangentially related to the core functionality of this PR. I made the change because there's no guarantee (that I'm aware of) that the `cwd` is the root folder of the app code, whereas the `app.getAppPath()` return value is normalized by Electron internals.

bin/install-studio-cli.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# Exit if any command fails
+set -e
+
+if [ -z "$INSTALLED_CLI_PATH" ] || [ -z "$PACKAGED_PATH" ]; then
```
nightnei: Maybe it's just personal, but for me, at a glance, it was a bit difficult to understand what these variables about.
It's NIT, but just as proposition to have:

`CLI_TARGET_PATH`
`CLI_SYMLINK_PATH`

Also, two script names as:
`bin/install-studio-cli.sh`
`src/lib/install-cli.ts`
looks a bit confusing. I would propose making the name of this script more specific,
`bin/install-studio-cli.ts`
`src/lib/install-studio-cli-symlink.sh`

src/lib/install-cli.ts
```diff
@@ -0,0 +1,74 @@
+import { dialog } from 'electron';
```
nightnei: Just thoughts out loud - we have `src` folder which contains totally everything, maybe we could split it a bit.
Actually, I had thoughts even before, that we don't have separating between frontend and backend. And now we are adding cli things there too.
For example, as for me, this hierarchy of folders looks not nice:
`src`:
--- components
--- about-menu
--- lib/<cli-related-things>

Since we are just starting CLI project, maybe we could consider separating at least CLI things from everything else?
For example create separate `cli` folder under `src`:
`src`:
--- cli:
--- --- install.ts
--- --- lib / sudo-exec.ts

Also, one more thing - `lib` folder is huge already, so looks like it's a sign to think about splitting it.

WDYT? Mostly interesting to see your thoughts, to be on the same page.

src/menu.ts
```diff
@@ -93,6 +94,14 @@ function getAppMenu(
 						sendIpcEventToRenderer( 'user-settings' );
 					},
 				},
+				...( process.platform === 'darwin' && process.env.NODE_ENV === 'development'
+					? [
+							{
+								label: __( 'Install CLI…' ),
```
nightnei, line 100: Are we going to have this button in prod when we ungate feature?
I thought that installing CLI will be done automattically, not manually. Maybe I missed this info somewhere, could you please explain?

So if this is tmp dev button, then we should avoid translation request.
```suggestion
								label: 'Install CLI…',
```

src/lib/install-cli.ts
nightnei: Also, I would propose to rename it as `install-macos-studio-cli.ts`

bin/install-studio-cli.sh
nightnei: Update: as I menthioned renaming below in the feedback, I would update it with:
`bin/install-macos-studio-cli.ts`
`src/lib/install-macos-studio-cli-symlink.sh`

src/lib/install-cli.ts
```diff
@@ -0,0 +1,74 @@
...
+import { getResourcesPath } from 'src/storage/paths';
+import packageJson from '../../package.json';
+
+const installedCLIPath = '/usr/local/bin/studio';
+
+const binPath = path.join( getResourcesPath(), 'bin' );
+const packagedPath = path.join( binPath, 'studio-cli.sh' );
+const installScriptPath = path.join( binPath, 'install-studio-cli.sh' );
+
+export async function installCLIWithConfirmation() {
```
nightnei: I think `WithConfirmation` is redundant, sicne it's specific part of installation for macOS. So I would replace this with menthioning `macOS` in the name. And it will be easier to read this funciton name inside `src/menu.ts`. 
```suggestion
export async function installMacOsCLI() {
```

src/lib/install-cli.ts
```diff
@@ -0,0 +1,74 @@
...
+	try {
+		await installCLI();
+		const mainWindow = await getMainWindow();
+		await dialog.showMessageBox( mainWindow, {
+			type: 'info',
+			title: __( 'CLI Installed' ),
+			message: __( 'The CLI has been installed successfully.' ),
+		} );
+	} catch ( error ) {
+		console.error( 'Failed to install CLI', error );
```
nightnei: I am thinking - should we log it to Sentry, WDYT?

nightnei (approved):
In general, looks great 👍
Since I saw this code first time, I noticed that some parts took some time to realize how it works. So I would like to propose some changes in naming. To be clear and simplify understand what is in my head, I created [this small PR](https://github.com/Automattic/studio/pull/1141) :) To simplify discussion.

Fredrik, feel totally free to merge/ignore/discuss the changes in that PR :) 

Only two things that are not covered in that PR is:
1. Do we need to log to Sentry if installation failed?
2. WDYT about my thoughts about encapsulation CLI code to separate folder?

In general the PR LGTM, so I am approving it 👍

bin/install-studio-cli.sh
nightnei: All the aboved covered in [this PR](https://github.com/Automattic/studio/pull/1141), to simplify understanding my proposition in practice

src/lib/install-cli.ts
nightnei: Covered in [this PR](https://github.com/Automattic/studio/pull/1141), to simplify understanding my proposition in practice and a few more changes to variable names

src/menu.ts
fredrikekelund: We can install the CLI automatically on Windows by modifying the `PATH` environment variable. However, not on macOS because we can't modify the `PATH` environment variable (persistently and reliably), and admin privileges are usually needed to write to directories that are already included in `$PATH`.

src/lib/install-cli.ts
fredrikekelund: We'll soon have more Studio app code that interfaces with the CLI, and I'm on board with logically grouping that stuff. 

I moved this module to `src/modules/cli/lib/install-macos.ts`.

src/lib/install-cli.ts
fredrikekelund: I quite like the `WithConfirmation` suffix. This function is simply a wrapper around `installCLI` with confirmation modals, which the name clarifies.

fredrikekelund: I'm onboard with adding the "macOS" part to the function name, though 👍

fredrikekelund:
Thanks for the review, @nightnei 👍 I took inspiration from your changes in https://github.com/Automattic/studio/pull/1141, added some comments, changed some symbol names, and moved some files. 

Hopefully, the changes should be more clear now.

src/menu.ts
nightnei: So installation on MacOS will be only manual, with clicking on "Install CLI..." button? We won't make attempt to install it on running Studio and showing popup with providing sudo access?

nightnei:
>Thanks for the review, @nightnei 👍 I took inspiration from your changes in https://github.com/Automattic/studio/pull/1141, added some comments, changed some symbol names, and moved some files.

> Hopefully, the changes should be more clear now.

Yeah, it's even better than my suggestions 👍

Just not to lose track of my comments, I’ll mention the two that are left:
1. https://github.com/Automattic/studio/pull/1132#discussion_r2016880028
2. https://github.com/Automattic/studio/pull/1132#discussion_r2016583115

src/menu.ts
fredrikekelund: That's correct. We'll announce the CLI in the What's New modal and include instructions for installing it, but we won't prompt users automatically. 

This decision was made, in part, because the CLI will have a limited feature set in this first iteration

src/lib/install-cli.ts
fredrikekelund: Yep, we should 👍 >90% of errors will likely be because users didn't give permissions, but we can silence those in Sentry.

nightnei:
Also tested with the latest changes ATM - everything works great:
<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2025-03-27 at 15 14 27
<td>Screenshot 2025-03-27 at 15 14 40
</tr>
</table>

Also, I see symlink in bin folder:
Screenshot 2025-03-27 at 15 15 19

src/lib/install-cli.ts
nightnei: >\>90% of errors will likely be because users didn't give permissions, but we can silence those in Sentry.

Also was thinking about it.
Actually, as an option, I think we can track not provided permissions and don't log it to Sentry, but it's not critical, so let's keep it as is.

src/menu.ts
nightnei: > because the CLI will have a limited feature set in this first iteration

Oh yeah, this point makes definitely more sense to not disturb users with early versions of the feature 👍
Thanks for the information.

bin/install-studio-cli.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ -z "$CLI_SYMLINK_SOURCE" ] || [ -z "$CLI_SYMLINK_TARGET" ]; then
+	echo "Error: CLI_SYMLINK_SOURCE and CLI_SYMLINK_TARGET environment variables must be set"
+	exit 1
+fi
+
+DIRECTORY_PATH=$(dirname "$CLI_SYMLINK_SOURCE")
+
+rm -f "$CLI_SYMLINK_SOURCE"
+mkdir -p "$DIRECTORY_PATH"
+ln -s "$CLI_SYMLINK_TARGET" "$CLI_SYMLINK_SOURCE"
```
mcsf, line 18: In `ln` terminology, the source is the original file and the target is the symlink, making this file particularly confusing to read. :) Let's flip the variable names.

bin/install-studio-cli.sh
```diff
@@ -0,0 +1,18 @@
...
+# This script is used to install the Studio CLI on macOS. It creates a symlink at CLI_SYMLINK_SOURCE
+# (e.g. /usr/local/bin/studio) pointing to the packaged Studio CLI JS file at CLI_SYMLINK_TARGET.
+
+# Exit if any command fails
+set -e
+
+if [ -z "$CLI_SYMLINK_SOURCE" ] || [ -z "$CLI_SYMLINK_TARGET" ]; then
+	echo "Error: CLI_SYMLINK_SOURCE and CLI_SYMLINK_TARGET environment variables must be set"
+	exit 1
+fi
```
mcsf, line 12: Nit, but writing to stderr should be good practice:

```sh
echo >&2 "Error: ..."
```

bin/install-studio-cli.sh
```diff
@@ -0,0 +1,18 @@
...
+
```
mcsf, line 13: I suggest adding a check for the existence of the file for which we want to create a symlink:

```sh
if ! [ -f "$file" ]; then
  echo >&2 "Error: ..."
  exit 1
fi
```

bin/install-studio-cli.sh
```diff
@@ -0,0 +1,18 @@
...
+DIRECTORY_PATH=$(dirname "$CLI_SYMLINK_SOURCE")
+
+rm -f "$CLI_SYMLINK_SOURCE"
```
mcsf, line 16: I understand the point of `-f` is to silently ignore cases where the symlink doesn't exist yet, but I wonder if we could guard the `rm` such that it will only delete if the file is a symlink:

```sh
[ -h "$file" ] && rm "$file"
```

(yes, this is me feeling nervous about a `rm -f` command being run as root 😄)

bin/install-studio-cli.sh
mcsf: OK, after reading `man ln` I also read `man readlink`, and now my world is a little gloomier, because these two programs seem to have opposite meanings for "target":

> The ln utility creates a new directory entry (linked file) for the file name specified by target_file.  The target_file will be created with the same file modes as the source_file. 
> — **ln**

> When invoked as readlink, only the target of the symbolic link is printed.
> — **readlink**

So I think we should avoid source/target terminology and choose something unambiguous like source/symlink, file/symlink, program/symlink, etc.

fredrikekelund: I didn't stop to think close enough about the docs when writing this, but I remember feeling confused about the terminology. Now I know why 😄 To add to the confusion, the GNU manpage for `ln` [apparently](https://unix.stackexchange.com/a/616744/106354) flips the meaning of `target`. The node.js `fs.symlink` function [does the same](https://nodejs.org/docs/latest/api/fs.html#fspromisessymlinktarget-path-type).

It seems much more commonsensical that `target` would refer to the destination rather than the origin, but I take your point that we need to consider this fact in a bash script running only on macOS.

bin/install-studio-cli.sh
fredrikekelund: The path provided by Studio is static, so I struggle to see how the current logic would go wrong. However, your approach is elegant, so I don't see why not 👍

bin/install-studio-cli.sh
fredrikekelund: Do you mind elaborating on your reasoning here?

mcsf: Of course, sorry. Many of these defensive gestures come from the fact that there is an unavoidable  "split" in the program whereby the JS runtime needs to call an external shell script. Even if the latter is only intended to be called with a very specific pair of arguments, it's still a shell script lying around in the filesystem, susceptible of being run by anyone unintentionally.

Thus, things like ensuring that the source file exists, or that the target is either a symlink or non-existent, all that gives me a little peace of mind. Feel free to disagree and dismiss my recommendation! :)

Despite my soft spot for shell scripts, the real issue here is that apparently we don't have a decent way of running JS jobs / microtasks with elevated privileges in a Node/Electron stack — that's disappointing.

bin/install-studio-cli.sh
mcsf: > The path provided by Studio is static

I suppose the strictest version would be:

```sh
[ "$1" = "/usr/local/bin/studio" ] || exit 1
```

You may ask: why not just skip passing that argument, then? Because this makes it immediately obvious to any reader that this script is tied to a particular use, prompting the reader to go look for its JS counterpart.

PR #1135, opened 2025-03-26
Fix notification wording to skip the part about the start

### LINKED ISSUE #1095: Studio incorrectly shows success notice after failed server start
### Quick summary

Try throwing an error somewhere during a site's initialisation. The error is correctly shown as an alert box, but a system notification claims "your new site is up and running":

Image

 

### Steps to reproduce

The way I stumbled upon it was that the `sudoExec` inside `writeHostsFile` was throwing. So try adding a throw anywhere in there, then create a new site with a custom domain.

No success notice shown, just the error.

### What actually happened

Both shown.

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Platform

_No response_

### Logs or notes

_No response_

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/studio/issues/1095

## Proposed Changes

Currently, when a user creates or imports a site that fails to be started, there is a system notification saying the site is "up and running". The purpose of that notification was to let the user that the site was created or imported in case they changed focus to another app.

I propose to keep things simple and fix the notification wording to:
- skip the part saying if the site was started or not. If it wasn't started, there would be an error dialog shown when user focused the Studio again.
- differentiate the notification for site created and imported

## Testing Instructions

1. Apply a code that throws an error during site start, e.g.:

```
diff --git a/src/site-server.ts b/src/site-server.ts
index ed5dc66a..9920d195 100644
--- a/src/site-server.ts
+++ b/src/site-server.ts
@@ -166,6 +166,10 @@ export class SiteServer {
                this.server = new SiteServerProcess( options );
                await this.server.start();
 
+               throw new Error(
+                       `Test throwing error.`
+               );
+
                if ( this.server.options.port === undefined ) {
                        throw new Error( 'Server started with no port' );
                }

```

2. Add a new site, and move focus to another app
3. Confirm notification was displayed
4. Go back to the Studio and confirm the error dialog
5. Repeat 2-4 by using import file

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-add-site.ts
+++ b/src/hooks/use-add-site.ts
@@ -130,14 +130,19 @@ export function useAddSite() {
 							isNewSite: true,
 						} );
 						clearImportState( newSite.id );
+
+						getIpcApi().showNotification( {
+							title: newSite.name,
+							body: __( 'Your new site was imported' ),
+						} );
 					} else {
+						getIpcApi().showNotification( {
+							title: newSite.name,
+							body: __( 'Your new site was created' ),
+						} );
+
 						await startServer( newSite.id );
 					}
-
-					getIpcApi().showNotification( {
-						title: newSite.name,
-						body: __( 'Your new site is up and running' ),
-					} );
 				}
 			);
 		} catch ( e ) {


fredrikekelund (approved):
LGTM 👍

Creating a new site is usually a very quick operation (provided the WP version is already downloaded), so the UX of displaying a success-ish notification 2 seconds after submitting the form and then immediately displaying an error modal isn't perfect (even if this PR is still an improvement from what we have now).

We could return a `success` boolean from `startServer` and `importFile`, and only display the notification if they run successfully. Did you consider this option, @wojtekn?

ivan-ottinger (approved):
> Creating a new site is usually a very quick operation (provided the WP version is already downloaded), so the UX of displaying a success-ish notification 2 seconds after submitting the form and then immediately displaying an error modal isn't perfect (even if this PR is still an improvement from what we have now).

One difference I see there is on larger imports where it takes longer than those two seconds. When I tested a 500 MB import, it took around thirty seconds.

But yeah, I share your point of view there.

wojtekn:
> Creating a new site is usually a very quick operation (provided the WP version is already downloaded), so the UX of displaying a success-ish notification 2 seconds after submitting the form and then immediately displaying an error modal isn't perfect (even if this PR is still an improvement from what we have now).

The purpose of that notification is to inform users about a site that was created or imported. If users have already switched focus to another app, a notification will let them easily return to Studio. If the start fails, they will see the dialog informing them.

> We could return a success boolean from startServer and importFile, and only display the notification if they run successfully. Did you consider this option, @wojtekn?

The notification is supposed to inform about the created or imported site, not about the site's start. I considered displaying different notifications based on the status of the start operation, but it seems overkill. Also, in such a case, users would see a notification reporting that the start failed, would return to Studio, and would see another dialog reporting a failed start again.

Given that it is rare for a site to fail to start after it was created or imported, I would leave it as is.

PR #1136, opened 2025-03-26
Studio: Add constant to force modal in select versions

katinthehatsite:
## Related issues

STU-271

## Proposed Changes

This PR adds a config file which allows you to control the `What's new modal` display for specific versions of Studio besides minor and major. 

## Testing Instructions

* Pull the changes of this branch
* Apply this diff

```diff --git a/src/modules/whats-new/config/whats-new-config.ts b/src/modules/whats-new/config/whats-new-config.ts
index 3a0e0384..42f41e60 100644
--- a/src/modules/whats-new/config/whats-new-config.ts
+++ b/src/modules/whats-new/config/whats-new-config.ts
@@ -11,7 +11,7 @@
  * regardless of the normal version comparison logic.
  */
 export const forceWhatsNewVersions = [
-       '1.3.9-beta1', // Example: Important security patch
+       '1.3.9-beta2', // Example: Important security patch
 ];
```
* Open package.json and change the app version to `1.3.9-beta2`
* Run `npm install`
* Start the app with `STUDIO_WHATS_NEW_SECTION=true npm start`
* Confirm that you can see the modal appearing automatically when the app starts
* Close the modal
* Type `rs` in the terminal where you have Studio open
* Confirm that once the app restarts, you do not see the modal again 

## Pre-merge Checklist

### DIFF
--- a/src/modules/whats-new/config/whats-new-config.ts
+++ b/src/modules/whats-new/config/whats-new-config.ts
@@ -1,0 +1,25 @@
+/**
+ * Configuration for the What's New feature
+ * 
+ * This file contains settings that control when the What's New modal appears.
+ * Add version numbers to the forceWhatsNewVersions array to make the modal
+ * appear for those specific versions, even if they're just patch updates.
+ */
+
+/**
+ * List of versions that should always show the What's New modal,
+ * regardless of the normal version comparison logic.
+ */
+export const forceWhatsNewVersions = [
+	'1.3.9-beta1', // Example: Important security patch
+];
+
+/**
+ * Determines if the What's New modal should be forced for a specific version
+ *
+ * @param version The current application version
+ * @returns True if the modal should be forced for this version
+ */
+export const shouldForceWhatsNew = ( version: string ): boolean => {
+	return forceWhatsNewVersions.includes( version );
+}; 
--- a/src/modules/whats-new/config/whats-new-config.ts
+++ b/src/modules/whats-new/config/whats-new-config.ts
@@ -1,6 +1,6 @@
 /**
  * Configuration for the What's New feature
- * 
+ *
  * This file contains settings that control when the What's New modal appears.
  * Add version numbers to the forceWhatsNewVersions array to make the modal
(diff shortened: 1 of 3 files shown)


src/modules/whats-new/config/whats-new-config.ts
```diff
@@ -0,0 +1,25 @@
...
+	'1.3.9-beta1', // Example: Important security patch
+];
+
+/**
+ * Determines if the What's New modal should be forced for a specific version
+ *
+ * @param version The current application version
+ * @returns True if the modal should be forced for this version
+ */
+export const shouldForceWhatsNew = ( version: string ): boolean => {
```
wojtekn, line 21: Why couldn't it be just a flag set to `true` or `false`?

katinthehatsite: I am considering this option as well. With this approach, I was thinking that it will be easier to see in what versions we forced the modal. However, this approach is more robust and might not be necessary.

katinthehatsite: Also, I commented here https://github.com/Automattic/studio/pull/1138#issuecomment-2757682930 - the advantage of this approach is that once the user has seen the modal with a specific version and closed it, they won't be seeing it again on app restart compared to using the feature flag that uses the local state to track that. However, it might not be a huge downside. 

In any case, I opened two different PR and explained suggestions in each of them, we can proceed with whichever one is more suitable.

wojtekn: I didn't mean feature flag - Studio uses them for a different purpose. 

I meant to have `shouldForceWhatsNew` a simple trust/false flag we could set in the code. 

See https://github.com/Automattic/studio/pull/1139

katinthehatsite: I see, I did not see this draft but the PR is here: https://github.com/Automattic/studio/pull/1140

katinthehatsite:
We will go for a simpler approach so let's close this one.

PR #1134, opened 2025-03-26
What's New Section: Fetch `lastSeenVersion` at the `App` component instead of the store

ivan-ottinger:
## Related issues

- Fixes STU-308-linear-issue.

## Proposed Changes

- fetch `lastSeenVersion` at the `App` component instead of the store

## Testing Instructions

1. Check out the PR branch and build the app with `STUDIO_WHATS_NEW_SECTION=true npm start`.
2. Head over to your `appdata-v1.json` and make sure there's no `lastSeenVersion` value stored in it.
3. The `What's New` modal should work as expected, i.e.: it should open on app start.
4. If you run `npm test`, there should be no errors similar to what has Fredrik encountered: https://github.com/Automattic/studio/pull/1110#issuecomment-2754740221.

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1137, opened 2025-03-27
Suggestion: Use RTK-Query for app version reducer

fredrikekelund:
## Related issues

- Related to https://github.com/Automattic/studio/pull/1134

## Proposed Changes

- Change `src/stores/app-version-slice.ts` into an RTK-Query-based store using `createApi`

## Testing Instructions

Follow the testing instructions in https://github.com/Automattic/studio/pull/1134

## Pre-merge Checklist


ivan-ottinger (approved):
That is looking great and all is working correctly. Thank you for suggesting the use of RTK-Query, Fredrik.

I had a moment yesterday when I was considering using it, but then discarded it quickly as it didn't feel like a right approach. But looking at it now from the perspective you have shared on the original PR today, I think it makes sense:

> It doesn't fetch data over the network, but the IPC communication works very similarly to a network request.

Thanks!

PR #1143, opened 2025-03-27
Revert tmp code from #1123

WunderBart:
## Proposed Changes

This reverts commit 036e841d72794d8d69725fafcc0906be0cea7bf1.

## Testing Instructions

CI should pass.

### DIFF
--- a/metrics/tests/site-startup.test.ts
+++ b/metrics/tests/site-startup.test.ts
@@ -1,4 +1,3 @@
-import path from 'path';
 import { test, expect } from '@playwright/test';
 import { E2ESession } from '../../e2e/e2e-helpers';
 import Onboarding from '../../e2e/page-objects/onboarding';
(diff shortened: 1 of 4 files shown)


metrics/tests/site-startup.test.ts
```diff
@@ -30,18 +29,6 @@ test.describe( 'Startup Metrics', () => {
 		} );
 
 		await session.cleanup();
-
-		const video = session.mainWindow.video();
```
youknowriad, line 34: I actually wonder if we should keep this somehow. But I agree it's better to remove the current tmp code :)

youknowriad (approved):
Sorry, my mistake, I merged too quickly :)

metrics/tests/site-startup.test.ts
WunderBart: Not sure how much of a performance impact would video recording be, but we can now actually measure it 😄

WunderBart: Anyhow, let's implement in a separate PR!

PR #1144, opened 2025-03-27
Performance Metrics: Log the results in codevitals

youknowriad:
## Related issues

Fixes DOTCOM-671-linear-issue

## Changes proposed in this Pull Request:

**Big Picture** My goal is to bring metrics tracking and performance dashboard to the tools Studio use daily to avoid performance regressions and work on performance improvements similarly to the [Gutenberg project](https://www.codevitals.run/project/gutenberg) for instance.

Implementing this require three complementary steps:

 1- First step is to have some test runner (e2e test here) that we can launch to compute the metrics.
 2- Second step is to be able to use the same test runner in two separate branches as part of the same CI job. The reason for this is that CI runners capacities are unstable and we can't rely on the absolute numbers. We have to compute relative numbers and compare each commit with a given "base commit".
 3- Third step is to send these values (base branch and commit) to the dashboard (codevitals) to log it and produce the graphs.

The current PR implements the last step. 

### How to test the changes in this Pull Request:

http://codevitals.run/project/studio

### DIFF
--- a/.github/workflows/metrics-comparison.yml
+++ b/.github/workflows/metrics-comparison.yml
@@ -48,7 +48,7 @@ jobs:
 
       # Log performance metrics to CodeVitals when running on trunk
       - name: Log performance metrics to CodeVitals
-        #if: github.event_name == 'push' && github.ref == 'refs/heads/trunk' && !cancelled()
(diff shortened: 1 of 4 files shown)


youknowriad:
Ok, this works I'm going to ship this to start collecting data in trunk.

PR #1141, opened 2025-03-27
Tiny readability changes proposition for: Implement CLI installer for macOS

nightnei:
@fredrikekelund  I was reviewing Implement [CLI installer for macOS](https://github.com/Automattic/studio/pull/1132) and at the end realized that it's better to show my thoughts with the help of PR, to show the clear picture from my head. So this PR is proposition and reflects my comments in that PR.

PR #1148, opened 2025-03-28
macOS CLI script tweaks

fredrikekelund:
## Related issues

- Follow-up to https://github.com/Automattic/studio/pull/1132 and https://github.com/Automattic/studio/pull/1111

## Proposed Changes

- In the POSIX `ln` utility documentation, the `source` and `target` terms refer to the opposite of what you'd expect (if you expect that "target" means link destination and "source" means symlink path). To avoid additional confusion, I've changed the terms to "symlink path" and "packaged path".
- As an additional safety measure, we now abort the installation if something other than a symlink already lives at `/usr/local/bin/studio`
- Tweaked the macOS CLI bootstrap script per @mcsf's recommendations in https://github.com/Automattic/studio/pull/1111.
  - Remove the use of `command` prefixes.
  - Use a static name for the Electron executable file.
  - Verify the existence of the Electron executable file.
  - Use `exec` to replace the shell process with the node process.

> [!TIP]
> An interesting piece of trivia is that the GNU `ln` utility [apparently](https://unix.stackexchange.com/a/616744/106354) uses the opposite meaning for "target".

## Testing Instructions

Same as https://github.com/Automattic/studio/pull/1132

## Pre-merge Checklist

### DIFF
--- a/bin/install-studio-cli.sh
+++ b/bin/install-studio-cli.sh
@@ -1,18 +1,18 @@
 #!/bin/sh
 
-# This script is used to install the Studio CLI on macOS. It creates a symlink at CLI_SYMLINK_SOURCE
-# (e.g. /usr/local/bin/studio) pointing to the packaged Studio CLI JS file at CLI_SYMLINK_TARGET.
+# This script is used to install the Studio CLI on macOS. It creates a symlink at CLI_SYMLINK_PATH
+# (e.g. /usr/local/bin/studio) pointing to the packaged Studio CLI JS file at CLI_SYMLINK_DESTINATION.
 
 # Exit if any command fails
 set -e
 
-if [ -z "$CLI_SYMLINK_SOURCE" ] || [ -z "$CLI_SYMLINK_TARGET" ]; then
-	echo "Error: CLI_SYMLINK_SOURCE and CLI_SYMLINK_TARGET environment variables must be set"
+if [ -z "$CLI_SYMLINK_PATH" ] || [ -z "$CLI_SYMLINK_DESTINATION" ]; then
+	echo >&2 "Error: CLI_SYMLINK_PATH and CLI_SYMLINK_DESTINATION environment variables must be set"
 	exit 1
 fi
 
-DIRECTORY_PATH=$(dirname "$CLI_SYMLINK_SOURCE")
+DIRECTORY_PATH=$(dirname "$CLI_SYMLINK_PATH")
 
-rm -f "$CLI_SYMLINK_SOURCE"
+rm -f "$CLI_SYMLINK_PATH"
 mkdir -p "$DIRECTORY_PATH"
-ln -s "$CLI_SYMLINK_TARGET" "$CLI_SYMLINK_SOURCE"
+ln -s "$CLI_SYMLINK_DESTINATION" "$CLI_SYMLINK_PATH"
--- a/bin/install-studio-cli.sh
+++ b/bin/install-studio-cli.sh
@@ -1,18 +1,18 @@
 #!/bin/sh
 
 # This script is used to install the Studio CLI on macOS. It creates a symlink at CLI_SYMLINK_PATH
-# (e.g. /usr/local/bin/studio) pointing to the packaged Studio CLI JS file at CLI_SYMLINK_DESTINATION.
+# (e.g. /usr/local/bin/studio) pointing to the packaged Studio CLI JS file at CLI_PACKAGED_PATH.
 
 # Exit if any command fails
 set -e
 
-if [ -z "$CLI_SYMLINK_PATH" ] || [ -z "$CLI_SYMLINK_DESTINATION" ]; then
-	echo >&2 "Error: CLI_SYMLINK_PATH and CLI_SYMLINK_DESTINATION environment variables must be set"
+if [ -z "$CLI_SYMLINK_PATH" ] || [ -z "$CLI_PACKAGED_PATH" ]; then
+	echo >&2 "Error: CLI_SYMLINK_PATH and CLI_PACKAGED_PATH environment variables must be set"
 	exit 1
 fi
 
-DIRECTORY_PATH=$(dirname "$CLI_SYMLINK_PATH")
+CLI_DIRECTORY_PATH=$(dirname "$CLI_SYMLINK_PATH")
 
-rm -f "$CLI_SYMLINK_PATH"
-mkdir -p "$DIRECTORY_PATH"
-ln -s "$CLI_SYMLINK_DESTINATION" "$CLI_SYMLINK_PATH"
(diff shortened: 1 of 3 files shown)


src/modules/cli/lib/install-macos.ts
```diff
@@ -27,52 +32,84 @@ export async function installCLIOnMacOSWithConfirmation() {
...
 	if ( process.platform !== 'darwin' ) {
-		return;
+		throw new Error( InstallError.WrongPlatform );
+	}
+
+	const fileType = await getFileType( cliSymlinkPath );
+
+	// Avoid overwriting an existing file if it's not a symlink.
+	if ( fileType === FileType.NonSymlink ) {
+		throw new Error( InstallError.FileAlreadyExists );
```
mcsf: Studio isn't my project, so please take this as opinion beyond the scope of reviewing the shell script. :)

IMO these changes go beyond [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and add unnecessary indirection. `getFileType` isn't used anywhere else, nor is `FileType`. There will be less code to maintain and fewer jumps required of the reader if we inline a throw/catch construct with the system call in plain sight.

src/modules/cli/lib/install-macos.ts
```diff
@@ -27,52 +32,84 @@ export async function installCLIOnMacOSWithConfirmation() {
...
+		if ( error instanceof Error && error.message === InstallError.FileAlreadyExists ) {
+			message = sprintf(
+				/* translators: 1: Installation path */
+				__(
+					'The installation path %1$s is already occupied by a file or directory. Please remove it and try again.'
+				),
+				cliSymlinkPath
+			);
+		}
+
```
mcsf: This fails to propagate legitimate errors occurred during execution of the sudo'd command. Try it yourself by exiting early from `install-studio-cli.sh` with `exit 1`, or even echoing an error and exiting with 1. The code assumes a failure will be due to permissions not being granted, but that can be verified by looking at `error.message`:

|Problem|`error.message`|
|-|-|
|Permission not granted|_User did not grant permission._|
|Non-zero exit with no echoed output|_Command failed: /bin/sh "/Applications/Studio.app/Contents/Resources/bin/install-studio-cli.sh"_|
|Non-zero exit with echoed output|_Command failed: /bin/sh "/Applications/Studio.app/Contents/Resources/bin/install-studio-cli.sh"`\n`Error: Foo_|

mcsf (commented):
Thanks for addressing my feedback! One of my comments here is opinion, but the other one should be addressed, I think.

src/modules/cli/lib/install-macos.ts
fredrikekelund: This is true, but I think it's fair to assume that most errors generated here will be due to missing permissions. Moreover, the logic in `install-studio-cli.sh` is unlikely to generate errors (especially considering that we verify that `cliSymlinkPath` is not occupied before running the script). 

Still, I added some additional error message parsing here for completeness' sake, so we now also distinguish between known permission errors and unknown errors.

src/modules/cli/lib/install-macos.ts
fredrikekelund: That's fair. I've inlined this logic

bin/install-studio-cli.sh
```diff
@@ -1,18 +1,18 @@
 #!/bin/sh
 
-# This script is used to install the Studio CLI on macOS. It creates a symlink at CLI_SYMLINK_SOURCE
-# (e.g. /usr/local/bin/studio) pointing to the packaged Studio CLI JS file at CLI_SYMLINK_TARGET.
+# This script is used to install the Studio CLI on macOS. It creates a symlink at CLI_SYMLINK_PATH
+# (e.g. /usr/local/bin/studio) pointing to the packaged Studio CLI JS file at CLI_PACKAGED_PATH.
```
nightnei, line 4: `CLI_SYMLINK_PATH` and `CLI_PACKAGED_PATH` are great names 👍
Similar to what I proposed [( CLI_SYMLINK_PATH / CLI_TARGET_PATH )](https://github.com/Automattic/studio/pull/1141/files), but `CLI_PACKAGED_PATH` is much clear 👍

src/modules/cli/lib/install-macos.ts
```diff
@@ -1,19 +1,26 @@
...
 
-const cliSymlinkSource = '/usr/local/bin/studio';
+const cliSymlinkPath = '/usr/local/bin/studio';
 
 const binPath = path.join( getResourcesPath(), 'bin' );
-const cliSymlinkTarget = path.join( binPath, 'studio-cli.sh' );
+const cliPackagedPath = path.join( binPath, 'studio-cli.sh' );
 const installScriptPath = path.join( binPath, 'install-studio-cli.sh' );
 
+enum InstallError {
```
nightnei: Hm, it seems, I remember that either Tumblr or Dotcom or both didn't recommend using enums, due to tree shaking issues or something like that. Do you remember something about it? Or maybe I remember something incorrectly 😅

nightnei: Let's wrap strings with translation function.

nightnei (approved):
LGTM and works still well 👍
Left just two comments about enum and translations.

Screenshot 2025-04-08 at 14 39 39

src/modules/cli/lib/install-macos.ts
```diff
@@ -27,52 +34,77 @@ export async function installCLIOnMacOSWithConfirmation() {
...
+
+	try {
+		const stats = await lstat( cliSymlinkPath );
+
+		if ( ! stats.isSymbolicLink() ) {
+			throw new Error( InstallError.FileAlreadyExists );
+		}
+	} catch ( error ) {
+		// File does not exist, do nothing
 	}
```
mcsf: Thanks for addressing my feedback in general.

I don't want to be the one delaying this PR any longer, so please use your best judgment here. I just wanted to point out, though, that — although they are unlikely — `lstat` can potentially throw [13 different errors](https://man7.org/linux/man-pages/man2/lstat.2.html#ERRORS). So this empty `catch` block seems a bit bold to me. This is what I meant earlier with propagating the error.

(Currently traveling back from a meetup, so I won't be very responsive.)

fredrikekelund: Checking for ENOENT errors specifically is fair 👍 I've made that change

src/modules/cli/lib/install-macos.ts
fredrikekelund: These error messages aren't displayed to users, so we shouldn't need to translate them. You're right that we don't need an enum, though. I've changed it to regular constants.

PR #1154, opened 2025-03-31
Settings: add section with database path and general connection instructions

ramonjd:
## Related issues

Temporary fix for https://github.com/Automattic/studio/issues/1033

## Proposed Changes

This PR adds a database section to the setting for users who wish to use a third-party application to manage the local SQLite database.

Such apps might include:

- https://sqlitebrowser.org/
- https://dbeaver.io/
- https://www.beekeeperstudio.io/

These apps are free. I've tested with all of them. Some are better than others. 

Changes:

- exposes the path to the local SQLite file in the settings tabs.
- adds a link to open the file in the local file system
- updates `<SettingsRow />` to conditionally render labels and adjust layout to accommodate new content
- adds basic tests

Screenshot 2025-03-31 at 3 49 22 pm

I've deliberately left out references to specific apps for fear of being accused of promotion. All the ones I tested will associate themselves with and open `*.db` and `*.sqlite` file.

> [!NOTE]
> This PR doesn't preclude bundling an browser-based app with Studio, only provides details for the sake of user choice.

## Testing Instructions

1. Fire up this branch
2. Open the settings tab in Studio
3. Ensure that the open in file system link works
4. Ensure that the copy link works

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1163, opened 2025-04-02
Add Adminer integration and database management features

### LINKED ISSUE #420: Feature Request: Access and edit the database

ramonjd:
## Proposed Changes

Introduces [Adminer Editor](https://www.adminer.org/en/editor/) for table viewing and record editing the SQLite file.

Introduces a "Database" link in the header to open the Adminer app.

It includes a custom configuration for WordPress Studio:

- only shows `wp_` tables
- doesn't allow inserting or deleting records

> [!NOTE]
> It's been advised to use MySQL tools (like Adminer or PhpMyAdmin) when the [SQLite <> MySQL integration](https://github.com/Automattic/sqlite-database-integration) is mature enough. This is to avoid data corruption when the database schema changes. However, I've had advice that it's okay to allow updating existing record values only using the SQLite driver. When the sqlite-database-integration is ready we could allow row deletion, table column updates and other functions I'm hoping.

## TODO

- [x] Add tests
- [ ] Test making some updates via Adminer and exporting database, importing to another WP site.
- [ ]  Go through exceptions and errors and make sure logging is better/appropriate to each case
- [ ] Check if it's the right way to go about it. Maybe it's better in a WP plugin that's pulled it alongside [sqlite-database-integration ](https://github.com/WordPress/sqlite-database-integration)
- [x] Always remove on site deletion (even if the user says to keep files?)
- [x]  Ensure each instance displays the updated site name

https://github.com/user-attachments/assets/ecb8ad0c-7c3b-4fd5-8650-c3a25e1be2dd

## How?

Right now, it copies a vendor package into each site's root folder, and removes it when the site is deleted (even if the user checks the box to keep the WP installation).

An alternative would be to create a dedicated WP mu-plugin and install it so that tool is accessible from WordPress Admin. I tested this method, however I had some trouble with the way Adminer serves its static files. I didn't go deep. 

## Related issues

- (To some extent) Resolves https://github.com/Automattic/studio/issues/420

## Testing Instructions

 Verify that Adminer opens correctly from the "Database" header link and displays only wp_ tables. Ensure no errors occur during data updates (in database and in site) and site deletions.

Run a few sites at a time and test that each has its own Adminer instance.

Delete a site, but keep the local files. Adminer should disappear always.

Test making some updates via Adminer and exporting database, importing to another WP site. (TODO - I haven't yet checked this myself)

## Pre-merge Checklist


src/site-server.ts
```diff
@@ -51,6 +51,7 @@ export async function createSiteWorkingDirectory(
 
 	await purgeWpConfig( wpVersion );
 	await recursiveCopyDirectory( getWordPressVersionPath( wpVersion ), path );
+	await recursiveCopyDirectory( 'vendor/adminer', path + '/adminer' );
```
ramonjd: Might need an adminer.config.php that defines some site constants. Mainly to populate Adminer's header/breadcrumb with the site name for the sake of differentiation.

vendor/adminer/version.txt
```diff
@@ -0,0 +1 @@
+1.0.0
```
ramonjd, line 1: This is so we can force updates in new versions.

vendor/adminer/translations.php
```diff
@@ -0,0 +1,6 @@
+<?php
+				return array(
+					'openSite' => 'Open site',
+					'siteTables' => 'Site tables', 
+				);
```
ramonjd, line 5: Is there a better way to do this? It's clunky, but works. I don't expect many other strings. Adminer has its own translations.

vendor/adminer/index.php
```diff
@@ -0,0 +1,179 @@
...
+			return 5;
+		}
+
+		/**
+		 * Returns Content Security Policy headers.
+		 *
+		 * @return array Array of arrays with directive name in key, allowed sources in value.
+		 */
+		public function csp( $csp = array() ) {
+			return array();
```
ramonjd, line 112: So we can inject inline JS

vendor/adminer/index.php
```diff
@@ -0,0 +1,179 @@
...
+			return ADMINER_SQLITE_DATABASE_PATH;
+		}
+
+		/**
+		 * Returns the custom name for the admin interface.
+		 *
+		 * @return string
+		 */
+		public function name() {
+			return 'WordPress Studio';
```
ramonjd, line 48: Just fallbacks here.

src/site-server.ts
```diff
@@ -231,6 +234,10 @@ export class SiteServer {
 				tlsCert: cert,
 			};
 		}
+
+		if ( oldName !== newName ) {
+			await updateAdminerConfig( this.details );
```
ramonjd, line 239: Ensure we display the right name in the database view always.

src/lib/adminer.ts
```diff
@@ -0,0 +1,128 @@
...
+			const translationsPath = 'vendor/adminer/translations.php';
+			try {
+				await fs.promises.writeFile( translationsPath, translationsPhpContent );
+			} catch ( translationsError ) {
+				throw new Error(
+					'Failed to update adminer translations.php: ' + ( translationsError as Error ).message
+				);
+			}
+			try {
+				await recursiveCopyDirectory( 'vendor/adminer', adminerPath );
```
ramonjd, line 68: This is the foundation: copying Adminer PHP files to every site folder.

As mentioned, I did explore an alternative by wrapping Adminer in a WP plugin and adding it to the mu-plugins folder during wp-env setup, but Adminer doesn't seem to like being rendered in an WP-Admin page. 🤔 It can't serve its own assets. Probably a work around though.

vendor/adminer/index.php
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		/**
+		 * Returns the server name display.
+         * Get server name displayed in breadcrumbs.
```
andrewserong: Tiny formatting nit (spaces vs tabs). I'm not 100% sure the code suggestion here is correct but I copy/pasted from the previous line:

```suggestion
		 * Get server name displayed in breadcrumbs.
```

ramonjd: Let's YOLO it 😄 Thanks!

src/lib/adminer.ts
```diff
@@ -0,0 +1,128 @@
...
+		const adminerPath = nodePath.join( siteDetails.path, 'adminer' );
+		const sourceVersionPath = 'vendor/adminer/version.txt';
+		const destVersionPath = `${ adminerPath }/version.txt`;
+
+		// Check if we need to update
+		let shouldUpdate = false;
+
+		try {
+			const exists = await pathExists( adminerPath );
+			console.log( 'exists', exists );
```
ramonjd, line 27:
```suggestion
```

src/lib/adminer.ts
```diff
@@ -0,0 +1,128 @@
...
+				await recursiveCopyDirectory( 'vendor/adminer', adminerPath );
+			} catch ( copyError ) {
+				throw new Error( `Failed to copy adminer directory: ${ ( copyError as Error ).message }` );
+			}
+
+			// Update config.php with site details.
+			await updateAdminerConfig( siteDetails );
+		}
+	} catch ( error ) {
+		console.error( 'Error setting up adminer:', error );
```
ramonjd, line 77: TODO - go through exceptions and errors and make sure logging is better/appropriate to each case

andrewserong (commented):
Just gave this a quick test and it's working pretty well for me so far, nice work!

> doesn't allow inserting or deleting records

Is this still the case? From the Select view I was able to use the duplicate and clone buttons on selected rows:

image

Other than that, it appears to work well for me. One wrinkle I ran into is with the loading state when clicking the Database link on a site that isn't running yet. Clicking the link starts the site, but since that takes a while, I found it easy to forget that I clicked the button, and then when the site eventually starts, the Database window opened unexpectedly for me. Expected in the sense that it _should_ open, but unexpected because it was a long time after I clicked it 😄

https://github.com/user-attachments/assets/577c4bca-d606-4b4b-82cf-4550d332cff4

I see this is the same behaviour as the _WP admin_ and _Open site_ buttons, though, so it's consistent 🤔

A couple of off the top-of-my-head thoughts:

* Longer-term could we add right-click behaviour to copy the WP admin, Open site, and Database links?
* Should the Database button be disabled if the site isn't running? (I suppose not, since if we change the behaviour we'd probably want to do it for the other buttons/links).
* Or, would it be better to have a Database tab, so we can list out the URL, etc as well as the button to open Adminer?

> Ensure no errors occur during data updates (in database and in site) and site deletions.

I ran into an issue only once where opening Adminer didn't automatically log in, and I was stuck on a login screen that flickered slightly. Unfortunately I haven't been able to reproduce this as when I closed the window and opened it again, the problem was gone. But I'll keep an eye out for it in subsequent testing.

Happy to test further next week!

ramonjd:
Thank you for testing @andrewserong !!

> I was able to use the duplicate and clone buttons on selected rows

Oh, I missed those. Good spotting. I think I'll have to hide them using CSS since Adminer has limited extension options.

>  Expected in the sense that it should open, but unexpected because it was a long time after I clicked 
> I see this is the same behaviour as the WP admin and Open site buttons, though, so it's consistent

Sound like a good follow up issue 🍺 

> I ran into an issue only once where opening Adminer didn't automatically log in, and I was stuck on a login screen that flickered slightly. 

Interesting. I might remove that auto redirect and just allow login. I've seen it once I think too.

> Longer-term could we add right-click behaviour to copy the WP admin, Open site, and Database links?

Another great follow up issue. I like!

> Should the Database button be disabled if the site isn't running? (I suppose not, since if we change the behaviour we'd probably want to do it for the other buttons/links).

Yeah, not sure either. I agree it should be consistent either way. Do you mean only visually disabled? Would we still want a click action to start the server?

> Or, would it be better to have a Database tab, so we can list out the URL, etc as well as the button to open Adminer?

I have https://github.com/Automattic/studio/pull/1154 going, so I could move the content there. 

I don't think it's a bad idea, though I think we shouldn't actively encourage editing the sqlite db directly as it can lead to corruption from what I've heard, but exposing the path isn't really doing much more than what the user can do with their OS's file system browser.

ramonjd:
Thank you for testing @andrewserong !!

> I was able to use the duplicate and clone buttons on selected rows

Oh, I missed those. Good spotting. I think I'll have to hide them using CSS since Adminer has limited extension options.

>  Expected in the sense that it should open, but unexpected because it was a long time after I clicked 
> I see this is the same behaviour as the WP admin and Open site buttons, though, so it's consistent

Sound like a good follow up issue 🍺 

> I ran into an issue only once where opening Adminer didn't automatically log in, and I was stuck on a login screen that flickered slightly. 

Interesting. I might remove that auto redirect and just allow login. I've seen it once I think too.

> Longer-term could we add right-click behaviour to copy the WP admin, Open site, and Database links?

Another great follow up issue. I like!

> Should the Database button be disabled if the site isn't running? (I suppose not, since if we change the behaviour we'd probably want to do it for the other buttons/links).

Yeah, not sure either. I agree it should be consistent either way. Do you mean only visually disabled? Would we still want a click action to start the server?

> Or, would it be better to have a Database tab, so we can list out the URL, etc as well as the button to open Adminer?

I have https://github.com/Automattic/studio/pull/1154 going, so I could move the content there. 

I don't think it's a bad idea, though I think we shouldn't actively encourage editing the sqlite db directly as it can lead to corruption from what I've heard, but exposing the path isn't really doing much more than what the user can do with their OS's file system browser.

andrewserong:
> Do you mean only visually disabled? Would we still want a click action to start the server?

Good question. I was imagining properly disabled, but maybe it's worth trying visually, and a single click starts the server. I think the main thing I'd avoid is where there can be a long time between that first click and a window opening, because by the time it eventually opens it can feel unexpected. This is possibly something to hack around with until it _feels_ right?

Thanks for thinking through all the details!

ramonjd:
Placing this PR on pause until Playground's SQL driver reaches adequate compatibility. 

Work is in progress over https://github.com/Automattic/sqlite-database-integration

The rationale is that to invest the effort into full integration with a mysql tool rather than release a half measure. We'd also be able to test phpMyAdmin over Adminer.

The following UX issues/questions have been identified in this PR:

- Improve user feedback in the UI after clicking links that start a paused site. Sometimes a site takes several seconds to boot up. 
- Optional/to debate: Depending on what we want to expose to the user, consider creating a database tab with a link to the local db file, help content.
- How to handle exceptions, possible file corruptions. Maybe Studio could trigger backups.

aminul-xs:
When might this feature be released?

ramonjd:
> When might this feature be released?

Work in [ongoing for the new SQLite driver](https://github.com/Automattic/sqlite-database-integration/commits/develop/). 

Once that's finished I think work on this feature will be continued.

This PR is a bit stale however. 

I will close it and leave it as instructive for a new PR once the new driver is ready and integrated into studio.

PR #1166, opened 2025-04-02
Auto-start previously running sites on application launch

sejas:
## Related issues

- Fixes STU-321

## Proposed Changes

- Added persistence of site running states across application restarts by saving `autoStart` site property into appData to track site were already running
- Implemented automatic startup of previously running sites when application launches
- Modified site server management to properly track and restore running states

## Testing Instructions

- Run `npm start`
- Create a new site and make sure it starts
- Close the application completely
- Reopen the application
- Verify the previously running site automatically starts up
- Create a second site and stop it
- Close and reopen the application
- Confirm only the first site auto-starts while the second remains stopped
- Stop the first site and restart the application
- Verify it remains stopped, demonstrating the state is properly tracked

https://github.com/user-attachments/assets/b94d931f-234a-4efe-98f0-be207bf7c024

## Pre-merge Checklist

### DIFF
--- a/src/storage/user-data.ts
+++ b/src/storage/user-data.ts
@@ -169,7 +169,7 @@ function fromDiskFormat( { version, sites, ...rest }: PersistedUserData ): UserD
 			.map( ( { path, name, autoStart, ...restOfSite } ) => ( {
 				name: name || nodePath.basename( path ),
 				path,
-				running: false, // Sites start as not running, they will be auto-started if needed
+				running: false,
 				autoStart: autoStart || false,
 				...restOfSite,
 			} ) ),
--- a/src/storage/user-data.ts
+++ b/src/storage/user-data.ts
@@ -130,6 +130,7 @@ function toDiskFormat( { sites, ...rest }: UserData ): PersistedUserData {
 				themeDetails,
 				customDomain,
 				enableHttps,
+				wasRunning,
 			} ) => {
 				// No object spreading allowed. TypeScript's structural typing is too permissive and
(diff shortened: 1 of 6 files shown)


src/storage/user-data.ts
```diff
@@ -164,10 +166,11 @@ function fromDiskFormat( { version, sites, ...rest }: PersistedUserData ): UserD
 	return {
 		sites: sites
 			.filter( ( site ) => fs.existsSync( site.path ) ) // Remove sites the user has deleted from disk
-			.map( ( { path, name, ...restOfSite } ) => ( {
+			.map( ( { path, name, autoStart, ...restOfSite } ) => ( {
 				name: name || nodePath.basename( path ),
 				path,
-				running: false,
+				running: false, // Sites start as not running, they will be auto-started if needed
```
katinthehatsite: Should not that comment be below with the `autoStart`? Although that is not a strong preference

katinthehatsite (approved):
The changes look good to me. I did not find any specific issues and the workflow worked as expected. I also liked that we were displaying "Starting" when the sites were starting. Nice work!!

src/storage/user-data.ts
sejas: good catch!

sejas: Removed the comment: https://github.com/Automattic/studio/pull/1166/commits/18e34da98a886688fe63ff1a27997af6f3d066bf

katinthehatsite: Looks good, I already approach so feel free to merge 🚢

PR #1178, opened 2025-04-09
What's New Section: Add tests for app version API and version comparison functionality

ivan-ottinger:
## Related issues

- Resolves STU-334-linear-issue

## Proposed Changes

- add tests for app version API and version comparison functionality

## Testing Instructions

1. Please review the proposed tests.
2. Newly-introduced tests should pass (`npm test src/stores/tests/app-version-api.test.ts`).
3. All tests should pass (`npm run test`).

## Pre-merge Checklist

### DIFF
--- a/src/stores/tests/app-version-api.test.ts
+++ b/src/stores/tests/app-version-api.test.ts
@@ -1,6 +1,6 @@
 import { configureStore } from '@reduxjs/toolkit';
 import { getIpcApi } from 'src/lib/get-ipc-api';
-import { appVersionApi, selectIsNewVersion } from '../app-version-api';
+import { appVersionApi, selectIsNewVersion } from 'src/stores/app-version-api';
 
 jest.mock( 'src/lib/get-ipc-api', () => ( {
 	getIpcApi: jest.fn(),
--- a/src/stores/tests/app-version-api.test.ts
+++ b/src/stores/tests/app-version-api.test.ts
@@ -1,0 +1,136 @@
+import { configureStore } from '@reduxjs/toolkit';
+import { getIpcApi } from 'src/lib/get-ipc-api';
+import { appVersionApi, selectIsNewVersion } from '../app-version-api';
+
+jest.mock( 'src/lib/get-ipc-api', () => ( {
+	getIpcApi: jest.fn(),
+} ) );
+
+const mockIpcApi = {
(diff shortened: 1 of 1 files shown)


gcsecsey (approved):
The new tests LGTM, although I am not very familiar with the code yet, so I would appreciate another 👀 if possible.

I tested the PR and both the newly introduced tests and all tests are passing for me.

image

CleanShot 2025-04-09 at 12 03 08@2x

ivan-ottinger:
> The new tests LGTM, although I am not very familiar with the code yet, so I would appreciate another 👀 if possible.
> 
> I tested the PR and both the newly introduced tests and all tests are passing for me.

Thank you, Gergely!

src/stores/tests/app-version-api.test.ts
```diff
@@ -0,0 +1,136 @@
...
+		it( 'should return false for a new patch version', () => {
+			const lastSeenVersion = '1.2.0';
+			const currentVersion = '1.2.1';
+
+			const result = selectIsNewVersion( createMockQueryResult( lastSeenVersion ), currentVersion );
+
+			expect( result ).toBe( false );
+		} );
+
+		it( 'should return false for prerelease versions', () => {
```
katinthehatsite, line 109: Perhaps we should rename "prerelease versions" to "beta versions" as we barely use that terminology. Although, it is not a strong opinion on my end.

katinthehatsite (approved):
The tests look good 👍 

I am wondering if we should also add a test case around `forceWhatNew` constant. What do you think?

src/stores/tests/app-version-api.test.ts
ivan-ottinger: Thanks, Kat! I have used prerelease as it was more general than beta - basically mirroring terminology used in semver already.

ivan-ottinger:
> I am wondering if we should also add a test case around forceWhatNew constant. What do you think?

Thank you Kat! I was considering this before, but since `forceWhatNew` is hardcoded, it is in a way covered by existing tests. 

However, while looking at this all again, I noticed one small bug in our display logic which I am currently fixing in another PR.

ivan-ottinger:
> However, while looking at this all again, I noticed one small bug in our display logic which I am currently fixing in another PR.

Here's the fix: https://github.com/Automattic/studio/pull/1186.

PR #1179, opened 2025-04-09
Implement CLI command for deleting preview sites

fredrikekelund:
## Related issues

- Fixes STU-245

## Proposed Changes

Add a `studio preview delete <host>` command for deleting preview sites.

## Testing Instructions

1. Run `npm run cli:build`
2. Have a preview site that you can delete for testing purposes
4. Make note of the hostname of that preview site
5. Run `node dist/cli/main.js preview delete HOST` where `HOST` is the previously noted hostname
6. Ensure that the command completes successfully
7. Run the exact same command again
8. Ensure that it prints `Preview site not found`

## Pre-merge Checklist


gcsecsey (approved):
I have tested the changes, and they work as described. Code changes also LGTM, although I am not very familiar with the code yet, so I would appreciate another 👀 if possible.

image

CleanShot 2025-04-09 at 12 16 56@2x

katinthehatsite:
In terms of functionality, everything seems to have worked as expected. As a sidenote, we should probably make it easier for users to copy the hostname from the Studio UI but it is not important for now. I will review the code next:

Screenshot 2025-04-09 at 3 32 06 PM

cli/commands/preview/delete.ts
```diff
@@ -0,0 +1,48 @@
...
+		const snapshotToDelete = snapshots.find( ( s ) => s.url === host );
+		if ( ! snapshotToDelete ) {
+			throw new LoggerError( 'Preview site not found' );
+		}
+		logger.reportSuccess( 'Validation successful' );
+
+		logger.reportStart( LoggerAction.DELETE, 'Deleting...' );
+		await deleteSnapshot( snapshotToDelete.atomicSiteId, token.accessToken );
+		await deleteSnapshotFromAppdata( snapshotToDelete.url );
+		logger.reportSuccess( 'Deletion successful' );
```
katinthehatsite: Are these strings user-facing and might need to have translations?

cli/commands/preview/delete.ts
```diff
@@ -0,0 +1,48 @@
...
+async function runCommand( host: string, outputFormat?: OutputFormat ): Promise< void > {
+	const logger = new Logger< LoggerAction >( outputFormat );
+
+	try {
+		logger.reportStart( LoggerAction.VALIDATE, 'Validating...' );
+		const token = await getAuthToken();
+		const snapshots = await getSnapshotsFromAppdata( token.id );
+		const snapshotToDelete = snapshots.find( ( s ) => s.url === host );
+		if ( ! snapshotToDelete ) {
+			throw new LoggerError( 'Preview site not found' );
```
katinthehatsite: If the user is typing the wrong hostname, what do you think about throwing the error message that suggests to use preview list command to inform them about sites available for deletion? 

Something like: `Preview site "${host}" not found. Use "studio preview list" to see available preview sites.`

katinthehatsite, line 21: If the user is typing the wrong hostname, what do you think about throwing the error message that suggests to use preview list command to inform them about sites available for deletion? 

Something like: `Preview site "${host}" not found. Use "studio preview list" to see available preview sites.`

cli/commands/preview/delete.ts
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+export const registerCommand: RegisterCommand = ( program ) => {
+	program
+		.command( 'delete <host>' )
+		.description( 'Delete a preview site' )
+		.action( async ( host: string ) => {
+			const options = program.opts();
+			const normalizedHost = host.replace( /^https?:\/\//, '' );
```
katinthehatsite: I am wondering if we might need to add some more validation here. For instance, I added the hostname in uppercase and it ended up not finding the site even though the hostname is correct:

Screenshot 2025-04-09 at 3 51 55 PM

This might be an edge case though so feel free to skip this. Just documenting the observations.

katinthehatsite (approved):
The command worked well for me, I did not observe any strange issues 👍 . I added a couple of minor suggestions (feel free to proceed with merging though).

cli/commands/preview/delete.ts
fredrikekelund: Suggesting to use the list command is a neat idea 👍

cli/commands/preview/delete.ts
fredrikekelund: It seems unlikely that this would trip users up in practice (neither the CLI nor the Studio UI prints the hostname in caps), but it still doesn't hurt to normalize the hostname to lowercase, as I see it.

cli/commands/preview/delete.ts
fredrikekelund: They do indeed 👍

I won't wrap every string in a `__()` call in this PR (I'll use a dedicated PR for that, since there's a backlog of strings that need updating), but I'll fix the ones in the command file and in the `deleteSnapshot` function.

cli/commands/preview/delete.ts
katinthehatsite: Yeah I don't think users actually do this  hence the edge case. I was mostly looking for ways to break the things in the code from different angles 😅

PR #1180, opened 2025-04-09
Implement CLI command for updating preview sites

fredrikekelund:
## Related issues

- Fixes STU-237

## Proposed Changes

Add a `studio preview update [folder] -h <host>` command for updating preview sites.

## Testing Instructions

1. Run `npm run cli:build`
2. Have a preview site that you can update for testing purposes
4. Make note of the hostname of a preview site you want to update
5. Run `node dist/cli/main.js preview update FOLDER -h HOST` where `FOLDER` is the path to the Studio site, and `HOST` is the previously noted hostname
6. Ensure that the command completes successfully
7. Run `node dist/cli/main.js preview update FOLDER` where `FOLDER` is the path to the Studio site
8. Ensure that you see a message about `-h` being a required option

## Pre-merge Checklist


cli/commands/preview/update.ts
```diff
@@ -0,0 +1,82 @@
...
+
+enum LoggerAction {
+	VALIDATE = 'validate',
+	ARCHIVE = 'archive',
+	UPLOAD = 'upload',
+	READY = 'ready',
+	APPDATA = 'appdata',
+}
+
+async function runCommand(
```
fredrikekelund, line 21: The logic in this function is very similar to that in `cli/commands/preview/create.ts`. The difference is only really that we pass a third argument to `uploadArchive`.

cli/commands/preview/update.ts
```diff
@@ -0,0 +1,86 @@
...
+	const logger = new Logger< LoggerAction >( outputFormat );
+
+	try {
+		logger.reportStart( LoggerAction.VALIDATE, __( 'Validating...' ) );
+		validateSiteFolder( siteFolder );
+		const token = await getAuthToken();
+		const snapshots = await getSnapshotsFromAppdata( token.id );
+		const snapshotToUpdate = snapshots.find( ( s ) => s.url === host );
+		if ( ! snapshotToUpdate ) {
+			throw new LoggerError( 'Preview site not found' );
```
katinthehatsite, line 39: I noticed that in this command, we do not seem to normalize the host by removing protocol compared to the delete command. As a result, when you pass the host with `https/`, you are getting `Preview site not found`. 

 Should we follow the same pattern in this command as in the `delete` command?

cli/commands/preview/update.ts
```diff
@@ -0,0 +1,86 @@
...
+		logger.reportSuccess( __( 'Preview site saved to Studio' ) );
+	} catch ( error ) {
+		if ( error instanceof LoggerError ) {
+			logger.reportError( error );
+		} else {
+			const loggerError = new LoggerError( __( 'Failed to update preview site' ), error );
+			logger.reportError( loggerError );
+		}
+	} finally {
+		setImmediate( () => cleanup( archivePath ) );
```
katinthehatsite: I see that we are not using `setImmediate` in the create command and it does not seem to be necessary there but I am wondering if it make the code more defensive and more consistent in both places. Although, this might be an overkill.

cli/commands/preview/update.ts
katinthehatsite: And another suggestion is that we could also possible suggest to use the command to list preview sites, similarly to delete command.

cli/logger.ts
fredrikekelund, line 1: I took the opportunity to make this class a bit easier to work with. We no longer have to check the type of `error` in catch clauses before passing it to `LoggerError`, but this class handles the type check internally. I've also updated the message formatting to automatically include the error message from `previousError` when available.

cli/lib/appdata.ts
```diff
@@ -92,9 +102,24 @@ export async function saveAppdata( userData: UserData ): Promise< void > {
 		const fileContent = JSON.stringify( userData, null, 2 ) + '\n';
 
 		fs.writeFileSync( appDataPath, fileContent, 'utf8' );
+	} catch ( error ) {
+		throw new LoggerError( __( 'Failed to save appdata file' ), error );
+	}
+}
+
+export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken' ] > > {
```
fredrikekelund, line 110: This function had already shrunken noticeably compared to the original implementation, so I figured moving it to `appdata.ts` was a better fit.

cli/lib/api.ts
```diff
@@ -12,7 +12,7 @@ export enum SnapshotStatus {
 
 const createSiteResponseSchema = z.object( {
 	domain_name: z.string().min( 1, 'Domain name is required' ),
-	atomic_site_id: z.number().int().positive( 'Site ID must be a positive integer' ),
+	atomic_site_id: z.coerce.number().int().positive( 'Site ID must be a positive integer' ),
```
fredrikekelund, line 15: The `/jurassic-ninja/update-site-from-zip` API endpoint returned a stringified number in my testing. The `coerce` prop makes it so zod tries to convert incoming strings to numbers.

cli/commands/preview/update.ts
fredrikekelund: Good catch 👍 I'll add a utility function and use the same approach for both commands

katinthehatsite (approved):
The changes look good. I tested with updating a preview site and got my changes with a new theme up and running. I also tried triggering different errors e.g. authentication missing, the required parameter missing, invalid folder and it worked for all cases:

Screenshot 2025-04-10 at 10 52 29 AM

I had one or two minor things but other than that, everything seemed to work on my end.

cli/lib/tests/utils.test.ts
```diff
@@ -0,0 +1,31 @@
...
+	it( 'should remove http protocol', () => {
+		expect( normalizeHostname( 'http://example.com' ) ).toBe( 'example.com' );
+	} );
+
+	it( 'should remove https protocol', () => {
+		expect( normalizeHostname( 'https://example.com' ) ).toBe( 'example.com' );
+	} );
+
+	it( 'should remove trailing slash', () => {
+		expect( normalizeHostname( 'example.com/' ) ).toBe( 'example.com' );
```
katinthehatsite, line 17: Nice, I got caught by this twice when testing 😅

cli/lib/appdata.ts
```diff
@@ -92,9 +102,24 @@ export async function saveAppdata( userData: UserData ): Promise< void > {
...
+	try {
+		const { authToken } = await readAppdata();
+
+		if ( ! authToken?.accessToken ) {
+			throw new Error( 'Authentication required' );
```
gavande1, line 115: Nit: Don't we need a translated string here?

cli/commands/preview/update.ts
fredrikekelund: I added the `setImmediate` call to mitigate some weird behavior I encountered during testing where `fs.existsSync` would return true in `cleanup`, but `fs.unlinkSync` would still throw an ENOENT error 🤷‍♂️

I later realized that this messed up the tests in more than one way, but I've added back the equivalent logic inside the `cleanup` function now again.

cli/lib/appdata.ts
fredrikekelund: We don't actually. We rely on the `LoggerError` thrown on L120 for the translated string.

PR #1133, opened 2025-03-26
Studio: Adjust logic for automatic modal opening

katinthehatsite:
## Related issues

STU-271

## Proposed Changes

This PR adjusts the logic for automatic opening of the `What's new` modal in the following way:

* When the version changes are major or minor, the modal should be opening automatically
* When the version changes are patch changes, the modal will not show automatically
* When the version is beta, the modal will also not show

## Testing Instructions

**Major version change**

* Pull the changes from this branch
* In `package.json`, change the version to a major one such as `2.0.0`
* Run `npm install`
* In the terminal where you have Studio running, type `rs` to restart the app
* Confirm that the modal appears automatically

**Minor version change**

* Pull the changes from this branch
* In `package.json`, change the version to a major one such as `2.1.0`
* Run `npm install`
* In the terminal where you have Studio running, type `rs` to restart the app
* Confirm that the modal appears automatically

**Patch version change**

* Pull the changes from this branch
* In `package.json`, change the version to a major one such as `2.1.1`
* Run `npm install`
* In the terminal where you have Studio running, type `rs` to restart the app
* Confirm that the modal does not appear automatically

**Beta versions**

* Pull the changes from this branch
* In `package.json`, change the version to a major one such as `2.1.1-beta1`
* Run `npm install`
* In the terminal where you have Studio running, type `rs` to restart the app
* Confirm that the modal does not appear automatically

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1140, opened 2025-03-27
Studio: Simplify constant approach

katinthehatsite:
## Related issues

STU-271

## Proposed Changes

This PR adds `forceNewVersion` constant to force the `What's New` modal to Studio versions other than minor or major. 

## Testing Instructions

* Pull the changes from this branch
* Start the app with `STUDIO_WHATS_NEW_SECTION=true npm start`
* Confirm that you do not see the modal
* Apply this diff

```diff --git a/src/stores/app-version-slice.ts b/src/stores/app-version-slice.ts
index aeebc032..76000bdf 100644
--- a/src/stores/app-version-slice.ts
+++ b/src/stores/app-version-slice.ts
@@ -52,7 +52,7 @@ const appVersionSlice = createSlice( {
                selectIsNewVersion: createSelector(
                        [ ( state ) => state.lastSeenVersion, ( _state, currentVersion: string ) => currentVersion ],
                        ( lastSeenVersion, currentVersion ) => {
-                               const forceNewVersion = false;
+                               const forceNewVersion = true;
 
                                if ( forceNewVersion && currentVersion !== lastSeenVersion ) {
                                        return true;
```

* Restart the app
* Confirm that you can now see the modal

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1138, opened 2025-03-27
Studio: Use feature flag to force the modal

katinthehatsite:
## Related issues

STU-271

## Proposed Changes

This PR adds forces the modal for when `whatsNewSectionEnabled` feature flag is set to true to be able to force it with app versions that are not included for automatic triggering of the modal. 

## Testing Instructions

* Pull the changes from this branch
* Apply the following diff:
```diff --git a/src/ipc-handlers.ts b/src/ipc-handlers.ts
index cc090c42..3324682f 100644
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -784,7 +784,7 @@ export async function getAppGlobals( _event: IpcMainInvokeEvent ): Promise< AppG
                appVersion: app.getVersion(),
                arm64Translation: app.runningUnderARM64Translation,
                terminalWpCliEnabled: process.env.STUDIO_TERMINAL_WP_CLI === 'true',
-               whatsNewSectionEnabled: process.env.STUDIO_WHATS_NEW_SECTION === 'true',
+               whatsNewSectionEnabled: true,
        };
 }
```
* Start the app with `npm start`
* Confirm that you see the modal when the app starts
* Confirm that the modal closes correctly

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1186, opened 2025-04-10
What's New Section: Fix modal display logic

ivan-ottinger:
## Related issues

While working on https://github.com/Automattic/studio/pull/1178 I observed a smaller issue with our modal opening logic. If `forceNewVersion` was `true` it ignored other checks (e.g. prerelease beta version check).

## Proposed Changes

- ensure `forceNewVersion` overrides the `selectIsNewVersion` output after all other conditions are evaluated

## Testing Instructions

1. Check out the branch PR and build the app with `npm start`.
2. The modal should work as expected.
3. All tests should pass (`npm test src/stores/tests/app-version-api.test.ts` and `npm run test`).
4. Change the `forceNewVersion` to `true` in `src/stores/app-version-api.ts`.
5. Run `npm test src/stores/tests/app-version-api.test.ts`  again. One test should fail - `should return false for a new patch version`.
6. Additionally, we could also consider the going through the testing instructions in https://github.com/Automattic/studio/pull/1133.

## Pre-merge Checklist

### DIFF
--- a/src/stores/app-version-api.ts
+++ b/src/stores/app-version-api.ts
@@ -52,9 +52,10 @@ export const selectIsNewVersion = createSelector(
 	( lastSeenVersion, currentVersion ) => {
 		const forceNewVersion = false;
 
-		if ( forceNewVersion && currentVersion !== lastSeenVersion ) {
-			return true;
+		if ( currentVersion === lastSeenVersion ) {
+			return false;
 		}
+
 		if ( ! currentVersion || ! lastSeenVersion ) {
 			return !! currentVersion && lastSeenVersion !== currentVersion;
 		}
@@ -79,7 +80,9 @@ export const selectIsNewVersion = createSelector(
 			}
 
 			return (
-				lastSeenParts.major !== currentParts.major || lastSeenParts.minor !== currentParts.minor
+				lastSeenParts.major !== currentParts.major ||
+				lastSeenParts.minor !== currentParts.minor ||
+				forceNewVersion
 			);
 		} catch ( error ) {
 			console.error( 'Error comparing versions:', error );


katinthehatsite (commented):
Thanks for making the changes. From what I understand, the goal is to ensure that the `forceNewVersion` does not get set to true without first checking for the difference in the `currentVersion` and `latestSeenVersion`, is that right?

src/stores/app-version-api.ts
```diff
@@ -52,9 +52,10 @@ export const selectIsNewVersion = createSelector(
 	( lastSeenVersion, currentVersion ) => {
 		const forceNewVersion = false;
```
gcsecsey, line 53: I saw that this flag was added [here](https://github.com/Automattic/studio/pull/1140) to simplify force-showing the modal. Is there any plans to toggle this more than one release in the future? If so, I think it could make sense to move it to an environment variable or feature flag, to make managing it easier.

gcsecsey (approved):
Changes LGTM, and everything works as described. I'd still appreciate another pair of 👀 as I am not yet too familiar with the code.

- Running `npm run test` passes all cases.
- Running `npm test src/stores/tests/app-version-api.test.ts` on this branch:
image

- Running `npm test src/stores/tests/app-version-api.test.ts` after setting `forceNewVersion` to `true`:
image

I added a minor comment about the flag, but it's non-blocking, so please feel free to merge without addressing it.

src/stores/app-version-api.ts
katinthehatsite: Nice suggestion! I had that idea initially[ in this PR ](https://github.com/Automattic/studio/pull/1138) but we decided it to be too complex and to keep it simpler. I also tried a more fancy approach with having a config with [the list of versions where that constant was forced. ](https://github.com/Automattic/studio/pull/1136/files)

ivan-ottinger: Thank you, Gergely! The core idea behind that flag is to let us force the modal for patch release versions (as normally we want to display the modal for major and minor versions only).

I expect the flag might be used a max. a couple of times a year - just setting it `true` for specific production release and then back to `false` again.

Because of that, I am not sure if other approach would be more suitable. 🤔

gcsecsey: Thank you both for the additional context!

> I expect the flag might be used a max. a couple of times a year - just setting it true for specific production release and then back to false again.

> Because of that, I am not sure if other approach would be more suitable. 🤔

Yeah, in this case it completely makes sense to me to use this flag as is.

ivan-ottinger:
> Thanks for making the changes. From what I understand, the goal is to ensure that the `forceNewVersion` does not get set to true without first checking for the difference in the `currentVersion` and `latestSeenVersion`, is that right?

Good question. Bfore the change, when `forceNewVersion` is set to `true`, the following tests fail as well:
- `should return false for prerelease versions`
- `should return false for invalid version formats`

The proposed change ensures `forceNewVersion` overrides the check for the patch version only. In other words, it will only make the `should return false for a new patch version` test fail. Which is what we expect.

PR #1187, opened 2025-04-10
Load translations for CLI

fredrikekelund:
## Related issues

- Fixes STU-335

## Proposed Changes

- Added `__()` wrappers around all relevant remaining strings
- Added a `loadTranslations` function called on startup that loads translations based on the preference from Studio or based on environment configuration.

## Testing Instructions

1. Open Studio and change the language preference to something other than English (ideally something you can read)
2. Open the corresponding JSON file in `src/translations`
3. Add the following message at the end: `"Studio by WordPress.com CLI":["YOUR TRANSLATION"]`
4. Run `npm run cli:watch`
5. Run `node dist/cli/main.js`
6. Ensure that the second row of the help text is translated according to what you added to the JSON file

## Pre-merge Checklist

### DIFF
--- a/cli/commands/preview/create.ts
+++ b/cli/commands/preview/create.ts
@@ -1,6 +1,6 @@
 import os from 'os';
 import path from 'path';
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 import { uploadArchive, waitForSiteReady } from 'cli/lib/api';
 import { getAuthToken } from 'cli/lib/appdata';
 import { createArchive, cleanup } from 'cli/lib/archive';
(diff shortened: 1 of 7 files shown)


gcsecsey (approved):
Changes LGTM, I didn't find any other strings that might need to be localized.

The changes also work as described, I was able to add my translation to the CLI output. Thanks for the great testing description! 🙌

image

katinthehatsite (approved):
It looks good. I tested with Ukrainian and the output looked good to me: 

Screenshot 2025-04-10 at 4 42 55 PM

I looked through the code as well and did not see improvements that would make it potentially better.

sejas (approved):
Looks good! I confirm the commands we've added are translated.
I wonder if we could and should also translate the Help default command output.

Maybe the `program.configureHelp` could help.

The underlined strings are from the default Commander Help and it seems we are not translating:
Screenshot 2025-04-14 at 17 32 59

fredrikekelund:
Excellent point about the default strings from the `commander` module, @sejas 👍

Unfortunately, `commander` doesn't support localization. I'll land this PR and look into switching `commander` for `yargs` in another PR. Judging by the [yargs docs](https://yargs.js.org/docs/#api-reference-localelocale), they support localization.

fredrikekelund:
For reference, I added STU-355 to track switching from commander to yargs

PR #1189, opened 2025-04-10
Fixes and improvements for selecting WP version

nightnei:
- Fixes STU-113

Followup to [Add mu-plugin to prevent updates on selected WordPress versions #1120](https://github.com/Automattic/studio/pull/1120)

## Proposed Changes
Previously, users could select specific version for their WP site, but it worked incorrectly - it was still autoupdated to the latest. With this PR we give opportunity to select specific version and opportunity to keep always latest (with auto-updating it). Also This PR adds a few UI improvements and some refacrtoring:
1. Created a new shared component for selecting WP version. So it will be consistently used in different places, w/o duplication of code.
2. Previously, we went with approach to store `wpVersion` inside appdata-v1.json. It's an option, but there are a few disadvantages:
	2.1. With wpVersion  were have 2 source of truth what is definitely not good. Much more reliable and stable to retrieve wp version from site instance, as we did it before.
	2.2 If user selected `latest` and we store it in appdata-v1.json, then to get specific version we always had to do soothing like `wpVersions.find(iteam => item.version === 'latest')`, what is not very nice.
	2.3 Folowing to 2.2 point above - even if we get the specific version with `item.value === 'latest'`, we can't be sure that it' exactly the version which is used on this specific wp instance, so this way we can bugs.
	2.4. Storing `isWpAutoUpdating: boolean` makes code much cleaner, as you can see in this PR, since we don't need to add extra condition and we can easily distinguish the version and auto-updating status, since it's 2 different things.

## Testing Instructions

I decided to record video, since there are a lot of testing instructions and video will simplify go through them.
Also, it will help you (the reviewer) to quickly understand how it works in all scenarios.

https://drive.google.com/file/d/1eXwUk_B8Gkd3-28cVIEj4Sg7YCV9t4Sq/view?usp=sharing


src/components/wp-version-selector/add-wp-version-to-list.ts
```diff
@@ -0,0 +1,27 @@
+import semver from 'semver';
+
+type Option = { label: string; value: string };
+
+export const addWpVersionToList = ( newOption: Option, options: Option[] ): Option[] => {
```
nightnei, line 5: 1. I rewrote this function from mutability to immutability.
2. Moved close to wpVersionsSelector, since it doesn';t make sense to keep it globally now, when we have only one pleace where we use it.

src/components/wp-version-selector/index.tsx
```diff
@@ -0,0 +1,121 @@
+import { SelectControl, Icon } from '@wordpress/components';
```
nightnei, line 1: Now we have separate component which doesn't store any context, so it's already used in 2 places and can be easily reused in the future, w/o extra conditions.
Also now we have consistency with rendering it for creating/updating website, so it works the same way, we don't need to copy-paste etc.

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -33,12 +28,11 @@ type EditSiteDetailsProps = {
 export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteDetailsProps ) {
 	const { __ } = useI18n();
 	const { updateSite, selectedSite, stopServer, startServer } = useSiteDetails();
-	const [ isChangeWpError, setIsChangeWpError ] = useState( '' );
+	const [ errorUpdatingWpVersion, setErrorUpdatingWpVersion ] = useState< string | null >( null );
```
nightnei, line 31: Just renamed, since it's not a predicate

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -49,8 +43,15 @@ export default function EditSiteDetails( { currentWpVersion, onSave }: EditSiteD
 	const [ selectedPhpVersion, setSelectedPhpVersion ] = useState< AllowedPHPVersion >(
 		( selectedSite?.phpVersion as AllowedPHPVersion ) ?? DEFAULT_PHP_VERSION
 	);
-	const siteWpVersion = selectedSite?.wpVersion === 'latest' ? 'latest' : currentWpVersion;
-	const [ selectedWpVersion, setSelectedWpVersion ] = useState( siteWpVersion );
+	const getEffectiveWpVersion = useCallback(
```
nightnei, line 46: Good example of benefit for using `isWpAutoUpdating` - we don't need any more extra conditions, just this one.

vendor/wp-now/src/wp-now.ts
```diff
@@ -603,8 +603,7 @@ async function mountInternalMuPlugins( php: PHP, options: WPNowOptions ) {
 		`
 	);
 
-	// Only add the auto-updates disabling plugin for non-latest WordPress versions
-	if ( options?.wordPressVersion && options.wordPressVersion !== 'latest' ) {
+	if ( ! options.isWpAutoUpdating ) {
```
nightnei, line 606: Also good example of benefit from `isWpAutoUpdating` - the condition is much more simple.

src/site-server.ts
```diff
@@ -70,11 +70,19 @@ function getAbsoluteUrl( details: SiteDetails ): string {
 	return `http://localhost:${ details.port }`;
 }
 
+// We use SiteDetails for storing it in appdata-v1.json, so this meta was introduced for extra data which is not stored locally
+type SiteServerMeta = {
+	wpVersion?: string;
```
nightnei, line 75: @sejas This is only one point where I am not sure that it's the best approach. I am not very familiar with wpnow etc, so could you please take a look on my approach to say - is it good, or maybe it's better to go another way.

src/components/site-form.tsx
```diff
@@ -418,73 +404,14 @@ export const SiteForm = ( {
...
-														) : (
-															<option value={ DEFAULT_WORDPRESS_VERSION }>
-																{ DEFAULT_WORDPRESS_VERSION }
-															</option>
-														) }
-													</SelectControl>
-												</Tooltip>
-											</div>
+
+											<WPVersionSelector
```
sejas, line 408: Nice! A single component 💪

src/hooks/use-add-site.ts
```diff
@@ -113,13 +113,6 @@ export function useAddSite() {
 						};
 					}
 
-					if ( newSite.wpVersion !== wpVersion ) {
```
sejas, line 116: Everything seems to be working correctly, creating new sites and editing sites.
So removing the test `should update site with WordPress version after creation` makes sense.

sejas (commented):
I tested multiple cases of existing and new sites, and everything worked correctly. I'll give it a second pass tomorrow. 

First, I like that we have a common component. We could rename the label `Keep Latest Stable` to `Latest` or `Autoupdate Latest`. Sites with older versions are marked as "Keep Latest Stable," which I believe is correct, as they will eventually be auto-updated.

For now, the edge case I'm uncertain about are Beta/Nightly versions. Existing sites with those versions are marked as `Keep Latest Stable` when they are not stable at all.
We add the version to the options, which is great. I think we should also mark that version as selected.
Additionally, I think we should continue auto-updating those beta versions, but it's not a strong opinion.

Screenshot 2025-04-14 at 18 29 58 2

wojtekn:
Can we display the actual version, e.g., "6.7.2  (latest)" instead of "Keep Latest Stable"? 

> For now, the edge case I'm uncertain about are Beta/Nightly versions. Existing sites with those versions are marked as Keep Latest Stable when they are not stable at all.
> We add the version to the options, which is great. I think we should also mark that version as selected.
Additionally, I think we should continue auto-updating those beta versions, but it's not a strong opinion.

I think those shouldn't be auto-updated, and the first one should be the only one that is getting auto-updated.

Regarding the edit form, it seems there is a bug there as the "Keep Latest Stable" is displayed for each site. I think the currently used version should be preselected.

sejas:
All existing sites are auto-updated, so eventually if those sites are used they will get updated to the latest version. That's the logic behind the picker showing "Keep Latest Stable" and not the specific version.

These existing sites are an edge case and any logic we add to handle them should be something "temporal". @nightnei, Maybe the **first time** Studio starts we could check the WP version in all existing sites and add  `isWpAutoUpdating: false` in all of them that doesn't have the 6.7.2 version.

> Regarding the edit form, it seems there is a bug there as the "Keep Latest Stable" is displayed for each site. I think the currently used version should be preselected.

Cool!, let's keep them as "pinned" versions.

nightnei:
> Can we display the actual version, e.g., "6.7.2 (latest)" instead of "Keep Latest Stable"?

@wojtekn regarding this point - I found it confusing to have both "6.7.2" and "6.7.2 latest", more details I descibed in this thread -p1744195024724019-slack-C04GESRBWKW.

Also, I was talking with ChatGPT about it and even w/o providing my thoughts he said that it's not clear for users to understand the difference bethween these options, since in both cases 6.7.2 is indeed latest. So he recommended to flag that the first option keeps site wiuth latest WP version and in teh second it uses static version.

Take a look please on the thread.
WDYT?

nightnei:
>All existing sites are auto-updated, so eventually if those sites are used they will get updated to the latest version. That's the logic behind the picker showing "Keep Latest Stable" and not the specific version.

> These existing sites are an edge case and any logic we add to handle them should be something "temporal".

@wojtekn @sejas I have teh same thoughts. Also, additionally, as @sejas mentioned before - we would have inconsistency with rendering this warning <br />Screenshot 2025-04-15 at 09 53 51

nightnei:
> @nightnei, Maybe the first time Studio starts we could check the WP version in all existing sites and add isWpAutoUpdating: false in all of them that doesn't have the 6.7.2 version.

@sejas hm, it seems it's great idea to achive consistency 👍 I will think about it

src/storage/user-data.ts
```diff
@@ -59,6 +59,79 @@ function legacyPopulateSnapshotUserIds( data: UserData ): void {
...
+
+	for ( let i = 0; i < Math.max( v1.length, v2.length ); i++ ) {
+		const num1 = v1[ i ] || 0;
+		const num2 = v2[ i ] || 0;
+		if ( num1 > num2 ) return true;
+		if ( num1 < num2 ) return false;
+	}
+
+	return true; // versions are equal
+}
```
sejas: What do you think about removing this function in favor of [`semver.gt`](https://www.npmjs.com/package/semver) ?

src/storage/user-data.ts
```diff
@@ -59,6 +59,79 @@ function legacyPopulateSnapshotUserIds( data: UserData ): void {
...
+			versionFileContent = fs.readFileSync(
+				nodePath.join( site.path, 'wp-includes', 'version.php' ),
+				'utf8'
+			);
+		} catch ( err ) {
+			return { ...site, isWpAutoUpdating: false };
+		}
+
+		const matches = versionFileContent.match( /\$wp_version\s*=\s*'([0-9a-zA-Z.-]+)'/ );
+		const version = matches?.[ 1 ];
```
sejas: Why not calling to `getWordPressVersionFromInstallation` directly?

src/storage/user-data.ts
```diff
@@ -59,6 +59,79 @@ function legacyPopulateSnapshotUserIds( data: UserData ): void {
...
+
+		if ( ! version ) {
+			return { ...site, isWpAutoUpdating: false };
+		}
+
+		let isVersionGreaterThan672;
```
sejas: Let's rename this variable in case wp 6.8.0 is out before our next Studio release.
For example isVersionGreaterThanThreshold ? or isVersionGreaterThanLatest?

src/storage/user-data.ts
```diff
@@ -67,13 +140,16 @@ export async function loadUserData(): Promise< UserData > {
...
-			const data = fromDiskFormat( parsed );
+			let data = fromDiskFormat( parsed );
 
 			// Temporarily populate old snapshots with userId of authenticated user.
 			// See PR #937 for more context.
 			legacyPopulateSnapshotUserIds( data );
 			sortSites( data.sites );
 			populatePhpVersion( data.sites );
+
+			data = await migrateWpAutoUpdating( data );
```
sejas: I think this function should be executed only once and not every time Studio starts.

sejas (commented):
I didn't test it yet. I left a few comments in the new code. Let me know what you think.

src/storage/user-data.ts
nightnei: I was thinking about it and why I went this approach:
1. to check - should we run this function, we should do at least 'deta.sites.filter()', so it means that we should run at least 1 loop always.
2. With the current approach, if all sites are immigrated, we run only one loop, so totally the same https://github.com/Automattic/studio/pull/1189/files#diff-2a78c3d26284d759ad9ec39370822d96d2dcbf38c383487f49c15f67192682daR85

So it's trade of - extra `filter` function outside or `map` to create new array inside. performance will be totally the same.

So it's only about readability and thinking about it... probably you are right, condition outside would keep better readability 👍 I will add it

nightnei: Done

src/storage/user-data.ts
nightnei: Didn't see this function, thanks, I will update it

wojtekn (approved):
Feature-wise works well.

src/components/wp-version-selector/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+	return (
+		<label className="flex flex-1 flex-col gap-1.5 leading-4">
+			<span className="font-semibold flex items-center gap-2">
+				{ __( 'WordPress version' ) }
+				{ selectedValue !== 'latest' && (
+					<Tooltip
+						text={ __( 'WordPress Core automatic updates will be disabled for this site.' ) }
+						placement="top"
+					>
+						<Icon icon={ warning } className="text-[#ae5c00]" size={ 16 } />
```
sejas: I don't think this color is required. I don't see any change in the UI after removing it.

```suggestion
						<Icon icon={ warning } size={ 16 } />
```

Screenshot 2025-04-21 at 11 52 02

sejas (approved):
Thanks for the changes.
I confirm new created sites work as expected and latest will be autoupdated, while betas and stables have `"isWpAutoUpdating": false,`

Existing sites are identified with the correct version too.

sejas:
I noticed the performance metrics is failing in this PR. I think we need to fix that before merging it.

PR #1196, opened 2025-04-11
Studio: Dynamically determine sidebar breakpoint based on tab width

epeicher:
## Related issues

- Fixes STU-203

https://github.com/user-attachments/assets/af710437-eb28-44ce-a9ef-f1eb3abdf43a

## Proposed Changes
- Introduced dynamic breakpoint calculation for sidebar visibility based on the width of the passed element identified by a selector
- Added state management for dynamic breakpoint using `dynamicBreakPoint` state
- Enhanced resize event handler to consider tab overflow conditions

> [!NOTE]
>  - The commit [cefee1c](https://github.com/Automattic/studio/pull/1196/commits/cefee1c06d0ce5879239950816acc6aa353212ea) used a simpler approach hardcoding the `div[role="tablist"]` selector, but I decided to elaborate more the changes and accept a selector as a parameter.
> - The selector used is a CSS class because the `TabPanel` does not accept other props like `id` and the element that was causing the overflow was the inner `TabList` component.
> - The breakpoint is not recalculated when changing the language, only when resizing the window. If required, this can be further investigated.
> - Added a docstring to the hook, but happy to delete it if it feels unnecessary.

## Testing Instructions
- Run `npm start`
- Change the language to one that has longer words, like `Svenska`
- Resize the window width to observe the sidebar behavior:
  - When tabs overflow, the sidebar should automatically hide at a wider breakpoint
  - The breakpoint should be calculated based on the tab width plus sidebar width plus the 10px right padding
  - Verify the sidebar toggle works correctly at different window sizes
- Change the language to other languages, refresh, and confirm that there are not unexpected behavior with the sidebar collapsing.

## Pre-merge Checklist


gavande1 (approved):
Thank you for your effort on this, @epeicher. Excellent work! I have tested this, and it works as described. I noticed the sidebar is not displayed when the tab size is larger. Looks good to me! 👍

src/hooks/use-sidebar-visibility.ts
```diff
@@ -1,22 +1,45 @@
...
-			setIsLowerThanBreakpoint( window.innerWidth < SIDEBAR_BREAKPOINT );
+			const el = document.querySelector( elementSelector );
+			if (
+				! isLowerThanBreakpoint && // Only recalculate if the sidebar is visible
+				el?.clientWidth &&
+				el?.scrollWidth &&
+				el?.clientWidth < el?.scrollWidth
+			) {
+				// The new breakpoint is the width of the element to measure plus the sidebar plus the right padding
+				setDynamicBreakPoint( el.clientWidth + SIDEBAR_WIDTH + 10 );
```
katinthehatsite: Should we move `10` from here to a constant? Not a strong preference on my end though.

katinthehatsite (approved):
It looks good, I did not find anything strange with the sidebar. Nice work! 👍 

I noticed one small thing and that is when the app get restarted, it does not preserve the resized view but goes back to a default with the sidebar:

https://github.com/user-attachments/assets/cc962a12-964d-43bb-a5a0-48f05c5f03c1

But I am not sure it is something that we need to handle in this PR or maybe something we might not need to handle at all. Just documenting it here as an observation in case someone else comes across this.

src/hooks/use-sidebar-visibility.ts
epeicher: Good catch, I thought about it and left it there for simplicity, but I think it can be moved, doing it now

epeicher: Added the constant here [d9cb7de](https://github.com/Automattic/studio/pull/1196/commits/d9cb7dee4baafa8bc17155c5a3d57b7631f5958f)

epeicher:
> I noticed one small thing and that is when the app get restarted, it does not preserve the resized view but goes back to a default with the sidebar:
> 
> But I am not sure it is something that we need to handle in this PR or maybe something we might not need to handle at all. Just documenting it here as an observation in case someone else comes across this.

That's right, that is related to one of my notes 
> The breakpoint is not recalculated when changing the language...

I have been looking at adding code to recalculate the breakpoint when starting the window. For example, adding a call to `handleResize` [before line 43 here](https://github.com/Automattic/studio/blob/d9cb7dee4baafa8bc17155c5a3d57b7631f5958f/src/hooks/use-sidebar-visibility.ts#L43) after defining the `handleResize` function but at that point, the element `app-measure-tabs-width` is not created yet so further code should be added. 

So I would say that can be tackled as a follow-up PR if we find that scenario happening a lot.

src/hooks/use-sidebar-visibility.ts
katinthehatsite: Nice 👍

PR #1200, opened 2025-04-14
fix commit count check

bcotrim:
## Related issues

- After a release, dev builds are throwing an error

## Proposed Changes

- Change to check to include 0 commit count scenarios:
 
image

## Testing Instructions

- N/A - Code review should be enough

## Pre-merge Checklist

### DIFF
--- a/scripts/generate-releases-manifest.mjs
+++ b/scripts/generate-releases-manifest.mjs
@@ -45,7 +45,6 @@ import https from 'https';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import packageJson from '../package.json' assert { type: 'json' };
-import { getLatestTag, getCommitCount } from './lib/git-utils.mjs';
 
 const cdnURL = 'https://cdn.a8c-ci.services/studio';
 const baseName = 'studio';
@@ -53,15 +52,12 @@ const baseName = 'studio';
 const currentCommit = child_process.execSync( 'git rev-parse --short HEAD' ).toString().trim();
 const { version } = packageJson;
 const isDevBuild = process.env.IS_DEV_BUILD;
-const latestTag = getLatestTag();
-const commitCount = getCommitCount( latestTag );
 
 const __dirname = path.dirname( fileURLToPath( import.meta.url ) );
 
 console.log( `Version: ${ version }` );
 console.log( `Is dev build: ${ isDevBuild }` );
 console.log( `Current commit: ${ currentCommit }` );
-console.log( `Commit count since ${ latestTag || 'start' }: ${ commitCount }` );
 
 console.log( 'Downloading current manifest ...' );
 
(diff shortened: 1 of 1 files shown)


scripts/generate-releases-manifest.mjs
```diff
@@ -127,7 +127,7 @@ const releasesData = JSON.parse( await fs.readFile( releasesPath, 'utf8' ) );
 if ( isDevBuild ) {
 	console.log( 'Overriding latest dev release ...' );
 
-	if ( ! commitCount ) {
+	if ( commitCount !== 0 && ! commitCount ) {
```
nightnei: Hm, is this check correct?
What I see - getCommitCount returns number, new check allows only `>0`, but second check disallows any number more than `0`.

Is it typo and you wanted to use `=== 0`? If so, then it's weird that `! commitCount` doesn't include it...

Hm, could you please explaine and probably we need to add some comment to teh code.

nightnei (changes requested):

scripts/generate-releases-manifest.mjs
bcotrim: I think so, it follows the same logic as the one we have on the https://github.com/Automattic/studio/blob/trunk/scripts/prepare-dev-build-version.mjs#L14

I did notice however, that we no longer need the `commitCount` here, so I cleanup the file I bit.
Let me know if it's better now.

nightnei (approved):
LGTM 👍
But regarding [this condition](https://github.com/Automattic/studio/blob/trunk/scripts/prepare-dev-build-version.mjs#L14) - I don't underatand how it can work...

bcotrim:
> LGTM 👍 But regarding [this condition](https://github.com/Automattic/studio/blob/trunk/scripts/prepare-dev-build-version.mjs#L14) - I don't underatand how it can work...

`! commitCount` returns `true` for all `falsy` values, which include `0`. However for our scenario `0` is a valid `commitCount`, that's why we need to check that.
We could use something like [isFinite](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite) to make it clearer, do you think it's better to change that?

nightnei:
>! commitCount returns true for all falsy values, which include 0. However for our scenario 0 is a valid commitCount, that's why we need to check that.
We could use something like [isFinite](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite) to make it clearer, do you think it's better to change that?

I see now, thanks for the explanation.
In this case, let's go with simpel and clear condition as `typeof commitCount !== 'number'`. Is there some reason why this check doesn't work for the case?
It covers teh same, but it shows explicitly that we throw error if it's not a number. Unless I am missing some scenario.

bcotrim:
> > ! commitCount returns true for all falsy values, which include 0. However for our scenario 0 is a valid commitCount, that's why we need to check that.
> > We could use something like [isFinite](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite) to make it clearer, do you think it's better to change that?
> 
> I see now, thanks for the explanation. In this case, let's go with simpel and clear condition as `typeof commitCount !== 'number'`. Is there some reason why this check doesn't work for the case? It covers teh same, but it shows explicitly that we throw error if it's not a number. Unless I am missing some scenario.

There are some cases where `typeof commitCount !== 'number'` won't work for this check, like `NaN`
image

nightnei:
>There are some cases where typeof commitCount !== 'number' won't work for this check, like NaN

In this case I would go with explicit check as `typeof commitCount !== 'number' || Number.isNaN(value)`, it's still more explicit then `commitCount !== 0 && ! commitCount`.

PR #1201, opened 2025-04-14
Refactor Studio to depend on CLI for preview site creation

fredrikekelund:
## Related issues

- Fixes STU-240

## Proposed Changes

- Add a Redux slice to hold state related to preview site CRUD operations.
- Refactor the preview site creation functionality to depend on the CLI.
- Add a `getSnapshotUsage` query to `src/stores/wpcom-api.ts` and update the UI to use this hook in the relevant places. 
- Remove the `useArchiveSite` hook. 

> [!NOTE]
> Snapshot state is now split between the new Redux reducer and the old `SnapshotProvider`. This means updating and deleting snapshots does not update the UI as expected. We will address this in separate PRs (potentially branching off this PR instead of merging this to `trunk` first). 

## Testing Instructions

1. Ensure that creating a preview site works as expected. 
2. Ensure that the Preview sites counter in the settings modal works as expected. 

## Pre-merge Checklist


cli/commands/preview/create.ts
```diff
@@ -54,6 +47,7 @@ async function runCommand( siteFolder: string, outputFormat?: OutputFormat ): Pr
 		}
 	} finally {
 		cleanup( archivePath );
+		process.exit( 0 );
```
fredrikekelund: I'm not sure why this is needed, but Electron wouldn't emit an `exit` event for the child process without it.

fredrikekelund: I was able to resolve this by using node's `child_process` module instead of Electron's `utilityProcess` module.

cli/commands/preview/logger-actions.ts
```diff
@@ -0,0 +1,28 @@
+// Store the actions in a separate file to avoid Webpack issues when importing them in the Studio
```
bcotrim, line 1: At some point I believe it would be better for us to start building a `common` folder for types and functions that are used in both CLI and Studio. I don't think it will make any difference behavior wise, but will make it more intentional and easier to read.

bcotrim (commented):
I don't really like the progress bar behavior for this scenario.
What happens on all my tests is that the bar progresses quickly until the last step, so it just gets stuck on this state for a while:
image

In my opinion, we could either use an indeterminate progress indicator (like a continuous progress bar) or display the same information we show on the CLI (Creating archive, Uploading archive, etc...) which seems far more valuable for our users compared to a stuck progress bar.

Code-wise I like this approach, looks clean and it's easy to read.
One minor comment would be to clear up the naming a bit. At least for me, the distinction between previews and snapshots seem confusing. Is there a benefit for keeping the 2, could we pick one?

Another thing on my mind is how we are currently cleaning up operations. Operations that fail, timeout, or are hanging for other reasons could potentially go on existing forever in our state, correct? I don't think this alone will be an issue, but it's something we should be mindful of, if this is going to be our approach for operations in Studio moving forward.

Is it possible to use a feature flag for this during the development?

fredrikekelund:
> Is it possible to use a feature flag for this during the development?

This PR touches so many parts of the code, so I was thinking it might be easier to branch off of this PR for the related tasks in STU-235

src/ipc-handlers.ts
```diff
@@ -17,12 +17,15 @@ import nodePath from 'path';
 import * as Sentry from '@sentry/electron/main';
 import { __, LocaleData, defaultI18n } from '@wordpress/i18n';
 import archiver from 'archiver';
+import { z } from 'zod';
+import { CreateLoggerAction } from 'cli/commands/preview/logger-actions';
```
youknowriad, line 21: If you're importing code from the "cli" app, why not import the command and run it as a function instead of a subprocess. I don't have a fully formed opinion at the moment but I think right now we're doing a bit of both which is kind of weird.

fredrikekelund: I suggested this approach of keeping the business logic squarely in the CLI when planning the project. It definitely adds complexity, but the upside is that it lets us avoid double bundling of PHP-WASM dependencies (once we get to moving that functionality from Studio to the CLI), which would increase the bundle size by ~700MB in the short term and ~400MB in the long term.

cli/commands/preview/logger-actions.ts
fredrikekelund: Good idea 👍

src/ipc-handlers.ts
fredrikekelund: As for code sharing, the idea is to only share modules without side effects. @bcotrim [suggested](https://github.com/Automattic/studio/pull/1201#discussion_r2044500053) moving those to a `common` folder for added clarity, which I think is a good idea.

fredrikekelund:
> progress bar

That's fair. I guess the current behavior in this PR would easily be perceived as a UX regression. I'll look into improving this. 

> clear up the naming

That's fair. We could probably stick with "snapshots", since that's what we use in the code already. 

> how we are currently cleaning up operations. Operations that fail, timeout, or are hanging for other reasons could potentially go on existing forever in our state, correct?

They could indeed. That's not unique for this architecture, though, even if I realize that I still need to add some error handling to this PR.

cli/lib/appdata.ts
```diff
@@ -11,15 +11,15 @@ export const snapshotSchema = z.object( {
...
+	name: z.string(),
 	userId: z.number().optional(),
 } );
 
 const siteSchema = z
 	.object( {
 		id: z.string(),
 		path: z.string(),
-		name: z.string().optional(),
+		name: z.string(),
```
fredrikekelund: I don't believe `name` props are actually optional for sites or snapshots. I made this change to accommodate the `reportKeyValuePair` logging of the snapshot name.

cli/lib/snapshots.ts
```diff
@@ -28,14 +21,11 @@ export async function addPreviewSiteToAppdata(
...
-): Promise< void > {
+): Promise< Snapshot > {
 	try {
 		const userData = await readAppdata();
-		const site = userData.sites?.find( ( s ) => s.path === siteFolder );
-		if ( ! site ) {
-			return;
-		}
-		const snapshot: z.infer< typeof snapshotSchema > = {
+		const site = await getSiteFromFolder( siteFolder );
```
fredrikekelund: This change means that this function now throws if the site isn't found in the appdata file.

src/ipc-handlers.ts
youknowriad: Sounds good to me ;)

fredrikekelund:
I've addressed your remarks, @bcotrim. 

- Instead of using static values for the progress bar, we now use a combination of changing status messages (inferred from the CLI output) and an animated progress bar (thanks to `ProgressBarWithAutoIncrement`). 
- We now consistently use the term "snapshot".
- I've improved the error handling to match the existing experience (errors display an error modal).
- I've also updated the success handler to display a system notification (matching the existing behavior). This turned out to require a rather large refactor, because we needed the snapshot name and URL as generated by the CLI. To support this, I added a `reportKeyValuePair` method to the logger, that only logs output when the output mode is set to `json`.

bcotrim (approved):
LGTM 👍 
Code and UI are looking great!

nightnei (changes requested):
I was testing and:
1. Counter looks broken for me
2. Deleting also broken (maybe it's expected for this PR and will be solved in taht separate PR for handling deletion of previews, but wanted to flag in case if it's not expected)

@bcotrim coukd you please test it one more time, to understand is it my local issue or it's indeed exist.

Recording of the issues: https://drive.google.com/file/d/1xd71MGjY8bU0ZFdEcbzrRKjNqounppAp/view?usp=sharing

bcotrim:
@nightnei I see that same issue but I assumed it was expected based on 

> [!NOTE]
> Snapshot state is now split between the new Redux reducer and the old `SnapshotProvider`. This means updating and deleting snapshots does not update the UI as expected. We will address this in separate PRs (potentially branching off this PR instead of merging this to `trunk` first). 

If you reload the app after creating a preview site, the count is updated as expected.

nightnei:
> If you reload the app after creating a preview site, the count is updated as expected.

It wasn't menthioned in testing instructions, I see only "... works es expected". But it doesn't work as expected.

Also, here (p1744904341127939-slack-C04GESRBWKW) he menthioned "I opted for this multi-PR approach to not break the preview site functionality on trunk while still making the PR size relatively manageable.", so looks liek it should work well even after merging just this one PR, w/o other PRs.

Let's wait for Fredrik's response to clarify things.

UPDATE:
I tested with the help of "Refactor preview site deleting to depend on CLI #1230" (it's chain of PRs) - we indeed have bugs:
1. Delete previeew site - counter isn't updated
2. Delete all preview sites - doesn't have any effect to UI, but under teh hood looks liek everything works well, so necessary to fix UI handling.

src/components/progress-bar.tsx
```diff
@@ -27,26 +25,39 @@ const ProgressBar = ( {
...
-} ) {
-	const [ value, setValue ] = useState( initialValue );
+}: ProgressBarWithAutoIncrementProps ) {
+	const [ animatedValue, setAnimatedValue ] = useState( value );
 
 	useEffect( () => {
+		if ( value > animatedValue ) {
+			setAnimatedValue( value );
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
nightnei, line 46: Let's add some description, otherwise anotehr A12N can make mistake with adding it `animatedValue`. And it will save time for A12S in the future with debugging to understand teh reason and logic.

src/components/progress-bar.tsx
```diff
@@ -27,26 +25,39 @@ const ProgressBar = ( {
...
+	}, [ value ] );
+
+	useEffect( () => {
+		const maximumAutoIncrementValue = maxValue * 0.95;
+
 		const interval = setInterval( () => {
```
nightnei: It seems it should be setTimeout, since we run this effect every time when we update `animatedValue`.
Theoreticaly, it works even now, since we clear interval every time when we rerun the effect, but a safer and correct way would be to use setTimeout.
Right or am I missing something?

src/stores/snapshot-slice.ts
```diff
@@ -0,0 +1,205 @@
...
+				};
+			} );
+	},
+	selectors: {
+		selectSnapshots: ( state ) => state.snapshots,
+		selectSnapshotsCount: ( state ) => state.snapshots.length,
+		selectActiveOperationForSite: ( state, siteId: string ) =>
+			Object.values( state.operations )
+				.filter( ( operation ) => operation.status === 'pending' )
+				.find( ( operation ) => operation.siteId === siteId ),
```
nightnei, line 81:
```suggestion
				.filter( ( operation ) => operation.status === 'pending' && operation.siteId === siteId )
```

src/stores/snapshot-slice.ts
```diff
@@ -0,0 +1,205 @@
...
+					progress: 0,
+					siteId: action.payload.siteId,
+					status: 'pending',
+					type: 'create',
+				};
+			} );
+	},
+	selectors: {
+		selectSnapshots: ( state ) => state.snapshots,
+		selectSnapshotsCount: ( state ) => state.snapshots.length,
```
nightnei, line 77: We use it only in 1 place.
Also it looks a bit redundant to create extra selector for length if we already have selector for teh list.
Also, it even simlifies using it and readign code:

```
BEFORE:
	const allSnapshots = useRootSelector( snapshotSelectors.selectSnapshots );
	const snapshotsCount = useRootSelector( snapshotSelectors.selectSnapshotsCount );
AFTER:
	const allSnapshots = useRootSelector( snapshotSelectors.selectSnapshots );
	const snapshotsCount = allSnapshots.length; // But actually, I think we can even avoid creating this extra variable and use "allSnapshots.length" directly in code
```

```suggestion
```

nightnei:
> At some point I believe it would be better for us to start building a common folder for types and functions that are used in both CLI and Studio.

It's great idea 👍
But also, when we proceed with it - necessary to take into account that we have 3 things: backend, cli and frontend. We already have mix of frontend and backend and one day I belive we should split it. So if we create kinda `common` folder, then necessary to think about more specific name and / or location, to avoid bounding all 3 things. Otherwise, it will be very messy.

bcotrim:
> > At some point I believe it would be better for us to start building a common folder for types and functions that are used in both CLI and Studio.
> 
> It's great idea 👍 But also, when we proceed with it - necessary to take into account that we have 3 things: backend, cli and frontend. We already have mix of frontend and backend and one day I belive we should split it. So if we create kinda `common` folder, then necessary to think about more specific name and / or location, to avoid bounding all 3 things. Otherwise, it will be very messy.

We already merged a change with the initial version of that: https://github.com/Automattic/studio/pull/1222
Let me know if you have any feedback, and I can address it in a follow-up

src/stores/snapshot-slice.ts
nightnei: I see that you already did it in the follow up PR https://github.com/Automattic/studio/pull/1223/files#diff-e7f3116a80918052c6f2b0b4cc2e22b2f8dc46e44d97249c6d49522bc2bf3b2cR127-R131

src/modules/cli/lib/execute-command.ts
```diff
@@ -0,0 +1,65 @@
...
+		}
+	} );
+
+	child.on( 'error', ( error ) => {
+		console.error( 'Child process error:', error );
+		eventEmitter.emit( 'error', error );
+	} );
+
+	child.on( 'exit', ( code: number | null ) => {
+		console.log( 'Preview site creation completed with code', code );
```
bcotrim, line 51: While reviewing https://github.com/Automattic/studio/pull/1223 I noticed that we will have a specific execute command for previews. It will likely make sense to remove this

src/components/progress-bar.tsx
fredrikekelund: Good call 👍

src/stores/snapshot-slice.ts
fredrikekelund: I agree. I actually removed this in https://github.com/Automattic/studio/pull/1230.

fredrikekelund:
> Delete previeew site - counter isn't updated

Good catch. We do actually invalidate the query correctly, but it looks like the deletion hasn't taken effect yet on the back-end when we do that, because the API still returns the old site count. Maybe we should delay the query invalidation to solve this…

> Delete all preview sites - doesn't have any effect to UI, but under teh hood looks liek everything works well, so necessary to fix UI handling.

Good catch. Looking into this 👍

src/modules/cli/lib/execute-command.ts
fredrikekelund: Good point 👍 I've removed it in https://github.com/Automattic/studio/pull/1230

fredrikekelund:
I've addressed your comments, @nightnei. You should probably re-review https://github.com/Automattic/studio/pull/1230, but this PR needs an approval before merging.

nightnei (approved):
LGTM 👍

PR #1202, opened 2025-04-14
Submit bump data allowing us to calculate WAUs for the CLI

bcotrim:
## Related issues
- Fixes STU-262

## Proposed Changes
- Add anonymous usage statistics tracking for Studio CLI commands
- Implement aggregated unique stat bumping functionality with daily/weekly/monthly intervals
- Store last bump timestamps in app data to prevent duplicate stat recording
- Add preAction hook to track CLI usage on command execution
- Update app data schema to include lastBumpStats storage

## Testing Instructions
- Run `npm run cli:build`
- Execute any Studio CLI command like `node dist/cli/main.js preview list`
- Verify in the code that stats are being tracked (in production) or logged (in development)
- Try running the same command multiple times within the same week to ensure stats are only bumped once per interval

## Pre-merge Checklist


bcotrim:
**Note**
I introduced some code duplication regarding bump stats since we have different ways to get and save app data between the CLI and the Studio app.
I think it's fine, but let me know if you have a better approach.

gavande1 (approved):
I have tested this on my local. The stat was printed on the console. The changes LGTM 👍

cli/index.ts
```diff
@@ -11,6 +12,9 @@ program
 	.name( 'studio' )
 	.description( 'Studio by WordPress.com CLI' )
 	.version( version )
+	.hook( 'preAction', () =>
+		bumpAggregatedUniqueStat( StatsGroup.STUDIO_CLI_USAGE_UNIQUE, StatsMetric.SUCCESS, 'weekly' )
+	)
```
fredrikekelund:
```suggestion
	.hook( 'preAction', () => {
		bumpAggregatedUniqueStat( StatsGroup.STUDIO_CLI_USAGE_UNIQUE, StatsMetric.SUCCESS, 'weekly' )
	} )
```
Nit, but since we're not intentionally returning anything, I say let's wrap the function body in curly braces.

cli/lib/stats.ts
```diff
@@ -0,0 +1,92 @@
+import { isSameDay, isSameMonth, isSameWeek } from 'date-fns';
+import fetch from 'node-fetch';
+import { readAppdata, saveAppdata } from './appdata';
+
+export enum StatsGroup {
+	STUDIO_CLI_USAGE_UNIQUE = 'studio-cli-usage-unique',
+}
```
fredrikekelund: I'm thinking we could add this to `StatsGroup` in `src/lib/bump-stats/types.ts`. It would clarify the structure of the `lastBumpStats` object.

cli/lib/stats.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export enum StatsMetric {
+	SUCCESS = 'success',
+	FAILURE = 'failure',
+}
```
fredrikekelund: Same thing here about reusing `StatsMetric` from `src/lib/bump-stats/types.ts` to clarify the expected structure of the `lastBumpStats` object.

cli/lib/appdata.ts
```diff
@@ -33,6 +33,7 @@ const userDataSchema = z
 			} )
 			.passthrough()
 			.optional(),
+		lastBumpStats: z.record( z.string(), z.record( z.string(), z.number() ) ).optional(),
```
```suggestion
		lastBumpStats: z
			.record( z.nativeEnum( StatsGroup ), z.record( z.nativeEnum( StatsMetric ), z.number() ) )
			.optional(),
```
If we use a single `StatsGroup` and `StatsMetric` definition, we could refine the schema.

cli/lib/stats.ts
```diff
@@ -0,0 +1,92 @@
...
+
+// Returns UTC timestamp of the last time the stat was bumped, or null if it has never been bumped.
+async function getLastBump( group: string, stat: string ): Promise< number | null > {
+	const { lastBumpStats } = await readAppdata();
+	return lastBumpStats?.[ group ]?.[ stat ] ?? null;
+}
+
+// Store this moment as the last time we bumped the state, in UTC time.
+async function updateLastBump( group: string, stat: string ) {
```
fredrikekelund:
```suggestion
async function getLastBump( group: StatsGroup, stat: StatsMetric ): Promise< number | null > {
	const { lastBumpStats } = await readAppdata();
	return lastBumpStats?.[ group ]?.[ stat ] ?? null;
}

// Store this moment as the last time we bumped the state, in UTC time.
async function updateLastBump( group: StatsGroup, stat: StatsMetric ) {
```
If we refine the schema, we should also refine the argument types here.

cli/lib/stats.ts
```diff
@@ -0,0 +1,92 @@
...
+		.then( ( didBump ) => {
+			if ( didBump ) {
+				updateLastBump( group, stat );
+			}
+		} );
+}
+
+// Returns true if we attempted to bump the stat
+export function bumpStat( group: StatsGroup, stat: StatsMetric, bumpInDev = false ) {
+	if ( process.env.E2E || ( process.env.NODE_ENV === 'development' && ! bumpInDev ) ) {
```
fredrikekelund:
```suggestion
	if ( process.env.NODE_ENV === 'development' && ! bumpInDev ) {
```
I guess we can probably remove the E2E env var check here?

cli/lib/stats.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export type AggregateInterval = 'daily' | 'weekly' | 'monthly';
+
+// Bumps a stat if it hasn't been bumped within the current aggregate interval.
+// This allows us to approximate a 1-count-per-user stat without recording which
+// user the event came from (bump stats are anonymous).
+//
+// We don't want to block the thread to record the stat, so this function doesn't
+// await promises before returning.
+export function bumpAggregatedUniqueStat(
```
fredrikekelund:
```suggestion
export async function bumpAggregatedUniqueStat(
```
Nit, but this could probably be an async function, right? Would make the function body slightly easier to read if we await'ed instead of then'ed.

fredrikekelund (commented):
Looks good and works as expected 👍 

However, I think we should consider using a shared definition of `StatsGroup` and `StatsMetric` across `cli` and `src`. I did some quick testing, and unfortunately it looks like we might have to move the `getPlatformMetric` and `getImporterMetric` functions out of `src/lib/bump-stats/types.ts` to make it work (because the imported importers cause trouble for Webpack when bundling the CLI code). If it turns out to be a hassle, I say let's skip it, but it's worth a shot, at least.

bcotrim:
Thanks for both reviews!

@fredrikekelund I think I addressed everything 👍 
While refactoring the stats function, I noticed an inconsistency in how we handle `updateLastBump`. When there's no previous stat data (`lastBump === null`), we always update the timestamp regardless of whether `bumpStat` succeeded. However, for subsequent calls, we only update if `bumpStat` returns true.
For example, in development mode where `bumpStat` returns false, we still record the timestamp on the first call, but not on subsequent calls. This seems like a bug.
I've fixed this in the CLI to be consistent (only update when `bumpStat` returns true). If you agree this is a bug, I can fix it in Studio.

fredrikekelund:
> When there's no previous stat data (lastBump === null), we always update the timestamp regardless of whether bumpStat succeeded

I don't know… Always persisting a value to the appdata file when there isn't one already seems like a sensible default to me.

fredrikekelund (approved):
LGTM 👍

gcsecsey (approved):
Changes make sense to me, and I'm getting the stat bumps logged as described:
CleanShot 2025-04-15 at 15 58 12@2x

PR #1207, opened 2025-04-15
First commit: just trying out an email server.

ramonjd:
## Related issues

- Fixes #

## Proposed Changes

Just testing forwarding WordPress emails to nodemailer.

The first iteration created a dedicated express server.

The second sets up a route on each site's express app. Each site has dedicated email host settings. This is more flexible as it allows users to set up different accounts for each site.

### Questions/TODOs:

- [ ] Add settings controls so that users can add their own email server
- [ ] Add https://ethereal.email/ as an option
- [ ] Support SMTP/IMAP/POP
- [ ] For user who can't or won't provide email server details, could we set up an endpoint (with limiting) using WordPress.com's WordPressMailer? OR alternatively, route the emails to a fake "inbox" in Studio?
- [ ] ~~Should the port be dynamic? What would this look like?~~ we can set up a route on each site's express server
- [ ] Is there a better way? Mailpit only runs on Docker/installation
- [ ] What about incoming emails, e.g., post by email. See: import { EmailSettings } from './email-server';??

## Testing Instructions

Add an email and app password (use Gmail for testing, it's easier)
Click on WP Admin link on a site. 
Then add a user, making sure to send them an email. 
See if it works.

-

## Pre-merge Checklist


vendor/wp-now/src/wp-now.ts
```diff
@@ -614,6 +614,27 @@ async function mountInternalMuPlugins( php: PHP, options: WPNowOptions ) {
...
 		);
 	}
+
+	// Intercept wp_mail and send email data via Electron service.
+	// @TODO If we need dynamic ports, think about bundling this in the site folder.
+	php.writeFile(
+		path.posix.join( PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER, '0-wp-mail.php' ),
+		`<?php
+			add_filter( 'wp_mail', function( $args ) {
+				$response = wp_remote_post( 'http://localhost:7777/email', array(
```
ramonjd: TODO: the port number should be dynamic.

It's set in the app.

- PHP const?
- WP option?
- ...

ramonjd: I'm thinking this should be copied to `/mu-plugins/` for each separate site.

This would simplify setting internal constants for the port number.

Also, users might want use a customized or other local email server.

src/lib/email-server.ts
```diff
@@ -0,0 +1,94 @@
...
+
+	async start(): Promise< number > {
+		if ( this.server ) {
+			return this.port!;
+		}
+
+		const app = express();
+		app.use( express.json() );
+
+		// @TODO: service should be configurable via Studio settings.
```
ramonjd: The default would be Ethereal catch all for all sites. 

Should sites be able to use separate email services? For example, a user might want site `x`  to use gmail and not care about other sites? 

Or should settings affect all sites?

wojtekn:
Closing as it's stale. Feel free to reopen if work is restarted.

PR #1211, opened 2025-04-15
Fix bump stat race condition

fredrikekelund:
## Related issues

- Fixes STU-368

## Proposed Changes

The `bumpAggregatedUniqueStat` function writes a `lastBumpStats` object to the appdata file to track when we last bumped a particular stat. This function is called when the app starts (see `src/index.ts`), which is a particularly sensitive time for race conditions in appdata handling (many functions are reading and writing the appdata file in short succession). 

Unfortunately, it looks like we now have such a race condition on our hands, and in my case, it prevented the `local-environment-launch-uniques` stat from persisting. If this issue is widespread, it could mean our analytics data is misrepresentative. 

In my testing, I found the `removeSitesWithEmptyDirectories` function in `src/main-window.ts` to be the culprit. To fix the issue, I moved that function to a standalone module in `src/migrations` and made it handle appdata loading internally. For good measure, I also moved the `moveDatabasesInSitu` to a module in `src/migrations`. 

## Testing Instructions

Code review is key, but to ensure that `local-environment-launch-uniques` is persisted as expected, follow these steps:

1. Open the appdata config file
2. Ensure there is no `local-environment-launch-uniques` key in the `lastBumpStats` object (remove it if there is one)
3. Launch Studio
4. Ensure that a `local-environment-launch-uniques` key is written to the appdata file

## Pre-merge Checklist


bcotrim (approved):
Works as expected and LGTM, nice find! 👍

PR #1215, opened 2025-04-16
Studio: Extend the list of available terminals / shells 

katinthehatsite:
## Related issues

Related to: STU-364

## Proposed Changes

This PR implements a logic for opening different terminals to be similar to the logic for opening different code editors. We save the preferred terminal to appData (with Terminal set as default). Then, we retrieve this data to display in the `Overview` tab. 

Screenshot 2025-04-16 at 1 15 31 PM

Screenshot 2025-04-16 at 1 17 33 PM

Screenshot 2025-04-16 at 1 18 03 PM

Screenshot 2025-04-16 at 1 18 16 PM

## Testing Instructions

* Run `npm install` 
* Navigate to the `User settings`
* Click on `Preferences` tab
* In the `Terminal` section, select Terminal (you should see only the Terminals you have installed, so if you have iTerm, you will see it there as well. I will change this later once I clean up the code)
* Save your settings
* Navigate to the `Overview` tab
* Confirm that you see Terminal in there in `Open in...` section
* Click on it and confirm it works
* Navigate to the `User settings` again
* In the `Terminal` section, select iTerm
* Save your settings
* Confirm that you see iTerm in there in `Open in...` section
* Confirm that you can open it and the site opens in iTerm

So far, this PR adds support only for iTerm but we will add more terminal apps in the followup. 

## Pre-merge Checklist


src/components/user-settings.tsx
```diff
@@ -221,22 +222,51 @@ const PreferencesTab = ( { onClose }: { onClose: () => void } ) => {
...
 		onClose();
 	};
 
-	const hasChanges = locale !== savedLocale;
+	const hasChanges = locale !== savedLocale || terminal !== savedTerminal;
 
 	return (
 		<>
 			<LanguagePicker value={ locale } onChange={ setLocale } />
+			<div className="mt-6">
```
katinthehatsite: I added this field just for testing for now so ignore the design here, we will use the design from Rahul's PR.

katinthehatsite:
@gcsecsey @gavande1 this is still a draft but feel free to give it an early test. As mentioned in the comments, we can skip the design itself, I simply added it for testing for now. We will take an actual design from Rahul's PR

gavande1:
Thanks @katinthehatsite, great work! I have tested this and everything seems to be working as expected. The only issue I noticed is that the terminal name didn't get updated in the overview tab after selecting the new shell and saving it. I had to switch tabs in order to see the updated terminal app. I think you mentioned the same problem in your test instructions.

katinthehatsite:
Yeah @gavande1  - I am looking into it, I think we can probably fetch the data on window focus. I will double-check what is best here though

katinthehatsite:
The UI change in the Overview tab should be working correctly with https://github.com/Automattic/studio/pull/1215/commits/3844d3ffc3210dba27f16170c6708933e94758f0

src/ipc-handlers.ts
```diff
@@ -1218,6 +1233,25 @@ export async function checkSyncBackupSize(
...
+	supportedTerminal: SupportedTerminal
+) {
+	const userData = await loadUserData();
+	await saveUserData( {
+		...userData,
+		supportedTerminal: supportedTerminal,
+	} );
+
+	// Notify renderer processes that the terminal preference has changed
+	sendIpcEventToRenderer( 'terminal-preference-changed' );
```
gavande1: What do you think about using a general name like `user-preferences-changed`? This could help us avoid the need to introduce a special event for each new setting!

gavande1:
@katinthehatsite, The UI change in the Overview tab is working as expected. I like the elegant solution. I have left a minor comment about the event name.

src/ipc-handlers.ts
katinthehatsite: Suggestion applied 👍

src/components/user-settings.tsx
katinthehatsite: Removed for now and integrated with Rahul's PR

gcsecsey (approved):
The changes LGTM and work well. Thanks for updating the tests as well.

src/ipc-handlers.ts
```diff
@@ -1232,6 +1247,25 @@ export async function checkSyncBackupSize(
...
+		supportedTerminal: supportedTerminal,
+	} );
+
+	// Notify renderer processes that the terminal preference has changed
+	sendIpcEventToRenderer( 'user-preference-changed' );
+}
+
+export async function getUserTerminal( _event: IpcMainInvokeEvent ): Promise< SupportedTerminal > {
+	const userData = await loadUserData();
+	return userData.supportedTerminal as SupportedTerminal;
```
fredrikekelund, line 1266: The `supportedTerminal` key returned by `loadUserData` has a `SupportedTerminal | undefined` type. Shouldn't we account for the possibility that the value is `undefined` here?

src/components/content-tab-overview.tsx
```diff
@@ -139,6 +140,21 @@ function CustomizeSection( {
...
+		setTerminalName( terminal === 'iterm' ? __( 'iTerm' ) : __( 'Terminal' ) );
+	};
+
+	useIpcListener( 'user-preference-changed', () => {
+		updateTerminalName();
+	} );
+
+	useEffect( () => {
+		updateTerminalName();
+	}, [] );
```
fredrikekelund, line 156: Consider this a suggestion, but it strikes me that storing the app key returned by `getIpcApi().getUserTerminal()` as state and using a map to associate app keys with app names would be easier to scale.

katinthehatsite: Yeah I am planning to adjust it in https://github.com/Automattic/studio/pull/1240/files#diff-656c73dd35e01f5f64e895d19dd484d168661e56deffe46d567e257b49ddc9a4R148 - it is still a draft but we can introduce it in that PR 👍

src/ipc-handlers.ts
katinthehatsite: Let's adjust it in https://github.com/Automattic/studio/pull/1265/commits/56bb680c1741ce9339c4c32bcda80b719fdcc255
Thanks for pointing this out 🙇

PR #1216, opened 2025-04-16
User Settings: Add Shell selector dropdown to the Preferences tab

gavande1:
## Related issues

- Fixes STU-360

## Proposed Changes

- Add terminal shell selector dropdown on preferences tab
- This PR does not handle:
     - Saving preference is app data.
     - Looking for actually installed terminals
     - Disable dropdown when none of the supported terminal is found.

## Testing Instructions

- Check out this branch
- Open settings
- Goto Preference Tab
- Make sure the terminal dropdown is visible
- Change the value and notice that the `Save` button is enabled. 
- You will be alerted with saved value when clicked on `Save`.

## Pre-merge Checklist

### DIFF
--- a/src/components/terminal-picker.tsx
+++ b/src/components/terminal-picker.tsx
@@ -1,0 +1,57 @@
+import { SelectControl } from '@wordpress/components';
+import { useI18n } from '@wordpress/react-i18n';
+import { useEffect, useState } from 'react';
+import { SupportedTerminal, supportedTerminalNames } from 'src/lib/terminal';
+
+interface TerminalPickerProps {
+	value: SupportedTerminal;
+	onChange: ( value: SupportedTerminal ) => void;
+}
+
+export const TerminalPicker = ( { value, onChange }: TerminalPickerProps ) => {
+	const { __ } = useI18n();
+	const [ installedTerminals, setInstalledTerminals ] = useState< {
+		iterm: boolean | null;
+		terminal: boolean | null;
+	} >( {
+		iterm: null,
+		terminal: null,
+	} );
+
+	useEffect( () => {
+		const fetchInstalledTerminals = async () => {
+			try {
+				const terminals = { iterm: true, terminal: true }; // TODO:: Fetch installed terminals from the API
+				setInstalledTerminals( terminals );
+			} catch ( error ) {
+				console.error( 'Failed to fetch installed terminals:', error );
+			}
+		};
+
+		fetchInstalledTerminals();
+	}, [] );
+
+	const options = Object.entries( supportedTerminalNames ).map( ( [ terminal, label ] ) => {
+		const terminalKey = terminal as SupportedTerminal;
+		const isInstalled = installedTerminals[ terminalKey as keyof typeof installedTerminals ];
+
+		return {
+			value: terminalKey,
+			label,
+			disabled: ! isInstalled,
+		};
+	} );
+
+	return (
+		<div className="flex gap-5 flex-col">
+			<h2 className="a8c-subtitle-small">{ __( 'Shell' ) }</h2>
+			<SelectControl
+				value={ value }
+				onChange={ onChange }
+				options={ options }
+				__nextHasNoMarginBottom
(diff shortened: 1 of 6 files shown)


src/modules/user-settings/hooks/use-terminal-data.ts
```diff
@@ -0,0 +1,54 @@
...
+		//TODO:: Save the terminal preference to the IPC API
+		alert( terminal );
+	};
+
+	// Reset terminal to saved value
+	const resetTerminal = () => {
+		setTerminal( savedTerminalValue );
+	};
+
+	// Check if terminal has changes
```
katinthehatsite: I think we can clean up these comments, they are pretty self-explanatory

src/modules/user-settings/components/terminal-picker.tsx
```diff
@@ -0,0 +1,57 @@
...
+		return {
+			value: terminalKey,
+			label,
+			disabled: ! isInstalled,
+		};
+	} );
+
+	return (
+		<div className="flex gap-5 flex-col">
+			<h2 className="a8c-subtitle-small">{ __( 'Shell' ) }</h2>
```
katinthehatsite: Ideally, we should capitalize the titles on the modal fields to be consistent with the Figma design: RToz6tIuQ7nlZrikBte4GU-fi-7079_212368

We can also open a separate PR to polish the modal after though. I also see we now end up with scrollbars which are also not in the original design

katinthehatsite (approved):
It looks good and works as expected. Some minor comments but we can handle the followup with modal cleanup in a separate issue

src/modules/user-settings/hooks/use-terminal-data.ts
gavande1: Fixed in [4214ee6](https://github.com/Automattic/studio/pull/1216/commits/4214ee647edd9cede49e98afef3d016a3e9a92a0)

src/modules/user-settings/components/terminal-picker.tsx
gavande1: Fixed in [b12c433](https://github.com/Automattic/studio/pull/1216/commits/b12c4336b10d1e3a81e312ec63ad40a6189e99f9)

src/modules/user-settings/components/terminal-picker.tsx
```diff
@@ -44,7 +44,7 @@ export const TerminalPicker = ( { value, onChange }: TerminalPickerProps ) => {
 
 	return (
 		<div className="flex gap-5 flex-col">
-			<h2 className="a8c-subtitle-small">{ __( 'Shell' ) }</h2>
+			<h2 className="a8c-subtitle-small">{ __( 'SHELL' ) }</h2>
```
katinthehatsite: Should we apply `uppercase` to the parent `div` instead so that the labels for each fields change consistently?

gavande1: I think we need a better solution to this. I am going to revert my change. We can follow up with a better solution to this.

katinthehatsite: Okay sounds good, if you want to merge this, I will rebase my PR ad we can test the full path.

katinthehatsite: I create a follow up issue for polishing modal design here: STU-374

gavande1: Perfect. Thanks for creating an issue.

gcsecsey (approved):
Changes LGTM and work great 

I have a minor comment: when using the default window size the app starts with, the preferences tab now overflows its original size, so the buttons aren't fully visible:
CleanShot 2025-04-16 at 15 11 55@2x

I think a good solution to this would be to reduce the margin between the sections, but we can do this in a follow-up.

gavande1:
Thanks @gcsecsey for pointing that out. We discussed it with @katinthehatsite and will handle it as part of STU-374.

Issue #63, opened 2024-04-30
Feature Request: Allow users to select code editor and terminal app


matt-west:
### What

Studio currently supports quick actions for VS Code, PhpStorm, and the Terminal app. Users should be able to choose any code editor/terminal they have installed so the quick actions match their preference.

### Why

* People are more likely to use these quick actions if they can customise them to open their favourite tools.
* Studio currently prioritises VS Code over PhpStorm for users that have both installed. This may not match the user's preference.

### How

Here’s how GitHub Desktop handles this:

<img width="1072" alt="Screenshot 2024-04-30 at 09 15 54" src="https://github.com/Automattic/studio/assets/417538/530c3af7-36e5-4363-9466-755bc52493af">

faisalahammad:
+1

bacoords:
+1

matt-west:
@michalczaplinski I did a quick pass of the app settings screen to facilitate adding the options to select a code editor and shell.

I think we’ll need to implement tabs inside the modal as there’s too much in here now to fit in one page comfortably. It adds some complexity to this task, but I foresee us adding more options to app settings in the future so I think it's worth the extra time investment.

We can place these new fields in a _Preferences_ tab alongside the Language selector:

![Image](https://github.com/user-attachments/assets/13e364ae-cf24-4326-94f8-a52bc3614ba9)

Usage metrics have moved to a dedicated tab. This would only be visible to users that have connected a WordPress.com account as these features are unavailable otherwise.

![Image](https://github.com/user-attachments/assets/e21a5c91-2359-42e9-bec1-6fc8d294c22c)

The default tab would be _Accounts_, with options to connect a WordPress.com or Pressable account. (@crisbusquets I’ve tried to integrate your work on Pressable Sync here.)

![Image](https://github.com/user-attachments/assets/f7afcc60-c80a-437b-ac9a-a0da9820726d)

![Image](https://github.com/user-attachments/assets/c609fb74-b2cd-4f2f-ad74-5fe79d2f8721)

Full designs can be found here: RToz6tIuQ7nlZrikBte4GU-fi-2658_81168

Let me know if you have any thoughts or feedback on these designs.

crisbusquets:
> The default tab would be Accounts, with options to connect a WordPress.com or Pressable account. (@crisbusquets I’ve tried to integrate your work on Pressable Sync here.)

[...]

PR #1217, opened 2025-04-16
User Settings: return the list of installed editors

gcsecsey:
## Related issues

Fixes STU-363

## Proposed Changes

> [!NOTE]  
> I tried out the new PR Cursor template (pesfPa-1sj-p2) to produce a more detailed description.

This PR enhances the IDE detection functionality to support a wider range of editors across different platforms (macOS, Windows, Linux) with improved reliability and test coverage.

The existing IDE detection was limited in scope and didn't account for the various ways different platforms store and name their applications. Each platform (macOS, Windows, Linux) has its own conventions for:
- Application naming (e.g., `.app` bundles on macOS vs executable names on Linux)
- Installation paths (e.g., `/Applications` vs `Program Files` vs `/usr/bin`)
- User-specific vs system-wide installations

While I attempted to find a more generalized solution, the fundamental differences between platforms made it necessary to implement platform-specific path detection. While more verbose, this approach ensures reliable detection across all supported platforms.

- Added support for detecting additional editors:
  - Cursor
  - Windsurf
  - Nova (macOS only)
  - WebStorm
  - Sublime Text
  - Atom

- Enhanced JetBrains IDE detection:
  - Added version-specific folder detection on Windows
  - Better path handling for PhpStorm and WebStorm

- Added comprehensive test coverage:
  - Platform-specific test suites
  - Mock filesystem and electron app paths
  - Test cases for all supported editors
  - Edge cases and platform-specific scenarios

#### Unit tests
The changes include a new test suite that covers:
- Editor detection on all supported platforms
- Various installation locations
- Platform-specific edge cases
- Mac-only editor handling on other platforms

Run the new tests using `npm run test -- src/lib/tests/is-installed.test.ts`. All tests should pass.

The new tests contain some path manipulation that were causing some issue when run on Windows. After fixing these, I ran the tests both on Mac and Windows to double-check:
On Mac:
CleanShot 2025-04-16 at 12 02 53@2x

On Windows:
CleanShot 2025-04-17 at 13 12 00@2x

#### Manual testing

- Start the app with `STUDIO_PREFERRED_EDITOR=true npm start`
- Go to user preferences
- Check that the list of editors contains the installed editors as active options
- Check that the list of terminals contains the installed terminals as active options

| editors | terminals |
|--------|--------|
| image | image | 

## Pre-merge Checklist

- [X] Have you checked for TypeScript, React or other console errors?


src/lib/is-installed.ts
```diff
@@ -2,39 +2,114 @@ import { app } from 'electron';
...
+		nova: [ 'Nova.app' ],
+		webstorm: [ 'WebStorm.app' ],
+		sublime: [ 'Sublime Text.app' ],
+		atom: [ 'Atom.app' ],
+	},
+	win32: {
 		vscode: [
-			path.join( systemApplications, 'Visual Studio Code.app' ),
-			path.join( userApplications, 'Visual Studio Code.app' ),
+			'C:\\Program Files\\Microsoft VS Code',
```
wojtekn: Can we use variables from electron app instead of hardcoding path and disk letter?

wojtekn (commented):
It works for my limited use case of having PhpStorm and VS Code, but I didn't have a chance to test for other editors or systems.

One thing I'm wondering about is the amount of editors. Do we want to add a support for a long list of editors, and then commit to maintain and test it for all for future releases?

Thanks for adding tests!

gcsecsey:
> It works for my limited use case of having PhpStorm and VS Code, but I didn't have a chance to test for other editors or systems.

Yes, I ran into the same issue, even having installed some apps I can only test on Mac. I'm working on setting up Windows and Linux VMs on my proxmox box, and will try to test my assumptions of these install paths before merging.

> One thing I'm wondering about is the amount of editors. Do we want to add a support for a long list of editors, and then commit to maintain and test it for all for future releases?

Yes, this is an interesting question. From a maintenance burden standpoint, I think it'd be better to have only a handful of options and add a custom option where the user can browse for an arbitrary executable on their system. On https://github.com/Automattic/studio/issues/63#issuecomment-2729071936 this was discussed earlier but you shared that there we have app-specific code which would limit us to do something like that. Can you elaborate on this a bit? Is this related to the opening logic we do currently in `openFileinIDE`?

https://github.com/Automattic/studio/blob/4aad06b5a71385944433dbabceeed38657378609/src/ipc-handlers.ts#L1102-L1128

src/lib/is-installed.ts
gcsecsey: That's a good point, updated it in a78fcca

src/lib/is-installed.ts
```diff
@@ -2,39 +2,139 @@ import { app } from 'electron';
...
-let appPaths: Record< keyof InstalledApps, string[] >;
+type InstalledApp =
+	| 'vscode'
+	| 'phpstorm'
+	| 'cursor'
+	| 'windsurf'
+	| 'nova'
+	| 'webstorm'
+	| 'sublime'
+	| 'atom';
```
gcsecsey: Couple of notes regarding this list:
- atom was archived in 2022
- nova is mac-only

wojtekn:
> Yes, I ran into the same issue, even having installed some apps I can only test on Mac. I'm working on setting up Windows and Linux VMs on my proxmox box, and will try to test my assumptions of these install paths before merging.

We can remove the Linux part at this point, or just leave what was supported there. We don't officially support Linux and we don't provide a Linux binary, yet, and I think we shouldn't dedicate too much time for Linux-specific features at this point.
 
> [...] you shared that there we have app-specific code which would limit us to do something like that. Can you elaborate on this a bit? Is this related to the opening logic we do currently in `openFileinIDE`?

Yes, the reason is that we couldn't find a unified way to open the project in the editor. I you look into `openFileinIDE`, you will see that we use the different deep links for PhpStorm and VS Code, and I could expect it will be different for other IDEs, too.

src/lib/is-installed.ts
```diff
@@ -2,53 +2,122 @@ import { app } from 'electron';
...
+		phpstorm: [ '/usr/bin/phpstorm' ],
+		cursor: [ '/usr/bin/cursor' ],
+		windsurf: [ '/usr/bin/windsurf' ],
+		nova: [],
+		webstorm: [ '/usr/bin/webstorm' ],
+		sublime: [ '/usr/bin/sublime' ],
+		atom: [ '/usr/bin/atom' ],
+		iterm: [],
+		terminal: [],
+	},
```
gcsecsey: @katinthehatsite when rebasing this PR after #1215 was merged, I found it easier to merge the list of terminals and apps into one list, and do the same path manipulations for these as the IDEs.

I only had some issues with the typing, currently I use `InstalledApps | InstalledTerminals` where we need to be able to handle both. Longer term, I think it'd make sense to maintain just one type of the possible apps/terminals we support.

gavande1 (commented):
Thanks, @gcsecsey, for working on this. It looks great to me!

I do have one concern similar to what Wojtek mentioned. Are we sure we want to support this many editors right from the start? It might make testing and maintenance more challenging.

Would it make sense to start with a smaller set of editors and gradually expand support, either later on or during the polishing phase?

gavande1 (approved):

src/lib/is-installed.ts
katinthehatsite: Yeah I am fine with this approach for now considering that the terminal apps are still applications and are located at the same path as other applications 👍

src/lib/tests/is-installed.test.ts
```diff
@@ -0,0 +1,132 @@
...
+	app: {
+		getPath: jest.fn(),
+	},
+} ) );
+
+describe( 'isInstalled', () => {
+	let isInstalled: ( key: string ) => boolean;
+	let mockPaths: string[];
+
+	// Reset mocks before each test
```
katinthehatsite: Let's clean up the comments as they seem to be pretty self explanatory unless it is something that is not evident

katinthehatsite (approved):
The changes look good to me and I can see, that for instance, it is now finding Cursor for me:

Screenshot 2025-04-22 at 11 54 48 AM

It would be nice to get this tested on Windows as well to see how it turns there.

gcsecsey:
> I do have one concern similar to what Wojtek mentioned. Are we sure we want to support this many editors right from the start? It might make testing and maintenance more challenging.
> 
> Would it make sense to start with a smaller set of editors and gradually expand support, either later on or during the polishing phase?

I agree that it might be too much to start off with this many editors. As I mentioned [above](https://github.com/Automattic/studio/pull/1217#discussion_r2046897787), I don't see too much benefits of supporting Nova and Atom at all, and I think we can start with a smaller subset and itroduce the rest later.

As we need to support PHPStorm and VSCode, I think adjacent editors to these would be the easiest to start off with. For now, I'll update the list to be these editors, then we can iterate on these later.
- VSCode
- PHPStorm
- Cursor
- Windsurf
- Webstorm

PR #1218, opened 2025-04-16
Update existing snapshots appdata entry when preview site is updated

fredrikekelund:
## Related issues

- Fixes STU-373

## Proposed Changes

Previously, the `preview update` command would insert a duplicate `snapshots` entry in the appdata when it finished. This PR makes it so the existing entry is updated instead. 

Because https://github.com/Automattic/studio/pull/1201 also touches a lot of the same files (the `upsertPreviewSiteInAppdata` function specifically), I checked out the changes from that branch, too. 

## Testing Instructions

1. Have an existing preview site
2. Open the appdata file in your code editor
3. Run `npm run cli:build`
4. Run `node dist/cli/main.js preview update PATH_TO_SITE -h PREVIEW_SITE_HOSTNAME`
5. Ensure that there are no duplicates in the `snapshots` array in the appdata file and that the `date` prop was updated

## Pre-merge Checklist

### DIFF
--- a/cli/lib/snapshots.ts
+++ b/cli/lib/snapshots.ts
@@ -1,11 +1,4 @@
-import { z } from 'zod';
-import {
-	getSiteIdFromFolder,
-	readAppdata,
-	saveAppdata,
-	snapshotSchema,
-	Snapshot,
-} from 'cli/lib/appdata';
+import { getSiteFromFolder, readAppdata, saveAppdata, Snapshot } from 'cli/lib/appdata';
 import { LoggerError } from 'cli/logger';
 
 export async function getSnapshotsFromAppdata(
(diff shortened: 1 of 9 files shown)


cli/lib/snapshots.ts
```diff
@@ -17,25 +10,23 @@ export async function getSnapshotsFromAppdata(
...
 	return snapshots;
 }
 
-export async function addPreviewSiteToAppdata(
-	previewUrl: string,
+// Insert or update a snapshot entry (matching on atomicSiteId) in the appdata file.
+export async function upsertPreviewSiteInAppdata(
+	siteFolder: string,
 	atomicSiteId: number,
-	siteFolder: string
```
fredrikekelund, line 30: Reordering the arguments wasn't necessary, it's just something I found intuitive. `siteFolder` comes first because it defines the site we're working with (and the function throws if this value doesn't map to an existing site). `atomicSiteId` comes second because that's the identifier used to update existing snapshots. `previewUrl` comes last because it doesn't have any side effects.

nightnei (approved):
LGTM and works as expoected - confirms that only `date` were updated and no duplicates 👍

PR #1219, opened 2025-04-16
CLI: Improve size validation error handling on preview site create

bcotrim:
## Proposed Changes
- Added `common` folder for code shared between CLI and Studio
- Added validation for file size when creating a preview site using the CLI

| Studio | CLI |
|--------|--------|
| image | image |

## Testing Instructions
- Run `npm start`
- Create a new site.
- Validate that you can create a preview site using Studio and CLI
- Create another site in Studio that exceeds 2GB (by adding a large file in wp-content for example)
- Validate that you can't create a preview site using Studio or CLI

## Pre-merge Checklist


cli/lib/validation.ts
```diff
@@ -1,9 +1,12 @@
 import fs from 'fs';
+import path from 'path';
 import { __, sprintf } from '@wordpress/i18n';
+import { DEMO_SITE_SIZE_LIMIT_BYTES, DEMO_SITE_SIZE_LIMIT_GB } from 'common/constants';
+import { calculateDirectorySize } from 'common/lib/fs-utils';
 import { isWordPressDirectory } from 'src/lib/fs-utils';
```
fredrikekelund, line 6: Maybe we could move this function to `common/lib/fs-utils`, too?

fredrikekelund (approved):
Looks good and tests well 👍 I left one suggestion, but it's not necessarily a blocker

cli/lib/validation.ts
bcotrim: I was thinking about adding a follow-up PR that moves all existing functions to the `common` folder if that sounds good.

fredrikekelund: Sounds good 👍

PR #1222, opened 2025-04-16
Add common folder structure & some adjustments

bcotrim:
## Related Issues

- Fixes STU-378

## Proposed Changes

-  Refactored code organization by moving shared types and constants to `common/` directory
  - Moved `Snapshot` type to `common/types/snapshot.ts`
  - Moved `StatsGroup` and `StatsMetric` types to `common/types/stats.ts`
  - Moved time constants (`HOUR_MS`, `DAY_MS`) to `common/constants.ts`
- Split validation logic in `cli/lib/validation.ts` into two separate functions:
  - `validateSiteFolder`: Synchronous validation of WordPress directory structure
  - `validateSiteSize`: Asynchronous validation of site size limits
- Consolidated filesystem utilities by moving `isWordPressDirectory` to `common/lib/fs-utils.ts`
- Updated translation file paths from `src/translations` to `common/translations`
- Fixed duplicate import in `cli/commands/preview/tests/list.test.ts`
- Updated `make-pot` script to include CLI files in translation extraction

## Testing Instructions

- Run `npm start`
- Create a new WordPress site or use an existing one
- Run the preview create command with the site path
- Verify that validation works correctly:
  - Site folder validation (WordPress structure)
  - Site size validation (2GB limit)
- Test with different locales to ensure translations are working
- Run tests to verify the refactored validation and utility functions
- Run `npm run make-pot` and confirm translation from cli code

## Pre-merge Checklist


common/types/snapshot.ts
```diff
@@ -0,0 +1,12 @@
...
+export const snapshotSchema = z.object( {
+	url: z.string(),
+	atomicSiteId: z.number(),
+	localSiteId: z.string(),
+	date: z.number(),
+	name: z.string(),
+	userId: z.number().optional(),
+} );
+
+export type Snapshot = z.infer< typeof snapshotSchema >;
```
fredrikekelund, line 12: The `Snapshot` type definition in `src` lives in `src/ipc-types.d.ts`, and it's not identical to this one (it also has `isLoading` and `isDeleting` props). https://github.com/Automattic/studio/pull/1201, https://github.com/Automattic/studio/pull/1223 and the soon-to-be deletion PR should remove the need for those props, though, so maybe we can proceed with using a shared type definition after all.

package.json
```diff
@@ -33,7 +33,7 @@
 		"test:watch": "cross-env NODE_OPTIONS='--no-deprecation' jest --watch",
 		"e2e": "npx playwright install && npx playwright test",
 		"test:metrics": "npx playwright test --config=./metrics/playwright.metrics.config.ts",
-		"make-pot": "wp-babel-makepot \"./src/**/*.{js,jsx,ts,tsx}\" --ignore \"**/*.d.ts\" --base \"./src\" --dir \"./out/pots\" --output \"./out/pots/bundle-strings.pot\""
+		"make-pot": "wp-babel-makepot \"{./src,./cli}/**/*.{js,jsx,ts,tsx}\" --ignore \"**/*.d.ts\" --base \".\" --dir \"./out/pots\" --output \"./out/pots/bundle-strings.pot\""
```
fredrikekelund, line 36: Nice 👍

common/constants.ts
```diff
@@ -1,2 +1,8 @@
 export const DEMO_SITE_SIZE_LIMIT_GB = 2;
 export const DEMO_SITE_SIZE_LIMIT_BYTES = DEMO_SITE_SIZE_LIMIT_GB * 1024 * 1024 * 1024; // 2GB
+export const HOUR_MS = 1000 * 60 * 60;
+export const DAY_MS = HOUR_MS * 24;
+// AI Assistant constants
+// IMPORTANT: When updating this value, we need to update the string located in `AIClearHistoryReminder` component.
+// Reference: https://github.com/Automattic/studio/blob/3dd5c58cdb7998e458d191e508e8e859177225a9/src/components/ai-clear-history-reminder.tsx#L78
+export const CLEAR_HISTORY_REMINDER_TIME = 2 * HOUR_MS; // In milliseconds
```
fredrikekelund: This doesn't strike me as a constant that needs to live in the `common` directory 🤔 It's only used in `src/components`

fredrikekelund (approved):
LGTM with the caveat that `CLEAR_HISTORY_REMINDER_TIME` probably shouldn't live in the `common` directory.

common/constants.ts
bcotrim: Having an alias import (like `common/constant`) in `src/constants` was causing some issues with the `postinstall` scripts.
I've made the adjustments in https://github.com/Automattic/studio/pull/1222/commits/c554a7dac118ad3dda573600d36139172400ce11 that seems to be an "ok workaround"
Let me know which one you prefer.

fredrikekelund: Ugh, that's frustrating. Do you know which scripts was causing trouble? The latest changes LGTM 👍

bcotrim: I think it was this one: https://github.com/Automattic/studio/blob/trunk/scripts/download-wp-server-files.ts#L5
I don't think the constant that we are importing in that script is used anywhere else, so we could also change the script, but it will always be a potential issue, so I think it's best to just reference it and have a relative import.

PR #1223, opened 2025-04-17
Refactor preview site updating to depend on CLI

fredrikekelund:
## Related issues

- Fixes STU-241

## Proposed Changes

- Refactor preview site updating functionality to depend on CLI.
- Persist changes to appdata file when preview sites are renamed. 

## Testing Instructions

1. Ensure that updating a preview site works as expected. 
2. Ensure that preview site renamings are persisted when reloading the app.

## Pre-merge Checklist


src/components/root.tsx
```diff
@@ -14,7 +14,6 @@ import { PromptUsageProvider } from 'src/hooks/use-prompt-usage';
 import { SiteDetailsProvider } from 'src/hooks/use-site-details';
 import { SnapshotProvider } from 'src/hooks/use-snapshots';
 import { ThemeDetailsProvider } from 'src/hooks/use-theme-details';
-import { DemoSiteUpdateProvider } from 'src/hooks/use-update-demo-site';
```
nightnei, line 17: 😍

src/stores/snapshot-slice.ts
```diff
@@ -45,6 +57,24 @@ const createSnapshot = createAsyncThunk(
...
 );
 
+const updateSnapshot = createAsyncThunk(
+	'snapshot/updateSnapshot',
+	async (
+		{ atomicSiteId, siteFolder }: { atomicSiteId: number; siteFolder: string },
+		thunkAPI
+	) => {
+		const state = thunkAPI.getState() as RootState;
+		const snapshot = state.snapshot.snapshots.find(
```
nightnei, line 67: `state.snapshot.snapshots.find(`

NIT but I am thinking - maybe we should use a less generic slice name, instead of `snapshot`, maybe to add some postfix, to simplify reading code. WDYT?

nightnei (approved):
LGTM and updating works es expected 👍

src/modules/cli/lib/execute-preview-command.ts
```diff
@@ -0,0 +1,79 @@
+import crypto from 'crypto';
+import { z } from 'zod';
+import { CreateLoggerAction } from 'cli/commands/preview/logger-actions';
```
bcotrim, line 3: Now that we have the `common` folder let's try to move code there.

bcotrim (approved):
Tested as described
LGTM 👍

src/stores/snapshot-slice.ts
fredrikekelund: `snapshot.snapshots` could be somewhat confusing at first glance, but changing the store name to something like `snapshotStore` or `snapshotSlice` seems redundant to me, and for consistency's sake, it would also require us to change the other slice names (adding even more redundant naming). Given that, I would prefer to keep the name as-is.

src/modules/cli/lib/execute-preview-command.ts
fredrikekelund: Fixed in https://github.com/Automattic/studio/pull/1230 👍

PR #1224, opened 2025-04-17
Add Cannot allocate Wasm memory error message

bgrgicak:
## Proposed Changes

- Add a error message for the `Cannot allocate Wasm memory` error

## Testing Instructions

- Add `throw new Error( 'failed to asynchronously prepare wasm: RangeError: WebAssembly.instantiate(): Out of memory: Cannot allocate Wasm memory for new instance' );` before this line https://github.com/Automattic/studio/blob/e720f7814ef117d6add98e1e395912e14052a638/src/site-server.ts#L131
- Start Studio
- Start a site
- See the new error modal

If you want to trigger a real error, [follow these instructions to recreate the issue](https://linear.app/a8c/issue/STU-110/rangeerror-webassemblyinstantiate-out-of-memory-cannot-allocate-wasm#comment-85e0a8a9) and use a [Studio build from Buildkite](https://buildkite.com/automattic/studio/builds/5606#019642fb-b050-4ecd-8f92-d135855e4747).

## Pre-merge Checklist

Screenshot 2025-04-17 at 11 33 49


ivan-ottinger (approved):
The proposed change looks good and works as expected. 👍🏼 

I have tested it manually on macOS (by throwing the error as you suggested in the test instructions):

Markup on 2025-04-17 at 12:40:30

bgrgicak:
@wojtekn when you get a chance, could you please take a look at the error message content and let me know if you have any improvement suggestions?

src/hooks/use-site-details.tsx
```diff
@@ -295,6 +295,17 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
 						showOpenLogs: true,
 					} );
+				} else if (
+					error instanceof Error &&
+					error.message.includes( 'Cannot allocate Wasm memory for new instance' )
+				) {
+					getIpcApi().showErrorMessageBox( {
+						title: __( 'Not enough memory to start the site server' ),
+						message: __(
+							'Please stop some of your running sites and try again. If this problem persists, try closing other apps that might be using memory and try again.'
```
gavande1: Thanks, @bgrgicak, for the improvements. Regarding the copy, I see that "try again" is repeated twice. I am suggesting an alternative below that avoids this. 

```suggestion
							'Please stop some of your running sites first. If this problem persists, try closing other apps that might be using memory and try again.'
```

bgrgicak: Thanks! Your version sounds better.

gcsecsey (approved):
These changes LGTM and work well. 🚢 

| macOS | Windows |
|--------|--------|
| CleanShot 2025-04-22 at 12 56 32@2x | CleanShot 2025-04-22 at 13 13 29@2x |

wojtekn:
@bgrgicak message looks good, thanks.

Would it make sense to catch this error in the IPC handler and throw the message/code like we do with Proxy errors? Could we ensure we don't log this error to Sentry?

bgrgicak:
> Would it make sense to catch this error in the IPC handler and throw the message/code like we do with Proxy errors? Could we ensure we don't log this error to Sentry?

@wojtekn I don't have much experience with Sentry but if I understand correctly this is done using `Sentry.captureException` and  it should work now after I moved the error check to `startServer`. https://github.com/Automattic/studio/pull/1224/files#diff-4897e47ea3fb6cdbb1798e8dce993fa538a1d6d56b425eabbf82d945889ec508R418-R428

I tested the latest commit on a slow Windows machine to endure I can recreate the real error.

wojtekn:
@bgrgicak I think that if we catch WASM error, rethrow custom exception and display nice message about the error, we shouldn't log the exception to Sentry:

```
```

bgrgicak:
> @bgrgicak I think that if we catch WASM error, rethrow custom exception and display nice message about the error, we shouldn't log the exception to Sentry:

I created a new PR to address this: https://github.com/Automattic/studio/pull/1302.

PR #1225, opened 2025-04-17
Update user-facing CLI preview copy

nightnei:
## Proposed Changes
In Studio UI we use "preview" instead of "snapshot", also CLI command is called "preview", so this PR renames output

Screenshot 2025-04-17 at 13 18 07
Screenshot 2025-04-17 at 13 18 15

## Testing Instructions

Visual review shoudl suffice

### DIFF
--- a/cli/lib/appdata.ts
+++ b/cli/lib/appdata.ts
@@ -120,7 +120,7 @@ export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken
 	}
 }
 
-export async function getSiteFromAppData(
+export async function getSiteByFolder(
 	siteFolder: string
 ): Promise< z.infer< typeof siteSchema > > {
 	const userData = await readAppdata();
--- a/cli/lib/appdata.ts
+++ b/cli/lib/appdata.ts
@@ -120,7 +120,7 @@ export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken
 	}
 }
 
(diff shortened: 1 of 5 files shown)


cli/lib/appdata.ts
```diff
@@ -6,7 +6,7 @@ import { StatsGroup, StatsMetric } from 'src/lib/bump-stats/types';
 import { z } from 'zod';
 import { LoggerError } from 'cli/logger';
 
-export const snapshotSchema = z.object( {
```
nightnei: it's not used outside

cli/lib/appdata.ts
```diff
@@ -129,7 +129,7 @@ export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken
 	}
 }
 
-export async function getSiteFromFolder(
+export async function getSiteFromAppData(
```
nightnei: `getSiteFromFolder` looks a bit confusing, I think it shoudl be eitehr `getSiteBYFolder` or `getSiteFromAPPDATA`  or `getSiteFromAPPDATAbyFOLDER

WDYT Fredrik?

fredrikekelund: `getSiteByFolder` sounds like a good option to me 👍

fredrikekelund (approved):
I would call the function `getSiteByFolder`, but other than that, this LGTM 👍

PR #1232, opened 2025-04-17
CLI: refactor cli from commander to yargs

bcotrim:
## Related issues

- Fixes STU-355.
- Fixes STU-379.

## Proposed Changes

- Refactor CLI to use `yargs`
- Adjusted unit tests
- Added locale config to CLI

You can check more details about supported locales [here](https://yargs.js.org/docs/#api-reference-localelocale)

## Testing Instructions

- Run `npm run cli:build`
- Run `npm run start` and change your language in Studio
- Run `node dist/cli/main.js --help` and confirm the default yargs strings are translated
- Test other cli preview commands to ensure it's working as before

## Pre-merge Checklist


cli/index.ts
```diff
@@ -1,45 +1,38 @@
...
+		} );
 
-	registerPreviewCreateCommand( program );
-	registerPreviewListCommand( previewCommand );
-	registerPreviewDeleteCommand( previewCommand );
-	registerPreviewUpdateCommand( previewCommand );
+	registerPreviewCommands( studioArgv ).strict().help();
 
-	program.parse( process.argv );
+	await studioArgv.argv;
```
fredrikekelund:
```ts
		} )
		.demandCommand( 1, __( 'You need to specify a command' ) )
		.strict();

	registerPreviewCommands( studioArgv );
```
When I run the CLI script without any arguments, the help page isn't printed right now. Adding a `demandCommand` call fixes that.

cli/commands/preview/index.ts
fredrikekelund, line 1: Not a requirement, but I'd argue that keeping this logic in `cli/index.ts` makes it easier to understand the structure of all the commands provided.

cli/commands/preview/create.ts
```diff
@@ -60,14 +63,21 @@ async function runCommand( siteFolder: string, outputFormat?: OutputFormat ): Pr
...
-		.description(
-			__( 'Create a preview site from the specified folder (defaults to current directory)' )
-		)
-		.action( async ( siteFolder: string = process.cwd() ) => {
-			const options = program.opts();
-			await runCommand( siteFolder, options.outputFormat );
-		} );
+export const registerCommand = ( yargs: StudioArgv ) => {
+	return yargs.command( {
+		command: 'go [folder]',
```
```suggestion
		command: 'go [folder=cwd]',
```
While we're at it, I think this would add clarity for users.

cli/commands/preview/list.ts
```diff
@@ -44,14 +47,19 @@ async function runCommand( siteFolder: string, outputFormat?: OutputFormat ): Pr
...
-		.description(
-			__( 'List preview sites for the specified folder (defaults to current directory)' )
-		)
-		.action( async ( siteFolder: string = process.cwd() ) => {
-			const options = program.opts();
-			await runCommand( siteFolder, options.outputFormat );
-		} );
+export const registerCommand = ( yargs: StudioArgv ) => {
+	return yargs.command( {
+		command: 'list [folder]',
```
fredrikekelund, line 48:
```suggestion
		command: 'list [folder=cwd]',
```

cli/commands/preview/update.ts
```diff
@@ -73,16 +74,26 @@ async function runCommand(
...
-		)
-		.requiredOption( '-h, --host <host>', __( 'Host of the preview site to update' ) )
-		.action( async ( siteFolder: string = process.cwd(), options ) => {
-			const parentOptions = program.opts();
-			const normalizedHost = normalizeHostname( options.host );
-			await runCommand( siteFolder, normalizedHost, parentOptions.outputFormat );
-		} );
+export const registerCommand = ( yargs: Argv< GlobalOptions > ) => {
+	return yargs.command( {
+		command: 'update [folder]',
```
fredrikekelund, line 75:
```suggestion
		command: 'update [folder=cwd]',
```

cli/commands/preview/update.ts
```diff
@@ -73,16 +74,26 @@ async function runCommand(
...
+		describe: __( 'Update preview site for the specified folder (defaults to current directory)' ),
+		builder: ( yargs: Argv< GlobalOptions > ) => {
+			return yargs
+				.positional( 'folder', {
+					type: 'string',
+					default: process.cwd(),
+					description: __( 'The folder to update the preview from' ),
+				} )
+				.option( 'host', {
+					type: 'string',
```
fredrikekelund, line 86:
```suggestion
					alias: 'h',
					type: 'string',
```
I'm not sure what I think the best approach is here, but I wanted to raise the question. `-h` is what we had before, but I guess you could argue that it could be confused with `--help`. WDYT, @bcotrim?

fredrikekelund (commented):
Core functionality works great 👍 I left a couple of small suggestion and one slightly bigger comment about printing a help page by default when no command is provided.

cli/commands/preview/update.ts
bcotrim: I added the alias for `H`.
I think it makes it less likely to be confused with `--help`
What do you think?

cli/commands/preview/create.ts
bcotrim: For some reason this breaks the command. (it always uses the cwd value)

cli/index.ts
```diff
@@ -1,47 +1,57 @@
...
-			new Option( '--output-format [format]', __( 'Specify a non-standard output format' ) )
-				.argParser( ( value: string ) => {
-					if ( value !== 'json' ) {
-						throw new Error( __( 'The only custom output format supported is "json"' ) );
-					}
-					return value;
-				} )
-				.hideHelp()
-		);
+		.middleware( async () =>
```
fredrikekelund:
```suggestion
		.middleware( () =>
```
Subjective nit, but since we aren't using `await` inside the function, I think it's clearer which promise is relevant if we remove the `async` keyword. Alternatively, we could add an `await` in the function body.

cli/commands/preview/update.ts
fredrikekelund: Good call 👍

cli/commands/preview/delete.ts
```diff
@@ -39,13 +40,20 @@ async function runCommand( host: string, outputFormat?: OutputFormat ): Promise<
...
-		.command( 'delete <host>' )
-		.description( __( 'Delete a preview site' ) )
-		.action( async ( host: string ) => {
-			const outputFormat = rootCommand.opts().outputFormat;
-			const normalizedHost = normalizeHostname( host );
-			await runCommand( normalizedHost, outputFormat );
-		} );
+export const registerCommand = ( yargs: Argv< GlobalOptions > ) => {
+	return yargs.command( {
+		command: 'delete <host>',
```
fredrikekelund, line 45:
```suggestion
		command: 'preview delete <host>',
```
I think it would be better if all commands were printed on the top-level help page. Having to explicitly explore the preview command to understand which commands are available will be cumbersome, especially for the first release (when there are so few commands in total).

This change would also require some changes in `src/index.ts`.

fredrikekelund (commented):
Changes are looking good 👍 I added a few more comments, with one substantial potential change that I thought of just now. Let's also add yargs (and the types package) as a top-level dependency in `package.json`.

cli/index.ts
```diff
@@ -1,47 +1,57 @@
...
+					throw new Error( __( 'The only custom output format supported is "json"' ) );
+				}
+				return value;
+			},
+		} );
 
-	const previewCommand = studioCommand
-		.command( 'preview' )
-		.description( __( 'Manage preview sites' ) );
+	registerCreateCommand( studioArgv );
```
fredrikekelund: IMO, we might as well address STU-379 in this PR, too, and rename this command to `preview create`.

cli/commands/preview/delete.ts
bcotrim: I couldn't make it work with "2 names commands". From my testing all the commands would register as "preview", so only the last one to be registered would be available.
I still moved to the logic to the main index, grouped under the `preview` command.
I explored options for displaying sub-commands in the main help output but couldn't find a clean way to do this without manually writing the help text. 
I think that's worse than having the different helps for the sub-commands. 
What do you think?

fredrikekelund: Ugh, pulling my hair over this issue 😭

I couldn't figure out a solution either, so let's stick with this approach for now.

package.json
fredrikekelund, line 1: Still missing the `yargs` dependency here 🙂

cli/index.ts
```diff
@@ -1,47 +1,45 @@
...
-
-	const previewCommand = studioCommand
-		.command( 'preview' )
-		.description( __( 'Manage preview sites' ) );
-
-	registerPreviewCreateCommand( studioCommand );
-	registerPreviewListCommand( previewCommand, studioCommand );
-	registerPreviewDeleteCommand( previewCommand, studioCommand );
-	registerPreviewUpdateCommand( previewCommand, studioCommand );
+		.command( 'preview', 'Preview commands', ( previewYargs ) => {
```
fredrikekelund:
```suggestion
		.command( 'preview', __( 'Manage preview sites' ), ( previewYargs ) => {
```

cli/index.ts
```diff
@@ -1,47 +1,45 @@
...
+			registerCreateCommand( previewYargs );
+			registerListCommand( previewYargs );
+			registerDeleteCommand( previewYargs );
+			registerUpdateCommand( previewYargs );
+		} );
```
fredrikekelund: I also tried hard getting these commands to display in the default help output, but couldn't figure out a solution…

fredrikekelund (approved):
Looking good and works as expected 👍

cli/index.ts
```diff
@@ -1,47 +1,45 @@
...
-
-	registerPreviewCreateCommand( studioCommand );
-	registerPreviewListCommand( previewCommand, studioCommand );
-	registerPreviewDeleteCommand( previewCommand, studioCommand );
-	registerPreviewUpdateCommand( previewCommand, studioCommand );
+		.command( 'preview', 'Preview commands', ( previewYargs ) => {
+			registerCreateCommand( previewYargs );
+			registerListCommand( previewYargs );
+			registerDeleteCommand( previewYargs );
+			registerUpdateCommand( previewYargs );
```
fredrikekelund, line 38:
```suggestion
			registerUpdateCommand( previewYargs );

			previewYargs.demandCommand( 1, __( 'You must provide a valid command' ) );
```
Forgot to add this suggestion. Without this, the help page doesn't display when passing just the `preview` command

cli/index.ts
fredrikekelund:
```suggestion
		} )
		.demandCommand( 1, __( 'You must provide a valid command' ) )
		.strict();
```
`demandCommand` needed to display the help page by default and `strict` seems like good practice.

package.json
bcotrim: We already had `yargs` in there: https://github.com/Automattic/studio/blob/trunk/package.json#L159

PR #1235, opened 2025-04-21
Studio: Polish User Settings Modal design

gavande1:
## Related issues

- Fixes STU-374

## Proposed Changes

- Update gaps between label and select control
- Update select components to use __next40pxDefaultSize to match with rest of the designs
- Update string case of "Code Editor" -> "Code editor". 

## Testing Instructions

- Check out this branch
- Compare the rendered layout with the Figma designs added below. 

| Trunk | Branch | Figma |
|--------|--------|--------|
| CleanShot 2025-04-21 at 15 24 46@2x | CleanShot 2025-04-21 at 15 23 22@2x | CleanShot 2025-04-21 at 15 27 40@2x |

## Pre-merge Checklist

### DIFF
--- a/src/modules/user-settings/components/editor-picker.tsx
+++ b/src/modules/user-settings/components/editor-picker.tsx
@@ -42,7 +42,7 @@ export const EditorPicker = ( { value, onChange }: EditorPickerProps ) => {
 
 	return (
 		<div className="flex gap-1.5 flex-col">
-			<h2 className="a8c-subtitle-small">{ __( 'Code editor' ) }</h2>
+			<label className="font-semibold">{ __( 'Code editor' ) }</label>
 			<SelectControl
 				value={ value }
 				onChange={ ( newValue ) => onChange( newValue as SupportedEditor ) }
--- a/src/modules/user-settings/components/editor-picker.tsx
+++ b/src/modules/user-settings/components/editor-picker.tsx
@@ -49,13 +49,11 @@ export const EditorPicker = ( { value, onChange }: EditorPickerProps ) => {
 				__nextHasNoMarginBottom
 				__next40pxDefaultSize
 			>
-				<optgroup label={ __( 'Installed' ) }>
-					{ installedEditors.map( ( [ editor, label ] ) => (
-						<option key={ editor } value={ editor }>
-							{ label }
-						</option>
-					) ) }
-				</optgroup>
+				{ installedEditors.map( ( [ editor, label ] ) => (
+					<option key={ editor } value={ editor }>
+						{ label }
+					</option>
+				) ) }
 				<optgroup label={ __( 'Not installed' ) }>
 					{ uninstalledEditors.map( ( [ editor, label ] ) => (
 						<option key={ editor } value={ editor } disabled>
--- a/src/modules/user-settings/components/editor-picker.tsx
+++ b/src/modules/user-settings/components/editor-picker.tsx
@@ -3,6 +3,7 @@ import { useI18n } from '@wordpress/react-i18n';
 import { useEffect, useState } from 'react';
 import { SupportedEditor, supportedEditorNames } from 'src/lib/editor';
 import { getIpcApi } from 'src/lib/get-ipc-api';
+import { SettingsFormField } from './settings-form-field';
 
 interface EditorPickerProps {
 	value: SupportedEditor;
@@ -41,8 +42,7 @@ export const EditorPicker = ( { value, onChange }: EditorPickerProps ) => {
 	);
 
 	return (
-		<div className="flex gap-1.5 flex-col">
-			<label className="font-semibold">{ __( 'Code editor' ) }</label>
+		<SettingsFormField label={ __( 'Code editor' ) }>
 			<SelectControl
 				value={ value }
 				onChange={ ( newValue ) => onChange( newValue as SupportedEditor ) }
@@ -64,6 +64,6 @@ export const EditorPicker = ( { value, onChange }: EditorPickerProps ) => {
 					) ) }
 				</optgroup>
 			</SelectControl>
-		</div>
+		</SettingsFormField>
 	);
 };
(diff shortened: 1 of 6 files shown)


epeicher (approved):
This looks much better now! I can see the changes as described

|Trunk | This branch|
|-------|------|
| CleanShot 2025-04-21 at 19 15 23@2x|  CleanShot 2025-04-21 at 19 14 12@2x|

I have one question about the capitalization, why are you not using the capitalized version as suggested in the linked task and in Figma? This is for the labels `Language`, `Editor` and `Shell`

gavande1:
Thanks, @epeicher, for the review and your observation about the label casing. I intentionally didn’t update the label case because the design for this screen differs from the others. For example, the settings screen (in Figma) uses uppercase labels, whereas screens like "Add a site" use sentence case. That inconsistency is why I left the label case unchanged here. I have also raised a similar concerns in Figma.
CleanShot 2025-04-22 at 10 00 26@2x

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -41,13 +41,13 @@ export const EditorPicker = ( { value, onChange }: EditorPickerProps ) => {
 	);
 
 	return (
-		<div className="flex gap-5 flex-col">
-			<h2 className="a8c-subtitle-small">{ __( 'Code Editor' ) }</h2>
+		<div className="flex gap-1.5 flex-col">
+			<h2 className="a8c-subtitle-small">{ __( 'Code editor' ) }</h2>
```
katinthehatsite: if we are going for consistency with other modals, I am wondering if we might want to set the size of the text to 16px on these labels to be consistent with, for example, _Add site_ modal

katinthehatsite (commented):
One thing that I noticed is that there is a bit of weird overlay when the scrollbar is present (see on the screenshot how there is a line over Language, Performance etc):

Screenshot 2025-04-22 at 9 35 20 AM

One thing that I am wondering is if we might be able to somehow avoid scrollbar at all. One approach that comes to mind if having `Advanced settings` section that can be then expanded (this is how it now works with `Add site`). However, the downside of this is that it will hide the settings one step further from the user.

cc @matt-west what do you think?

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -41,13 +41,13 @@ export const EditorPicker = ( { value, onChange }: EditorPickerProps ) => {
...
 			<SelectControl
 				value={ value }
 				onChange={ ( newValue ) => onChange( newValue as SupportedEditor ) }
 				__nextHasNoMarginBottom
-				className="mb-2"
+				__next40pxDefaultSize
 			>
 				<optgroup label={ __( 'Installed' ) }>
```
katinthehatsite: I am wondering if we should remove the label for this section - although this is not a strong preference.

Current:

Screenshot 2025-04-22 at 9 46 05 AM

Figma: 

Screenshot 2025-04-22 at 9 46 15 AM

src/modules/user-settings/components/language-picker.tsx
```diff
@@ -10,7 +10,7 @@ interface LanguagePickerProps {
 export const LanguagePicker = ( { value, onChange }: LanguagePickerProps ) => {
 	const { __ } = useI18n();
 	return (
-		<div className="flex gap-5 flex-col">
+		<div className="flex gap-1.5 flex-col">
 			<h2 className="a8c-subtitle-small">{ __( 'Language' ) }</h2>
```
katinthehatsite: I see that we are repeating this styling for LanguagePicker, EditorPicker and Terminal Picker. To improve maintainability of this in the future, we could possibly extract it into a component that can be reused for all these pickers e.g. something like:

```
   export const FormField = ({ label, children }) => (
     <div className="flex gap-1.5 flex-col">
       <h2 className="a8c-subtitle-small">{label}</h2>
       {children}
     </div>
   );
 ```

And then use it with each picker. Not a big deal though, I am mostly thinking of the case when someone updates some styling for one field but forgets to do it for other pickers. This could also help reduce code duplication. Feel free to skip though, if you think this might be an overkill.

matt-west:
> One thing that I noticed is that there is a bit of weird overlay when the scrollbar is present

Is this caused by a set height on the modal content? The modal should expand vertically so that we only need to show scroll-bars if it's a tall as the Studio window.

src/modules/user-settings/components/editor-picker.tsx
gavande1: Thanks for the suggestion. Addressed in c34e1cb9f9f13c7830af2c37e3d2890c70ddefaa!

src/modules/user-settings/components/editor-picker.tsx
gavande1: Fixed in 5de00a3d51d89821ddbc0d1ffe7e96f34d1afea5

src/modules/user-settings/components/language-picker.tsx
gavande1: Replace with common component in 347acf3bf69e54d80462424d430dc9a8ae811ec2!

gcsecsey (commented):
> Is this caused by a set height on the modal content? The modal should expand vertically so that we only need to show scroll-bars if it's a tall as the Studio window.

The modal already expands vertically until it reaches the `[height of the viewport] - 120px` height. I tested this by duplicating the inner content:
CleanShot 2025-04-22 at 13 35 11@2x

I think we have two issues here:
- The title div is rendered over the scrollable content when scrolling
- The tab bar is part of the modal content, so it also scrolls away. It should be fixed on top with the modal title.

gavande1:
I have removed the additional margin fix that was added previously. The overlay issue should be fixed now.

gcsecsey (approved):
Thanks for following up @gavande1! 🙌 

The rendering of the modal title looks good to me now. The tab bar still scrolls under the title when the modal content is scrolled. But as the modal is not scrollable with the minimum window height, I don't think this is a blocker anymore.

image

PR #1236, opened 2025-04-22
Studio: Check if port is open before starting servers

gavande1:
## Related issues

- Fixes STU-103

## Proposed Changes

- Added port availability checking before starting servers
- Improved error handling for busy ports

## Testing Instructions

- Check out this branch and run `npm start`
- Start the server for your site and confirm that the server starts normally. 
- Note the port number. e.g. 8881 and stop the server. 
- Open a new terminal and run `php -S localhost:{PORT_NUMBER}`. e.g. `php -S localhost:8881`
- Come back to the studio and try to start the server. It should show the following alert:

CleanShot 2025-04-22 at 10 24 46@2x

- Confirm that the site running status returns to normal and shows the `Start` button. 
- Go back to the terminal where you started `php -S localhost:8881`, press cmd + c to terminate the server. This would free up port 8881. 
- Come back to the studio and start the server.
- Confirm the site server started without any problem, and you can browse your test site. 

### Bonus Test
- Open your `appdata-v1.json` (cat "~/Library/Application Support/Studio/appdata-v1.json")
- Look at the `sites` key, note down the highest occupied port number. For example, in my case it's `8883`. 
- Add 1 to that port number. So becomes `8884`
- Open a new terminal session and run `php -S localhost:8884`
- Come back to the studio and add a new site. 
- Check the port number for the new site; it should be 8885. When creating the site, port 8884 was busy, so Studio skipped it and chose the next available port. If 8885 is also busy, the port finder will increment by 1 to find the next available port. 

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -302,7 +302,7 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 						message: __(
 							`The site server failed to start because the port is already in use. Please close any local development apps that may be using port ${ port } and try again.`
 						),
-						showOpenLogs: true,
+						showOpenLogs: false,
 					} );
 				} else {
 					getIpcApi().showErrorMessageBox( {
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -295,6 +295,15 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 						),
 						showOpenLogs: true,
 					} );
+				} else if ( error instanceof Error && error.message.includes( 'ERROR_PORT_IN_USE' ) ) {
+					const port = error.message.match( /\d+/ );
+					getIpcApi().showErrorMessageBox( {
+						title: __( 'Failed to start the site server' ),
+						message: __(
+							`The site server failed to start because the port is already in use. Please close any local development apps that may be using port ${ port } and try again.`
+						),
+						showOpenLogs: true,
+					} );
 				} else {
 					getIpcApi().showErrorMessageBox( {
 						title: __( 'Failed to start the site server' ),
--- a/src/lib/port-finder.ts
+++ b/src/lib/port-finder.ts
@@ -1,4 +1,5 @@
 import http from 'http';
+import net from 'net';
 import killPort from 'cross-port-killer';
 
 const DEFAULT_PORT = 8881;
@@ -26,16 +27,28 @@ class PortFinder {
 
 	#isPortFree(): Promise< boolean > {
 		return new Promise( ( resolve ) => {
-			const server = http.createServer();
+			// First try to connect to the port
+			const socket = new net.Socket();
+			socket.on( 'error', () => {
+				// If we can't connect, try to bind to the port
+				const server = http.createServer();
+				server
+					.listen( this.#searchPort, () => {
+						server.close();
+						setTimeout( () => {
+							resolve( true );
+						}, 50 ); // Add a small delay to ensure port is released
+					} )
+					.on( 'error', () => {
+						resolve( false );
+					} );
+			} );
 
-			server
-				.listen( this.#searchPort, () => {
-					server.close();
-					resolve( true );
-				} )
-				.on( 'error', () => {
-					resolve( false );
-				} );
+			// Try to connect to the port
+			socket.connect( this.#searchPort, 'localhost', () => {
+				socket.destroy();
+				resolve( false );
+			} );
 		} );
 	}
 
@@ -92,6 +105,20 @@ class PortFinder {
 					this.#openPort = DEFAULT_PORT;
 				} );
 		}
+	}
+
+	/**
+	 * Checks if a specific port is available.
+	 *
+	 * @param {number} port - The port number to check.
+	 * @returns {Promise<boolean>} A promise that resolves to true if the port is available, false otherwise.
+	 **/
+	public async isPortAvailable( port: number ): Promise< boolean > {
+		const localPort = this.#searchPort;
+		this.#searchPort = port;
+		const result = await this.#isPortFree();
+		this.#searchPort = localPort;
+		return result;
 	}
 }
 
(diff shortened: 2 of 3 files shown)


src/lib/port-finder.ts
```diff
@@ -26,16 +27,28 @@ class PortFinder {
 
 	#isPortFree(): Promise< boolean > {
 		return new Promise( ( resolve ) => {
-			const server = http.createServer();
+			// First try to connect to the port
+			const socket = new net.Socket();
```
gavande1, line 31: The previous `isPortFree()` implementation used `http.createServer().listen()` to check port availability. However, this was leading to false positives. The port was showing free when it was actually in use. This might be happening because listen() may not immediately fail if the port is in a TIME_WAIT state or due to OS-level race conditions.

The updated logic first attempts to connect to the port using a socket to improve reliability. A successful connection confirms the port is in use. If the connection fails, we try to bind to the port to to make sure it's truly available.

src/lib/port-finder.ts
```diff
@@ -93,6 +106,20 @@ class PortFinder {
...
+	 * Checks if a specific port is available.
+	 *
+	 * @param {number} port - The port number to check.
+	 * @returns {Promise<boolean>} A promise that resolves to true if the port is available, false otherwise.
+	 **/
+	public async isPortAvailable( port: number ): Promise< boolean > {
+		const localPort = this.#searchPort;
+		this.#searchPort = port;
+		const result = await this.#isPortFree();
+		this.#searchPort = localPort;
```
epeicher: This pattern of saving/restoring the global variable `this.#searchPort` seems a bit fiddly to me. What do you think? I would suggest updating the `this.#isPortFree()` to accept a port as a parameter, maybe using `this.#searchPort` as a default, or even better using either `this.#searchPort` as a parameter in the other calls and the `port` in this call.

epeicher (approved):
I have tested it, and it works as advertised. I love this change as it improves the UX when there are some errors, and those errors of port in use are very common so this is 💯 . I have left a non-blocking comment about the style of a method but overall this LGTM!

|Port in use for existing site | Having the next port in use for new site|
|-------|------|
|  CleanShot 2025-04-22 at 09 18 40@2x |    CleanShot 2025-04-22 at 09 25 45@2x|

src/hooks/use-site-details.tsx
```diff
@@ -295,6 +295,15 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
 						showOpenLogs: true,
 					} );
+				} else if ( error instanceof Error && error.message.includes( 'ERROR_PORT_IN_USE' ) ) {
+					const port = error.message.match( /\d+/ );
+					getIpcApi().showErrorMessageBox( {
+						title: __( 'Failed to start the site server' ),
+						message: __(
+							`The site server failed to start because the port is already in use. Please close any local development apps that may be using port ${ port } and try again.`
+						),
+						showOpenLogs: true,
```
wojtekn: Will logs be helpful in this case?

gavande1: Not really. I was skeptical about removing it. I will go ahead and remove it.

gavande1: Addressed in [a22a794](https://github.com/Automattic/studio/pull/1236/commits/a22a794e34cae0e9579bcd12e72931b3bcaf772b)

src/lib/port-finder.ts
gavande1: Thanks @epeicher for the suggestion. I was a bit sceptical about making changes to the #isPortFree function. But your suggestion makes sense to me. I have updated that in [211b83b](https://github.com/Automattic/studio/pull/1236/commits/211b83b8dc2e2af4355f8ad6c450b862419b4181)

src/hooks/use-site-details.tsx
```diff
@@ -295,6 +295,15 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 						),
 						showOpenLogs: true,
 					} );
+				} else if ( error instanceof Error && error.message.includes( 'ERROR_PORT_IN_USE' ) ) {
+					const port = error.message.match( /\d+/ );
+					getIpcApi().showErrorMessageBox( {
+						title: __( 'Failed to start the site server' ),
+						message: __(
+							`The site server failed to start because the port is already in use. Please close any local development apps that may be using port ${ port } and try again.`
```
katinthehatsite, line 303: One idea that comes to mind as a potential improvement even further could be that instead of the `Ok` button as shown now, we could add a button that says something like `Try another port` and it could automatically attempt to start the server on a different port.

katinthehatsite:
The testing steps worked for me and the implementation works as expected 👍 I added one suggestion that we can consider to potentially improve the user experience further.

src/hooks/use-site-details.tsx
gcsecsey: This pattern is also used by some build tools, without prompting the user. Eg. when using Vite, if the default port is occupied, it just increments the dev port until it finds a free port.
image

gcsecsey (approved):
This works well for me too.

image

src/hooks/use-site-details.tsx
katinthehatsite: Sounds good 👍

gavande1: Thanks for the suggestion, @katinthehatsite. This was one of the suggestions in the linked issue. However, it would require a larger effort than the current PR. I believe this PR improves the experience significantly.

fredrikekelund:
We also have `checkPortInWindows` in `src/lib/proxy-server.ts`. There's an obvious difference in that the `checkPortInWindows` function only runs on Windows, but I still wonder if we shouldn't refactor that function and `PortFinder::isPortFree` to share the same underlying logic. WDYT, @gavande1?

gavande1:
Thanks @fredrikekelund for bringing it to my attention. I don't see any reason not to refactor `src/lib/proxy-server.ts` to share similar logic. I will create a follow-up PR for that.

gavande1:
For posterity, I have created a PR https://github.com/Automattic/studio/pull/1281 to address the above suggestion.

PR #339, opened 2024-07-04
Add feature flag to control integration of `wp-cli` commands in terminal

fluiddot:
Related to 8097-gh-Automattic/dotcom-forge.

## Proposed Changes

- Add feature flag `terminalWpCliEnabled` to control when the `wp-cli` integration should be enabled in the terminal.
- Expand IPC handler `openTerminalAtPath` to enable/disable `wp-cli` integration.
- Add test case to cover if `wp-cli` integration is enabled via feature flag.

## Testing Instructions

> [!NOTE]
> The following testing instructions are meant to be followed in macOS. For other platforms, the paths and commands might differ.

### `wp-cli` integration is disabled (Default behavior)
- Run the app with the command `npm start`.
- Click on the Terminal button.
- Within the terminal, run the command `which wp`.
- If the `wp` command was installed globally, observe that the command returns its location (e.g. `/usr/local/bin/wp`)
- If the `wp` command is not installed, observe that running `wp` results in a failure.

### `wp-cli` integration is enabled via feature flag
- Run the app with the command `STUDIO_TERMINAL_WP_CLI=true npm start`.
- Click on the Terminal button.
- Within the terminal, run the command `which wp`.
- Observe that the command returns the location `<PROJECT_PATH>/local-environment/bin/wp`.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #1240, opened 2025-04-22
Studio: Add more terminal apps to Studio

katinthehatsite:
## Related issues

Related to STU-364

## Proposed Changes

This PR adds `Ghostty` and `Warp` to the selection of available terminal apps. `Ghostty` is available on MacOs (and Linux) and `Warp` should be available for both Windows and MacOs. 

## Testing Instructions

* Open `User settings` by clicking your profile picture
* Click on `Preferences` tab
* In the `Terminal` section, observe that you should see  `Ghostty` and `Warp` (if you have them installed, you should be able to select them. If not, they should appear as inactive)
* Install both of these apps
* Restart Studio
* Open `User settings` by clicking your profile picture
* Click on `Preferences` tab
* Select one of the newly installed apps
* Navigate to the `Overview` tab
* Click on the terminal app you just selected
* Confirm the site open's in it
* Repeat the same steps with the second app

**Note**: I could not test Warp with Windows so it would be nice if we could run some testing first before merging the PR

## Pre-merge Checklist


katinthehatsite:
I used this list as a guide for the list of missing terminal apps: https://docs.google.com/spreadsheets/d/1a49RBWqIay-ynbFZQgCyO0-Dp3FH1jvAFt1KansawVE/edit?gid=812510103#gid=812510103
Not sure if we want to start with the smaller list first and then expand or if we want to add anything else.

src/ipc-handlers.ts
```diff
@@ -990,7 +990,11 @@ export async function openTerminalAtPath(
 		const userData = await loadUserData();
 		const preferredTerminal = userData.supportedTerminal || 'terminal';
 
-		if ( preferredTerminal === 'iterm' ) {
+		if ( preferredTerminal === ( 'warp' as SupportedTerminal ) ) {
+			return promiseExec( `open -a Warp "${ targetPath }"` );
```
katinthehatsite, line 994: I was not able to use the same AppleScripting approach as with, for example, iTerm - it seems that it is not supported yet https://github.com/warpdotdev/Warp/issues/3364 . I went for a simpler approach but not sure what is best here. I will leave it as is for further discussion

katinthehatsite: I checked a bit further and it seems that we are using `wpCliEnabled` flag for now to hide WP-CLI integration for now and it is not available for users (as per this PR: https://github.com/Automattic/studio/pull/339 ) so I think it might be fine for these two terminal apps

epeicher: I actually prefer this approach as it is much simpler 😀. Looking at the code, that is used for exporting those two environment variables `PATH` and `STUDIO_APP_PATH` so maybe there is an option to export those variables as initial steps? Anyway, this can be considered as a follow-up task as it only works when `wpCliEnabled` as you mention.

epeicher: Additionally, when the CLI flag is enabled, the terminal opens with that big PATH executed, and it looks a bit weird for the user (I have blurred the image, but it can be observed 😛 )
CleanShot 2025-04-23 at 15 59 01@2x

epeicher (approved):
I have tested with both terminals installed, and it works as expected. LGTM!

|Terminals | Ghostty selected| Open Ghostty |Warp selected| Open Warp|
|-------|------|------|------|------|
|  CleanShot 2025-04-23 at 15 31 38@2x|CleanShot 2025-04-23 at 15 33 13@2x |CleanShot 2025-04-23 at 15 33 55@2x |  CleanShot 2025-04-23 at 15 34 12@2x |    CleanShot 2025-04-23 at 15 34 26@2x|

src/lib/is-installed.ts
```diff
@@ -65,6 +69,8 @@ const installationPaths: Record< string, PlatformPaths > = {
 		],
 		iterm: [],
 		terminal: [],
+		warp: [ path.win32.join( app.getPath( 'appData' ), 'Warp' ) ],
```
gcsecsey: On Windows Warp gets installed to `%USERPROFILE%\AppData\Local\Warp` so we'll have to use that path for it to be correctly shown. On https://github.com/Automattic/studio/pull/1255 I've made some changes to this path, so we'll just need to use `getLocalProgramsPath()`.

CleanShot 2025-04-23 at 15 55 36@2x

gcsecsey (approved):
Changes LGTM and work great! I could open Ghostty, Warp, iTerm, and Terminal successfully using the button on the overview tab.

https://github.com/user-attachments/assets/ea3ed9b2-a697-4e5f-b870-6c82c6336ea0

src/lib/is-installed.ts
katinthehatsite: Sounds good, I adjusted this in https://github.com/Automattic/studio/pull/1240/commits/d48555f35ff7453ae152d82f26de5b8d2b83dd8e

katinthehatsite:
> I actually prefer this approach as it is much simpler 😀. Looking at the code, that is used for exporting those two environment variables PATH and STUDIO_APP_PATH so maybe there is an option to export those variables as initial steps? Anyway, this can be considered as a follow-up task as it only works when wpCliEnabled as you mention.

Makes sense, I also noticed that currently we disabled `wpCliEnabled`  because [php-wasm doesn't support all the commands.](https://github.com/Automattic/dotcom-forge/issues/8097) In that case, I think it makes sense to focus on simplicity and I can perhaps create a follow up task to check on this for when the `wpCliEnabled` becomes enabled again.

katinthehatsite:
Added follow up issue here: STU-397

src/components/content-tab-overview.tsx
```diff
@@ -146,7 +147,8 @@ function ShortcutsSection( { selectedSite }: Pick< ContentTabOverviewProps, 'sel
 
 	const updateTerminalName = useCallback( async () => {
 		const terminal = await getIpcApi().getUserTerminal();
-		setTerminalName( terminal === 'iterm' ? __( 'iTerm' ) : __( 'Terminal' ) );
+
+		setTerminalName( __( supportedTerminalNames[ terminal ] ) );
```
fredrikekelund, line 151: We can only pass static strings to `__()`, so the `__()` calls should happen in `src/lib/terminal.ts`.

src/lib/terminal.ts
fredrikekelund, line 1: I don't have a good overview of the files being touched in this project, but would it perhaps make sense to group them in a `src/modules/SOMETHING` folder..?

src/components/content-tab-overview.tsx
katinthehatsite: Nice catch, I made some adjustments for this in https://github.com/Automattic/studio/pull/1265

src/lib/terminal.ts
katinthehatsite: We have `src/modules/user-settings` for this but I think we move only frontend components there and not these functions. We can move it there I guess, I do not have a strong preference. I can add a follow-up PR to adjust the structure

katinthehatsite: Added PR here: https://github.com/Automattic/studio/pull/1266

PR #1243, opened 2025-04-22
Studio Logs: Could not open log file on Studio when error occurred on Windows

gavande1:
## Related issues

- Fixes STU-102

## Proposed Changes

- Do not try to create symlink for logger file on windows

## Testing Instructions

- TBA

## Pre-merge Checklist

### DIFF
--- a/src/logging.ts
+++ b/src/logging.ts
@@ -1,6 +1,7 @@
 import { app } from 'electron';
 import path from 'path';
 import * as FileStreamRotator from 'file-stream-rotator';
+import { platform } from 'os';
 
 let logStream: ReturnType< typeof FileStreamRotator.getStream > | null = null;
 let currentLogFile = '';
@@ -25,6 +26,8 @@ export function setupLogging( {
 		app.setAppLogsPath();
 	}
 
+	const isWindows = platform() === 'win32';
+	
 	logStream = FileStreamRotator.getStream( {
 		filename: path.join( logDir, 'studio-%DATE%' ),
 		date_format: 'YYYYMMDD',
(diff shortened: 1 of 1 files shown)


src/logging.ts
```diff
@@ -25,6 +26,8 @@ export function setupLogging( {
 		app.setAppLogsPath();
 	}
 
+	const isWindows = platform() === 'win32';
```
sejas, line 29: We have a function [`isWindows`](https://github.com/Automattic/studio/blob/93adc785c3485fd0c6f5fca7c946be4ecddf2792/src/lib/app-globals.ts#L12-L14) but it seems it only works in the front-end. So It's ok to make the comparison here.

src/logging.ts
```diff
@@ -33,12 +36,18 @@ export function setupLogging( {
...
+		create_symlink: ! isWindows, // Only create symlinks on non-Windows platforms
 		audit_hash_type: 'sha256',
 		verbose: true, // file-stream-rotator itself will log to console too
 	} );
 
-	currentLogFile = path.join( logDir, 'current.log' );
+	// On Windows, we'll use the most recent log file as the current log file
+	if ( isWindows ) {
+		const today = new Date().toISOString().split( 'T' )[ 0 ].replace( /-/g, '' );
+		currentLogFile = path.join( logDir, `studio-${ today }.0.log` );
```
sejas, line 47: I know our current log file size is 5MB, and the chances of reaching that limit are very low, but in theory, it's possible, and in those cases, the `.0.log` will fail. Building the date from scratch seems a bit flaky. What do you think about setting the current log file dynamically?

sejas (commented):
Thank you for creating this PR. I tested it, and it works as expected on Windows. My only concern is hardcoding the filename to `.0.log`.
I went ahead and created a PR saving the log file returned by `FileStreamRotator`, so it will work even if it rotates the file multiple times. Let me know what you think.

- https://github.com/Automattic/studio/pull/1245

gavande1:
Thanks, @sejas, for your review. Your PR https://github.com/Automattic/studio/pull/1245 is undoubtedly a better approach in terms of cross-platform parity.

gavande1:
I am closing this PR in favour of https://github.com/Automattic/studio/pull/1245.

PR #1244, opened 2025-04-22
Preview sites: watch for changes on appdata file

bcotrim:
## Related issues

- Fixes STU-383

## Proposed Changes

- Add file watcher for the app data file to detect external changes (CLI commands, manual edits)
- Create new user data slice in Redux to handle app data state
- Refactor snapshot storage to use the user data slice instead of a separate slice
- Add IPC events for user data updates and errors
- Move snapshots from state.snapshot.snapshots to state.userData.snapshots

## Testing Instructions

- Run `npm start`
- Create a Studio site
- Navigate to the preview tabs
- In another terminal run `npm run cli:build`
- Run `node dist/cli/main.js go <STUDIO SITE FOLDER>` and wait for completion
- Without refreshing or doing anything else in Studio confirm that the new preview site appear in the list
- On Studio open user settings and check if the number of preview sites is correct
- Without closing user settings
- In another terminal create a studio preview site with the CLI again and wait for completion
- Validate that the number of preview sites in the Studio user settings got updated
- All preview sites operation in Studio work as expected.

## Pre-merge Checklist


src/lib/user-data-watcher.ts
```diff
@@ -0,0 +1,57 @@
...
+		const mainWindow = await getMainWindow();
+		if ( ! mainWindow.isDestroyed() && ! mainWindow.webContents.isDestroyed() ) {
+			mainWindow.webContents.send( 'user-data-updated', userData );
+		}
+	} catch ( error ) {
+		const errMsg = error instanceof Error ? error.message : 'Failed to load user data';
+		const mainWindow = await getMainWindow();
+		if ( ! mainWindow.isDestroyed() && ! mainWindow.webContents.isDestroyed() ) {
+			mainWindow.webContents.send( 'user-data-error', errMsg );
+		}
```
fredrikekelund:
```suggestion
		sendIpcEventToRendererWithWindow( mainWindow, 'user-data-updated', userData );
	} catch ( error ) {
		const errMsg = error instanceof Error ? error.message : 'Failed to load user data';
		const mainWindow = await getMainWindow();
		sendIpcEventToRendererWithWindow( mainWindow, 'user-data-error', errMsg );
```
We should use this utility function. It helps reduce repetition, but above all, it ensures type safety.

src/lib/user-data-watcher.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	getMainWindow().then( ( mainWindow ) => {
+		if ( mainWindow.webContents.isLoading() ) {
+			mainWindow.webContents.once( 'did-finish-load', () => {
+				loadAndUpdateUserData();
+			} );
+		} else {
+			loadAndUpdateUserData();
+		}
+	} );
```
fredrikekelund:
```suggestion
	const mainWindow = await getMainWindow();

	if ( mainWindow.webContents.isLoading() ) {
		mainWindow.webContents.once( 'did-finish-load', () => {
			loadAndUpdateUserData();
		} );
	} else {
		loadAndUpdateUserData();
	}
```
Nit, but I would make the function async and `await` the result of `getMainWindow`

src/lib/user-data-watcher.ts
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	const filePath = getUserDataFilePath();
+
+	const fsEventHandler = ( eventType: string ) => {
+		loadAndUpdateUserData();
+		if ( eventType === 'rename' && watcher ) {
+			watcher.close();
+			watcher = fs.watch( filePath, fsEventHandler );
+		}
```
fredrikekelund, line 21: Does `fs.watch` "follow" a file when it's renamed?

src/lib/user-data-watcher.ts
fredrikekelund, line 1: I feel like this approach unlocks a lot of potential simplifications around state handling 🙂

src/stores/user-data-slice.ts
fredrikekelund, line 1: Considering how much state is stored in the appdata file, and that we want to continue implementing separate Redux slices for the different parts of that state, I would advise that we implement multiple `user-data-updated` and `user-data-error` event handlers—one in each relevant Redux slice.

In concrete terms, I advise moving the `user-data-updated` and `user-data-error` event handlers to `src/stores/snapshot-slice.ts` and removing this Redux slice altogether.

fredrikekelund (commented):
This works beautifully and the diff is nice and simple 👍 My biggest concern is that I'm not convinced we should have a dedicated Redux slice for holding all state from the appdata file. Continuing with separate Redux slices that each listen to the relevant IPC events seems like a more scaleable approach.

src/stores/snapshot-slice.ts
```diff
@@ -82,13 +79,12 @@ const updateSnapshot = createAsyncThunk(
...
-		);
-		if ( ! snapshot ) {
+		const found = state.snapshot.snapshots.find( ( snap ) => snap.atomicSiteId === atomicSiteId );
+		if ( ! found ) {
 			throw new Error( 'Snapshot not found' );
 		}
-		const { operationId } = await getIpcApi().updateSnapshot( siteFolder, snapshot.url );
+		const snapshotUrl = found.url;
+		const { operationId } = await getIpcApi().updateSnapshot( siteFolder, snapshotUrl );
 		return { operationId };
```
fredrikekelund: Nit, but this looks to be entirely cosmetic, and we could probably revert this change.

src/modules/user-settings/components/user-settings.tsx
```diff
@@ -68,6 +72,12 @@ export default function UserSettings() {
 		}
 	}, [ __, dispatch, user?.id ] );
 
+	useEffect( () => {
+		if ( needsToOpenUserSettings ) {
+			refetchSnapshotUsage();
+		}
+	}, [ snapshotsByUser.length, needsToOpenUserSettings, refetchSnapshotUsage ] );
```
fredrikekelund, line 79: What does this change accomplish?

src/stores/snapshot-slice.ts
```diff
@@ -295,6 +293,11 @@ const selectSnapshotsBySiteAndUser = createSelector(
 		)
 );
 
+window.ipcListener.subscribe( 'user-data-updated', ( _, payload ) => {
+	console.log( 'user-data-updated', payload );
```
fredrikekelund:
```suggestion
```

src/tests/index.test.ts
```diff
@@ -12,11 +12,16 @@ jest.mock( 'src/main-window' );
...
+jest.mock( 'src/lib/user-data-watcher' );
 jest.mock( 'src/setup-wp-server-files', () => ( {
 	setupWPServerFiles: jest.fn( () => Promise.resolve() ),
 	updateWPServerFiles: jest.fn( () => Promise.resolve() ),
 } ) );
 
+type MockedFs = typeof fs & {
+	__setFileContents: ( path: string, contents: string ) => void;
+};
+
```
fredrikekelund:
```suggestion
```
We don't need to redefine this type – it's already available globally.

src/stores/snapshot-slice.ts
```diff
@@ -145,6 +141,12 @@ const snapshotSlice = createSlice( {
 				( snapshot ) => snapshot.atomicSiteId !== action.payload.atomicSiteId
 			);
 		},
+		setSnapshots: ( state, action: PayloadAction< Snapshot[] > ) => {
+			if ( ! fastDeepEqual( state.snapshots, action.payload ) ) {
+				state.snapshots = action.payload;
+				state.isLoaded = true;
```
fredrikekelund:
```suggestion
				state.isInitialSnapshotsLoaded = true;
```
Not directly related to this PR, but I'm thinking that `isLoaded` may be too generic a name. Might be worth clarifying.

fredrikekelund (approved):
I left a number of comments that we should address, but fundamentally, this is working well 👍 In the interest of moving things along, I'm approving now.

src/modules/user-settings/components/user-settings.tsx
bcotrim: Without this change, we would never fetch the snapshot usage again, since `UserSettings` are mounted in `app.tsx`.
The usage shown on the user settings gives preference to the value returned from the snapshotUsage API call.
I wanted to refetch that call whenever we open user settings, or the snapshots length changes.
Does it make sense or is my understanding wrong?

bcotrim: Since the other comments on this PR are smaller issues and this code shouldn't cause any major issues, I'll merge this PR and can follow-up in a new one if needed.

fredrikekelund: `src/stores/snapshot-slice.ts` [optimistically updates](https://github.com/Automattic/studio/blob/007426f6d69c8199aedb07f85c84d2d29f221ead/src/stores/snapshot-slice.ts#L460-L464) the query data when snapshots are created or deleted and then [invalidates the query](https://github.com/Automattic/studio/blob/007426f6d69c8199aedb07f85c84d2d29f221ead/src/stores/snapshot-slice.ts#L479-L482) after a delay (to allow the changes to take effect on the server). The component shouldn't have to concern itself with updating the query data. 

I'll put together a PR to fix this.

PR #1242, opened 2025-04-22
Add sequence numbers to preview site names

fredrikekelund:
## Related issues

- Fixes STU-381

## Proposed Changes

- Previously, preview site names would be `{{ siteName }} Preview {{ sequenceNumber }}` where `sequenceNumber` was based on the highest sequence number for the existing preview sites connected to the current site. However, this behavior was never implemented in the CLI. This PR fixes that. 
- I also took the opportunity to rename functions in `cli/lib/snapshots.ts`, ensuring we follow the rule of using `snapshot` as the code term and `preview sites` as the user-facing term. 

## Testing Instructions

1. Start Studio with `npm start`.
2. Create a new preview site.
3. Ensure that the name is `{{ siteName }} Preview {{ sequenceNumber }}` where `sequenceNumber` is based either on the highest pre-existing sequence number or the existing number of preview sites. 

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #775, opened 2025-01-03
Studio: Allow to start the import when there is already a process in progress from onClick

katinthehatsite:
## Related issues

Closes https://github.com/Automattic/dotcom-forge/issues/9759

## Proposed Changes

This PR ensures that multiple import processes can be started at the same time from the `Import / Export` tab for different sites. It is currently possible with the drag and drop functionality but not with file selector. 

## Testing Instructions

* Pull the changes from this branch
* Start the app with `npm start`
* Navigate to the `Import/Export` tab on Studio site A
* Start the import process with file selector e.g. by clicking to select a file (and not drag and drop)
* Switch to the `Import/Export` tab on Studio site B
* Start the import process with file selector e.g. by clicking to select a file (and not drag and drop)
* Confirm the second import starts

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1247, opened 2025-04-23
Reset component state when switching sites

fredrikekelund:
## Related issues

- Fixes STU-180

## Proposed Changes

In https://github.com/Automattic/studio/pull/775, we worked on an issue caused by React state not resetting when selecting another site in Studio. This happens because the `key` prop passed to `TabPanel` references only the active tab and not the selected site. 

@katinthehatsite raised another issue in https://github.com/Automattic/studio/pull/1242#issuecomment-2823510545 that I suspect is related to the same thing. 

This PR fixes the problem by including the selected site ID in the `key` prop passed to `TabPanel`.

See [the React docs](https://react.dev/learn/preserving-and-resetting-state#option-2-resetting-state-with-a-key) for more details on resetting state with a `key` prop. 

## Testing Instructions

Ideally, @katinthehatsite would retry the same steps as in https://github.com/Automattic/studio/pull/1242#issuecomment-2823510545 and ensure that it works as expected now. 

For other testers, do a light smoke test of the Sync tab, the Previews tab, the Import/Export tab, and the Settings tab. While this should be a low-risk change, it also has far-reaching effects on state handling.

## Pre-merge Checklist

### DIFF
--- a/src/components/site-content-tabs.tsx
+++ b/src/components/site-content-tabs.tsx
@@ -41,7 +41,7 @@ export function SiteContentTabs() {
 				orientation="horizontal"
 				onSelect={ ( tabName ) => setSelectedTab( tabName as TabName ) }
 				initialTabName={ selectedTab }
-				key={ selectedTab }
+				key={ selectedTab + selectedSite.id }
 			>
 				{ ( { name } ) => (
 					<div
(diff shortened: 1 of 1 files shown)


epeicher (approved):
Thanks for the changes! They make sense to me and after some general testing, I have not found any issues. LGTM! :shipit: 


https://github.com/user-attachments/assets/e8679885-b543-473b-a63a-627755178b6a

wojtekn (approved):
Works fine in my tests.

katinthehatsite (approved):
Nice, thanks for the fix!

I (of course 🤦 ) was testing something else in the meantime so I removed the expired sites on my end so I can not test with the original issue. But from what I saw earlier, the changes make sense to me.

PR #48, opened 2024-04-26
i18n: Make "VS Code" and "PHPStorm" translatable

t-hamano:
## Proposed Changes

This PR makes the texts "VS Code" and "PHPstorm" translatable. Since these terms are proper nouns, the text may be the same in any language, but I think all text present in the app should be translatable.

## Testing Instructions

If it is the default language, the Overview tab content should continue to display as before.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #1252, opened 2025-04-23
Site overview: Adjust additional code editors

gavande1:
## Related issues

- Fixes STU-387.

## Proposed Changes

- Add config for handling additional editors on sites overview section.

## Testing Instructions

- Check out this branch
- Select PhpStorm or Cursor editor from preferences
- Make sure Open In... "YOUR SELECTED EDITOR" buttons appear
CleanShot 2025-04-23 at 19 02 09@2x
- Click on the button and make sure the editor opens with the site folder selected. 

## Pre-merge Checklist

### DIFF
--- a/src/lib/editor.ts
+++ b/src/lib/editor.ts
@@ -1,3 +1,5 @@
+import { __ } from '@wordpress/i18n';
+
 export type SupportedEditor = 'vscode' | 'phpstorm' | 'cursor' | 'windsurf' | 'webstorm';
 
 export type SupportedEditorConfig = {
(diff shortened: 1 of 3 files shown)


src/lib/editor.ts
```diff
@@ -1,20 +1,37 @@
-export type SupportedEditor =
-	| 'vscode'
-	| 'phpstorm'
-	| 'cursor'
-	| 'windsurf'
-	| 'nova'
-	| 'webstorm'
-	| 'sublime'
-	| 'atom';
+export type SupportedEditor = 'vscode' | 'phpstorm' | 'cursor' | 'windsurf' | 'webstorm';
```
gavande1, line 3: We have decided to continue with the five listed editors only. Read more: STU-354.

src/lib/editor.ts
```diff
@@ -1,20 +1,37 @@
...
 	windsurf: 'Windsurf',
-	nova: 'Nova',
 	webstorm: 'WebStorm',
-	sublime: 'Sublime',
-	atom: 'Atom',
+};
+
+export const supportedEditorConfig: Record< SupportedEditor, SupportedEditorConfig > = {
+	vscode: {
+		label: 'VS Code',
```
gavande1: We don't want to translate the label. So it's fine to keep here as config.

wojtekn: Can we make those translatable, but do not translate them by default? See https://github.com/Automattic/studio/pull/48 for context.

Also, should we keep the names in sync so we have VS Code in both places instead of Visual Studio Code in one place and VS Code in another?

gcsecsey: It seems to me that `supportedEditorConfig` has the same mappings as `supportedEditorNames` but it contains more details.

`supportedEditorNames` has only been used here previously:
https://github.com/Automattic/studio/blob/93adc785c3485fd0c6f5fca7c946be4ecddf2792/src/modules/user-settings/components/editor-picker.tsx#L36-L42

 I'd suggest we refactor the usages of `supportedEditorNames` to use `supportedEditorConfig` instead, and we can remove `supportedEditorNames` altogether.

gcsecsey (commented):
Changes overall look good to me, I have a suggestion to refactor some things.

Functionally it works great, the button label is updated and I can open the project in the selected IDE:
CleanShot 2025-04-23 at 16 09 25@2x

CleanShot 2025-04-23 at 16 09 11@2x

src/lib/editor.ts
gavande1: Thanks for sharing the context. I have added it in [856c0f9](https://github.com/Automattic/studio/pull/1252/commits/856c0f9b4ccc2a0725a8f1952e4039d14105469b)

gcsecsey (approved):
Thanks for carrying out the refactoring and adding translations. Changes LGTM and still work great. 🚢

PR #1253, opened 2025-04-23
Enable CLI installation

fredrikekelund:
## Related issues

- Fixes STU-394

## Proposed Changes

Update the conditionals limiting the CLI installation feature so they are also enabled in development and beta builds of Studio.

## Testing Instructions

Code review should suffice

## Pre-merge Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -298,10 +298,7 @@ async function appBoot() {
 			'weekly'
 		);
 
-		// temporary hidden since in development yet
-		if ( process.env.NODE_ENV === 'development' ) {
-			await installCLIOnWindows();
-		}
+		await installCLIOnWindows();
 
 		finishedInitialization = true;
 	} );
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -298,10 +298,7 @@ async function appBoot() {
 			'weekly'
 		);
 
-		// temporary hidden since in development yet
-		if ( process.env.NODE_ENV === 'development' ) {
-			await installCLIOnWindows();
-		}
+		await installCLIOnWindows();
```
wojtekn: Should we use a feature flag instead?

fredrikekelund: Could you elaborate? At this point, we're enabling the feature for all users.

From a technical standpoint, we went with the "cheap" `process.env.NODE_ENV === 'development'` approach previously because the only existing logic we have for loading feature flags lives in the renderer, whereas this feature lives in the main process and runs on initialization.

wojtekn: > Could you elaborate? At this point, we're enabling the feature for all users.

If we remove this code and do not use the concept of feature flag, we will release this feature to all users with the next release on Monday. I wanted to confirm if it's what we are planning.

fredrikekelund: Ah, I see. I overlooked the fact that the 1.4.1 release target date is next Monday. Give me a minute to think about how to approach this…

fredrikekelund: I've added a feature gate based on the app version that enables the CLI installation for development and beta builds.

gcsecsey (approved):
LGTM, thanks for the detailed reasoning behind these changes above. 🙌

nightnei (approved):
LGTM 👍

PR #1254, opened 2025-04-23
Sync: Filter out a8c sites from the sync sites listing

ivan-ottinger:
## Related issues

- Resolves STU-390-linear-issue

## Proposed Changes

- filter out sites owned by a8c

## Testing Instructions

2. Head over to the Sync tab of any local site and click on the `Connect site` button.
3. Observe that the list of sites in the modal does not include any a8c sites.

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-fetch-wpcom-sites/index.tsx
+++ b/src/hooks/use-fetch-wpcom-sites/index.tsx
@@ -16,6 +16,7 @@ export const sitesEndpointSiteSchema = z.object( {
 	jetpack: z.boolean().optional(),
 	is_deleted: z.boolean(),
 	hosting_provider_guess: z.string().optional(),
+	site_owner: z.number().optional(),
 	options: z
 		.object( {
 			created_at: z.string(),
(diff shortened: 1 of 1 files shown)


wojtekn (approved):
The list of sites looks much cleaner now. 🥳

src/hooks/use-fetch-wpcom-sites/index.tsx
```diff
@@ -51,6 +52,7 @@ export const sitesEndpointResponseSchema = z.object( {
 } );
 
 const STUDIO_SYNC_FEATURE_NAME = 'studio-sync';
+const A8C_SITE_OWNER_ID = 26957695;
```
bcotrim, line 55: I don't think exposing this ID is a good approach.
We also do it in calypso, but I believe it would be safer if we could handle this in the API instead.

ivan-ottinger: I was thinking about the same as well, but since I have seen the ID used in public Calypso repo already, I followed the same approach. Let's continue the discussion in STU-390-linear-issue.

PR #1258, opened 2025-04-24
Assorted CLI tweaks

fredrikekelund:
## Related issues

- Fixes #

## Proposed Changes

- Type tweak in `src/stores/snapshot-slice.ts`
- Added a standalone `getCliPath` function
- Removed refetch call for `useGetSnapshotUsage` query in `UserSettings` (see https://github.com/Automattic/studio/pull/1244#discussion_r2057697355)
- Added `--avoid-telemetry` option to CLI for when CLI is initiated internally from Studio

## Testing Instructions

Code review should suffice

## Pre-merge Checklist

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -18,9 +18,6 @@ async function main() {
 		.usage( __( 'Studio by WordPress.com CLI' ) )
 		.locale( locale )
 		.version( version )
-		.middleware( () =>
-			bumpAggregatedUniqueStat( StatsGroup.STUDIO_CLI_USAGE_UNIQUE, StatsMetric.SUCCESS, 'weekly' )
-		)
 		.option( 'output-format', {
 			type: 'string',
 			hidden: true,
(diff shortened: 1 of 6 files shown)


cli/index.ts
```diff
@@ -31,6 +28,19 @@ async function main() {
 				return value;
 			},
 		} )
+		.option( 'avoid-telemetry', {
```
bcotrim, line 31: Nice one!

bcotrim (changes requested):
The number of active sites now doesn't get updated when we use the CLI until we refresh the app.


https://github.com/user-attachments/assets/21af8d9a-f16e-462f-930b-063058a79309

fredrikekelund:
Ah, I see what you were getting at, @bcotrim… Give me a minute to see if I can address that

fredrikekelund:
And just to clarify: the previous solution also has a bug in that deleting all sites from the settings modal refetches `useGetSnapshotUsage` too soon (i.e. before the changes have taken effect on the server). The result is that the preview site counter isn't decremented after the sites have been deleted.

fredrikekelund:
This should be good for another look now, @bcotrim!

bcotrim (commented):
My user has 2 preview sites outside of Studio.
So if I create a preview site with CLI, while having the user settings open, the number of sites will be wrong.
Opening and closing the user settings fixes this issue.

I believe it is an edge case and something we can live with for now, as I don't see a perfect solution. What do you think?

fredrikekelund:
Fair point, @bcotrim. I've updated the optimistic update to use an increment operation based on the difference between the old snapshot count and the new one. This accounts for scenarios like yours.

cli/index.ts
```diff
@@ -31,6 +28,19 @@ async function main() {
...
+			type: 'boolean',
+			hidden: true,
+		} )
+		.middleware( async ( argv ) => {
+			if ( ! argv.avoidTelemetry ) {
```
nightnei, line 36: Do we need this condition? AFAIU `avoidTelemetry` is always `true`? Or I misunderstand something?

nightnei (approved):
LGTM 👍
Left onr NIT comment

cli/index.ts
fredrikekelund: It's not true when the CLI is initiated directly from the terminal

PR #1259, opened 2025-04-24
Refactor `InstalledAppsProvider` to a new RTK api

gcsecsey:
## Related issues

- Fixes [STU-382 User settings: Refactor InstalledAppsProvider to a Redux slice](https://linear.app/a8c/issue/STU-382/user-settings-refactor-installedappsprovider-to-a-redux-slice)
- Fixes [STU-435 Site Overview: Open in Button Flicker on Site Switch](https://linear.app/a8c/issue/STU-435/site-overview-open-in-button-flicker-on-site-switch)

## Proposed Changes

- Add a new installed apps slice in Redux
- Refactor use-check-installed-apps hook to use the new slice
- Update tests to mock the installed apps data using Redux state

## Testing Instructions

- Check that the `Code editor` dropwdown still shows the correct list of installed apps
- Check that the `Terminal` dropdown still shows the correct list of installed terminals
- Kill the app and check that the selected apps are still saved to `appdata-v1.json`
CleanShot 2025-04-24 at 11 33 15@2x

## Pre-merge Checklist

- [X] Have you checked for TypeScript, React or other console errors?


src/stores/installed-apps-slice.ts
```diff
@@ -0,0 +1,45 @@
...
+			// Update the state with the fetched installed apps
+			Object.assign( state, action.payload );
+		} );
+	},
+} );
+
+export const installedAppsActions = installedAppsSlice.actions;
+export const reducer = installedAppsSlice.reducer;
+export const selectInstalledApps = ( state: { installedApps: InstalledApps } ) =>
+	state.installedApps;
```
fredrikekelund: Given how simple this selector is, I don't really think it's needed. For future reference, `createSlice` accepts a `selectors` object where each selector automatically receives just the local state slice.

src/stores/installed-apps-slice.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+const initialState: InstalledApps = {
+	vscode: false,
+	phpstorm: false,
+	webstorm: false,
+	windsurf: false,
+	cursor: false,
+	iterm: false,
+};
```
fredrikekelund:
```suggestion
type InstalledAppsState = InstalledApps & InstalledTerminals;

const initialState: InstalledAppsState = {
	vscode: false,
	phpstorm: false,
	webstorm: false,
	windsurf: false,
	cursor: false,
	iterm: false,
	terminal: false,
};
```
You know the details of this project better than I, @gcsecsey, but I would use the global `InstalledApps` and `InstalledTerminals` type definitions to define the state. When using those types, it looks like we also need to add the `terminal` key to the state.

src/stores/installed-apps-slice.ts
fredrikekelund, line 1: I'm thinking that `createApi` might actually be a better fit than `createSlice` here… `src/stores/app-version-api.ts` should offer up some good inspiration.

In short, `createApi` (RTK Query) is Redux Toolkit's answer to react-query. It would allow us to drop the `useCheckInstalledApps` hook altogether, since the fetching and caching logic is handled internally by RTK Query. 

Typically, `createApi` calls would fetch stuff over the network, but calling IPC methods turns out to be a great use case, too.

src/components/tests/content-tab-overview-shortcuts-section.test.tsx
fredrikekelund, line 1: Given how many times we repeat the `configureStore` call in this file, I'm thinking it might be better to follow the example from `src/components/tests/content-tab-settings.test.tsx` and augment `installedAppsReducer` with a `setInstalledApps` action. 

We may have to alter this if we go with the `createApi` approach that I suggested, though.

fredrikekelund (commented):
Thanks for tackling this, @gcsecsey 🙏 This is looking good overall, but I realize now that `createApi` ([docs](https://redux-toolkit.js.org/rtk-query/api/createApi)) would probably be a better fit here than `createSlice`. As I mentioned in another comment, `src/stores/app-version-api.ts` should offer up some good inspiration

src/stores/installed-apps-slice.ts
gcsecsey: Thanks @fredrikekelund for the suggestion, I like that approach much better too!

src/stores/installed-apps-slice.ts
gcsecsey: Thanks, I didn't know that we have that option for `createSlice`, that's pretty handy! 👍

src/stores/installed-apps-slice.ts
gcsecsey: Thanks for the suggestion, I looked into `createApi` a bit and how it's used in `src/stores/app-version-api.ts`, and I agree that it'd be a more streamlined solution to this use case. I'll update this file to use it and refactor the usages/tests soon.

src/stores/installed-apps-slice.ts
fredrikekelund: `createSlice` is honestly a dream to work with compared to "classic" Redux 😅

src/components/tests/content-tab-overview-shortcuts-section.test.tsx
gcsecsey: I updated these tests to use a similar approach to `content-tab-settings.test.tsx` in [2e31df2](https://github.com/Automattic/studio/pull/1259/commits/2e31df2ad643a69735bf1e845386b2541b21b5b9)

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -12,31 +11,16 @@ interface EditorPickerProps {
...
-		getIpcApi()
-			.getInstalledApps()
-			.then( ( apps ) => {
-				setInstalledApps( apps );
-			} )
-			.catch( ( error ) => {
-				console.error( 'Failed to fetch installed apps:', error );
-			} );
-	}, [] );
+	console.log( installedApps );
```
fredrikekelund:
```suggestion
```

src/stores/installed-apps-api.ts
```diff
@@ -0,0 +1,39 @@
...
+						data: installedAppsInitialState,
+					};
+				}
+			},
+			providesTags: [ 'InstalledApps' ],
+		} ),
+	} ),
+} );
+
+export const { useGetInstalledAppsQuery } = installedAppsApi;
```
fredrikekelund:
```suggestion
export const { useGetInstalledAppsQuery } = installedAppsApi;

export const selectInstalledEditors = createSelector(
	[ ( data?: InstalledAppsState ) => data ],
	( installedApps ) => {
		const entries = Object.entries( supportedEditorConfig ) as [
			SupportedEditor,
			SupportedEditorConfig,
		][];

		return entries.filter( ( [ editor ] ) => installedApps && installedApps[ editor ] );
	}
);

export const selectUninstalledEditors = createSelector(
	[ ( data?: InstalledAppsState ) => data ],
	( installedApps ) => {
		const entries = Object.entries( supportedEditorConfig ) as [
			SupportedEditor,
			SupportedEditorConfig,
		][];

		return entries.filter( ( [ editor ] ) => ! installedApps || ! installedApps[ editor ] );
	}
);

```
Consider this a suggestion, but I would add selectors here to select the installed editors. The `createSelector` API isn't the most intuitive, but it gives us built-in caching that's useful.

src/stores/installed-apps-api.ts
```diff
@@ -0,0 +1,39 @@
...
+export const installedAppsApi = createApi( {
+	reducerPath: 'installedAppsApi',
+	baseQuery: fetchBaseQuery(),
+	tagTypes: [ 'InstalledApps' ],
+	endpoints: ( builder ) => ( {
+		getInstalledApps: builder.query< InstalledAppsState, void >( {
+			queryFn: async () => {
+				try {
+					const installedApps = await getIpcApi().getInstalledApps();
+					return { data: installedApps as InstalledAppsState };
```
fredrikekelund:
```suggestion
					return { data: installedApps };
```
`getIpcApi().getInstalledApps()` seems to only return data about editors, not about terminals. We could either call `getIpcApi().getInstalledTerminals()`, too, and await both promises, or we refactor `ipc-handlers` to return data about editors and terminals from a single function.

Or, alternatively, we add a second query/endpoint to fetch data about installed terminals.

src/stores/installed-apps-api.ts
```diff
@@ -0,0 +1,39 @@
...
+				} catch ( error ) {
+					return {
+						data: installedAppsInitialState,
+					};
+				}
```
fredrikekelund:
```suggestion
				const installedApps = await getIpcApi().getInstalledApps();
				return { data: installedApps };
```
Unless we have a specific reason to return the default state when exceptions are caught, I wouldn't add error handling inside `queryFn`.

src/stores/installed-apps-api.ts
```diff
@@ -0,0 +1,39 @@
...
+	phpstorm: false,
+	webstorm: false,
+	windsurf: false,
+	cursor: false,
+	terminal: false,
+	iterm: false,
+	ghostty: false,
+	warp: false,
+};
+
```
fredrikekelund:
```suggestion
```
Assuming we follow my suggestion about error handling, we can remove this part.

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -12,31 +11,16 @@ interface EditorPickerProps {
...
 export const EditorPicker = ( { value, onChange }: EditorPickerProps ) => {
 	const { __ } = useI18n();
-	const [ installedApps, setInstalledApps ] = useState< {
-		vscode: boolean | null;
-		phpstorm: boolean | null;
-	} >( {
-		vscode: null,
-		phpstorm: null,
-	} );
+	const { data: installedApps } = useGetInstalledAppsQuery();
```
fredrikekelund:
```suggestion
	const { installedEditors, uninstalledEditors } = useGetInstalledAppsQuery( undefined, {
		selectFromResult: ( result ) => ( {
			installedEditors: selectInstalledEditors( result.data ),
			uninstalledEditors: selectUninstalledEditors( result.data ),
		} ),
	} );
```
If we follow my suggestion of adding selectors to `src/stores/installed-apps-api.ts`, then we can update this call like so.

src/modules/user-settings/components/editor-picker.tsx
fredrikekelund, line 1: `src/modules/user-settings/components/terminal-picker.tsx` should also receive similar edits. We just need to make it so `installedAppsApi` also handles terminals.

fredrikekelund (changes requested):
This is shaping up nicely! I'm requesting changes because in my view, `src/stores/installed-apps-api.ts` should also handle terminals.

I don't want to blow the scope here completely, but I would also argue that `useEditorData` and `useTerminalData` should be folded into `installed-apps-api.ts`. There may be some complexity around handling state changes until the user saves their changes, but fundamentally, RTK Query has [queries](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usequery) and [mutations](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usemutation) to handle both reads and writes, which should cover our needs.

src/stores/installed-apps-api.ts
gcsecsey: Both `getInstalledApps()` and `getInstalledTerminals()` are using `isInstalled`, so I'd much prefer going the refactoring `ipc-handlers` route and moving everything to one place.

gcsecsey:
Thanks for all your helpful comments and suggestions @fredrikekelund! 🙌 I committed some of the suggestions. I'll continue working on folding both terminals and editors into `installed-apps-api.ts` and ask for a new review when ready.

gcsecsey:
I made the changes and updated the tests, this is now ready for another review.

fredrikekelund:
Taking a look at this now.

src/components/content-tab-overview.tsx
```diff
@@ -142,18 +140,10 @@ function CustomizeSection( {
...
-	useEffect( () => {
-		handleEditorChange( editor );
-		handleTerminalChange( terminal );
-	}, [ editor, handleEditorChange, handleTerminalChange, terminal ] );
-
-	useIpcListener( 'user-preference-changed', async () => {
-		handleEditorChange( await getSavedEditor() );
-		handleTerminalChange( await getSavedTerminal() );
-	} );
+	const terminalName = supportedTerminalNames[ terminal as keyof typeof supportedTerminalNames ];
```
fredrikekelund:
```suggestion
	const terminalName = getTerminalName( terminal );
```
Consider this a suggestion, but a helper function could help us do away with the type casting. This could go in `src/modules/user-settings/lib/terminal.ts`:

```ts
export function getTerminalName( terminal: SupportedTerminal | undefined ): string {
	if ( ! terminal ) {
		return '';
	}

	return supportedTerminalNames[ terminal ];
}
```

src/modules/user-settings/components/tests/terminal-picker.test.tsx
```diff
@@ -1,24 +1,127 @@
...
+				phpstorm: false,
+				webstorm: false,
+				windsurf: false,
+				cursor: false,
+				terminal: true,
+				iterm: true,
+				warp: false,
+				ghostty: false,
+			} )
+		);
```
fredrikekelund:
```suggestion
		store.dispatch(
			installedAppsApi.util.updateQueryData( 'getInstalledApps', undefined, ( data ) => {
				return {
					vscode: false,
					phpstorm: false,
					webstorm: false,
					windsurf: false,
					cursor: false,
					terminal: true,
					iterm: true,
					warp: false,
					ghostty: false,
				};
			} )
		);
```
RTK Query comes with an `updateQueryData` helper that seems to do the trick here.

src/modules/user-settings/components/tests/terminal-picker.test.tsx
```diff
@@ -1,24 +1,127 @@
...
+				webstorm: false,
+				windsurf: false,
+				cursor: false,
+				// Terminal properties
+				terminal: true,
+				iterm: true,
+				warp: false,
+				ghostty: false,
+			} )
+		);
```
fredrikekelund:
```suggestion
		store.dispatch(
			installedAppsApi.util.updateQueryData( 'getInstalledApps', undefined, ( data ) => {
				return {
					// Editor properties
					vscode: false,
					phpstorm: false,
					webstorm: false,
					windsurf: false,
					cursor: false,
					// Terminal properties
					terminal: true,
					iterm: true,
					warp: false,
					ghostty: false,
				};
			} )
		);
```

src/modules/user-settings/components/tests/terminal-picker.test.tsx
```diff
@@ -1,24 +1,127 @@
...
+
+// Create test actions for installedAppsApi
+const installedAppsTestActions = {
+	setInstalledApps: ( installedApps: InstalledAppsState ) => {
+		return { type: 'installedApps/setInstalledApps', payload: { installedApps } };
+	},
+};
+
+// Replace the store's reducer with our test reducer
+store.replaceReducer( installedAppsTestReducer as Reducer< RootState > );
```
fredrikekelund:
```suggestion
store.replaceReducer( testReducer );
```
If we use the `updateQueryData` API, then it looks like `installedAppsTestReducer` is redundant.

src/components/tests/content-tab-overview-shortcuts-section.test.tsx
```diff
@@ -16,11 +23,53 @@ const selectedSite: StartedSiteDetails = {
...
+
+// Create test actions for installedAppsApi
+const installedAppsTestActions = {
+	setInstalledApps: ( installedApps: InstalledAppsState ) => {
+		return { type: 'installedApps/setInstalledApps', payload: { installedApps } };
+	},
+};
+
+// Replace the store's reducer with our test reducer
+store.replaceReducer( installedAppsTestReducer as Reducer< RootState > );
```
fredrikekelund:
```suggestion
store.replaceReducer( testReducer );
```
Looks like `installedAppsTestReducer` isn't used here, and we can stick with just `testReducer`

src/stores/installed-apps-api.ts
fredrikekelund, line 1: This entire file is looking great 👍

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -1,65 +1,52 @@
...
 					</option>
 				) ) }
-				<optgroup label={ __( 'Not installed' ) }>
-					{ uninstalledEditors.map( ( [ editorKey, editorConfig ] ) => (
-						<option key={ editorKey } value={ editorKey } disabled>
-							{ editorConfig.label }
-						</option>
-					) ) }
-				</optgroup>
+				{ uninstalledEditors.length > 0 && (
```
fredrikekelund: 👍

src/ipc-handlers.ts
```diff
@@ -106,13 +106,17 @@ export async function getSiteDetails( _event: IpcMainInvokeEvent ): Promise< Sit
...
 
-export function getInstalledApps(): InstalledApps {
+export function getInstalledAppsAndTerminals(): InstalledApps & InstalledTerminals {
 	return {
 		vscode: isInstalled( 'vscode' ),
 		phpstorm: isInstalled( 'phpstorm' ),
 		webstorm: isInstalled( 'webstorm' ),
 		windsurf: isInstalled( 'windsurf' ),
 		cursor: isInstalled( 'cursor' ),
+		terminal: true, // Terminal.app is always available on macOS
```
fredrikekelund, line 116: Beside the point of this PR, but I assume this key means `Terminal.app` on macOS and `Command Prompt` on Windows, right? I was mostly thinking of the comment.

src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -5,51 +5,57 @@ import { useI18nData } from 'src/hooks/use-i18n-data';
...
 			<LanguagePicker value={ locale } onChange={ setLocale } />
-			<EditorPicker value={ editor } onChange={ handleEditorChange } />
-			<TerminalPicker
-				value={ terminal }
-				onChange={ handleTerminalChange }
-				availableTerminals={ availableTerminals }
+			<EditorPicker
+				value={ currentEditor }
+				onChange={ setCurrentEditor }
+				disabled={ editor === undefined }
```
fredrikekelund, line 56:
```suggestion
				disabled={ isEditorLoading }
```
This line had me confused to begin with because of the "fallback logic" in `getUserEditor`. I realize now that what we're saying here is "disable this control while the editors are loading". I think it would be clearer if we used the `isLoading` prop returned by `useGetUserEditorQuery` for that purpose.

src/components/content-tab-overview.tsx
```diff
@@ -142,18 +140,10 @@ function CustomizeSection( {
...
 
-	useEffect( () => {
-		handleEditorChange( editor );
-		handleTerminalChange( terminal );
-	}, [ editor, handleEditorChange, handleTerminalChange, terminal ] );
-
-	useIpcListener( 'user-preference-changed', async () => {
-		handleEditorChange( await getSavedEditor() );
-		handleTerminalChange( await getSavedTerminal() );
-	} );
```
fredrikekelund, line 156: Did we rely on this logic before, or was it just a kind of nice to have?

fredrikekelund (approved):
This is looking great 👍 Nice work!

I left a couple of comments, but the biggest ones are all just related test setup logic. Approving now, since none of my remarks are blocking, even if they'd be good to address.

src/ipc-handlers.ts
gcsecsey: Yes, we use the same `terminal` key to refer to either default option. We'll show the appropriate label for the app after https://github.com/Automattic/studio/pull/1275 is merged.

src/components/content-tab-overview.tsx
gcsecsey: We did previously, this was introduced when we updated the settings screen: https://github.com/Automattic/studio/pull/1215

src/modules/user-settings/components/preferences-tab.tsx
gcsecsey: We had a quick call with @gavande1 yesterday and I realized that I forgot to move over the fallback logic from `getUserEditor`. My intention with these changes was to account for a case where there's no selected editors, and no supported editors are installed on the system either.

For this case, we have a Figma design based on which we should disable the dropdown, and show a CTA to install VSCode. @gavande1 worked on updating the UI with these in [STU-436 User Settings: Add CTA for installing VSCode if none of supported editor is installed](https://linear.app/a8c/issue/STU-436/user-settings-add-cta-for-installing-vscode-if-none-of-supported), but in this PR I `getUserEditor` with the fallback logic and added a `undefined` case.

src/components/tests/content-tab-overview-shortcuts-section.test.tsx
gcsecsey: Nice catch, thanks!

src/modules/user-settings/components/tests/terminal-picker.test.tsx
gcsecsey: This is a very handy util, thanks for highlighting! 🙌

src/modules/user-settings/components/tests/terminal-picker.test.tsx
gcsecsey: Thanks for the suggestion, updated in b89185c

src/modules/user-settings/components/tests/terminal-picker.test.tsx
gcsecsey: That's a very handy util, thanks for highlighting! 🙌

sejas:
Let's not merge this PR until we release 1.5.0. Thank you!

src/ipc-handlers.ts
```diff
@@ -106,13 +106,17 @@ export async function getSiteDetails( _event: IpcMainInvokeEvent ): Promise< Sit
 	return mergeSiteDetailsWithRunningDetails( sites );
 }
 
-export function getInstalledApps(): InstalledApps {
+export function getInstalledAppsAndTerminals(): InstalledApps & InstalledTerminals {
```
fredrikekelund: It seems to me that the `InstalledApps` and `InstalledTerminals` types are always used together now. Maybe we should just merge them in `src/ipc-types.d.ts`?

gcsecsey: That's a good point, thanks. I updated it in [2735521](https://github.com/Automattic/studio/pull/1259/commits/2735521498b95d7ba3a57d2623f829c2215e42a2)

PR #1261, opened 2025-04-24
Build the CLI with `mode: production` in the Webpack config

fredrikekelund:
## Related issues

- Fixes STU-396

## Proposed Changes

When building the CLI as part of the electron-forge packaging flow, we should use the `mode: production` Webpack config option (see [docs](https://webpack.js.org/configuration/mode/)) to ensure that `process.env.NODE_ENV` is interpolated to the correct value.

This fixes an issue where `Would have bumped stat…` is printed by the CLI in production environments.

## Testing Instructions

1. Download a dev build from Buildkite
2. Explore the package contents and open `Contents/Resources/cli/main.js`
3. Find the `bumpStat` function
4. Ensure it doesn't contain a `console.info( 'Would have bumped stat…` call

## Pre-merge Checklist

### DIFF
--- a/cli/lib/appdata.ts
+++ b/cli/lib/appdata.ts
@@ -114,8 +114,7 @@ export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken
 		return authToken;
 	} catch ( error ) {
 		throw new LoggerError(
-			__( 'Authentication required. Please run the Studio app and log in to WordPress.com first.' ),
-			error
+			__( 'Authentication required. Please run the Studio app and log in to WordPress.com first.' )
 		);
 	}
 }
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -132,7 +132,7 @@ const config: ForgeConfig = {
 			}
 
 			console.log( 'Building CLI ...' );
-			const compiler = webpack( cliConfig );
+			const compiler = webpack( { ...cliConfig, mode: 'production' } );
 
 			await new Promise< void >( ( resolve, reject ) => {
 				compiler.run( ( error ) => {


forge.config.ts
```diff
@@ -132,7 +132,7 @@ const config: ForgeConfig = {
 			}
 
 			console.log( 'Building CLI ...' );
-			const compiler = webpack( cliConfig );
+			const compiler = webpack( { ...cliConfig, mode: 'production' } );
```
epeicher, line 135: It seems that the `mode` is coming from [this line of code](https://github.com/Automattic/studio/blob/8ab546db73a64e3cfd964a6cb95d2b37a7b65cff/webpack.cli.config.ts#L15) so we are actually defaulting to `development` when there are no `process.env.NODE_ENV` set. Do you think it could be better to refine that `mode` in `cliConfig` instead of overriding here the `mode` to `production` always?

fredrikekelund: There are four ways to build the CLI right now:

- Run `npm start`, which calls the `prestart` npm script (in turn running the `cli:build` npm script)
- Run `npm run cli:build`
- Run `npm run cli:watch`
- Build the app with Electron Forge

IMO, the last case is the only time when we can be sure that the CLI is being packaged for the "end consumer", which is why I modified the `mode` parameter in this way. 

Running `NODE_ENV=production npm run cli:build`, for example, would produce the equivalent output.

epeicher: Thanks for the clarification, as I'm still getting familiar with the build process. My point was that if we are using `mode: 'production'` here, we are not using [mode in the webpack.cli.config.js](https://github.com/Automattic/studio/blob/8ab546db73a64e3cfd964a6cb95d2b37a7b65cff/webpack.cli.config.ts#L15), right? 🤔. It is a bit of a nitpick though 🤓

cli/lib/appdata.ts
```diff
@@ -114,8 +114,7 @@ export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken
 		return authToken;
 	} catch ( error ) {
 		throw new LoggerError(
-			__( 'Authentication required. Please run the Studio app and log in to WordPress.com first.' ),
-			error
+			__( 'Authentication required. Please run the Studio app and log in to WordPress.com first.' )
```
fredrikekelund, line 117: This change is unrelated to the core purpose of this PR, but it's so small I didn't think it deserved its own PR. Not passing the previous error to `LoggerError` means that the CLI will print the following error message:

```
✖ Authentication required. Please run the Studio app and log in to WordPress.com first.
```

Previously, it would print:

```
✖ Authentication required. Please run the Studio app and log in to WordPress.com first.: Authentication required
```

forge.config.ts
fredrikekelund: Hmm, I don't think we need to adjust `webpack.cli.config.ts`. Currently, the `mode` option defaults to `development`, which is what we want when developing the app/CLI. The only build scenario where I think it makes sense to default to `mode: production` is when we're building and packaging the app with Electron Forge (which is what this change accomplishes). 

Let me know if I'm misunderstanding something.

epeicher (approved):
I have tested it, and it works as advertised. I cannot see the `console.info` in this version of the package while I can see it in other builds. LGTM!

|Other builds | This build|
|-------|------|
|   CleanShot 2025-04-25 at 10 53 25@2x|CleanShot 2025-04-25 at 10 49 20@2x|

forge.config.ts
epeicher: Thanks @fredrikekelund for the additional explanations. I was incorrectly assuming that the `webpack.cli.config.js` was used only in this line of code, but it is actually being used in the package.json scripts `npm run cli:build` and `npm run cli:watch` as you indicated above. This makes total sense now ✅

PR #1265, opened 2025-04-25
Studio: Fix translations related to terminal apps

katinthehatsite:
## Related issues

Related to https://github.com/Automattic/studio/pull/1240#discussion_r2058546234

## Proposed Changes

This PR cleans up some translations e.g. redundant translation functions related to terminal apps. In terms of app functionality, nothing should be changing with how the app works. 

## Testing Instructions

* Open `User settings` by clicking your profile picture
* Click on `Preferences` tab
* Try changing to different terminals and confirm you can open your Studio sites with different terminals and they correctly update in the `Overview` tab

## Pre-merge Checklist


ivan-ottinger (commented):
I wonder if we need to translate those terminal names in the first place. Wouldn't the app be called the same in the OS regardless of the locale the user is using? 🤔 

One thing I noticed is that if I haven't set the preferred terminal before, the button on the Overview tab is empty:

Markup on 2025-04-25 at 10:22:18

katinthehatsite:
Thanks for the review!

> I wonder if we need to translate those terminal names in the first place. Wouldn't the app be called the same in the OS regardless of the locale the user is using? 🤔

I am not sure either but we already are doing it with the code editors so I think it makes sense to keep it for consistency.

> One thing I noticed is that if I haven't set the preferred terminal before, the button on the Overview tab is empty 

Let me check. I think the intended behaviour is to default to terminal initially if nothing is selected and for the code editors, we should not display the button at all.

katinthehatsite:
@ivan-ottinger I believe this should be fixed now. Feel free to give it another review 🙇

src/ipc-handlers.ts
```diff
@@ -1282,7 +1282,7 @@ export async function saveUserTerminal(
 
 export async function getUserTerminal( _event: IpcMainInvokeEvent ): Promise< SupportedTerminal > {
 	const userData = await loadUserData();
-	return userData.supportedTerminal as SupportedTerminal;
+	return ( userData.supportedTerminal as SupportedTerminal ) || DEFAULT_TERMINAL;
```
ivan-ottinger: Nice! This fixes the issue with empty terminal indeed.

src/lib/terminal.ts
```diff
@@ -1,10 +1,15 @@
...
+
 export type SupportedTerminal = 'terminal' | 'iterm' | 'warp' | 'ghostty';
 
 export const supportedTerminalNames: Record< SupportedTerminal, string > = {
-	terminal: 'Terminal',
-	iterm: 'iTerm',
-	warp: 'Warp',
-	ghostty: 'Ghostty',
+	terminal: __( 'Terminal' ),
+	// translators: "iTerm" is the brand name for a terminal app and does not need to be translated
```
ivan-ottinger: Do we want to add  a similar translators comment for the "Terminal" terminal?

ivan-ottinger (approved):
Thank you for the updates, Kat! 🙂 

> I am not sure either but we already are doing it with the code editors so I think it makes sense to keep it for consistency.

Makes sense. 👍🏼 

> I believe this should be fixed now. Feel free to give it another review 🙇

I can confirm this is fixed now. ✅ 

---

The terminal switcher works well and the preferred terminal app is being stored in the `appdata-v1.json`.

src/ipc-handlers.ts
```suggestion
	return userData.supportedTerminal || DEFAULT_TERMINAL;
```
No need for the `as` statement anymore.

fredrikekelund:
We could give `getUserEditor` in `src/ipc-handlers.ts` a similar treatment to `getUserTerminal`. There's no `DEFAULT_EDITOR` constant, but I would just remove the `as` statement in that function _and_ the return type. That way, Typescript knows that `getIpcApi().getUserEditor()` can return `SupportedEditor | undefined`.

src/ipc-handlers.ts
katinthehatsite: Hmm I am unsure as it is not something that we had before as we use Terminal to mark both command line on Windows and Terminal app for macOS 🤔

src/lib/terminal.ts
katinthehatsite: Hmm I am unsure as it is not something that we had before as we use Terminal to mark both command line on Windows and Terminal app for macOS 🤔

src/ipc-handlers.ts
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/1265/commits/769193f3f79728852a8396f540635c62f5fe2c7a

src/lib/terminal.ts
gavande1: @katinthehatsite I was wondering if we should refer to the Terminal as "Command Prompt" on Windows since we are defaulting to it and not the Terminal (PowerShell). What do you think?

katinthehatsite: I am not sure 😅 I simply left what was there before but it makes sense to reconsider. Maybe @gcsecsey can offer an insight as a project lead

katinthehatsite: Let's discuss it here: STU-405.

gcsecsey: It's a bit tricky situation, I added my thoughts and investigation to the new issue (STU-405). Overall I agree with @gavande1 that we should just rename the default option to "Command Prompt".

PR #1268, opened 2025-04-25
Print URL to log in to WordPress.com in the CLI

fredrikekelund:
## Related issues

- Related to pdtkmj-3Am-p2#comment-6888

## Proposed Changes

When the user isn't signed into WordPress.com in Studio, the `studio preview` commands don't work (except for `list`). However, printing a URL that can be used to sign into WordPress.com in Studio is straightforward, and this PR implements that change. 

## Testing Instructions

1. Be logged out of WordPress.com in Studio
2. Run `npm run cli:build`
3. Run `node node dist/cli/main.js preview create [folder]`
4. Ensure that the command prints an error message including a URL to sign in to WordPress.com
5. Ensure that signing in to WordPress.com works when opening the URL

## Pre-merge Checklist


sejas (changes requested):
It works as expected, but we need to change the source of the refactored constant imports in:
- ./src/components/content-tab-previews.tsx:12:2
- ./src/components/content-tab-sync.tsx:1:10
- ./src/index.ts:15:10

For example in `src/components/content-tab-sync.tsx`

```diff
- import { CLIENT_ID, PROTOCOL_PREFIX, SCOPES, WP_AUTHORIZE_ENDPOINT } from 'src/constants';
+ import { CLIENT_ID, PROTOCOL_PREFIX, SCOPES, WP_AUTHORIZE_ENDPOINT } from 'common/constants';
```


CLI works as expected:

Screenshot 2025-04-25 at 12 10 59

sejas (approved):
I've updated the constants on https://github.com/Automattic/studio/pull/1268/commits/70de5b4854b31cabb8cab595ac6ebbe5e43227ec . It should be good to go! 👍

fredrikekelund:
Thanks for pushing a fix, @sejas 👍

fredrikekelund:
I took another look at the components you edited, @sejas, and thought that now would be a good time to centralize the logic for generating signup URLs. 

So, I updated the `authenticate` function in `src/ipc-handlers.ts` to take an `isSignup` argument that determines if we should generate a signup URL or a login URL. 

I'd appreciate another review when you have a moment 🙏

common/constants.ts
```diff
@@ -2,3 +2,9 @@ export const DEMO_SITE_SIZE_LIMIT_GB = 2;
 export const DEMO_SITE_SIZE_LIMIT_BYTES = DEMO_SITE_SIZE_LIMIT_GB * 1024 * 1024 * 1024; // 2GB
 export const HOUR_MS = 1000 * 60 * 60;
 export const DAY_MS = HOUR_MS * 24;
+
+// OAuth constants
+export const CLIENT_ID = '95109';
+export const PROTOCOL_PREFIX = 'wpcom-local-dev';
+export const WP_AUTHORIZE_ENDPOINT = 'https://public-api.wordpress.com/oauth2/authorize';
+export const SCOPES = 'global';
```
sejas: Should we move `WP_AUTHORIZE_ENDPOINT` and `SCOPES` into `common/lib/oauth.ts`? It seems that they are not used anywhere else now. Or maybe even hardcode them.

sejas (approved):
Beautiful refactor ✨.  I left a suggestion about moving `WP_AUTHORIZE_ENDPOINT` and `SCOPES` into `common/lib/oauth.ts`.  

I tested the login flow from the CLI, preview sites, sync, and assistant 👍. I also tested the signup flow from the preview sites. All of them worked correctly.  

I also noticed that creating a new account from the Assistant tab opens a different page. That behavior is also present in the trunk, so I created a separate issue to address it: STU-407.

common/constants.ts
fredrikekelund: Makes sense 👍 I made a few more changes:

- Inlined `WP_AUTHORIZE_ENDPOINT` in `getAuthenticationUrl`
- Moved `SCOPES` to `common/lib/oauth.ts`
- Refactored `getAuthenticationUrl` to use a URL constructor for added clarity and correctness

sejas (approved):
Looks good.
I also noticed we've removed `&from-calypso=1` which makes sense 👌 .

PR #1270, opened 2025-04-25
Wire up void IPC handlers using `ipcMain.on`

fredrikekelund:
## Related issues

- Fixes a regression from https://github.com/Automattic/studio/pull/1086

## Proposed Changes

- Use a constant to define the list of void IPC handlers instead of a type. 
- Derive the `IpcApi` directly from this constant. 
- Use this list to determine which channels should use `ipcMain.on` and `ipcMain.handle` respectively in the `setupIpc` function in `src/index.ts`. 

## Testing Instructions

1. Start Studio
2. Ensure that you can open any URL (a site homepage, for example)

## Pre-merge Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,7 +12,7 @@ import path from 'path';
 import * as Sentry from '@sentry/electron/main';
 import { __ } from '@wordpress/i18n';
 import { StatsGroup } from 'common/types/stats';
-import { PROTOCOL_PREFIX } from 'src/constants';
+import { PROTOCOL_PREFIX, IPC_VOID_HANDLERS } from 'src/constants';
 import * as ipcHandlers from 'src/ipc-handlers';
 import { hasActiveSyncOperations } from 'src/lib/active-sync-operations';
 import { bumpAggregatedUniqueStat, bumpStat } from 'src/lib/bump-stats';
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -148,29 +148,27 @@ async function appBoot() {
...
-		for ( const [ key, handler ] of Object.entries( ipcHandlers ) ) {
-			if ( typeof handler === 'function' && key !== 'logRendererMessage' ) {
-				ipcMain.handle( key, function ( event, ...args ) {
+		const ipcHandlerEntries = Object.entries( ipcHandlers ) as [
+			keyof typeof ipcHandlers,
+			( ...args: unknown[] ) => unknown,
+		][];
+
+		for ( const [ key, handler ] of ipcHandlerEntries ) {
+			if ( IPC_VOID_HANDLERS.find( ( handler ) => handler === key ) ) {
```
fredrikekelund, line 157: I would have used `IPC_VOID_HANDLERS.includes`, but Typescript is very picky about the types that can be passed to that method, so I went with this approach instead.

katinthehatsite: Yeah I saw the same issue when trying a quick fix on my end 👍

katinthehatsite (approved):
The changes look good to me and work as expected 👍 I tried a similar solution so I think it makes sense.

One thing is that I do not love is that `IPC_VOID_HANDLERS` now live somewhat further from ipc-types but I see there are comments on how it works so hopefully it won't be confusing for folks when adding new IPC handlers.

fredrikekelund:
> `IPC_VOID_HANDLERS` now live somewhat further from ipc-types

I initially tried placing that constant in `src/preload.ts`, but that didn't work because the main process can't import members of that module. I think we would run into similar issues by putting the constant in a main-process file. `src/constants.ts` is neutral territory 😄

Agreed that it's kind of spread out right now, but it should hopefully be relatively easy to backtrack.

katinthehatsite:
> I initially tried placing that constant in src/preload.ts, but that didn't work because the main process can't import members of that module.

Yeah that is the idea I had too 😅 but I think comments are clear enough to make it work 😄

fredrikekelund:
The Sentry issue is not new. It was previously tracked in STUDIO-1M6, but this PR changed the exception signature. I'm looking into what happens when the exception is thrown.

PR #1271, opened 2025-04-25
Site Overview: setup default state for editor and terminal

gavande1:
## Related issues

- Fixes STU-409

## Proposed Changes

- Use data provider hooks to populate the default editor and terminal on the overview screen. 
- Remove duplicate implementation for getting preferred editor and terminal. 
- Revert saved terminal to `preferredTerminal`

## Testing Instructions

- Check out this branch
- Remove preferred editor and terminal from appdata-v1.json. (/Users/[USERNAME]/Library/Application Support/Studio/appdata-v1.json)
- Run `npm start`
- Visit the site overview and see that the `VS Code` and `Terminal` buttons are visible.

CleanShot 2025-04-25 at 19 08 09@2x

- Terminate `npm start`. 
- Run `STUDIO_PREFERRED_EDITOR=true npm start`. 
- Check site overview for `VS Code` and `Terminal` buttons. 
- Update preferred editor or terminal in settings. 
- Ensure overview updates with the correct editor or terminal after saving settings. 

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-overview.tsx
+++ b/src/components/content-tab-overview.tsx
@@ -26,7 +26,7 @@ import { cx } from 'src/lib/cx';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { useEditorData } from 'src/modules/user-settings/hooks/use-editor-data';
 import { useTerminalData } from 'src/modules/user-settings/hooks/use-terminal-data';
-import { SupportedEditor, supportedEditorConfig } from 'src/modules/user-settings/lib/editor';
+import { supportedEditorConfig } from 'src/modules/user-settings/lib/editor';
 import { supportedTerminalNames } from 'src/modules/user-settings/lib/terminal';
 
 interface ContentTabOverviewProps {
@@ -170,7 +170,7 @@ function ShortcutsSection( { selectedSite }: Pick< ContentTabOverviewProps, 'sel
 		},
 	];
 
-	const editorConfig = editor ? supportedEditorConfig[ editor as SupportedEditor ] : false;
+	const editorConfig = editor ? supportedEditorConfig[ editor ] : false;
 	if ( editorConfig ) {
 		buttonsArray.push( {
 			label: editorConfig.label,
--- a/src/components/content-tab-overview.tsx
+++ b/src/components/content-tab-overview.tsx
@@ -14,19 +14,20 @@ import {
 	widget,
 } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
-import { useState, useEffect, useCallback } from 'react';
+import { useState, useEffect } from 'react';
 import { ArrowIcon } from 'src/components/arrow-icon';
 import { ButtonsSection, ButtonsSectionProps } from 'src/components/buttons-section';
-import { useCheckInstalledApps } from 'src/hooks/use-check-installed-apps';
+import { useEditorData } from 'src/hooks/use-editor-data';
 import { useFeatureFlags } from 'src/hooks/use-feature-flags';
 import { useIpcListener } from 'src/hooks/use-ipc-listener';
 import { useSiteDetails } from 'src/hooks/use-site-details';
+import { useTerminalData } from 'src/hooks/use-terminal-data';
 import { useThemeDetails } from 'src/hooks/use-theme-details';
 import { isMac } from 'src/lib/app-globals';
 import { cx } from 'src/lib/cx';
 import { getIpcApi } from 'src/lib/get-ipc-api';
-import { supportedEditorConfig } from 'src/modules/user-settings/lib/editor';
-import { supportedTerminalNames, DEFAULT_TERMINAL } from 'src/modules/user-settings/lib/terminal';
+import { SupportedEditor, supportedEditorConfig } from 'src/modules/user-settings/lib/editor';
+import { SupportedTerminal, supportedTerminalNames } from 'src/modules/user-settings/lib/terminal';
 
 interface ContentTabOverviewProps {
 	selectedSite: SiteDetails;
@@ -141,34 +142,28 @@ function CustomizeSection( {
 
 function ShortcutsSection( { selectedSite }: Pick< ContentTabOverviewProps, 'selectedSite' > ) {
 	const { terminalWpCliEnabled } = useFeatureFlags();
-	const installedApps = useCheckInstalledApps();
-	const [ terminalName, setTerminalName ] = useState( supportedTerminalNames[ DEFAULT_TERMINAL ] );
-	const [ editorName, setEditorName ] = useState( '' );
+	const { terminal, getSavedTerminal } = useTerminalData();
+	const { editor, getSavedEditor } = useEditorData();
+	const [ terminalName, setTerminalName ] = useState( supportedTerminalNames[ terminal ] );
+	const [ editorName, setEditorName ] = useState( editor );
 
-	const updateTerminalName = useCallback( async () => {
-		const terminal = await getIpcApi().getUserTerminal();
-		setTerminalName( supportedTerminalNames[ terminal ] );
-	}, [ setTerminalName ] );
+	const handleEditorChange = ( newEditor: SupportedEditor ) => {
+		setEditorName( newEditor );
+	};
 
-	const updateEditorName = useCallback( async () => {
-		const editor = await getIpcApi().getUserEditor();
-
-		if ( editor && installedApps[ editor as keyof typeof installedApps ] ) {
-			setEditorName( editor );
-		} else {
-			setEditorName( '' );
-		}
-	}, [ setEditorName, installedApps ] );
-
-	useIpcListener( 'user-preference-changed', () => {
-		void updateTerminalName();
-		void updateEditorName();
-	} );
+	const handleTerminalChange = ( newTerminal: SupportedTerminal ) => {
+		setTerminalName( supportedTerminalNames[ newTerminal ] );
+	};
 
 	useEffect( () => {
-		void updateTerminalName();
-		void updateEditorName();
-	}, [ updateTerminalName, updateEditorName ] );
+		handleEditorChange( editor );
+		handleTerminalChange( terminal );
+	}, [ editor, terminal ] );
+
+	useIpcListener( 'user-preference-changed', async () => {
+		handleEditorChange( await getSavedEditor() );
+		handleTerminalChange( await getSavedTerminal() );
+	} );
 
 	const buttonsArray: ButtonsSectionProps[ 'buttonsArray' ] = [
 		{
--- a/src/components/content-tab-overview.tsx
+++ b/src/components/content-tab-overview.tsx
@@ -147,22 +147,22 @@ function ShortcutsSection( { selectedSite }: Pick< ContentTabOverviewProps, 'sel
 	const [ terminalName, setTerminalName ] = useState( supportedTerminalNames[ terminal ] );
 	const [ editorName, setEditorName ] = useState( editor );
 
-	const handleEditorChange = ( newEditor: SupportedEditor ) => {
+	const updateEditorName = ( newEditor: SupportedEditor ) => {
 		setEditorName( newEditor );
 	};
 
-	const handleTerminalChange = ( newTerminal: SupportedTerminal ) => {
+	const updateTerminalName = ( newTerminal: SupportedTerminal ) => {
 		setTerminalName( supportedTerminalNames[ newTerminal ] );
 	};
 
 	useEffect( () => {
-		handleEditorChange( editor );
-		handleTerminalChange( terminal );
+		updateEditorName( editor );
+		updateTerminalName( terminal );
 	}, [ editor, terminal ] );
 
 	useIpcListener( 'user-preference-changed', async () => {
-		handleEditorChange( await getSavedEditor() );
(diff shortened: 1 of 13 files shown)


src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -38,8 +37,6 @@ export const PreferencesTab = ( { onClose }: { onClose: () => void } ) => {
 		onClose();
 	};
 
-	const hasChanges = locale !== savedLocale || hasEditorChanges || hasTerminalChanges;
```
gavande1: This change was necessary because I encountered a dead block. I did not have VS Studio installed, but Cursor. When I visited the settings screens, the cursor was displayed as the selected option. However, since there was no change selection, the Save button was disabled for me, and I could not edit anything.

gcsecsey: We can take a look into fixing this edge case in `use-editor-data` and/or `use-terminal-data`. Alternatively, if we always show the Save button, then we can also remove the `hasEditorChanges` and  `hasTerminalChanges` functions from these hooks.

gcsecsey (commented):
Thanks for working on these changes @gavande1! 🙌 I like the approach of setting up the default editor and terminal on the overview tab. I found some potential simplifications that were easier for me to push than to start describing in the review, but please feel free to revert these if needed.

src/modules/user-settings/components/preferences-tab.tsx
gcsecsey: I tried replicating this issue by returning false for VS Code in `ipc-handlers`:
```diff
diff --git a/src/ipc-handlers.ts b/src/ipc-handlers.ts
index 87e67bf..803e2f3 100644
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -107,7 +107,7 @@ export async function getSiteDetails( _event: IpcMainInvokeEvent ): Promise< Sit
 
 export function getInstalledApps(): InstalledApps {
        return {
-               vscode: isInstalled( 'vscode' ),
+               vscode: false,
                phpstorm: isInstalled( 'phpstorm' ),
                webstorm: isInstalled( 'webstorm' ),
                windsurf: isInstalled( 'windsurf' ),
```

But for me, I'm able to select Cursor then save the dialog. Are you referring to another issue above?

https://github.com/user-attachments/assets/0dfbcf59-92cb-4d72-bba0-79f177f16854

gavande1: @gcsecsey I experienced it on Windows. Have you tried it on?

src/modules/user-settings/lib/editor.ts
```diff
@@ -2,6 +2,8 @@ import { __ } from '@wordpress/i18n';
 
 export type SupportedEditor = 'vscode' | 'phpstorm' | 'cursor' | 'windsurf' | 'webstorm';
 
+export const DEFAULT_EDITOR = 'vscode';
```
katinthehatsite: If we want to use some constants, I think we can place them in the constants file. What do you think?

katinthehatsite: Also, if I am not mistaken, I think that with the previous implementation if the user had PhpStorm installed, they would PhpStorm displayed instead of `VsCode` but `VsCode` would be default otherwise. If this the behaviour we are accounting for in the implementation when the app is started without the feature flag?

src/components/content-tab-overview.tsx
```diff
@@ -14,19 +14,20 @@ import {
...
 import { useFeatureFlags } from 'src/hooks/use-feature-flags';
 import { useIpcListener } from 'src/hooks/use-ipc-listener';
 import { useSiteDetails } from 'src/hooks/use-site-details';
+import { useTerminalData } from 'src/hooks/use-terminal-data';
 import { useThemeDetails } from 'src/hooks/use-theme-details';
 import { isMac } from 'src/lib/app-globals';
 import { cx } from 'src/lib/cx';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { supportedEditorConfig } from 'src/modules/user-settings/lib/editor';
-import { supportedTerminalNames, DEFAULT_TERMINAL } from 'src/modules/user-settings/lib/terminal';
```
katinthehatsite, line 29: GH does not allow me to choose the line where `DEFAULT_TERMINAL` is declared but we should also move this to constants file to keep them all organized in one place

src/ipc-handlers.ts
```diff
@@ -1282,7 +1282,7 @@ export async function saveUserTerminal(
 
 export async function getUserTerminal( _event: IpcMainInvokeEvent ): Promise< SupportedTerminal > {
 	const userData = await loadUserData();
-	return userData.supportedTerminal || DEFAULT_TERMINAL;
+	return userData.preferredTerminal as SupportedTerminal;
```
katinthehatsite: I am wondering if we should keep the default here and add a feature flag to set the default for when the app is started without the feature flag. I think this way, the two functions (this one and `openTerminalAtPath`) will keep two different parts of logic and `openTerminalAtPath` would not need to handle default. 

What do you think?

src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -53,7 +49,7 @@ export const PreferencesTab = ( { onClose }: { onClose: () => void } ) => {
 				<Button variant="tertiary" onClick={ cancelChanges }>
 					{ __( 'Cancel' ) }
 				</Button>
-				<Button variant="primary" onClick={ savePreferences } disabled={ ! hasChanges }>
```
katinthehatsite: Is not the purpose of this code to keep the `Save` button disabled when there are no changes to either of the fields? It seems that it currently shows as active even if you did not make any changes which could be slightly confusing for the user and is not consistent with other modals that we have in the app e.g. `Edit site` modal.

src/modules/user-settings/lib/editor.ts
gavande1: > If we want to use some constants, I think we can place them in the constants file. What do you think?

Thanks for the suggestion. I have moved in [a4b5aa7](https://github.com/Automattic/studio/pull/1271/commits/a4b5aa7f1981976c2487d6a5288406979ffac3c0).

>  if I am not mistaken, I think that with the previous implementation if the user had PhpStorm installed, they would PhpStorm displayed instead of VsCode but VsCode would be default otherwise. If this the behaviour we are accounting for in the implementation when the app is started without the feature flag?

No. We are not. I plan to follow up on that after we land this PR.

src/components/content-tab-overview.tsx
gavande1: I have moved in [a4b5aa7](https://github.com/Automattic/studio/pull/1271/commits/a4b5aa7f1981976c2487d6a5288406979ffac3c0).

src/modules/user-settings/components/preferences-tab.tsx
gavande1: Yes. It is. But on Windows, I hit a dead block where I could not save anything.  https://github.com/Automattic/studio/pull/1271#discussion_r2060306082

src/ipc-handlers.ts
gavande1: Thanks for highlighting this. I added it back in [b00cf0a](https://github.com/Automattic/studio/pull/1271/commits/b00cf0a567450624f6c2af9bd8f9e50b0bf71954)!

src/modules/user-settings/lib/editor.ts
gavande1: @katinthehatsite I created follow-up PR: https://github.com/Automattic/studio/pull/1284

src/modules/user-settings/components/preferences-tab.tsx
gavande1: @gcsecsey I ended up addressing the edge case, then removing it completely. I had to add an empty option when no editor is selected in the preferences. Let me know what you think about it.

src/modules/user-settings/components/preferences-tab.tsx
gavande1: @katinthehatsite, I was able to revert this behavior by adding an additional empty Select when no editor is saved in preferences.

src/modules/user-settings/hooks/use-editor-data.ts
```diff
@@ -3,15 +3,30 @@ import { getIpcApi } from 'src/lib/get-ipc-api';
 import { SupportedEditor } from 'src/modules/user-settings/lib/editor';
 
 export function useEditorData() {
-	const [ editor, setEditor ] = useState< SupportedEditor >( 'vscode' );
-	const [ savedEditorValue, setSavedEditorValue ] = useState< SupportedEditor >( 'vscode' );
+	const [ editor, setEditor ] = useState< SupportedEditor | string >( '' );
+	const [ savedEditorValue, setSavedEditorValue ] = useState< SupportedEditor | string >( '' );
```
fredrikekelund:
```suggestion
	const [ editor, setEditor ] = useState< SupportedEditor >();
	const [ savedEditorValue, setSavedEditorValue ] = useState< SupportedEditor >();
```
Initiating the state like so sets the variable types to `SupportedEditor | undefined`. 

We can please Typescript if we change `getSavedEditor` to default to return `undefined` and the two following functions like so:

```ts
	const handleEditorChange = ( newEditor: SupportedEditor | undefined ) => {
		if ( newEditor ) {
			setEditor( newEditor );
		}
	};

	const saveEditorPreference = async () => {
		if ( editor ) {
			await getIpcApi().saveUserEditor( editor );
		}
	};
```

The current type definitions force us to use the `as` operator all over the place, which is essentially short-circuiting type checking. Let's try to update the type definitions so we don't have to use the `as` operator.

gavande1: Thanks for highlighting it. I was literaly thinking about the change I made last evening related to using `<string>`. I am in the process of refactoring it.

fredrikekelund: Nice 👍

gavande1: I factored the code to use `undefined`. Added in [1dcdca3](https://github.com/Automattic/studio/pull/1271/commits/1dcdca351bb2bc4bc31b18ae803479c0d48dc3e2)!

src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -15,6 +15,7 @@ export const PreferencesTab = ( { onClose }: { onClose: () => void } ) => {
 
 	const { editor, handleEditorChange, saveEditorPreference, resetEditor, hasEditorChanges } =
 		useEditorData();
+
```
katinthehatsite, line 18: This seems to be intended line. Also, it looks like we are not using `hasEditorChanges` in that file, is that right?

katinthehatsite: I realized that I did not actually have the most recent commit, so ignore this comment. We are using it 👍

src/components/tests/content-tab-overview-shortcuts-section.test.tsx
```diff
@@ -147,24 +139,95 @@ describe( 'ShortcutsSection', () => {
...
+			<ContentTabOverview selectedSite={ selectedSite } />
+		);
+
+		await findByLabelText( 'Terminal' );
+		await findByLabelText( 'PhpStorm' );
+
+		expect( queryByLabelText( 'VS Code' ) ).toBeNull();
+	} );
+
+	it( 'shows users preffered editor', async () => {
```
katinthehatsite:
```suggestion
	it( 'shows users preferred editor', async () => {
```

Just small typo 😅

src/constants.ts
```diff
@@ -27,6 +27,7 @@ export const FEATURE_REQUEST_URL =
 export const WPCOM_PROFILE_URL = 'https://wordpress.com/me';
 export const LOCAL_STORAGE_CHAT_MESSAGES_KEY = 'ai_chat_messages';
 export const LOCAL_STORAGE_CHAT_API_IDS_KEY = 'ai_chat_ids';
+export const DEFAULT_TERMINAL = 'terminal';
```
katinthehatsite, line 30: Nice!

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -47,6 +47,7 @@ export const EditorPicker = ( { value, onChange }: EditorPickerProps ) => {
 				__nextHasNoMarginBottom
 				__next40pxDefaultSize
 			>
+				{ ! value && <option value={ '' }>{ __( 'Select' ) }</option> }
```
katinthehatsite, line 50: Is there a possibility to handle this in a similar manner to `Edit site` modal where we apply a similar behavior?

katinthehatsite (approved):
i think the changes look good, I left a couple of minor comments but feel free to proceed with merging 👍  Nice work!

gcsecsey (approved):
These changes work well for me too.

src/modules/user-settings/components/editor-picker.tsx
gavande1: @katinthehatsite, can you please elaborate more on your suggestion? I checked the `Edit site` modal but could not find anything closely related to this implementation. Happy to follow up on it if needed.

PR #1272, opened 2025-04-25
Add site to Studio when creating preview from CLI

bcotrim:
## Related issues

- Fixes STU-395

## Proposed Changes

- Added `newSites` field to user data schema to track sites pending creation
- Implemented handling for the case when a preview is created for a site that doesn't exist in Studio
- Created a Redux slice to manage new site processing
- Added IPC handlers and listeners to automatically create sites when detected
- Updated relevant tests to cover new functionality

## Testing Instructions

- Run `npm run start` and create a new site in Studio
- Go to settings, copy the path of the site and delete it from Studio with the option to keep files in the system
- Run `npm run cli:build`
- Run `node dist/cli/main.js preview create <STUDIO SITE FOLDER>`
- Wait for preview creation success
- Go back to Studio
- There should be a new site, with the name = pathname
- That site should have the newly created preview

## Pre-merge Checklist

### DIFF
--- a/cli/lib/appdata.ts
+++ b/cli/lib/appdata.ts
@@ -1,3 +1,4 @@
+import crypto from 'crypto';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
@@ -16,8 +17,16 @@ const siteSchema = z
 	} )
 	.passthrough();
 
+const newSiteSchema = z
+	.object( {
+		id: z.string(),
+		path: z.string(),
+	} )
+	.passthrough();
+
 const userDataSchema = z
 	.object( {
+		newSites: z.array( newSiteSchema ).optional(),
 		sites: z.array( siteSchema ).optional(),
 		snapshots: z.array( snapshotSchema ).optional(),
 		locale: z.string().optional(),
@@ -128,14 +137,24 @@ export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken
 
 export async function getSiteByFolder(
 	siteFolder: string
-): Promise< z.infer< typeof siteSchema > > {
+): Promise< z.infer< typeof siteSchema > | null > {
 	const userData = await readAppdata();
 	const sites = userData.sites ?? [];
 	const site = sites.find( ( site ) => site.path === siteFolder );
 
 	if ( ! site ) {
-		throw new LoggerError( __( 'The specified folder is not added to Studio.' ) );
+		return null;
 	}
 
 	return site;
+}
+
+export function getNewSiteAppData( siteFolder: string ): z.infer< typeof siteSchema > {
+	const newSite = {
+		id: crypto.randomUUID(),
+		path: siteFolder,
+		name: path.basename( siteFolder ),
+	};
+
+	return newSite;
 }
--- a/cli/lib/appdata.ts
+++ b/cli/lib/appdata.ts
@@ -17,16 +17,9 @@ const siteSchema = z
 	} )
 	.passthrough();
 
-const newSiteSchema = z
-	.object( {
-		id: z.string(),
-		path: z.string(),
-	} )
-	.passthrough();
-
 const userDataSchema = z
 	.object( {
-		newSites: z.array( newSiteSchema ).optional(),
+		newSites: z.array( siteSchema ).optional(),
 		sites: z.array( siteSchema ).optional(),
 		snapshots: z.array( snapshotSchema ).optional(),
 		locale: z.string().optional(),
@@ -137,19 +130,19 @@ export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken
 
 export async function getSiteByFolder(
 	siteFolder: string
-): Promise< z.infer< typeof siteSchema > | null > {
+): Promise< z.infer< typeof siteSchema > > {
 	const userData = await readAppdata();
 	const sites = userData.sites ?? [];
 	const site = sites.find( ( site ) => site.path === siteFolder );
 
 	if ( ! site ) {
-		return null;
+		throw new LoggerError( __( 'The specified folder is not added to Studio.' ) );
 	}
 
 	return site;
 }
 
-export function getNewSiteAppData( siteFolder: string ): z.infer< typeof siteSchema > {
+export function getNewSitePartial( siteFolder: string ): z.infer< typeof siteSchema > {
 	const newSite = {
 		id: crypto.randomUUID(),
 		path: siteFolder,
--- a/src/ipc-types.d.ts
+++ b/src/ipc-types.d.ts
@@ -45,6 +45,12 @@ interface StartedSiteDetails extends StoppedSiteDetails {
 
 type SiteDetails = StartedSiteDetails | StoppedSiteDetails;
 
+type NewSiteDetails = {
+	id: string;
+	path: string;
+	name: string;
+};
+
 type InstalledApps = {
 	vscode: boolean;
 	phpstorm: boolean;
--- a/src/ipc-types.d.ts
+++ b/src/ipc-types.d.ts
@@ -45,11 +45,7 @@ interface StartedSiteDetails extends StoppedSiteDetails {
 
 type SiteDetails = StartedSiteDetails | StoppedSiteDetails;
 
-type NewSiteDetails = {
-	id: string;
-	path: string;
-	name: string;
-};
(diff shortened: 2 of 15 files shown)


src/stores/new-sites-slice.ts
```diff
@@ -0,0 +1,48 @@
...
+		setIsProcessing: ( state, action: PayloadAction< boolean > ) => {
+			state.isProcessing = action.payload;
+		},
+	},
+} );
+
+window.ipcListener.subscribe( 'user-data-updated', ( _, payload ) => {
+	const state = store.getState() as RootState & { newSites?: NewSitesState };
+
+	const newSites = payload.newSites;
```
fredrikekelund:
```suggestion
	const state = store.getState();
	const newSites = payload.newSites;
```
No need for `as`

src/stores/new-sites-slice.ts
```diff
@@ -0,0 +1,48 @@
...
+					console.error(
+						`[New Sites Slice] Failed to create site for folder: ${ site.path }`,
+						error
+					);
+				}
+			} )
+		).finally( () => {
+			store.dispatch( newSitesSlice.actions.setIsProcessing( false ) );
+		} );
+	}
```
fredrikekelund:
```suggestion
	if ( ! state.newSites.isProcessing && newSites ) {
		store.dispatch( handleNewSite( newSites ) );
	}
```

I would make the event handler smaller and add a `handleNewSite` thunk like this:

```ts
const handleNewSite = createAsyncThunk( 'newSites/handleNewSite', ( sites: NewSiteDetails[] ) => {
	return Promise.all(
		sites.map( async ( site ) => {
			try {
				await getIpcApi().handleNewSite( site );
			} catch ( error ) {
				console.error(
					`[New Sites Slice] Failed to create site for folder: ${ site.path }`,
					error
				);
			}
		} )
	);
} );
```

Lastly, we should update the `isProcessing` state depending on the thunk state:

```ts
	extraReducers: ( builder ) => {
		builder
			.addCase( handleNewSite.pending, ( state ) => {
				state.isProcessing = true;
			} )
			.addCase( handleNewSite.fulfilled, ( state ) => {
				state.isProcessing = false;
			} )
			.addCase( handleNewSite.rejected, ( state ) => {
				state.isProcessing = false;
			} );
	},
```

cli/lib/appdata.ts
```diff
@@ -122,14 +131,24 @@ export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken
...
 	siteFolder: string
-): Promise< z.infer< typeof siteSchema > > {
+): Promise< z.infer< typeof siteSchema > | null > {
 	const userData = await readAppdata();
 	const sites = userData.sites ?? [];
 	const site = sites.find( ( site ) => site.path === siteFolder );
 
 	if ( ! site ) {
-		throw new LoggerError( __( 'The specified folder is not added to Studio.' ) );
+		return null;
```
fredrikekelund: I guess this is ultimately subjective, but I would rather continue throwing an error from this function and instead wrap the call in `saveSnapshotToAppdata` in a `try .. catch`

cli/lib/appdata.ts
```diff
@@ -122,14 +131,24 @@ export async function getAuthToken(): Promise< NonNullable< UserData[ 'authToken
...
 	}
 
 	return site;
 }
+
+export function getNewSiteAppData( siteFolder: string ): z.infer< typeof siteSchema > {
```
```suggestion
export function getNewSitePartial( siteFolder: string ) {
```
I feel like we could clarify the name here. I would also let TS automatically intuit the return type.

cli/lib/appdata.ts
```diff
@@ -15,8 +16,16 @@ const siteSchema = z
...
+const newSiteSchema = z
+	.object( {
+		id: z.string(),
+		path: z.string(),
+	} )
+	.passthrough();
+
 const userDataSchema = z
 	.object( {
+		newSites: z.array( newSiteSchema ).optional(),
```
```suggestion
const userDataSchema = z
	.object( {
		newSites: z.array( siteSchema ).optional(),
```
In my quick testing, it looked like we could use `siteSchema` here, too. Maybe the current code is a leftover from a previous iteration?

fredrikekelund (commented):
I haven't tested the changes yet, but judging by the code, this looks like a solid solution 👍

fredrikekelund (commented):
I've tested the changes now, and everything works well apart from the fact that the list of sites isn't automatically refreshed. Maybe we could export a `refreshSitesList` function from `src/hooks/use-site-details.tsx` to resolve that?

src/ipc-types.d.ts
```diff
@@ -45,6 +45,12 @@ interface StartedSiteDetails extends StoppedSiteDetails {
 
 type SiteDetails = StartedSiteDetails | StoppedSiteDetails;
 
+type NewSiteDetails = {
+	id: string;
+	path: string;
+	name: string;
+};
```
```suggestion
type NewSiteDetails = Pick< SiteDetails, 'id' | 'path' | 'name' >;
```
Nit, but this allows us to continue using a canonical type definition for the sites objects.

src/stores/new-sites-slice.ts
```diff
@@ -0,0 +1,59 @@
...
+			}
+		} )
+	);
+} );
+
+window.ipcListener.subscribe( 'user-data-updated', ( _, payload ) => {
+	const state = store.getState();
+	const newSites = payload.newSites;
+
+	if ( ! state.newSites.isProcessing && newSites ) {
```
fredrikekelund: I'm curious, what would happen if we didn't have the `! state.newSites.isProcessing` check here? It's not immediately clear to me why there might be multiple events in such short succession and why we shouldn't process all of them if that happens.

fredrikekelund (commented):
Tested again, and it works nicely with auto-refreshing the sites list 👍 Impressively small diff for a relatively complex change, too. Nice job!

It would be good to iron out the question about whether the `isProcessing` state is truly needed, but other than that, this LGTM.

src/stores/new-sites-slice.ts
bcotrim: Since this event subscription handles all user data changes, and not only `newSites` ones, I think it is crucial to have some tracking mechanism to avoid site duplication. However I see your point, and issue, that a single flag would probably lead to some unwanted behaviors, like new sites not being picked up if some are currently processing, and staying pending, until another user data event (probably unrelated) gets triggered.

I've updated the implementation to track individual site IDs instead of a single processing flag. Now we filter out only sites already being processed while allowing new ones to proceed immediately. 

Let me know what you think about this approach.

src/stores/new-sites-slice.ts
```diff
@@ -0,0 +1,70 @@
...
+		},
+		removeProcessingSites: ( state, action: PayloadAction< string[] > ) => {
+			state.processingSiteIds = state.processingSiteIds.filter(
+				( id ) => ! action.payload.includes( id )
+			);
+		},
+	},
+} );
+
+export const { addProcessingSites, removeProcessingSites } = newSitesSlice.actions;
```
fredrikekelund:
```suggestion
export const newSitesActions = newSitesSlice.actions;
```
Nit, but it'd be nice to stay consistent with other slices and export the actions inside a "namespace".

src/stores/tests/new-sites-slice.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+describe( 'newSitesSlice', () => {
+	describe( 'reducers', () => {
+		it( 'should add site IDs to processingSiteIds when addProcessingSites is dispatched', () => {
+			const initialState = { processingSiteIds: [] };
+
+			const action = {
+				type: 'newSites/addProcessingSites',
+				payload: [ 'site-1', 'site-2' ],
+			};
```
fredrikekelund:
```suggestion
			const action = newSitesActions.addProcessingSites( [ 'site-1', 'site-2' ] );
```
If we export the actions in a single namespace, we can create the action in this way.

fredrikekelund (approved):
LGTM 🚀

src/stores/new-sites-slice.ts
fredrikekelund: Makes sense. If a single `user-data-updated` event carried multiple `newSites`, then this logic will be required to avoid adding duplicate site entries. 

The updated solution looks great to me 👍

PR #1279, opened 2025-04-27
Studio: Build windows and mac artifact only on trunk branch

gavande1:
## Proposed Changes

- Skip Mac and Windows builds when PR is created. 

## Why changes are being made

- The Mac and Windows builds run on every PR. I think this is not useful in any case. 

## Testing Instructions

- Make sure that while creating this PR, the Windows and Mac build pipelines are not triggered. 

## Pre-merge Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -125,7 +125,7 @@ steps:
         notify:
           - github_commit_status:
               context: All Mac Dev Builds
-    if: build.tag !~ /^v[0-9]+/
+    if: build.branch == 'trunk' && build.tag !~ /^v[0-9]+/
 
   - label: 🔨 Windows Dev Build
     key: dev-windows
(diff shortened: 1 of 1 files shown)


bcotrim (commented):
Thanks for making this change and opening the PR for discussion.
I understand the argument that running Mac and Windows builds on every PR consumes time and resources. However, I also see the value in these checks—removing them could eventually lead to broken builds in trunk that would have been caught earlier.
My suggestion is to make a data-driven decision. Are we actually saving significant CI costs by removing these checks? If so, by how much? Are we saving developer time, and can we quantify that? Having this information will help us make a more informed decision, in my opinion.

gavande1:
Thank you for your thoughtful feedback.
>  also see the value in these checks—removing them could eventually lead to broken builds in trunk that would have been caught earlier.

I completely agree that running Mac and Windows builds on PRs helps in catching failures early, and that's a strong reason for me to discard this PR.

I just wanted to share that my initial intent was mainly to save developer time and avoid unnecessary computational resource usage, but I now realise that the benefit of early detection outweighs those concerns.

PR #1281, opened 2025-04-27
Studio: Refactor windows port checker to use `portFinder` helpers

gavande1:
## Related issues

- Fixes STU-103

> [!NOTE]
> We also have checkPortInWindows in src/lib/proxy-server.ts. There's an obvious difference in that the checkPortInWindows function only runs on Windows, but I still wonder if we shouldn't refactor that function and PortFinder::isPortFree to share the same underlying logic.
> - https://github.com/Automattic/studio/pull/1236#issuecomment-2826996329 

## Proposed Changes

- Refactor Windows port checker to use `portFinder` helpers

## Testing Instructions

1. Check out the branch on a Windows machine.
2. Open a new terminal tab and run:
   ```bash
   php -S localhost:80
   ```
3. In another terminal tab, start Studio by running:
   ```bash
   npm start
   ```
4. Confirm that you receive an alert indicating the port is busy (similar to the screenshots provided).

CleanShot 2025-04-27 at 18 44 30@2x

CleanShot 2025-04-27 at 18 46 30@2x

5. Stop Studio (`Ctrl+C` in the terminal).
6. Cancel the PHP server process (`Ctrl+C` in the other terminal).
7. Wait a few seconds to ensure both processes have fully stopped.
8. Start Studio again by running:
   ```bash
   npm start
   ```
9. Confirm that this time, you **do not** receive any alert about the port being busy.

## Pre-merge Checklist

### DIFF
--- a/src/lib/proxy-server.ts
+++ b/src/lib/proxy-server.ts
@@ -102,15 +102,11 @@ async function handleProxyRequest(
  * On Windows, node doesn't throw an error if port is busy, so we use portFinder to check
  * if the port is available.
  */
-export async function checkPortInWindows( port: number ): Promise< boolean > {
-	if ( process.platform !== 'win32' ) {
-		return true;
-	}
-
+export async function checkIfPortIsFree( port: number ): Promise< boolean > {
 	const portAvailable = await portFinder.isPortAvailable( port );
 
 	if ( ! portAvailable ) {
-		throw new Error( 'EADDRINUSE' );
+		throw new Error( 'PROXY_ERROR_PORT_IN_USE' );
 	}
 
 	return portAvailable;
@@ -124,7 +120,7 @@ export async function startProxyServer(): Promise< boolean > {
 	try {
 		// Start HTTP server if not already running
 		if ( ! isHttpProxyRunning ) {
-			await checkPortInWindows( 80 );
+			await checkIfPortIsFree( 80 );
 			httpProxyServer = http.createServer( ( req, res ) => handleProxyRequest( req, res, false ) );
 			await new Promise< void >( ( resolve, reject ) => {
 				httpProxyServer!
@@ -142,7 +138,7 @@ export async function startProxyServer(): Promise< boolean > {
 
 		// Start HTTPS server if not already running
 		if ( ! isHttpsProxyRunning ) {
-			await checkPortInWindows( 443 );
+			await checkIfPortIsFree( 443 );
 			const defaultOptions: https.ServerOptions = {
 				SNICallback: async ( servername, cb ) => {
 					try {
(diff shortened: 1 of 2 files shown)


fredrikekelund (approved):
Elegant and works as expected 👍

Not directly related to this PR, but I noticed this error message:

> Studio needs to use port 80 and 443 to enable custom domains and SSL, but one of both of these ports are already in use by another app. Close any local development apps and restart Studio.

I would remove "of both" from that message to make it sound more idiomatic. I.e., it would read:

> Studio needs to use port 80 and 443 to enable custom domains and SSL, but one of these ports are already in use by another app. Close any local development apps and restart Studio.

src/lib/proxy-server.ts
```diff
@@ -99,35 +99,21 @@ async function handleProxyRequest(
...
 /**
- * On Windows, node doesn't throw an error if port is busy, so we use the net module to explicitly check
- * if it's possible to establish a TCP connection to that port (meaning it's busy).
+ * On Windows, node doesn't throw an error if port is busy, so we use portFinder to check
+ * if the port is available.
  */
 export async function checkPortInWindows( port: number ): Promise< boolean > {
 	if ( process.platform !== 'win32' ) {
 		return true;
 	}
```
fredrikekelund: Not a requirement, but coming back to this code, I would argue that we should run this logic on all platforms (not just on Windows). Yes, `http.createServer` already throws an error on macOS, but making Studio's business logic consistent across platforms is a way to avoid future headaches. This PR seems like a good opportunity to address that.

fredrikekelund: If we make this change, I'd also throw an error with the `PROXY_ERROR_PORT_IN_USE` directly from this function, and adjust the `catch` clause in `startProxyServer` to rethrow the error.

bcotrim (approved):
LGTM 👍

src/lib/proxy-server.ts
gavande1: Excellent suggestions @fredrikekelund! I will implement those changes.

gavande1: Addressed your feedback in [b1f2b46](https://github.com/Automattic/studio/pull/1281/commits/b1f2b46de512fc4ad21be85a403620a298ae2ca9)!

src/lib/proxy-server.ts
```diff
@@ -209,10 +191,7 @@ export async function startProxyServer(): Promise< boolean > {
 
 		return true;
 	} catch ( error ) {
-		if (
-			( isErrnoException( error ) && error.code === 'EADDRINUSE' ) ||
-			( error instanceof Error && error.message === 'EADDRINUSE' )
-		) {
+		if ( error instanceof Error && error.message === 'PROXY_ERROR_PORT_IN_USE' ) {
 			throw new Error( 'PROXY_ERROR_PORT_IN_USE' );
```
```suggestion
			throw error;
```
No need to construct a new error instance

src/lib/proxy-server.ts
```diff
@@ -1,11 +1,11 @@
 import http from 'http';
 import https from 'https';
-import { createConnection } from 'node:net';
 import { createSecureContext } from 'node:tls';
 import { domainToASCII } from 'node:url';
 import * as Sentry from '@sentry/electron/main';
 import httpProxy from 'http-proxy';
 import { isErrnoException } from 'src/lib/is-errno-exception';
```
```suggestion
```
This import is unused now

fredrikekelund (approved):
LGTM with two nits 👍

PR #1284, opened 2025-04-28
Site Overview: Check installed editors before setting default state for editor

gavande1:
## Related issues

- Fixes STU-409

## Proposed Changes

- Check installed apps before setting the default editor. If VS Code is installed, use it as the editor; otherwise, look for phpStorm if it is installed. If neither is available, set an empty string. 
- Introduce some more tests
- Remove default editor constant

## Testing Instructions

Refer: https://github.com/Automattic/studio/pull/1271

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-overview.tsx
+++ b/src/components/content-tab-overview.tsx
@@ -26,7 +26,7 @@ import { cx } from 'src/lib/cx';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { useEditorData } from 'src/modules/user-settings/hooks/use-editor-data';
 import { useTerminalData } from 'src/modules/user-settings/hooks/use-terminal-data';
-import { supportedEditorConfig } from 'src/modules/user-settings/lib/editor';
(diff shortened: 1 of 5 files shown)


katinthehatsite:
I think it works as described in the PR description - I have both VsCode and PhpStorm installed and the app ended up using VsCode. I am wondering though if this was the default behaviour before we introduced the feature flag and all the related changes?

katinthehatsite:
I found this PR: https://github.com/Automattic/studio/commit/357befcc45da961f943e2f555b119cae42fe4a4 which seems to clarify the original implementation:

```
Make smart decision on MAC os:
* Rename labels, from Shortcuts -> Open in…
* Display one IDE button:
* Displaying PHP STORM button if the user has it installed and does not have VS Code
* Displaying VSCode if user has it installed
```

In that case, I think this implementation looks good 👍

katinthehatsite (approved):

gavande1:
Thanks @katinthehatsite for the review. Yes. I looked at the same implementation that you mentioned in your comment.

PR #1289, opened 2025-04-29
Updated command signatures for CLI

fredrikekelund:
## Related issues

- Fixes STU-419
- Fixes STU-418

## Proposed Changes

- Change the folder/path option from a positional to a `--path` option. This harmonises the `studio preview update` and `studio preview delete` command signatures, and aligns with the WP-CLI convention.
- Resolve potentially relative paths before passing them to the command handler.

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js preview list --path PATH` where `PATH` is an **absolute** path to a Studio site with preview sites
3. Ensure that the preview sites are logged as expected
4. Run `node dist/cli/main.js preview list --path PATH` where `PATH` is an **relative** path to a Studio site with preview sites
5. Ensure that the preview sites are logged as expected
6. `cd` into the aforementioned `PATH` and run `node PATH_TO_MAIN_JS`
7. Ensure that the preview sites are logged as expected

## Pre-merge Checklist

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1,3 +1,4 @@
+import path from 'node:path';
 import { __ } from '@wordpress/i18n';
 import { StatsGroup, StatsMetric } from 'common/types/stats';
 import yargs from 'yargs';
@@ -31,6 +32,13 @@ async function main() {
 		.option( 'avoid-telemetry', {
 			type: 'boolean',
 			hidden: true,
+		} )
+		.option( 'path', {
+			type: 'string',
+			default: process.cwd(),
+			defaultDescription: __( 'Current directory' ),
+			description: __( 'Path to the WordPress files' ),
+			coerce: ( value ) => path.resolve( process.cwd(), value ),
 		} )
 		.middleware( async ( argv ) => {
 			if ( ! argv.avoidTelemetry ) {
--- a/cli/commands/preview/create.ts
+++ b/cli/commands/preview/create.ts
@@ -65,19 +65,10 @@ export async function runCommand(
 
 export const registerCommand = ( yargs: StudioArgv ) => {
 	return yargs.command( {
-		command: 'create [folder]',
-		describe: __(
-			'Create a preview site from the specified folder (defaults to current directory)'
-		),
-		builder: ( yargs ) => {
-			return yargs.positional( 'folder', {
-				type: 'string',
-				default: process.cwd(),
-				description: __( 'The folder to create a preview site from' ),
-			} );
-		},
+		command: 'create',
+		describe: __( 'Create a preview site' ),
 		handler: async ( argv ) => {
-			await runCommand( argv.folder, argv.outputFormat );
+			await runCommand( argv.path, argv.outputFormat );
 		},
 	} );
 };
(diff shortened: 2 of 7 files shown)


cli/index.ts
```diff
@@ -32,6 +33,13 @@ async function main() {
 			type: 'boolean',
 			hidden: true,
 		} )
+		.option( 'path', {
```
bcotrim, line 36: With this change, the `--path` option is now available for all commands, but for some—like preview delete—the option isn’t actually used. I think this could be confusing for users, since they might expect the option to have an effect everywhere it’s shown.
Would it be possible to only include the --path option for commands that actually require it? I think this would help keep the CLI interface more intuitive.

bcotrim (commented):
Maybe it's because I'm already used to the CLI, but to me this change makes the path usage less clear:
image
Could we perhaps add some more details to the help text? For example, mentioning that the --path option can be used to specify the target directory, and that it defaults to the current directory.

cli/index.ts
fredrikekelund: I'm of two minds about this. Most commands need a reference to local WP files, and I believe this will remain true as we add more commands, too. Making `--path` a global option makes it more discoverable and easier to learn. It also aligns with the WP-CLI convention. 

At the same time, sure, being able to specify an option that doesn't have any effect is a little weird.

wojtekn: > I think this could be confusing for users, since they might expect the option to have an effect everywhere it’s shown

In the case of `preview delete,` we use a little shortcut if we don't check the path. Technically, we should check it there, too. It's not a big deal, though, as currently, preview site hosts are unique across the whole app.

However, it makes sense to consider `--path` a global option. In the WP-CLI users can also provide a path to commands that don't use that.

fredrikekelund: > In the case of preview delete, we use a little shortcut if we don't check the path. Technically, we should check it there, too

How so? Hostnames are globally unique identifiers for preview sites, so there's no need to run the command in the context of a particular Studio site.

wojtekn: >  It's not a big deal, though, as currently, preview site hosts are unique across the whole app.

fredrikekelund: > However, it makes sense to consider --path a global option. In the WP-CLI users can also provide a path to commands that don't use that.

This is a strong argument, IMO. The vast majority of WP-CLI commands are indeed affected by `--path`, but `wp package`, for example, is not.

fredrikekelund:
> Could we perhaps add some more details to the help text? For example, mentioning that the --path option can be used to specify the target directory, and that it defaults to the current directory.

I've already included help text specifically for the default option (`default: Current directory`). I'm open to tweaking the `description` config param for the `path` option, but I think I need some suggestions to understand which direction you'd like to go, @bcotrim

bcotrim:
> I've already included help text specifically for the default option (default: Current directory). I'm open to tweaking the description config param for the path option, but I think I need some suggestions to understand which direction you'd like to go, @bcotrim

After testing this version more I think it is great as is. The `path` option description shows up in the `--help` command for every context, so I think it's clear.

bcotrim (approved):
Approving after our discussion. 
LGTM 👍

PR #1296, opened 2025-04-29
Use `process.send` over stdio

fredrikekelund:
## Related issues

–

## Proposed Changes

I reevaluated how we communicate from the CLI to Studio and found that using `process.send` allows us to simplify the communication. See the `child_process.fork` [docs](https://nodejs.org/docs/latest/api/child_process.html#child_processforkmodulepath-args-options) for more details about this. 

## Testing Instructions

Ensure that creating, updating, and deleting preview sites from Studio still works as expected.

## Pre-merge Checklist


bcotrim (approved):
LGTM 👍
Personally, I think this change is great—it makes the code a lot easier to maintain and read, and the connection between Studio and the CLI is a lot better.

PR #1300, opened 2025-04-30
Additional safety around user config writes and watcher

fredrikekelund:
## Related issues

- Fixes STU-427

## Proposed Changes

@sejas noticed an intermittent bug when deleting all snapshots from the settings modal, where Studio would get stuck in an infinite write/watch loop for the user config file. 

This infinite loop is caused by `src/lib/user-data-watcher.ts` reacting to changes to `appdata-v1.json` as caused by the CLI delete processes. The watcher triggers the `user-data-updated` event in `snapshot-slice.ts` that updates the snapshots state (if it has changed). When the snapshots state changes, a Redux middleware function in `src/stores/index.ts` calls `getIpcApi().saveSnapshotsToStorage()` to write back to the user config file.

Normally, we curb the infinite loop by comparing the incoming payload in the `user-data-updated` event handler with the existing state before making any updates. However, sometimes the multiple CLI delete processes lead to a situation where the state and user config file infinitely alternate between states A and B.

This PR fixes the bug by adding the following safety measures:

- Use the `atomically` library to read and write the user config file in `./cli` (we already do this in the main app). This guarantees atomic writes, which could otherwise be an issue with multiple processes writing to the same file. 
- Use the `lockfile` module to prevent the `deleteSnapshotFromAppdata` function from running in parallel between two separate processes.
- Update the Redux listener that writes the snapshots array to the config file to only run in response to the `updateSnapshotLocally` action. The `setSnapshots` action is currently only used in response to the `user-data-updated` event, meaning we only run it when the user config file **has already changed**. There's no need to write back to the user config file if the Studio app didn't make changes to the snapshots state. 

## Testing Instructions

1. Run `npm start`
2. Create two preview sites
3. Rename one of the sites
4. Open the settings modal
5. Click on vertical ellipsis and delete all preview sites
6. Ensure that the operation completes successfully, without any glitches

## Pre-merge Checklist


cli/lib/snapshots.ts
```diff
@@ -20,6 +24,25 @@ import {
...
+	return new Promise< void >( ( resolve, reject ) => {
+		lockfile.lock( path, options, ( err ) => {
+			if ( err ) {
+				reject( err );
+			} else {
+				resolve();
+			}
+		} );
+	} );
+}
```
fredrikekelund, line 42: The type definitions for `util.promisify` don't play nice with functions that accept more than one argument.

sejas (approved):
Thanks for the quick PR. I tested it and I cannot reproduce the glitch anymore.

I'm still seeing some negative numbers for a brief period, but I think it's and independent low priority issue. After a few seconds, it display the 0 sites correctly.

Screenshot 2025-04-30 at 11 37 03

src/stores/snapshot-slice.ts
```diff
@@ -301,7 +301,10 @@ window.ipcListener.subscribe( 'user-data-updated', ( _, payload ) => {
 		const countDiff = snapshots.length - state.snapshot.snapshots.length;
 		store.dispatch(
 			wpcomApi.util.updateQueryData( 'getSnapshotUsage', undefined, ( data ) => {
-				data.siteCount += countDiff;
+				// There's a risk that more sites are deleted locally than the count returned by the
+				// API, because expired sites are preserved locally. Therefore, we need to ensure
+				// the count is non-negative.
+				data.siteCount = Math.max( 0, data.siteCount + countDiff );
```
sejas, line 307: ❤️

fredrikekelund: 👍

cli/lib/snapshots.ts
```diff
@@ -114,6 +137,7 @@ export async function saveSnapshotToAppdata(
 }
 
 export async function deleteSnapshotFromAppdata( snapshotUrl: string ): Promise< void > {
+	await lock( UPDATE_SNAPSHOTS_LOCKFILE_PATH, { wait: 1000 } );
```
bcotrim: I might be missing something here, but aren't we missing some error handling? In case of an exception or an early return, doesn't the lock stay in place?

fredrikekelund: The lockfile module states that:

> All known locks are removed when the process exits. Of course, it's possible for certain types of failures to cause this to fail, but a best effort is made to not be a litterbug.

Still, a `try .. finally` statement wouldn't hurt. I went the extra mile and added a `withLock` function in e09b871

PR #1302, opened 2025-04-30
Don't capture Wasm memory errors in Sentry

bgrgicak:
## Related issues

- [Fixes](https://github.com/Automattic/studio/pull/1224#issuecomment-2841143166)

## Proposed Changes

- Don't capture Wasm memory errors in Sentry

## Testing Instructions

- CI

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -415,17 +415,19 @@ export async function startServer(
 	try {
 		await server.start();
 	} catch ( error ) {
+		if (
+			error instanceof Error &&
+			error.message.includes( 'Cannot allocate Wasm memory for new instance' )
+		) {
+			throw new Error( 'WASM_ERROR_NOT_ENOUGH_MEMORY' );
+		}
+
 		Sentry.captureException( error );
 		if (
 			error instanceof Error &&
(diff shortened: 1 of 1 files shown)


src/ipc-handlers.ts
```diff
@@ -415,17 +415,19 @@ export async function startServer(
 	try {
 		await server.start();
 	} catch ( error ) {
+		if (
```
wojtekn, line 426: Would it make sense to add a comment shortly explaining that we don't need to track memory errors anymore, but we want to track unreachable instruction ones? Also, do we need to track the latter?

wojtekn (approved):

src/ipc-handlers.ts
bgrgicak: Good point, I documented it in [8c09798](https://github.com/Automattic/studio/pull/1302/commits/8c09798c84fac6ddb331cba35dd166d6a927174c).

I think that it's still good to collect unreachable instruction errors in Sentry. It's technically not a Studio issue, but it's a great source of feedback for Playground and a useful metric for us as we should aim to get this error to 0.

PR #1306, opened 2025-05-01
Fix assorted CLI CfT issues

fredrikekelund:
## Related issues

- Fixes STU-434
- Fixes STU-433

## Proposed Changes

- Don't persist `Validation successful` spinner message
- Display expired count in `preview list` command output
- Use default values for `newSites`, `sites`, and `snapshots` in `userDataSchema` to simplify code
- Update zod to newest patch version

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js preview list --path PATH` where `PATH` is the path to a Studio site with more than one preview site where at least one preview site is expired
3. Ensure that no `Validation successful` message is persisted in the terminal
4. Ensure that the output includes `(N expired)`

## Pre-merge Checklist

### DIFF
--- a/cli/lib/appdata.ts
+++ b/cli/lib/appdata.ts
@@ -20,9 +20,9 @@ const siteSchema = z
 
 const userDataSchema = z
 	.object( {
-		newSites: z.array( siteSchema ).optional(),
-		sites: z.array( siteSchema ).optional(),
-		snapshots: z.array( snapshotSchema ).optional(),
+		newSites: z.array( siteSchema ).default( () => [] ),
+		sites: z.array( siteSchema ).default( () => [] ),
+		snapshots: z.array( snapshotSchema ).default( () => [] ),
 		locale: z.string().optional(),
 		authToken: z
 			.object( {
@@ -66,9 +66,7 @@ export async function readAppdata(): Promise< UserData > {
 	try {
 		const fileContent = await readFile( appDataPath, { encoding: 'utf8' } );
 		const userData = JSON.parse( fileContent );
-		const result = userDataSchema.parse( userData );
-
-		return result;
+		return userDataSchema.parse( userData );
 	} catch ( error ) {
 		if ( error instanceof LoggerError ) {
 			throw error;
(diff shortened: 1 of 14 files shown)


cli/lib/appdata.ts
```diff
@@ -20,9 +20,9 @@ const siteSchema = z
 
 const userDataSchema = z
 	.object( {
-		newSites: z.array( siteSchema ).optional(),
-		sites: z.array( siteSchema ).optional(),
-		snapshots: z.array( snapshotSchema ).optional(),
+		newSites: z.array( siteSchema ).default( () => [] ),
+		sites: z.array( siteSchema ).default( () => [] ),
+		snapshots: z.array( snapshotSchema ).default( () => [] ),
```
fredrikekelund, line 25: This is no must-have, but it helps simplify the code elsewhere. It means that the appdata file will always contain empty array values for all these keys after the CLI touched it, but I didn't think that was really an issue.

cli/logger.ts
```diff
@@ -51,9 +50,11 @@ export class Logger< T extends string > {
 		this.spinner.text = message;
 	}
 
-	public reportSuccess( message: string ) {
+	public reportSuccess( message: string, shouldClearSpinner = false ) {
```
bcotrim, line 53: Nit: I think this logic could be a bit cleaner.
Could we maybe make the clear spinner decision based on the current action?
From what I could tell, all the cases for clear are on `LoggerAction.VALIDATE`

bcotrim (approved):
LGTM and works as described
image

One minor comment, but it shouldn't be a blocker in any way

cli/logger.ts
fredrikekelund: I actually opted for this approach initially. In short, I added a `shouldClearSpinnerForAction` argument to the `Logger` constructor. This argument accepted a function that mapped `T` to a boolean. This approach worked fine, I just found it kinda contrived compared to the current approach. 

Happy to discuss this further and potentially adjust in a follow-up PR.

PR #1309, opened 2025-05-01
User Settings: Add CTA for installing VSCode if none of supported editor is installed

gavande1:
> [!IMPORTANT]
> This PR has a base of https://github.com/Automattic/studio/pull/1259 until it gets to production. But its there, I will update the destination branch. 

## Related issues

- Fixes STU-436

## Proposed Changes

- Add a recommendation to download VS Code when no editor is installed. 
- Add tests for the editor picker component

## Testing Instructions

- Make the following changes to mimic uninstalled editors. 

```diff
diff --git a/src/ipc-handlers.ts b/src/ipc-handlers.ts
index e0e5308..e07afdb 100644
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -108,11 +108,11 @@ export async function getSiteDetails( _event: IpcMainInvokeEvent ): Promise< Sit
 
 export function getInstalledAppsAndTerminals(): InstalledApps & InstalledTerminals {
        return {
-               vscode: isInstalled( 'vscode' ),
-               phpstorm: isInstalled( 'phpstorm' ),
-               webstorm: isInstalled( 'webstorm' ),
-               windsurf: isInstalled( 'windsurf' ),
-               cursor: isInstalled( 'cursor' ),
+               vscode: false, // isInstalled( 'vscode' ),
+               phpstorm: false, // isInstalled( 'phpstorm' ),
+               webstorm: false, // isInstalled( 'webstorm' ),
+               windsurf: false, // isInstalled( 'windsurf' ),
+               cursor: false, // isInstalled( 'cursor' ),
                terminal: true, // Terminal.app is always available on macOS
                iterm: isInstalled( 'iterm' ),
                warp: isInstalled( 'warp' ),
```

- Make sure you can see the recommendation to download VS Code; the design is as per Figma.

CleanShot 2025-05-02 at 11 53 12@2x

## Pre-merge Checklist


src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -23,16 +24,43 @@ export const EditorPicker = ( { value, onChange, disabled }: EditorPickerProps )
...
+		return (
+			<p className="text-gray-500">
+				{ /* translators: "Visual Studio Code" is a trademarked brand name by Microsoft. Do not translate brand names. */ }
+				{ createInterpolateElement( __( 'We recommend using <a>Visual Studio Code ↗ </a>.' ), {
+					a: (
+						<Button
+							onClick={ () => getIpcApi().openURL( 'https://code.visualstudio.com/' ) }
+							variant="link"
+						/>
+					),
```
fredrikekelund:
```suggestion
					a: <VisualStudioCodeLink />,
```
Consider this an optional suggestion, but the `↗` struck me as more of an icon, that'd be better to keep out of the translation string. Following my suggestion and adding a `VisualStudioCodeLink` component like so worked fine in my quick testing:

```tsx
function VisualStudioCodeLink( props: PropsWithChildren ) {
	return (
		<Button
			onClick={ () => getIpcApi().openURL( 'https://code.visualstudio.com/' ) }
			variant="link"
		>
			{ props.children } ↗
		</Button>
	);
}
```

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -23,16 +24,43 @@ export const EditorPicker = ( { value, onChange, disabled }: EditorPickerProps )
...
+
 	return (
 		<SettingsFormField label={ __( 'Code editor' ) }>
 			<SelectControl
 				value={ value }
 				onChange={ ( newValue ) => onChange( newValue as SupportedEditor ) }
 				__nextHasNoMarginBottom
 				__next40pxDefaultSize
-				disabled={ disabled }
+				disabled={ ! value && ! hasEditorInstalled }
```
fredrikekelund, line 63: Disabling the control altogether when no supported editors are installed doesn't strike me as an obviously better user experience. The dropdown menu already has the `Not installed` optgroup, and it might be interesting to know which editors Studio supports. 

Granted, it's now clear that VS Code is supported, but I still don't really see it as an improvement to disable the control in this way.

src/modules/user-settings/components/tests/editor-picker.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+				phpstorm: false,
+				webstorm: false,
+				windsurf: false,
+				cursor: false,
+				terminal: false,
+				iterm: false,
+				warp: false,
+				ghostty: false,
+			} )
+		);
```
fredrikekelund:
```suggestion
		store.dispatch(
			installedAppsApi.util.updateQueryData( 'getInstalledApps', undefined, ( data ) => {
				return {
					vscode: false,
					phpstorm: false,
					webstorm: false,
					windsurf: false,
					cursor: false,
					terminal: false,
					iterm: false,
					warp: false,
					ghostty: false,
				};
			} )
		);
```
I commented on this in my review of https://github.com/Automattic/studio/pull/1259, too. RTK Query comes with an `updateQueryData` helper that works fine for these tests.

src/modules/user-settings/components/tests/editor-picker.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+// Create test actions for installedAppsApi
+const installedAppsTestActions = {
+	setInstalledApps: ( installedApps: InstalledAppsState ) => {
+		return { type: 'installedApps/setInstalledApps', payload: { installedApps } };
+	},
+};
+
+// Replace the store's reducer with our test reducer
+store.replaceReducer( installedAppsTestReducer as Reducer< RootState > );
```
fredrikekelund:
```suggestion
store.replaceReducer( testReducer );
```
By using the `updateQueryData` helper, we can do away with the `installedAppsTestReducer`.

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -11,10 +13,9 @@ import { SettingsFormField } from './settings-form-field';
 interface EditorPickerProps {
 	value: SupportedEditor | undefined;
 	onChange: ( value: SupportedEditor ) => void;
-	disabled?: boolean;
```
fredrikekelund: I believe the idea in https://github.com/Automattic/studio/pull/1259 was to disable this control while the editor data is loading (see https://github.com/Automattic/studio/pull/1259#discussion_r2071206344). I would advise against removing that prop.

fredrikekelund (commented):
Looking good and works as expected 👍 The only thing that'd be good to clear up before approving are my comments about the `disabled` prop.

src/modules/user-settings/components/editor-picker.tsx
gavande1: I might be wrong, though. But I don't see any reason to keep it. We used `disabled` to disable the dropdown and not show the loading state. We can derive the same state without needing the `disabled` prop. 

@gcsecsey What do you think about it?

katinthehatsite:
It seems that the PR https://github.com/Automattic/studio/pull/1259 was already merged, what do you think about rebasing on trunk so that the changes are slightly easier to review?

One thing that I am wondering about (I already mentioned this in Linear) is whether we might want to say something like `View the list of supported editors here` and link to our support documentation that mentions them instead of focusing on one specific code editor 🤔

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -1,43 +1,57 @@
...
-	);
+	const renderEditorRecommendation = () => {
+		if ( hasEditorInstalled ) {
+			return null;
+		}
+
+		return (
+			<p className="text-gray-500">
+				{ /* translators: "Visual Studio Code" is a trademarked brand name by Microsoft. Do not translate brand names. */ }
+				{ createInterpolateElement( __( 'We recommend using <a>Visual Studio Code</a>.' ), {
```
katinthehatsite, line 49: Should we drop the size of this to be slightly smaller and to be consistent with other modals - this is an example from `Add site` modal:

Screenshot 2025-05-06 at 3 16 08 PM

gavande1:
> It seems that the PR https://github.com/Automattic/studio/pull/1259 was already merged, what do you think about rebasing on trunk so that the changes are slightly easier to review?

Yes. that's the next step todo. 

> One thing that I am wondering about (I already mentioned this in Linear) is whether we might want to say something like View the list of supported editors here and link to our support documentation that mentions them instead of focusing on one specific code editor 🤔

I think that is an excellent idea. We should probably do that instead.

src/modules/user-settings/components/editor-picker.tsx
gcsecsey: Sorry @fredrikekelund for the confusion, I added the disabled prop in https://github.com/Automattic/studio/pull/1259/files#diff-ddc1a73fcb85252ce120a2916367500bc0f46db3976a1674dad6c81d9a6a7adbR33 as a placeholder for this PR, I was under the impression that the `disabled` prop sets the control to disabled.

@gavande1 I agree that we can remove it. 👍

gcsecsey:
> > One thing that I am wondering about (I already mentioned this in Linear) is whether we might want to say something like View the list of supported editors here and link to our support documentation that mentions them instead of focusing on one specific code editor 🤔
> 
> I think that is an excellent idea. We should probably do that instead.

Huge +1 to this, really good idea @katinthehatsite! 🎉

src/modules/user-settings/components/tests/editor-picker.test.tsx
gcsecsey: I carried out these changes in #1259 and now it's merged to trunk, so these changes should be picked up after a rebase.

src/modules/user-settings/components/tests/editor-picker.test.tsx
gcsecsey: I carried out these changes in https://github.com/Automattic/studio/pull/1259 and now it's merged to trunk, so these changes should be picked up after a rebase.

gavande1:
I am closing this PR in favour of the idea of providing a link to the support doc, which lists supported editors.

gavande1:
Let's reopen if required.

PR #1313, opened 2025-05-02
Only allow a single `startProxyServer` call at a time

fredrikekelund:
## Related issues

- Fixes STU-440

## Proposed Changes

`startProxyServer` is called in `SiteServer::start`. It's called every time a server with a custom domain config starts and returns early if the proxy server is already initiated. 

This can become an issue when the Studio app starts, because Studio will re-initiate the servers that were running when the app quit, and if there were multiple servers with custom domain configs, there will be multiple `startProxyServer` calls in rapid succession. This leads to errors like this:

This PR fixes the problem by adding a `sequential` helper in `src/lib/proxy-server.ts` that makes it so `startProxyServer` calls are executed sequentially.

## Testing Instructions

1. Apply the following diff

```diff
diff --git a/src/lib/proxy-server.ts b/src/lib/proxy-server.ts
index bf861da9..1efd425c 100644
--- a/src/lib/proxy-server.ts
+++ b/src/lib/proxy-server.ts
@@ -145,6 +145,7 @@ export async function checkIfPortIsFree( port: number ): Promise< boolean > {
  * This requires admin/root privileges
  */
 export const startProxyServer = sequential( async (): Promise< boolean > => {
+       console.log( 'startProxyServer' );
        try {
                // Start HTTP server if not already running
                if ( ! isHttpProxyRunning ) {
@@ -227,6 +228,8 @@ export const startProxyServer = sequential( async (): Promise< boolean > => {
                console.error( 'Failed to start proxy servers:', error );
 
                throw new Error( 'PROXY_ERROR_START_FAILED' );
+       } finally {
+               console.log( 'startProxyServer done' );
        }
 } );
```

2. Run `npm start`
3. Ensure that you have at least three sites with custom domains
4. Ensure that those sites are all started
5. Quit the app
6. Run `npm start`
7. Ensure that the app starts without any warnings
8. Looking at the terminal logs, ensure that the `startProxyServer` calls were executed sequentially (by looking at the output from the `console.log` calls you added)

## Pre-merge Checklist

### DIFF
--- a/src/lib/proxy-server.ts
+++ b/src/lib/proxy-server.ts
@@ -13,6 +13,22 @@ let httpsProxyServer: https.Server | null = null;
 let isHttpProxyRunning = false;
 let isHttpsProxyRunning = false;
 
+const locks = new Map< () => Promise< unknown >, Promise< unknown > >();
+
+// Ensures that only one instance of the function is running at a time
+function sequential< Args extends unknown[], Return >(
+	fn: ( ...args: Args ) => Promise< Return >
+) {
+	return async ( ...args: Args ) => {
+		const lock = locks.get( fn ) ?? Promise.resolve();
+		await Promise.allSettled( [ lock ] );
+		locks.delete( fn );
+		const promise = fn( ...args );
+		locks.set( fn, promise );
+		return await promise;
+	};
+}
+
 const proxy = httpProxy.createProxyServer();
 
 // Setup error handling for the proxy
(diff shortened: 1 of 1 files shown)


gavande1 (approved):
Thanks for the quick fix, @fredrikekelund. It's working as expected. Applying the patch, the error message wasn't shown. 

CleanShot 2025-05-02 at 15 05 34@2x

LTGM 👍

PR #1288, opened 2025-04-29
Refactor `PromptUsageProvider` as RTK Query hook

fredrikekelund:
## Related issues

- Fixes STU-416

## Proposed Changes

This quick refactor follows up on #826 by also moving `PromptUsageProvider` to `src/stores/wpcom-api.ts`. The prompt usage state is closely tied to the Assistant state, and it makes sense to keep them both in the Redux tree. 

## Testing Instructions

1. Open the Assistant tab
2. Send a message
3. Quickly open the settings modal
4. Ensure that the prompt usage meter is updated immediately when the AI response is received
5. Refresh the app and stay in the Overview tab
6. Open the settings modal
7. Ensure that the prompt usage meter displays the correct value now, too

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1318, opened 2025-05-05
Fix issue where snapshot bulk operations never finish

fredrikekelund:
## Related issues

- Fixes STU-443

## Proposed Changes

When deleting a site from Studio, we always dispatch a `deleteAllSnapshotsForSite` action. These "delete all snapshots" operations are handled by adding a "bulk operation" to the `operations` record in `snapshotSlice`. When all associated delete operations (for individual preview sites) have completed, we mark the bulk operation as completed, too. 

This PR addresses an issue where bulk operations are created even if the Studio site doesn't have any associated preview sites. The problem in these cases is that the bulk operation won't ever complete, because there are no associated delete operations for individual preview sites. 

I also took the opportunity to fix an incorrect change I made in https://github.com/Automattic/studio/pull/1288 where I increased the hard-coded default Assistant prompt limit from 200 to 1000. I've since seen that 200 is indeed the prompt limit for regular users. 

## Testing Instructions

1. Have at least two Studio sites. One of them does not have any preview sites.
2. Delete the Studio site without any preview sites from Studio.
3. Go to the `Previews` tab for any site.
4. The `Create preview site` button should be enabled and clickable.

## Pre-merge Checklist


src/stores/snapshot-slice.ts
```diff
@@ -195,6 +195,10 @@ const snapshotSlice = createSlice( {
 			.addMatcher(
 				isAnyOf( deleteAllSnapshotsForSite.fulfilled, deleteAllSnapshotsForUser.fulfilled ),
 				( state, action ) => {
+					if ( ! action.payload.operations.length ) {
+						return;
+					}
```
fredrikekelund, line 200: This is the key fix.

cli/lib/appdata.ts
```diff
@@ -168,9 +168,6 @@ export async function getOrCreateSiteByFolder(
 		}
 		const userData = await readAppdata();
 		site = getNewSitePartial( siteFolder );
-		if ( ! userData.newSites ) {
-			userData.newSites = [];
-		}
```
fredrikekelund, line 173: This is just following up my changes from https://github.com/Automattic/studio/pull/1306. `newSites` will always be an array now, so there's no need for this check.

src/stores/snapshot-slice.ts
```diff
@@ -355,10 +359,10 @@ function getAssociatedBulkOperation(
 	return [];
 }
 
-function isBulkOperationFulfilled( bulkOperation: BulkOperation ) {
+function isBulkOperationSettled( bulkOperation: BulkOperation ) {
 	return bulkOperation.operationIds.every( ( operationId ) => {
 		const operation = getOperation( operationId );
-		return operation?.status === 'fulfilled';
+		return operation && operation.status !== 'pending';
```
fredrikekelund, line 365: This is not directly related to the issue I observed, but it adds more safety and could potentially fix similar problems in the future.

src/constants.ts
```diff
@@ -11,7 +11,7 @@ export const MIN_WIDTH_SELECTOR_TO_MEASURE = `.${ MIN_WIDTH_CLASS_TO_MEASURE }`;
 export const SCREENSHOT_WIDTH = 1040;
 export const SCREENSHOT_HEIGHT = 1248;
 export const LIMIT_OF_ZIP_SITES_PER_USER = 10;
-export const LIMIT_OF_PROMPTS_PER_USER = 1000;
+export const LIMIT_OF_PROMPTS_PER_USER = 200;
```
fredrikekelund, line 14: This is not related to the main purpose of this PR, but it fixes an incorrect change I made in https://github.com/Automattic/studio/pull/1288. 200 is the prompt limit for regular Studio users.

sejas: Correct. Thanks for the fix.

sejas (approved):
I followed the testing steps and confirmed that before this PR, the creation of preview sites was unexpectedly disabled, and this PR fixes the issue.

**Before**

https://github.com/user-attachments/assets/4d9ce5b4-81c8-4f5f-8518-954816dea499

**After**

https://github.com/user-attachments/assets/22803d39-e447-49cb-8a11-b7a6beddf180

ivan-ottinger (commented):
Thank you for the fix, Fredrik. I can also confirm the proposed change fixes the issue and I cannot reproduce it anymore. The buttons does not get disabled anymore.

PR #1319, opened 2025-05-05
Appdata lockfile

fredrikekelund:
## Related issues

- Related to pesfPa-1tY-p2 and p1746194187596159-slack-C04GESRBWKW
- Fixes STU-500

## Proposed Changes

> [!IMPORTANT]  
> I'm currently seeking feedback on the implementation here. I'm aware that the tests are failing, but I don't want to work on fixing those before we decide on a direction for the implementation. The actual CLI and app code should work normally. 
> **Please remember to view the diff with whitespace changes hidden.** See [this comment](https://github.com/Automattic/studio/pull/1319#issuecomment-2894434924) for details. 

This PR implements a mandatory lockfile mechanism that's used when writing the Studio config file (`appdata-v1.json`) from the CLI or app code. It fixes race conditions that can lead to data loss, especially when the CLI and the app interact. 

This PR evolved from a simpler concept of using function arguments to `loadUserData` and `saveUserData` to acquire and release the lock, to a slightly larger refactor that makes acquiring the lock mandatory to write the config file. The current solution is a bigger change, but it's also safer. 

This PR now adds a higher-order function (`withUserDataWrite`) that wraps functions wanting to write to the Studio config file. `withUserDataWrite` does a couple of things:

- Acquires a lock
- Injects a parsed instance of `userData` as the first argument to the callback function
- Writes the return value of the callback to the Studio config file
- Releases the lock

In the CLI code, we use a slightly more advanced higher-order function that supports the callbacks returning tuples (arrays with a fixed numbers of members). The first member is a `userData` object that gets written to the Studio config file, the second member is returned to the function caller. This way, we get added flexibility.

## Testing Instructions

TBD

## Pre-merge Checklist


src/ipc-handlers.ts
```diff
@@ -153,7 +153,7 @@ export async function createSite(
 	enableHttps?: boolean,
 	siteId?: string
 ): Promise< SiteDetails[] > {
-	const userData = await loadUserData();
+	const userData = await loadUserData( true );
```
fredrikekelund: The `createSite` function can probably take more than 1000ms to execute, so maybe we need to hold the lockfile for longer in this case.

bcotrim (commented):
I've run some smoke tests on the app and everything seems to be working correctly.

I would suggest changing the current approach of passing a lock flag to something more like the withLock function. I think that approach would help guarantee a few things:
- Writes are always performed with the lock held
- Error handling is more robust, ensuring the file is always unlocked on exceptions or when the process completes

This could help reduce the risk of accidental misuse and make the locking logic easier to maintain.

For example:

```typescript
function withAppdataLock(fn) {
  return async (...args) => {
    await lock(LOCKFILE_PATH, { wait: 1000, stale: 1000 });
    try {
      const data = await readAppdata();
      const updated = await fn(data, ...args);
      await saveAppdata(updated);
      return updated;
    } finally {
      await unlock(LOCKFILE_PATH);
    }
  };
}
```

fredrikekelund:
> I would suggest changing the current approach of passing a lock flag to something more like the withLock function

I like your idea, @bcotrim, but I'm not convinced it's flexible enough for our existing codebase. 

The appealing aspects with your suggestion, as I see it, are:

- It guarantees that locks are held and released in `withAppdataLock` functions
- It looks clean. 

I've refactored `ipc-handlers.ts` to use that approach. For some functions, the difference from before is negligible. For other functions that previously both updated the Studio config file _and_ returned a different value, I had to come up with workarounds. In several cases, I think the end result was better than before (the `deleteSite` function returning an array of all remaining sites is perhaps not the best approach, IMO).

Still, there is one remaining case in `src` where the `withAppdataLock` approach would be more cumbersome: https://github.com/Automattic/studio/blob/4d0cf79b589f904113e8adf1bee7b34cbf06aed3/src/lib/windows-helpers.ts#L38-L64

For the `withAppdataLock` approach to be really effective, we shouldn't have to export the `saveUserData` function from `src/storage/user-data.ts`, and I've yet to think of a practical way to refactor `src/lib/windows-helpers.ts` to accommodate this approach. 

**More importantly**, the CLI code also has several functions that update the Studio config file _and_ return a different value. Here, for example: https://github.com/Automattic/studio/blob/4d0cf79b589f904113e8adf1bee7b34cbf06aed3/cli/lib/snapshots.ts#L39-L55

I don't think the `updateSnapshotInAppdata` function is really calling for a refactor for any other reason, and following the `withAppdataLock` approach introduces relatively cumbersome hoops that we need to jump through.

src/storage/user-data.ts
```diff
@@ -1,15 +1,60 @@
...
+			} else {
+				resolve();
+			}
+		} );
+	} );
+}
+
+type MaybePromise< T > = T | Promise< T >;
+
+export function withAppdataLock< Args extends unknown[] >(
```
bcotrim: Could we do something like this

```typescript
export function withAppdataLock< Args extends unknown[], ReturnType = void >(
	fn: ( userData: UserData, ...args: Args ) => MaybePromise< [ UserData, ReturnType ] >
```

This would allow for the flexibility we need, while keeping everything clean.
What do you think?

fredrikekelund:
I added an `updateAppdata` function to `src/storage/user-data.ts` to solve the problem in `src/lib/windows-helpers.ts`. That function writes only primitive values with a low chance of conflicts and can only be used to write known values (i.e. not values that are derived from the existing ones in the config file, like modifications to the `sites` array, for example.)

For the CLI code, I first took a shot at a generator-based approach where we `yield` updates to the config file. That felt overly complex, though, so I ended up implementing your suggestion, @bcotrim. `withAppdataWrite` callbacks can now return tuples, where the second value (if there is one) is passed back to the caller. I was hesitant at first, thinking this would not be easy to understand, but I've since come around, and I think it's OK now. 

I haven't touched the tests, so it's expected that they are still failing. The CLI and the app should (hopefully) be fully functional, though. 

I'd appreciate another round of feedback, @bcotrim!

fredrikekelund:
Tagging you for feedback, @nightnei, since Bernardo's AFK

fredrikekelund:
If the tuple pattern feels weird, we can also make the callbacks return a schema like so:

```ts
{
	userData?: UserData;
	return?: unknown;
}
```

src/hooks/use-site-details.tsx
```diff
@@ -227,7 +224,7 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 					customDomain,
 					enableHttps
 				);
-				const newSite = data.find( ( site ) => site.path === path );
+				const newSite = data.sites.find( ( site ) => site.path === path );
```
fredrikekelund: Not 100% happy about the `createSite` function returning the entire user config object… It might be worthwhile to update `withUserDataWrite` function to also support returning values to the caller (the same way that `withAppdataWrite` works).

cli/lib/utils.ts
```diff
@@ -1,7 +1,24 @@
...
+
 export function normalizeHostname( hostname: string ): string {
 	return hostname
 		.trim()
 		.toLowerCase()
 		.replace( /^https?:\/\//, '' )
 		.replace( /\/$/, '' );
 }
+
+export function lock( path: string, options: lockfile.Options ) {
```
nightnei: `lock` looks very general. what about `lockFileAsync` & `unlockFileAsync`? adding Async mostly due to distinguish between library name `lockfile` and court custom functions in codebase.

cli/lib/appdata.ts
```diff
@@ -111,6 +112,40 @@ export async function saveAppdata( userData: UserData ): Promise< void > {
 	}
 }
 
+const LOCKFILE_PATH = path.join( getAppdataDirectory(), 'appdata-v1.lock' );
```
nightnei: NIT but I would go with `appdata-v1.json.lock`, IMO it looks more clear that it lock specifically for `appdata-v1.json`.
```suggestion
const LOCKFILE_PATH = path.join( getAppdataDirectory(), 'appdata-v1.json.lock' );
```

nightnei (commented):
I was reviewing, and more and more I was wondering - would it be simpler if we don't introduce HOF and make fewer changes with just modifying `saveUserData`? Since now code looks much more complicated, especially with nuances like [updateAppdata](https://github.com/Automattic/studio/pull/1319/files#diff-2a78c3d26284d759ad9ec39370822d96d2dcbf38c383487f49c15f67192682daR180)

And yeah, tuple pattern looks weird, I think idea with object is great 👍

fredrikekelund:
> I was reviewing, and more and more I was wondering - would it be simpler if we don't introduce HOF and make fewer changes with just modifying saveUserData?

It would indeed be simpler, but it would also be less secure, because there's no guarantee that we obtain a look before writing the config file. 

This is the solution I explored originally, though. If you want to see what that looked like, you can go to https://github.com/Automattic/studio/pull/1319/files/559b01c9717d46ea236666bec13583e2b7865ae8

nightnei:
I reviewed the old approach and IMO it was cool. Of course, providing `true` everywhere looked really not nice and just duplicating. But I think that previous solution could be improved with moving `loadUserData` inside `saveUserData` and additionally we could pass callback as an argument to `saveUserData` and we would kill two birds with one stone. Actually, it's just mixing the latest changes in this PR with those changes before HOF:


~const data = await loadUserData( true );~
~data.devToolsOpen = false;~
~await saveUserData( data, true );~
```
await saveUserData( currentData => ({
	 ... currentData,
	 devToolsOpen: false
 }));
 
 const saveUserData = (cb) => {
 	lock;
 	data = read;
 	cb(data);
 	unlock;
 }
```

As a result:
1. We don't do big refactoring
2. YOLO members are already used to the current behaviour of `saveUserData`, so the behaviour literally is not changed 
3. Code is easier to read
4. And last but not least - we are considering moving to real DB or something like that in the future, so in the future I believe that it will be easier to move to DB with this approach, since we will just adjust encapsulated saveUserData with replacing file lock mechanism to DB locking

HOF is really nice idea, but IMO not in this case, it looks overcomplicated.

WDYT?

fredrikekelund:
> I think that previous solution could be improved with moving loadUserData inside saveUserData and additionally we could pass callback as an argument to saveUserData and we would kill two birds with one stone

Plenty of functions modify the return value of `loadUserData` in complex enough ways that the majority of the function body would consist of the callback to `saveUserData` in this case. I am convinced that this approach would be more cumbersome to work with.

I'm open to reverting to the original solution for increased simplicity if multiple team members feel this is the right thing to do, but we should understand that this removes the static analysis benefits of the current approach (`loadUserData` doesn't know if the lock should be acquired, and `saveUserData` doesn't know if the lock is acquired). With the old approach, we sacrifice correctness for a more simplistic change. 

> We don't do big refactoring

This PR does indeed touch a lot of lines, but hiding whitespace (see screenshot below) when viewing the diff reveals that relatively little logic has actually changed.

Screenshot 2025-05-20 at 15 10 32

> YOLO members are already used to the current behaviour of saveUserData, so the behaviour literally is not changed

The HOF pattern is uncommon—especially when injecting arguments—but it's not complex to the point of being difficult to understand or reason about. 

> Code is easier to read

[My original solution](https://github.com/Automattic/studio/pull/1319/files/559b01c9717d46ea236666bec13583e2b7865ae8) is indeed easier to read. However, see my previous points about tradeoffs and how to view the diff.

> we are considering moving to real DB or something like that in the future, so in the future I believe that it will be easier to move to DB with this approach, since we will just adjust encapsulated saveUserData with replacing file lock mechanism to DB locking

We might move to a per-site config architecture and/or adopt a SQLite database for the config. Whichever way we go, we'll need to make changes in many places again. I don't believe any of the solutions proposed here meaningfully pave the way for splitting the config file and/or moving to SQLite.

fredrikekelund:
@bcotrim @wojtekn, I'd appreciate a second review of the suggested approach here to confirm whether I should proceed with it or not. 

Per the PR description, please remember to view the diff with whitespace changes hidden. See https://github.com/Automattic/studio/pull/1319#issuecomment-2894434924 for details.

bcotrim:
> @bcotrim @wojtekn, I'd appreciate a second review of the suggested approach here to confirm whether I should proceed with it or not.

I understand that the HOF could make the code harder to read, so I have another suggestion:
- We could create and use an ESLint rule to ensure calls to `updateAppdata` always have a `lock` call before.

As I never did something like this, I asked Cursor to implement a rule for that use case and it worked!

image 

I would still prefer the HOF approach as it keeps the locking logic encapsulated and reduces the chance of forgetting to add the lock call, but it's a good middle ground if the team agrees to keep the code easier to read.

fredrikekelund:
The thought of using an ESLint rule struck me, too. Your screenshot looks promising, @bcotrim. Thinking out loud: if we choose a linter-based approach, the rules should go like this:

`saveUserData` and `updateAppdata` can only be called if:
1. We update with data that's not derived from the existing file (i.e., we don't modify the `sites` array, for example), OR
2. We need to have acquired a lock first

This would indeed lead to simpler code, even if I'm unsure how difficult it would be to encode those exact rules in ESLint. @bcotrim, could you share the code behind your experiment? Maybe branch off this PR.

bcotrim:
> The thought of using an ESLint rule struck me, too. Your screenshot looks promising, @bcotrim. Thinking out loud: if we choose a linter-based approach, the rules should go like this:
> 
> `saveUserData` and `updateAppdata` can only be called if:
> 
>     1. We update with data that's not derived from the existing file (i.e., we don't modify the `sites` array, for example), OR
> 
>     2. We need to have acquired a lock first
> 
> 
> This would indeed lead to simpler code, even if I'm unsure how difficult it would be to encode those exact rules in ESLint. @bcotrim, could you share the code behind your experiment? Maybe branch off this PR.

I've cleaned the code a bit and adjusted the condition to the behavior you mentioned.
https://github.com/Automattic/studio/pull/1430

fredrikekelund:
I worked on fixing the tests for this PR, but realized that there's a fundamental challenge in the proposed architecture: if a `withAppdataWrite` function calls another `withAppdataWrite` function, the lock mechanism does not work as intended. In these cases, the second callback waits until the lock of the first callback has been released, and the first callback unlocks the lock after completing, despite no longer holding the lock (in the worst case, another function could already have acquired the lock).

I propose taking a step back and going with the simpler architecture that I originally suggested (see https://github.com/Automattic/studio/pull/1319/files/559b01c9717d46ea236666bec13583e2b7865ae8). This would be a pragmatic way of fixing the issue at hand without spending too much more time on this. 

We can potentially improve that solution later by adding an ESLint rule. I think manual code reviews would go a long way, though.

nightnei:
> I propose taking a step back and going with the simpler architecture that I originally suggested (see [Automattic/studio/pull/1319/files/559b01c9717d46ea236666bec13583e2b7865ae8](https://github.com/Automattic/studio/pull/1319/files/559b01c9717d46ea236666bec13583e2b7865ae8)). This would be a pragmatic way of fixing the issue at hand without spending too much more time on this.

Let's do it.
And taking into account that it's a temporary solution (AFAIU), then I think we should not worry a lot about making this solution perfect.  Just keep it simple and clear.

bcotrim:
@nightnei @fredrikekelund This PR should be ready for review. As we discussed I changed this back to a simpler approach without the HOF.
I would still recommend reviewing/merging this one first https://github.com/Automattic/studio/pull/1430 as the rule really makes it complete, to enforce the lock/unlock mechanism.

Let me know if you have any comments or feedback.

nightnei:
> This PR should be ready for review. As we discussed I changed this back to a simpler approach without the HOF.

@bcotrim The current code LGTM

> I would still recommend reviewing/merging this one first https://github.com/Automattic/studio/pull/1430 as the rule really makes it complete, to enforce the lock/unlock mechanism.

Let me try to update the code with my [suggestion above](https://github.com/Automattic/studio/pull/1319#issuecomment-2893741600). If it works well, then we don't need ESLint rule and code will be even more simple and w/o duplications of calling `loadUserData`

nightnei:
> Let me try to update the code with my https://github.com/Automattic/studio/pull/1319#issuecomment-2893741600. If it works well, then we don't need ESLint rule and code will be even more simple and w/o duplications of calling loadUserData

~So, the idea is [this](https://github.com/Automattic/studio/pull/1443) - to avoid copy-paste, avoid the eslint rule and make a reliable solution. I see that updating it in some places will not be very easy, so we can try to proceed with changes in my PR and update all calls of `saveUserData` or we can proceed with current changes in this PR as the first step and then we will see - should we add eslint rule or maybe we will soon consider using DB or something else.~

~Folks, what are your thoughts about it?~

UPDATE: I am ok with the reverted latest changes in this PR 👍
Regarding ESLint rule - it also makes sense to make these changes complete, so if Fredrik also ok with it, then let's merge, do final review/test and let's merge it.

cli/lib/appdata.ts
```diff
@@ -163,20 +173,29 @@ export function getNewSitePartial( siteFolder: string ): z.infer< typeof siteSch
 	return newSite;
 }
 
-export async function getOrCreateSiteByFolder(
+const createNewSite = async ( siteFolder: string ): Promise< z.infer< typeof siteSchema > > => {
```
```suggestion
const createNewSite = async ( siteFolder: string ): Promise< SiteData > => {
```
Nit, but there's a dedicated type for this now.

cli/lib/appdata.ts
```diff
@@ -163,20 +173,29 @@ export function getNewSitePartial( siteFolder: string ): z.infer< typeof siteSch
...
+	} finally {
+		await unlockAppdata();
+	}
+};
+
+export const getOrCreateSiteByFolder = async (
 	siteFolder: string
-): Promise< z.infer< typeof siteSchema > > {
-	let site;
+): Promise< z.infer< typeof siteSchema > > => {
```
```suggestion
): Promise< SiteData > => {
```

cli/lib/appdata.ts
```diff
@@ -163,20 +173,29 @@ export function getNewSitePartial( siteFolder: string ): z.infer< typeof siteSch
...
 	try {
-		site = await getSiteByFolder( siteFolder );
+		const site = await getSiteByFolder( siteFolder );
+		return site;
```
```suggestion
		return await getSiteByFolder( siteFolder );
```
Nit, suggestion

cli/lib/tests/appdata.test.ts
```diff
@@ -106,54 +105,6 @@ describe( 'Appdata Module', () => {
...
-			};
-
-			await saveAppdata( mockUserData );
-
-			expect( writeFile ).toHaveBeenCalled();
-			const savedData = JSON.parse( ( writeFile as jest.Mock ).mock.calls[ 0 ][ 1 ] );
-			expect( savedData.version ).toBe( 1 );
-		} );
-	} );
-
```
fredrikekelund: I believe we could bring back these tests, no? I removed them when we stopped exporting `saveAppdata`

common/lib/lockfile.ts
```diff
@@ -0,0 +1,25 @@
+import lockfile from 'lockfile';
+
+export function lockFileAsync( LOCKFILE_PATH: string, options: lockfile.Options ) {
```
fredrikekelund:
```suggestion
export function lockFileAsync( lockfilePath: string, options: lockfile.Options ) {
```
Let's not capitalize this, since this is no longer a constant.

common/lib/lockfile.ts
```diff
@@ -0,0 +1,25 @@
...
+			if ( err ) {
+				reject( err );
+			} else {
+				resolve();
+			}
+		} );
+	} );
+}
+
+export function unlockFileAsync( LOCKFILE_PATH: string ) {
```
fredrikekelund:
```suggestion
export function unlockFileAsync( lockfilePath: string ) {
```

src/hooks/use-site-details.tsx
fredrikekelund: Let's change the `createSite` IPC function to only return the newly created site.

src/ipc-handlers.ts
```diff
@@ -174,124 +179,132 @@ export async function createSite(
...
 
-	userData.sites.push( server.details );
-	sortSites( userData.sites );
-	await saveUserData( userData );
+		userData.sites.push( server.details );
+		sortSites( userData.sites );
 
-	return mergeSiteDetailsWithRunningDetails( userData.sites );
+		await saveUserData( userData );
+		return userData;
```
```suggestion
		return server.details;
```
Per my other comment, let's return just the newly created site here.

src/ipc-handlers.ts
```diff
@@ -300,91 +313,98 @@ export async function disconnectWpcomSites(
...
+		}
 
-	const connections = userData.connectedWpcomSites?.[ currentUserId ] || [];
+		const connections = userData.connectedWpcomSites?.[ currentUserId ] || [];
 
-	if ( ! connections.length ) {
-		return;
-	}
+		if ( ! connections.length ) {
+			return userData;
```
```suggestion
			return;
```
This function should return `void` now.

src/ipc-handlers.ts
```diff
@@ -682,26 +694,29 @@ export function removeTemporalFile( event: IpcMainInvokeEvent, path: string ) {
...
+			// Move files to trash
+			if ( deleteFiles ) {
+				await shell.trashItem( server.details.path );
+			}
+		} catch ( error ) {
+			/* We want to exit gracefully if the there is an error deleting the site files */
+			Sentry.captureException( error );
+		}
+		const newSites = userData.sites.filter( ( site ) => site.id !== id );
+		return { ...userData, sites: newSites };
```
```suggestion
```
We call `getIpcApi().getSiteDetails()` in the renderer process now to query the sites after this function completes, so let's remove the return value.

src/ipc-handlers.ts
```diff
@@ -1357,13 +1355,24 @@ export async function deleteSnapshot(
...
-	await saveUserData( {
-		...userData,
-		newSites: userData.newSites?.filter( ( s ) => s.id !== newSite.id ),
-	} );
+		const newSites = userData.newSites?.filter( ( s ) => s.id !== newSite.id );
+		await saveUserData( { ...updatedUserData, newSites } );
+	} finally {
+		await unlockAppdata();
+	}
 }
```
fredrikekelund: There's a logic problem here similar to what I described on Monday https://github.com/Automattic/studio/pull/1319#issuecomment-2909711610:
1. We acquire a lock
2. We call `createSite`
3. `createSite` acquires a lock (it waits for the existing lock to go stale after 1000ms)
4. `createSite` returns
5. We update the Studio config file without a lock
6. We release the lock (despite no longer actually holding it)

We could fix this by acquiring the lock after `createSite` returns, and then calling `loadUserData` again to get a fresh copy of the config file.

src/lib/bump-stats/index.ts
```diff
@@ -36,6 +36,8 @@ export function bumpAggregatedUniqueStat(
 				return false;
 			}
 
+			console.log( 'bumpStat', group, stat, bumpInDev );
+
```
fredrikekelund:
```suggestion
```

src/storage/user-data.ts
```diff
@@ -101,18 +103,45 @@ export async function loadUserData(): Promise< UserData > {
...
 	const filePath = getUserDataFilePath();
-
 	const asString = JSON.stringify( toDiskFormat( data ), null, 2 ) + '\n';
+	await writeFile( filePath, asString, 'utf-8' );
+}
+
+const LOCKFILE_PATH = nodePath.join( getResourcesPath(), LOCKFILE_NAME );
+
+export async function lockAppdata() {
+	return lockFileAsync( LOCKFILE_PATH, { wait: 1000, stale: 1000 } );
```
fredrikekelund: I'm not sure what the right configuration is, but I have a feeling that we should be more strict with this configuration. 

Firstly, I believe it's redundant to specify the `wait` option if we also have `stale`. The current configuration is saying "proceed anyway after 1000ms" AND "throw an error if the lock isn't released within 1000ms". AFAICT, the lockfile module ignores the `wait` option when `stale` is also specified. 

I don't think our implementation is robust enough to use the `wait` option exclusively. I recommend proceeding by removing `wait` and increasing the value of `stale`.

src/ipc-handlers.ts
```diff
@@ -174,124 +179,132 @@ export async function createSite(
...
 
-	const parentWindow = BrowserWindow.fromWebContents( event.sender );
-	sendIpcEventToRendererWithWindow( parentWindow, 'theme-details-updating', { id: details.id } );
+		const parentWindow = BrowserWindow.fromWebContents( event.sender );
+		sendIpcEventToRendererWithWindow( parentWindow, 'theme-details-updating', { id: details.id } );
 
-	userData.sites.push( server.details );
-	sortSites( userData.sites );
-	await saveUserData( userData );
+		userData.sites.push( server.details );
```
fredrikekelund, line 241:
```suggestion
		await lockAppdata();
		const userData = await loadUserData();
		userData.sites.push( server.details );
```
I recommend acquiring the lock here. `SiteServer.create` can take a long time to resolve. Holding the lock for such a long time is risky, since it may go stale.

fredrikekelund (commented):
This is looking really good overall 👍 I feel confident that this is the right path forward. 

I left some comments with varying degrees of importance. On a more birds-eye-view note, I also recommend adding something like this to `cli/index.ts`:

```ts
process.on( 'uncaughtExceptionMonitor', async () => {
	await unlockAppdata();
} );
```

We should do the same for the Studio app and add an equivalent handler in `src/index.ts`.

bcotrim:
>  On a more birds-eye-view note, I also recommend adding something like this to `cli/index.ts`:
> 
> ```ts
> process.on( 'uncaughtExceptionMonitor', async () => {
> 	await unlockAppdata();
> } );
> ```
> 
> We should do the same for the Studio app and add an equivalent handler in `src/index.ts`.

We always wrap the unlock call in a `finally` and are using the `stale` option from `lockfile` so I don't think this is necessary. Is there any scenario I'm missing?

fredrikekelund:
> Is there any scenario I'm missing?

I was thinking of a scenario where an unexpected exception is thrown from somewhere other than the function holding the lock. This could theoretically happen if we're juggling simultaneous async operations.

src/ipc-handlers.ts
```diff
@@ -174,124 +179,134 @@ export async function createSite(
 		// Form validation should've prevented a non-empty directory from being selected
 		throw new Error( 'The selected directory is not empty nor an existing WordPress site.' );
 	}
+	try {
```
fredrikekelund: Let's move the opening `try {` clause closer to the `lockAppdata` call. If an error is thrown before we acquire the lock, that could lead to incorrect behavior.

src/ipc-handlers.ts
```diff
@@ -174,124 +179,134 @@ export async function createSite(
...
-export async function updateSite(
-	event: IpcMainInvokeEvent,
-	updatedSite: SiteDetails
-): Promise< SiteDetails[] > {
-	const userData = await loadUserData();
-	const updatedSites = userData.sites.map( ( site ) =>
-		site.id === updatedSite.id ? updatedSite : site
-	);
-	userData.sites = updatedSites;
+export async function updateSite( event: IpcMainInvokeEvent, updatedSite: SiteDetails ) {
```
```suggestion
export async function updateSite( event: IpcMainInvokeEvent, updatedSite: SiteDetails ): Promise< void > {
```
This is most definitely optional, but I find it helpful to add the explicit return types back here to clarify the intention. I removed the return types to accommodate the `withUserDataWrite` type definition.

src/ipc-handlers.ts
```diff
@@ -161,8 +167,7 @@ export async function createSite(
 	customDomain?: string,
 	enableHttps?: boolean,
 	siteId?: string
-): Promise< SiteDetails[] > {
-	const userData = await loadUserData();
+) {
```
```suggestion
): Promise< SiteDetails > {
```
Per my other comment about return types.

src/lib/oauth.ts
```diff
@@ -34,15 +34,7 @@ async function getToken(): Promise< StoredToken | null > {
...
-		const userData = await loadUserData();
-		await saveUserData( {
-			...userData,
-			authToken: token,
-		} );
-	} catch ( error ) {
-		console.error( 'Failed to store token', error );
-	}
+	await updateAppdata( { authToken: token } );
 }
```
fredrikekelund: Given how succinct this function is now and that it isn't exported, I'd consider inlining it in `onOpenUrlCallback`

src/lib/oauth.ts
```diff
@@ -51,15 +43,7 @@ export function getSignUpUrl( locale: SupportedLocale ) {
...
-		const userData = await loadUserData();
-		await saveUserData( {
-			...userData,
-			authToken: undefined,
-		} );
-	} catch ( error ) {
-		return;
-	}
+	await updateAppdata( { authToken: undefined } );
 }
```
fredrikekelund: Optional, but we could also inline this in the `clearAuthenticationToken` function in `ipc-handlers`

src/storage/tests/user-data.test.ts
```diff
@@ -69,60 +55,18 @@ platformTestSuite( 'User data', () => {
...
-							themeDetails: defaultThemeDetails,
-						} ) ),
-						snapshots: [],
-					},
-					null,
-					2
-				) + '\n',
-				'utf-8'
-			);
-		} );
```
fredrikekelund: We could bring back this test case, too. Not critical, though.

cli/lib/tests/appdata.test.ts
fredrikekelund, line 1: The changes to this file do not _need_ to be reverted, but the tests still pass without them.

fredrikekelund:
I realize that the `uncaughtExceptionMonitor` could potentially lead to issues where the current process unlocks the lockfile despite not holding the lock to begin with… Maybe it's better to avoid it for the time being. The other changes here still go a very long way towards fixing the concurrent write problem.

fredrikekelund (commented):
Moving the `try {` clause in the `createSite` function is the only relatively high-priority change I see remaining. Aside from that, this PR LGTM, and I've confirmed that it fixes the issue I reported to begin with 👍

Apparently, I can't approve the PR because I'm also the PR author, but feel free to add your own approval in my stead, @bcotrim

bcotrim (approved):

cli/lib/appdata.ts
```diff
@@ -163,20 +173,26 @@ export function getNewSitePartial( siteFolder: string ): z.infer< typeof siteSch
 	return newSite;
 }
 
-export async function getOrCreateSiteByFolder(
-	siteFolder: string
-): Promise< z.infer< typeof siteSchema > > {
-	let site;
+const createNewSite = async ( siteFolder: string ): Promise< SiteData > => {
```
youknowriad, line 176: I have a few questions here:

 - This function creates sites right? And it's called in CLI, do we currently create sites in the CLI (not preview sites), normal local sites. If I'm following the flow, on each preview command, this is supposed to create a new empty site right if non existant? Is it something else?

 - The second remark is that this function is inconsistent with the Electron App. The site type in the electron app has the following properties as mandatory: id, path, name, port, phpVersion while this one only has id, path and name. So is this a bug? Why are we missing these mandatory properties here, or do you think it's ok to update the "app" types to make port and phpVersion as optional as well?

youknowriad: Maybe @bcotrim as @fredrikekelund is AFK

bcotrim: @youknowriad 

> This function creates sites right? And it's called in CLI, do we currently create sites in the CLI (not preview sites), normal local sites. If I'm following the flow, on each preview command, this is supposed to create a new empty site right if non existant? Is it something else?

Exactly, this function will add an entry to `newSites` in appdata. When Studio boots up, it will automatically pick up on the list of sites and create them using Studio's workflow.
Yes, it only happens when a preview site is created for a non existing site in Studio.

> The second remark is that this function is inconsistent with the Electron App. The site type in the electron app has the following properties as mandatory: id, path, name, port, phpVersion while this one only has id, path and name. So is this a bug? Why are we missing these mandatory properties here, or do you think it's ok to update the "app" types to make port and phpVersion as optional as well?

It has the minimal necessary data to create a site in Studio. You can check the handler here: https://github.com/Automattic/studio/blob/trunk/src/ipc-handlers.ts#L1413

youknowriad: So does it mean that we need to update the `SiteDetails` type to make these extra properties as "optional". Or do we have another "type" that represents what can get saved to appData?

bcotrim: We already have a separate type for this exact purpose. There's a NewSiteDetails type defined as:

```typescript
type NewSiteDetails = Pick< SiteDetails, 'id' | 'path' | 'name' >;
```

This is the minimal data needed to create a site in Studio. The CLI's SiteData type only includes the required fields for site creation. 
The additional fields like port and phpVersion are assigned by Studio during the actual site creation process in the handler.

So the types are intentionally different:
- NewSiteDetails / CLI's SiteData: Minimal data for site creation (id, path, name)
- SiteDetails (StoppedSiteDetails): Full site information including port, phpVersion, etc.

The CLI doesn't need to know about ports or PHP versions since Studio handles that automatically when it processes the newSites array on startup.

youknowriad: >The CLI doesn't need to know about ports or PHP versions since Studio handles that automatically when it processes the newSites array on startup.

For now :) but if we are to make the CLI create and launch sites independently from the App, it should. That said, no worries. The above answer my question. 

So would it be ok to say that the entries in app data under `newSites` should be of type `NewSiteDetails` and the entries under `sites` should be of type `SiteDetails` ?

bcotrim: Assuming this discussion is around the CLI project, I think we should disregard the `newSites` entirely.
If the CLI is able to create sites, we no longer need that workaround in the CLI `preview` command.

PR #1325, opened 2025-05-05
SSL Support: Refactor certificate trust selector

gavande1:
## Related issues

- Fixes https://github.com/Automattic/studio/pull/1117#pullrequestreview-2814292909

## Proposed Changes

- Refactor certificate trust select to use Redux Store

## Testing Instructions

- Follow https://github.com/Automattic/studio/pull/1117

## Pre-merge Checklist


src/hooks/use-certificate-trust.ts
```diff
@@ -1,35 +1,27 @@
...
+import { checkCertificateTrust, selectIsRootCATrusted } from 'src/store/slices/certificate-trust';
+import { useAppDispatch, useRootSelector } from 'src/stores';
 
 /**
  * Custom hook that checks if the Studio CA certificate is trusted on the system
  * @returns A boolean indicating if the certificate is trusted
  */
 export function useCertificateTrust(): boolean {
-	const [ isTrusted, setIsTrusted ] = useState< boolean >( false );
+	const dispatch = useAppDispatch();
```
gavande1: @fredrikekelund, I have been thinking about your comment, and I was also unhappy with how `useCertificateTrust()` was implemented. It was creating a flicker issue with showing https help text when changing sites, so I implemented it using the Redux store.

This is the first time I have implemented it. Let me know what you think about it.

src/store/slices/certificate-trust.ts
fredrikekelund, line 1: Much like https://github.com/Automattic/studio/pull/1259, I think Redux Toolkit's `createApi` API would be a better fit here. It gives us react-query-like hooks to be used in the components, handling fetching and caching internally. The typical use case is fetching something over the network, but IPC communication turns out to be a great fit, too. 

See `src/stores/installed-apps-api.ts` in https://github.com/Automattic/studio/pull/1259 for inspiration on what that might look like.

fredrikekelund, line 1: The path to this file should be `src/stores/slices/certificate-trust.ts` (`stores` instead of `store`).

src/hooks/use-certificate-trust.ts
fredrikekelund: Just to clarify, my comment https://github.com/Automattic/studio/pull/1117#discussion_r2073151154 was not meant to say that the current approach was wrong per se (though I generally support refactoring this using Redux Toolkit). I mostly wanted to point out that using an `isMounted` ref might not be such a bad idea.

fredrikekelund (commented):
I support refactoring the "is trusted check" to be part of the Redux state tree 👍 I suggest we look into using the `createApi` API instead, though ([docs](https://redux-toolkit.js.org/rtk-query/api/createApi)).

As I already mentioned, Gergely's https://github.com/Automattic/studio/pull/1259 PR should offer up some good inspiration.

src/hooks/use-certificate-trust.ts
gavande1: I agree. I also don't mean to say that the previous approach was bad; I just wasn't happy with the flicker issue.

src/store/slices/certificate-trust.ts
gavande1: Thanks for the suggestion. Will look into it. 👍

gavande1: I removed the file eventually.

src/hooks/use-certificate-trust.ts
```diff
@@ -1,35 +1,22 @@
...
+import { useCheckCertificateTrustQuery } from 'src/stores/certificate-trust-api';
 
 /**
  * Custom hook that checks if the Studio CA certificate is trusted on the system
  * @returns A boolean indicating if the certificate is trusted
  */
 export function useCertificateTrust(): boolean {
-	const [ isTrusted, setIsTrusted ] = useState< boolean >( false );
+	const { data: isTrusted = false, refetch: checkCertificateTrust } =
+		useCheckCertificateTrustQuery();
```
fredrikekelund:
```suggestion
	const { data: isTrusted = false } = useCheckCertificateTrustQuery( undefined, {
		refetchOnFocus: true,
	} );
```
I suggest doing away with this file and replacing the `useCertificateTrust` calls with this ☝️ code.

src/stores/certificate-trust-api.ts
```diff
@@ -0,0 +1,25 @@
...
+					return { data: false };
+				}
+			},
+			providesTags: [ 'CertificateTrust' ],
+		} ),
+	} ),
+} );
+
+export const { useCheckCertificateTrustQuery } = certificateTrustApi;
+export default certificateTrustApi.reducer;
```
fredrikekelund:
```suggestion
```
No need to export the reducer.

fredrikekelund (commented):
This is shaping up nicely 👍 I left a couple of quick suggestions.

src/stores/certificate-trust-api.ts
gavande1: Removed!

src/hooks/use-certificate-trust.ts
gavande1: Was able to get rid of the file.

src/stores/certificate-trust-api.ts
```diff
@@ -0,0 +1,24 @@
...
+	endpoints: ( builder ) => ( {
+		checkCertificateTrust: builder.query< boolean, void >( {
+			queryFn: async () => {
+				try {
+					const isTrusted = await getIpcApi().isCATrusted();
+					return { data: isTrusted };
+				} catch ( error ) {
+					console.error( 'Failed to check certificate trust:', error );
+					return { data: false };
+				}
```
fredrikekelund:
```suggestion
				try {
					const isTrusted = await getIpcApi().isCATrusted();
					return { data: isTrusted };
				} catch ( error ) {
					return { error: { status: 500, data: error } };
				}
```
In #1259, I advised removing `catch` handlers inside `queryFn` functions, but I realize now that this ☝️ is more in line with what RTK Query expects.

src/stores/certificate-trust-api.ts
```diff
@@ -0,0 +1,24 @@
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react';
+import { getIpcApi } from 'src/lib/get-ipc-api';
+
+export const certificateTrustApi = createApi( {
+	reducerPath: 'certificateTrustApi',
+	baseQuery: fetchBaseQuery(),
+	tagTypes: [ 'CertificateTrust' ],
+	endpoints: ( builder ) => ( {
```
fredrikekelund, line 9:
```suggestion
	refetchOnFocus: true,
	endpoints: ( builder ) => ( {
```
I realize now that we can configure this at the global level and avoid some repetition in the components.

fredrikekelund (commented):
A couple of final tweaks and this should be good to go 👍

We need to add the following to `src/stores/index.ts` to make the `refetchOnFocus` option work:

```ts
import { setupListeners } from '@reduxjs/toolkit/query';

setupListeners( store.dispatch );
```

src/stores/certificate-trust-api.ts
gavande1: I have added it and it works like a magic. 👍

src/stores/certificate-trust-api.ts
```diff
@@ -0,0 +1,29 @@
...
+				try {
+					const isTrusted = await getIpcApi().isCATrusted();
+					return { data: isTrusted };
+				} catch ( error ) {
+					console.error( 'Failed to check certificate trust:', error );
+					const errorMessage =
+						error instanceof Error ? error.message : 'Failed to check certificate trust';
+					return {
+						error: { data: false, error: errorMessage } as FetchBaseQueryError,
+					};
```
gavande1, line 21: @fredrikekelund, Returning error as you suggested was throwing an error in the console. RTK expects errors in current format.

src/stores/certificate-trust-api.ts
gavande1: Replied in https://github.com/Automattic/studio/pull/1325#discussion_r2077448826

src/stores/certificate-trust-api.ts
fredrikekelund:
```suggestion
					return {
						error: { status: 500, data: errorMessage },
					};
```
This seems to match the `FetchBaseQueryError` type.

fredrikekelund (approved):
LGTM with a small comment about the structure of the error object returned by `checkCertificateTrust` 👍

src/stores/certificate-trust-api.ts
gavande1: Thanks! Updated in bd34055cec87fff47b4a13adf8da74bc3a8dec42!

fredrikekelund:
The Sentry issue appears to be unrelated to this PR.

PR #1327, opened 2025-05-06
SSL Support: Install root certificate directly on Windows instead of opening folder

gavande1:
## Related issues

- Fixes STU-446

## Proposed Changes

- Attempt to install Root SSL certificate on Windows

## Why changes are being made

On Windows, the certificate installation process can be automated. Previously, we opened the folder containing the certificate, requiring manual installation. This PR updates the behavior to call trustRootCA() on Windows, automatically installing the certificate when it’s not trusted. 

## Testing Instructions

- Apply this PR on Windows
- If your certificate is already trusted, remove it from the certificate manager. 
- run `npm start`
- Add HTTPS to your site. 
- Go to site settings, you should see the "Trust certificate" link
- When you click on it, you should see a dialogue box asking permission to execute the script.
- On trust, it will just open the file explorer in the certificates folder.

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-settings.tsx
+++ b/src/components/content-tab-settings.tsx
@@ -78,7 +78,7 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) 
 						<div>
 							<span>{ selectedSite.enableHttps ? __( 'Enabled' ) : __( 'Disabled' ) }</span>{ ' ' }
 							{ ! isCertificateTrusted && selectedSite.enableHttps && (
-								<Button variant="link" onClick={ () => getIpcApi().openCertificate() }>
+								<Button variant="link" onClick={ () => getIpcApi().trustCertificate() }>
 									{ __( 'Trust Certificate' ) }
 								</Button>
 							) }
--- a/src/components/content-tab-settings.tsx
+++ b/src/components/content-tab-settings.tsx
@@ -8,7 +8,10 @@ import { useGetWpVersion } from 'src/hooks/use-get-wp-version';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { decodePassword } from 'src/lib/passwords';
 import EditSiteDetails from 'src/modules/site-settings/edit-site-details';
-import { certificateTrustApi, useCheckCertificateTrustQuery } from 'src/stores/certificate-trust-api';
+import {
+	certificateTrustApi,
+	useCheckCertificateTrustQuery,
+} from 'src/stores/certificate-trust-api';
 
 interface ContentTabSettingsProps {
 	selectedSite: SiteDetails;
@@ -27,7 +30,8 @@ function SettingsRow( { children, label }: PropsWithChildren< { label: string } 
 
 export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) {
 	const { __ } = useI18n();
-	const { data: isCertificateTrusted, refetch: refetchCertificateTrust } = useCheckCertificateTrustQuery();
+	const { data: isCertificateTrusted, refetch: refetchCertificateTrust } =
+		useCheckCertificateTrustQuery();
 	const username = 'admin';
 	// Empty strings account for legacy sites lacking a stored password.
 	const storedPassword = decodePassword( selectedSite.adminPassword ?? '' );
--- a/src/components/content-tab-settings.tsx
+++ b/src/components/content-tab-settings.tsx
@@ -38,20 +38,6 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps ) 
 		: `localhost:${ selectedSite.port }`;
 	const protocol = selectedSite.customDomain && selectedSite.enableHttps ? 'https' : 'http';
 
-	const handleTrustCertificate = async () => {
-		const result = await getIpcApi().trustCertificate();
-		if ( ! result ) {
-			// show error message using electron dialog
-			await getIpcApi().showErrorMessageBox( {
-				title: __( 'Certificate Trust Failed' ),
-				message: __(
-					'Studio was unable to trust the certificate automatically. You may need to trust it manually using certificate manager.'
-				),
-				showOpenLogs: true,
-			} );
-		}
-	};
-
 	return (
 		<div className="p-8 ltr:pr-4 rtl:pl-4">
 			<div className="flex justify-between items-center mb-4">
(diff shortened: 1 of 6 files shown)


src/ipc-handlers.ts
```diff
@@ -1267,6 +1268,15 @@ export async function isCATrusted(): Promise< boolean > {
 	return isRootCATrusted();
 }
 
+export async function trustCertificate(): Promise< void > {
+	const platform = process.platform;
+	if ( platform === 'win32' ) {
+		return trustRootCA();
```
fredrikekelund: I think we should consider returning a boolean from `trustRootCA` to indicate if the operation was successful. If it fails, we can display an error modal. Silently failing would not be very helpful for users.

fredrikekelund (commented):
This is getting there, but I have two comments:

1. The thing about not failing silently if `trustRootCA` fails. 
2. We should automatically refresh the trust status after `getIpcApi().trustCertificate()` finishes executing. After landing https://github.com/Automattic/studio/pull/1325, we can do this like so:

```tsx
onClick={ async () => {
	await getIpcApi().trustCertificate();
	await dispatch(
		certificateTrustApi.util.invalidateTags( [ 'CertificateTrust' ] )
	);
} }
```

src/ipc-handlers.ts
gavande1: That's a very good suggestion. I have addressed it.

src/ipc-handlers.ts
```diff
@@ -1267,6 +1268,17 @@ export async function isCATrusted(): Promise< boolean > {
...
+export async function trustCertificate(): Promise< boolean > {
+	const platform = process.platform;
+	if ( platform === 'win32' ) {
+		return trustRootCA();
+	}
+
+	void openCertificateDialog();
+
+	return true;
+}
```
```suggestion
export async function trustCertificate( event: IpcMainInvokeEvent ): Promise< void > {
	if ( process.platform === 'win32' ) {
		const result = await trustRootCA();

		if ( ! result ) {
			await showMessageBox( event, {
				type: 'error',
				message: __( 'Certificate Trust Failed' ),
				detail: __(
					'Studio was unable to trust the certificate automatically. You may need to trust it manually using certificate manager.'
				),
			} );
		}
	}

	await openCertificateDialog();
}
```
I'd advise us to keep the logic for displaying the message box in the IPC handler function, and keep the component focused on calling this single IPC function.

fredrikekelund (commented):
Let's land https://github.com/Automattic/studio/pull/1325 first and also make the `invalidateTags` change I [suggested](https://github.com/Automattic/studio/pull/1327#pullrequestreview-2820620472) before landing this.

src/ipc-handlers.ts
nightnei: I am thinking - why do we go with returning boolean, instead of typical approach to throw error from `trustRootCA`? It looks redundant.
I noticed that in PHP world (at least in dotcom) we have these approaches, but I have never seen that it was used in JS world, maybe because JS has good errors ecosystem and async nature. If we look at `handleTrustCertificate` - the handling with `! result` looks weird.
Would be cool to not mute errors inside `trustRootCA`, always throw errors outside and consumers should decide what to do with errors - mute / show dialog / etc.

src/components/content-tab-settings.tsx
```diff
@@ -37,6 +37,21 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
 		? `${ selectedSite.customDomain }`
 		: `localhost:${ selectedSite.port }`;
 	const protocol = selectedSite.customDomain && selectedSite.enableHttps ? 'https' : 'http';
+
+	const handleTrustCertificate = async () => {
+		const result = await getIpcApi().trustCertificate();
+		if ( ! result ) {
+			// show error message using electron dialog
```
nightnei: `getIpcApi().showErrorMessageBox` is self-explanatory
```suggestion
```

nightnei (changes requested):

src/components/content-tab-settings.tsx
gavande1: You're right. The current approach of returning a boolean from `trustRootCA` and handling errors internally doesn't follow JavaScript/TypeScript best practices. It makes sense to throw an error from `trustRootCA` and handle it in `trustCertificate` with try-catch.

src/components/content-tab-settings.tsx
```diff
@@ -78,7 +79,7 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
 						<div>
 							<span>{ selectedSite.enableHttps ? __( 'Enabled' ) : __( 'Disabled' ) }</span>{ ' ' }
 							{ ! isCertificateTrusted && selectedSite.enableHttps && (
-								<Button variant="link" onClick={ () => getIpcApi().openCertificate() }>
+								<Button variant="link" onClick={ () => getIpcApi().trustCertificate() }>
```
fredrikekelund:
```suggestion
								<Button
									variant="link"
									onClick={ async () => {
										await getIpcApi().trustCertificate();
										dispatch( certificateTrustApi.util.invalidateTags( [ 'CertificateTrust' ] ) );
									} }
								>
```
Sorry if you're already working on this and I'm just nagging, @gavande1, but this change (that I mentioned [previously](https://github.com/Automattic/studio/pull/1327#pullrequestreview-2820620472)) is the only thing that I see remaining here.

gavande1: Yes. Was working on it. It seems I had to explicitly call `refetch` to update the UI. Addressed in [8e3e367](https://github.com/Automattic/studio/pull/1327/commits/8e3e367dfa493943ffef9fda8e6eee694c7d3a33).

gavande1:
Thanks both of you, @nightnei and @fredrikekelund, for amazing suggestions. The PR is ready to be reviewed again.

src/components/content-tab-settings.tsx
```diff
@@ -37,6 +41,16 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
...
 		: `localhost:${ selectedSite.port }`;
 	const protocol = selectedSite.customDomain && selectedSite.enableHttps ? 'https' : 'http';
+
+	const handleTrustCertificate = async () => {
+		await getIpcApi().trustCertificate();
+
+		// Invalidate the query to refresh the data
+		certificateTrustApi.util.invalidateTags( [ 'CertificateTrust' ] );
+
+		await refetchCertificateTrust();
```
gavande1: I think I know what happened, yet. I need to wrap in `dispatch` to invalidate tags

fredrikekelund: Yup, that's right 👍 You need the `dispatch` call. Use the `useAppDispatch` hook from `src/stores` to get a dispatch function.

nightnei (approved):
LGTM and works well 👍
Screenshot 2025-05-07 at 15 53 56

src/components/content-tab-settings.tsx
gavande1: I have tried using `dispatch` and it works. I have updated the PR.

PR #1344, opened 2025-05-07
Update Customize buttons to make them always active

wojtekn:
## Related issues

- Fixes STU-357

## Proposed Changes

- I propose to change Customize buttons to make them always active and auto-start site when clicked.

Screenshot 2025-05-07 at 18 49 22

## Testing Instructions

1. Stop the site
2. Click Customize buttons and confirm they open site after it is started
3. Test Classic theme, too.

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-overview.tsx
+++ b/src/components/content-tab-overview.tsx
@@ -50,51 +50,49 @@ function CustomizeSection( {
 	themeDetails?: SiteDetails[ 'themeDetails' ];
 	loading?: boolean;
 } ) {
+	const { startServer, loadingServer } = useSiteDetails();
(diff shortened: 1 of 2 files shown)


katinthehatsite (approved):
The functionality works as expected for me and the buttons now start a site.

I noticed one small thing is that if you start and stop the site quickly, the buttons look like they somewhat glitch and go from enabled to disabled to enabled too fast. It seems like the loading state might be too short and gives this effect (see on the video below when I am clicking the second time):

https://github.com/user-attachments/assets/812c761a-9f4e-4cdc-8dd4-eacacc243e3c

I do not think it is a common user behaviour though so I am approving the changes 👍

PR #1332, opened 2025-05-06
Studio Sync: Improve the percentage progress and add more states into account in PUSH progress bar

epeicher:
## Related issues

- Resolves STU-399

## Proposed Changes

- Implement `onEvent` for the export ipc handler so it emits the server events to the client.
- Refactor `getStatusWithProgress` to `getPullStatusWithProgress` and move it to the `useSyncStatesProgressInfo` hook.
- Refactor `getBackupStatusWithProgress` by moving it to the `useSyncStatesProgressInfo` hook.
- Create `getPushStatusWithProgress` by following a similar (but slightly different) approach than `getPullStatusWithProgress`.
- Use the new `getPushStatusWithProgress` method to update the progress and messages for the `Push` action.

## Testing Instructions

- Checkout this branch and run `npm start`
- Navigate to the Sync tab and Connnect a site or use an existing connected site
- Perform a Pull operation and make sure you see the same progress and messages than in Production, i.e. there are no regressions.
- Perform a Push operation and check that the progress and messages for the initial steps are more granular than in production, i.e. you can see more messages and a more linear progress for the initial steps

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #1345, opened 2025-05-07
Import/Export: Display progress only if not syncing

epeicher:
## Related issues

- Fixes STU-465

## Proposed Changes

- Refine the condition to display the Export progress bar so it is not displayed when Syncing
- Additionally, do not display the Export progress bar when it reaches 100 as it will be disabled anyway.

## Testing Instructions

1. Run `npm start`
3. Go to Sync tab
5. Connect a site
7. Click on Push
9. Observe the progress bar appears with messages
11. Go to Import/Export tab
13. The progress bar should not be visible
15. When the Push finishes, the export progress bar should not be visible and the button should be enabled

In the linked task, there is a Video with the expected behaviour, [link here](https://linear.app/a8c/issue/STU-465/when-pushing-a-site-a-progress-bar-appears-in-the-importexport-tab#heading-expectation-9c56de4f)

#### Regression testing
11. Go to Import/Export tab
12. Export a site
13. The progress bar should be work as before. It should be displayed and progressed.
15. When the Export finishes, the export progress bar should not be visible and the button should be enabled

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-import-export.tsx
+++ b/src/components/content-tab-import-export.tsx
@@ -34,8 +34,7 @@ export const ExportSite = ( {
 	const { [ selectedSite.id ]: currentProgress } = exportState;
 	const isImporting = importState[ selectedSite.id ]?.progress < 100;
 	const isExportDisabled = isImporting || isThisSiteSyncing;
-	const shouldDisplayProgress =
-		currentProgress && currentProgress.progress < 100 && ! isThisSiteSyncing;
+	const shouldDisplayProgress = currentProgress && ! isThisSiteSyncing;
 
 	let tooltipText;
 	if ( isThisSiteSyncing ) {
--- a/src/components/content-tab-import-export.tsx
+++ b/src/components/content-tab-import-export.tsx
@@ -34,6 +34,8 @@ export const ExportSite = ( {
 	const { [ selectedSite.id ]: currentProgress } = exportState;
 	const isImporting = importState[ selectedSite.id ]?.progress < 100;
 	const isExportDisabled = isImporting || isThisSiteSyncing;
+	const shouldDisplayProgress =
+		currentProgress && currentProgress.progress < 100 && ! isThisSiteSyncing;
 
 	let tooltipText;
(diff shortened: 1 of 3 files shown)


sejas (commented):
It seems to solve the issue, but I have found another edge case where it fails.

1. Push the site.
2. Wait for the push to finish
3. Got to Import/Export tab and Try exporting the site
4. Observe the dialog opens but the export doesn't run for that site. The export works for the rest of sites.

See the vide at 1min 19seg.

https://github.com/user-attachments/assets/c0eff690-82eb-4dc2-addf-8559fe367e8a

sejas:
@epeicher , maybe we should revert https://github.com/Automattic/studio/pull/1332 ? I don't see much improvement in the progress bar, and it seems it produces other issues.

epeicher:
> @epeicher , maybe we should revert #1332 ? I don't see much improvement in the progress bar, and it seems it produces other issues.

I have addressed the edge case [reported here](https://github.com/Automattic/studio/pull/1345#pullrequestreview-2822352448) , it was caused by the state of the Export not being reset so [this callback](https://github.com/Automattic/studio/blob/ef98a9a28848ca00cd94c98e3b934573739f59e0/src/hooks/use-import-export.tsx#L259) was returning early. The changes are similar to the [`clearImportState` here](https://github.com/Automattic/studio/blob/ef98a9a28848ca00cd94c98e3b934573739f59e0/src/hooks/sync-sites/use-sync-pull.ts#L215). The commit is ef98a9a.

Given said that. Reverting is always an option. I would continue improving as part of a follow-up task. But I trust your instincts here 😄 (as they are not instinct but knowledge 😉) , if you think there could be other issues, or you don't think the additional complexity is justified, let's revert it.

UPDATE: I have added a revert PR https://github.com/Automattic/studio/pull/1349 to err on the safe side. I close this PR, but save it for future reference.

epeicher:
Closing this as the revert https://github.com/Automattic/studio/pull/1349 has been merged. These changes can be used as a reference if the export states need to be emitted for the Sync Push functionality.

PR #1335, opened 2025-05-06
Open editors using bundle IDs

gcsecsey:
## Related issues

- Partly addresses [STU-450 Use CLI to open IDE instead of app URI](https://linear.app/a8c/issue/STU-450/use-cli-to-open-ide-instead-of-app-uri)

## Proposed Changes

This PR changes how we open IDEs on macOS by using bundle IDs instead of URL schemes. This change helps avoid the warning thrown by VSCode and forked editors when first opening a directory.

- Add `bundleId` property to editor configurations
- Add a new `openAppAtPath` IPC handler for macOS
- Update the editor opening logic to use bundle IDs on macOS

## Testing Instructions

Prerequisites: either VSCode, Cursor, or Windsurf installed

- Run `STUDIO_PREFERRED_EDITOR=true npm start`
- Go to preferences and choose VSCode, Cursor, or Windsurf
- On a site's Overview tab, try opening the directory using the selected editor
- Check that the app opens the directory, and that this warning isn't shown:

image

#### Regression
- If you have Jetbrains IDEs installed on Mac, test that there are no changes to opening directories
- If you have Windows, test that there are no changes to opening directories

## Pre-merge Checklist

- [X] Have you checked for TypeScript, React or other console errors?

PR #1346, opened 2025-05-07
Use executables to open editors on Windows

gcsecsey:
## Related issues

- Fixes [STU-450 Use CLI to open IDE instead of app URI](https://linear.app/a8c/issue/STU-450/use-cli-to-open-ide-instead-of-app-uri)
- Follow-up to https://github.com/Automattic/studio/pull/1335

## Proposed Changes

This PR changes how we open IDEs on Windows by using the executable path primarily, and falling back to URI schemes if needed. This change helps avoid the warning thrown by VSCode and forked editors when first opening a directory.

- Update `openAppAtPath` to handle both platforms
- Added Windows-specific paths to `supportedEditorConfig`

## Testing Instructions

- Run `npm start`
- Go to preferences and choose VSCode, Cursor, or Windsurf
- On a site's Overview tab, try opening the directory using the selected editor
- Check that the app opens the directory, and that this warning isn't shown:
- If you have Jetbrains IDEs installed, test that there are no changes to opening directories

VSCode forks (Cursor doesn't have a Windows arm64 build):

https://github.com/user-attachments/assets/d126542d-78ad-4b2b-b0fd-39b38ae9ccaa

JetBrains IDEs:

https://github.com/user-attachments/assets/231c31e5-27e4-48ae-9e13-5dc31621097a

#### Regression
- On Mac, test that there are no changes to opening directories

## Pre-merge Checklist


src/modules/user-settings/lib/editor.ts
```diff
@@ -14,29 +15,50 @@ export const supportedEditorConfig: Record< SupportedEditor, SupportedEditorConf
 		label: __( 'VS Code' ),
 		url: ( path: string ) => `vscode://file/${ path }?windowId=_blank`,
 		bundleId: 'com.microsoft.VSCode',
+		winPaths: [
+			'%LOCALAPPDATA%\\Programs\\Microsoft VS Code\\code.exe',
+			'%PROGRAMFILES%\\Microsoft VS Code\\code.exe',
+			'%PROGRAMFILES(X86)%\\Microsoft VS Code\\code.exe',
+		],
```
gcsecsey, line 22: These paths somewhat duplicate what we were already doing in [is-installed](https://github.com/Automattic/studio/blob/036d7ae7651ff3e3d7c1cc24b5e1c6458fa9b556/src/lib/is-installed.ts#L64-L90), but the path here needed to be more specific. I thought it was beyond the scope of this PR to combine the two solutions.

However, it'd be nice to use `supportedEditorConfig` as the source of truth, and update `is-installed` to work with these paths. I'll create a separate PR for this.

src/modules/user-settings/lib/editor.ts
```diff
@@ -6,6 +6,7 @@ export type SupportedEditorConfig = {
 	label: string;
 	url: ( path: string ) => string;
 	bundleId: string;
```
fredrikekelund:
```suggestion
	macOSBundleId: string;
```
Just a suggestion, but we might as well clarify that this is also platform-specific.

src/ipc-handlers.ts
```diff
@@ -1081,6 +1070,95 @@ END` );
...
+							.catch( () => false )
+					) {
+						return fullPath;
+					}
+				}
+			} catch ( error ) {
+				// Skip if directory doesn't exist
+				continue;
+			}
+		} else {
```
fredrikekelund: Is this needed? None of the paths in `src/modules/user-settings/lib/editor.ts` contain a wildcard character.

src/ipc-handlers.ts
```diff
@@ -1081,6 +1070,95 @@ END` );
...
+ * Finds the executable path for a given editor on Windows
+ */
+async function winFindEditorPath( editorKey: SupportedEditor ): Promise< string | null > {
+	const editor = supportedEditorConfig[ editorKey ];
+	if ( ! editor || ! editor.winPaths ) {
+		return null;
+	}
+
+	for ( const possiblePath of editor.winPaths ) {
+		const expandedPath = winExpandPath( possiblePath );
```
fredrikekelund: Optional, but given how short `winExpandPath` is, how about we just inline it?

src/ipc-handlers.ts
```diff
@@ -1081,6 +1070,95 @@ END` );
...
+ * Expands Windows environment variables in a path
+ */
+function winExpandPath( path: string ): string {
+	return path.replace( /%([^%]+)%/g, ( _, n ) => process.env[ n ] || '' );
+}
+
+/**
+ * Finds the executable path for a given editor on Windows
+ */
+async function winFindEditorPath( editorKey: SupportedEditor ): Promise< string | null > {
```
fredrikekelund: I would advise moving this function to a module in `src/modules/user-settings/lib`.

src/ipc-handlers.ts
```diff
@@ -1081,6 +1070,95 @@ END` );
...
+				// Skip if directory doesn't exist
+				continue;
+			}
+		} else {
+			if (
+				await fs.promises
+					.access( expandedPath )
+					.then( () => true )
+					.catch( () => false )
+			) {
```
```suggestion
		if ( fs.existsSync( expandedPath ) ) {
```
Given Electron's multi-process architecture (meaning sync calls here don't interrupt the renderer process) and the speed of just checking for the existence of a file, I think `fs.existsSync` is worth considering.

src/ipc-handlers.ts
```diff
@@ -1081,6 +1070,95 @@ END` );
...
+	editorKey: SupportedEditor,
+	filePath: string
+): Promise< void > {
+	const platform = process.platform;
+	const editor = supportedEditorConfig[ editorKey ];
+
+	if ( ! editor ) {
+		throw new Error( `Editor ${ editorKey } not found or not supported` );
+	}
+
```
```suggestion
```
Typescript already guarantees that `editor` is a `SupportedEditorConfig` object.

fredrikekelund (approved):
Works beautifully 👍 I suggested a couple of changes, but nothing is blocking, so in the interest of moving things along, I'll approve straight away.

src/ipc-handlers.ts
```diff
@@ -58,12 +58,12 @@ import { getLogsFilePath, writeLogToFile, type LogLevel } from 'src/logging';
 import { getMainWindow } from 'src/main-window';
 import { popupMenu, setupMenu } from 'src/menu';
 import { executePreviewCliCommand } from 'src/modules/cli/lib/execute-preview-command';
-import { SupportedEditor } from 'src/modules/user-settings/lib/editor';
-import { SupportedTerminal } from 'src/modules/user-settings/lib/terminal';
 import { SiteServer, createSiteWorkingDirectory } from 'src/site-server';
 import { DEFAULT_SITE_PATH, getResourcesPath, getSiteThumbnailPath } from 'src/storage/paths';
 import { loadUserData, saveUserData } from 'src/storage/user-data';
 import { DEFAULT_PHP_VERSION, DEFAULT_WORDPRESS_VERSION } from 'vendor/wp-now/src/constants';
+import { supportedEditorConfig, SupportedEditor } from './modules/user-settings/lib/editor';
```
gavande1: Nit: Let's use full namespace
```suggestion
import { supportedEditorConfig, SupportedEditor } from 'src/modules/user-settings/lib/editor';
```

src/ipc-handlers.ts
```diff
@@ -58,12 +58,12 @@ import { getLogsFilePath, writeLogToFile, type LogLevel } from 'src/logging';
...
+import { SupportedTerminal } from './modules/user-settings/lib/terminal';
```
gavande1: Same as above

src/ipc-handlers.ts
```diff
@@ -1081,6 +1070,95 @@ END` );
 	}
 }
 
+export async function openAppAtPath(
+	_event: IpcMainInvokeEvent,
```
gavande1: @fredrikekelund I am curious, do we need `_event` in this case?

src/ipc-handlers.ts
```diff
@@ -1081,6 +1070,95 @@ END` );
...
+		const expandedPath = winExpandPath( possiblePath );
+
+		// Handle wildcards in paths (for JetBrains Toolbox installations)
+		if ( expandedPath.includes( '*' ) ) {
+			const basePath = nodePath.dirname( expandedPath );
+			const pattern = nodePath.basename( expandedPath );
+
+			try {
+				const files = await fs.promises.readdir( basePath );
+				const matchingFiles = files.filter( ( f ) =>
```
gavande1: We should use the full name whenever possible. 
```suggestion
				const matchingFiles = files.filter( ( file ) => 
				file.match( new RegExp( pattern.replace( '*', '.*' ) ) )
```

src/ipc-handlers.ts
fredrikekelund: All IPC functions always receive an `IpcMainInvokeEvent` object as the first parameter, so we need to include it. However, we don't need to prefix the parameter name with an underscore.

gavande1: Thanks for explaining that.

src/modules/user-settings/lib/editor.ts
```diff
@@ -14,29 +15,50 @@ export const supportedEditorConfig: Record< SupportedEditor, SupportedEditorConf
...
+		],
 	},
 	phpstorm: {
 		// translators: "PhpStorm" is the brand name for an IDE and does not need to be translated
 		label: __( 'PhpStorm' ),
 		url: ( path: string ) => `phpstorm://open?file=${ path }`,
 		bundleId: 'com.jetbrains.PhpStorm',
+		winPaths: [
+			'%LOCALAPPDATA%\\Programs\\PhpStorm\\bin\\phpstorm64.exe',
+			'%PROGRAMFILES%\\JetBrains\\PhpStorm\\bin\\phpstorm64.exe',
```
gavande1: I am afraid that these absolute paths are not guaranteed. I just installed PhpStorm on my Windows, and it was installed in `JetBrains\PhpStorm 2025.1.0.1\bin\phpstorm64.exe`. 

JetBrains products use `<product><version>` pattern for directories. You can read more about it at: https://www.jetbrains.com/help/phpstorm/directories-used-by-the-ide-to-store-settings-caches-plugins-and-logs.html#config-directory

src/modules/user-settings/lib/editor.ts
```diff
@@ -14,29 +15,50 @@ export const supportedEditorConfig: Record< SupportedEditor, SupportedEditorConf
...
+		],
 	},
 	cursor: {
 		// translators: "Cursor" is the brand name for an IDE and does not need to be translated
 		label: __( 'Cursor' ),
 		url: ( path: string ) => `cursor://file/${ path }?windowId=_blank`,
 		bundleId: 'com.todesktop.230313mzl4w4u92',
+		winPaths: [
+			'%LOCALAPPDATA%\\Programs\\Cursor\\Cursor.exe',
+			'%PROGRAMFILES%\\Cursor\\Cursor.exe',
```
gavande1: I think it doesn't matter on Windows, but I still wanted to point it out. For me, the Cursor files in the Programs directory are in small case. Overall path is `C:\Users\gavande\AppData\Local\Programs\cursor\Cursor.exe` 

CleanShot 2025-05-08 at 14 58 07@2x

src/ipc-handlers.ts
gcsecsey: I was experimenting with recognizing Jetbrains IDEs in their versioned folders, and for these, I included a `*` in the path to match them. Later I removed the wildcards from the paths because I thought Jetbrains doesn't use the versioned folders anymore, but that seem to only be the case when installed for the current user only.

As @gavande1 [pointed out below](https://github.com/Automattic/studio/pull/1346#discussion_r2079276669), this might be needed for system-wise installations of the Jetbrains IDEs after all. I'll revert that path to include a wildcard.

src/modules/user-settings/lib/editor.ts
gcsecsey: Good point, done in b6528ade

src/ipc-handlers.ts
gcsecsey: Yep, there were multiple uses of this ealier, but now it makes sense. Done in d1b7f20f.

src/ipc-handlers.ts
gcsecsey: Done in 06bc7ac0.

src/ipc-handlers.ts
gcsecsey: I wanted to prefix the parameter with `_` in this case to signal that it's not being used/referenced within the function. But I now realized that I ended using it when falling back to `openURL`, so I updated this in 06bc7ac0.

src/ipc-handlers.ts
gcsecsey: Thanks, done in a6824001.

src/ipc-handlers.ts
gcsecsey: I agree, done in bacb2ac210261e9e0a32632201cbacef4007abaf.

src/ipc-handlers.ts
gcsecsey: Done in bacb2ac210261e9e0a32632201cbacef4007abaf

src/ipc-handlers.ts
gcsecsey: Done in bacb2ac210261e9e0a32632201cbacef4007abaf

src/modules/user-settings/lib/editor.ts
gcsecsey: Thanks for highlighting, yes, Windows handles paths in a "case insensitive, case preserving" way. I've still updated the path in 81e5bd6c84659fdf2d9884c8caa07ac0cac2d6a8 just for correctness.

src/ipc-handlers.ts
gcsecsey: Good point, thanks. Done in 761352569211bb0534b859ae57bca9e8604ff471.

src/modules/user-settings/lib/editor.ts
gcsecsey: Thanks for sharing, I also saw this docs page earlier. Just a small clarification, this page refers to the directories Jetbrains uses to store config, caches, and logs, not the installation paths.

Could you share if `JetBrains\PhpStorm 2025.1.0.1\bin\phpstorm64.exe` is within Program Files or your user App Data folder for you?
 
Based on what I find on the internet, I believe the path doesn't contain a version number for IDEs installed for the current user only (this is the default in the Toolbox app). I could also verify this on my VMs.

For IDEs installed for all users, it should go to Program Files, but I'm unsure about the exact folder structure here.

gcsecsey: Using the [standalone installer](https://www.jetbrains.com/help/webstorm/installation-guide.html#standalone) I could verify that for system-wide installs, the path should be `%PROGRAMFILES\JetBrains\<product> <version>`.

CleanShot 2025-05-08 at 11 50 32@2x

gavande1: > Could you share if JetBrains\PhpStorm 2025.1.0.1\bin\phpstorm64.exe is within Program Files or your user App Data folder for you?

Yes. This is in the program files.

gcsecsey: I updated the matching logic in 7ca0c5a4e4a62b531c599c87f278aaad8681c357

src/ipc-handlers.ts
gcsecsey: I updated the matching logic in 7ca0c5a4e4a62b531c599c87f278aaad8681c357

gcsecsey:
Thanks @gavande1 and @fredrikekelund for your insights! 🙌 

I considered removing the JetBrains IDEs from using the executables altogether, because these don't show the warning we opened this issue for. However, while testing falling back to the app URIs, I noticed that eg. the WebStorm standalone installer doesn't set up WebStorm to handle the `webstorm://` URIs, and I get this modal in this case:
CleanShot 2025-05-08 at 15 42 56@2x

Because of this, I tweaked the path detection for JetBrains IDEs further, and now it should work correctly whether the app is installed only for the user, or system-wide. Here's a video of opening PhpStorm from the user app data, and WebStorm from Program Files:

https://github.com/user-attachments/assets/eeb88309-545c-49a6-9822-4977215ae407

I'll merge this once CI is green, but we can continue iterating on this as needed.

PR #1347, opened 2025-05-07
Improve localization process to remove old phrases

wojtekn:
## Related issues

- N/A

## Proposed Changes

I propose to improve the localization process to remove old phrases when new bundle is generated. It seems that if previously created `out/pots/bundle-strings.pot` is not removed, running `npm run make-pot` will add phrases to the duplicate files. It results in leaving old phrases there.

To test the issue:

1. Run `wp-babel-makepot`
2. Update any phrase
3. Run `wp-babel-makepot` again
4. Confirm that both old and new phrase appear in the file

I checked `wp-babel-makepot` and I couldn't find any option that would allow me to purge the file.

## Testing Instructions

Review new steps to ensure they are clear.

## Pre-merge Checklist

### DIFF
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -17,9 +17,9 @@ If you want to add support for another language you will need to add it to the
 
 #### Step 1: Extract Strings:
 
-   1. Run `npm run make-pot` to get the text out of the source
-   files.
-   
+   1. Remove `out/pots/` directory if it exists.
+   2. Run `npm run make-pot` to get the text out of the source files.
+
    This will create a `*.pot` file for each module, as well as a bundle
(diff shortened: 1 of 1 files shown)


fredrikekelund (approved):
Good catch 👍 Would it make sense to even include an `rm -rf` step in the `package.json` `make-pot` script, perhaps? I can't imagine a scenario where we actually want the current concat behavior…

wojtekn:
> Would it make sense to even include an rm -rf step in the package.json make-pot script, perhaps? I can't imagine a scenario where we actually want the current concat behavior…

I thought about that and explored the options provided by the tool, but couldn't see anything for that. I didn't feel good about including `rm -rf` as a part of those scripts. 🤷‍♂️

PR #1350, opened 2025-05-08
User Settings: Render help when no editors are installed

gavande1:
## Related issues

- Fixes STU-436

## Proposed Changes

- Add a recommendation to download VS Code when no editor is installed. 
- Add tests for the editor picker component

## Testing Instructions

- Make the following changes to mimic uninstalled editors. 

```diff
diff --git a/src/ipc-handlers.ts b/src/ipc-handlers.ts
index e0e5308..e07afdb 100644
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -108,11 +108,11 @@ export async function getSiteDetails( _event: IpcMainInvokeEvent ): Promise< Sit
 
 export function getInstalledAppsAndTerminals(): InstalledApps & InstalledTerminals {
        return {
-               vscode: isInstalled( 'vscode' ),
-               phpstorm: isInstalled( 'phpstorm' ),
-               webstorm: isInstalled( 'webstorm' ),
-               windsurf: isInstalled( 'windsurf' ),
-               cursor: isInstalled( 'cursor' ),
+               vscode: false, // isInstalled( 'vscode' ),
+               phpstorm: false, // isInstalled( 'phpstorm' ),
+               webstorm: false, // isInstalled( 'webstorm' ),
+               windsurf: false, // isInstalled( 'windsurf' ),
+               cursor: false, // isInstalled( 'cursor' ),
                terminal: true, // Terminal.app is always available on macOS
                iterm: isInstalled( 'iterm' ),
                warp: isInstalled( 'warp' ),
```

- Run `npm start`
- Make sure you can see the help text pointing to https://developer.wordpress.com/docs/developer-tools/studio/sites/#site-settings

CleanShot 2025-05-08 at 12 05 51@2x

## Pre-merge Checklist

### DIFF
--- a/src/modules/user-settings/components/editor-picker.tsx
+++ b/src/modules/user-settings/components/editor-picker.tsx
@@ -35,6 +35,7 @@ export const EditorPicker = ( { value, onChange, disabled }: EditorPickerProps )
 					)
 				}
 				variant="link"
+				className="text-xs"
 			>
 				{ props.children } ↗
 			</Button>
@@ -47,7 +48,7 @@ export const EditorPicker = ( { value, onChange, disabled }: EditorPickerProps )
 		}
 
 		return (
-			<p className="text-gray-500">
+			<p className="text-gray-500 text-xs">
 				{ createInterpolateElement(
 					__( 'You can find a list of supported code editors <a>here</a>.' ),
 					{
--- a/src/modules/user-settings/components/editor-picker.tsx
+++ b/src/modules/user-settings/components/editor-picker.tsx
@@ -1,5 +1,8 @@
-import { SelectControl } from '@wordpress/components';
+import { Button, SelectControl } from '@wordpress/components';
+import { createInterpolateElement } from '@wordpress/element';
 import { useI18n } from '@wordpress/react-i18n';
+import { PropsWithChildren } from 'react';
+import { getIpcApi } from 'src/lib/get-ipc-api';
 import { SupportedEditor } from 'src/modules/user-settings/lib/editor';
 import {
 	useGetInstalledAppsQuery,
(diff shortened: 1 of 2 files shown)


wojtekn:
I'm wondering if we ever need this message. The field already lists editors under the 'Not installed' group. Should we skip it to avoid cluttering the UI? @Marinatsu anu thoughts?

src/modules/user-settings/components/editor-picker.tsx
```diff
@@ -39,6 +75,12 @@ export const EditorPicker = ( { value, onChange, disabled }: EditorPickerProps )
 							{ editorConfig.label }
 						</option>
 					) ) }
+
+					{ installedEditors.length === 0 && (
+						<option value={ '' } disabled>
+							{ __( 'None' ) }
+						</option>
+					) }
```
gavande1, line 83: I have added this `None` option when no editors are installed to clearly differentiate between the `Available` and `Not Installed` groups. 

CleanShot 2025-05-08 at 12 33 14@2x

gavande1:
> I'm wondering if we ever need this message. The field already lists editors under the 'Not installed' group. Should we skip it to avoid cluttering the UI? @Marinatsu anu thoughts?

We discussed adding this on Linear. I think it makes sense to add this help text and the idea originated from Fredrik's comment on the [previous PR](https://github.com/Automattic/studio/pull/1309). Let me know if you think otherwise. 

> The dropdown menu already has the Not installed optgroup, and it might be interesting to know which editors Studio supports.

cc @gcsecsey

gcsecsey (approved):
Thanks for tackling this @gavande1! The changes LGTM and work well.

gcsecsey:
> > I'm wondering if we ever need this message. The field already lists editors under the 'Not installed' group. Should we skip it to avoid cluttering the UI? @Marinatsu anu thoughts?
> 
> We discussed adding this on Linear. I think it makes sense to add this help text and the idea originated from Fredrik's comment on the [previous PR](https://github.com/Automattic/studio/pull/1309). Let me know if you think otherwise.

Just for context, we started working on handling this state based on RToz6tIuQ7nlZrikBte4GU-fi-7122_211912. If none of the editors are supported, IMO it makes sense to point the user to the documentation with more details, but it'd be great to get some UX guidance on this @Marinatsu @matt-west.

image

gavande1:
Thanks @gcsecsey for the review and providing additional context. I will wait for the review from the design team.

matt-west:
In the original design we were going to disable the field if no supported editors were installed which is why the note was added. It looks like we keep the field enabled though?

I think disabling the field is cleaner, but if we don't go with that direction, the note isn't strictly needed.

wojtekn:
Thanks @matt-west , the note makes sense indeed if the field is disabled.

As we are now close to the release date, we will skip it at this point and will leave the current solution.

PR #1359, opened 2025-05-09
Add translations for 1.5.1

wojtekn:
## Related issues

- pfHvTO-ws-p2

## Proposed Changes

I propose adding translations for 1.5.1.

## Testing Instructions

- Test if Studio works when different language is selected

## Pre-merge Checklist


epeicher (commented):
Thanks @wojtekn , I have identified one string in Spanish that does not seem to be translated in this version, while I can see it in the Production version:

|This branch | Production and trunk|
|-------|------|
|CleanShot 2025-05-09 at 09 54 52@2x| CleanShot 2025-05-09 at 09 51 27@2x|

katinthehatsite:
I was also reviewing this and checked that string and it seems to be translated in GlotPress: https://translate.wordpress.com/projects/studio/es/default/?filters%5Bterm%5D=Creating+backup&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc

I see that it is also translated in `common/translations/studio-es.jed.json`. Perhaps these translations are not applying specifically for that bit?

epeicher:
I am comparing the `translations-es.json` before and after and I think the issue is that the version `Creating backup...` with three dots instead of the ellipsis has been deleted. I think this could be fixed by replacing the three dots by an ellipsis [here](https://github.com/Automattic/studio/blob/570b92ec03c21c3f246f8d815aec9e2c786b34b1/src/hooks/use-import-export.tsx#L399-L400)

epeicher:
I have also added some Spanish translations 
CleanShot 2025-05-09 at 10 16 13@2x

katinthehatsite:
@wojtekn before you merge, could you please add the translations once again. I just added a fix for STU-472 in GlotPress that I spotted while testing this. All the updates should be made so if we add a new version of translations, I think that issue will be solved 🙇

PR #1361, opened 2025-05-09
Studio: Fix misc warnings thrown during test execution

katinthehatsite:
## Related issues

Fixes STU-468

## Proposed Changes

This PR cleans up different warnings for tests such as `listenerMiddleware/error` and ` An update to EditSiteDetails inside a test was not wrapped in act(...` and similar. 

## Testing Instructions

* Run tests with `npm run test`
* Confirm the warnings are not present

* You can also run the tests on trunk with `npm run test` to confirm that the warnings are gone

## Pre-merge Checklist


gavande1 (commented):
Thanks @katinthehatsite for improving the test. However, I am still seeing some errors while running the tests. Refer attached video:

https://github.com/user-attachments/assets/48e39beb-3003-4c1a-bf78-3e20e95a54df

katinthehatsite:
@gavande1 I mentioned those in the PR description:

```
As a sidenote: You will still see some console error such as:

An unhandled error occurred processing a request for the endpoint "getWordPressVersions".  In the case of an unhandled error, no tags will be "provided" or "invalidated". Error: Network error

or An unhandled error occurred processing a request for the endpoint "getWordPressVersions". In the case of an unhandled error, no tags will be "provided" or "invalidated". ZodError: [.

I think these errors are expected because we are setting up the tests in wordpress-versions-api.test.ts to handle them like so. We can change approach here if there are other suggestions.
```

I believe these should be expected as we are setting up the tests to throw those.

wojtekn:
Can we change tests so they do not rely on unhandled exceptions?

gavande1:
Thanks for pointing that out. I somehow read it as if the following error should not be present.

fredrikekelund:
> Can we change tests so they do not rely on unhandled exceptions?

You could argue that the offending tests (`should handle schema validation error for both channels` in `src/stores/tests/wordpress-versions-api.test.ts`, for example) are too granular in what they test, but principally, I think it's great to test that the code can handle unexpected inputs.

fredrikekelund:
However, it would be nice to silence the output so the errors we test aren't displayed in the console.

wojtekn:
> However, it would be nice to silence the output so the errors we test aren't displayed in the console.

Yes, this was my main point.

katinthehatsite:
> However, it would be nice to silence the output so the errors we test aren't displayed in the console.

I looked into addressing that comment on Friday briefly but I did not decide on the approach yet and this was one of the options I considered as well. I think I will proceed with this 👍

katinthehatsite:
@gavande1 @wojtekn I made some adjustments for that error, feel free to give it another review. 🙇

src/stores/tests/wordpress-versions-api.test.ts
```diff
@@ -144,6 +144,9 @@ describe( 'WordPress Versions API', () => {
 	} );
 
 	it( 'should handle schema validation error for both channels', async () => {
+		const originalConsoleError = console.error;
+		console.error = jest.fn();
```
fredrikekelund: Maybe we should consider running Jest with the `--silent` option ([docs](https://jestjs.io/docs/cli#--silent)) instead? Seems like a good approach to me.

gavande1:
@katinthehatsite, I am getting the following error message while running the test. Should we address it as well?

CleanShot 2025-05-12 at 15 31 17@2x

src/stores/tests/wordpress-versions-api.test.ts
katinthehatsite: I looked into this but if I am not mistaken, I think this silences the results for all tests, not just the ones that are expected to produce errors. I think I would prefer to keep the logs working for tests that are not expected to produce errors so that we can be alerted if something gets messed with a specific setup. 

I went with `spyOn` to silence errors for specific tests where they are expected. We already use this approach in other tests as well.

katinthehatsite:
@gavande1 I adjusted this in https://github.com/Automattic/studio/pull/1361/commits/c39aaaf85d38400abcfbcd96ee512928dba4364c , feel free to give this another review. Initially, I did not silence this error because it falls in the same category of being expected but since we are going for the approach of silencing expected errors, I applied the changes to it as well.

wojtekn (approved):
Looks good to me.

It is not really scope of this issue, but what do you think about silencing `console.log` output? Do we see it useful?

src/stores/tests/wordpress-versions-api.test.ts
fredrikekelund: Per the docs, it silences all console output from tests. Seems like it could save us some repeated boilerplate code in the tests. 

> We already use this approach in other tests as well.

Yeah, I know I've implemented this approach in the past, too. Defaulting to not printing console output seems like a better approach to me, though. We could discuss this separately if you prefer, though

katinthehatsite: > Yeah, I know I've implemented this approach in the past, too. Defaulting to not printing console output seems like a better approach to me, though. We could discuss this separately if you prefer, though

I will open a separate issue and we can discuss it there 👍

katinthehatsite: I added STU-493 to discuss this.

fredrikekelund:
> what do you think about silencing console.log output?

@wojtekn, see https://github.com/Automattic/studio/pull/1361#discussion_r2084261850 and STU-493

PR #1362, opened 2025-05-09
Change all three dots to ellipsis char for consistency

wojtekn:
## Related issues

- Fixes STU-476

## Proposed Changes

- I propose to change all three dots to ellipsis character for consistency.

## Testing Instructions

- test if character is displayed correctly in the app e.g. during sync

## Pre-merge Checklist

### DIFF
--- a/cli/commands/preview/create.ts
+++ b/cli/commands/preview/create.ts
@@ -18,27 +18,27 @@ export async function runCommand( siteFolder: string ): Promise< void > {
 	const logger = new Logger< LoggerAction >();
 
 	try {
-		logger.reportStart( LoggerAction.VALIDATE, __( 'Validating...' ) );
+		logger.reportStart( LoggerAction.VALIDATE, __( 'Validating…' ) );
 		validateSiteFolder( siteFolder );
 		await validateSiteSize( siteFolder );
 		const token = await getAuthToken();
 		logger.reportSuccess( __( 'Validation successful' ), true );
 
-		logger.reportStart( LoggerAction.ARCHIVE, __( 'Creating archive...' ) );
+		logger.reportStart( LoggerAction.ARCHIVE, __( 'Creating archive…' ) );
 		await createArchive( siteFolder, archivePath );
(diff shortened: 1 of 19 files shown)


sejas (commented):
Changes look good. Thanks for replacing ellipsis.

@wojtekn, I also found the three dots in cli/commands files. I think we should replace them:

```
cli/commands/preview/create.ts:
cli/commands/preview/delete.ts:
cli/commands/preview/list.ts:
cli/commands/preview/update.ts:
cli/commands/preview/tests/create.test.ts:
cli/commands/preview/tests/delete.test.ts:
cli/commands/preview/tests/list.test.ts:
cli/commands/preview/tests/update.test.ts:
```

sejas:
After merging this PR, we might need to update translations from GlotPress.

gcsecsey (commented):
LGTM and these now look consistent. I agree with @sejas that we should also update the usages in CLI.

CleanShot 2025-05-09 at 12 51 53@2x

sejas (approved):
Thanks for the updating the ellipsis in the command! ✨

epeicher:
Do you think we could add a linting rule for ensuring ellipsis in translatable strings? We can use something similar [to this one](https://github.com/Automattic/wp-calypso/blob/trunk/packages/eslint-plugin-wpcalypso/docs/rules/i18n-ellipsis.md), but I have quickly tried and it does not work _out of the box_ because it uses a [deprecated rule format](https://eslint.org/docs/v8.x/extend/custom-rules-deprecated) so it implies some additional effort.

wojtekn:
@epeicher, if there is no ready-to-use rule, I'm unsure if it's worth the effort. 🤷‍♂️

epeicher:
> @epeicher, if there is no ready-to-use rule, I'm unsure if it's worth the effort. 🤷‍♂️

I actually agree 😄  I explored the idea, assuming it would be a 5-minute improvement, but after realising it involved more, I don't think it is worth the effort. Anyway, I wanted to hear your opinion. Thanks for sharing your view!

PR #1363, opened 2025-05-09
Fix snapshot schema format - name optional

bcotrim:
## Related issues

- Fixes STU-475

## Proposed Changes

- Make the snapshot name optional in the schema to allow for older `appdata` file formats

## Testing Instructions

- Edit your `appdata-v1.json` file and add a snapshot like this:
```

  "snapshots": [
    {
      "url": "secretly-cognitive.jurassic.ninja",
      "atomicSiteId": 150570991,
      "localSiteId": "64e2ffab-590c-455a-ac98-cb1f61f46d9e",
      "date": 1711097456449
    }
  ],
```
- Run `npm run start`
- Create a new site and login to WordPress.com
- Try to create a preview site
- Preview site is created successfully

## Pre-merge Checklist

### DIFF
--- a/common/types/snapshot.ts
+++ b/common/types/snapshot.ts
@@ -5,7 +5,7 @@ export const snapshotSchema = z.object( {
 	atomicSiteId: z.number(),
 	localSiteId: z.string(),
 	date: z.number(),
-	name: z.string(),
+	name: z.string().optional(),
 	userId: z.number().optional(),
 	sequence: z.number().optional(),
 } );
--- a/cli/commands/preview/create.ts
+++ b/cli/commands/preview/create.ts
@@ -46,7 +46,7 @@ export async function runCommand( siteFolder: string ): Promise< void > {
 		);
 		logger.reportSuccess( __( 'Preview site saved to Studio' ) );
 
-		logger.reportKeyValuePair( 'name', snapshot.name );
+		logger.reportKeyValuePair( 'name', snapshot.name ?? '' );
 		logger.reportKeyValuePair( 'url', snapshot.url );
 	} catch ( error ) {
 		if ( error instanceof LoggerError ) {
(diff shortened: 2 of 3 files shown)


common/types/snapshot.ts
```diff
@@ -5,7 +5,7 @@ export const snapshotSchema = z.object( {
 	atomicSiteId: z.number(),
 	localSiteId: z.string(),
 	date: z.number(),
-	name: z.string(),
+	name: z.string().optional(),
```
sejas, line 8: 👍

epeicher (approved):
I have tested it, and it works as advertised. I can see the error in `trunk` while this branch fixes it.

|Trunk | This branch|
|-------|------|
| CleanShot 2025-05-09 at 14 01 06@2x|CleanShot 2025-05-09 at 13 59 59@2x|



LGTM! :shipit:

wojtekn (approved):
It fixes the issue for me.

gcsecsey (approved):
Thanks @bcotrim for working on this. These changes LGTM, and I was able to create a site and preview site after having a snapshot without a name in my appdata.

CleanShot 2025-05-09 at 13 02 08@2x
CleanShot 2025-05-09 at 13 02 15@2x

sejas (approved):
I tested it by modifying my appData and it works as expected, and I don't see the error anymore.
I confirm delete and update also work 👍 

I reproduced the error on trunk:
Screenshot 2025-05-09 at 13 02 06

common/types/snapshot.ts
fredrikekelund, line 8:
```suggestion
	name: z.string().default( '' ),
```
A suggestion for future similar changes is to use zod's `default` method. It's not always the right choice, but in cases like this, where we always (or mostly, at least) default to an empty string anyway, it just makes things easier.

PR #1368, opened 2025-05-09
Translate what's new links

sejas:
## Related issues

- Fixes STU-470

## Proposed Changes

- Create a new function to translate links to blog posts

## Testing Instructions

- Run `npm start`
- Click on your profile name at the top right corner
- Change the language to Spanish or Japanese
- Click on Help on the system menu 
- Click on "What's new" to open the modal
- Navigate to the last card about Preview Sites.
- Click on the link.
- Observe the browser opens on the right blog post

https://github.com/user-attachments/assets/27b0bad8-701c-4d31-aeed-0bfee8e3cdbe

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1375, opened 2025-05-12
Refactor getDocsLink in favor of getLocalizedLink

sejas:
## Related issues

- Fixes STU-470

## Proposed Changes

- Unify the getBlogLink and translateLink functions that translate links into a single, easy-to-use function called `getLocalizedLink`, maintain a dictionary for centralized link usage, and preserve the essence suggested in https://github.com/Automattic/studio/pull/1368#discussion_r2084437501

## Testing Instructions

- Run `npm start`
- Confirm the links in what's new are still working. You can follow the tests on #1368 
- Confirm the links pointing to the docs are still working in English, and Spanish.
  - Sections: 'system menu', 'content-tab-import-export', 'site-form', 'sync-connected-sites', and 'top-bar'

## Pre-merge Checklist


gcsecsey (approved):
Thanks for making this follow-up @sejas! 🙌 

The refactor looks good to me and works well. I can still open the Spanish blog post from the What's New section using this branch.

https://github.com/user-attachments/assets/ce3e9c1c-4c1a-4cfc-8b79-b0b40ccb2ad1

src/lib/translate-link.ts
```diff
@@ -1,16 +1,40 @@
...
+export const BLOG_LINKS = {
+	'php-versions': 'https://wordpress.com/blog/2025/03/17/studio-wordpress-php-versions/',
+	'preview-sites': 'https://wordpress.com/blog/2025/02/24/studio-preview-sites/',
+	'custom-domains-https': 'https://wordpress.com/blog/2025/03/31/studio-custom-domains-https/',
+};
+
+const BLOG_TRANSLATIONS: Map< string, Partial< Record< SupportedLocale, string > > > = new Map( [
 	[
-		'https://wordpress.com/blog/2025/03/17/studio-wordpress-php-versions/',
+		BLOG_LINKS[ 'php-versions' ],
```
wojtekn: Would it be more transparent and easier to read if we stored the default one in the `en` key along with translations?

src/menu.ts
```diff
@@ -236,7 +236,7 @@ function getAppMenu(
 					label: __( 'Studio Help' ),
 					click: async () => {
 						const locale = await getUserLocaleWithFallback();
-						void shellOpenExternalWrapper( getDocsLink( locale, 'studio' ) );
+						void shellOpenExternalWrapper( translateLink( locale, DOCS_LINKS.studio ) );
```
wojtekn: I think it was clearer when function accepted the link name as a parameter.

src/lib/tests/translate-link.test.ts
```diff
@@ -14,32 +14,71 @@ describe( 'translateLink', () => {
...
+			expect( translateLink( 'uk', DOCS_LINKS.sites ) ).toBe(
+				'https://developer.wordpress.com/docs/developer-tools/studio/sites/'
+			);
+			expect( translateLink( 'fr', DOCS_LINKS.sync ) ).toBe(
+				'https://developer.wordpress.com/docs/developer-tools/studio/sync/'
+			);
+		} );
+
+		it( 'should return Spanish URLs when locale is es', () => {
+			expect( translateLink( 'es', DOCS_LINKS.studio ) ).toBe(
```
wojtekn: What's the point of having all links here? Wouldn't be enough to test if the method returns the link for existing translation, and defaults to `en` for a missing one?

src/lib/translate-link.ts
```diff
@@ -28,7 +52,11 @@ const URL_TRANSLATIONS: Map< string, Partial< Record< SupportedLocale, string >
 ] );
 
 export function translateLink( locale: SupportedLocale, url: string ): string {
-	const translation = URL_TRANSLATIONS.get( url )?.[ locale ];
```
wojtekn, line 31: Following the single responsibility principle, and considering that the two implementations differ significantly, I think it would make sense to have two separate files, one for `getDocsLink` and another for `getBlogLink`. Alternatively, have `getDocsLink` which dynamically changes the language, and another one, e.g. `getLink` that uses the mapping array.

src/lib/translate-link.ts
sejas: Yep, it makes sense. Initially I stored as a key so the translateLink function could work with any URL, and not a subset.

src/menu.ts
sejas: 👍 changed!

src/lib/tests/translate-link.test.ts
sejas: I removed duplicated tests, but I preferred to leave a couple of examples per each test.

src/lib/translate-link.ts
sejas: Good point. So I opted to share the same logic and architecture. Now the translatable links use a dict/object. Types force to have the english link as a type requirement. We only have a unique function to "translate" links, which I think is important for consistency.

sejas:
@wojtekn , thanks for the feedback. I really appreciate it. Now the code it's cleaner. We are forced to add all the links into the translation array. The sam function `getLink` can be used to generate docs and blog links, and potentially any kind of links.

Let me know what you think.

src/lib/get-link.ts
```diff
@@ -0,0 +1,63 @@
...
+		en: 'https://developer.wordpress.com/docs/developer-tools/studio/cli/',
+	},
+} satisfies Record< string, TranslatedLink >;
+
+const BLOG_LINKS = {
+	blogPhpVersions: {
+		en: 'https://wordpress.com/blog/2025/03/17/studio-wordpress-php-versions/',
+		es: 'https://wordpress.com/es/blog/2025/04/02/modifica-las-versiones-de-wordpress-y-php-de-tu-sitio-local-con-studio/',
+		fr: 'https://wordpress.com/fr/blog/2025/03/28/studio-wordpress-php-versions/',
+		'pt-br': 'https://wordpress.com/pt-br/blog/2025/04/02/estudio-wordpress-php-versoes/',
```
sejas: Having the full links makes it easier to test whether they work or return a 404.

src/lib/get-link.ts
```diff
@@ -0,0 +1,63 @@
...
+		fr: 'https://wordpress.com/fr/blog/2025/03/04/sites-de-previsualisation-studio/',
+		ja: 'https://wordpress.com/ja/blog/2025/03/28/studio-preview-sites/',
+		'pt-br': 'https://wordpress.com/pt-br/blog/2025/03/06/estudio-visualizar-sites/',
+	},
+	blogCustomDomainsHttps: {
+		en: 'https://wordpress.com/blog/2025/03/31/studio-custom-domains-https/',
+		es: 'https://wordpress.com/es/blog/2025/03/31/studio-custom-domains-https/',
+		'pt-br': 'https://wordpress.com/pt-br/blog/2025/04/03/estudio-dominios-personalizados-https/',
+	},
+} satisfies Record< `blog${ string }`, TranslatedLink >;
```
sejas: These types force the link keys to start with `blog` and `docs`, respectively.

src/lib/get-link.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const LINKS = {
+	...BLOG_LINKS,
+	...DOCS_LINKS,
+} as const satisfies Record< string, TranslatedLink >;
```
fredrikekelund: Judging by some quick experimentation, it looks like using the `satisfies` operator here was a good call 👍 I haven't used it before, though, so it'd be interesting to understand better why it's a good fit in this particular case.

src/lib/get-link.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Returns the link for the given locale if it exists, otherwise, returns the English link.
+ */
+export function getLink( locale: SupportedLocale, linkKey: keyof typeof LINKS ): string {
```
fredrikekelund:
```suggestion
export function getWordPressComUrl( locale: SupportedLocale, linkKey: keyof typeof LINKS ): string {
```
Just a suggestion, but `getLink` is not easily understandable. A more verbose name could help make the code easier to understand.

src/lib/tests/get-link.test.ts
```diff
@@ -0,0 +1,60 @@
+// To run tests, execute `npm run test -- src/lib/tests/translate-link.test.ts` from the root directory
+
```
fredrikekelund:
```suggestion
```
I suggest just removing this part.

fredrikekelund (approved):
Nice improvement 👍 Left some minor suggestions.

src/lib/get-link.ts
wojtekn: One last nit on method signature - if we want to enforce a prefix, would it be clearer to change the method signature and add another parameter for that?

```
 getLink( 'en', 'docs', 'importExport' )
 ```

Alternatively, what if we skipped enforcing that, and used one array for both docs and blog links, making it generic and not tied to the resource type it stores?

```
const LINKS = {
	blogPhpVersions: {
		[...]
	},
	docsPhpVersions: {
		[...]
	},
	dotOrgBlogGutenber: {
		[...]
	},
}

wojtekn (approved):
Looks good, thanks for simplifying that!

src/lib/get-link.ts
wojtekn: We can consider a more detailed name, but still generic, e.g., `getLocalizedLink`.

nightnei (approved):
LGTM and works well 👍
TIL `satisfies`

src/lib/tests/get-link.test.ts
sejas: Removed! https://github.com/Automattic/studio/pull/1375/commits/08f3cb29a78961829290e59c3d56160fe0428afe

src/lib/get-link.ts
sejas: Great naming! Changed here https://github.com/Automattic/studio/pull/1375/commits/a29f92f8cd904278c83f4b8a7634eda7a147781c
When creating the PR I was thinking on `getTranslatedLInk` 👎 and that's why I opted for `getLink`.

src/lib/get-link.ts
sejas: I prefer having everything in the same object. I'm not sure if forcing a prefix is useful, but I see that it can help with decision-making when seeing the autocomplete while calling the function.

Screenshot 2025-05-13 at 09 49 07

I'll leave it as it is for now. We can always merge the objects and remove the prefix check later.

src/lib/get-link.ts
sejas: Yeah! Satisfies is great for this use case because I wanted to enforce using the locale keys and ensure having at least the `en` translation while keeping the magic of autocomplete for main keys. With regular types, the object keys would be defined as dull strings, which would break the autocomplete.

src/lib/get-link.ts
wojtekn: Sounds good, thanks for sharing this use case.

PR #1376, opened 2025-05-12
Studio: Handle missing WP server-files

katinthehatsite:
## Related issues

Fixes STU-456

## Proposed Changes

Currently, if there are some WP server-files are missing, Studio will attempt to create a site and then site start will fail. 
I was not able to reproduce the exact error reported by users but I think we can enhance the current behaviour by running checksum with CLI before creating a site. This will ensure that all the WP files are in order. If some files are missing, we can attempt to download a specific version of WordPress again. 

## Testing Instructions

* Navigate to `/Users/{yourUser}/Library/Application Support/Studio/server-files/wordpress-versions/latest`
* Delete `wp-config-sample.php` from there
* Click on `Add site` 
* Select the WP version to be `latest`
* Observe that the checksum check being outputted in the terminal
* Confirm that the site is created and started successfully:

Screenshot 2025-05-13 at 4 51 34 PM

## Pre-merge Checklist


src/lib/wp-versions.ts
```diff
@@ -97,3 +99,35 @@ export async function purgeWpConfig( wpVersion: string ) {
...
+
+/**
+ * Verifies the checksums of a WordPress installation.
+ * If checksums don't match, it will retry the download once.
+ *
+ * @param wpVersion - The WordPress version to verify
+ */
+export async function verifyWordPressChecksums( wpVersion: string ): Promise< void > {
+	if ( ! net.isOnline() ) {
+		console.log( 'Skipping WordPress checksum verification - offline mode' );
```
katinthehatsite, line 111: I added this check to make sure it does not run when the user is offline as they would need to have connection for the checksums to run correctly. Maybe there is a better approach. I will leave it for now

src/lib/wp-versions.ts
```diff
@@ -97,3 +99,35 @@ export async function purgeWpConfig( wpVersion: string ) {
...
+		] );
+
+		if ( result.exitCode !== 0 ) {
+			console.log( `Checksums don't match for WordPress ${ wpVersion }, retrying download...` );
+			await fs.remove( getWordPressVersionPath( wpVersion ) );
+			// Retry download one more time
+			await downloadWordPress( wpVersion, { overwrite: true } );
+		}
+	} catch ( error ) {
+		console.error( `Failed to verify WordPress checksums:`, error );
```
nightnei, line 131: If there are already issues with checksum and if downloading Wordpress is failed, probably we should not mute the error of downloading and throw it outside, so that it could be watched [here](https://github.com/Automattic/studio/blob/trunk/src/ipc-handlers.ts#L186).
WDYT?
```suggestion
		console.error( `Failed to verify WordPress checksums:`, error );
		throw error;
```

nightnei: I have tested - we indded have godd handling erors here and if we add `throw error;` then we will get this result in UI: Screenshot 2025-05-14 at 14 18 16

So I would throw error outside to avoid muting it.

katinthehatsite: Nice, I will make adjustments based on your suggestion 👍 It would definitely to avoid muting it. 
Yeah, I discovered a couple of strange things with downloading WordPress so I am slowly uncovering those 😅

nightnei (approved):
Works well and LGTM 👍
Just let's throw error outside (I tested it, it will work well).

src/lib/wp-versions.ts
katinthehatsite: @nightnei I think actually your initial suggestion to throw the error inside try/catch was good. I found that we were swallowing this error later in the try/catch in `site-server.ts`. I fixed it there and now the error is properly propagating with your initial suggestion. You can test it by applying this diff: 


```diff --git a/src/lib/wp-versions.ts b/src/lib/wp-versions.ts
index ffa2033d..b19332e5 100644
--- a/src/lib/wp-versions.ts
+++ b/src/lib/wp-versions.ts
@@ -114,6 +114,7 @@ export async function verifyWordPressChecksums( wpVersion: string ): Promise< vo
 
        try {
                console.log( `Verifying checksums for WordPress version ${ wpVersion }...` );
+               throw new Error( 'Forced error for testing checksum verification' );
                const result = await executeWPCli( getWordPressVersionPath( wpVersion ), [
                        'core',
                        'verify-checksums',
```

PR #1380, opened 2025-05-13
Pull: Move progress functions to the existing progress custom hook

epeicher:
## Related issues

- Related to STU-399

## Proposed Changes

- Move the existing `getBackupProgress` and `getStatusWithProgress` to the existing `useSyncStatesProgressInfo` hook, which exposes utility functions to return Progress. This seems to be a better location for those functions, more consistent with existing code. 

## Testing Instructions

- Apply this changes and run `npm start`
- Check the Pull process work exactly as in Production and no unexpected errors are triggered.

## Pre-merge Checklist


bcotrim (approved):
Works as described
LGTM 👍

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -130,12 +137,59 @@ export function useSyncStatesProgressInfo() {
...
+				return { message: sitePullState.message, progress: sitePullState.progress };
+			}
+			if ( importState ) {
+				if ( importState.progress === 100 ) {
+					return { message: __( 'Applying final details…' ), progress: 99 };
+				}
+				return {
+					message: importState.statusMessage,
+					progress:
+						IMPORTING_INITIAL_VALUE + IMPORTING_TO_FINISHED_STEP * ( importState.progress / 100 ),
```
fredrikekelund, line 175: What do we accomplish with this calculation? `IMPORTING_INITIAL_VALUE` is always 80 and `IMPORTING_TO_FINISHED_STEP` is always 20, so it looks like we're saying `100 * (80 / 100)`. Why not just return `importState.progress`..?

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -130,12 +137,59 @@ export function useSyncStatesProgressInfo() {
...
+			response: SyncBackupResponse
+		) => {
+			const frontendStatus = hasBackupCompleted
+				? pullStatesProgressInfo.downloading.key
+				: response.status;
+			let newProgressInfo: PullStateProgressInfo | null = null;
+			if ( response.status === 'in-progress' ) {
+				newProgressInfo = pullStatesProgressInfo[ frontendStatus ];
+				newProgressInfo.progress =
+					IN_PROGRESS_INITIAL_VALUE + IN_PROGRESS_TO_DOWNLOADING_STEP * ( response.percent / 100 );
```
fredrikekelund, line 155: Same thing here.

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -130,12 +137,59 @@ export function useSyncStatesProgressInfo() {
...
+			if ( importState ) {
+				if ( importState.progress === 100 ) {
+					return { message: __( 'Applying final details…' ), progress: 99 };
+				}
+				return {
+					message: importState.statusMessage,
+					progress:
+						IMPORTING_INITIAL_VALUE + IMPORTING_TO_FINISHED_STEP * ( importState.progress / 100 ),
+				};
+			}
```
fredrikekelund:
```suggestion
			if ( importState ) {
				if ( importState.progress === 100 ) {
					return { message: __( 'Applying final details…' ), progress: 99 };
				}
				return {
					message: importState.statusMessage,
					progress:
						IMPORTING_INITIAL_VALUE + IMPORTING_TO_FINISHED_STEP * ( importState.progress / 100 ),
				};
			}
			if ( sitePullState ) {
				return { message: sitePullState.message, progress: sitePullState.progress };
			}
```
Nitpicking, but changing the order of the `if` statements clarifies intent.

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -130,12 +137,59 @@ export function useSyncStatesProgressInfo() {
 		[]
 	);
 
+	const getBackupStatusWithProgress = useCallback(
+		(
+			hasBackupCompleted: boolean,
```
fredrikekelund, line 142: Couldn't we intuit this value inside `getBackupStatusWithProgress` by checking if `response.status === 'finished'`?

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -130,12 +137,59 @@ export function useSyncStatesProgressInfo() {
...
+			let newProgressInfo: PullStateProgressInfo | null = null;
+			if ( response.status === 'in-progress' ) {
+				newProgressInfo = pullStatesProgressInfo[ frontendStatus ];
+				newProgressInfo.progress =
+					IN_PROGRESS_INITIAL_VALUE + IN_PROGRESS_TO_DOWNLOADING_STEP * ( response.percent / 100 );
+			}
+			const statusWithProgress =
+				newProgressInfo ||
+				pullStatesProgressInfo[ frontendStatus ] ||
+				pullStatesProgressInfo.failed;
```
fredrikekelund:
```suggestion
			const statusWithProgress =
				newProgressInfo ||
				pullStatesProgressInfo[ frontendStatus ];
```
Typescript already guarantees that `frontendStatus` is a key of `pullStatesProgressInfo`, so we don't need to default to `pullStatesProgressInfo.failed`.

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -17,6 +18,12 @@ export type PullStateProgressInfoValues = Record<
 	PullStateProgressInfo
 >;
 
+export type SyncBackupResponse = {
+	status: 'in-progress' | 'finished' | 'failed';
+	download_url: string;
+	percent: number;
+};
+
```
fredrikekelund, line 26: This type is also defined in `src/hooks/sync-sites/use-sync-pull.ts`. Could we use a single definition, and, while we're at it, maybe convert it to a zod schema?

fredrikekelund (commented):
I left a couple of comments on the code. I know a lot of that stuff was the same prior to this PR, but this seems like a good opportunity to address it. 

If you're not already familiar with the `ProgressBarWithAutoIncrement` component, I'd also like to mention that. We use it for preview sites and it provides some nice interactivity. Granted, sync operations typically take longer, but you could give it a small `increment` prop to account for that.

src/hooks/use-sync-states-progress-info.ts
epeicher: Not really, `80 + 20 * progress / 100 != progress` 😛 . This tries to move from 80% to 100% progressively determined by `importState.progress`. I think I can add a comment there to clarify, as it can be difficult to read.

src/hooks/use-sync-states-progress-info.ts
epeicher: Same reasoning, I think some comments will help here, I'm adding them

src/hooks/use-sync-states-progress-info.ts
epeicher: I agree, the suggesions has less conditions

src/hooks/use-sync-states-progress-info.ts
epeicher: Good catch, I meant to delete it from `src/hooks/sync-sites/use-sync-pull.ts` and declare it here instead to avoid a circular dependency. Alternatively, I can move it to an external types file and import it from there in both places (here and in `use-sync-pull`). Do you prefer one option over the other?

src/hooks/use-sync-states-progress-info.ts
epeicher: Good point, I originally passed it as a parameter because it was already obtained [here](https://github.com/Automattic/studio/blob/3594007042cb3a53a76207654ab90de9566c0d42/src/hooks/sync-sites/use-sync-pull.ts#L284), but in this case, it could be better to calculate it the function and avoid the extra parameter, do you agree?

src/hooks/use-sync-states-progress-info.ts
epeicher: Good point, I've accepted the suggestion

src/hooks/use-sync-states-progress-info.ts
fredrikekelund: Ofc, my bad 🤦‍♂️ Doesn't seem as bad now that I come back to it, though. Maybe assigning the `IMPORTING_TO_FINISHED_STEP * ( importState.progress / 100 )` part to a separate variable first would be enough to make it easier to understand.

src/hooks/use-sync-states-progress-info.ts
epeicher: For reference, I've deleted the redundant definition in this commit [2241fd6](https://github.com/Automattic/studio/pull/1380/commits/2241fd6791019219722eb00aa79c85f9e16c638f)

epeicher:
I will merge this as I have addressed most of the comments. If there are any other suggestions, please let me know, and I will create a follow-up task for that

Issue #1377, opened 2025-05-12
Studio CLI commands don't work from root folders of projects


juanmaguitar:
### Quick summary

In the [Studio CLI Docs](https://developer.wordpress.com/docs/developer-tools/studio/cli/) it states 
> You can run these commands from the root directory of any Studio site

If I open the Terminal on any Studio site, the Studio CLI commands should directly work on the opened folder (root folder of the project). Instead i get the following error:

```
⬢  django  studio preview list                              
✖ The specified folder is not added to Studio.
```

<details>
<summary>See the content of <code>~/Library/Application\ Support/Studio/appdata-v1.json</code></summary>

```
{
  "version": 1,
  "sites": [
    {
      "id": "9eba78a1-f861-47a3-b0c2-301424ef81e3",
      "name": "Django",
      "path": "/Users/juanmanuelgarrido/Studio/django",
      "adminPassword": "NElQcXl2dHhuSjEkZ1FLcypSbiNkZTIp",
      "port": 8885,
      "phpVersion": "8.2",
      "isWpAutoUpdating": true,
      "customDomain": "django.wp.local",
      "enableHttps": true,
      "autoStart": true,
      "themeDetails": {
        "name": "George Lois",
        "path": "/var/www/html/wp-content/themes/george-lois",
        "slug": "george-lois",
        "isBlockTheme": true,
        "supportsWidgets": false,
        "supportsMenus": false
      }
    },
    {
      "id": "d6cf80f9-aeeb-4ecc-b96d-698df01e4a34",
      "name": "Murakami",
      "path": "/Users/juanmanuelgarrido/Studio/murakami",
      "adminPassword": "VGRLNXchNm1HeGdiVlVeKHZydCF6Wk8p",
      "port": 8883,
      "phpVersion": "8.2",
      "isWpAutoUpdating": true,
      "enableHttps": false,
      "autoStart": false,
      "themeDetails": {
        "name": "Twenty Twenty-Five",
        "path": "/var/www/html/wp-content/themes/twentytwentyfive",
        "slug": "twentytwentyfive",
        "isBlockTheme": true,
        "supportsWidgets": false,
        "supportsMenus": false
      }
    },
    {
      "id": "9c80f0d8-e1bd-4c11-8dd9-217bc4ab7ded",
      "name": "Tarantino",
      "path": "/Users/juanmanuelgarrido/Studio/tarantino",
[...]

PR #1381, opened 2025-05-13
Compare paths in a case-insensitive manner

fredrikekelund:
## Related issues

- Fixes STU-487

## Proposed Changes

Our existing way of looking up Studio sites produces unexpected results on case-**insensitive** file systems if users have renamed any of the directories in the path to use a different case. Here's a concrete example:

- I'm using macOS's default file system.
- I have a Studio site in `/Users/fredrik/Studio/my-wordpress-website`.
- I rename the `Studio` folder to `STUDIO`.
- Studio continues to function normally, since the path still resolves to an actual location on disk.
- However, the CLI no longer works, since `process.cwd()` returns `/Users/fredrik/STUDIO/my-wordpress-website` and my Studio config file contains `/Users/fredrik/Studio/my-wordpress-website`.

This PR fixes the issue by using `fs.stat` and comparing `fs.Stats.dev` and `fs.Stats.ino` between the CLI path and the config file path. 

**Moreover**, I've updated `src/hooks/use-add-site.ts`. This hook would previously rely on comparing paths by always converting them to lowercase. We now use the same logic there as we do in the CLI.

## Testing Instructions

Test the CLI:
1. Run `npm run cli:build`.
2. Run `node /dist/cli/main.js preview list --path ~/Studio/my-wordpress-website` (update the `--path` option to point to a Studio website).
3. Run `node /dist/cli/main.js preview list --path ~/STUDIO/my-wordpress-website` (use the same `--path` option, but change the case in some part of it).

Test the app:
1. Run `npm start`.
2. Click the `Add site` button.
3. Enter the name of a site that's already added to Studio (and that resides in Studio's default site directory).
4. Ensure that an error is displayed.
5. Change the site name to something else.
6. Manually specify a site path. Choose a site path that's already occupied by an existing Studio site. 
7. Ensure that an error is displayed.

## Pre-merge Checklist

### DIFF
--- a/common/lib/fs-utils.ts
+++ b/common/lib/fs-utils.ts
@@ -51,6 +51,8 @@ export function isWordPressDirectory( projectPath: string ): boolean {
 
 // Compare paths in a case-insensitive manner. `fs.Stats.dev` signifies the device ID, and
 // `fs.Stats.ino` signifies the inode number that uniquely identifies the file or directory.
+// The benefit of this approach over converting the entire path to lowercase is that it respects
+// the current file system's case sensitivity.
 export function arePathsEqual( path1: string, path2: string ) {
 	try {
 		const stats1 = fs.statSync( path.resolve( path1 ) );
--- a/common/lib/fs-utils.ts
+++ b/common/lib/fs-utils.ts
@@ -1,6 +1,5 @@
 import fs from 'fs';
 import path from 'path';
-import { existsSync } from 'fs-extra';
 
 /**
  * Calculates the total size of a directory by recursively traversing its contents.
@@ -44,8 +43,21 @@ export function calculateDirectorySize( directoryPath: string ): Promise< number
 
 export function isWordPressDirectory( projectPath: string ): boolean {
 	return (
-		existsSync( path.join( projectPath, 'wp-content' ) ) &&
-		existsSync( path.join( projectPath, 'wp-includes' ) ) &&
-		existsSync( path.join( projectPath, 'wp-load.php' ) )
+		fs.existsSync( path.join( projectPath, 'wp-content' ) ) &&
+		fs.existsSync( path.join( projectPath, 'wp-includes' ) ) &&
+		fs.existsSync( path.join( projectPath, 'wp-load.php' ) )
 	);
+}
+
+// Compare paths in a case-insensitive manner. `fs.Stats.dev` signifies the device ID, and
+// `fs.Stats.ino` signifies the inode number that uniquely identifies the file or directory.
+export function arePathsEqual( path1: string, path2: string ) {
+	try {
+		const stats1 = fs.statSync( path.resolve( path1 ) );
+		const stats2 = fs.statSync( path.resolve( path2 ) );
+
+		return stats1.ino === stats2.ino && stats1.dev === stats2.dev;
+	} catch ( error ) {
+		return false;
+	}
 }
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -17,7 +17,7 @@ import nodePath from 'path';
 import * as Sentry from '@sentry/electron/main';
 import { __, LocaleData, defaultI18n } from '@wordpress/i18n';
 import archiver from 'archiver';
-import { calculateDirectorySize, isWordPressDirectory } from 'common/lib/fs-utils';
+import { calculateDirectorySize, isWordPressDirectory, arePathsEqual } from 'common/lib/fs-utils';
 import { SupportedLocale } from 'common/lib/locale';
 import { getAuthenticationUrl } from 'common/lib/oauth';
 import { Snapshot } from 'common/types/snapshot';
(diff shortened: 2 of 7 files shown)


gavande1 (approved):
It works as expected. LGTM 👍 
CleanShot 2025-05-13 at 15 14 43@2x

fredrikekelund:
@wojtekn, just confirming that you are also OK with this approach, given our discussion in https://github.com/Automattic/studio/issues/1377?

wojtekn:
@fredrikekelund sure, looks good. What do you think about updating the older code to use the unified approach?

fredrikekelund:
Makes sense 👍 I'll take a stab at that in this PR before landing.

fredrikekelund:
I've updated the PR to also cover `useAddSite`. Would appreciate another review 🙏

wojtekn (approved):
The code change looks good, and it works as expected.

common/lib/fs-utils.ts
```diff
@@ -44,8 +43,23 @@ export function calculateDirectorySize( directoryPath: string ): Promise< number
...
-		existsSync( path.join( projectPath, 'wp-content' ) ) &&
-		existsSync( path.join( projectPath, 'wp-includes' ) ) &&
-		existsSync( path.join( projectPath, 'wp-load.php' ) )
+		fs.existsSync( path.join( projectPath, 'wp-content' ) ) &&
+		fs.existsSync( path.join( projectPath, 'wp-includes' ) ) &&
+		fs.existsSync( path.join( projectPath, 'wp-load.php' ) )
 	);
 }
+
+// Compare paths in a case-insensitive manner. `fs.Stats.dev` signifies the device ID, and
```
wojtekn, line 52: Thanks for adding the explanation.

src/ipc-handlers.ts
```diff
@@ -1409,3 +1409,7 @@ export async function handleNewSite( event: IpcMainInvokeEvent, newSite: NewSite
 		newSites: userData.newSites?.filter( ( s ) => s.id !== newSite.id ),
 	} );
 }
+
+export function comparePaths( event: IpcMainInvokeEvent, path1: string, path2: string ) {
```
nightnei, line 1413: I am thinking - should we go with the same naming for internal and for IPC?
Or, mauyb e as option woudl be better to come up with some rule to follow inside our team, e.g. - if you want to expose some function, use the same name but with some prefix or something like that.
WDYT?
```suggestion
export function arePathsEqual( event: IpcMainInvokeEvent, path1: string, path2: string ) {
```

common/lib/fs-utils.ts
```diff
@@ -44,8 +43,23 @@ export function calculateDirectorySize( directoryPath: string ): Promise< number
...
+// `fs.Stats.ino` signifies the inode number that uniquely identifies the file or directory.
+// The benefit of this approach over converting the entire path to lowercase is that it respects
+// the current file system's case sensitivity.
+export function arePathsEqual( path1: string, path2: string ) {
+	try {
+		const stats1 = fs.statSync( path.resolve( path1 ) );
```
nightnei, line 58: I noticed that we are friquently use `sync` options of functions. In this case I think it's not a problem, but I am thinking - should we have some rule to try to always use async versions, to avoid any potential cases to block event loop?

nightnei (approved):
LGTM, just left few comments

src/ipc-handlers.ts
fredrikekelund: I agree they would ideally use the same name, this was just my lazy choice to work around the problem of name collisions in `ipc-handlers.ts`. I hit "merge" _just_ before you submitted your review, so I would just leave this as is.

nightnei: Sure, let's leave it as is, I just wanted to raise this discussion for future cases

common/lib/fs-utils.ts
fredrikekelund: `fs.exists` is deprecated, but `fs.existsSync` is not, so that's why we'd use the sync version of that particular function. As for `fs.statSync`, we could technically use the async version. Given that there's only a single client to this code (this isn't a web server handling multiple clients in parallel), I think we can be more liberal with using sync functions than we would otherwise.

Still, it would be very simple to change these calls to their async counterparts, so I agree it would have been better if I had done that. This code lived in just the CLI to begin with, where the demand for asynchrony is even lower, and I think I just overlooked that part when moving it to `./common`

nightnei: Yeah, totally agree about not having any issues with CLI.
It was more general discussion - to have an agreement inside our team to try to always use async versions, to have consistency and avoid potential cases when some sync version can be unintentionally used inside main app.

Btw, interesting thing - https://stackoverflow.com/questions/31799274/node-js-fs-exists-will-be-deprecated-what-to-use-instead

```
The [io.js docs](https://iojs.org/api/fs.html#fs_fs_exists_path_callback#fs_fs_exists_path_callback) mention the use of fs.stat() or fs.access() in place of fs.exists().
```

fredrikekelund: > some sync version can be unintentionally used inside main app

Since the node.js APIs aren't available in the renderer process, I don't think it's technically possible to block the renderer by using those functions. 

I think we can be pragmatic about using sync functions, but I agree that we should always default to the async versions.

PR #1383, opened 2025-05-13
Studio: Force latest WP version for creating site if the user goes offline

katinthehatsite:
## Related issues

Fixes STU-494

## Proposed Changes

Currently, if the user opens the `Add site` form, chooses a custom WP version that needs to be downloaded and then their connection cuts, Studio attempts to download that WP version and this results in a gibberish error. 

I propose to force the `latest` version for creating a site if the user is offline since this comes with Studio already and should work without connection.

## Testing Instructions

* Click on `Add site` in the sidebar
* Select the WP version from the dropdown that is not `latest`
* Keep the form open
* Cut your internet access
* Observe that the option in the dropdown gets set to `latest`
* Confirm that the site gets created and starts successfully with the latest version
* You can repeat the same steps on trunk and confirm that you can attempt site creation with a custom WP version which results into gibberish error when the app attempts downloading WordPress

## Pre-merge Checklist

### DIFF
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -424,6 +424,9 @@ export const SiteForm = ( {
 												fallbackOptions={ [
 													{ label: __( 'Latest' ), value: DEFAULT_WORDPRESS_VERSION },
 												] }
+												offlineMessage={ __(
+													'You are currently offline so your site will be created with the latest version. Selecting a different WordPress version requires an internet connection.'
+												) }
 											/>
 										</div>
 
--- a/src/components/tests/add-site.test.tsx
+++ b/src/components/tests/add-site.test.tsx
@@ -416,7 +416,9 @@ describe( 'AddSite', () => {
 		await user.hover( wpVersionSelect );
 
 		expect(
-			screen.getByText( 'Changing WordPress version requires an internet connection.' )
+			screen.getByText(
+				'You are currently offline so your site will be created with the latest version. Selecting a different WordPress version requires an internet connection.'
+			)
 		).toBeInTheDocument();
 	} );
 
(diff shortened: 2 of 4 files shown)


katinthehatsite:
I left the tooltip as currently is when offline e.g. `Changing WP version requires an internet connection` because I thought it is fitting but we can refine it if we consider this necessary.

bcotrim (commented):
Worked as described, nice catch!

Have you considered treating it as an error and disable the button if the version is not `latest` and is offline? All of this behavior is an edge case, but I could see a scenario where the connection fails for a short time and we change the version without the user noticing it. Having an error would make it more visible.

katinthehatsite:
> Have you considered treating it as an error and disable the button if the version is not latest and is offline? All of this behavior is an edge case, but I could see a scenario where the connection fails for a short time and we change the version without the user noticing it. Having an error would make it more visible.

I am in two minds about this because we sort of block the changing of versions already when editing site instead of allowing the user to trigger the error. Perhaps, a middle ground could be to change the tooltip to something like "You are currently offline and can create a site only with the latest version. Selecting a different version requires an internet connection".

Alternatively, we can also source more opinions. On my end, I don't have a strong preference.

bcotrim:
I like that suggestion!
Just to confirm that I understood correctly, we would adjust the tooltip as you suggested, and on submission always use the `latest` version if the user is offline, is that correct?

katinthehatsite:
> Just to confirm that I understood correctly, we would adjust the tooltip as you suggested, and on submission always use the latest version if the user is offline, is that correct?

Yeah exactly!

gavande1 (approved):
It works as expected. Thanks for the improvement. LGTM 👍 

CleanShot 2025-05-13 at 19 02 18@2x

wojtekn (approved):
The code change looks clear, works as expected. Nice improvement!

PR #1388, opened 2025-05-13
Make default test run command silent and add verbose one

wojtekn:
## Related issues

- Fixes STU-493

## Proposed Changes

- I propose to make default test run command silent and add a verbose one.

## Testing Instructions

Run tests:
```
npm run test
npm run test:verbose
```

## Pre-merge Checklist


gcsecsey (approved):
LGTM

| `npm run test` | `npm run test:verbose` |
|--------|---|
| CleanShot 2025-05-13 at 17 30 25@2x | CleanShot 2025-05-13 at 17 30 58@2x |

bcotrim (commented):
I don't think running the tests silently is a good approach. At some point it will hide errors, that we could have fixed.
At the time of reviewing this PR, the difference between running `npm run test` and `npm run test:verbose` are some console logs. If those console.logs aren't offering anything to the test, we should mock them (we could maybe assert if the messages are being written if that's important) and achieve the same "silent" state for our tests.

This is a much better state to be in, our tests don't produce any "noise" and when they do we will immediately notice it and have the opportunity to fix it.

katinthehatsite:
> I don't think running the tests silently is a good approach. At some point it will hide errors, that we could have fixed.

+ 1

wojtekn:
Closing in favor of https://github.com/Automattic/studio/pull/1393

PR #1389, opened 2025-05-13
Push: Read and update the progress from the sync endpoint

epeicher:
## Related issues

- Relates STU-399

## Proposed Changes

- Calls the GET `/sync/import` using the `importId` to obtain the progress
- Use the progress from the `/sync/import` endpoint to update the progress in the `importing` state 
- Splits the progress of the `importing` state between `backup` and `import`
- The `backup` step will take 10 steps (from 100) in the progress bar, from 60 to 70, and it will be increased by the `backup_progress` returned by the endpoint.
- The `import` step will take 30 steps in the progress bar, from 70 to 100, and it will be increased by the `import_progress` returned by the endpoint.

This is a video with the numeric progress added for debugging purposes:

https://github.com/user-attachments/assets/17d89f2b-f922-4bdd-8952-7309983197c6

## Testing Instructions

- Apply the changes in 182803-ghe-Automattic/wpcom and sandbox your environment
- Apply changes in this branch and run `npm start`
- Create a site with studio or use an existing one, and connect it to a site in WordPress.com
- Go to the Sync tab and run a Push
- Check the progress moves in the `Applying changes…` after some seconds
- Additionally, you can apply this diff for debugging purposes to see a numerical progress
```diff
diff --git i/src/components/sync-connected-sites.tsx w/src/components/sync-connected-sites.tsx
index c4bb29b1..a94b7c0c 100644
--- i/src/components/sync-connected-sites.tsx
+++ w/src/components/sync-connected-sites.tsx
@@ -248,6 +248,7 @@ const SyncConnectedSitesList = ( { selectedSite }: SyncConnectedSitesListProps )
 											sitePullState.status.message }
 									</div>
 									<ProgressBar value={ sitePullState.status.progress } maxValue={ 100 } />
+									<p>{ sitePullState.status.progress }</p>
 								</div>
 							) }
 							{ isPullError && (

```

### Regression testing
- Apply the changes in 182803-ghe-Automattic/wpcom and sandbox your environment
- Open a production version of Studio
- Create a site with studio or use an existing one, and connect it to a site in WordPress.com
- Go to the Sync tab and run a Push
- Check there are no unexpected errors

## Pre-merge Checklist


src/hooks/use-sync-states-progress-info.ts
```diff
@@ -24,6 +29,18 @@ export type SyncBackupResponse = {
 	percent: number;
 };
 
+export type ImportResponse = {
+	status:
+		| 'finished'
+		| 'failed'
+		| 'initial_backup_started'
+		| 'archive_import_started'
```
sejas, line 37: I see this status is new.

sejas (commented):
I tested it with the backend API PR and it worked great. Here is a short video displaying the progress in number.

https://github.com/user-attachments/assets/33d3ae8b-cf0e-4369-a5b2-2ba6f1c7b81f

Let's not merge this PR until the backend PR is merged.

wojtekn:
Is it okay that we display '100%' and 'Applying changes...' for a while?

epeicher:
> Is it okay that we display '100%' and 'Applying changes...' for a while?

For the Pull process, [we added one step](https://github.com/Automattic/studio/blob/7540a42357ae5e1c3fb261ef507bc45628762ec7/src/hooks/use-sync-states-progress-info.ts#L188) `Applying final details…` for that. We can do a similar thing here, but in principle that 100% should not take more than some seconds.

src/hooks/use-sync-states-progress-info.ts
epeicher: Yes, I added `archive_import_finished` to [handle it here](https://github.com/Automattic/studio/blob/7540a42357ae5e1c3fb261ef507bc45628762ec7/src/hooks/use-sync-states-progress-info.ts#L207), but as @wojtekn  [asks here](https://github.com/Automattic/studio/pull/1389#issuecomment-2880570952), we can keep the progress under 100 % during that status and the `finished` one. What do you think?

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -17,12 +17,14 @@ import {
...
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import type { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
+import type { ImportResponse } from 'src/hooks/use-sync-states-progress-info';
 
 export type SyncPushState = {
 	remoteSiteId: number;
 	status: PushStateProgressInfo;
 	selectedSite: SiteDetails;
 	isStaging: boolean;
+	importId?: number;
```
sejas: I think we should remove this type option.

```suggestion
```

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -24,11 +29,24 @@ export type SyncBackupResponse = {
...
+	import_progress: number;
+	error?: string;
+};
+
 export const IN_PROGRESS_INITIAL_VALUE = 30;
 const DOWNLOADING_INITIAL_VALUE = 60;
 export const IN_PROGRESS_TO_DOWNLOADING_STEP =
 	DOWNLOADING_INITIAL_VALUE - IN_PROGRESS_INITIAL_VALUE;
-export const IMPORTING_INITIAL_VALUE = 80;
+export const IMPORTING_INITIAL_VALUE = 60;
```
sejas: Let's not modify this variable as it's used also in the Pull process:

https://github.com/Automattic/studio/blob/23e81af92f8bce48d5baec708ae20dcec15ed653/src/hooks/use-sync-states-progress-info.ts#L67-L71

We can use the PUSH_ prefix for new constants to avoid this confusion in the future.

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -73,17 +91,17 @@ export function useSyncStatesProgressInfo() {
...
 			uploading: {
 				key: 'uploading',
-				progress: 50,
+				progress: 40,
 				message: __( 'Uploading Studio site…' ),
 			},
 			importing: {
 				key: 'importing',
-				progress: 80,
+				progress: IMPORTING_INITIAL_VALUE,
```
sejas: Similarly to the previous comment, This variable is being used also in the Pull process.

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -96,7 +114,7 @@ export function useSyncStatesProgressInfo() {
 				progress: 100,
 				message: __( 'Error pushing changes' ),
 			},
-		} as const;
+		} satisfies PushStateProgressInfoValues;
```
sejas: Instead of using satisfies, let's use a regular type when declaring the variable.
```suggestion
		}
```
Something like:

```tsx
	const pullStatesProgressInfo: PullStateProgressInfoValues = useMemo( () => {
```

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -244,8 +252,8 @@ export function useSyncPush( {
 
 		Object.entries( pushStates ).forEach( ( [ key, state ] ) => {
 			if ( state.status.key === pushStatesProgressInfo.importing.key ) {
-				intervals[ key ] = setTimeout( () => {
-					void getPushProgressInfo( state.remoteSiteId, state );
+				intervals[ key ] = setTimeout( async () => {
+					await getPushProgressInfo( state.remoteSiteId, state );
```
sejas: Do we need this change?

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -182,6 +200,36 @@ export function useSyncStatesProgressInfo() {
 		[ __ ]
 	);
 
+	const getPushStatusWithProgress = useCallback(
+		( status: PushStateProgressInfo, response: ImportResponse ) => {
+			if ( status.key === 'importing' ) {
+				const backupStep = 10;
+				const archiveInitialValue = IMPORTING_INITIAL_VALUE + backupStep;
+				const archiveStep = IMPORTING_TO_FINISHED_STEP - backupStep;
```
sejas: Probably we need need a different constant for `IMPORTING_TO_FINISHED_STEP`, which is also used on `getPullStatusWithProgress`.

sejas (commented):
@epeicher , great work!! 🙌 
Thanks for iterating on this, I tested it with the endpoint and it works great.

I would like to divide the `Applying changes…` step by adding two more messages?

1. I would like to add a state saying `Keeping your site safe…` when `initial_backup_started`.
2. Let's add a final message when `archive_import_finished` saying something like: `Last touches…` and return progress of 99.

The last state will fix the [problem mentioned by Wojtek](https://github.com/Automattic/studio/pull/1389#issuecomment-2880570952) where the bar is kept in 100% for a few seconds.


https://github.com/user-attachments/assets/55db7dfc-7c36-41b3-a8a7-859db6e514e5

src/hooks/use-sync-states-progress-info.ts
epeicher: The problem with using a type here in the `useMemo` is that typescript will apply _type widening_, to the return type, meaning that it will be more flexible. So for example:
#### 1. Using a type
```jsx
	const pullStatesProgressInfo = useMemo< PullStateProgressInfoValues >( () => {
```
We could have later on something that would be valid for Typescript:
```jsx
	pullStatesProgressInfo.downloading = {
		key: 'in-progress',
		progress: 100,
		message: __( 'Hello' ),
	};
```
Please note that we are assigning to `key` a different key than `downloading`

#### 2. Using satisfies
CleanShot 2025-05-15 at 16 03 53@2x


There is a much longer explanation in this [chatgpt thread](https://chatgpt.com/share/6825f192-8694-800d-902f-757ccea0eb31) but I think the above is the gist of it. I think this applies here because it is the return of `useMemo`, usually defining and using types is the common approach.

And a [ts playground to try ](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBDAnmYcAKBXAzgCwMowCGMwaUEA5lMFlgJIB2AZhHALxwDeAUAJADWwRAC44AcgDG1YgEsGFAEKEJ-DGDFwAPuLUAbCIQAmcihoDcfMOSo0sohhgC2AI2BQLvR7cIVgorDBQJhYAvhbcoJCwCMiomLgExKTW1LSMLABqhLoYNOxwAErAEtCGADx88fhEJGSUqfTMEADa4oKIGgC6ADSV2NVJdTZpTdwAfOHc3CUMAXCOiFlQAKpYJgAqsXCiVYm1KbbpEFk5eRw8cHBSwLLySipqoheXcO2iktIwJveq6r0vcCs9VsogADP8Xl5aD4-OIcMBdPoxNxLiEIXoDMZ5E8US83jowPojCYxBDLkDhnY4ABGcG4y5QrAw97wjq4kLcDnTCCzeALJarEx4WRYJgyM5cXHXW6KZS-HEA-EfG5fO5ytSk+mAg60MFk+beXwshFI9nowmYkwKvFCd4Y4nyTUAikNUS0-WM5lwoTI1GcuBMr6i8VYdD9PbJYEjTLZXJYKb8wgrNbyTYoAB00tVsoeYHyzyV9qxpghLpBdIZhthYlZyK5ieTQpFYpomc+33VefOuMLFodJdxZd1NIrBuhRrhEDrUxnM6AA)

sejas: Thanks for sharing more context. It makes sense.
What do you think if we leave also the `as const` and prevent any object manipulation?
Something like:
`} as const satisfies PushStateProgressInfoValues;`

src/hooks/sync-sites/use-sync-push.ts
epeicher: Not as part of these changes, reverting it. Although I think we can discuss generically if using `void` is a good pattern 🤔

src/hooks/use-sync-states-progress-info.ts
epeicher: Yep. Done in [6bb6aa5](https://github.com/Automattic/studio/pull/1389/commits/6bb6aa5c3ed77c2c5475745a1dfe2b08103a9e0e)

src/hooks/use-sync-states-progress-info.ts
epeicher: Great suggestion, I have added that as part of [332752a](https://github.com/Automattic/studio/pull/1389/commits/332752a4189beedef28add4f03474e2b18b1d64f) and removed two unnecessary exports

epeicher:
Thanks @sejas  for all your suggestions! I think I've applied most of them, and this is ready for another review. I like that we are leaving the code in a much better position. And I also love the additional steps that will provide more feedback to the user.

sejas:
@epeicher , should we update the base of this PR to point to trunk?
Screenshot 2025-05-15 at 16 39 07

epeicher:
> @epeicher , should we update the base of this PR to point to trunk?

Yep, I had stacked these changes on top of https://github.com/Automattic/studio/pull/1380 to avoid distracting this with unrelated changes. Now that I have merged the base branch, this has been rebased to trunk and there should not be any additional changes

sejas (commented):
The new messages set better expectations for the user. Thanks for adding them!

Would be possible to add the messages inside the `pushStatesProgressInfo` object?
I was thinking maybe instead of having different variables like `archiveInitialValue`, we could use the object `pushStatesProgressInfo.archiving.progress + STEP*X`, `pushStatesProgressInfo.importing.progress`

Around the minute 1:30 , I see the progress goes backwards a bit.

https://github.com/user-attachments/assets/7fa870fc-9753-4068-a212-5acad76277a3

epeicher:
>Would be possible to add the messages inside the pushStatesProgressInfo object?

Sure, I thought about that but discarded initially because I didn't wanted to affect other areas and I interepreted them as _substeps_ of the pushStatesProgressInfo object. But I totally see the reasoning of using that object as the _source of truth_ of the states and avoiding the _substeps_ approach. I will work on that.

> I was thinking maybe instead of having different variables like archiveInitialValue, we could use the object pushStatesProgressInfo.archiving.progress + STEP*X, pushStatesProgressInfo.importing.progress

I will iterate over this, maybe I will ping you to check before committing any changes.

> Around the minute 1:30 , I see the progress goes backwards a bit.

I see it in the video, thanks for that. I will have a look at it.

sejas:
Sure! happy to pair or give early feedback.<br>Maybe the issue that the progress goes backwards is solved once we move the "base" progress to the object.

epeicher:
@sejas a quick question on naming 😅 , what do you think about naming the two new steps:
- creatingRemoteBackup 
- applyingChanges

They will be added to `PushStateProgressInfo` instead of the current `importing` step 
CleanShot 2025-05-16 at 11 03 25@2x

Happy to update them to other suggestions

sejas:
They look great! I think applying changes is better than importing 👍.<br>Could we also add the last step of last touches?<br>For that step the label could say: `Almost there…` and maybe the key… `finishing` or `waitingToFinish` …<br>Naming variables is hard! xD

epeicher:
> Could we also add the last step of last touches?

You read my mind! While changing the code I was just about to ask you about it 😄 . I think the approach of centralizing all the states in `pushStatesProgressInfo` is the proper one

epeicher:
@sejas, I have applied your suggestions about the additional steps and the progress changes in [aa2a0b3](https://github.com/Automattic/studio/pull/1389/commits/aa2a0b3cdb36a273738f0009553b3069875fceb8). I think is better than using the constants, we can apply the same approach to the `Pull` process (as part of a follow-up 😄).

I plan to add some tests that check at least that the ProgressBar appears when in progress, what do you think?

Regarding the progress moving backwards, I have not been able to reproduce the exact issue, but one hypothesis could be that the backend reports a decreasing percentage, in that case, the progress would represent that. If you are able to reproduce that consistently, could you please check the `/sync/import` endpoint call and check the `backup_progress` field in the Network tab?

Additionally, we can add a guard [here](https://github.com/Automattic/studio/blob/aa2a0b3cdb36a273738f0009553b3069875fceb8/src/hooks/use-sync-states-progress-info.ts#L227) that ensures that the progress only progresses forward but I would only add that if we confirm that the backend reports incorrect percentages.

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -183,14 +222,58 @@ export function useSyncStatesProgressInfo() {
 		[ __ ]
 	);
 
+	const getPushStatusWithProgress = useCallback(
+		( status: PushStateProgressInfo, response: ImportResponse ) => {
+			if ( status.key === pushStatesProgressInfo.creatingRemoteBackup.key ) {
+				const creatingRemoteBackupStep =
```
sejas: Nit. Let's clarify that this variable refers to the "space/gap/range" available for this step in the progress bar. I think it's ok if we call the variable the same in each step.
```suggestion
				const progressRange =
```

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -121,7 +126,12 @@ export function useSyncPush( {
 							  ),
 					showOpenLogs: true,
 				} );
+			} else if ( response.success && response.status === 'archive_import_started' ) {
+				status = pushStatesProgressInfo.applyingChanges;
+			} else if ( response.success && response.status === 'archive_import_finished' ) {
+				status = pushStatesProgressInfo.finishing;
```
sejas, line 132: 😚 👌

sejas (approved):
I tested it by pushing a Pressable site and the progress bar works great! No more doubts that the push is progressing and not stuck at all. Great work!

https://github.com/user-attachments/assets/4bb077a2-96aa-45fb-b3df-ae803c382ee2

sejas:
For comparison, here is how the progress bar was before:


https://github.com/user-attachments/assets/a80484c9-89d2-4bc7-b25c-4e84a7b69dc5

PR #1393, opened 2025-05-14
Silence console.log for tests

wojtekn:
## Related issues

- Fixes STU-493

## Proposed Changes

- I propose to mock `console.log` for tests to avoid that output visible in test execution log.

## Testing Instructions

1. Run tests:
```
npm run test
```
2. Confirm there is no output from `console.log`

## Pre-merge Checklist

### DIFF
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -5,6 +5,11 @@ import '@testing-library/jest-dom';
 import 'web-streams-polyfill/polyfill';
 import nock from 'nock';
 
+// Silence console.log for all tests
+beforeEach(() => {
+	console.log = jest.fn();
+});
+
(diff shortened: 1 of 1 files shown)


bcotrim (approved):
All tests passed, and I got a `console.warn` which is a nice confirmation that we will see see messages like this one and allow us to investigate it:
image

I would personally prefer to mock the console.log closer to the tests where it is needed but this approach also looks good. 👍

gcsecsey (approved):
This seems to be a better approach to me as well, thanks for iterating on it!

katinthehatsite (approved):
I think this is a much better approach that will allow us not to miss any important errors or warnings 👍

PR #1400, opened 2025-05-15
HTTPS: Convert domain name Unicode characters to ASCII characters

gavande1:
## Related issues

- Fixes STU-497

## Proposed Changes

Convert the domain name Unicode characters to ASCII characters for the SSL certificate name. 

## Testing Instructions

* Check out this branch.
* Run `npm start`.
* Change your locale to Japanese (日本語).
* Click **Add new site**, choose **Custom domain**, and enable **HTTPS**. 
Keep the default value for the domain. (E.g. `マイ-wordpress-サイト.wp.local`)
* Click **Add site**.
* Confirm that the site is created successfully and opens in the browser.
* Verify that there are no errors in the console.

## Pre-merge Checklist

### DIFF
--- a/src/lib/certificate-manager.ts
+++ b/src/lib/certificate-manager.ts
@@ -236,10 +236,9 @@ export async function generateSiteCertificate(
 	domain: string
 ): Promise< { cert: string; key: string } > {
 	try {
-		// Convert domain to punycode for certificate generation
 		const punycodeDomain = domainToASCII( domain );
-		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ punycodeDomain }.crt` );
-		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ punycodeDomain }.key` );
+		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.crt` );
+		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.key` );
 
 		// If the certificate already exists, no need to generate a new one
 		if ( fs.existsSync( siteCertPath ) && fs.existsSync( siteKeyPath ) ) {
@@ -319,9 +318,8 @@ export async function generateSiteCertificate(
  */
 export function deleteSiteCertificate( domain: string ): boolean {
 	try {
-		const punycodeDomain = domainToASCII( domain );
-		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ punycodeDomain }.crt` );
-		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ punycodeDomain }.key` );
+		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.crt` );
+		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.key` );
 		let deletedFiles = false;
 		if ( fs.existsSync( siteCertPath ) ) {
 			fs.unlinkSync( siteCertPath );
--- a/src/lib/certificate-manager.ts
+++ b/src/lib/certificate-manager.ts
@@ -3,6 +3,7 @@ import { execFile } from 'node:child_process';
 import crypto from 'node:crypto';
 import fs from 'node:fs';
 import path from 'node:path';
+import { domainToASCII } from 'node:url';
 import { promisify } from 'node:util';
 import * as Sentry from '@sentry/electron/main';
 import sudo from '@vscode/sudo-prompt';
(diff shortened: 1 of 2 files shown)


ivan-ottinger (approved):
Nice! Thanks for the fix! The changes look good and I can confirm they fix the issue with Japanese:

Markup on 2025-05-15 at 17:29:48

There seems to be one more issue when trying to start RTL sites with SSL though (e.g. Hebrew or Arabic):

Markup on 2025-05-15 at 17:57:42

I am approving the PR since we may want to look into this in a separate task.

src/lib/certificate-manager.ts
```diff
@@ -235,8 +236,10 @@ export async function generateSiteCertificate(
 	domain: string
 ): Promise< { cert: string; key: string } > {
 	try {
-		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.crt` );
-		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.key` );
+		// Convert domain to punycode for certificate generation
```
nightnei: Let's either jsut remove this comment or create a function with wrapping `domainToASCII` + adding JSDoc with explanation why do we need it.
Why:
1. now we have the comment only in 1 place, so if we keep it then it's necessary to copy-paster everywhere
2. The current comment is slightly unclear - how punycode influence in a good way on certificate generation.

nightnei:
The code works well 👍
Just left 1 comment here https://github.com/Automattic/studio/pull/1400#discussion_r2091607586
Screenshot 2025-05-15 at 17 50 24

src/lib/proxy-server.ts
```diff
@@ -170,7 +170,8 @@ export const startProxyServer = sequential( async (): Promise< boolean > => {
 			const defaultOptions: https.ServerOptions = {
 				SNICallback: async ( servername, cb ) => {
 					try {
-						const site = await getSiteByHost( servername );
+						const punycodeServerName = domainToASCII( servername );
+						const site = await getSiteByHost( punycodeServerName );
```
```suggestion
						const site = await getSiteByHost( servername );
```
`getSiteByHost` already converts the domain name to punycode internally.

src/lib/certificate-manager.ts
```diff
@@ -235,8 +236,10 @@ export async function generateSiteCertificate(
...
+		const punycodeDomain = domainToASCII( domain );
+		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ punycodeDomain }.crt` );
+		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ punycodeDomain }.key` );
```
```suggestion
		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.crt` );
		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.key` );
```
The path doesn't have to use the punycode-encoded domain name. I suggest sticking with the regular domain name for added readability.

src/lib/certificate-manager.ts
```diff
@@ -316,8 +319,9 @@ export async function generateSiteCertificate(
  */
 export function deleteSiteCertificate( domain: string ): boolean {
 	try {
-		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.crt` );
-		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.key` );
+		const punycodeDomain = domainToASCII( domain );
+		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ punycodeDomain }.crt` );
+		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ punycodeDomain }.key` );
```
```suggestion
		const siteCertPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.crt` );
		const siteKeyPath = path.join( CERT_DIRECTORY, 'domains', `${ domain }.key` );
```
Same thing here about the path not needing to be punycode-encoded.

fredrikekelund (commented):
Nice improvement 👍 I left a couple of suggestions that would be nice to address before landing this, though.

src/lib/certificate-manager.ts
wojtekn: Do we even need this comment? It seems clear what the line is doing.

nightnei: Agree, let's just remove it 👍

src/lib/proxy-server.ts
gavande1: Great catch! I removed it.

src/lib/certificate-manager.ts
gavande1: TIL. thanks for suggestion. I have removed it.

fredrikekelund (approved):
LGTM 👍

gcsecsey (approved):
LGTM and works as described 🚢

PR #1402, opened 2025-05-15
Set WP_ENVIRONMENT_TYPE to local by default

### LINKED ISSUE #165: WP_ENVIRONMENT_TYPE is unset / i.e. set to "production"
### Quick summary

If you run `wp_get_environment_type();` you'll see that because it's not set, the environment type is `production`. Just asking if it makes more sense to set `WP_ENVIRONMENT_TYPE to `local` by default for Studio sites since they are on a local environment.

(I would also consider setting WP_DEVELOPMENT_MODE to 'theme' or 'all', if the intended user is going to be building/updating custom themes, but I can understand if that's not the target audience or too opinionated.)

https://developer.wordpress.org/reference/functions/wp_get_environment_type/

If this ends up getting closed, users can always update their environment type by running ` wp config set WP_ENVIRONMENT_TYPE local` or updating the `wp-config.php` file.

### Steps to reproduce

1. Check wp-config or use wp-cli to check the environment type value

The environment type is `local`

### What actually happened

The environment type is `production`

### Impact

One

### Available workarounds?

Yes, easy to implement

### Platform

_No response_

### Logs or notes

_No response_

sejas:
## Related issues

- Fixes STU-17
- Fixes https://github.com/Automattic/studio/issues/165

## Proposed Changes

- Set the default environment type to `local`.  
- If the user defines the WP_ENVIRONMENT_TYPE constant in wp-config.php, respect that value and do not override it.

I also considered setting it in `defineWpConfigConsts`, but that would override the constants defined by the user, potentially causing issues when trying to replicate production environments.

https://github.com/Automattic/studio/blob/18a8e7cf111f555db6d38da33e91b744b6b47de0/vendor/wp-now/src/wp-now.ts#L418

More information  about [wp_get_environment_type](https://developer.wordpress.org/reference/functions/wp_get_environment_type/).

## Testing Instructions

- Run `npm start`
- Install and activate this plugin or similar:
```php
<?php
/**
* Plugin Name: Env Type
* Description: This plugin is used to display the current environment type.
* Author: Automattic
* Version: 1.0.0
*/

function env_type_notice() { ?>
    <div class="notice notice-info">
        <p>Current environment type: 
            <?php echo wp_get_environment_type(); ?>
        </p>
    </div>
<?php }
add_action( 'admin_notices', 'env_type_notice' );
```
- Observe the displayed value is `local`
- Define the constant at the top of wp-config.php `define( 'WP_ENVIRONMENT_TYPE', 'production' );`
- Observe the displayed value is `production`

https://github.com/user-attachments/assets/4fb1abe0-7c7d-4da4-a835-c4b32e0bc765

## Pre-merge Checklist

### DIFF
--- a/vendor/wp-now/src/wp-now.ts
+++ b/vendor/wp-now/src/wp-now.ts
@@ -603,6 +603,15 @@ async function mountInternalMuPlugins( php: PHP, options: WPNowOptions ) {
 		`
 	);
 
+	php.writeFile(
+		path.posix.join( PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER, '0-wp-environment-type.php' ),
+		`<?php
+		if ( ! defined( 'WP_ENVIRONMENT_TYPE' ) ) {
+			putenv( 'WP_ENVIRONMENT_TYPE=local' );
+		}
+		`
+	);
+
 	if ( ! options.isWpAutoUpdating ) {
 		php.writeFile(
 			path.posix.join( PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER, '0-disable-auto-updates.php' ),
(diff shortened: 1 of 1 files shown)


ivan-ottinger (approved):
The change looks good and works as expected. 👍🏼 

If the constant isn't defined, the default `local` is used:

Markup on 2025-05-16 at 09:51:44

Otherwise, the defined value (e.g. `define('WP_ENVIRONMENT_TYPE', 'staging');`) is applied:

Markup on 2025-05-16 at 09:53:55

PR #1411, opened 2025-05-19
Sanitize regex in Win editor path

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/studio/security/code-scanning/8

## Proposed Changes

I propose to sanitize the regex. It doesn't seem there is a known attack vectore, but I want to test CodeQL integration.

## Testing Instructions

1. Confirm editor can still be opened for Studio site on Windows

## Pre-merge Checklist


src/modules/user-settings/lib/win-editor-path.ts
```diff
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import nodePath from 'path';
+import _ from 'lodash';
```
nightnei: I am not a fan of lodash, sicne at my previous job, we had some pain around this lib. It's mostly personal opinion, since in general this lib is good.
So if we proceed with including it in our codebase, I would advocate for installing not the whole lib, but just specific functions, to avoid potential issues with tree shaking - https://www.npmjs.com/package/lodash.escaperegexp

nightnei (changes requested):

src/modules/user-settings/lib/win-editor-path.ts
nightnei: Update after discussion here p1747666104280689-slack-C06DRMD6VPZ:

At least, let's import the package directly from lodash, to improve tree-shaking:
`import escapeRegExp from 'lodash.escapeRegExp';`

wojtekn: In this case, requiring lodash doesn't change anything as it's already required by eight dependencies, including `electron-forge`. I adjusted the import, though.

epeicher (approved):
I know that this is still in Draft mode, but I have reviewed the changes and tested in Windows and changes LGTM, so providing my update here 😄

nightnei (approved):
LGTM 👍

PR #1412, opened 2025-05-19
Sync: Update notification to include Hostname

epeicher:
## Related issues

- Fixes STU-510

## Proposed Changes

- Add the Hostname of the connected site to the Push Notification
- Add the Hostname of the connected site to the Pull Notification

|Operation|Before | After|
|-------|-------|------|
|Pull|  CleanShot 2025-05-19 at 18 19 18@2x| CleanShot 2025-05-19 at 17 13 33@2x|
|Push|     CleanShot 2025-05-19 at 18 22 40@2x|CleanShot 2025-05-19 at 17 00 05@2x|

## Testing Instructions

- Apply this patch and run `npm start`
- Navigate to the Sync tab and connect a site or use one of your connected sites
- Click on Push
- Check the notification includes the URL of the connected site
* Click on Pull
* Check the notification includes the URL of the connected site

## Pre-merge Checklist

### DIFF
--- a/src/hooks/sync-sites/use-sync-push.ts
+++ b/src/hooks/sync-sites/use-sync-push.ts
@@ -24,7 +24,6 @@ export type SyncPushState = {
 	remoteSiteId: number;
 	status: PushStateProgressInfo;
 	selectedSite: SiteDetails;
-	isStaging: boolean;
 	remoteSiteUrl: string;
 };
 
@@ -181,7 +180,6 @@ export function useSyncPush( {
 				remoteSiteId,
 				status: pushStatesProgressInfo.creatingBackup,
 				selectedSite,
-				isStaging: connectedSite.isStaging,
 				remoteSiteUrl,
 			} );
 
--- a/src/hooks/sync-sites/use-sync-push.ts
+++ b/src/hooks/sync-sites/use-sync-push.ts
@@ -110,6 +110,7 @@ export function useSyncPush( {
 				getIpcApi().showNotification( {
 					title: syncPushState.selectedSite.name,
 					body: sprintf(
+						// translators: %s is the site url without the protocol.
 						__( '%s has been updated' ),
 						getHostnameFromUrl( syncPushState.remoteSiteUrl )
 					),
--- a/src/hooks/sync-sites/use-sync-push.ts
+++ b/src/hooks/sync-sites/use-sync-push.ts
@@ -16,6 +16,7 @@ import {
 	PushStateProgressInfo,
 } from 'src/hooks/use-sync-states-progress-info';
 import { getIpcApi } from 'src/lib/get-ipc-api';
+import { getHostnameFromUrl } from 'src/lib/url-utils';
 import type { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
 import type { ImportResponse } from 'src/hooks/use-sync-states-progress-info';
 
@@ -24,6 +25,7 @@ export type SyncPushState = {
 	status: PushStateProgressInfo;
 	selectedSite: SiteDetails;
 	isStaging: boolean;
+	remoteSiteUrl: string;
 };
 
 export type PushStates = Record< string, SyncPushState >;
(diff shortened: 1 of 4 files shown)


gcsecsey (approved):
Thanks for making these changes @epeicher, the notifications are much more descriptive now! 🙌 

These changes LGTM and work as described. 🚢

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -108,9 +109,10 @@ export function useSyncPush( {
 				onPushSuccess?.( remoteSiteId, syncPushState.selectedSite.id );
 				getIpcApi().showNotification( {
 					title: syncPushState.selectedSite.name,
-					body: syncPushState.isStaging
-						? __( 'Staging has been updated' )
-						: __( 'Production has been updated' ),
+					body: sprintf(
```
katinthehatsite, line 112: I am wondering if we should add a comment for translators here what % refers to so that they know it is a site URL. Mostly thinking that in some languages a "site" can be masculine or feminine so it might be useful for translators to know that it is a site and so that they can apply the right translation.

katinthehatsite (approved):
The changes look good and notifications are much clearer now. Thanks for making these improvements. I added one minor comment 👍 

Screenshot 2025-05-20 at 9 57 18 AM

src/hooks/sync-sites/use-sync-push.ts
epeicher: Good point, I'm adding a comment for translators in [8a5041d](https://github.com/Automattic/studio/pull/1412/commits/8a5041d7eab2dbdb2619b5ad2525ec0b6adfc004). Thanks for the suggestion!

ivan-ottinger (approved):
Nice! The changes look good and provide a great improvement to the notifications!

One thing we could consider is to shorten the URL with `…` if it is too long, e.g. in a case like this:

Markup on 2025-05-20 at 11:05:31

But that would probably work for the Pull operation only - since the Push notification has the address displayed in the beginning. 🤔

epeicher:
> One thing we could consider is to shorten the URL with … if it is too long. But that would probably work for the Pull operation only - since the Push notification has the address displayed in the beginning. 🤔

Mmmm... I personally don't like the `…` pattern, but I understand it is required for very long strings. I would say we can create a follow-up task for that if we receive further feedback about it being too wordy. What do you think @ivan-ottinger?

Update: I will merge this for now as it improves the notification a bit, we can always create follow-ups for further improvements 😄

ivan-ottinger:
> I personally don't like the … pattern, but I understand it is required for very long strings.

Yeah, similarly. I would like to see the full address in general. It just when it is too long it made the notification not that pretty. Regardless, at least full information is there.

> I would say we can create a follow-up task for that if we receive further feedback about it being too wordy. What do you think @ivan-ottinger?

That sounds good to me. 👍🏼

src/lib/url-utils.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Extracts the hostname from a URL string.
+ * @param url The URL string to extract the hostname from
+ * @returns The hostname (domain and port if specified), or an empty string if the URL is invalid
+ */
+export function getHostnameFromUrl( url: string ): string {
+	try {
+		const parsedUrl = new URL( url );
+		return parsedUrl.hostname;
+	} catch ( error ) {
+		return '';
```
sejas, line 11: @epeicher , should we return the whole `url` param if we cannot parse it?

sejas (commented):
The notification look great. Thanks for updating them.
I think it's ok to display the whole domain.

Screenshot 2025-05-20 at 11 39 38

src/lib/url-utils.ts
epeicher: Thinking again about it, I think it is a better UX than displaying an empty string. Let me create a quick PR for that.

epeicher: I have created https://github.com/Automattic/studio/pull/1420 for that

PR #1417, opened 2025-05-20
Remove `downloadWordPress` call in `startWPNow`

gcsecsey:
## Related issues

Fixes [STU-509 Studio: Check if we need to call downloadWordPress in the StartWPNow function](https://linear.app/a8c/issue/STU-509/studio-check-if-we-need-to-call-downloadwordpress-in-the-startwpnow)

## Proposed Changes

This PR removes the `downloadWordPress` call from the `startWPNow` function, as it seems unnecessary:
- WordPress core files are downloaded and cached in a shared location (under `getWordpressVersionsPath()`)
- The function was called after site creation, where it's already called from `createSiteWorkingDirectory`.
- Note that the `downloadWordPress` has a check not to re-download WordPress if the version already exists, so this extra call was always returning early.

## Testing Instructions

- Check your appdata folder (`~/Library/Application Support/Studio/server-files/wordpress-versions` on Mac), and note the already downloaded WordPress versions
- Create new sites using different WordPress versions than the cached ones
- For each new site:
  - Check that the WordPress version is downloaded to `wordpress-versions`
  - Check that the site is created successfully and opens in the browser
  - Check that there are no errors in the console

## Pre-merge Checklist

### DIFF
--- a/vendor/wp-now/src/wp-now.ts
+++ b/vendor/wp-now/src/wp-now.ts
@@ -33,7 +33,7 @@ import {
 	SQLITE_FILENAME,
 	SQLITE_PLUGIN_FOLDER,
 } from './constants';
-import { downloadWordPress, removeDownloadedMuPlugins } from './download';
+import { removeDownloadedMuPlugins } from './download';
 import getSqlitePath from './get-sqlite-path';
 import getWordpressVersionsPath from './get-wordpress-versions-path';
 import { output } from './output';
--- a/vendor/wp-now/src/wp-now.ts
+++ b/vendor/wp-now/src/wp-now.ts
@@ -90,7 +90,6 @@ export default async function startWPNow(
 		return { php, options };
 	}
 	output?.log( `wp: ${ options.wordPressVersion }` );
-	await downloadWordPress( options.wordPressVersion );
 
 	if ( options.reset ) {
 		fs.removeSync( options.wpContentPath );


ivan-ottinger (commented):
The proposed change looks good and I can confirm the sites are being created and started correctly in WP versions that I haven't had downloaded previously. 👌🏼🙂 

I can see that WP is getting downloaded already in `createSiteWorkingDirectory` (called in `createSite`), so another run in `startWPNow` shouldn't be needed.

---

Just noticed one thing: We can also remove the `downloadWordPress` import from the line `36`:

```
import { downloadWordPress, removeDownloadedMuPlugins } from './download';
```

gcsecsey:
> The proposed change looks good and I can confirm the sites are being created and started correctly in WP versions that I haven't had downloaded previously. 👌🏼🙂
> 
> I can see that WP is getting downloaded already in `createSiteWorkingDirectory` (called in `createSite`), so another run in `startWPNow` shouldn't be needed.
> 
> Just noticed one thing: We can also remove the `downloadWordPress` import from the line `36`:
> 
> ```
> import { downloadWordPress, removeDownloadedMuPlugins } from './download';
> ```

Thanks for reviewing this @ivan-ottinger and for highlighting the unnecessary import. I removed it in 47278b64.

ivan-ottinger (approved):
Approved! 🙂

nightnei:
We are testing only creating new sites, but I see that it's used also for executing PHP in cli https://github.com/Automattic/studio/blob/trunk/vendor/wp-now/src/execute-php.ts#L21, won't we break it?

nightnei:
Also, I am thinking - we have opportunity to change WP version. Don't we use this call of downloading WP to update files?

nightnei:
Update: tested, looks liek everything works well

nightnei (approved):

PR #1419, opened 2025-05-20
Fix hiding plugins spinner

nightnei:
## Related issues

- Fixes STU-267

## Proposed Changes
We have a bug - after pulling a site from dotcom to Studio - appears a spinner for updating all plugins, but it should be hidden. This is a dotcom bug, specifically in [page-optimize](https://wordpress.org/plugins/page-optimize/), but we have temporary workarounds [here](https://github.com/Automattic/wpcomsh/pull/975) and [here](https://github.com/Automattic/jetpack/blob/550550264caaaea98ca7527b7615cd14257d0aa8/projects/plugins/wpcomsh/plugin-hotfixes.php#L82). So I added this temporary fix to Studio to avoid confusion for our users with using Sync functionality. Later we can remove it, as soon as the initial issue is fixed. I pinged folks [here](https://github.com/Automattic/wpcomsh/pull/975#issuecomment-2893992482) about it.

## Testing Instructions

1. Create a site on dotcom (I haven't managed to reproduce it with Business plan, only Commerce)
2. Creat a site in Studio
3. Connect them
4. Pull site from dotcom to Studio
5. Wait till it's finished
6. Open WP admin of the local Studio's site
7. Open "Plugins" -> "Installed plugins" page
8. Assert that you don't see spinner prev to "Enable auto-updates" link

<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2025-05-20 at 14 53 46
<td>Screenshot 2025-05-20 at 14 53 54
</tr>
</table>

### DIFF
--- a/vendor/wp-now/src/wp-now.ts
+++ b/vendor/wp-now/src/wp-now.ts
@@ -639,8 +639,8 @@ async function mountInternalMuPlugins( php: PHP, options: WPNowOptions ) {
 		`<?php
 			// This is a temporary fix for a page-optimize bug that causes spinner icons to show all the time in the plugins list auto-update column
 
-			add_action( 'admin_enqueue_scripts', 'wpcomsh_patch_auto_update_spinner_style', 999 );
-			function wpcomsh_patch_auto_update_spinner_style() {
+			add_action( 'admin_enqueue_scripts', 'studio_patch_auto_update_spinner_style', 999 );
+			function studio_patch_auto_update_spinner_style() {
 				$current_screen = get_current_screen();
 				if ( isset( $current_screen->id ) && 'plugins' === $current_screen->id ) {
 					wp_add_inline_style(
--- a/vendor/wp-now/src/wp-now.ts
+++ b/vendor/wp-now/src/wp-now.ts
@@ -633,6 +633,24 @@ async function mountInternalMuPlugins( php: PHP, options: WPNowOptions ) {
 		}, 1, 3);
 		`
 	);
+
+	php.writeFile(
+		path.posix.join( PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER, '0-tmp-fix-hide-plugins-spinner.php' ),
+		`<?php
+			// This is a temporary fix for a page-optimize bug that causes spinner icons to show all the time in the plugins list auto-update column
+			// @see https://github.com/Automattic/wpcomsh/pull/975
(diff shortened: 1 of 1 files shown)


katinthehatsite:
I did not test this yet but wanted to mention that it was nice research to find all the temporary fixes in all these different places 👏

epeicher (approved):
This was indeed a great sleuthing, thanks for that! 🕵️ I have tested it, and it works as advertised, while I see the spinner in `trunk`, I cannot see it anymore in this branch. LGTM! :shipit: 

|Trunk | This branch|
|-------|------|
|CleanShot 2025-05-20 at 16 59 32@2x|  CleanShot 2025-05-20 at 16 58 32@2x|

vendor/wp-now/src/wp-now.ts
```diff
@@ -633,6 +633,24 @@ async function mountInternalMuPlugins( php: PHP, options: WPNowOptions ) {
...
 		`
 	);
+
+	php.writeFile(
+		path.posix.join( PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER, '0-tmp-fix-hide-plugins-spinner.php' ),
+		`<?php
+			// This is a temporary fix for a page-optimize bug that causes spinner icons to show all the time in the plugins list auto-update column
+			// @see https://github.com/Automattic/wpcomsh/pull/975
+			add_action( 'admin_enqueue_scripts', 'wpcomsh_patch_auto_update_spinner_style', 999 );
+			function wpcomsh_patch_auto_update_spinner_style() {
```
wojtekn: Would it make sense to rename this function to avoid potential conflicts for someone working on wpcomsh development in Studio?

Also, link points to private repository, so let's remove it.

nightnei: >Would it make sense to rename this function to avoid potential conflicts for someone working on wpcomsh development in Studio?

I intentionally decided to keep the sema name, to simplify for other developers to find this function in Studio, since it's the same fix.

>Also, link points to private repository, so let's remove it.

Ok

wojtekn: It might be safer to use our own prefix, e.g., `studio_patch_auto_update_spinner_style`.

nightnei: I see, updated

PR #1427, opened 2025-05-22
Studio: Follow symlinks when exporting site

epeicher:
## Related issues

- Solves STU-309

## Proposed Changes

- Add a `getSymlinks` function that returns recursively all the symlinks of a folder
- Avoid adding symlinks when adding files to compress using `archive.directory`
- Add the symlink real paths to the `archiver` zip file as an additional step

## Testing Instructions

This can be tested using Studio or the CLI

- Apply this branch and run `npm start` or `npm run cli:watch`
- Create a site and inside the `wp-content` folder, add some symlinks for either `plugins` or `themes` (although they can be on any other folder). For example, if you download the [following repo](https://github.com/juanma-wp/wp-data-layer-app-pages) to your `~/github` folder, you could add a symbolic link to it in your `plugins` folder using `ln -s ~/github/wp-data-layer-app-pages ~/Studio/my-symlinked-website/wp-content/plugins/wp-data-layer-app-pages` (replacing `my-symlinked-website` by your Studio site).
- On the Studio app, navigate to Previews and create a preview site or on a terminal, open the root folder of the studio repo and run `node dist/cli/main.js preview create --path ~/Studio/my-symlinked-website`
- Once created, navigate to the preview site, and open `/wp-admin/plugins.php` or navigate to Plugins -> Intalled Plugins
- Check you can see the symlinked plugin, if you followed the example, you should see `WP Data Layer App Pages`

**Optional Suggestion**: Apart from running one E2E test as explained above, for faster iterations to run multiple different tests, you can see the commit [797a39b](https://github.com/Automattic/studio/pull/1427/commits/797a39bfb65e0147ff01ef9d6fa9fd27ff8411df) that includes a `test.ts` command that can be used to only run the zip step. You can then check that the zip contains the expected symlinks resolved to their target paths. That test can be executed by `node dist/cli/main.js preview test --path ~/Studio/my-symlinked-website`

## Pre-merge Checklist


cli/commands/preview/test.ts
epeicher, line 1: This file will not be merged, added for convenience

cli/index.ts
epeicher, line 1: These changes will note be merged, added only for convenience

cli/lib/archive.ts
```diff
@@ -28,13 +32,32 @@ export async function createArchive(
...
 
+		for ( const symlink of symlinks ) {
+			const { symbolicPath, realPath } = symlink;
+			const archivePath = path.relative( siteFolder, symbolicPath );
+			if ( symlink.isDirectory ) {
+				archive.directory( realPath, archivePath, ( entry: EntryData ) => {
+					if ( shouldExcludeEntry( entry.name ) ) {
+						return false;
+					}
+					return entry;
```
epeicher: I am not excluding symlinks in the symlinked directories to allow for potential inner symlinks. 

Additionally, in this initial version, I am not following the symlinks inside the symlinked directories to reduce complexity. If that is a common use case, it could be tackled as a follow-up.

cli/lib/archive.ts
```diff
@@ -55,3 +78,33 @@ export async function cleanup( archivePath: string ): Promise< void > {
...
+				return await getSymlinks( filePath );
+			}
+			// Regular file
+			return [];
+		} )
+	);
+	return results.flat();
+}
+
+function shouldExcludeEntry( entryName: string ): boolean {
```
bcotrim: Have you considered calling `shouldExcludeEntry` in the `getSymlinks` function?
I think it could save us some execution time.

bcotrim (commented):
Code LGTM.
Added a minor comment, but it shouldn't be a blocker.

Testing instructions worked as described. Something I found during tests that might be an issue: I tried to use the export site functionality but noticed the symlinked plugin isn't included. Creating a preview site works as expected (the symlinked plugin is visible and functional).
I'm not sure if this is the expected behavior, if it is consider the PR approved.

cli/lib/archive.ts
fredrikekelund, line 1: This is definitely optional, but seeing as both `archive.directory` and `getSymlinks` iterate over the `wp-content` directory, we could collapse that into a single operation by using `readdir( 'wp-content', { recursive: true, withFileTypes: true } )`. 

The `recursive` option iterates over subdirectories _and_ resolves symlinks. The `withFileTypes` options makes it so a list of `fs.Dirent` objects is returned instead of a list of strings. We can then iterate over that list, and use the `archive.file` method to add files to the archive (filtering files by using the `fs.Dirent.isFile()` method).

fredrikekelund (approved):
Thanks for tackling this! This PR works as expected 👍

I think there's an opportunity to simplify the logic for iterating over the `wp-content` directory (per my inline comment), but that's only a suggestion. Feel free to merge the current changes if you prefer, @epeicher.

cli/lib/archive.ts
epeicher: Good point, I didn't think about that in the symlinks, but it's worth filtering out those, let me add it.

epeicher: Suggestion added here [5ac3d81](https://github.com/Automattic/studio/pull/1427/commits/5ac3d81e9f9fdfd363f8b64e4a476a43ac83f610)

cli/lib/archive.ts
epeicher: Thanks for the suggestion. That's an alternative approach that can be considered, and it is the [option used in the default-exporter.ts class](https://github.com/Automattic/studio/blob/5ac3d81e9f9fdfd363f8b64e4a476a43ac83f610/src/lib/import-export/export/exporters/default-exporter.ts#L218). 

In principle, it seems more verbose, and it seems more difficult to adapt if `archiver` implements the suggestion to follow symlinks. **But** it will make the `export` and `create preview` implementations very similar, so I can apply a similar approach to follow symlinks for both functionalities, so I am going to give it a try 👍

epeicher: @fredrikekelund, I have implemented the suggested approach as part of [5ac3d81](https://github.com/Automattic/studio/pull/1427/commits/5ac3d81e9f9fdfd363f8b64e4a476a43ac83f610). I think it could be more straightforward as it only needs one pass through the file tree, although it adds more conditionals and it uses [realPathSync and statSync](https://github.com/Automattic/studio/pull/1427/commits/354c0b317d6efa07e78c0053cda7014390d1bfd3#diff-a209913d07f0fe8b84dfa0623e37c62a3ace4bc4f40cd7d3849ae17e62abb9ddR43-R44). What do you think? Please let me know any suggestions.

If you are happy with the approach, we can progress this task, and I can tackle the changes required[ for the export functionality](https://github.com/Automattic/studio/blob/e5a266c39fe6c4cc5212b9e61a12e467d1dafe32/src/lib/import-export/export/exporters/default-exporter.ts#L218) as part of another task, as it requires slightly different changes.

epeicher:
@bcotrim, thanks for your review! 
>  Something I found during tests that might be an issue: I tried to use the export site functionality but noticed the symlinked plugin isn't included.

Well spotted! The export functionality is slightly different than creating the preview archive, it uses the default-exporter.ts. So I think it will be better to tackle that as part of a follow-up task.

Also, please note that I've changed the approach for this PR as [suggested here](https://github.com/Automattic/studio/pull/1427#discussion_r2108462293), so the code is now slightly different. Happy to have another round of review.

bcotrim:
@epeicher while re-reviewing the changes, I tested a new scenario with a broken symlink in the plugins folder.
I created the synlink as usual, confirmed it was working, and then moved the target file. 
The preview command now fails
image

Is this the expected behavior? 
To me it seems like ideally, we would want to proceed with the preview site creation since their WP site isn't broken and give our user a warning.
What do you think?

epeicher:
> I tested a new scenario with a broken symlink in the plugins folder.
I created the synlink as usual, confirmed it was working, and then moved the target file.
The preview command now fails

> Is this the expected behavior?
To me it seems like ideally, we would want to proceed with the preview site creation since their WP site isn't broken and give our user a warning.
What do you think?

@bcotrim that is a very good point, to cover the errors and edge cases that the users will face for sure. The current implementation assumes the links point to correct targets, but it can be updated, so when the target does not exist, symlinks can be ignored. I think this is better than the production behavior, which adds the symlinks (which could be broken), what do you think?

cli/lib/archive.ts
```diff
@@ -10,6 +11,12 @@ export async function createArchive(
 	siteFolder: string,
 	archivePath: string
 ): Promise< archiver.Archiver > {
+	const wpContentFolder = path.join( siteFolder, 'wp-content' );
+	const directoryContents = await fsPromises.readdir( wpContentFolder, {
+		recursive: true,
+		withFileTypes: true,
```
fredrikekelund:
```suggestion
```
Sorry for getting this wrong in my previous comment, but I see now that `readdir` only resolves symlinks if `recursive: true` and `withFileTypes: false`. 

I suggest simplifying the logic by removing `withFileTypes`.

bcotrim:
> @bcotrim that is a very good point, to cover the errors and edge cases that the users will face for sure. The current implementation assumes the links point to correct targets, but it can be updated, so when the target does not exist, symlinks can be ignored. I think this is better than the production behavior, which adds the symlinks (which could be broken), what do you think?

That sounds good to me!

cli/lib/archive.ts
```diff
@@ -24,17 +31,32 @@ export async function createArchive(
...
+						}
+						return entry;
+					} );
+				} else {
+					archive.file( realPath, { name: archivePath } );
+				}
+			} else if ( dirEnt.isFile() ) {
+				archive.file( filePath, { name: archivePath } );
+			}
+		}
```
fredrikekelund:
```ts
		for ( const entryPath of directoryContents ) {
			if ( entryPath.includes( '.git' ) || entryPath.includes( 'node_modules' ) ) {
				continue;
			}

			const absolutePath = path.join( wpContentFolder, entryPath );
			const stat = fs.statSync( absolutePath );

			if ( stat.isFile() ) {
				const archivePath = path.relative( siteFolder, absolutePath );
				archive.file( absolutePath, { name: archivePath } );
			}
		}
```
Without the `withFileTypes` option to `readdir`, the `recursive` option is much more useful, because it also resolves symlinks. 

Here's a suggestion for how we could simplify this loop more along the lines of what I had in mind to begin with.

fredrikekelund (commented):
Thanks for considering my suggestion, @epeicher. I realized that I got confused about the behavior of `readdir` when passing just `recursive` compared to when `withFileTypes` is also passed. I left a suggestion for how we could simplify this even further by using only the `recursive` option.

fredrikekelund:
I believe my latest suggestion should also fix the issue @bcotrim raised about broken symlinks.

cli/lib/archive.ts
epeicher: Oh, that's a very key point. I tested that to check if `readdir` followed symlinks, but didn't think about changing `withFileTypes` to `false`. Let me do a quick iteration as I think it will simplify the code a lot.

cli/lib/archive.ts
epeicher: Indeed, let me apply that or a similar one and test.

epeicher: Thanks for the suggestion @fredrikekelund! I have applied a similar approach but I've needed to check for symlinks that point to files as they are not followed by `readdir`, you can see my changes in this commit [a1ea258](https://github.com/Automattic/studio/pull/1427/commits/a1ea258c8fff0b73618c6df64a6b626a9f3cdc1e). I have added a `try..catch` to handle errors if the path does not exist. ~~That can be always extracted to an external function that returns the `filePath` if it exists.~~ Extracted to an external function  in [683e81d](https://github.com/Automattic/studio/pull/1427/commits/683e81d3d9d90fd172f5951c3fbf370999d6511a)

epeicher:
> I believe my latest suggestion should also fix the issue @bcotrim raised about broken symlinks.

The suggestion was great, but I have needed to follow symlinks using `realpathSync` because the target files are not returned by `readdir` for symlinked _files_. I am also checking errors for broken symlinks and handling them gracefully [in here](https://github.com/Automattic/studio/blob/d603b339081a8a5a25b7ff0fc76ba1de37ba02c9/cli/lib/archive.ts#L75).

@bcotrim , @fredrikekelund , this is ready for another review 👀

fredrikekelund (approved):
> I have needed to follow symlinks using realpathSync because the target files are not returned by readdir for symlinked files

Good catch 👍

This is looking great and works as expected!

bcotrim (approved):
LGTM 👍

epeicher:
Closing this in favour of https://github.com/Automattic/studio/pull/1447

For reference, this other PR follows symlinks for Export and Push functionality https://github.com/Automattic/studio/pull/1448

PR #1428, opened 2025-05-23
Terminal Apps: Use URI scheme to open Warp terminal on Windows

gavande1:
## Related issues

- Fixes STU-519

## Proposed Changes

I propose using the URI Scheme to open Warp Terminal to fix the issue of not opening on Windows. The documentation for URI schemes can be found at https://docs.warp.dev/features/uri-scheme. 

## Testing Instructions

- Check out this branch
- Install Warp if it is not installed. 
- Select Warp from Preferences as your terminal. 
- Click on the Warp option on the site overview.
- Confirm that the Warp terminal opens correctly with the correct site directly. 

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -1051,7 +1051,8 @@ END` );
 			: {};
 
 		if ( preferredTerminal === ( 'warp' as SupportedTerminal ) ) {
-			return promiseExec( `start "" "warp" "--cwd=${ targetPath }"`, {
+			const encodedPath = encodeURIComponent( targetPath );
+			return promiseExec( `start "" "warp://action/new_tab?path=${ encodedPath }"`, {
 				env: { ...process.env, ...env },
 			} );
 		}


src/ipc-handlers.ts
```diff
@@ -1051,7 +1051,8 @@ END` );
 			: {};
 
 		if ( preferredTerminal === ( 'warp' as SupportedTerminal ) ) {
-			return promiseExec( `start "" "warp" "--cwd=${ targetPath }"`, {
+			const encodedPath = encodeURIComponent( targetPath );
+			return promiseExec( `start "" "warp://action/new_tab?path=${ encodedPath }"`, {
```
gavande1, line 1055: I chose to open a new tab instead of a new window, as the user might already have window open.

epeicher (approved):
Thanks for fixing this Rahul! I have tested it, and it works as advertised. I can now open Warp on Windows 🤓 Also I like the _touch_ of opening in a new tab and is working well 👍. LGTM! :shipit: 

|Opening Warp | Warp already opened|
|-------|------|
|CleanShot 2025-05-23 at 10 03 38@2x| CleanShot 2025-05-23 at 10 07 32@2x|

PR #1430, opened 2025-05-23
ESLint rule for user data lockfile

bcotrim:
## Related issues

- https://github.com/Automattic/studio/pull/1319

## Proposed Changes

- Add ESLint rule to enforce `lock` and `unlock` when making changes to the app data file

Code errors examples:
| Without changes | With changes to data | With changes, lock but no unlock | No error |
| image | image | image | image | 

## Testing Instructions

- Code review and checking the ESLint behavior mentioned above

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -237,18 +237,13 @@ export async function createSite(
 
 	const parentWindow = BrowserWindow.fromWebContents( event.sender );
 	sendIpcEventToRendererWithWindow( parentWindow, 'theme-details-updating', { id: details.id } );
-	try {
-		await lockAppdata();
-		userData = await loadUserData();
+	userData = await loadUserData();
 
-		userData.sites.push( server.details );
-		sortSites( userData.sites );
+	userData.sites.push( server.details );
+	sortSites( userData.sites );
 
-		await saveUserData( userData );
-		return server.details;
-	} finally {
-		await unlockAppdata();
-	}
+	await saveUserData( userData );
+	return server.details;
 }
 
 export async function updateSite(
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -237,7 +237,6 @@ export async function createSite(
 
 	const parentWindow = BrowserWindow.fromWebContents( event.sender );
 	sendIpcEventToRendererWithWindow( parentWindow, 'theme-details-updating', { id: details.id } );
-
 	try {
 		await lockAppdata();
 		userData = await loadUserData();
@@ -246,11 +245,10 @@ export async function createSite(
 		sortSites( userData.sites );
 
 		await saveUserData( userData );
+		return server.details;
 	} finally {
 		await unlockAppdata();
 	}
-
-	return server.details;
 }
 
 export async function updateSite(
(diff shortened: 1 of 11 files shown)


src/storage/tests/user-data.test.ts
```diff
@@ -7,7 +7,7 @@ import { normalize } from 'path';
 import * as atomically from 'atomically';
 import { getUserDataFilePath } from 'src/storage/paths';
 import { UserData } from 'src/storage/storage-types';
-import { loadUserData, saveUserData } from 'src/storage/user-data';
```
nightnei, line 7: @bcotrim  I am reviewing, but initially wanted to say that here is something weird:
Screenshot 2025-05-23 at 15 36 06
Screenshot 2025-05-23 at 15 36 16

nightnei (changes requested):
The general idea is awesome 👍
But I can't  understand the usecase - in Frdrik's PR we are moving to using `withUserDataWrite` as the wrapper for `saveUserData`, so we won't use it directly anymore. And if somebody need to update appdata, they will definitely use only `withUserDataWrite` as the single option which includes lock/unlock out of the box, so we don't need eslint for this approach.
Could you please elaborate on the idea, maybe I am missing something, but can't understand how this linter will be used with Fredrik's changes.

bcotrim:
Thanks for the review @nightnei 

In Fredrik's PR you had some concerns about the readability and usage of the HOF, the ESLint rule allows us to not use that while ensuring we "force" developers to implement the lock/unlock in their code when they change the appdata.

nightnei:
@bcotrim oh, I missed that it's a proposition of an alternative approach to HOF. The idea is really cool 👍

src/ipc-handlers.ts
```diff
@@ -235,18 +235,13 @@ export async function createSite(
...
+	userData.sites.push( server.details );
+	sortSites( userData.sites );
 
-		await saveUserData( userData );
-		return server.details;
-	} finally {
-		await unlockAppdata();
-	}
+	await saveUserData( userData );
+	return server.details;
```
epeicher: Just curious as I'm not very familiar with the changes, why is not required here to lock the app data file?

epeicher (commented):
Thanks Bernardo for improving the maintainability of this! I have tested it, and it works as advertised, I can see the lint rule applied. However, when I run the command, it returns the line `1:1` and I think that is why the whole file gets the warning. Please see the video below:

CleanShot 2025-06-23 at 13 28 15@2x

https://github.com/user-attachments/assets/a87a3184-1544-42e5-aa85-3a33a7e95235


It could be something with my environment, for context, I have deleted `node_modules` and run `npm install` again, and I have closed and opened Cursor

src/ipc-handlers.ts
bcotrim: thanks for catching this one, we should not remove this.

bcotrim:
@nightnei @epeicher thanks for reviewing!
There was an issue with the rule scope. Can you take another look please?

epeicher (approved):
Thanks Bernardo for addressing the feeback! I have tested it, and it works as expected. LGTM! :shipit: 

|Not getting the lock  | Removing the try-finally |Not calling the unlock|
|-------|------|------|
|CleanShot 2025-06-23 at 16 24 44@2x| CleanShot 2025-06-23 at 16 29 59@2x| CleanShot 2025-06-23 at 16 26 11@2x|

nightnei (approved):
LGTM 👍

PR #910, opened 2025-02-10
Fix some RTL styles

bcotrim:
## Related issues

- Fixes https://github.com/Automattic/dotcom-forge/issues/10374

## Proposed Changes

<details>
<summary> Modal title alignment</summary>

image

</details>

<details>
<summary>Sync pull/push buttons</summary>

image

</details>

<details>
<summary>Sync progress bar</summary>

image

</details>

<details>
<summary>Sync clear button</summary>

image

</details>

## Testing Instructions

1. Open studio and change the language to an RTL language
2. Check a modal title (user settings, create site, etc...)
3. Navigate to Sync tab
4. Check the icons alignment
5. Do a push/pull and check the progress bar layout
6. After the push/pull check the clear button layout

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1435, opened 2025-05-26
Use the same style for the Preview and Sync tabs

sejas:
## Related issues

- Fixes p1748248185249619-slack-C06DRMD6VPZ

## Proposed Changes

- Let's make the tabs Sync and Preview tabs consistent with styling and fix the two lines title when the scrollbar is visible
- Removes `rtl:order-first`, which was added in https://github.com/Automattic/studio/pull/910, although there is no mention of making that change intentionally.

## Testing Instructions

- Run `npm start`
- Set your Mac Preferences > Appearance > Show Scroll bar  > Always
- Navigate to the sync tab when not authenticated
- Observe the screen looks good.

| Before | After |
|--------|--------|
| Screenshot 2025-05-26 at 10 30 57 | Screenshot 2025-05-26 at 10 30 40 |

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-sync.tsx
+++ b/src/components/content-tab-sync.tsx
@@ -17,8 +17,8 @@ import type { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
 function SiteSyncDescription( { children }: PropsWithChildren ) {
 	const { __ } = useI18n();
 	return (
-		<div className="flex justify-between max-w-3xl gap-4">
-			<div className="flex flex-col p-8 pr-7">
+		<div className="p-8 flex justify-between max-w-3xl gap-4">
+			<div className="flex flex-col">
 				<div className="flex items-center mb-1">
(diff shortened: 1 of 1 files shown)


ivan-ottinger (approved):
Thank you for the fix! The proposed changes look good and resolve the issue:

Markup on 2025-05-26 at 11:42:20

When the scrollbars are set to display `always`, the Sync tab is formatted correctly.

I did not observe any regressions and the tab content is rendering correctly when when resized, with or without the tab toggled.

epeicher (approved):
Thanks for the fix! I have tested it, and it works as advertised. I can see the bottom message in one line, and it works fine when resizing. LGTM! :shipit: 


https://github.com/user-attachments/assets/8cf472fe-4d95-4914-ad22-00352814a132

|Before | After|
|-------|------|
| CleanShot 2025-05-26 at 11 52 05@2x| CleanShot 2025-05-26 at 11 47 37@2x|

PR #1443, opened 2025-05-27
refactor: isolate locking appdata

nightnei:
## Related issues

- Proposition PR for https://github.com/Automattic/studio/pull/1319

## Proposed Changes

- Isolate locking/unlocking of appdata, avoid duplications and keep code reliable.


bcotrim:
I'm confused. Isn't this just going back to the HOF approach?

nightnei:
> I'm confused. Isn't this just going back to the HOF approach?

It's not HOF, since we don't have a wrapper function. But thinking more about it - probably it will again be cumbersome changes... maybe we should just start with at least something, for example with what we already have in [this PR](https://github.com/Automattic/studio/pull/1319) and then we will see.

PR #1447, opened 2025-05-28
Follow symlinks in Preview sites when archiving directories

epeicher:
## Related issues

- Fixes STU-309

## Proposed Changes

- Create a patch for the `archiver` third party package so it follows the symlinks when archiving directories

### This can be tested using Studio or the CLI

- Apply this branch and run `npm install`
- Start Studio with `npm start` or if using the CLI run `npm run cli:watch`
- Create a site and inside the `wp-content` folder, add some symlinks for either `plugins` or `themes` (although they can be on any other folder). For example, if you download the [following repo](https://github.com/juanma-wp/wp-data-layer-app-pages) to your `~/github` folder, you could add a symbolic link to it in your `plugins` folder using `ln -s ~/github/wp-data-layer-app-pages ~/Studio/my-symlinked-website/wp-content/plugins/wp-data-layer-app-pages` (replacing `my-symlinked-website` by your Studio site).
- Create a symlink to a file, for example, create a symlink that points to a dummy plugin file outside of the site folder, .e.g. save the following file and add a symlink to it in the plugins folder:
```
<?php
/**
 * Plugin Name: My Test Plugin
 * Plugin URI: https://example.com/
 * Description: A simple test plugin that doesn't do anything.
 * Version: 1.0
 * Author: Your Name
 * Author URI: https://example.com/
 * License: GPL2
 */

// Silence is golden.
```
- On the Studio app, navigate to Previews and create a preview site or on a terminal, open the root folder of the studio repo and run `node dist/cli/main.js preview create --path ~/Studio/my-symlinked-website`
- Once created, navigate to the preview site, and open `/wp-admin/plugins.php` or navigate to Plugins -> Intalled Plugins
- Check you can see the symlinked plugins, if you followed the example, you should see `WP Data Layer App Pages` and you should also see `My Test Plugin`, for example:
CleanShot 2025-05-28 at 18 00 38@2x

#### Test broken symlinks
- Following the steps above, delete or move the target folders and file of the symlinks so the symlinks will point to a non-existing directory and file.
- Create a preview site and make sure that there are no unexpected errors.
- Open the preview site, and confirm that you cannot see the Plugins, but you don't see any errors.

## Pre-merge Checklist

### DIFF
--- a/patches/archiver+6.0.2.patch
+++ b/patches/archiver+6.0.2.patch
@@ -1,0 +1,14 @@
+diff --git a/node_modules/archiver/lib/core.js b/node_modules/archiver/lib/core.js
+index 7c0a74d..5ac4e3a 100644
+--- a/node_modules/archiver/lib/core.js
++++ b/node_modules/archiver/lib/core.js
+@@ -631,7 +631,8 @@ Archiver.prototype.directory = function(dirpath, destpath, data) {
+ 
+   var globOptions = {
+     stat: true,
+-    dot: true
++    dot: true,
++    follow: true,
+   };
+ 
+   function onGlobEnd() {
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 Unreleased
 ==========
 * Miscellaneous security improvements #1411
+* Follow symlinks when creating preview sites #1447
 
 1.5.2
 =====
--- a/cli/lib/archive.ts
+++ b/cli/lib/archive.ts
@@ -14,6 +14,7 @@ export async function createArchive(
 		const output = fs.createWriteStream( archivePath );
 		const archive = archiver( 'zip', {
 			zlib: { level: ZIP_COMPRESSION_LEVEL },
+			followSymlinks: true,
 		} );
 
 		output.on( 'close', () => {
--- a/cli/lib/tests/archive.test.ts
+++ b/cli/lib/tests/archive.test.ts
@@ -48,7 +48,10 @@ describe( 'Archive Module', () => {
 			const result = await createArchive( mockSiteFolder, mockArchivePath );
 
 			expect( fs.createWriteStream ).toHaveBeenCalledWith( mockArchivePath );
-			expect( archiver ).toHaveBeenCalledWith( 'zip', { zlib: { level: 9 } } );
+			expect( archiver ).toHaveBeenCalledWith( 'zip', {
+				followSymlinks: true,
+				zlib: { level: 9 },
+			} );
 			expect( mockArchiver.pipe ).toHaveBeenCalledWith( mockWriteStream );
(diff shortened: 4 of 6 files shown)


patches/archiver+6.0.2.patch
```diff
@@ -0,0 +1,14 @@
...
+index 7c0a74d..5ac4e3a 100644
+--- a/node_modules/archiver/lib/core.js
++++ b/node_modules/archiver/lib/core.js
+@@ -631,7 +631,8 @@ Archiver.prototype.directory = function(dirpath, destpath, data) {
+ 
+   var globOptions = {
+     stat: true,
+-    dot: true
++    dot: true,
++    follow: true,
```
sejas: We could consider using the same code from your node-archiver PR: https://github.com/archiverjs/node-archiver/pull/810

sejas (approved):
Works as expected, and now Preview sites follow symlinks when generating the zip file.

Let's rename the patch file from `patches/archiver+6.0.2.patch` to `patches/archiver+6.0.1.patch`. I got a warning when installing the dependencies with `npm install`.

```
patch-package 8.0.0
Applying patches...
@automattic/generate-password@0.1.0 ✔
@wordpress/components@29.9.0 ✔
@wordpress/compose@6.27.0 ✔
archiver@6.0.2 ✔

Warning: patch-package detected a patch file version mismatch

  Don't worry! This is probably fine. The patch was still applied
  successfully. Here's the deets:

  Patch file created for

    archiver@6.0.2

  applied to

    archiver@6.0.1
  
  At path
  
    node_modules/archiver

  This warning is just to give you a heads-up. There is a small chance of
  breakage even though the patch was applied successfully. Make sure the package
  still behaves like you expect (you wrote tests, right?) and then run

    patch-package archiver

  to update the version in the patch file name and make this warning go away.

---
patch-package finished with 1 warning(s).
```

patches/archiver+6.0.2.patch
epeicher: That's a good point. I decided to use this simple approach because it will be discarded if there is a potential upgrade. But thinking about it, and related to this other PR https://github.com/Automattic/studio/pull/1448, it is more explicit to set the option from the calling code, i.e. `archiver('zip', {followSymlinks: true})`, and the changes in the calling code will reflect better why the symlinks are followed, so I will apply your suggested changes instead 👍

epeicher: Done as part of latest commits [f10e4c8](https://github.com/Automattic/studio/pull/1447/commits/f10e4c852065b2d7e6147e5abb167ee8689f8000), [be3b8a6](https://github.com/Automattic/studio/pull/1447/commits/be3b8a63327eb1a2c4dfe7bfb77a646fd7393728) and [51437f9](https://github.com/Automattic/studio/pull/1447/commits/51437f9fe9e2a21ae31405e4c8b5afcb4d080b37)

PR #1448, opened 2025-05-28
Follow symlink directories for Jetpack backups

epeicher:
## Related issues

- Fixes STU-392
- Related PR https://github.com/Automattic/studio/pull/1447

## Proposed Changes

- Use `archive.directory` instead of `archive.file` for creating the Jetpack backup file so the symlinks can be followed. This works because of [the following patch in archiver](https://github.com/Automattic/studio/blob/a6cc70ce9a756b2bcfadcd6640dd5bdd26cd5352/patches/archiver%2B6.0.2.patch).

### Export
- Create a site that contains symlinks to external folders for example for plugins.
- Ensure you create both symlinks pointing to directories and symlinks pointing to external files.
- Navigate to the Import/Export tab and do an export of the site
- Open the site in Finder, unzip it, and check that the symlinked folders and files contain the actual target folders and dirs and not the symlinks.
- Repeat the export but ensure you add some symlinks that are pointing to non-existent folders
- Check there are not unexpected errors.
### Push
* Navigate to the Sync tab
* Connect a site to a WordPress.com or Pressable site
* Run a Push
* Check the pushed site contains the symlinked plugins

## Pre-merge Checklist


src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -162,11 +162,16 @@ export class DefaultExporter extends EventEmitter implements Exporter {
...
-				this.archive.file( file, { name: relativePath } );
-				this.emit( ExportEvents.WP_CONTENT_EXPORT_PROGRESS, { file: relativePath } );
-			}
+			const folderName = category === 'muPlugins' ? 'mu-plugins' : category;
+			const absolutePath = path.join( this.options.site.path, 'wp-content', folderName );
+			const archivePath = path.relative( this.options.site.path, absolutePath );
+			this.archive.directory( absolutePath, archivePath, ( entry ) => {
+				if ( entry.name.includes( '.git' ) || entry.name.includes( 'node_modules' ) ) {
+					return false;
+				}
```
sejas: Let's also exclude the [pathsToExclude](https://github.com/Automattic/studio/blob/db3060cb639a24e1d0a3fa88ed7d60c8831efc2e/src/lib/import-export/export/exporters/default-exporter.ts#L31-L41).

Currently I see a regression where the `sqlite-database-integration` mu-plugin is being exported.

sejas (commented):
This almost works as expected. I confirm the export includes/follow symlinks, but it's not excluding `pathsToExclude` and I see the `sqlite-database-integration` folder in the final zip.

Screenshot 2025-05-29 at 14 27 56

I think we should remove the [getSiteFiles](https://github.com/Automattic/studio/blob/db3060cb639a24e1d0a3fa88ed7d60c8831efc2e/src/lib/import-export/export/exporters/default-exporter.ts#L226) but we can tackle that in a separate issue. I created STU-536.

src/lib/import-export/export/exporters/default-exporter.ts
sejas: Maybe if mu-plugins is empty we avoid adding it too 🤔

epeicher: Good catch! I have excluded them as part of [51c1983](https://github.com/Automattic/studio/pull/1448/commits/51c198366c39c9200b95fe659f42c4adb0d05af7). 

> Maybe if mu-plugins is empty we avoid adding it too 🤔

A nice side effect is that empty directories are automatically excluded, for example, now that `pathsToExclude` filtering is applied, I cannot see either `mu-plugins` or `fonts` because they are empty in my site 
CleanShot 2025-05-29 at 17 55 18@2x

epeicher:
> I think we should remove the [getSiteFiles](https://github.com/Automattic/studio/blob/db3060cb639a24e1d0a3fa88ed7d60c8831efc2e/src/lib/import-export/export/exporters/default-exporter.ts#L226) but we can tackle that in a separate issue. I created STU-536

I also thought about that and decided to suggest it as a follow-up, but you beat me to it 😄. Thank you very much for creating the task

sejas (approved):
Thanks for the update. It works great! Now the pathsToExclude paths are excluded, and the symlinks are resolved and included in the zip 👍 .

**Before**

https://github.com/user-attachments/assets/c09d50f3-a75f-4593-a8e0-e21be9345910



**After**


https://github.com/user-attachments/assets/20250a2c-b01b-4567-91d3-7713ef2627e7

PR #1452, opened 2025-05-29
Update translations for 1.5.3

gavande1:
## Related issues

- pfHvTO-xT-p2

## Proposed Changes

- Add translations for 1.5.3.

## Testing Instructions

- Test if Studio works when different language is selected

## Pre-merge Checklist


ivan-ottinger (commented):
Looks like we might be missing a couple of language files (at least I am missing Ukrainian, Vietnamese, Hebrew, Korean, Italian and Polish and Russian).

gavande1:
@ivan-ottinger I think this is because those language files are not updated on GlotPress. And, I am doing this a bit early. The strings need to be added in the release stage and not in the beta release stage. Right?

ivan-ottinger:
> @ivan-ottinger I think this is because those language files are not updated on GlotPress. I think I am doing this a bit early. The strings need to be added in the release stage and not in the beta release stage. Right?

During the initial phase of the release rotation I only export strings from Studio and import them to GlotPress. In other words, I follow these steps [from the docs](https://github.com/Automattic/studio/blob/trunk/docs/localization.md#extract-and-import):
Markup on 2025-05-29 at 11:56:42

I repeat the process if new strings are added to Studio throughout the week.

Then, when it is close to the actual release, I follow the remaining steps [on the doc](https://github.com/Automattic/studio/blob/trunk/docs/localization.md#export-and-add) to get the translated strings out of GlotPress and to the Studio repo.

---

But I don't think there's anything wrong with getting fresh translations into Studio - especially if we want to test / see the UI with the new translations in. 🙂

ivan-ottinger (approved):
> @ivan-ottinger I think this is because those language files are not updated on GlotPress.

Good point! I missed that.

The app is loading correctly in different locales. 👍🏼

gavande1:
> But I don't think there's anything wrong with getting fresh translations into Studio - especially if we want to test / see the UI with the new translations in. 🙂

Thanks for the detailed explanation. It makes sense to me. Also, TIL.

wojtekn (approved):

gcsecsey (approved):
Works well for me 🚢

PR #1454, opened 2025-05-29
Fix blueprints support in wp-now code

wojtekn:
## Related issues

- N/A

## Proposed Changes

I propose to fix blueprints support in the `wp-now` code. It doesn't cause any issue in the current Studio, but it prevents as from exploring blueprint in Studio.

## Testing Instructions

1. Confirm that the site can still be created and started without any issues
2. Apply a patch that hardcodes the blueprint (feel free to use different set of plugins):

```
diff --git a/src/site-server.ts b/src/site-server.ts
index 14869c1e..04177f08 100644
--- a/src/site-server.ts
+++ b/src/site-server.ts
@@ -177,6 +177,38 @@ export class SiteServer {
 			isWpAutoUpdating: this.details.isWpAutoUpdating,
 		} );
 
+		options.blueprintObject = JSON.parse(
+			`{
+			"steps": [
+				{
+					"step": "installPlugin",
+					"pluginData": {
+						"resource": "wordpress.org/plugins",
+						"slug": "akismet"
+					}
+				}, {
+					"step": "installPlugin",
+					"pluginData": {
+						"resource": "wordpress.org/plugins",
+						"slug": "crowdsignal-forms"
+					}
+				}, {
+					"step": "installPlugin",
+					"pluginData": {
+						"resource": "wordpress.org/plugins",
+						"slug": "polldaddy"
+					}
+				}, {
+					"step": "installPlugin",
+					"pluginData": {
+						"resource": "wordpress.org/plugins",
+						"slug": "gutenberg"
+					}
+				}
+			]
+		}`
+		);
+
 		options.absoluteUrl = getAbsoluteUrl( this.details );
 		options.siteLanguage = await getPreferredSiteLanguage( options.wordPressVersion );
 
```

3. Create a new Studio site
4. Confirm it was created and started successfully
5. Open WP Admin and confirm plugins were installed

## Pre-merge Checklist

### DIFF
--- a/vendor/wp-now/src/wp-now.ts
+++ b/vendor/wp-now/src/wp-now.ts
@@ -100,9 +100,11 @@ export default async function startWPNow(
 
 	await prepareWordPress( php, options );
 
+	await installationSteps( php, options );
+
 	if ( options.blueprintObject ) {
 		output?.log( `blueprint steps: ${ options.blueprintObject.steps.length }` );
-		const compiled = compileBlueprint( options.blueprintObject, {
+		const compiled = await compileBlueprint( options.blueprintObject, {
 			onStepCompleted: ( result, step: StepDefinition ) => {
 				output?.log( `Blueprint step completed: ${ step.step }` );
 			},
@@ -110,7 +112,6 @@ export default async function startWPNow(
 		await runBlueprintSteps( compiled, php );
 	}
 
-	await installationSteps( php, options );
 	await login( php, options );
 
 	if ( isFirstTimeProject && [ WPNowMode.PLUGIN, WPNowMode.THEME ].includes( options.mode ) ) {


vendor/wp-now/src/wp-now.ts
```diff
@@ -100,17 +100,18 @@ export default async function startWPNow(
 
 	await prepareWordPress( php, options );
 
+	await installationSteps( php, options );
+
 	if ( options.blueprintObject ) {
 		output?.log( `blueprint steps: ${ options.blueprintObject.steps.length }` );
-		const compiled = compileBlueprint( options.blueprintObject, {
+		const compiled = await compileBlueprint( options.blueprintObject, {
```
wojtekn, line 107: The `compileBlueprint()` function returns a promise.

vendor/wp-now/src/wp-now.ts
```diff
@@ -100,17 +100,18 @@ export default async function startWPNow(
...
 			onStepCompleted: ( result, step: StepDefinition ) => {
 				output?.log( `Blueprint step completed: ${ step.step }` );
 			},
 		} );
 		await runBlueprintSteps( compiled, php );
 	}
 
-	await installationSteps( php, options );
```
wojtekn, line 113: In the current implementation, it breaks site creation when blueprint is used as it tries to activate plugins before installing WordPress. The change won't affect behavior for Studio sites, as `options.blueprintObject ` will remain empty for all sites for now, and the issue only pops out when we modify code to provide blueprint.

ivan-ottinger (approved):
This is exciting! Can't wait until we will be adding full blueprints support.

The proposed changes look good and work as expected. The site is being created correctly with plugins installed and activated:

Markup on 2025-05-29 at 13:10:08

Markup on 2025-05-29 at 13:10:56

I did not observe any regressions.

ivan-ottinger:
Although, the E2E tests are failing 🤔 

```
Error: page.goto: net::ERR_CONNECTION_REFUSED at http://localhost:8881/wp-admin/options-general.php?playground-auto-login=true
```

epeicher (approved):
Thanks for progressing this, it is indeed exciting! I have tested it, and I can see the site correctly created with the plugins installed. LGTM! :shipit: 

CleanShot 2025-05-29 at 14 19 29@2x

epeicher:
> Although, the E2E tests are failing 🤔

I have tested it in my local, and they work as expected. Maybe now it takes a bit more time to start the sites and the timeout needs to be increased? 🤔

wojtekn:
Thanks for reviewing and testing.

> This is exciting! Can't wait until we will be adding full blueprints support.

We will still need to migrate to Playground CLI to use blueprints v2, but it's a good start, as we could even start working on that while still using wp-now and blueprints v1.

sejas (commented):
Nice! thanks for fixing the execution of blueprints. 👍

sejas:
It's worth mentioning that before we start supporting them more broadly, we should check what steps are not compatible with Playground nodejs

wojtekn:
> It's worth mentioning that before we start supporting them more broadly, we should check what steps are not compatible with Playground nodejs

Yes, we would likely need to have a validator or preprocessor that skips or removes unsupported steps.

PR #1407, opened 2025-05-16
Increase HTTP request timeout to 60 seconds

gcsecsey:
## Related issues

Fixes [STU-323 Feature Request: Plugin Updater Setting (to Fix Timeout)](https://linear.app/a8c/issue/STU-323/feature-request-plugin-updater-setting-to-fix-timeout)

## Proposed Changes
This PR addresses timeout issues with plugin updates by increasing the default HTTP request timeout to 60 seconds.

- Added a new MU plugin `0-http-request-timeout.php` that sets the default HTTP request timeout and curl timeouts to 60 seconds

## Testing Instructions

1. Temporarily decrease the timeouts to 1 second:
```diff
--- a/vendor/wp-now/src/wp-now.ts
+++ b/vendor/wp-now/src/wp-now.ts
@@ -623,13 +623,13 @@ async function mountInternalMuPlugins( php: PHP, options: WPNowOptions ) {
                `<?php
                // Set the default HTTP request timeout to 60 seconds
                add_filter( 'http_request_timeout', function() {
-                       return 60;
+                       return 1;
                } );
 
                // Set the default cURL timeout to 60 second
                add_action('http_api_curl', function($curl, $url, $options) {
-                       curl_setopt( $curl, CURLOPT_CONNECTTIMEOUT, 60 );
-                       curl_setopt($curl, CURLOPT_TIMEOUT, 60);
+                       curl_setopt( $curl, CURLOPT_CONNECTTIMEOUT, 1 );
+                       curl_setopt($curl, CURLOPT_TIMEOUT, 1);
                        return $curl;
                }, 1, 3);
```
2. Run `npm start`
3. In one of your sites, open the `functions.php` of the active theme, and add this snippet:
```
add_action(
    'wp_head', function () {
        $response = wp_remote_get('https://httpstat.us/200?sleep=5000'); // 5-second delay

        if (is_wp_error($response) ) {
            echo '<div style="background: #fee; color: #a00; padding: 10px; font-weight: bold;">';
            echo 'Request failed: ' . esc_html($response->get_error_message());
            echo '</div>';
        } else {
            echo '<div style="background: #efe; color: #060; padding: 10px; font-weight: bold;">';
            echo 'Request succeeded: HTTP ' . esc_html(wp_remote_retrieve_response_code($response));
            echo '</div>';
        }
    }
);
```
4. Go to the site's homepage and check that you receive a timeout error
5. Remove the changes from `wp-now.ts`, and kill and restart Studio
[...]

Issue #1167, opened 2025-04-04
Feature Request: Plugin Updater Setting (to Fix Timeout)


swamiofsuccess:
### What

Provide a way to override the default cURL timeout. Issue: Update failed: cURL error 28: Operation timed out after 5011 milliseconds with 683623 out of 861127 bytes received

### Why

Updating plugins on a local site often times out. Since there's no way to modify the php.ini settings, a setting to override the default timeout would be great! It's frustrating to have to manually download the plugin instead of being able to use the WP built-in updater.

gcsecsey:
@swamiofsuccess Thanks for your feedback and for highlighting the timeout issue.

I recently merged https://github.com/Automattic/studio/pull/1407 that increases the default timeout to 60 seconds, which should help in scenarios like the above. This improvement will be included in the upcoming v1.5.2 release.

PR #1456, opened 2025-05-29
Reduce timeout for HTTP request to 30 seconds

epeicher:
## Related issues

- Fixes STU-523
- Related task https://github.com/Automattic/studio/pull/1407

## Proposed Changes

- Reduces the timeout of the HTTP requests from 60 seconds to 30 seconds because in offline mode, this is the timeout it takes to perform some installation steps [step 0 here](https://github.com/Automattic/studio/blob/b8c1ecb8c70ab5af30c089d22fc7bb01e0221c76/vendor/wp-now/src/wp-now.ts#L701) and [step 1 here](https://github.com/Automattic/studio/blob/b8c1ecb8c70ab5af30c089d22fc7bb01e0221c76/vendor/wp-now/src/wp-now.ts#L702). Moreover, the start server default timeout is [2 minutes](https://github.com/Automattic/studio/blob/b8c1ecb8c70ab5af30c089d22fc7bb01e0221c76/src/lib/site-server-process.ts#L14) so it sometimes triggers and the start server fails.

## Testing Instructions

- Disconnect internet so Studio works in Offline mode
- Create a site and make sure the site starts successfully
- This was originally found in Windows, as it was slightly slower, so testing in Windows would also be beneficial.

## Pre-merge Checklist

### DIFF
--- a/src/lib/site-server-process.ts
+++ b/src/lib/site-server-process.ts
@@ -11,7 +11,7 @@ declare const SITE_SERVER_PROCESS_MODULE_PATH: string;
 
 export type MessageName = 'start-server' | 'stop-server' | 'run-php';
 
-const DEFAULT_RESPONSE_TIMEOUT = 180000;
+const DEFAULT_RESPONSE_TIMEOUT = 150000;
 
 export default class SiteServerProcess {
 	lastMessageId = 0;
--- a/src/lib/site-server-process.ts
+++ b/src/lib/site-server-process.ts
@@ -11,7 +11,7 @@ declare const SITE_SERVER_PROCESS_MODULE_PATH: string;
 
 export type MessageName = 'start-server' | 'stop-server' | 'run-php';
 
-const DEFAULT_RESPONSE_TIMEOUT = 150000;
+const DEFAULT_RESPONSE_TIMEOUT = 120000;
 
 export default class SiteServerProcess {
 	lastMessageId = 0;
--- a/src/lib/site-server-process.ts
+++ b/src/lib/site-server-process.ts
@@ -11,7 +11,7 @@ declare const SITE_SERVER_PROCESS_MODULE_PATH: string;
 
 export type MessageName = 'start-server' | 'stop-server' | 'run-php';
 
-const DEFAULT_RESPONSE_TIMEOUT = 120000;
+const DEFAULT_RESPONSE_TIMEOUT = 180000;
 
 export default class SiteServerProcess {
 	lastMessageId = 0;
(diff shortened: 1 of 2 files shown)


epeicher:
@Automattic/yolo, I have increased the timeout for [the waitForResponse](https://github.com/Automattic/studio/blob/51c198366c39c9200b95fe659f42c4adb0d05af7/src/lib/site-server-process.ts#L34) method that is used when starting, stopping, and getting the theme details from 2 minutes to 2.5 minutes. 

The timeout errors were triggered in Offline mode in Windows, which I don't think is a common scenario, while increasing this could mask other potential errors as the user will need thirty seconds more to identify them. 

What do you think about this? 

Note: In this scenario, the majority of the time is spent during [startWPNow](https://github.com/Automattic/studio/blob/51c198366c39c9200b95fe659f42c4adb0d05af7/vendor/wp-now/src/start-server.ts#L46) that takes around `startWPNow: 2:17.323 (m:ss.mmm)`

wojtekn:
@epeicher, we recently increased HTTP request time to 1 min (https://github.com/Automattic/studio/pull/1407). Is this change possible to have caused the issues we are trying to fix here? If yes, is increasing the timeout to three minutes a most optimal solution?

epeicher:
> @epeicher, we recently increased HTTP request time to 1 min (#1407). Is this change possible to have caused the issues we are trying to fix here? If yes, is increasing the timeout to three minutes a most optimal solution?

Thanks @wojtekn for the info! That is exactly the reason. The first two steps [step 0 here](https://github.com/Automattic/studio/blob/b8c1ecb8c70ab5af30c089d22fc7bb01e0221c76/vendor/wp-now/src/wp-now.ts#L701) and [step 1 here](https://github.com/Automattic/studio/blob/b8c1ecb8c70ab5af30c089d22fc7bb01e0221c76/vendor/wp-now/src/wp-now.ts#L702) take to complete the timeout configured in #1407 (around 1 minute). The step 2 only takes seconds after that. The interesting thing is that those steps return a successful response, but I think is due to how [php.requestHandler.request works](https://github.com/Automattic/studio/blob/b8c1ecb8c70ab5af30c089d22fc7bb01e0221c76/vendor/wp-now/src/wp-now.ts#L679).

So given the above, I think we could _reduce_ the timeout changed as part of #1407 from 1 minute to 30 seconds, as [the user reported a timeout after 5 seconds](https://github.com/Automattic/studio/issues/1167), so in principle, there is plenty of time. That way, the estimate for starting the server (in offline mode) is less than 2 minutes, and we don't need to change the [default response timeout](https://github.com/Automattic/studio/blob/b8c1ecb8c70ab5af30c089d22fc7bb01e0221c76/src/lib/site-server-process.ts#L14). What do you think?

gcsecsey (approved):
Thanks for looking into this @epeicher and for iterating on the solution. The reduced timeout works well for me on Windows with the network disabled. The site is created successfully and works as expected.

CleanShot 2025-06-03 at 12 20 50@2x
CleanShot 2025-06-03 at 12 21 58@2x

wojtekn:
Thanks for the fix @epeicher.

Ideally, Studio would detect offline mode and pass information about it to the installer script, so the installer doesn't attempt to make any requests. As far as I see, the installer doesn't support something like that, so we might need to do something else to prevent it from running those calls, e.g., default to English when Studio is in offline mode. However, as it's an edge case, we may leave it for now.

PR #1470, opened 2025-06-06
Add selective sync dialog

nightnei:
## Related issues

- Fixes STU-541, STU-542

## Proposed Changes
With this PR we are adding dialof for Selective sync feature hidden by feature flag.

*Note that*:
1. This PR handles both push and pull
2. This PR, in advance, provides the opportunity for "a more granular selection for syncing individual plugins and themes" (p1749048852126039-slack-C04GESRBWKW).
3. Two things which doesn't follow design:
	3.1 I added indentation for tree item checkboxes (STU-541).
	3.2 I decided to use `SelectControl` from `@wordpress/components` for consistency and it supports control from keyboard, instead of custom dropdown:<br >
	<table>
	<tr>
		<td>DESIGN
		<td>IMPLEMENTATION
	</tr>
	<tr>
		<td>Screenshot 2025-06-06 at 16 14 13
		<td>Screenshot 2025-06-06 at 16 15 44Screenshot 2025-06-06 at 16 16 04
	</tr>
</table>

### Test feature flag
1. run `npm start`
2. Open sync tab
3. Try `Pull` / `Push`
4. Assert that everyhting works well as before
### Test new dialog
3. Try to click on `Pull` / `Push`
4. Assert that you always see "selective sync" dialog and assert that copy are rendered correctly (You see appropriate keywords as: pull, push, production, staging)<br />Screenshot 2025-06-06 at 16 25 51
5. Open browser's dev tool and try to provide long and short site names to check how UI behaves and assert that it works well, examples:<br />Screenshot 2025-06-06 at 16 27 26<br />Screenshot 2025-06-06 at 16 31 01
6. Switch "All files and folders" to "Specific files and folders" and assert that folders tree appeared and it disappears after selecting "All files and folders" back<br />Screenshot 2025-06-06 at 16 33 45
7. Try to play with selecting/deselecting folders to test that everythign works well - `selected`, `deselected` and `indeterminate` states<br />Screenshot 2025-06-06 at 16 35 02
8. Click on Pull or Push button at teh end of teh dialog and assert that you see `console.log` with correct information about selected / deselected items, etc.


sejas:
Thank you for working on this PR and supporting nested directories. I will review it thoroughly on Monday. It looks promising!

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,226 @@
...
+		const element = document.querySelector(
+			'.treeItemIdfilesAndFolders > div:nth-child(2)'
+		) as HTMLElement;
+
+		if ( ! element ) {
+			return;
+		}
+
+		element.style.display = dialogMode === 'advanced' ? 'block' : 'none';
+	}, [ dialogMode ] );
```
nightnei: It looks not very nice, but I decided to go this way, since:
1. the logic is encapsulated inside useEffect, so no problems with it
2. tree-view component looks beautiful, w/o extra functionality specific for this case

src/components/sync-connected-sites.tsx
```diff
@@ -42,6 +44,8 @@ const SyncConnectedSiteControls = ( {
 } ) => {
 	const { __ } = useI18n();
 	const isOffline = useOffline();
+	const { selectiveSyncEnabled } = useFeatureFlags();
+	const [ syncDialog, setSyncDialog ] = useState< { type: 'pull' | 'push' } | null >( null );
```
bcotrim: Why not just `pull | push | null`?

src/components/tree-view.tsx
```diff
@@ -0,0 +1,126 @@
...
+		indeterminate: false,
+		children: node.children ? updateAllChildren( node.children, checked ) : undefined,
+	} ) );
+};
+
+const getParentStateFromChildren = (
+	children: TreeNode[]
+): { checked: boolean; indeterminate: boolean } => {
+	const total = children.length;
+	const checkedCount = children.filter( ( c ) => c.checked ).length;
```
bcotrim: We could do both counts in the same loop using reduce, without making the code harder to read.
```typescript
const { checkedCount, indeterminateCount } = children.reduce(
        (acc, child) => ({
            checkedCount: acc.checkedCount + (child.checked ? 1 : 0),
            indeterminateCount: acc.indeterminateCount + (child.indeterminate ? 1 : 0),
        }),
        { checkedCount: 0, indeterminateCount: 0 }
    );
```

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,226 @@
...
+							<div className="leading-[32px]">{ copy.toLabel }</div>
+							<div className="border border-gray-300 rounded-[2px] min-h-12 px-[19px] flex items-center py-2 break-all">
+								{ syncTo }
+							</div>
+						</div>
+					</div>
+				</div>
+				<div className="px-8 pt-7 pb-3">{ copy.subtitleSelector }</div>
+				<div className="px-8 pb-6 relative">
+					<SelectControl
```
bcotrim: Could this be a toggle?
We only have 2 options but we require the user to open and select the other option from the dropdown, when in fact there's is just 1 option to pick.
In my opinion a toggle would be better UI.

src/components/tree-view.tsx
```diff
@@ -0,0 +1,126 @@
...
+	);
+};
+
+export type TreeViewProps = {
+	tree: TreeNode[];
+	setTree: React.Dispatch< React.SetStateAction< TreeNode[] > >;
+	className?: string;
+};
+
+export const TreeView = ( { tree, setTree, className }: TreeViewProps ) => {
```
bcotrim: Probably resulting from our recent project around accessibility, but I think we should improve this component in that regard.
As it is, VoiceOver has no context of parent/children relations, making it difficult for a user to navigate and understand our options.

src/index.css
```diff
@@ -294,4 +294,20 @@ blockquote {
 .components-button.components-guide__back-button {
 	outline: none;
 	border: 1px solid theme( 'colors.a8c-blueberry' );
-}
\ No newline at end of file
+}
+
+/* Looks not nice, but it's small and better to modify UI of component here to keep JS code clear and easy reusable in the future */
+.sync-tree .treeItemLevel0 > div {
```
bcotrim: I understand we have more exceptions, but I would prefer to do this with tailwind inside the component.
I don't see the benefits of having the layout logic spread across different files/languages.
I do understand we might have different needs in the future, but let's not make the code harder to read today to fix a potential issue in the future.

src/components/sync-dialog.tsx
bcotrim: I agree it's not a great approach.
Have you considered adding a prop to `TreeNode` like `showChildren` or `hideChildren` instead?
I feel it would be a good improvement to the `TreeView` component, easier to read and understand and follows along with React's best practices better.

bcotrim (commented):
Works and looks good! 👍 
Tested both push and pull, playing around with checking/unchecking children and parent nodes, and didn't run into any issues.
Added a few comments, but overall this looks like a solid change.

src/components/tree-view.tsx
nightnei: We have 2 loops, but the current approach is much clearer at first glance than using `reduce` with ternary operations, etc.
```
const { checkedCount, indeterminateCount } = children.reduce(
        (acc, child) => ({
            checkedCount: acc.checkedCount + (child.checked ? 1 : 0),
            indeterminateCount: acc.indeterminateCount + (child.indeterminate ? 1 : 0),
        }),
        { checkedCount: 0, indeterminateCount: 0 }
    );
```
VS
```
const checkedCount = children.filter( ( c ) => c.checked ).length;
const indeterminateCount = children.filter( ( c ) => c.indeterminate ).length;
```

src/components/sync-dialog.tsx
nightnei: I think this question is better to discuss in the design P2 (p58i-ltm-p2). But for teh first iteration I would definitely leave it as is, to speed things up.

src/components/tree-view.tsx
nightnei: It's nice to have. But, I AFAIR, I saw in some P2, that Studio is not obligated to support it.
So, I would go with either supporting the whole app or not focusing on it.

src/index.css
nightnei: Thinking more about it - I don't find it dificult, since it's very common approach, when:
1. We have common component
2. We add some className to make some CSS overrides for specific case. 
CSS changes are very small here, but from JS perspective we would need to add a few conditions which would uglify codebase much more then CSS override.

src/components/sync-dialog.tsx
nightnei: I don't like `showChildren ` / `hideChildren ` sicne in context of compoent it makes no sense, w/o context of how it's used.
But I see your point, I will think about it and more approaches

src/components/sync-connected-sites.tsx
```diff
@@ -42,6 +44,8 @@ const SyncConnectedSiteControls = ( {
 } ) => {
 	const { __ } = useI18n();
 	const isOffline = useOffline();
+	const { selectiveSyncEnabled } = useFeatureFlags();
+	const [ syncDialogType, setSyncDialogType ] = useState< 'pull' | 'push' | null >( null );
```
gavande1, line 48: Nit: Should we use a custom type rather than using hardcoded inline options?

wojtekn:
Great start! I'm sharing a few UX-related comments:

1. I think we should display the provider name (WordPress.com or Pressable) and not the domain here, or clean it in another way. I will open another thread to discuss that:

Screenshot 2025-06-09 at 11 49 04

2. Can we make a bucket or directory name clickable, so it could be used to select or deselect the item?

3.  Do we need a 'Specific files and folders' dropdown at all? What if we made it consistent with child items and had the same arrow selector for the first item, too?

4. Let's link the environment push to https://developer.wordpress.com/docs/developer-tools/studio/sync/#push and https://developer.wordpress.com/docs/developer-tools/studio/sync/#pull respectively.

5. We can replace "Pull from" with a "Pull" to make it consistent with "Push" windows. 

6. I would play with spacing to have the whole content visible in the default screen state, when all items are collapsed.

src/components/icons/circle-staging.tsx
```diff
@@ -0,0 +1,24 @@
+export const CircleStagingIcon = () => (
+	<svg xmlns="http://www.w3.org/2000/svg" width="13" height="13" viewBox="0 0 13 13" fill="none">
+		<rect
+			x="0.432861"
+			y="0.613525"
+			width="12.2075"
+			height="12.2075"
+			rx="6.10373"
+			fill="#F7BA42"
+			fill-opacity="0.4"
```
sejas:
```suggestion
			fillOpacity="0.4"
```

src/components/icons/circle-staging.tsx
```diff
@@ -0,0 +1,24 @@
...
+		/>
+		<rect
+			x="1.38657"
+			y="1.56723"
+			width="10.3001"
+			height="10.3001"
+			rx="5.15003"
+			stroke="white"
+			stroke-opacity="0.6"
+			stroke-width="1.90742"
```
sejas:
```suggestion
			strokeOpacity="0.6"
			strokeWidth="1.90742"
```

src/components/icons/folder.tsx
```diff
@@ -0,0 +1,10 @@
+export const FolderIcon = () => (
+	<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20" fill="none">
+		<path
+			fill-rule="evenodd"
+			clip-rule="evenodd"
```
sejas:
```suggestion
			fillRule="evenodd"
			clipRule="evenodd"
```

src/components/icons/circle-prod.tsx
```diff
@@ -0,0 +1,16 @@
+export const CircleProdIcon = () => (
+	<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
+		<rect width="16" height="16" rx="8" fill="#4AB866" fill-opacity="0.4" />
```
sejas:
```suggestion
		<rect width="16" height="16" rx="8" fill="#4AB866" fillOpacity="0.4" />
```

sejas (commented):
I noticed a few error logs in the console. Let's use the correct React attribute names in camel case in the SVG files.

wojtekn:
> Do we need a 'Specific files and folders' dropdown at all? What if we made it consistent with child items and had the same arrow selector for the first item, too?

I discussed this with @Marinatsu more and we agreed we should leave it there. In other case, users might miss this feature completely.

> I would play with spacing to have the whole content visible in the default screen state, when all items are collapsed.

Instead of playing with spacing, let's make the footer sticky, similar to Connect site window. We would have "Read more..," and actions in the same sticky line, with grey border above.

src/components/tree-view.tsx
sejas: As discussed during team meeting, let's work on A11y improvements for these components in a new issue STU-561.

src/components/sync-connected-sites.tsx
nightnei: We don't reuse this type. 
Also, the type definition is very small and keeps good readability - we immidiatelly see possible options.
So, from my perspective - creating a separate type would just add lines of code and make code less readable.

src/components/tree-view.tsx
wojtekn: Thinking about the tree more, should we consider using an existing component for that, e.g., https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-treegrid--docs instead of writing our own? It seems it has accessibility built-in.

nightnei: >Thinking about the tree more, should we consider using an existing component for that, e.g., [wordpress.github.io/gutenberg?path=/docs/components-experimental-treegrid--docs](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-treegrid--docs) instead of writing our own? It seems it has accessibility built-in.

Why I don't like this component:
1. It's created with the help of `<table>`. I have no idea what the reason was to go with a table, but it's a painful limitation for HTML flexibility, which I would like not to bring to Studio. Maybe this component is very-very old.
2. a11y - I looked at the html and tested this component with VoiceOver - it doesn't provide information about parents.

So I would leave it as is. Our implementation is small, w/o redundant things and workarounds.

nightnei:
> 2. Can we make a bucket or directory name clickable, so it could be used to select or deselect the item?

> 4. Let's link the environment push to [developer.wordpress.com/docs/developer-tools/studio/sync#push](https://developer.wordpress.com/docs/developer-tools/studio/sync/#push) and [developer.wordpress.com/docs/developer-tools/studio/sync#pull](https://developer.wordpress.com/docs/developer-tools/studio/sync/#pull) respectively.

> 5. We can replace "Pull from" with a "Pull" to make it consistent with "Push" windows.

> 6. Instead of playing with spacing, let's make the footer sticky, similar to Connect site window. We would have "Read more..," and actions in the same sticky line, with grey border above.

@wojtekn JIC it's done

src/components/icons/circle-prod.tsx
```diff
@@ -0,0 +1,16 @@
+export const CircleProdIcon = () => (
+	<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
+		<rect width="16" height="16" rx="8" fill="#4AB866" fillOpacity="0.4" />
```
bcotrim: Adding a `title` is a good a11y practice.
```suggestion
        <title>Production Environment</title>
		<rect width="16" height="16" rx="8" fill="#4AB866" fillOpacity="0.4" />
```

src/components/icons/circle-staging.tsx
```diff
@@ -0,0 +1,24 @@
+export const CircleStagingIcon = () => (
```
bcotrim: Same title suggestion as on the other `svg` file.

src/components/icons/folder.tsx
```diff
@@ -0,0 +1,10 @@
+export const FolderIcon = () => (
```
bcotrim, line 1: Same `title` suggestion as the ones above

src/components/icons/right-arrow.tsx
```diff
@@ -0,0 +1,11 @@
+export const RightArrowIcon = ( { width = 24 } ) => (
```
bcotrim, line 1: Same `title` suggestion

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,214 @@
...
+				<div className="px-8">
+					<div className="flex items-start gap-1 pb-7 border-b border-a8c-gray-5">
+						<div className="flex-1">
+							<div className="leading-[32px]">{ copy.fromLabel }</div>
+							<div className="border border-gray-300 rounded-[2px] min-h-12 px-[19px] flex items-center py-2 break-all">
+								{ syncFrom }
+							</div>
+						</div>
+						<div className="w-10 mt-[32px] h-12 flex items-center justify-center">
+							<RightArrowIcon />
```
bcotrim, line 235: Using VoiceOver this reads as "image".
The `title` comment above should fix it

src/components/tree-view.tsx
```diff
@@ -0,0 +1,151 @@
...
+				</label>
+				{ node.children && ! node.customExpanderOptions && (
+					<button onClick={ () => setExpanded( ! expanded ) }>
+						<div className={ expanded ? 'rotate-90' : '' }>
+							<RightArrowIcon width={ 16 } />
+						</div>
+					</button>
+				) }
+
+				{ node.customExpanderOptions && (
```
bcotrim: I still think this implementation has some issues.
For example, if multiple "nodes" have `customExpanderOptions` this component will render multiple `customExpanderOptions` on top of each other.
I would still prefer the `TreeView` component to expose a prop that allows this behavior, but leaving the logic and UI responsibility to the parent component. Trying to build everything inside this component will make it harder to read, understand and implement.

Another approach could be to make this component specifically to this use case, and remove all of this abstraction we are trying to create.

src/index.css
sejas: I agree with @bcotrim , Let's try to avoid using regular CSS when we have the possibility to use Tailwind inside the component.

Here is a diff that can be used as reference:

```diff
diff --git a/src/components/tree-view.tsx b/src/components/tree-view.tsx
index 658d8e28..0622b153 100644
--- a/src/components/tree-view.tsx
+++ b/src/components/tree-view.tsx
@@ -84,9 +84,11 @@ const TreeItem = ( {
 		node.defaultExpanded !== undefined ? node.defaultExpanded : true
 	);
 
+	const isLevel0 = level === 0;
+
 	return (
-		<div className={ cx( `treeItemLevel${ level }` ) }>
-			<div className="flex items-center py-2">
+		<div className={ cx( isLevel0 && 'border-b border-gray-300 py-2' ) }>
+			<div className={ cx( 'flex items-center py-2', isLevel0 && 'py-4' ) }>
 				<label className="flex items-center">
 					<CheckboxControl
 						checked={ node.checked }

```

src/components/icons/circle-staging.tsx
sejas: Do we need different components/icons for each environment, or can we have a single component that changes color based on the passed environment?

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,214 @@
...
+			description: __(
+				"Pulling will replace your Studio site's files and database with a copy from your production site."
+			),
+		},
+		fromLabel: __( 'Pull' ),
+		toLabel: __( 'To' ),
+		subtitleSelector: __( 'What would you like to pull?' ),
+		envSync: __( 'Read more about <a>environment pull</a> ↗' ),
+		submit: __( 'Pull' ),
+	},
```
sejas: Can we move these copies into a hook , and use the translation function `const { __ } = useI18n();` to make the translations reactive?

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,214 @@
...
+type SyncDialogProps = {
+	type: 'push' | 'pull';
+	localSite: SiteDetails;
+	remoteSite: SyncSite;
+	onSubmit: ( syncData: TreeNode[] ) => void;
+	onRequestClose: () => void;
+};
+
+export function SyncDialog( {
+	type = 'push',
```
sejas: Any reason to default to push?
I suggest deleting the default value to avoid future bugs.

```suggestion
	type,
```

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,214 @@
...
+		},
+		{
+			id: 'database',
+			label: 'Database',
+			checked: true,
+		},
+	] );
+
+	const copy = allCopy[ type ];
+	const remoteSiteType = remoteSite.isStaging ? 'staging' : 'production';
```
sejas: Let's take into account also the `sandbox` environment type. We can use `environmentType` to infer it.

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+	const localSiteName = localSite.name;
+	const remoteSiteName = (
+		<div className="flex items-center gap-2 flex-wrap">
+			<span className="flex-shrink-0">
+				{ remoteSite.isStaging ? <CircleStagingIcon /> : <CircleProdIcon /> }
+			</span>
+			<span>{ remoteSite.isStaging ? 'Staging' : 'Production' }</span>
```
sejas: Let's take into account also sandbox sites.

sejas: Let's translate the environment type:

```suggestion
			<span>{ remoteSite.isStaging ? __( 'Staging' ) : __( 'Production' ) }</span>
```

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,214 @@
...
+	remoteSite,
+	onSubmit,
+	onRequestClose,
+}: SyncDialogProps ) {
+	const { locale } = useI18nData();
+
+	const [ treeState, setTreeState ] = useState< TreeNode[] >( [
+		{
+			id: 'filesAndFolders',
+			label: 'Files and folders',
```
sejas: Let's translate all labels:

```suggestion
			label: __( 'Files and folders' ),
```

sejas (commented):
I left a few comments.
Let's also add the cases to display sandbox environments.

From my point of view, we could separate the logic of the files tree and the first "Files and Folders" checkbox. If we have separate states, we could keep the user's file selections in one state and the "Files and Folders" checkbox in a different state. Unchecking the "Files and Folders" checkbox could leave the files as they are selected but mark them as disabled.

Screenshot 2025-06-10 at 10 51 51

src/components/tree-view.tsx
sejas: Agree. I suggest having two components. One for the main checkbox with the dropdown and simplify the logic of the TreeView to just render the folders structure and nothing else.

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,214 @@
...
+		production: {
+			title: __( 'Pull from Production' ),
+			description: __(
+				"Pulling will replace your Studio site's files and database with a copy from your production site."
+			),
+		},
+		fromLabel: __( 'Pull' ),
+		toLabel: __( 'To' ),
+		subtitleSelector: __( 'What would you like to pull?' ),
+		envSync: __( 'Read more about <a>environment pull</a> ↗' ),
```
sejas: Let's move the arrow inside the link.

```suggestion
		envSync: __( 'Read more about <a>environment pull ↗</a>' ),
```

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,214 @@
...
+		production: {
+			title: __( 'Push to Production' ),
+			description: __(
+				'Pushing will replace the existing files and database with a copy from your local site.\n\n The production site will be backed-up before any changes are applied.'
+			),
+		},
+		fromLabel: __( 'Push' ),
+		toLabel: __( 'To' ),
+		subtitleSelector: __( 'What would you like to push?' ),
+		envSync: __( 'Read more about <a>environment push</a> ↗' ),
```
sejas:
```suggestion
		envSync: __( 'Read more about <a>environment push ↗</a>' ),
```

src/components/tree-view.tsx
wojtekn: Yeah, it's marked as experimental. However, we can at least look there for ideas for accessibility support.

wojtekn:
@nightnei, thanks for implementing those. Small follow-ups related to styling:

1. Can we have a cursor pointer for the directory name, too, as it's clickable now?
2. We should remove the line below "Database" or add more space there, as it gets glued to the footer border
3. Can we add a heading border and move the scrollbar outside the footer, so it would be consistent with the 'Connect site' dialog?

Screenshot 2025-06-10 at 13 00 03

src/components/tree-view.tsx
sejas: I took a brief look at the TreeGrid component. My main concern is that it is experimental and would require considerable effort to adapt to our needs. I think it's okay to continue with our current implementation.

sejas (commented):
@nightnei , I noticed a few issues with RTL languages and created a quick PR against this branch to fix them: https://github.com/Automattic/studio/pull/1474

src/components/sync-dialog.tsx
sejas: This will be fixed once https://github.com/Automattic/studio/pull/1474 is merged.

src/components/sync-dialog.tsx
sejas: This will be fixed once https://github.com/Automattic/studio/pull/1474 is merged.

src/components/tree-view.tsx
nightnei: > For example, if multiple "nodes" have customExpanderOptions this component will render multiple customExpanderOptions on top of each other.

Fixed in the current approach

> @bcotrim  Trying to build everything inside this component will make it harder to read, understand and implement.
@sejas I suggest having two components. One for the main checkbox with the dropdown and simplify the logic of the TreeView to just render the folders structure and nothing else.

Yes, that's why I tried initially to encapsulate this behavior inside useEffect to keep it small and simple. However the code wasn't perfect, it still was good tradeoff https://github.com/Automattic/studio/pull/1470/commits/dfade6c242eadd66647c361820954a9046f9e627#diff-04017bc3527cf42a0d80444eea8588a2f1f58bd4e38d095ad006aa8b20580309L142-L152

> I would still prefer the TreeView component to expose a prop that allows this behavior, but leaving the logic and UI responsibility to the parent component.

What is your proposition for comunication between these two components?
As I said above, it was my initial approach and I wanted to keep the simple logic of TreeView to avoid extra functionality, with the help of manipulating the DOM. Otherwise, we would need to provide some toggler to switch teh state of specific nodes, which would make code not simple.

src/components/icons/circle-prod.tsx
nightnei: Mentioned it in this issue, since necessary to test it carefully and also address localization
STU-561

nightnei:
> 1. Can we have a cursor pointer for the directory name, too, as it's clickable now?

> 2. We should remove the line below "Database" or add more space there, as it gets glued to the footer border

> 3. Can we add a heading border and move the scrollbar outside the footer, so it would be consistent with the 'Connect site' dialog?

@wojtekn done

Screenshot 2025-06-11 at 15 36 44
Screenshot 2025-06-11 at 15 37 10

src/components/icons/circle-staging.tsx
nightnei: I didn't analyze it, but svg code is different. If you prefer to have one icon, I think we can agree it with Marina, to unify and just change color.

nightnei:
@sejas @bcotrim I addressed everything in this PR, except things around rendering TreeView. I propose to have a call tomorrow, otherwise, I anticipate a long writing discussion here. So let's speed thigns up with a call, to quckly align us.

src/components/tree-view.tsx
bcotrim: @nightnei I have made a quick example with how my approach would look like:
https://github.com/Automattic/studio/pull/1476

src/index.css
```diff
@@ -62,7 +62,7 @@ blockquote {
 	top: 16px;
 }
 
-.components-tab-panel__tabs--assistant:hover:not(:active)::before {
```
sejas: I think we can undo these CSS changes.

src/components/sync-connected-sites.tsx
```diff
@@ -189,6 +203,23 @@ const SyncConnectedSiteControls = ( {
 						</Button>
 					</DynamicTooltip>
 				) }
+				{ syncDialogType && (
+					<SyncDialog
+						type={ syncDialogType }
+						localSite={ selectedSite }
+						remoteSite={ connectedSite }
+						onSubmit={ ( dataToSync ) => {
+							console.log( dataToSync );
```
sejas: We can remove this console.log
```suggestion
```

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,253 @@
...
+			<span className="flex-shrink-0">
+				{ <CircleEnvIcon color={ remoteSiteType === 'production' ? '#069e08' : '#f7ba42' } /> }
+			</span>
+			<span>
+				{ remoteSiteType === 'staging'
+					? __( 'Staging' )
+					: remoteSiteType === 'sandbox'
+					? __( 'Sandbox' )
+					: __( 'Production' ) }
+			</span>
```
sejas: We need to take into account the CircleEnvIcon sandbox color.
Let's avoid using nested ternary conditions.
Let's declare the colors in tailwind instead of hardcoding them in the UI.

Here is a diff that could be used as reference to address the mentioned comments above.

 ```diff
diff --git a/src/components/icons/circle-env.tsx b/src/components/icons/circle-env.tsx
index aba42fd9..8bd3a1fc 100644
--- a/src/components/icons/circle-env.tsx
+++ b/src/components/icons/circle-env.tsx
@@ -1,4 +1,4 @@
-export const CircleEnvIcon = ( { color }: { color: string } ) => (
+export const CircleEnvIcon = ( { fillClass }: { fillClass: string } ) => (
 	<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
 		<rect width="16" height="16" rx="8" fill="#4AB866" fillOpacity="0.4" />
 		<rect
@@ -11,6 +11,6 @@ export const CircleEnvIcon = ( { color }: { color: string } ) => (
 			strokeOpacity="0.6"
 			strokeWidth="2.5"
 		/>
-		<rect x="2" y="2" width="12" height="12" rx="6" fill={ color } />
+		<rect x="2" y="2" width="12" height="12" rx="6" className={ fillClass } />
 	</svg>
 );
diff --git a/src/components/sync-dialog.tsx b/src/components/sync-dialog.tsx
index b0928a86..99b64603 100644
--- a/src/components/sync-dialog.tsx
+++ b/src/components/sync-dialog.tsx
@@ -73,16 +73,36 @@ const useCopy = ( type: 'pull' | 'push' ) => {
 
 type EnvType = 'sandbox' | 'staging' | 'production';
 
-const getEnvType = ( connectedSite: SyncSite ): EnvType => {
+const useEnvDetails = (
+	connectedSite: SyncSite
+): { label: string; envType: EnvType; fillClass: string } => {
+	const { __ } = useI18n();
+
+	const envTypeValues = {
+		production: {
+			label: __( 'Production' ),
+			envType: 'production',
+			fillClass: 'fill-circle-env-production',
+		},
+		staging: {
+			label: __( 'Staging' ),
+			envType: 'staging',
+			fillClass: 'fill-circle-env-staging',
+		},
+		sandbox: {
+			label: __( 'Sandbox' ),
+			envType: 'sandbox',
+			fillClass: 'fill-sandbox-text',
+		},
+	} as const;
+
 	if ( connectedSite.isPressable ) {
-		return connectedSite.environmentType as EnvType;
+		return envTypeValues[ connectedSite.environmentType as EnvType ] ?? envTypeValues.production;
+	} else if ( connectedSite.isStaging ) {
+		return envTypeValues.staging;
 	}
 
-	if ( connectedSite.isStaging ) {
-		return 'staging';
-	} else {
-		return 'production';
-	}
+	return envTypeValues.production;
 };
 
 type SyncDialogProps = {
@@ -161,21 +181,15 @@ export function SyncDialog( {
 		},
 	] );
 
-	const remoteSiteType = getEnvType( remoteSite );
+	const envDetails = useEnvDetails( remoteSite );
 
 	const localSiteName = localSite.name;
 	const remoteSiteName = (
 		<div className="flex items-center gap-2 flex-wrap">
 			<span className="flex-shrink-0">
-				{ <CircleEnvIcon color={ remoteSiteType === 'production' ? '#069e08' : '#f7ba42' } /> }
-			</span>
-			<span>
-				{ remoteSiteType === 'staging'
-					? __( 'Staging' )
-					: remoteSiteType === 'sandbox'
-					? __( 'Sandbox' )
-					: __( 'Production' ) }
+				<CircleEnvIcon fillClass={ envDetails.fillClass } />
 			</span>
+			<span>{ envDetails.label }</span>
 			<span className="text-gray-600 break-all">
 				{ remoteSite.url.replace( /https?:\/\//, '' ) }
 			</span>
@@ -195,10 +209,10 @@ export function SyncDialog( {
 		<Modal
 			className="w-3/5 min-w-[550px] h-full max-h-[84vh] [&>div]:!p-0"
 			onRequestClose={ onRequestClose }
-			title={ copy[ remoteSiteType ].title }
+			title={ copy[ envDetails.envType ].title }
 		>
 			<div className="pb-[70px]">
-				<div className="px-8 pb-6 pt-2">{ copy[ remoteSiteType ].description }</div>
+				<div className="px-8 pb-6 pt-2">{ copy[ envDetails.envType ].description }</div>
 				<div className="px-8">
 					<div className="flex items-start gap-1 pb-7 border-b border-a8c-gray-5">
 						<div className="flex-1">
diff --git a/tailwind.config.js b/tailwind.config.js
index 909b9ebf..d0928df4 100644
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -148,6 +148,8 @@ module.exports = {
 				'chrome-inverted': '#fff',
 				'sandbox-bg': 'hsl(200, 95%, 85%)',
 				'sandbox-text': 'hsl(200, 95%, 28%)',
+				'circle-env-production': '#069e08',
+				'circle-env-staging': '#f7ba42',
 			},
 			spacing: {
 				chrome: `${ APP_CHROME_SPACING }px`,

```

File: src/components/icons/circle-env.tsx [hunk 28]
```diff
@@ -0,0 +1,16 @@
+export const CircleEnvIcon = ( { color }: { color: string } ) => (
+	<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
+		<rect width="16" height="16" rx="8" fill="#4AB866" fillOpacity="0.4" />
```
sejas: We should display a different glow, `#4AB866`, for each environment.  
Maybe instead of receiving the color, the CircleEnvIcon could receive the EnvironmentType, similarly to `EnvironmentBadge`.

https://github.com/Automattic/studio/blob/913a2070fa0d27fb2b46156353334bf2276b361f/src/components/environment-badge.tsx#L5-L15

```suggestion
		<rect width="16" height="16" rx="8" fill="#4AB866" fillOpacity="0.4" />
```

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,253 @@
...
+			fromLabel: __( 'Push' ),
+			toLabel: __( 'To' ),
+			subtitleSelector: __( 'What would you like to push?' ),
+			envSync: __( 'Read more about <a>environment push <ArrowIcon /></a>' ),
+			submit: __( 'Push' ),
+		};
+	}
+};
+
+type EnvType = 'sandbox' | 'staging' | 'production';
```
sejas: These types are being also declared on  EnvironmentBadge. Let's declare them only once.

https://github.com/Automattic/studio/blob/913a2070fa0d27fb2b46156353334bf2276b361f/src/components/environment-badge.tsx#L5

sejas (commented):
Vova, thanks for updating the PR, and discussing it during huddle.
I left a few suggestions related to the EnvType.

src/components/icons/circle-env.tsx
nightnei: Looking at the desiugn and Staging border color, I suppose that the goal is to keep teh same color but 0.4 opacity.
Done this way, it looks good.

sejas (approved):
Great work @nightnei !, Thanks for the updates and separating the logic of the files and folders dropdown.

https://github.com/user-attachments/assets/3a0cc1d6-f18d-4950-b67d-392713f9084c

My only concern is that the URL is displayed on multiple lines. Have we considered Wojtek's comment? or using ellipsis or any other solution?

> 1. I think we should display the provider name (WordPress.com or Pressable) and not the domain here, or clean it in another way. I will open another thread to discuss that:
> Screenshot 2025-06-09 at 11 49 04

Let's also fix the tests.

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,288 @@
...
+	localSite,
+	remoteSite,
+	onSubmit,
+	onRequestClose,
+}: SyncDialogProps ) {
+	const locale = useI18nLocale();
+	const copy = useCopy( type );
+
+	const [ showAllFiles, setShowAllFiles ] = useState( false );
+	const [ treeState, setTreeState ] = useState< TreeNode[] >( [
```
bcotrim, line 133: Some of these labels are not translated.
I'm not sure if it makes sense to translate everything, but some I think should be, "Database" for example

src/components/sync-dialog.tsx
```diff
@@ -0,0 +1,288 @@
...
+import Modal from 'src/components/modal';
+import { TreeView, TreeNode, updateNodeById } from 'src/components/tree-view';
+import { getIpcApi } from 'src/lib/get-ipc-api';
+import { getLocalizedLink } from 'src/lib/get-localized-link';
+import { useI18nLocale } from 'src/stores';
+import { CircleEnvIcon } from './icons/circle-env';
+import type { EnvironmentType } from 'src/components/environment-badge';
+import type { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
+
+const useCopy = ( type: 'pull' | 'push' ) => {
```
bcotrim, line 18: Since this hook returns an object, I would consider memoizing the return value, otherwise we have a new object on every render (even if the type hasn't changed)

bcotrim (approved):
Code is looking great. Good job on the changes!
I added a couple of very minor comments but 👍 

During testing I noticed something:
I would expect the checkbox to be checked after selection "All files and folders". 

https://github.com/user-attachments/assets/fc568e1e-f61e-4f9d-80b2-8fa3b1e27447

src/components/sync-dialog.tsx
nightnei: I think only Database should be translated, sin ce otehr folders are physical and not translated for customers on OS.
Translated Database, good catch 👍

sejas:
> My only concern is that the URL is displayed on multiple lines. Have we considered Wojtek's comment? or using ellipsis or any other solution?
> 
> > 1. I think we should display the provider name (WordPress.com or Pressable) and not the domain here, or clean it in another way. I will open another thread to discuss that:
> > 
> > Screenshot 2025-06-09 at 11 49 04
> 
> Let's also fix the tests.

Let's fix it in a separate issue, STU-572.

wojtekn:
Thanks for adding another issue for changing site box design @sejas . We can definitely work on this separately.

@nightnei I've reviewed the current UI, and it looks good. Thanks for implementing the proposed changes. The only small thing I noticed is that the top border for the scrollable area appears only after I start scrolling from the top position. The bottom border is displayed immediately, and the "Connect site" has both borders visible immediately.

PR #1478, opened 2025-06-13
 [Experiment] Update network timeout to 1 second when network is offline

gavande1:
## Related issues

- Fixes STU-550

## Proposed Changes

> [!WARNING]
> This is an experimental PR. I would like to know what others think about the approach first. 

I propose updating the network request timeout to 1 second when the studio is running in offline mode. By default, WordPress is not built for offline mode, and it assumes that it is running on a server with an internet connection. That's why it doesn't have checks in place for network status. The reason for the slow site creation is the network requests that WordPress makes during the initial load. While we can not prevent WordPress from making those requests, But we can reduce the timeout for network requests when the site is running in offline mode. So this PR does the following:

1. Updates the timeout based on network status. 1 second when offline and restores it to 30 seconds when online. 
2. Automatically restart running servers when the network changes.
    - I am aware this has some consequences. Like logging out from WP Admin on reload. 
    - Network fluctuations will automatically restart servers, even in the event of shorter disruptions. 

In tests:
This change has drastically improved timing in offline mode. The new sites are created almost instantly, and WP Admin loads instantly. 

https://github.com/user-attachments/assets/44400964-d93b-4a59-82e1-e5913b534a9a

## Testing Instructions

- Check out this branch
- Turn off the Wi-Fi adaptor or the network
- Create a new site and observe the timing
- Open WP Admin and observe the timing
- Both of the above should be improved in comparison to the trunk
- Turn on internet
- Observe the studio UI; all running servers will be automatically restarted. 

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-network-connection.ts
+++ b/src/hooks/use-network-connection.ts
@@ -1,6 +1,6 @@
 import { useEffect } from 'react';
-import { useSiteDetails } from './use-site-details';
 import { useOffline } from './use-offline';
+import { useSiteDetails } from './use-site-details';
 
 export function useNetworkConnection() {
 	const { data: sites, stopServer, startServer } = useSiteDetails();
--- a/src/hooks/use-network-connection.ts
+++ b/src/hooks/use-network-connection.ts
@@ -1,0 +1,23 @@
+import { useEffect } from 'react';
+import { useSiteDetails } from './use-site-details';
+import { useOffline } from './use-offline';
+
+export function useNetworkConnection() {
+    const { data: sites, stopServer, startServer } = useSiteDetails();
+    const isOffline = useOffline();
+
+    const restartRunningSites = async () => {
+        const runningSites = sites.filter(site => site.running);
+        await Promise.all(runningSites.map(site => stopServer(site.id)));
+        await Promise.all(runningSites.map(site => startServer(site.id)));
+    };
+
+    useEffect(() => {
+        window.addEventListener('online', restartRunningSites);
+        window.addEventListener('offline', restartRunningSites);
+        return () => {
+            window.removeEventListener('online', restartRunningSites);
+            window.removeEventListener('offline', restartRunningSites);
+        };
+    }, [sites, stopServer, startServer, isOffline]);
+} 
--- a/src/hooks/use-network-connection.ts
+++ b/src/hooks/use-network-connection.ts
@@ -3,21 +3,21 @@ import { useSiteDetails } from './use-site-details';
 import { useOffline } from './use-offline';
 
 export function useNetworkConnection() {
-    const { data: sites, stopServer, startServer } = useSiteDetails();
-    const isOffline = useOffline();
+	const { data: sites, stopServer, startServer } = useSiteDetails();
+	const isOffline = useOffline();
 
-    const restartRunningSites = async () => {
-        const runningSites = sites.filter(site => site.running);
-        await Promise.all(runningSites.map(site => stopServer(site.id)));
-        await Promise.all(runningSites.map(site => startServer(site.id)));
-    };
+	const restartRunningSites = async () => {
+		const runningSites = sites.filter( ( site ) => site.running );
+		await Promise.all( runningSites.map( ( site ) => stopServer( site.id ) ) );
+		await Promise.all( runningSites.map( ( site ) => startServer( site.id ) ) );
+	};
 
-    useEffect(() => {
-        window.addEventListener('online', restartRunningSites);
-        window.addEventListener('offline', restartRunningSites);
-        return () => {
-            window.removeEventListener('online', restartRunningSites);
-            window.removeEventListener('offline', restartRunningSites);
-        };
-    }, [sites, stopServer, startServer, isOffline]);
-} 
+	useEffect( () => {
+		window.addEventListener( 'online', restartRunningSites );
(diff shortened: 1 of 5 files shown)


src/hooks/use-network-connection.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react';
+import { useOffline } from './use-offline';
+import { useSiteDetails } from './use-site-details';
+
+export function useNetworkConnection() {
+	const { data: sites, stopServer, startServer } = useSiteDetails();
+	const isOffline = useOffline();
+
+	const restartRunningSites = async () => {
```
gavande1, line 9: This would update the timeout to 30 or 1 based on network status.

bcotrim (commented):
Nice improvement!
The difference in creating a site offline is very noticeable.

Have you considered displaying a notification to the user instead of automatically restarting? To me it feels risky to restart the server without any user action. As you mentioned it could be a side effect of an unstable network.
I was thinking of something like: "Have you gone offline/online? Restarting running server could improve performance."

Anyways big 👍 for this approach, I think it's a great improvement to our user experience.

wojtekn:
Nice experiment!

> While we can not prevent WordPress from making those requests, But we can reduce the timeout for network requests when the site is running in offline mode.

Thinking about alternative solutions - would it be possible to identify the most common requests WordPress makes during installation, and disable them using a filter/action hook when Studio is in offline mode?

Also, how can the 1s timeout affect local requests that would exceed it?

gavande1:
> Have you considered displaying a notification to the user instead of automatically restarting? To me it feels risky to restart the server without any user action. As you mentioned it could be a side effect of an unstable network.
I was thinking of something like: "Have you gone offline/online? Restarting running server could improve performance."

Thanks @bcotrim for reviewing. I thought about it, so I implemented it to demonstrate what can be done. It would be a good idea to bring in designers to discuss the best way to communicate those changes. 

> would it be possible to identify the most common requests WordPress makes during installation, and disable them using a filter/action hook when Studio is in offline mode?

@wojtekn  I identified the request to obtain translations from core servers, which is made by [wp_get_available_translations](https://github.com/WordPress/WordPress/blob/217f3d106fd825d2b42391e17b1b225f4f065bd5/wp-admin/includes/translation-install.php#L167), which affects the installation process. That request is only sent if [wp_can_install_language_pack](https://github.com/WordPress/WordPress/blob/217f3d106fd825d2b42391e17b1b225f4f065bd5/wp-admin/includes/translation-install.php#L302) returns true. We can apply a filter on `wp_can_install_language_pack`, but that would require us to add the `DISALLOW_FILE_MODS` constant somewhere, which leads to more complications. In fact, that was my first approach to this problem. 

> how can the 1s timeout affect local requests that would exceed it?

AFAIK, it shouldn't affect local requests as the 1 sec limit would apply to any code that uses [WP_Http](https://github.com/WordPress/wordpress-develop/blob/6.8.1/src/wp-includes/class-wp-http.php#L34C7-L34C14) class to make requests. There might be a possibility where the developer writes code that communicates with other services running locally.

wojtekn:
@gavande1 thanks for sharing details.

Thinking from another side - is it something that affects other local environment tools? 

Also, what if we made the 1s timeout change only for sites being created and only for the time they are being created? That way, we would speed up site creation, but we wouldn't change the behavior of sites that already exist, and we wouldn't need to add any UI to inform about the restart.

gavande1:
> is it something that affects other local environment tools?

I am sorry, but I didn't understand. What do you mean by other local environment tools? 

> what if we made the 1s timeout change only for sites being created and only for the time they are being created?

Absolutely, we can try to apply this 1-second limit more specifically, especially during the installation of WordPress.

wojtekn:
> I am sorry, but I didn't understand. What do you mean by other local environment tools?

Local, wp-env etc.

> Absolutely, we can try to apply this 1-second limit more specifically, especially during the installation of WordPress.

Sounds good, I think making the change for this part would solve >80% of the problem without introducing confusion.

gavande1:
> Local, wp-env etc.

Please correct me if I am wrong, but I believe the shared folder is site-specific. This means that files under `PLAYGROUND_INTERNAL_MU_PLUGINS_FOLDER` are not shared across sites and have a 1:1 relationship with each site. As a result, each site can have its own customized MU plugins.

If that is accurate, then the changes introduced in this PR should not affect any other tools.

wojtekn:
@gavande1 no, I mean how this issue appears in other local development environments. Does it affect them too? Or if they also rely on timeouts and site creation in offline mode takes longer? Or maybe they don't allow to create site in offline mode?

gavande1:
> I mean how this issue appears in other local development environments. Does it affect them too? Or if they also rely on timeouts and site creation in offline mode takes longer? Or maybe they don't allow to create site in offline mode?

Thanks for the clarification. I will study other tools for their behaviour in offline and post my findings here.

gavande1:
@wojtekn I created [another PR](https://github.com/Automattic/studio/pull/1491) after studying different tools. We have kind of reached the limitations of PHP Wasm. I added an extended explanation in the [PR](https://github.com/Automattic/studio/pull/1491).

gavande1:
Closing this in favor of https://github.com/Automattic/studio/pull/1491

PR #1480, opened 2025-06-13
Send selected sync data

nightnei:
## Related issues

- Fixes STU-567

## Proposed Changes
With this PR we send Selective sync option to backend

## Testing Instructions
Apply this diff:
```
diff --git a/src/components/sync-connected-sites.tsx b/src/components/sync-connected-sites.tsx
index de9c0791..448b62f5 100644
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -211,6 +211,8 @@ const SyncConnectedSiteControls = ( {
                                                remoteSite={ connectedSite }
                                                onSubmit={ ( tree ) => {
                                                        const optionsToSync = getCheckedTreeItems( tree );
+                                                       console.log( optionsToSync );
+                                                       return;
 
                                                        if ( syncDialogType === 'push' ) {
                                                                void pushSite( connectedSite, selectedSite, { optionsToSync } );
@@ -218,7 +220,7 @@ const SyncConnectedSiteControls = ( {
                                                                pullSite( connectedSite, selectedSite );
                                                        }
                                                } }
-                                               onRequestClose={ () => setSyncDialogType( null ) }
+                                               onRequestClose={() => { } }
                                        />
                                ) }
                        </div>
```
2. Open Sync tab for any site
4. Click `Push`
5. Open Browser's console
6. Select/deselect some options from `What would you like to push?` tree
7. Click `Push`
8. Assert that everything works well:
	1. By defualt you see  `['all']`
	2. If you deselect something then you will see array with the remaining selected items

### DIFF
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -15,6 +15,7 @@ import ProgressBar from 'src/components/progress-bar';
 import { SyncDialog } from 'src/components/sync-dialog';
 import { SyncPullPushClear } from 'src/components/sync-pull-push-clear';
 import { Tooltip, DynamicTooltip } from 'src/components/tooltip';
+import { getCheckedTreeItems } from 'src/components/tree-view';
 import { WordPressLogoCircle } from 'src/components/wordpress-logo-circle';
 import { useSyncSites } from 'src/hooks/sync-sites';
 import { useConfirmationDialog } from 'src/hooks/use-confirmation-dialog';
@@ -208,9 +209,11 @@ const SyncConnectedSiteControls = ( {
 						type={ syncDialogType }
 						localSite={ selectedSite }
 						remoteSite={ connectedSite }
-						onSubmit={ () => {
+						onSubmit={ ( tree ) => {
+							const optionsToSync = getCheckedTreeItems( tree );
+
 							if ( syncDialogType === 'push' ) {
-								void pushSite( connectedSite, selectedSite );
+								void pushSite( connectedSite, selectedSite, { optionsToSync } );
 							} else {
 								pullSite( connectedSite, selectedSite );
 							}
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -26,6 +26,7 @@ import { cx } from 'src/lib/cx';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { getLocalizedLink } from 'src/lib/get-localized-link';
 import { useI18nLocale } from 'src/stores';
+import type { SyncPart } from 'src/hooks/sync-sites/use-sync-push';
 import type { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
 
 interface ConnectedSiteSection {
@@ -209,7 +210,7 @@ const SyncConnectedSiteControls = ( {
 						localSite={ selectedSite }
 						remoteSite={ connectedSite }
 						onSubmit={ ( tree ) => {
-							const optionsToSync = [];
+							const optionsToSync: SyncPart[] = [];
 
 							const isAll = tree.every( ( node ) => node.checked );
 							if ( isAll ) {
@@ -228,7 +229,7 @@ const SyncConnectedSiteControls = ( {
 
 								wpContent.forEach( ( item ) => {
 									if ( item.checked ) {
-										optionsToSync.push( item.id );
+										optionsToSync.push( item.id as SyncPart );
 									}
 								} );
 							}
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -17,6 +17,7 @@ import { SyncPullPushClear } from 'src/components/sync-pull-push-clear';
 import { Tooltip, DynamicTooltip } from 'src/components/tooltip';
 import { WordPressLogoCircle } from 'src/components/wordpress-logo-circle';
 import { useSyncSites } from 'src/hooks/sync-sites';
+import { SyncPart, isSyncPart } from 'src/hooks/sync-sites/sync-part';
 import { useConfirmationDialog } from 'src/hooks/use-confirmation-dialog';
 import { useFeatureFlags } from 'src/hooks/use-feature-flags';
 import { useImportExport } from 'src/hooks/use-import-export';
@@ -27,7 +28,6 @@ import { getIpcApi } from 'src/lib/get-ipc-api';
 import { getLocalizedLink } from 'src/lib/get-localized-link';
 import { useI18nLocale } from 'src/stores';
 import type { TreeNode } from 'src/components/tree-view';
-import { SyncPart, isSyncPart } from 'src/hooks/sync-sites/sync-part';
 import type { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
 
 interface ConnectedSiteSection {
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -15,7 +15,6 @@ import ProgressBar from 'src/components/progress-bar';
 import { SyncDialog } from 'src/components/sync-dialog';
 import { SyncPullPushClear } from 'src/components/sync-pull-push-clear';
 import { Tooltip, DynamicTooltip } from 'src/components/tooltip';
-import { getCheckedTreeItems } from 'src/components/tree-view';
 import { WordPressLogoCircle } from 'src/components/wordpress-logo-circle';
 import { useSyncSites } from 'src/hooks/sync-sites';
 import { useConfirmationDialog } from 'src/hooks/use-confirmation-dialog';
@@ -210,7 +209,29 @@ const SyncConnectedSiteControls = ( {
 						localSite={ selectedSite }
 						remoteSite={ connectedSite }
 						onSubmit={ ( tree ) => {
-							const optionsToSync = getCheckedTreeItems( tree );
+							const optionsToSync = [];
+
+							const isAll = tree.every( ( node ) => node.checked );
+							if ( isAll ) {
+								optionsToSync.push( 'all' );
+							} else {
+								const isDatabaseSelected = tree.find( ( node ) => node.id === 'sqls' )?.checked;
+
+								if ( isDatabaseSelected ) {
+									optionsToSync.push( 'sqls' );
+								}
+
+								const filesAndFolders =
+									tree.find( ( node ) => node.id === 'filesAndFolders' )?.children || [];
+								const wpContent =
+									filesAndFolders.find( ( node ) => node.id === 'wp-content' )?.children || [];
+
+								wpContent.forEach( ( item ) => {
+									if ( item.checked ) {
+										optionsToSync.push( item.id );
+									}
+								} );
+							}
 
 							if ( syncDialogType === 'push' ) {
 								void pushSite( connectedSite, selectedSite, { optionsToSync } );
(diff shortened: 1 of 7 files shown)


src/components/sync-dialog.tsx
```diff
@@ -132,7 +132,7 @@ export function SyncDialog( {
 	const [ showAllFiles, setShowAllFiles ] = useState( false );
 	const [ treeState, setTreeState ] = useState< TreeNode[] >( [
 		{
-			id: 'filesAndFolders',
+			id: 'all',
```
sejas: I think this should be `contents` instead of `all`. It seems that `all` will also restore SQLs, while `contents` will restore only the wp-content files.
Alternatively, we may need to build a more hardcoded system instead of generating the fields dynamically.

sejas (commented):
Great start! Thanks for adding the options to the request to the backend 💪.  
I think we need to build a less dynamic way to send the options.  
For example, if both files and database are checked, we can send `all`.  
If files are checked but not database, then we'll send `contents`.  
If only one plugin is checked, then we'll send `plugins`.  
If one plugin and the database are selected, then we'll send `plugins,sqls`.  

Let's also add tests so we can validate each state.

wojtekn:
> For example, if both files and database are checked, we can send `all`.
> If files are checked but not database, then we'll send `contents`.

@sejas it doesn't seem to be correct. `contents` doesn't include plugins, themes or uploads. 

> If only one plugin is checked, then we'll send `plugins`.
> If one plugin and the database are selected, then we'll send `plugins,sqls`.

sejas:
Right! Thanks for clarifying. I misinterpreted the key name. All the keys are well explained in Staging Sync UI ([code](https://github.com/Automattic/wp-calypso/blob/a5f7a7c42184d8928602eadc48ca8ec6a3ae4bcb/client/sites/staging-site/components/staging-site-card/card-content/staging-sync-card.tsx#L175-L226)):



> If files are checked but not database:

So, for this case we'll need to send the "children" values, `plugins,themes,uploads,contents`

nightnei:
@sejas I am looking at staging site selectors and Selective Sync selectors and I think that we should reconsider the tree for Selective Sync. At least, we should, probably, put `wp-content` alongside plugins, etc, similarly to Staging sites. Since we just will confuse ourselves, instead of keeping consistency and the same behaviour. I don't see any point in coming up with different approaches/configurations.

So I propose:
Screenshot 2025-06-16 at 13 33 36

sejas:
That is interesting. It simplifies the logic, removes the duplication of files and folders, and wp-content having the same purpose, and is consistent with the current implementation of Staging Sync, which WPcom users might be familiar with. It's worth mentioning that we are redesigning that flow too. (Figma ASqz3fW9Go9m5bHL2XiuUR-fi-1_49864)



My concern is whether it provides a good user interface or if we are exposing our API limits. Either way, we will still need to display the plugins and theme child folders, and make both syncs consistent (Studio Selective Sync, and Staging revamp).<br>I'm curious about what @wojtekn and @marina.matijaca1 think about replacing a parent `wp-content` with a sibling `wp-content (Apart from themes, plugins, and uploads)`.

nightnei:
> I'm curious about what @wojtekn and @marina.matijaca1 think about replacing a parent wp-content with a sibling wp-content (Apart from themes, plugins, and uploads).

Just wanted to add, that we can improve UI here, but the main goal is to keep it simple and the same structure for "Studio Selective Sync" and "dotcom Staging Sites".

Also wanted to highlight important thing, that the current tree with `wp-content` as parent for themes/plugins - **disallows** to select wp-content w/o themes/plugins/uploads. I mean - it's not posible to keep `wp-content` selected, but all children deselected. Even if we implement such behaviour, it will be very confusing from UX standpoint, since it's not obvious what does it mean when `wp-content` is selected but all children are deselected. So IMO it's a good way to go with this structure:
Screenshot 2025-06-16 at 15 00 37

bcotrim:
> Right! Thanks for clarifying. I misinterpreted the key name. All the keys are well explained in Staging Sync UI ([code](https://github.com/Automattic/wp-calypso/blob/a5f7a7c42184d8928602eadc48ca8ec6a3ae4bcb/client/sites/staging-site/components/staging-site-card/card-content/staging-sync-card.tsx#L175-L226)):
> 
> 
> 
> > If files are checked but not database:
> 
> So, for this case we'll need to send the "children" values, `plugins,themes,uploads,contents`

Thanks for sharing @sejas 
I added a quick PR to add those details to our API so it's easier to understand: 185599-ghe-Automattic/wpcom

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -27,9 +27,17 @@ export type SyncPushState = {
 	remoteSiteUrl: string;
 };
 
+type PushSiteOptions = {
+	optionsToSync?: string[];
```
bcotrim: Let's improve the this type from `string` to the known values.

sejas:
> Also wanted to highlight important thing, that the current tree with `wp-content` as parent for themes/plugins - **disallows** to select wp-content w/o themes/plugins/uploads. I mean - it's not posible to keep `wp-content` selected, but all children deselected. Even if we implement such behaviour, it will be very confusing from UX standpoint, since it's not obvious what does it mean when `wp-content` is selected but all children are deselected. So IMO it's a good way to go with this structure:

Yep, supporting content needs specific options, otherwise, it can be confusing for users. I asked on Slack: p1750088098010279-slack-C06DRMD6VPZ

sejas:
@nightnei, let's keep wp-content parent folder, and add a new option for `contents`. The parent wp-content won't actually carry any state information, only it's children.

nightnei:
> @nightnei, let's keep wp-content parent folder, and add a new option for contents. The parent wp-content won't actually carry any state information, only it's children.

I already finished implementing the approach w/o wp-content parent folder 😅 Will update it now 🙂

UPDATE: Waiting for the answer here p1750091580053289/1750088098.010279-slack-C06DRMD6VPZ

wojtekn:
Thanks @sejas and @nightnei for exploring different options. The current state, which mimics the real directory structure, looks clean. 👍

src/components/sync-dialog.tsx
```diff
@@ -168,12 +164,18 @@ export function SyncDialog( {
 							checked: true,
 							type: 'folder',
 						},
+						{
+							id: 'contents',
+							label: __( 'other files and directories' ),
```
wojtekn: As this isn't the directory name, but a custom label, we could start with a capital letter like we do with 'Files and folders' or 'Database'.

src/components/sync-connected-sites.tsx
```diff
@@ -208,9 +209,33 @@ const SyncConnectedSiteControls = ( {
 						type={ syncDialogType }
 						localSite={ selectedSite }
 						remoteSite={ connectedSite }
-						onSubmit={ () => {
+						onSubmit={ ( tree ) => {
```
bcotrim, line 239: Could we extract this as a separate function? I think it would make it easier to read and understand. Maybe calling it `getSyncOptionsFromTree` or something similar.
What do you think?

bcotrim (changes requested):
Code LGTM 👍 

Following up on @sejas comment, I also think we should add some unit tests to validate the tree -> request workflow.

src/components/sync-connected-sites.tsx
```diff
@@ -208,9 +209,33 @@ const SyncConnectedSiteControls = ( {
...
+								}
+
+								const filesAndFolders =
+									tree.find( ( node ) => node.id === 'filesAndFolders' )?.children || [];
+								const wpContent =
+									filesAndFolders.find( ( node ) => node.id === 'wp-content' )?.children || [];
+
+								wpContent.forEach( ( item ) => {
+									if ( item.checked ) {
+										optionsToSync.push( item.id as SyncPart );
```
sejas: I'm not fan of overriding types. Is there a way we can avoid forcing this `SyncPart` type?

src/components/sync-dialog.tsx
```diff
@@ -1,6 +1,5 @@
 import { SelectControl } from '@wordpress/components';
 import { createInterpolateElement } from '@wordpress/element';
-import { __ } from '@wordpress/i18n';
```
sejas, line 3: ❤️

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -225,7 +235,15 @@ export function useSyncPush( {
...
 				type: 'application/gzip',
 			} );
-			const formData = [ [ 'import', file ] ];
+
+			const formData = [];
+
+			formData.push( [ 'import', file ] );
+
+			if ( options?.optionsToSync ) {
+				formData.push( [ 'options', options.optionsToSync ] );
```
sejas: I think sending an array won't work. The backend will override the latest value. We'll need to join the elements with `,`.

sejas (commented):
It works as expected; thanks for the changes. I tested it holistically, and it worked with all my combinations. I left a couple of comments. We need to send the options joined by a comma. ✅ `'sqls,themes'` instead of ❌ `['sqls', 'themes']`
Let's also add test cases to validate that the tree is converted correctly into the expected options.

**All**

Screenshot 2025-06-17 at 10 34 22

Screenshot 2025-06-17 at 10 34 28

**Database**

sqls
sqls-options

**Database + themes**

sqls-themes
sqls-themes-options

**Only files**

files
files-options

src/components/sync-connected-sites.tsx
sejas: Great idea, let's isolate that function.

src/components/sync-connected-sites.tsx
nightnei: > I'm not fan of overriding types. Is there a way we can avoid forcing this SyncPart type?

I am also not a fan, and I would revert `optionsToSync` to `string`, it's one option.
The second option is to define `id`s with type SyncPart, BUT we have `id` as `filesAndFolders`, which actually is not allowed to send to backend, so I see two options:
1. use type assertion
2. return `optionsToSync ` to `string` type

Do you have otehr ideas? 
@sejas @bcotrim

sejas: Let's not use `string` type.<br>I would love if we can define a more structured/fixed type using the SyncPart for wpContent keys.

nightnei: > I would love if we can define a more structured/fixed type using the SyncPart for wpContent keys.

And what about `id` as `filesAndFolders` which is forbidden to send?

sejas: `filesAndFolders` is the parent and not the child of wp-content.

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -225,7 +234,15 @@ export function useSyncPush( {
...
 				type: 'application/gzip',
 			} );
-			const formData = [ [ 'import', file ] ];
+
+			const formData = [];
+
+			formData.push( [ 'import', file ] );
+
+			if ( options?.optionsToSync ) {
+				formData.push( [ 'options', options.optionsToSync.join( ',' ) ] );
```
sejas, line 243: 👍

src/hooks/sync-sites/sync-part.ts
```diff
@@ -0,0 +1,7 @@
+const SYNC_PARTS = [ 'all', 'themes', 'plugins', 'uploads', 'sqls', 'contents' ] as const;
+
+export type SyncPart = ( typeof SYNC_PARTS )[ number ];
```
sejas: @nightnei, Can we rename `SyncPart` with `SyncOption` ? , and also the other constants, functions and filename.

sejas (approved):
Thank you for the changes, including the addition of the type check `isSyncPart` and the tests.
I suggested renaming those variables to `SyncOption`. Please let me know what you think.

I confirm that the previous push works as expected and does not include any options. The new selective sync feature send the options in the format that the backend expects. Example: `themes, plugins, sqls`.

src/hooks/sync-sites/sync-part.ts
nightnei: Good idea 👍 Renamed

bcotrim (approved):
Thanks for the adjustments and unit tests.
Good job! LGTM 👍

PR #1483, opened 2025-06-15
Add feature flags menu

bcotrim:
## Related issues

- N/A

## Proposed Changes

- Add a feature flags menu to make it easier to toggle feature flags

## Testing Instructions

- `yarn start`
- At the time of this PR the only available feature flag is "Selective Sync" so confirm that new feature is not available
- Open "Electron" menu
- Confirm the Feature Flags group shows with the option "Selective Sync" and toggle it
- Confirm the new sync feature is not available.

- Start the project with the env var flag to confirm the existing behavior still works. 
For example`STUDIO_SELECTIVE_SYNC=true npm start`

image

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-feature-flags.tsx
+++ b/src/hooks/use-feature-flags.tsx
@@ -1,23 +1,24 @@
 import * as Sentry from '@sentry/react';
 import React, { createContext, useContext, ReactNode, useState, useEffect } from 'react';
-import { z } from 'zod';
 import { useAuth } from 'src/hooks/use-auth';
+import { useIpcListener } from 'src/hooks/use-ipc-listener';
 import { getAppGlobals } from 'src/lib/app-globals';
+import { FEATURE_FLAGS, FeatureFlags } from 'src/lib/feature-flags';
+import { getIpcApi } from 'src/lib/get-ipc-api';
 
-// In PHP, empty associative arrays are encoded as regular arrays when converted to JSON.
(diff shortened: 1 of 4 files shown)


epeicher (approved):
Thanks Bernardo for enhancing this! This will be handy for development. I have tested it, and it works as advertised. I can enable the feature flags without stopping and starting Studio in development. I have also tested, and the menu item does not appear when building the production version. LGTM! :shipit: 

|Enabling Selective Sync feature flag | Disabling Selective Sync feature flag|Starting Studio with STUDIO_SELECTIVE_SYNC=true|
|-------|------|------|
| CleanShot 2025-06-16 at 16 02 21@2x|CleanShot 2025-06-16 at 16 02 37@2x|CleanShot 2025-06-16 at 16 03 41@2x|

gcsecsey (approved):
Thanks @bcotrim for adding this menu, this is a very nice improvement!

The changes LGTM and I can also verify that I can instantly toggle the feature flag using the menu, no restart/refresh needed. I also found no regressions. 👍

wojtekn:
Thanks for the addition @bcotrim  - it works great. I tested if we can still run the production build with flags and it worked fine:

```
STUDIO_SELECTIVE_SYNC=true open /Applications/Studio.app
```

We could consider enabling the menu also when `IS_DEV_BUILD`, so dev builds would display it, too. However, we would still need to run the above command during CfT, as we are using production builds there.

PR #1484, opened 2025-06-17
Read the real site folders to display a granular selection

bcotrim:
## Related issues

- Fixes STU-570

## Proposed Changes

- Sync dialog now reads and displays the real file structure for plugins and themes.
- Only for the `push` type 
- Added loading state on `TreeView` to correctly display spinner while node children are loading

## Testing Instructions

- Navigate to the Sync tab, connect a site, and press "Push"
- Confirm the real list of plugins and themes are displayed
- Close the dialog
- Press "Pull"
- Confirm there are not displayed list of plugins and themes

Note: At this time I didn't make any changes so the API call. I can add to this PR or as a follow-up. My initial goal is to discuss and decide on the approach to fetch and display the real list of plugins/themes.

| Push | Pull |
|--------|--------|
| image | image | 

## Pre-merge Checklist


src/components/sync-dialog.tsx
```diff
@@ -179,6 +188,66 @@ export function SyncDialog( {
 		},
 	] );
 
+	// Update the plugins tree state when the plugins are loaded
+	useEffect( () => {
+		if ( type === 'pull' ) {
+			return;
+		}
+
+		setTreeState( ( prev ) => {
```
nightnei: We can simplify and make it more readable with `updateNodeById`

src/components/sync-dialog.tsx
```diff
@@ -151,16 +162,14 @@ export function SyncDialog( {
 							label: 'plugins',
 							checked: true,
 							type: 'folder',
-							children: [
-								{ id: 'akismet', label: 'akismet', checked: true, type: 'folder' },
-								{ id: 'jetpack', label: 'jetpack', checked: true, type: 'folder' },
-							],
+							children: undefined,
```
nightnei: `undefined` doesn't give any benefits and it's not common approach to keep it explicitly, let's just remove it
```suggestion
```

src/components/sync-dialog.tsx
```diff
@@ -128,6 +129,16 @@ export function SyncDialog( {
 }: SyncDialogProps ) {
 	const locale = useI18nLocale();
 	const copy = useCopy( type );
+	const {
+		items: plugins,
+		isLoading: isLoadingPlugins,
+		error: pluginsError,
+	} = useWpList( localSite.id, 'plugin' );
+	const {
```
nightnei: Duplication looks not nice, WDYT about providing array, e.g. `['plugins', 'theme']` and expecte them directly, instead of property `items`

bcotrim: Could you share a bit more details on how you would approach it, please?

src/components/sync-dialog/index.tsx
```diff
@@ -190,6 +40,66 @@ export function SyncDialog( {
...
+						type: 'folder',
+				  } ) );
+			const updated = updateNodeById( newState, SYNC_OPTIONS.themes, {
+				loading: isLoadingThemes,
+				children: themesChildren,
+			} );
+
+			return updated;
+		} );
+	}, [ type, themes, isLoadingThemes, themesError ] );
```
sejas: Let's combine, or create a generic hook to avoid duplicating the same code. I feel both useEffects are almost the same code.

src/components/sync-dialog/use-copy.tsx
```diff
@@ -0,0 +1,62 @@
+import { useI18n } from '@wordpress/react-i18n';
+import { useMemo } from 'react';
+
+export const useCopy = ( type: 'pull' | 'push' ) => {
```
sejas: Let's rename this hook to something like `useSyncTexts`. Now that we are exporting the hook, it may cause some confusion.

```suggestion
export const useSyncTexts = ( type: 'pull' | 'push' ) => {
```

src/components/sync-dialog/use-env-details.tsx
```diff
@@ -0,0 +1,39 @@
+import { useI18n } from '@wordpress/react-i18n';
+import type { EnvironmentType } from 'src/components/environment-badge';
+import type { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
+
+export const useEnvDetails = (
```
sejas: Let's move this hook to the general `src/hooks` as we can re-use it in other components.

src/hooks/use-wp-list.ts
```diff
@@ -0,0 +1,44 @@
...
+	const [ error, setError ] = useState< Error | null >( null );
+
+	useEffect( () => {
+		async function fetchPlugins() {
+			try {
+				const { stdout, stderr } = await getIpcApi().executeWPCLiInline( {
+					siteId,
+					args: wpListCommands[ type ],
+					skipPluginsAndThemes: true,
+				} );
```
sejas: I'm not sure about using `wp-cli` command here. I see it's practical and almost accurate, but I see a few inconsistencies. Probably we should use a `fs` function to read the directories and files.

* Other "phantom" files are listed as plugins like `db.php`
* Not all files appear in the list. Maybe some devs have "libraries" or `.php` files that are necessary for their sites. Not a good practice, but it's a possibility.
* Let's differentiate between folders and files. See the `other-plugin` in the screenshot.

Screenshot 2025-06-18 at 12 21 24

sejas (commented):
@bcotrim , it works pretty good! I like it.
Thanks for separating the useCopy, useEnvDetails, and other hooks 👍 , now it's cleaner.
Let's differentiate between plugin files and folders with a different icon.
My biggest concern is using wp-cli instead of reading the file system. What do you think?

sejas:
Let's also order the directories and files alphabetically.

src/hooks/use-wp-list.ts
wojtekn: Agreed, we should work on files and directories and not the WP CLI result here.

src/components/sync-dialog/use-copy.tsx
nightnei: If we are going to extract it to a separate file, then we should create `tree-view` folder and keep it inside of this folder, instead of `sync-dialog` (a level above).
But amyway, I would make these changes in a separate PR, to keep this PR simple and focused on specific changes. So let's revert these changes and make refactoring in a complex way, if we want to split tree-view component to a few files.

src/components/sync-dialog/use-env-details.tsx
nightnei: As I mentioned [here](https://github.com/Automattic/studio/pull/1484#discussion_r2154424096) - I propose to do such changes separately.
And also keep this hook inside `tree-view` folder, instead of globally, to avoid flooding of global hooks. If we need to reuse it somewhere in the future, it will be easy to move it to the global folder, if we indeed need it.

src/components/sync-dialog/index.tsx
nightnei: Also, we can consider creating a hook that returns themes and plugins with the help of 1 hook call. e.g. `const { plugin, themes } = useWpList( localSite.id, [ 'plugin', 'themes' ] );`

sejas: Yep, that's a possibility, but we would still need to merge it with setTreeState. Maybe we could pass that setState as a callback to the hook.

src/components/sync-dialog/use-copy.tsx
sejas: Ideally, we refactor the code in separate PRs, but it’s fine to clean the code as we progress. So, there’s no need to undo these changes. I’ll defer the decision to @bcotrim .

src/components/sync-dialog/index.tsx
nightnei: I mean:

```
//wpList is object as { themes, plugins }
const wpList = useWpList( localSite.id, [ 'plugin', 'themes' ] );

useEffect( () => {
		if ( type === 'pull' ) {
			return;
		}
		
		setTreeState( ( prev ) => {
			// set themes and plugins from wpList to treeState
		} );
}, [ type, wpList, isLoadingWpList, wpListError ] );
```

src/components/sync-dialog/use-copy.tsx
bcotrim: I agree we could probably do a big code cleanup before completing the project.
To make things easier to review, I reverted the changes in https://github.com/Automattic/studio/pull/1484/commits/7dc7227152c17bbd1a3af6672e026c5397726be4

nightnei: Thank you, it's really much easier to review the PR now

nightnei: > If we are going to extract it to a separate file, then we should create tree-view folder and keep it inside of this folder, instead of sync-dialog (a level above).

@bcotrim BTW, regarding this ^^^ - my bad, I mixed up `tree-view` and `sync-dialog`. The changes were good 👍 But still, let's do it separatelly.

src/hooks/use-wp-list.ts
bcotrim: thanks for the feedback!
My worry was about knowing which files to show, and relying on `wp-cli` to decide what is a theme and a plugin.
I did a refactor to use `fs` and display all folders and `.php` files. I added this constraint to avoid showing files that don't make sense like `.DS_Store` for example.
https://github.com/Automattic/studio/pull/1484/commits/ca9cd62029d9ebe434cceb9985b67a0067aa2fe6

bcotrim:
> Let's also order the directories and files alphabetically.

I sorted by type (folders first) and alphabetically, please let me know if this sounds good.

src/components/sync-dialog/index.tsx
bcotrim: I have refactored the logic in https://github.com/Automattic/studio/pull/1484/commits/e30f52f9c1a7ae03acc4b4692800ee2a1938108e, can you take another look, please?

src/components/sync-dialog/use-copy.tsx
sejas: I created a issue to track it: STU-580

src/components/sync-dialog.tsx
```diff
@@ -176,6 +177,50 @@ export const useDefaultTree = (): TreeNode[] => {
...
+				newState = updateNodeById( newState, syncOptions[ wpType ], {
+					loading: isLoading,
+					children,
+				} );
+			} );
+
+			return newState;
+		} );
+	}, [ type, wpListResults, setTreeState ] );
+};
```
sejas: `useWpList` is only used in `useDynamicTreeState`. I wonder if it makes sense to have both hooks defined in the same file and exporting only one of them. Not a strong opinion though.

src/components/sync-dialog.tsx
```diff
@@ -176,6 +177,50 @@ export const useDefaultTree = (): TreeNode[] => {
...
+			return;
+		}
+
+		setTreeState( ( prev ) => {
+			let newState = [ ...prev ];
+
+			const syncOptions = {
+				plugins: SYNC_OPTIONS.plugins,
+				themes: SYNC_OPTIONS.themes,
+			} as const;
```
sejas: I think we can remove the options object.

```suggestion
```

src/components/sync-dialog.tsx
```diff
@@ -176,6 +177,50 @@ export const useDefaultTree = (): TreeNode[] => {
...
+				const children: TreeNode[] | undefined = error
+					? undefined
+					: items.map( ( item ) => ( {
+							id: item.name,
+							label: item.name,
+							checked: true,
+							type: item.type,
+					  } ) );
+
+				newState = updateNodeById( newState, syncOptions[ wpType ], {
```
sejas: And use `SYNC_OPTIONS` directly.
```suggestion
				newState = updateNodeById( newState, SYNC_OPTIONS[ wpType ], {
```

src/components/sync-dialog.tsx
```diff
@@ -176,6 +177,50 @@ export const useDefaultTree = (): TreeNode[] => {
...
 };
 
+const WP_LIST_ITEMS: WpListType[] = [ 'plugins', 'themes' ];
+
+const useDynamicTreeState = (
+	type: 'push' | 'pull',
+	localSiteId: string,
+	setTreeState: React.Dispatch< React.SetStateAction< TreeNode[] > >
+) => {
+	const wpListResults = useWpList( localSiteId, WP_LIST_ITEMS );
```
sejas: Can we move the list of folders constant inside the hook?
WP_LIST_ITEMS or WpListType can be confusing. I prefer naming them "folder" or "subdir," similar to the naming used in `listWpContentFolders`.

```suggestion
const useDynamicTreeState = (
	type: 'push' | 'pull',
	localSiteId: string,
	setTreeState: React.Dispatch< React.SetStateAction< TreeNode[] > >
) => {
	const foldersToRead: WpListType[] = [ 'plugins', 'themes' ];
	const wpListResults = useWpList( localSiteId, foldersToRead );
```

src/hooks/use-wp-list.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export function useWpList(
+	siteId: string,
+	types: WpListType[]
+): Record< WpListType, WpListResult > {
+	const [ results, setResults ] = useState< Record< WpListType, WpListResult > >( () => {
+		const initialResults = {
+			plugins: { items: [], isLoading: true, error: null },
+			themes: { items: [], isLoading: true, error: null },
+		} as Record< WpListType, WpListResult >;
```
sejas: Let's avoid casting types when possible:

```suggestion
		const initialResults: Record< WpListType, WpListResult > = {
			plugins: { items: [], isLoading: true, error: null },
			themes: { items: [], isLoading: true, error: null },
		};
```

src/preload.ts
```diff
@@ -118,6 +118,8 @@ const api: IpcApi = {
 	saveUserEditor: ( editor ) => ipcRendererInvoke( 'saveUserEditor', editor ),
 	handleNewSite: ( newSite ) => ipcRendererInvoke( 'handleNewSite', newSite ),
 	comparePaths: ( path1, path2 ) => ipcRendererInvoke( 'comparePaths', path1, path2 ),
+	listWpContentFolders: ( siteId, type ) =>
+		ipcRenderer.invoke( 'listWpContentFolders', siteId, type ),
```
sejas: Let's use the same parameter name used during implementation.
```suggestion
	listWpContentFolders: ( siteId, subdir ) =>
		ipcRenderer.invoke( 'listWpContentFolders', siteId, subdir ),
```

src/hooks/use-wp-list.ts
```diff
@@ -0,0 +1,60 @@
+import * as Sentry from '@sentry/electron/renderer';
+import { useState, useEffect } from 'react';
+import { getIpcApi } from 'src/lib/get-ipc-api';
+
+export type WpListType = 'plugins' | 'themes';
+export type WpListEntry = { name: string; type: 'file' | 'folder' };
+export type WpListResult = { items: WpListEntry[]; isLoading: boolean; error: Error | null };
+
+export function useWpList(
```
sejas: We can consider renaming this hook to something more descriptive. Maybe `useWpContentFolders` or just `useContentFolders`.

sejas (commented):
It works great!, I tested it by switching sites and opening the push sync dialog and I confirm I see all my files and folders in plugins and themes.

I noticed that we could use Gutenberg icons, I created a separate issue for that STU-581

I left a few suggestions. Feel free to apply them if they make sense to you.

sejas:
Can we add some tests in a follow-up PR?

src/components/sync-dialog.tsx
nightnei: useDynamicTreeState is very specific and bound to the component, so I would propose to keep it this way, close to the component definition. Actually, I don't like hooks which return nothing, so I would even remove it and would define useEffect directly inside the component, but no preference, Bernardo up to you.

`useWpList` is designed as common hook, so I would propose to leave it as is or put as:
```
sync-dialog(folder)
--- sync-dialog.tsx -> rename to  `index.tsx`
--- use-wp-list.tsx
```

src/ipc-handlers.ts
```diff
@@ -1376,3 +1376,30 @@ export async function handleNewSite( event: IpcMainInvokeEvent, newSite: NewSite
 export function comparePaths( event: IpcMainInvokeEvent, path1: string, path2: string ) {
 	return arePathsEqual( path1, path2 );
 }
+
+function isVisibleEntry( entry: { name: string; type: string } ) {
```
nightnei: I haven't checked how it works, but is it correct name? Looks weird that visibility of entry is checked with `folder || php file`.

Also, I am thinking - it's a small function and used only in one place, so maybe woudl be better to define it dirrectly inside `.filter()` and add comment to explain what and why we arew filtering?

nightnei: Looking more into it - looks like indeed we should not put this function to the main scope of `ipc-handlers.ts`, especially with such a generic name. Since this "visibility" is about only reading `wp content folders`.

src/hooks/use-wp-list.ts
```diff
@@ -0,0 +1,60 @@
...
+					Sentry.captureException( err );
+					const error = err instanceof Error ? err : new Error( `Failed to fetch ${ type }s` );
+					setResults( ( prev ) => ( {
+						...prev,
+						[ type ]: { ...prev[ type ], isLoading: false, error },
+					} ) );
+				}
+			} );
+
+			await Promise.allSettled( promises );
```
nightnei: Hm, do we really need to wrap it with async function and run `Promise.allSettled`? It seems it should work work with simple `forEach(async () => {....})` since we set state inside and do it separately
```suggestion
```

nightnei (changes requested):
In general looks great 👍
Left a few comments

sejas:
@bcotrim , let's address the feedback and try to merge this PR on Monday after v1.5.4 is released.

src/components/sync-dialog.tsx
bcotrim: I would suggest reviewing them as we do the final code cleanup in STU-580
Does that sound good?

src/hooks/use-content-folders.ts
```diff
@@ -0,0 +1,60 @@
...
+export type WpContentFolderEntry = { name: string; type: 'file' | 'folder' };
+export type WpContentFolderResult = {
+	items: WpContentFolderEntry[];
+	isLoading: boolean;
+	error: Error | null;
+};
+
+export function useContentFolders(
+	siteId: string,
+	types: WpContentFolder[]
```
sejas, line 15: I'm not sure if we need to pass types. I like the flexibility of the hook, but I think it's not necessary, and useContentFolders can always return the plugins and themes children. Let's reconsider if we need pas the types parameter or not on STU-580.

sejas (approved):
Great work! Thanks for the code refactors 👌 
Works as expected.

src/hooks/use-content-folders.ts
```diff
@@ -0,0 +1,60 @@
...
+	const [ results, setResults ] = useState< Record< WpContentFolder, WpContentFolderResult > >(
+		() => {
+			const initialResults: Record< WpContentFolder, WpContentFolderResult > = {
+				plugins: { items: [], isLoading: true, error: null },
+				themes: { items: [], isLoading: true, error: null },
+			};
+
+			return initialResults;
+		}
+	);
```
nightnei: Hm, coudl you please explain this way of defining default lvalue? Instead of classic way as:
```suggestion
const [ results, setResults ] = useState< Record< WpContentFolder, WpContentFolderResult > >( {
		plugins: { items: [], isLoading: true, error: null },
		themes: { items: [], isLoading: true, error: null },
	} );
```

nightnei (approved):
Looks great 👍 Left 1 last commenty

bcotrim:
@sejas I've changed the icon to use `wordpress/components` similar to https://github.com/Automattic/studio/pull/1496

PR #1485, opened 2025-06-18
Update the `make-pot` command to open the browser and select the POT file to simplify uploading localized strings

sejas:
## Related issues

- pfHvTO-yN-p2 

## Proposed Changes

- This PR adds a `make-pot.mjs` script to simplify the process of uploading `bundle-strings.pot` to [Studio GlotPress](https://translate.wordpress.com/projects/studio/). Logic described in [localization.md](https://github.com/Automattic/studio/blob/trunk/docs/localization.md#extract-and-import).
- Update documentation describing the new suggested way.

## Testing Instructions

- Run `npm run make-pot`
- Wait a few seconds
- Observe the browser correctly opens GlotPress import page so you can simply drag and drop the selected file.

https://github.com/user-attachments/assets/f735e67e-9be1-467d-a186-ed1f2c6bed15

**Output**

Screenshot 2025-06-18 at 18 21 07

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1489, opened 2025-06-19
Open browser as the last step to ensure it works with Choosy

wojtekn:
## Related issues

- Follow up to https://github.com/Automattic/studio/pull/1485

## Proposed Changes

I propose to open the browser as the last step to ensure it works with Choosy. In other cases, opening Finder immediately closes Choosy selector, so the browser is not opened.

## Testing Instructions

- Test if the browser and Finder are opening.

## Pre-merge Checklist

### DIFF
--- a/scripts/make-pot.mjs
+++ b/scripts/make-pot.mjs
@@ -38,10 +38,6 @@ const commands = [
 		description: `Revealing pot file in Finder: ${ POT_FILE }`,
 	},
 	{
-		command: 'sleep 0.5',
-		description: 'Waiting for the Finder to open',
-	},
-	{
 		command: `open "${ IMPORT_PAGE }"`,
 		description: `Opening the translation import page ${ IMPORT_PAGE } in the browser`,
 	}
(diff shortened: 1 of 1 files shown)


scripts/make-pot.mjs
```diff
@@ -34,17 +34,17 @@ const commands = [
...
-		command: `open "${ IMPORT_PAGE }"`,
-		description: `Opening the translation import page ${ IMPORT_PAGE } in the browser`,
+		command: `open -R "${ POT_FILE }"`,
+		description: `Revealing pot file in Finder: ${ POT_FILE }`,
 	},
 	{
 		command: 'sleep 0.5',
-		description: 'Waiting for the browser to open',
+		description: 'Waiting for the Finder to open',
 	},
```
sejas: Let's remove the sleep step as I think it's not necessary anymore. I originally added it to "force" the finder to open on top of the browser.

sejas (approved):
The change makes sense to make it compatible with Choosy.
Let's just remove the sleep step.

epeicher (approved):
This works as expected. LGTM! :shipit: 

One small thing to note: the Finder window can sometimes end up behind the browser instead of in the foreground. Not a blocker, just an observation 🤓

PR #1492, opened 2025-06-20
Remove backslashes when creating host entry as security measure

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/studio/security/code-scanning/11

## Proposed Changes

I propose to remove backslashes when creating host entry as security measure.

## Testing Instructions

- Check if a site with a custom domain can still be created.

## Pre-merge Checklist

### DIFF
--- a/src/lib/hosts-file.ts
+++ b/src/lib/hosts-file.ts
@@ -67,9 +67,13 @@ export const writeHostsFile = async ( content: string ): Promise< void > => {
  * Create a regular expression matching the hosts entry for a given domain:
  *
  * 	127.0.0.1 foo.wp.cloud # Port 8000
+ *
+ * 	Remove backslashes and escape dots as a security measure.
  */
-function createHostsEntryPattern( domain: string ): RegExp {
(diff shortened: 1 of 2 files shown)


src/lib/hosts-file.ts
```diff
@@ -4,6 +4,7 @@ import { platform, tmpdir } from 'os';
 import path from 'path';
 import { promisify } from 'util';
 import * as Sentry from '@sentry/electron/main';
+import escapeRegExp from 'lodash/escapeRegExp';
```
wojtekn, line 7: Studio uses Electron 35, which uses Chrome 134. There is `RegExp.escape()` added in Chrome 136 so we can move to this one when we upgrade Electron.

ivan-ottinger (approved):
The proposed changes look good. Site with a custom domain can still be created and works correctly. 👍🏼  I did not observe any regressions.

epeicher (approved):
I have tested it, and it works as expected. I can create a site with a custom domain with no issues. LGTM! :shipit: 

CleanShot 2025-06-20 at 18 18 28@2x

PR #1493, opened 2025-06-20
Add Jest Preview to simplify tests debugging

nightnei:
## Proposed Changes

While writing tests, sometimes we face a case when some DOM node is not found. And it's difficult to understand what's wrong, w/o looking at the real DOM tree in a browser.
So I propose to add [jest-preview](https://www.jest-preview.com/) to simplify the process.
If you are using another/better tool/approach - share it please in the comments to this PR.

## Testing Instructions
Make some mistake in tests, e.g.:
```
diff --git a/src/components/tests/content-tab-sync.test.tsx b/src/components/tests/content-tab-sync.test.tsx
index 7cbc2294..ebc01644 100644
--- a/src/components/tests/content-tab-sync.test.tsx
+++ b/src/components/tests/content-tab-sync.test.tsx
@@ -93,7 +93,7 @@ describe( 'ContentTabSync', () => {
                fireEvent.click( loginButton );
                expect( useAuth().authenticate ).toHaveBeenCalled();
 
-               const freeAccountButton = screen.getByRole( 'button', { name: /Create a free account/i } );
+               const freeAccountButton = screen.getByRole( 'button', { name: /Create a free account123123/i } );
                expect( freeAccountButton ).toBeInTheDocument();
 
                fireEvent.click( freeAccountButton );
```
1. Run `npm run jest-preview`
2. Run `npm t src/components/tests/content-tab-sync.test.tsx`
3. Assert that you see teh DOM state catched automatically at teh point whern error happend:<br />Screenshot 2025-06-20 at 13 06 53

### DIFF
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -3,6 +3,7 @@ import '@testing-library/jest-dom';
 // used by `@php-wasm/universal` and it's not available in the Jest environment.
 // eslint-disable-next-line import/no-unresolved
 import 'web-streams-polyfill/polyfill';
+import { jestPreviewConfigure } from 'jest-preview';
 import nock from 'nock';
 
 // Silence console.log for all tests
@@ -71,3 +72,5 @@ jest.mock( './src/hooks/use-ai-icon', () => ( {
 } ) );
 
 global.ResizeObserver = require( 'resize-observer-polyfill' );
+
+jestPreviewConfigure( { autoPreview: true } );
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -73,4 +73,6 @@ jest.mock( './src/hooks/use-ai-icon', () => ( {
 
 global.ResizeObserver = require( 'resize-observer-polyfill' );
 
-jestPreviewConfigure( { autoPreview: true } );
+if ( ! process.env.CI ) {
+	jestPreviewConfigure( { autoPreview: true } );
+}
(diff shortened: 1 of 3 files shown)


ivan-ottinger (approved):
The changes look good and the tool works well. 👍🏼 

Markup on 2025-06-20 at 17:24:43

> If you are using another/better tool/approach - share it please in the comments to this PR.

I haven't used any specific tool before so curious to hear from other as well.

jest-setup.ts
```diff
@@ -71,3 +72,5 @@ jest.mock( './src/hooks/use-ai-icon', () => ( {
 } ) );
 
 global.ResizeObserver = require( 'resize-observer-polyfill' );
+
+jestPreviewConfigure( { autoPreview: true } );
```
bcotrim: Should we avoid setting this on CI?
I'm not sure if there's a direct side effect on having this configuration, but I also don't see any reason why we would benefit enabling it in CI.

package.json
```diff
@@ -92,6 +93,7 @@
 		"fork-ts-checker-webpack-plugin": "^7.2.13",
 		"jest": "^29.7.0",
 		"jest-environment-jsdom": "^29.7.0",
+		"jest-preview": "^0.3.1",
```
bcotrim, line 96: Should we rename it to `test:preview` to match the other scripts format?

bcotrim (approved):
Added a couple comments for consideration, but overall I think this is a nice tool to have.
LGTM 👍

nightnei:
Thanks folks, for reviews. Comments are addressed.

PR #1494, opened 2025-06-20
Disable topbar log in button when offline

sejas:
## Related issues

- Fixes STU-582

## Proposed Changes

- Disable main login button when offline and add tooltip.

## Testing Instructions

- Run `npm start`
- Make sure you are logged out
- Disable your internet connection
- Observe the login button has a lower opacity
- Move the mouse hover
- Observe the tooltip that says "You're currently offline."

https://github.com/user-attachments/assets/fdae4476-f81b-4115-862b-39a830dae137

## Pre-merge Checklist

### DIFF
--- a/src/components/tests/topbar.test.tsx
+++ b/src/components/tests/topbar.test.tsx
@@ -10,6 +10,7 @@ jest.mock( 'src/hooks/use-auth' );
 jest.mock( 'src/hooks/use-offline' );
 
 const mockOpenURL = jest.fn();
+const mockAuthenticate = jest.fn();
 const toggleMinWindowWidth = jest.fn();
 jest.mock( 'src/lib/get-ipc-api', () => ( {
 	__esModule: true,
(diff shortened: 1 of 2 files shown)


bcotrim (approved):
Code LGTM and works as described.
Tested on different languages, including RTL.
👍

wojtekn (approved):
It works as expected. It now makes sense as we added another button to access Preferences.

katinthehatsite:
Should this button then be cleaned up under the account tab:

Screenshot 2025-06-23 at 10 38 56 AM

It does come up as disabled but it seems that the same setting is now present in two different places.

wojtekn:
> @katinthehatsite I think it should staIt does come up as disabled but it seems that the same setting is now present in two different places.

This duplication doesn't look harmful as in Settings the authenticated state changes to unauthenticated state when users logs out. In think we can leave it as is for now, and we could potentially simplify it in the future e.g. by making Log out action closing the Preferences dialog.

PR #1495, opened 2025-06-20
Sync: Implement the selective sync archive to prepare only the selected parts on UI

bcotrim:
## Related issues

- Fixes STU-560

## Proposed Changes

- Added selected options to the exporter

## Testing Instructions

- Comment this line: https://github.com/Automattic/studio/blob/trunk/src/hooks/sync-sites/use-sync-push.ts#L272
- Start a Studio site push
- In your terminal check for the message like `Backup created at: ...`
- Confirm the backup has the selected options in your test
- After your test disable the STUDIO_SELECTIVE_SYNC or stop and start the server with `npm start`
- Repeat the test and check the backup contains all files

## Pre-merge Checklist

### DIFF
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -37,8 +37,17 @@ interface ConnectedSiteSection {
 	connectedSites: SyncSite[];
 }
 
-export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOption[] => {
+export type SyncOptionsWithSelections = {
+	optionsToSync: SyncOption[];
+	specificSelections?: {
+		plugins?: string[];
+		themes?: string[];
+	};
+};
+
+export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOptionsWithSelections => {
 	const optionsToSync: SyncOption[] = [];
+	const specificSelections: { plugins?: string[]; themes?: string[] } = {};
 
 	const isAll = tree.every( ( node ) => node.checked );
 	if ( isAll ) {
@@ -54,13 +63,41 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOption[] => 
 		const wpContent = filesAndFolders.find( ( node ) => node.id === 'wp-content' )?.children || [];
 
 		wpContent.forEach( ( item ) => {
-			if ( item.checked && isSyncOption( item.id ) ) {
-				optionsToSync.push( item.id );
+			if ( isSyncOption( item.id ) ) {
+				const hasCheckedChildren = item.children?.some( ( child ) => child.checked );
+				const shouldInclude = item.checked || hasCheckedChildren;
+
+				if ( shouldInclude ) {
+					optionsToSync.push( item.id );
+
+					if ( item.id === SYNC_OPTIONS.plugins && item.children ) {
+						const selectedPlugins = item.children
+							.filter( ( plugin ) => plugin.checked )
+							.map( ( plugin ) => plugin.id );
+						if ( selectedPlugins.length > 0 && selectedPlugins.length < item.children.length ) {
+							specificSelections.plugins = selectedPlugins;
+						}
+					}
+
+					if ( item.id === SYNC_OPTIONS.themes && item.children ) {
+						const selectedThemes = item.children
+							.filter( ( theme ) => theme.checked )
+							.map( ( theme ) => theme.id );
+						if ( selectedThemes.length > 0 && selectedThemes.length < item.children.length ) {
+							specificSelections.themes = selectedThemes;
+						}
+					}
+				}
 			}
 		} );
 	}
 
-	return optionsToSync;
+	const hasSpecificSelections = specificSelections.plugins || specificSelections.themes;
+
+	return {
+		optionsToSync,
+		specificSelections: hasSpecificSelections ? specificSelections : undefined,
+	};
 };
 
 const SyncConnectedSiteControls = ( {
@@ -237,12 +274,14 @@ const SyncConnectedSiteControls = ( {
 						localSite={ selectedSite }
 						remoteSite={ connectedSite }
 						onSubmit={ ( tree ) => {
-							const optionsToSync = convertTreeToOptionsToSync( tree );
+							const syncOptions = convertTreeToOptionsToSync( tree );
 
 							if ( syncDialogType === 'push' ) {
-								void pushSite( connectedSite, selectedSite, { optionsToSync } );
+								void pushSite( connectedSite, selectedSite, syncOptions );
 							} else {
-								pullSite( connectedSite, selectedSite, { optionsToSync } );
+								pullSite( connectedSite, selectedSite, {
+									optionsToSync: syncOptions.optionsToSync,
+								} );
 							}
 						} }
 						onRequestClose={ () => setSyncDialogType( null ) }
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -47,7 +47,7 @@ export type SyncOptionsWithSelections = {
 
 export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOptionsWithSelections => {
 	const optionsToSync: SyncOption[] = [];
-	const specificSelections: { plugins?: string[]; themes?: string[] } = {};
+	let specificSelections: { plugins?: string[]; themes?: string[] } | undefined = undefined;
 
 	const isAll = tree.every( ( node ) => node.checked );
 	if ( isAll ) {
@@ -79,17 +79,18 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOptionsWithS
 					.filter( ( child ) => child.checked )
 					.map( ( child ) => child.id );
 				if ( selectedItems.length > 0 && selectedItems.length < item.children.length ) {
-					specificSelections[ item.id ] = selectedItems;
+					specificSelections = {
+						...specificSelections,
+						[ item.id ]: selectedItems,
+					};
 				}
 			}
 		} );
 	}
 
-	const hasSpecificSelections = specificSelections.plugins || specificSelections.themes;
-
 	return {
 		optionsToSync,
-		specificSelections: hasSpecificSelections ? specificSelections : undefined,
+		specificSelections,
 	};
 };
 
@@ -272,9 +273,7 @@ const SyncConnectedSiteControls = ( {
 							if ( syncDialogType === 'push' ) {
 								void pushSite( connectedSite, selectedSite, syncOptions );
 							} else {
-								pullSite( connectedSite, selectedSite, {
-									optionsToSync: syncOptions.optionsToSync,
-								} );
+								pullSite( connectedSite, selectedSite, syncOptions );
 							}
 						} }
 						onRequestClose={ () => setSyncDialogType( null ) }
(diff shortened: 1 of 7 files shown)


ivan-ottinger (approved):
The changes look good and work as described. 👌🏼🙂 

Exported archive:

| `Plugins` only selected | normal Push (not using selective sync) |
|--------|--------|
| Markup on 2025-06-23 at 11:32:45 | Markup on 2025-06-23 at 11:34:08 |

epeicher (approved):
Thanks Bernardo for progressing this! I have tested it, and it works as advertised. I can see my selected changes when enabling the feature flag, while I see all the archives synced in production. Changes also LGTM! :shipit: 

|Selective sync only selecting plugins and Others | Production version|
|-------|------|
| CleanShot 2025-06-23 at 12 39 21@2x| CleanShot 2025-06-23 at 12 39 50@2x|

sejas (approved):
Thanks for creating this PR. It's easy to read and address the selective sync for the global scopes.

As a follow-up, let's implement the granular selection:
> For example, if the user chooses to push only a plugin or a theme, the zip file will contain only the plugin or theme and the meta.json file.

Let's add tests when ready, maybe after the granular selection is complete.

I confirm every option worked as expected, with the caveat that "Other files and directories" are only including mu-plugins and fonts. I create this issue (STU-584) to discuss how we should handle the `contents` option.

**Screenshots:**



| all | contents | plugins-and-sqls |
|:----------------------------------------------------------------------------------------:|:--------------------------------------------------------------------------------------------:|:---------------------------------------------------------------------------------------------------------:|
| plugins | sqls | themes |
| uploads |                                                                                             |                                                                                                          |

bcotrim:
@nightnei @epeicher @ivan-ottinger I adjusted the PR to respect the granular selection.
Can you take another look please?

src/components/sync-connected-sites.tsx
```diff
@@ -54,13 +63,41 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOption[] =>
 		const wpContent = filesAndFolders.find( ( node ) => node.id === 'wp-content' )?.children || [];
 
 		wpContent.forEach( ( item ) => {
-			if ( item.checked && isSyncOption( item.id ) ) {
-				optionsToSync.push( item.id );
+			if ( isSyncOption( item.id ) ) {
+				const hasCheckedChildren = item.children?.some( ( child ) => child.checked );
+				const shouldInclude = item.checked || hasCheckedChildren;
+
+				if ( shouldInclude ) {
```
nightnei: If parent is checked, then it means that at least one child is checked, and if it's not true, then it means that the tree has some bug and works incorrectly.
So, it's better to keep code more simple and at the same time don't mute bugs with tree state.
Does it make sense or am I missing something?
```suggestion
				if ( item.checked ) {
```

src/components/sync-connected-sites.tsx
```diff
@@ -54,13 +63,41 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOption[] =>
...
+					if ( item.id === SYNC_OPTIONS.themes && item.children ) {
+						const selectedThemes = item.children
+							.filter( ( theme ) => theme.checked )
+							.map( ( theme ) => theme.id );
+						if ( selectedThemes.length > 0 && selectedThemes.length < item.children.length ) {
+							specificSelections.themes = selectedThemes;
+						}
+					}
+				}
 			}
```
nightnei: I think we can remove double-conditions, code-duplications and simplify it with:
```suggestion
			if ( ! isSyncOption( item.id ) ) {
				return;
			}

			if ( item.checked ) {
				optionsToSync.push( item.id );
				return;
			}

			const checkedChildren = item.children
				?.filter( ( child ) => child.checked )
				.map( ( plugin ) => plugin.id );
				
			if ( checkedChildren?.length ) {
				if ( item.id === SYNC_OPTIONS.plugins  ) {
					specificSelections.plugins = checkedChildren;
				} else if ( item.id === SYNC_OPTIONS.themes ) {
					specificSelections.themes = checkedChildren;
				}
			}
```

src/components/sync-connected-sites.tsx
```diff
@@ -54,13 +63,41 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOption[] =>
...
 		} );
 	}
 
-	return optionsToSync;
+	const hasSpecificSelections = specificSelections.plugins || specificSelections.themes;
+
+	return {
+		optionsToSync,
+		specificSelections: hasSpecificSelections ? specificSelections : undefined,
+	};
```
nightnei: WDYT about removign extra 2 conditions for `hasSpecificSelections` with defining undefined by default `let specificSelections: { plugins?: string[]; themes?: string[] } | undefined;` and then we can set children with parent object together, for example `specificSelections = { ...specificSelections, plugins: checkedChildren };`


```suggestion
	return {
		optionsToSync,
		specificSelections,
	};
```

src/components/sync-connected-sites.tsx
```diff
@@ -237,12 +274,14 @@ const SyncConnectedSiteControls = ( {
...
+							const syncOptions = convertTreeToOptionsToSync( tree );
 
 							if ( syncDialogType === 'push' ) {
-								void pushSite( connectedSite, selectedSite, { optionsToSync } );
+								void pushSite( connectedSite, selectedSite, syncOptions );
 							} else {
-								pullSite( connectedSite, selectedSite, { optionsToSync } );
+								pullSite( connectedSite, selectedSite, {
+									optionsToSync: syncOptions.optionsToSync,
+								} );
```
nightnei: Not critical, but for consistency, I would keep it this way. WDYT?
```suggestion
								pullSite( connectedSite, selectedSite, syncOptions );
```

src/preload.ts
```diff
@@ -22,7 +22,8 @@ function ipcRendererSend< T extends keyof IpcHandlers >(
 
 const api: IpcApi = {
 	archiveSite: ( id, format ) => ipcRendererInvoke( 'archiveSite', id, format ),
-	exportSiteToPush: ( id ) => ipcRendererInvoke( 'exportSiteToPush', id ),
+	exportSiteToPush: ( id, optionsToSync?, specificSelections? ) =>
```
nightnei: It's not a good practice to have optional parameters one by one, since it disallows providing the third argument w/o providing the second one. In our PHP codebase I saw a lot of such sad cases e.g. `fucntion(title, null, description)`. 
So I propose to combine them and wrap with parent object:
	
```suggestion
	exportSiteToPush: ( id, configuration ) =>
```

src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -160,20 +160,54 @@ export class DefaultExporter extends EventEmitter implements Exporter {
...
+				( ( category === 'plugins' && this.options.specificSelections.plugins ) ||
+					( category === 'themes' && this.options.specificSelections.themes ) );
+
+			if ( hasSpecificSelections ) {
+				const selectedItems =
+					category === 'plugins'
+						? this.options.specificSelections!.plugins!
+						: this.options.specificSelections!.themes!;
+
+				for ( const itemName of selectedItems ) {
```
nightnei: `!` is not a good practice and better to try to avoid using it, since we lie to TS that it's not an optional field.
But, actually we can remove it and duplicatrion of the condition and simplify the code at the same time with:
```suggestion
			const partialFolderItems =
				this.options.specificSelections &&
				( ( category === 'plugins' && this.options.specificSelections.plugins ) ||
					( category === 'themes' && this.options.specificSelections.themes ) );

			if ( partialFolderItems ) {
				for ( const itemName of partialFolderItems ) {
```

nightnei (changes requested):
In general looks great 👍 Left a few comments

src/components/sync-connected-sites.tsx
bcotrim: I don't think that is correct. The parent is `checked` only if all children are selected.
This change handles the scenario where we have some children checked but not all. In this case we still want to include `plugins` for example, if only of the plugins was selected.

src/components/sync-connected-sites.tsx
bcotrim: I've made some adjustments to the code to follow your suggestion.
Let me know what you think.

src/preload.ts
bcotrim: I've made the adjustment.
Just FYI, that approach is not always the best, or should at least be used with caution.
On the frontend, for example, always using objects as arguments can lead to unwanted re-rendering due to equality issues. This doesn't apply to this context, so your suggestion is good. 👍

src/components/sync-connected-sites.tsx
```diff
@@ -54,13 +63,35 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOption[] =>
...
 
 		wpContent.forEach( ( item ) => {
-			if ( item.checked && isSyncOption( item.id ) ) {
+			if ( ! isSyncOption( item.id ) ) {
+				return;
+			}
+
+			if ( item.checked || item.indeterminate ) {
 				optionsToSync.push( item.id );
 			}
```
nightnei, line 72: Hm, I still don't understand this point.
1. If, for example, `plugins` item is selected, it means that all children are selected and we need just: push the `id` to `optionsToSync`, `return` and go with this flow https://github.com/Automattic/studio/pull/1495/files#diff-00ad6cb3b7847f2a153b53c275d15f0d2991f5a336d2f057cbff1af5adc0141dR189

```suggestion
			if ( item.checked ) {
				optionsToSync.push( item.id );
				return;
			}
```

2. If, for example, `plugins` item is not selected, then we should check - if at least one plugin is selected and add selected plugins to `specificSelections`. And, we don't need to push id `plugins` to `optionsToSync`, since [here](https://github.com/Automattic/studio/pull/1495/files#diff-00ad6cb3b7847f2a153b53c275d15f0d2991f5a336d2f057cbff1af5adc0141dR169) we don't use it

nightnei (commented):
Thanks, everything looks great 👍
Left the last comment

src/components/sync-connected-sites.tsx
```diff
@@ -54,13 +63,35 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOption[] =>
...
+				const selectedItems = item.children
+					.filter( ( child ) => child.checked )
+					.map( ( child ) => child.id );
+				if ( selectedItems.length > 0 && selectedItems.length < item.children.length ) {
+					specificSelections = {
+						...specificSelections,
+						[ item.id ]: selectedItems,
+					};
+				}
+			}
```
nightnei, line 87: I have tested one more time and still can't understand why do we need those extra conditions:

1. If parent item(`plugins` | `themes`) is selected, then we don't need to check for children, it's redundant operation
2. If parent is not selected, then we shoudl check for any selected child.
```suggestion
			if ( item.checked ) {
				optionsToSync.push( item.id );
				return;
			}

			if (
				item.children &&
				( item.id === SYNC_OPTIONS.plugins || item.id === SYNC_OPTIONS.themes )
			) {
				const selectedItems = item.children
					.filter( ( child ) => child.checked )
					.map( ( child ) => child.id );
				if ( selectedItems.length > 0 ) {
					specificSelections = {
						...specificSelections,
						[ item.id ]: selectedItems,
					};
				}
			}
```

nightnei (changes requested):

src/components/sync-connected-sites.tsx
nightnei: We did huge investigation and:
1. `plugins` should be selected, even if we do granular sync, otherwise they will be ignored. So we should proceed with what we have now.
2. But we have a bug with selective sync and it exists even at production now - STU-587

So it's safe to merge this PR and proceed wioth this feature, but we should fix taht existing bug (or maybe agree that it's expected behaviour).

nightnei (approved):
Great job Bernardo 👍

PR #1505, opened 2025-06-24
Remove jest-preview

nightnei:
## Proposed Changes

Jest-preview (from [this PR](https://github.com/Automattic/studio/pull/1493)) introduced some dependencies that have security vulnerabilities, so we decided to revert it and later consider using another lib or approach.

## Testing Instructions
Visual review should suffice


bcotrim (approved):
LGTM 👍

PR #1506, opened 2025-06-24
Update SQLite plugin to 2.2.3 and enable new driver

wojtekn:
## Related issues

- Fixes STU-586

## Proposed Changes

I propose to update the SQLite plugin to 2.2.3 and enable the new driver.

The driver uses constants to enable the new driver, but due to Playground/Studio/WP CLI specifics, the WP CLI command doesn't have access to the constants defined through Playground code. Thus, I used a custom WP CLI parameter to enable the driver explicitly. We can deprecate that parameter when we enable the driver by default.

Release plan:
1. Merge `wp-cli-sqlite-command` PR and release new version of library. Any Studio installation will automatically upgrade to the latest version, and the latest version of the command will still work with Studio that uses the SQLite driver 2.1.17-alpha.1
2. Merge this PR and release a new Studio version. It will upgrade the SQLite driver to 2.2.3 for its sites and will start using AST for the sites.

## Testing Instructions

1. Pull the branch and run `npm install`
2. Clone the `wp-cli-sqlite-command` PR locally (https://github.com/Automattic/wp-cli-sqlite-command/pull/10)
3. Run `composer install --no-dev --optimize-autoloader --ignore-platform-reqs` in the `wp-cli-sqlite-command` directory
4. Replace `wp-files/sqlite-command` in the Studio code with a symlink to your local `wp-cli-sqlite-command` directory
5. Apply the patch to the Studio repo:
```
diff --git a/src/setup-wp-server-files.ts b/src/setup-wp-server-files.ts
index a81b7026..536c14a1 100644
--- a/src/setup-wp-server-files.ts
+++ b/src/setup-wp-server-files.ts
@@ -86,17 +86,11 @@ async function copyBundledWPCLI() {
 
 async function copyBundledSQLiteCommand() {
        const bundledSqliteCommandPath = path.join( getResourcesPath(), 'wp-files', 'sqlite-command' );
-       const bundledSqliteCommandVersion = await getSQLiteCommandVersion( bundledSqliteCommandPath );
-       if ( ! bundledSqliteCommandVersion ) {
-               return;
-       }
+       const bundledSqliteCommandVersion = '1.0.7';
        const installedSqliteCommandPath = getSqliteCommandPath();
        const isSqliteCommandInstalled = await fs.pathExists( installedSqliteCommandPath );
 
-       const installedSqliteCommandVersion = await getSQLiteCommandVersion( installedSqliteCommandPath );
-       const isBundledVersionNewer =
-               installedSqliteCommandVersion &&
-               semver.gt( bundledSqliteCommandVersion, installedSqliteCommandVersion );
+       const isBundledVersionNewer = true;
        if ( ! isSqliteCommandInstalled || isBundledVersionNewer ) {
                console.log( `Copying bundled SQLite command version ${ bundledSqliteCommandVersion }…` );
                await recursiveCopyDirectory( bundledSqliteCommandPath, installedSqliteCommandPath );
@@ -114,5 +108,5 @@ export async function updateWPServerFiles() {
        await updateLatestWordPressVersion();
        await updateLatestSqliteVersion();
        await updateLatestWPCliVersion();
-       await updateLatestSQLiteCommandVersion();
+       //await updateLatestSQLiteCommandVersion();
 }

```
6. Remove current command files:
```
~/Library/Application\ Support/Studio/server-files/sqlite-command 
```
5. Run `npm start`. It will copy the symlinked command files to directory under `server-files/`
7. Click on the Import / Export tab
8. Click on the "Export database"
9. Observe `.sql` file is created
10. Open the `.sql` file and validate that the dump is correct - dump which uses AST driver will have `wp_commentmeta` as the first dumped table. Confirm that field has types with length set, that `ENGINE=InnoDB` exists.

In steps 1-5, we are ensuring that Studio uses the development version of the WP CLI SQLite command. Those won't be needed when we merge the release `wp-cli-sqlite-command` PR and release a new version.

The test cases worth testing:
- Studio `trunk` with SQLite driver 2.1.17-alpha.1 and WP CLI SQLite v1.0.6 - current version, to confirm everything works.
- Studio `trunk` with SQLite driver 2.1.17-alpha.1 and WP CLI SQLite from `new-sqlite-driver` branch - to confirm current Studio installation will work when we release new WP CLI SQLite and when Studio automatically upgrades it.
- Studio `update/sqlite-plugin-to-2.2.0` branch with SQLite driver 2.2.3 and WP CLI SQLite from `new-sqlite-driver` branch - to confirm everything works fine with AST driver

## Pre-merge Checklist


src/constants.ts
```diff
@@ -58,9 +58,9 @@ export const WP_CLI_IMPORT_EXPORT_RESPONSE_TIMEOUT =
 	WP_CLI_IMPORT_EXPORT_RESPONSE_TIMEOUT_IN_HRS * 60 * 60 * 1000; // 6hr
 
 // SQLite
-export const SQLITE_DATABASE_INTEGRATION_VERSION = 'v2.1.17-alpha.1';
+export const SQLITE_DATABASE_INTEGRATION_VERSION = 'v2.2.0';
```
JanJakes: @wojtekn I think it's better to use `v2.2.2` now, there are a few changes, especially `ADD INDEX` & `DROP INDEX` statements.

wojtekn: Good point, fixed.

wojtekn:
I tested importing, exporting, and sync for combinations:
- `trunk` with SQLite driver 2.1.17-alpha.1 and wp-cli-sqlite v1.0.6 (current version)
- `trunk` with SQLite driver 2.1.17-alpha.1 and wp-cli-sqlite from `new-sqlite-driver` branch
- `update/sqlite-plugin-to-2.2.0` with SQLite driver 2.2.2 and wp-cli-sqlite from `new-sqlite-driver` branch

The pull action for the last one still fails, but it's being addressed under https://github.com/WordPress/sqlite-database-integration/pull/203. If I manually apply this change to the driver on my site before the pull, the issue with undefined DB_NAME is resolved. However, another problem, `Call to a member function get_charset_collate() on null`, arises. I'm checking it with @JanJakes .

vendor/wp-now/src/wp-now.ts
```diff
@@ -380,7 +380,9 @@ async function initWordPress( php: PHP, wordPressVersion: string, vfsDocumentRoo
 		initializeDefaultDatabase = true;
 	}
 
-	const wpConfigConsts = {};
+	const wpConfigConsts = {
+		WP_SQLITE_AST_DRIVER: true,
```
wojtekn, line 384: We set the constant to enable the new driver for the site's WP admin and frontend.

src/lib/import-export/export/export-database.ts
```diff
@@ -18,7 +18,7 @@ export async function exportDatabaseToFile(
 
 	// Execute the command to export directly to the temp file
 	const { stderr, exitCode } = await server.executeWpCliCommand(
-		`sqlite export ${ tempFileName } --require=/tmp/sqlite-command/command.php`,
+		`sqlite export ${ tempFileName } --require=/tmp/sqlite-command/command.php --enable-ast-driver`,
```
wojtekn, line 21: We set the parameter to enable the AST driver for CLI actions.

wojtekn:
Issues covered in previous comment are fixed now.

src/lib/import-export/export/export-database.ts
sejas: We probably won't add more commands, but I wonder if we should  `opt-in` to use the AST driver more broadly. Like enable it by default, or auto-append the parameter in the wp-cli-process, similarly to how we increase the timeout:
https://github.com/Automattic/studio/blob/48bf68bb7a235a94c84272979e1f587dca17a0b3/src/lib/wp-cli-process.ts#L67-L70

sejas (commented):
I followed the test instructions but I'm getting the `DB_NAME` error. @wojtekn , do you know if that's expected?

Node error:
```
Success: Exporting database...
<br />
<b>Fatal error</b>:  Uncaught Error: Undefined constant &quot;Automattic\WP_CLI\SQLite\DB_NAME&quot; in /tmp/sqlite-command/src/SQLiteDriverFactory.php:24
```

Screenshot 2025-07-07 at 23 48 14

src/lib/import-export/export/export-database.ts
wojtekn: We could, but we should remove this parameter as soon as the SQLite driver enables it by default. What do you think?

wojtekn:
Thanks for testing it @sejas .

> I followed the test instructions but I'm getting the DB_NAME error. @wojtekn , do you know if that's expected?

Could you open this site's SQLite driver in `wp-content/mu-plugins/sqlite-database-integration` and confirm that `readme.txt` shows `2.2.3` version and that `constants.php` file sets `DB_NAME` if it's not set?

src/constants.ts
```diff
@@ -58,9 +58,9 @@ export const WP_CLI_IMPORT_EXPORT_RESPONSE_TIMEOUT =
 	WP_CLI_IMPORT_EXPORT_RESPONSE_TIMEOUT_IN_HRS * 60 * 60 * 1000; // 6hr
 
 // SQLite
-export const SQLITE_DATABASE_INTEGRATION_VERSION = 'v2.1.17-alpha.1';
+export const SQLITE_DATABASE_INTEGRATION_VERSION = 'v2.2.2';
```
sejas: We can use the new version v2.2.3
```suggestion
export const SQLITE_DATABASE_INTEGRATION_VERSION = 'v2.2.3;
```

src/lib/import-export/export/export-database.ts
sejas: Then, let's leave it as it is. There is no risk that we'll introduce more SQLite commands and forget to add the parameter in the meantime. 👍

sejas:
> Thanks for testing it @sejas .
> 
> > I followed the test instructions but I'm getting the DB_NAME error. @wojtekn , do you know if that's expected?
> 
> Could you open this site's SQLite driver in `wp-content/mu-plugins/sqlite-database-integration` and confirm that `readme.txt` shows `2.2.3` version and that `constants.php` file sets `DB_NAME` if it's not set?

Actually I have the version `2.2.2`. I tried to reproduce the testing steps from scratch and I'm getting another error: `Database export failed: [31;1mError:[0m Parameter errors: unknown --enable-ast-driver parameter`. So I cannot verify the correct behaviour.

Also we'll need to test the push process. I got an error when importing a dump that used the SQLite 2.2.2 without the new parameter in the command.

sejas:
I think we need to add another test instruction at the beginning. We need to run `npm install` to download the new SQLite plugin.

wojtekn:
@sejas all the instructions were written for 2.2.3, but I haven't pushed the code change for that. I've pushed 2.2.3 now.

I've added an explicit `npm install` step.

sejas (commented):
@wojtekn , I was able to execute the testing steps and export the database without any "errors".
But I see the database dump is almost empty. It has only one line:
```
-- Dump completed on 2025-07-10T12:15:57+00:00
```
The file has only 46 Bytes size:
Screenshot 2025-07-10 at 13 18 56

sejas:
After some more testing, the empty dump only happens in one of my sites. I confirmed the site has the 2.2.3 plugin, and removed other plugins like Gutenberg. The site starts correctly in Studio and I can access wp-admin. I'll test the site with the old driver.

The rest of sites export the database correctly and I was able to pull and push to production. 👍

sejas:
On trunk, and after replacing the sqlite plugin to 2.1.17-alpha, the database is exported correctly.

JanJakes:
@sejas Is there a way for me to reproduce that? Perhaps a DB dump? (We can also share it on Slack.)

sejas:
I'll share the database. Let me know if you need the whole site.

JanJakes:
@sejas  So I checked this, and for me, the issue seems to be those problems with `DB_NAME` and `after_wp_config_load` that we’re discussing lately.

It seems to be like this:
1. The DB you sent to me seems to already have information schema tables from the new SQLite Driver.
2. Those information schema tables have DB name value `local` in them.
3. If I try to run the without `after_wp_config_load`, then the DB name will be different (fallback to `database_name_here`). This will result in the information schema finding no tables.

I think the solution is:
1. Put back the `after_wp_config_load` — [I just did that](https://github.com/Automattic/wp-cli-sqlite-command/pull/10/commits/957e06a236640d6e61a1d0436040e9f87a6419da). This should fix the export for you.
2. Fail the export if it cannot be done rather than making it silently empty. [PR here](https://github.com/WordPress/sqlite-database-integration/pull/213).

sejas (approved):
I tried this PR with the latest commit https://github.com/Automattic/wp-cli-sqlite-command/pull/10/commits/957e06a236640d6e61a1d0436040e9f87a6419da on sqlite command and I confirm it works in all my sites, even the one that was failing.
I also have to say the export database is much faster than the regular driver.

bcotrim (approved):
I tested this PR in the context of STU-611 and it seems to be working without issues.
👍

wojtekn:
> If I try to run the without after_wp_config_load, then the DB name will be different (fallback to database_name_here). This will result in the information schema finding no tables.

@JanJakes @sejas there may still be a use case for which it won't work - for the site that was imported with a custom database name in `wp-config.php` before we changed the driver. The database uses `database_name_here` name, but `DB_NAME` in config points to a custom name, resulting in an empty dump.

However, the fix for such a case is straightforward - the user should change the database name in the config file to `database_name_here`. As it would be an edge case, with an easy workaround, we can live with that. Any thoughts?

JanJakes:
@wojtekn

> However, the fix for such a case is straightforward - the user should change the database name in the config file to database_name_here. As it would be an edge case, with an easy workaround, we can live with that. Any thoughts?

In SQLite Database Integration 2.2.4, this should fail with `Incorrect database name. The database was created with name '', but 'database_name_here' is used in the current session.`. The 2.2.4 version was not released yet, but I can tag it any time, if needed.

PR #1509, opened 2025-06-25
Normalize and validate path names when extracting downloaded tools

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/studio/security/code-scanning/2

## Proposed Changes

I propose normalizing and validating path names when extracting downloaded tools.

## Testing Instructions

1. Remove `sqlite-command` from `~/Library/Application\ Support/Studio/server-files`
2. Apply patch to always download the library:
```
diff --git a/src/lib/sqlite-command-versions.ts b/src/lib/sqlite-command-versions.ts
index 23bbb7ea..8d6ae9ab 100644
--- a/src/lib/sqlite-command-versions.ts
+++ b/src/lib/sqlite-command-versions.ts
@@ -42,10 +42,10 @@ export async function updateLatestSQLiteCommandVersion() {
                return;
        }
 
-       if ( ! distributionCheck.needsDownload ) {
-               console.log( 'SQLite Command is up to date.' );
-               return;
-       }
+       // if ( ! distributionCheck.needsDownload ) {
+       //      console.log( 'SQLite Command is up to date.' );
+       //      return;
+       // }
 
        try {
                await downloadSQLiteCommand( distributionCheck.downloadUrl, getSqliteCommandPath() );

```
3. Start Studio in dev mode
4. Confirm output logs `Downloading SQLite Command...` and that tool is downloaded

I haven't tested the fix with a ZIP file that has paths that exploit path traversal. However, I tested modifying `! filePath.startsWith( destinationFolder )` condition to include one of the files included in the sqlite-command. It worked well - the file was skipped and the message was logged.

## Pre-merge Checklist

### DIFF
--- a/vendor/wp-now/src/download.ts
+++ b/vendor/wp-now/src/download.ts
@@ -120,7 +120,15 @@ async function downloadFileAndUnzip( {
 		await response
 			.pipe( unzipper.Parse() )
 			.on( 'entry', ( entry ) => {
-				const filePath = path.join( destinationFolder, entry.path );
+				const normalizedPath = path.normalize( entry.path );
+				const filePath = path.join( destinationFolder, normalizedPath );
+
+				if ( ! filePath.startsWith( destinationFolder ) ) {
+					output?.log( `Skipping invalid path: ${ entry.path }` );
+					entryPromises.push( entry.autodrain().promise() );
+					return;
+				}
+
 				/*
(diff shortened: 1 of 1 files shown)


epeicher (approved):
I have tested it, and it works as advertised. I can see the expected log in the terminal after commenting those lines of code. LGTM! :shipit: 
CleanShot 2025-06-26 at 09 21 11@2x

PR #1510, opened 2025-06-25
Add axe DevTools to Studio

bcotrim:
## Related issues

- Related to STU-561

## Proposed Changes

- This tool helped us identify and fix some a11y issues in dotcom, I propose adding it to Studio so we can easily identify a11y issues.

## Testing Instructions

- `npm run start`
- Open dev tools and ensure axe dev tools exist and work as in the browser

**Note:** I had to restart the server once before seeing the changes. I assume it was due to the download of axe dev tools since it never occured again on the following tests.

## Pre-merge Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -220,6 +220,7 @@ async function appBoot() {
 		if ( process.env.NODE_ENV === 'development' ) {
 			await installExtension( REACT_DEVELOPER_TOOLS );
 			await installExtension( REDUX_DEVTOOLS );
+			await installExtension( 'lhdoppojpmngadmnindnejefpokejbdd' ); // Axe DevTools
 			await launchExtensionBackgroundWorkers();
 		}
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -45,6 +45,8 @@ import { setupUpdates } from 'src/updates';
 // eslint-disable-next-line import/order
 import packageJson from '../package.json';
 
+const AXE_DEVTOOLS = 'lhdoppojpmngadmnindnejefpokejbdd';
+
 if ( ! process.env.IS_DEV_BUILD ) {
 	const { sentryRelease, isDevEnvironment } = getSentryReleaseInfo( app.getVersion() );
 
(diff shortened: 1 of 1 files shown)


nightnei (commented):
Works well, as expected:
Screenshot 2025-06-25 at 16 03 28

But I see the warning. Could you please take a look, is it critical?
Screenshot 2025-06-25 at 15 51 18

But actually, I see similar for Redux DevTools, so maybe it's not a problem:
Screenshot 2025-06-25 at 15 52 00

src/index.ts
```diff
@@ -220,6 +220,7 @@ async function appBoot() {
 		if ( process.env.NODE_ENV === 'development' ) {
 			await installExtension( REACT_DEVELOPER_TOOLS );
 			await installExtension( REDUX_DEVTOOLS );
+			await installExtension( 'lhdoppojpmngadmnindnejefpokejbdd' ); // Axe DevTools
```
wojtekn: Can we use a self-documenting constant?

bcotrim:
> Works well, as expected: Screenshot 2025-06-25 at 16 03 28
> 
> But I see the warning. Could you please take a look, is it critical? Screenshot 2025-06-25 at 15 51 18
> 
> But actually, I see similar for Redux DevTools, so maybe it's not a problem: Screenshot 2025-06-25 at 15 52 00

I took a quick look. It seems Electron doesn't recognize these permissions (maybe they are specific to chrome?)
The extensions seem to work fine, and since it's a dev only feature, I would say it's ok to ignore the warnings unless we find something that is broken. What do you think?

gavande1:
Thanks for proposing Axe Devtools @bcotrim. It would be a perfect addition to dev workflows. However, I encountered multiple issues while using it within Studio. 

1. The sign-in doesn't work. It's stuck with the following screen. Probably waiting for the URL open call that never completed. 

CleanShot 2025-06-26 at 10 57 27@2x


2. The scan doesn't work when the screen is updated. It shows the following message. 

CleanShot 2025-06-26 at 10 57 38@2x

The above errors give me a hint that Axe Devtools might not work in a Studio setup. I might be wrong, though.

bcotrim:
@gavande1 thanks for testing!
It does seem that axe DevTools will not work properly in Electron for now.
I'll close this PR for now until we can find a better solution or spend more time investigating.

PR #1511, opened 2025-06-25
Improve a11y on sync dialog and tree view component

bcotrim:
## Related issues

- Fixes STU-561

## Proposed Changes

- Add screen reader section
- Add aria properties and role to tree view component

## Testing Instructions

- Turn VoiceOver on
- Go to Sync and open push/pull dialog
- Check if navigation and VoiceOver work

## Pre-merge Checklist

### DIFF
--- a/src/components/tree-view.tsx
+++ b/src/components/tree-view.tsx
@@ -74,27 +74,20 @@ const TreeItem = ( {
 	node,
 	onPatchNode,
 	level,
-	isLast,
-	index = 0,
-	setsize = 1,
+	index,
+	isLast = false,
+	siblingsLength,
 }: {
 	node: TreeNode;
 	onPatchNode: ( id: string, patchNode: Partial< TreeNode > ) => void;
 	level: number;
+	index: number;
+	siblingsLength?: number;
 	isLast?: boolean;
-	index?: number;
-	setsize?: number;
 } ) => {
 	const { __ } = useI18n();
 	const isLevel0 = level === 0;
 	const expanded = node.expanded ?? true;
-
-	const handleKeyDown = ( event: React.KeyboardEvent ) => {
-		if ( event.key === ' ' || event.key === 'Enter' ) {
-			event.preventDefault();
-			onPatchNode( node.id, { checked: node.indeterminate ? true : ! node.checked } );
-		}
-	};
 
 	return (
 		<div>
@@ -102,7 +95,7 @@ const TreeItem = ( {
 				role="treeitem"
 				aria-level={ level }
 				aria-expanded={ node.children ? expanded : undefined }
-				aria-setsize={ setsize }
+				aria-setsize={ siblingsLength }
 				aria-posinset={ index + 1 }
 				aria-checked={ node.indeterminate ? 'mixed' : node.checked }
 				aria-label={ node.label }
@@ -154,7 +147,7 @@ const TreeItem = ( {
 							onPatchNode={ onPatchNode }
 							level={ level + 1 }
 							index={ idx }
-							setsize={ node.children ? node.children.length : 0 }
+							siblingsLength={ node.children?.length }
 						/>
 					) ) }
 				</div>
@@ -182,7 +175,7 @@ export const TreeView = ( { tree, setTree }: TreeViewProps ) => {
 					onPatchNode={ handlePatchNode }
 					level={ 1 }
 					index={ index }
-					setsize={ tree.length }
+					siblingsLength={ tree.length }
 					isLast={ index === tree.length - 1 }
 				/>
 			) ) }
--- a/src/components/tree-view.tsx
+++ b/src/components/tree-view.tsx
@@ -105,7 +105,7 @@ const TreeItem = ( {
 					isLast ? 'border-white' : ''
 				) }
 			>
-				<div className="flex items-center cursor-pointer">
+				<label className="flex items-center cursor-pointer">
 					<CheckboxControl
 						id={ node.id }
 						checked={ node.checked }
@@ -113,18 +113,16 @@ const TreeItem = ( {
 						onChange={ ( checked: boolean ) => onPatchNode( node.id, { checked } ) }
 						__nextHasNoMarginBottom
 					/>
-					<label htmlFor={ node.id } className="flex items-center cursor-pointer">
-						{ node.type && (
-							<Icon
-								aria-hidden
-								icon={ TREE_NODE_ICONS[ node.type ] }
-								size={ 20 }
-								className="me-1.5"
-							/>
-						) }
-						{ node.label }
-					</label>
-				</div>
+					{ node.type && (
+						<Icon
+							aria-hidden
+							icon={ TREE_NODE_ICONS[ node.type ] }
+							size={ 20 }
+							className="me-1.5"
+						/>
+					) }
+					{ node.label }
+				</label>
 				{ node.loading && <Spinner className="!w-[9px] !h-[9px] !m-0" /> }
 				{ ! node.loading && node.children && ! node.hideExpandButton && (
 					<button
--- a/src/components/tree-view.tsx
+++ b/src/components/tree-view.tsx
@@ -1,5 +1,6 @@
 import { CheckboxControl, Icon, Spinner } from '@wordpress/components';
 import { file, page } from '@wordpress/icons';
+import { useRef, useState, useEffect } from 'react';
 import { RightArrowIcon } from 'src/components/icons/right-arrow';
 import { cx } from 'src/lib/cx';
 
@@ -75,6 +76,10 @@ const TreeItem = ( {
 	isLast,
 	index = 0,
 	setsize = 1,
+	focusedId,
+	setFocusedId,
+	nodeRefs,
+	visibleNodeIds,
 }: {
 	node: TreeNode;
 	onPatchNode: ( id: string, patchNode: Partial< TreeNode > ) => void;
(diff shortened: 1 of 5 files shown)


src/components/sync-dialog.tsx
```diff
@@ -277,7 +292,13 @@ export function SyncDialog( {
 			<div className="pb-[70px]">
 				<div className="px-8 pb-6 pt-2">{ copy[ envDetails.envType ].description }</div>
 				<div className="px-8">
-					<div className="flex items-start gap-1 pb-7 border-b border-a8c-gray-5">
+					<span className="sr-only">
+						{ sprintf( __( 'From %s to %s' ), syncFromText, syncToText ) }
```
gavande1: Nitpick: Should we add a translator comment?

gavande1 (approved):
This works as expected. LGTM 👍 

However, I noticed one issue when using VoiceOver. I was not able to go to the next section (folder tree) without pressing the tab key. I would expect it to go to the next section with control + command + right arrow key. Am I missing anything? 

https://github.com/user-attachments/assets/fb57901e-50ef-424d-8d14-599549a43dae

bcotrim:
@gavande1 thanks for the review!
I spent some more time testing and validating the VoiceOver navigation on `tree` and `treeitems` and I believe it is working as expected. I didn't find any big improvements for the navigation, but I did improve the visibility of some of the labels and images for a clearer navigation.
Using `control + option + arrow` I was able to navigate to the tree and, as the reader announced, I am able to "drill down" using `shift + control + option + down arrow`.
Inside the tree items I was also able to navigate to the items and drill down to check/uncheck and expand/collapse.


https://github.com/user-attachments/assets/8de7fc6d-2481-4eb7-9ae6-8ca21750c40d

Can you test again and see if it looks good? If it doesn't, do you have any suggestions on how to approach it?
Thanks again for bringing this up and doing a thorough test test.

src/components/tree-view.tsx
```diff
@@ -73,53 +75,86 @@ const TreeItem = ( {
...
+	const { __ } = useI18n();
 	const isLevel0 = level === 0;
 	const expanded = node.expanded ?? true;
 
+	const handleKeyDown = ( event: React.KeyboardEvent ) => {
+		if ( event.key === ' ' || event.key === 'Enter' ) {
+			event.preventDefault();
+			onPatchNode( node.id, { checked: node.indeterminate ? true : ! node.checked } );
+		}
+	};
```
nightnei: We don't use it
```suggestion
```

src/components/tree-view.tsx
```diff
@@ -73,53 +75,86 @@ const TreeItem = ( {
...
+				>
+					{ node.children.map( ( child, idx ) => (
 						<TreeItem
 							key={ child.id }
 							node={ child }
 							onPatchNode={ onPatchNode }
-							level={ ++level }
+							level={ level + 1 }
+							index={ idx }
+							setsize={ node.children ? node.children.length : 0 }
```
nightnei: 1. Here we use `0` as default value
2. But for the `TreeItem` arguments we have useless default value
3. I see that we don't have fallback inside `TreeView`

So we shoudl either have default value for TreeItem areguments and don't provide here `0` as fallback or provide `0` fallback inside TreeView

src/components/tree-view.tsx
```diff
@@ -73,53 +75,86 @@ const TreeItem = ( {
 	onPatchNode,
 	level,
 	isLast,
+	index = 0,
+	setsize = 1,
```
nightnei: Hm, I don't understand why it's `1`... If we don't have `children` then it should be `0`

src/components/tree-view.tsx
```diff
@@ -73,53 +75,86 @@ const TreeItem = ( {
 	onPatchNode,
 	level,
 	isLast,
+	index = 0,
```
nightnei: I thiunk we don't need default value
```suggestion
```

src/components/tree-view.tsx
```diff
@@ -139,13 +174,15 @@ export const TreeView = ( { tree, setTree }: TreeViewProps ) => {
...
+		<div role="tree">
 			{ tree.map( ( node, index ) => (
 				<TreeItem
 					key={ node.id }
 					node={ node }
 					onPatchNode={ handlePatchNode }
-					level={ 0 }
+					level={ 1 }
+					index={ index }
+					setsize={ tree.length }
```
nightnei: I see that `setsize` is the name for aria, but we should not depend on names where it's used. In our codebase it would be clear to show that `setsize` is the number of all siblings, otherwise it's not clear what `set` means.
```suggestion
					siblingsLength={ tree.length }
```

src/components/tree-view.tsx
```diff
@@ -73,53 +75,86 @@ const TreeItem = ( {
...
 				className={ cx(
 					'flex items-center py-2 relative gap-2',
 					isLevel0 ? 'border-b border-gray-300 py-4' : '',
 					isLast ? 'border-white' : ''
 				) }
 			>
-				<label className="flex items-center cursor-pointer">
+				<div className="flex items-center cursor-pointer">
 					<CheckboxControl
+						id={ node.id }
```
nightnei, line 111: `id` in general is not a good practice. And according to my outdated experience - there is no difference between wrapping with `<label>...</label>` and using `htmlFor`, but in case of creating wrapper, we don't need to add redundant `id` property.

Do we really need to have `id` + `htmlFor` instead of previous approach with just wrapping with `label` teh content?

nightnei: Also, with this change - teh folder icon is not clickable

nightnei (changes requested):

src/components/tree-view.tsx
bcotrim: If we have the label on everything the VoiceOver navigation doesn't select the items in the same way.
label + htmlFor:

https://github.com/user-attachments/assets/5fc47654-7e48-477f-865f-8a18209c3f5e

label wrapper:

https://github.com/user-attachments/assets/48e4fe91-f5b1-4fcf-b85e-d0c027fdaa6e

Edit: The sound was not recorded, in both recordings I am trying to select using VoiceOver (control + option + space), in the second one, only if I drill down to the checkbox am I able to do it.

bcotrim: > Also, with this change - teh folder icon is not clickable

I've addressed this in https://github.com/Automattic/studio/pull/1511/commits/1658b26dfcd61bf03e35bf0a261be15d9b8c7df2

gavande1:
Thanks @bcotrim for the efforts you're putting in to improve a11y. I have tested this again and almost everything is working as expected except one thing. 

I was not able to open the 2nd-level tree menu using the keyboard. .i.e. themes

https://github.com/user-attachments/assets/a53cdd79-8f16-4e00-bb64-ba529b5c92ae

nightnei:
> I was not able to open the 2nd-level tree menu using the keyboard. .i.e. themes

Yep, same

bcotrim:
> Thanks @bcotrim for the efforts you're putting in to improve a11y. I have tested this again and almost everything is working as expected except one thing.
> 
> I was not able to open the 2nd-level tree menu using the keyboard. .i.e. themes

> Yep, same

The best solution I found was to navigate inside the `treeitem` with shift+control+option+down and navigate to the expand/collapse button (see video)
The only other alternative I can think of is implementing full keyboard navigation as I did in https://github.com/Automattic/studio/pull/1511/commits/ff18ac4ac757163e10ea0706f3248a4d01c6140d but we decided it was out of scope for this PR.
Please let me know what you think.

https://github.com/user-attachments/assets/2f8c8913-433a-4652-ba06-04f810187d20

nightnei:
@bcotrim @gavande1 I have a question - why do we test with such dificult combinations? I see that typical approach with `tab` / `shift tab` also works perfectly


@bcotrim and in case of using `tab` / `shift tab` everything works perfectly with avoiding usibng `htmlFor`:
I see that visual selection doesn't match, sometimes, what VoiceOver sees with `tab` / `shift tab`, but - VoiceOver is for people with eye issues, so it means that the app should work perfectly either visually or with teh help of sounds (VoiceOver). So we can proceed with using <label> as a wrapper:

```diff
--- a/src/components/tree-view.tsx
+++ b/src/components/tree-view.tsx
@@ -105,15 +105,14 @@ const TreeItem = ( {
                                        isLast ? 'border-white' : ''
                                ) }
                        >
-                               <div className="flex items-center cursor-pointer">
+                               <label className="flex items-center cursor-pointer">
                                        <CheckboxControl
-                                               id={ node.id }
                                                checked={ node.checked }
                                                indeterminate={ node.indeterminate }
                                                onChange={ ( checked: boolean ) => onPatchNode( node.id, { checked } ) }
                                                __nextHasNoMarginBottom
                                        />
-                                       <label htmlFor={ node.id } className="flex items-center cursor-pointer">
+                                       <div className="flex items-center cursor-pointer">
                                                { node.type && (
                                                        <Icon
                                                                aria-hidden
@@ -123,8 +122,8 @@ const TreeItem = ( {
                                                        />
                                                ) }
                                                { node.label }
-                                       </label>
-                               </div>
+                                       </div>
+                               </label>
                                { node.loading && <Spinner className="!w-[9px] !h-[9px] !m-0" /> }
```


https://github.com/user-attachments/assets/0d461b51-05b3-4b0c-b55f-62226e1254b1

bcotrim:
> @bcotrim and in case of using tab / shift tab everything works perfectly with avoiding usibng htmlFor:
I see that visual selection doesn't match, sometimes, what VoiceOver sees with tab / shift tab, but - VoiceOver is for people with eye issues, so it means that the app should work perfectly either visually or with teh help of sounds (VoiceOver). So we can proceed with using as a wrapper:

Done

src/modules/sync/lib/convert-tree-to-options-to-sync.ts
```diff
@@ -49,7 +49,10 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOptionsWithS
 			) {
 				const selectedItems = item.children
 					.filter( ( child ) => child.checked )
-					.map( ( child ) => child.id );
+					.map( ( child ) => {
+						const prefix = `${ item.id }-`;
+						return child.id.startsWith( prefix ) ? child.id.slice( prefix.length ) : child.id;
+					} );
```
nightnei: The last one thing - I was thinking, how to avoid this code and - WDYT about adding property name to the items of the tree?
So we have:
1. `id` what should be uniqueue and is used for updating tree items
2. `name` which is original items id (e.g. `index.php` or `akismet`)
As a result - we don't mix logic and can keep everything clear and separated.

```suggestion
.map( ( child ) => child.name );
```

nightnei (changes requested):

nightnei:
> The best solution I found was to navigate inside the treeitem with shift+control+option+down and navigate to the expand/collapse button (see video)

@bcotrim BTW, I was playign more with it and it was the best approach to enter the table, since it's exectly what VoicOver recommends to enter inside of it 👍 
Screenshot 2025-07-01 at 16 02 40

nightnei (approved):
Oh, I missed that we need to add id/name duplication for upper levels...
As for me, it still better then slicing id etc

LGTM, thanks for addressing 👍

bcotrim:
@gavande1 going to merge this one since we have 2 approval, but please let me know if you have any additional feedback and I'll address it in a follow up PR.

gavande1:
@bcotrim Thanks for merging the PR. I am fine with the changes as long as other feels okay with them. Thanks for letting me know as well before merging. 👍

PR #1512, opened 2025-06-26
Update Sync site budges

nightnei:
## Related issues

- Fixes STU-572

## Proposed Changes
Comment from Marina:
```
In order to simplify and avoid putting very lengthy names in the field, we aligned on a couple of things:

We’re only going to be showing the site name, and not the staging site link

We’re going to use the labels currently used in Studio for Production and Staging until the new hosting dashboard gets implemented. Then, we’ll unify Studio and wordpress.com to have a consistent UI
```

## Testing Instructions
3. Open Sync tab
4. Connect Studio and dotcom site
5. Click on Pull or Push
6. Assert that you see dialog with:
	6.1 Instead of domains, you see local's and dotcom's site names
	6.2 You see our default env badges

<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2025-06-26 at 13 24 37
<td>Screenshot 2025-06-26 at 13 24 14
</tr>
</table>


src/components/environment-badge.tsx
```diff
@@ -37,3 +39,32 @@ export function EnvironmentBadge( { type, selected }: EnvironmentBadgeProps ) {
 
 	return <Badge className={ cx( getClassName() ) }>{ labels[ type ] }</Badge>;
 }
+
+export const useSiteEnvDetails = (
```
bcotrim: Have you considered refactoring this hook as a component?
It seems strange to me to have a hook just to return a component function.
I think we could achieve the same thing using a component. What do you think?

src/components/environment-badge.tsx
```diff
@@ -37,3 +39,32 @@ export function EnvironmentBadge( { type, selected }: EnvironmentBadgeProps ) {
...
+		},
+		sandbox: {
+			AssignedEnvironmentBadge: () => <EnvironmentBadge type="sandbox" />,
+			envType: 'sandbox',
+		},
+	} as const;
+
+	if ( connectedSite.isPressable ) {
+		return (
+			envTypeValues[ connectedSite.environmentType as EnvironmentType ] ?? envTypeValues.production
```
bcotrim: Could we adjust the type on `SyncSite` so we don't have to do this cast?

src/components/environment-badge.tsx
nightnei: I considered it, but we also need to generate env name `staging` | `production` | `sandbox`

src/components/environment-badge.tsx
nightnei: But it's `SyncSite`

nightnei: The problem is - `environmentType` doesn't exists for dotcom sites. Theoretically we can try to improve typing to distinguish dotcom and pressable sites, but it will take more effort, so I would keep it as is, sicne it's out of project. We shoudl focus only on project progress ATM.

src/components/environment-badge.tsx
bcotrim: Could that be a helper function? the logic seems relatively small.

```typescript
export const getEnvironmentType = (connectedSite: SyncSite): EnvironmentType => {
	if (connectedSite.isPressable) {
		return (connectedSite.environmentType as EnvironmentType) ?? 'production';
	}
	return connectedSite.isStaging ? 'staging' : 'production';
};
```

nightnei: Makes sense, good idea 👍

src/components/environment-badge.tsx
bcotrim: Wouldn't this be enough then?

```typescript
export type EnvironmentType = 'production' | 'staging' | 'sandbox';

export type SyncSite = {
	id: number;
	localSiteId: string;
	name: string;
	url: string;
	isStaging: boolean;
	isPressable: boolean;
	environmentType?: EnvironmentType | null;
	stagingSiteIds: number[];
	syncSupport: SyncSupport;
	lastPullTimestamp: string | null;
	lastPushTimestamp: string | null;
};
```

nightnei: I think it's also not nice approach, since we are mixing types, since `environmentType` is null only for dotcom but opposite for Pressable. So I think that if we want to improve it, then we should do it in correct way with:

```
{
environmentType: 'production' | 'staging' | 'sandbox'
isPressable: true,
} |
{
isPressable: false,
}
```

But it will take much more effort to test teh whole codebase and make fixes. Let's focus on the project.

bcotrim (approved):
Looks clean and works as described!
LGTM 👍

sejas (commented):
Thanks for updating the site boxes. I tested it with fake long site names and  it seems to work better than before. I wonder if there are real sites with loooooooong words, in that case I think we might face the same issue as before.

Screenshot 2025-07-02 at 21 09 55
Screenshot 2025-07-02 at 21 11 23

PR #1514, opened 2025-06-26
Clean up sync-dialog

nightnei:
## Related issues

- Fixes STU-580

## Proposed Changes
With this PR we are cleaning up sync-dialog with moving hooks outside of module and wrapping it with a folder.

## Testing Instructions
1. Open Sync tab
2. Connect dotcom site
3. Click on Pull or Push button and assert that you see popup and everything looks good. Also try to click buttons to make sure that everything works as before.


src/components/sync-dialog/use-default-sync-tree.tsx
epeicher, line 1: Just curious, the linked task mentions moving these hooks to `src/hooks` and also mentions any separate file. Is there any reason you preferred them in the same folder? (It's a genuine question to know more about best practices in the codebase, not a suggestion 😄)

epeicher (approved):
Thanks @nightnei for the refactor, code looks cleaner now 🙌 . I have tested it, and it works as expected, I have not found any issues. LGTM! :shipit: 

|Push | Pull|
|-------|------|
| CleanShot 2025-06-27 at 09 43 59@2x| CleanShot 2025-06-27 at 09 40 25@2x|

src/components/sync-dialog/use-default-sync-tree.tsx
nightnei: 1. One day before, on our weekly meeting, I raised the issue that we have very linear structure, instead of hierarchical (very huge `components`, `hooks` etc common folders). Everybody agreed and also we agreed to have specific structure (can't find the task with that structure). Long story short - initially it was good to have global common folders, but now Studio is pretty big, so it's better to keep hierarchical structure.
2. `src/hooks` is folder for common hooks, but hooks in this PR is used only in one place ATM and additionally they are very specific only for sync-dialog (I don't anticipate to have them used in other places), so I don't see any argument to keep them globally, instead of locally for sync-dialog. Moreover, now they are bounded to sync-dialog with the help of `sinc-dialog` dir, so it will simplify navigation for developers and it will be easier to visually find all functions connected to this dialog.

Somethign like this :)

nightnei: As option, we coudl also create `sync-dialog/hooks/*` dir, but ATM we have just 2 hooks ano no other types of modules for sync-dialog, so I decided that extra `hooks` folder will be redundent ATM.

epeicher: I 100% agree with your reasoning here! Those are very valid reasons in my opinion:
- hooks in this PR is used only in one place ATM and additionally they are very specific only for sync-dialog
- I don't anticipate to have them used in other places
- they are bounded to sync-dialog with the help of sinc-dialog dir, so it will simplify navigation for developers and it will be easier to visually find all functions connected to this dialog.

Thanks for the detailed explanation 🫶

wojtekn: See https://github.com/Automattic/studio/blob/trunk/docs/code-contributions.md#project-structure

bcotrim (commented):
Would it make sense to move these to a new folder in `modules`?
From my understanding `components` should be for re-usable components, and a lot of this logic is specific to the `sync-sites` feature.
Maybe the `tree-view` component would be an exception, it is a re-usable component.

bcotrim (approved):

wojtekn:
> Would it make sense to move these to a new folder in modules?
> From my understanding components should be for re-usable components, and a lot of this logic is specific to the sync-sites feature.
> Maybe the tree-view component would be an exception, it is a re-usable component.

Agreed. Module could be `preview-site`, `user-settings`, `sync`, or `import-export`. It makes sense to move, or start moving, all the code specific to Sync to its own module.

sejas (commented):
Thanks for start moving the new sync files to the modules folder.

PR #1516, opened 2025-06-26
Add selective sync to uploads folders

bcotrim:
## Related issues

- Fixes STU-579

## Proposed Changes

- Allows selective sync for `uploads` folder content
- `plugins`, `themes` and `uploads` are collapsed by default
- Display more files types in the sync dialog tree

## Testing Instructions

- Create a site or use an existing one and connect it with a remote site.
- Use "Push" option
- Validate `wp-content` is expanded but `plugins`, `themes` and `uploads` are collapsed by default
- Expand `uploads` and verify the options match the file contents from your site folder
- Do the push operation and validate only the selected options were pushed to the remote site.

## Pre-merge Checklist

### DIFF
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -74,9 +74,9 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOptionsWithS
 
 			if (
 				item.children &&
-				( item.id === SYNC_OPTIONS.plugins ||
-					item.id === SYNC_OPTIONS.themes ||
-					item.id === SYNC_OPTIONS.uploads )
+				[ SYNC_OPTIONS.plugins, SYNC_OPTIONS.themes, SYNC_OPTIONS.uploads ].includes(
+					item.id as 'plugins' | 'themes' | 'uploads'
+				)
 			) {
 				const selectedItems = item.children
 					.filter( ( child ) => child.checked )
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -42,12 +42,15 @@ export type SyncOptionsWithSelections = {
 	specificSelections?: {
 		plugins?: string[];
 		themes?: string[];
+		uploads?: string[];
 	};
 };
 
 export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOptionsWithSelections => {
 	const optionsToSync: SyncOption[] = [];
-	let specificSelections: { plugins?: string[]; themes?: string[] } | undefined = undefined;
+	let specificSelections:
+		| { plugins?: string[]; themes?: string[]; uploads?: string[] }
+		| undefined = undefined;
 
 	const isAll = tree.every( ( node ) => node.checked );
 	if ( isAll ) {
@@ -73,7 +76,9 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOptionsWithS
 
 			if (
 				item.children &&
-				( item.id === SYNC_OPTIONS.plugins || item.id === SYNC_OPTIONS.themes )
+				( item.id === SYNC_OPTIONS.plugins ||
+					item.id === SYNC_OPTIONS.themes ||
+					item.id === SYNC_OPTIONS.uploads )
 			) {
 				const selectedItems = item.children
 					.filter( ( child ) => child.checked )
--- a/src/components/sync-connected-sites.tsx
+++ b/src/components/sync-connected-sites.tsx
@@ -48,9 +48,7 @@ export type SyncOptionsWithSelections = {
 
 export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOptionsWithSelections => {
 	const optionsToSync: SyncOption[] = [];
-	let specificSelections:
-		| { plugins?: string[]; themes?: string[]; uploads?: string[] }
-		| undefined = undefined;
+	let specificSelections: SyncOptionsWithSelections[ 'specificSelections' ] = undefined;
 
 	const isAll = tree.every( ( node ) => node.checked );
 	if ( isAll ) {
--- a/src/components/tests/convert-tree-to-options-to-sync.tsx
+++ b/src/components/tests/convert-tree-to-options-to-sync.tsx
@@ -65,4 +65,252 @@ describe( 'convertTreeToOptionsToSync', () => {
 			specificSelections: undefined,
 		} );
 	} );
+
+	describe( 'partial selections', () => {
+		it( 'returns partial plugins selection when only some plugins are selected', () => {
+			const { result } = renderHook( () => useDefaultTree() );
+			let tree = result.current;
+
+			tree = updateNodeById( tree, 'filesAndFolders', { checked: false } );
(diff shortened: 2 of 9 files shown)


src/components/tests/convert-tree-to-options-to-sync.tsx
```diff
@@ -37,6 +37,21 @@ describe( 'convertTreeToOptionsToSync', () => {
 		} );
 	} );
 
+	it( 'returns ["uploads"] when only uploads are selected', () => {
+		const { result } = renderHook( () => useDefaultTree() );
+		let tree = result.current;
+
+		tree = updateNodeById( tree, 'filesAndFolders', { checked: false } );
+		tree = updateNodeById( tree, 'sqls', { checked: false } );
```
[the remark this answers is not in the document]
bcotrim: Unchecking the parent will also uncheck its children, so I don't think this is necessary.

src/components/sync-connected-sites.tsx
```diff
@@ -73,7 +74,9 @@ export const convertTreeToOptionsToSync = ( tree: TreeNode[] ): SyncOptionsWithS
 
 			if (
 				item.children &&
-				( item.id === SYNC_OPTIONS.plugins || item.id === SYNC_OPTIONS.themes )
+				( item.id === SYNC_OPTIONS.plugins ||
+					item.id === SYNC_OPTIONS.themes ||
+					item.id === SYNC_OPTIONS.uploads )
```
nightnei:
```suggestion
				[ SYNC_OPTIONS.plugins, SYNC_OPTIONS.themes , SYNC_OPTIONS.uploads ].includes( item.id )
```

src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -164,7 +164,8 @@ export class DefaultExporter extends EventEmitter implements Exporter {
 			const partialFolderItems =
 				this.options.specificSelections &&
 				( ( category === 'plugins' && this.options.specificSelections.plugins ) ||
-					( category === 'themes' && this.options.specificSelections.themes ) );
+					( category === 'themes' && this.options.specificSelections.themes ) ||
+					( category === 'uploads' && this.options.specificSelections.uploads ) );
```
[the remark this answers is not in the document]
nightnei: `const partialFolderItems = this.options.specificSelections?.[ category ];`

src/ipc-handlers.ts
```diff
@@ -1415,9 +1422,8 @@ export async function listWpContentFolders(
 				name: e.name.toString(),
 				type: e.isDirectory() ? ( 'folder' as const ) : ( 'file' as const ),
 			} ) )
-			.filter( ( entry: { name: string; type: string } ) => {
-				if ( entry.type === 'folder' ) return true;
-				return entry.type === 'file' && entry.name.toLowerCase().endsWith( '.php' );
+			.filter( ( entry: { name: string } ) => {
+				return ! entry.name.startsWith( '.' );
```
nightnei: Could you please elaborate on this change?
I rtemember I googled another solutions even during previous PR and saw that even libs are using approach with `startsWith( '.' )`, but I had impression that it works for macOS but doesn't work for Windows.

Also, let's add some comment with explanation e.g. `We are good to show all files and folders, with exception of hidden items. And `startsWith( '.')` is enough to check check for hidden items`.

src/constants.ts
```diff
@@ -78,3 +78,7 @@ export const IPC_VOID_HANDLERS = < const >[
 	'showNotification',
 	'authenticate',
 ];
+
+// Folders in wp-content that can be selectively synced
+export const WP_CONTENT_FOLDERS = [ 'plugins', 'themes', 'uploads' ] as const;
+export type WpContentFolder = ( typeof WP_CONTENT_FOLDERS )[ number ];
```
nightnei, line 84: This file only about `const` variables, so I would not start mixing them with TS types.

Also, `WP_CONTENT_FOLDERS` is specific for sync-functionality, since it contains folders only for granular sync, so I want to propose to rename it and move inside sync-dialog folder. 

e.g.
```
// src/components/sync-dialog/constants.ts
export const GRANULAR_SYNC_FOLDERS = [ 'plugins', 'themes', 'uploads' ] as const;
``` 
And:
```
// src/components/sync-dialog/types.ts
import { GRANULAR_SYNC_FOLDERS } from './contsatnt';

export type GranularSyncFolders = ( typeof GRANULAR_SYNC_FOLDERS )[ number ];
```

nightnei (changes requested):
Works perfectly and teh code in general LGTM 👍 Just left a few comments

src/components/sync-connected-sites.tsx
bcotrim: This will force us to use `as` for the type assertion, which I think it's a bad approach, for no benefit besides a little bit of syntax sugar, but I'll make the change.
```
[ SYNC_OPTIONS.plugins, SYNC_OPTIONS.themes, SYNC_OPTIONS.uploads ].includes( item.id as 'plugins' | 'themes' | 'uploads' )
```

src/ipc-handlers.ts
bcotrim: Previously we were only displaying folders and `.php` files on the tree.
In the upload folder I believe it makes sense to change that condition to allow other file types.

src/constants.ts
bcotrim: Could we address this in https://github.com/Automattic/studio/pull/1514 ?
Since it's just a matter of where constants and types are defined, and we are revisiting the structure of this feature there, it seems to make sense.
What do you think?

src/lib/import-export/export/exporters/default-exporter.ts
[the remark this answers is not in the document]
bcotrim: Can you take a look at https://github.com/Automattic/studio/pull/1516/commits/24aa38f3a7bcd5002729b9f75309608acad3f459 and let me know if this approach looks better, please?

src/ipc-handlers.ts
bcotrim: Instead of adding the comment, I tried to make it more explicit that we are filtering hidden files: https://github.com/Automattic/studio/pull/1516/commits/731e0b4c1c671d17bf95c8db78072758d1f8cd92
What do you think?

src/lib/import-export/export/exporters/default-exporter.ts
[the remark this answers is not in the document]
nightnei: I don't see benefits of creating separate fucntion and handling every case manually with copuy-pasting it, btu it's not critical

src/ipc-handlers.ts
nightnei: > Instead of adding the comment, I tried to make it more explicit that we are filtering hidden files: https://github.com/Automattic/studio/commit/731e0b4c1c671d17bf95c8db78072758d1f8cd92
What do you think?

Personally for me, this is the case when comment fits the best. Since `hidden` property looks extra. But I don't have strong preference.

src/constants.ts
nightnei: Ok, I will do it there

nightnei (approved):
LGTM

sejas (commented):
Thanks for adding support to uploads child folders and adding tests. I confirm it works as expected on trunk.

PR #1518, opened 2025-06-27
Remove selective sync feature flag

bcotrim:
## Related issues

- Fixes STU-552

## Proposed Changes

- Remove selective sync feature flag

## Testing Instructions

- `npm run start`
- Try to push/pull a site
- Selective sync dialog should display
- Pushing or pulling should work as expected
- `npm run test`
- All tests should pass

## Pre-merge Checklist

### DIFF
--- a/src/lib/feature-flags.ts
+++ b/src/lib/feature-flags.ts
@@ -6,24 +6,22 @@ export interface FeatureFlagDefinition {
 }
 
 export interface FeatureFlags {
-	selectiveSyncEnabled: boolean;
+	placeholder: boolean; // placeholder to avoid type errors
 }
 
 export const FEATURE_FLAGS: Record< keyof FeatureFlags, FeatureFlagDefinition > = {
-	selectiveSyncEnabled: {
-		label: 'Selective Sync',
-		env: 'STUDIO_SELECTIVE_SYNC',
-		flag: 'selectiveSyncEnabled',
-		default: false,
-	},
-};
+	// No feature flags currently defined
+	// Add new feature flags here as needed
+} as Record< keyof FeatureFlags, FeatureFlagDefinition >;
 
 export function getFeatureFlagFromEnv( flag: keyof FeatureFlags ): boolean {
-	const envKey = FEATURE_FLAGS[ flag ].env;
-	return process.env[ envKey ] === 'true';
+	const envKey = FEATURE_FLAGS[ flag ]?.env;
+	return envKey ? process.env[ envKey ] === 'true' : false;
 }
 
 export function setFeatureFlagInEnv( flag: keyof FeatureFlags, value: boolean ): void {
-	const envKey = FEATURE_FLAGS[ flag ].env;
-	process.env[ envKey ] = value ? 'true' : 'false';
+	const envKey = FEATURE_FLAGS[ flag ]?.env;
+	if ( envKey ) {
+		process.env[ envKey ] = value ? 'true' : 'false';
+	}
 }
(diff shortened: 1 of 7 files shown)


src/lib/feature-flags.ts
```diff
@@ -6,24 +6,22 @@ export interface FeatureFlagDefinition {
 }
 
 export interface FeatureFlags {
-	selectiveSyncEnabled: boolean;
+	placeholder: boolean; // placeholder to avoid type errors
```
nightnei: Hm, AFAIR previously we just left empty as `FeatureFlags {}`, but actually even `placeholder: boolean` is ok

src/lib/feature-flags.ts
```diff
@@ -6,24 +6,22 @@ export interface FeatureFlagDefinition {
...
-	selectiveSyncEnabled: {
-		label: 'Selective Sync',
-		env: 'STUDIO_SELECTIVE_SYNC',
-		flag: 'selectiveSyncEnabled',
-		default: false,
-	},
-};
+	// No feature flags currently defined
+	// Add new feature flags here as needed
+} as Record< keyof FeatureFlags, FeatureFlagDefinition >;
```
nightnei: Check please, what did we have before, since I don't recall so dificultiers with empty features

src/lib/feature-flags.ts
```diff
@@ -6,24 +6,22 @@ export interface FeatureFlagDefinition {
...
+	return envKey ? process.env[ envKey ] === 'true' : false;
 }
 
 export function setFeatureFlagInEnv( flag: keyof FeatureFlags, value: boolean ): void {
-	const envKey = FEATURE_FLAGS[ flag ].env;
-	process.env[ envKey ] = value ? 'true' : 'false';
+	const envKey = FEATURE_FLAGS[ flag ]?.env;
+	if ( envKey ) {
+		process.env[ envKey ] = value ? 'true' : 'false';
+	}
```
nightnei: Hm, could you please check and make sure that it's exectly the way of removing feature, which we took before. Since I recall that we typically removed features with less changes. But maybe I am wrong 😅

nightnei (changes requested):

bcotrim:
@nightnei thanks for the review!
Some of the adjustment to the empty types result from the new feature flag menu: https://github.com/Automattic/studio/pull/1483
When I initially made the PR I didn't account for empty feature flags state. This should be a one time adjustment, next time it will be just adding or remove entries from the types.

We did have some unnecessary changes I reverted, I was trying to allow the type to be empty, but the `placeholder` is a lot simpler to achieve and doesn't hurt.

Let me know what you think.

nightnei (approved):
LGTM and Pull/Push work correctly 👍

bcotrim:
@sejas @nightnei leaving this open until it makes sense merging.
Feel free to merge it, or let me know when we should, please.

src/lib/feature-flags.ts
wojtekn: If the type can be empty, could we prepare it to allow for that?

bcotrim: > If the type can be empty, could we prepare it to allow for that?

Yes, the feature flag types need an improvement. 
I would suggest doing that in a follow-up PR to not increase the scope of this PR.
What do you think?

wojtekn: I think it makes sense to fix handling for empty type as a part of the PR that causes the type to be empty.

sejas (approved):
Thanks for removing the feature flag. I tested it by running `npm start` and confirm that selective sync works as expected by default.
Let's merge it so we can share the beta with the feature active for CfT.

PR #1521, opened 2025-06-30
Refactor structure of Sync dirs

nightnei:
## Proposed Changes
With this PR we are moving all Sync related files inside `src/modules/content-tab-sync`

## Testing Instructions
1. run `STUDIO_SELECTIVE_SYNC=true npm start`
2. Open Sync tab and open browser's console
3. Click on all buttons and assert that you don't see errors


wojtekn:
@nightnei to make it consistent with current naming convention, could we make it `src/modules/sync`?

nightnei:
> @nightnei to make it consistent with current naming convention, could we make it src/modules/sync?

@wojtekn 👍 Done

wojtekn (approved):

bcotrim (commented):
Some comments:
- I think we are adding unnecessary depth with the `sync-dialog` folder.
- `SyncPullPushClear` could be moved from `src/components` to `src/modules/sync/components`
- We could create a `lib` folder to hold some file, like `convert-tree-to-options-to-sync.ts`

Here is an example of the structure how I see it, and following the code-contributions guidelines

```
src/modules/sync/
  ├── components/
  │   ├── sync-dialog.tsx 
  │   ├── connect-button.tsx
  │   ├── create-button.tsx
  │   ├── sync-connected-sites.tsx
  │   ├── sync-sites-modal-selector.tsx
  │   ├── sync-tab-image.tsx
  │   └── sync-pull-push-clear.tsx
  ├── hooks/
  │   ├── use-default-sync-tree.tsx
  │   └── use-sync-texts.tsx
  ├── lib/
  │   ├── convert-tree-to-options-to-sync.ts
  │   ├── sync-options.ts        # Constants and types
  │   └── granular-sync-folders.ts
  └── tests/
      └── ...
```

What do you think about this?

nightnei:
Done:
1. Moved sync-tab-image.tsx from `images` to `components`
2. Moved sync-pull-push-clear.tsx

For discussion:
1. use-sync-texts.tsx is copy just specifically for teh dialog, they shoudl; be as close as possible and this hook is not used on sync page directly and won't be used anywhere, so I would keep it as is. The same for use-default-sync-tree.tsx.
2. `lib/` I am ok to go with this folder, but we should agree this namer across whole codebase, for similar purposes. i woudl propose something like `utils`, but if you prefer `lib`, then I am ok with it.
3. `sync-options.ts        # Constants and types` - is used across the whole codebase. And we agreed that - if something is used locally in one place, then it shoudl be inside module etc, but if necessary to use it in a few places, then we will move it outside to global places. I think it's not only about components, but also about such independent types and constants.

bcotrim:
> use-sync-texts.tsx is copy just specifically for teh dialog, they shoudl; be as close as possible and this hook is not used on sync page directly and won't be used anywhere, so I would keep it as is. The same for use-default-sync-tree.tsx.

I understand it's just used on this component. My suggestion would be to avoid having another depth level. To me it makes sense to have all the feature specific hooks in `modules/sync/hooks`.  Following the structure here: https://github.com/Automattic/studio/blob/trunk/docs/code-contributions.md#feature-modules

> lib/ I am ok to go with this folder, but we should agree this namer across whole codebase, for similar purposes. i woudl propose something like utils, but if you prefer lib, then I am ok with it.

`lib` is the definition according to https://github.com/Automattic/studio/blob/trunk/docs/code-contributions.md#feature-modules

> sync-options.ts        # Constants and types - is used across the whole codebase. And we agreed that - if something is used locally in one place, then it shoudl be inside module etc, but if necessary to use it in a few places, then we will move it outside to global places. I think it's not only about components, but also about such independent types and constants.

Sounds good 👍

bcotrim (approved):
Thanks for the adjustments, it looks good!
Tested the app and didn't find any regressions.
👍

PR #1523, opened 2025-07-01
Create dependabot.yml

wojtekn:
## Related issues

- Related to STU-496

## Proposed Changes

- I propose to create `dependabot.yml` file to enable automated dependency updates for npm packages.

## Testing Instructions

- N/A

## Pre-merge Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,26 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "npm"
+    # Look for `package.json` and `lock` files in the `root` directory
+    directory: "/"
+    # Check the npm registry for updates every day (weekdays)
+    schedule:
+      interval: "weekly"
+    allow:
+      - dependency-name: "@automattic/*"
+      - dependency-name: "@electron-forge/*"
+
+  # Enable version updates for GitHub Actions
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


bcotrim (approved):
LGTM
Not sure how to test it besides code review,  but this looks like a solid starting point. Thanks for adding it!
I would update the code comments, some don't match the values (weekly for example), and I think most are unnecessary as it is self explanatory.

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "npm"
+    # Look for `package.json` and `lock` files in the `root` directory
+    directory: "/"
+    # Check the npm registry for updates every day (weekdays)
```
katinthehatsite: Maybe I am misreading this but why does this comment says every day but the interval is set to `weekly`? Should the comment also say weekly? Or the schedule to be `daily` instead?

katinthehatsite (approved):
The changes look good to me 👍  I left one question about one of the comments in the file.

gcsecsey (approved):
LGTM and the timing is consistent now 👍

wojtekn:
Thanks for the reviews.

I haven't merged it because I was unsure if defining two packages for auto-updates in dependabot.yml will affect automatic PRs for security updates. I checked it more with the GH team and confirmed we are good here - Dependabot security updates will apply to all packages and will ignore the scope of packages configured in this file for that setting, while honoring the UX and access described in it.

It was not super clear in [Dependabot options reference](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference), however, now I know that some configuration options will influence security update PRs differently:

> In addition, all options marked with a  icon also change how Dependabot creates pull requests for security updates, except where target-branch is used.

The note with shield icon for `allow` config says:

> All dependencies defined in lock files with vulnerable dependencies are updated by security updates.

We should be good to go.

PR #1524, opened 2025-07-01
Force display the Selective Sync "What's New" modal for the patch version

sejas:
## Related issues

- Related to STU-555
- Fixes STU-592

## Proposed Changes

- Fixed a bug where the "What's New" modal was displayed after each Studio start instead of only when the patch version increased.
- Set the force new version patch to display the selective sync "What's New" modal.
- Updated tests to make them flexible based on patch condition.

## Testing Instructions

- Make sure the appdata has `"lastSeenVersion": "1.5.3",`
- Run `npm start`
- Observe the what's new opens and the `lastSeenVersion` is updated to `1.5.4`
- Kill the app, it doesn't matter if you close the modal or not.
- Run `npm start`
- Observe the modal doesn't appear for second time.

## Pre-merge Checklist


ivan-ottinger (approved):
The changes look good and work as expected: The modal opens up only - when forced with the constant. The `lastSeenVersion` value in `appdata-v1.jsoon` updates correctly.

Thank you for catching and fixing that bug, Antonio!

katinthehatsite (approved):
It worked as described. I did not see the modal appearing the second time after restarting the app 👍

wojtekn:
> Skipped tests that check the patch condition.

Should we either remove tests or fix them to account for the constant?

sejas:
Great question. I skipped the tests so that we can enable them again after we restore the default behavior in the next release. [STU-595](https://linear.app/a8c/issue/STU-595/studio-disable-whats-new-modal)

It's true that we could change the expected value and the test description to accommodate the temporal behavior.

sejas:
@wojtekn, I've updated the tests to match the current behaviour. Thanks for your suggestion: [https://github.com/Automattic/studio/pull/1524/commits/388c78b148d4c93518121080787ffcc2b4ddbb82](https://github.com/Automattic/studio/pull/1524/commits/388c78b148d4c93518121080787ffcc2b4ddbb82)

wojtekn:
Thanks @sejas . I meant whether we should treat constants as an integral part of the application and make tests aware of that. So they would account for the constant value.

sejas:
💡 Great idea! Thanks for clarifying it. I made the tests flexible so we won't need to update them when changing the constant value. [https://github.com/Automattic/studio/pull/1524/commits/6eb9136a03e4a2f20c686eff456287aa07576d7d](https://github.com/Automattic/studio/pull/1524/commits/6eb9136a03e4a2f20c686eff456287aa07576d7d)

PR #1532, opened 2025-07-03
Update Ukrainian - July 3

nightnei:
## Proposed Changes
Keep Ukrainian translation up to date.

## Testing Instructions
1. Run Studio locally
2. Turn on Ukrainian translation
3. Assert that it works


katinthehatsite (commented):
The translations work as expected but I am not sure these are real words 🤔 

Screenshot 2025-07-03 at 4 40 20 PM

Maybe we can use the same words as on the button, for example for `Push`?

katinthehatsite (approved):
I will approve to move the translations forward but I have a question/ concern about some wording. Not a big deal though, users might be able to understand

nightnei:
> I will approve to move the translations forward but I have a question/ concern about some wording. Not a big deal though, users might be able to understand

I was thinking a lot about the case (and before about similar cases)  - I have no idea what is teh best way 😅 "Запушити" is typical pronunciation for this action, what is used among developers, so I decided - maybe it's better to try to go with dev-slang words in Studio 🤔 Since "Завантажити" (and similar) - are very generic and not typical among developers.

I have thoughts about one day spending some time to reconsider namings in general and to choose the best way.

PR #1538, opened 2025-07-07
add delete token request to logout

bcotrim:
## Related issues

- Fixes STU-488

## Proposed Changes

- Add revoke token call during logout

## Testing Instructions

- check 187112-ghe-Automattic/wpcom
- Run Studio and log in to WordPress.com
- Cut your internet connection
- Confirm logout button is disabled
- Restore your internet connection
- Confirm logout button is enabled
- Logout and confirm the request was made to revoke the token
- Using the token from Studio, try to make a request to WordPress.com API (you can just copy/paste the auth info from appdata from before you logged out)

## Pre-merge Checklist

### DIFF
--- a/src/components/auth-provider.tsx
+++ b/src/components/auth-provider.tsx
@@ -66,6 +66,36 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 
 	const logout = useCallback( async () => {
 		try {
+			// Get the token before clearing authentication
+			const token = await getIpcApi().getAuthenticationToken();
+
+			if ( token?.accessToken ) {
+				try {
+					// Call the token revocation endpoint directly using fetch
+					// client.req.del is not working, so we need to use fetch
+					const response = await fetch(
+						'https://public-api.wordpress.com/wpcom/v2/studio-app/token',
+						{
+							method: 'DELETE',
+							headers: {
+								Authorization: `Bearer ${ token.accessToken }`,
+								'Content-Type': 'application/json',
+							},
+							signal: AbortSignal.timeout( 5000 ),
+						}
+					);
+
+					if ( ! response.ok ) {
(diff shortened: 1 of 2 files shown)


src/components/auth-provider.tsx
```diff
@@ -66,6 +66,36 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
...
+								Authorization: `Bearer ${ token.accessToken }`,
+								'Content-Type': 'application/json',
+							},
+							signal: AbortSignal.timeout( 5000 ),
+						}
+					);
+
+					if ( ! response.ok ) {
+						console.error( 'Failed to revoke token:', response.status, response.statusText );
+					}
```
sejas: ha! it seems the wpcom req.delete sends the POST method instead. In theory if you pass the method as param it should override it, but I couldn't make it work. See the [code](https://github.com/Automattic/wp-calypso/blob/trunk/packages/wpcom.js/src/lib/util/request.js#L60-L78).

An alternative if we want to avoid passing the bearer token is using client.request instead which I tried it and it works. Something like:

```ts
					client?.request(
						{
							apiNamespace: 'wpcom/v2',
							method: 'DELETE',
							path: '/studio-app/token',
						},
						( err, response ) => {
							if ( err ) {
								console.error( 'Failed to revoke token:', err );
							}
							console.log( 'Token revoked:', response );
						}
					);
```

sejas (approved):
I sandboxed the API and tried logging out. I confirm the token is revoked and it works as expected:

https://github.com/user-attachments/assets/2c474b4e-3599-4ee6-8605-00c91fc4065e

src/components/auth-provider.tsx
bcotrim: I also couldn't overwrite the method.
But using `client.request` is a great approach, thanks @sejas

PR #1539, opened 2025-07-08
Fix issue with unchecking empty folder

bcotrim:
## Related issues

- Fixes STU-607

## Proposed Changes

- Fix an issue where it wasn't possible to uncheck an empty folder

## Testing Instructions

- Open Studio, log-in to WordPress.com and connect an existing or new site to a WordPress.com site
- Empty one of site folders like `wp-content/uploads`
- Select option to "Push" changes
- Confirm you can check/uncheck the empty folder

## Pre-merge Checklist

### DIFF
--- a/src/components/tree-view.tsx
+++ b/src/components/tree-view.tsx
@@ -143,7 +143,7 @@ const TreeItem = ( {
 					className={ cx( 'ps-6', isLevel0 ? 'border-b border-gray-300 py-2' : '' ) }
 				>
 					{ node.children.length === 0 ? (
-						<div className="py-2 ps-6 text-gray-500 italic" aria-label={ __( 'Empty folder' ) }>
+						<div className="text-gray-500 italic" aria-label={ __( 'Empty folder' ) }>
 							{ __( 'Empty' ) }
 						</div>
 					) : (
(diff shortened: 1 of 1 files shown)


sejas (approved):
Thanks for catching this issue and fixing it ! 👌 

Do you think we should remove the chevron if the folder doesn't have any files?

Screenshot 2025-07-08 at 12 13 48


**Before** 

https://github.com/user-attachments/assets/6616d10c-b32e-4580-b510-89ec31bda408


**After**

https://github.com/user-attachments/assets/15a5cc26-a4c0-4805-8511-a84f48428559

bcotrim:
> Do you think we should remove the chevron if the folder doesn't have any files?

I was thinking about it but couldn't find a perfect solution. I considered not showing the chevron, or even the folder but I think that makes it unclear to the user that the folder is empty.
What do you think about keep the chevron but disabling the checkbox?

nightnei:
@bcotrim Good catch 👍

@bcotrim @sejas WDYT about keeping chevron but when it's expanded then show text "Empty". It's universal, short and clear.
Or maybe something similar, like "No items".

sejas:
Yeah, probably displaying a message like "Empty" or "No files found" when expanding the folder is the best solution.

bcotrim:
@sejas @nightnei thanks for the feedback!
I've added the "Empty" message.
Can you give it a quick review to see if the style looks good?

sejas (approved):
Works great! thank you!

Screenshot 2025-07-09 at 11 55 45

PR #1544, opened 2025-07-08
Studio Sync: add more granular selection to wp-content folders

bcotrim:
## Related issues

- Fixes STU-584

## Proposed Changes

- Adds `mu-plugins` and `fonts` to the sync granular selection

## Testing Instructions

- start Studio
- using a connected site, start a push operation
- `mu-plugins` and `fonts` folders and their content appear on the tree
- Selection works as expected

## Pre-merge Checklist

### DIFF
--- a/src/modules/sync/components/sync-dialog.tsx
+++ b/src/modules/sync/components/sync-dialog.tsx
@@ -81,7 +81,7 @@ export function SyncDialog( {
 	const locale = useI18nLocale();
 	const { __ } = useI18n();
 	const copy = useSyncDialogTexts( type );
-	const defaultTree = useDefaultSyncTree();
+	const defaultTree = useDefaultSyncTree( type );
 
 	const [ showAllFiles, setShowAllFiles ] = useState( false );
 	const [ treeState, setTreeState ] = useState< TreeNode[] >( defaultTree );
--- a/src/modules/sync/components/sync-dialog.tsx
+++ b/src/modules/sync/components/sync-dialog.tsx
@@ -27,6 +27,10 @@ type SyncDialogProps = {
 	onRequestClose: () => void;
 };
 
+const wpTypeIsFolder = ( wpType: string ): wpType is keyof typeof SYNC_OPTIONS => {
+	return Object.keys( SYNC_OPTIONS ).includes( wpType );
+};
+
 const useDynamicTreeState = (
 	type: 'push' | 'pull',
 	localSiteId: string,
@@ -55,7 +59,8 @@ const useDynamicTreeState = (
 							type: item.type,
 					  } ) );
 
-				newState = updateNodeById( newState, SYNC_OPTIONS[ wpType ], {
+				const nodeId = wpTypeIsFolder( wpType ) ? SYNC_OPTIONS[ wpType ] : wpType;
+				newState = updateNodeById( newState, nodeId, {
 					loading: isLoading,
 					children,
 				} );
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -56,7 +56,7 @@ export const SYNC_OPTIONS = {
 } as const;
 
 export const CONTENTS_SYNC_OPTIONS = {
-	muPlugins: 'mu-plugins',
+	'mu-plugins': 'mu-plugins',
 	fonts: 'fonts',
 } as const;
 
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -55,6 +55,11 @@ export const SYNC_OPTIONS = {
 	contents: 'contents',
 } as const;
 
+export const CONTENTS_SYNC_OPTIONS = {
+	muPlugins: 'mu-plugins',
+	fonts: 'fonts',
+} as const;
+
 // AI Assistant constants
 // IMPORTANT: When updating this value, we need to update the string located in `AIClearHistoryReminder` component.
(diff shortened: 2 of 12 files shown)


sejas (commented):
Let's hold off on merging this PR.  
I think adding mu-plugins and fonts is an improvement, but we could build the structure dynamically.  
Not all sites will have those folders, so we shouldn't hard-code those options. For example, classic themes won't have the fonts folder, and if the user is using MySQL, then they probably won't have mu-plugins either.  
If we build the tree dynamically, we can expand and display to the user any other file or folder inside the wp-content folder, and remove the `Other files and folders` option.

What do you think?


https://github.com/user-attachments/assets/7e0d26ce-4449-41bf-8602-6d70e70f87e6

bcotrim:
> I think adding mu-plugins and fonts is an improvement, but we could build the structure dynamically.
Not all sites will have those folders, so we shouldn't hard-code those options. For example, classic themes won't have the fonts folder, and if the user is using MySQL, then they probably won't have mu-plugins either.

This makes sense, we could easily check if the folders exist before adding them.

> If we build the tree dynamically, we can expand and display to the user any other file or folder inside the wp-content folder, and remove the Other files and folders option.

This was my initial idea, but then I noticed we don't actually export anything else inside `wp-content` (even if `all` files selected). Example: 

| Site folder | Exported |
|--------|--------|
| image | image | 

So we could in fact remove "Other files and folders" with this PR, or make some changes to our exporter to pick up everything.

sejas:
That's a good point. I think we should expand the export logic to include any file in wp-content by default. Maybe not as part of this issue though.<br>@wojtekn, What do you think?

wojtekn:
@sejas does it mean that currently, when "Other files and folders" is selected, Studio sets "contents" option of Jetpack restore, but doesn't include those directories and files in the archive?

sejas:
> Studio sets "contents" option of Jetpack restore, but doesn't include those directories and files in the archive?

Studio only includes [mu-plugins and fonts](https://github.com/Automattic/studio/blob/685f79a2e8e0234a1e8d5f5a713690e69af7aaed/src/lib/import-export/export/exporters/default-exporter.ts#L155-L156) as part of other files.<br>I wonder if we should include this PR and remove the `other files and folders` option. It would make the pull and push a bit more asymmetric but more accurate. In the next releases we can expand the push selection to all the other files and folders from the file system.

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -27,6 +27,10 @@ type SyncDialogProps = {
 	onRequestClose: () => void;
 };
 
+const wpTypeIsFolder = ( wpType: string ): wpType is keyof typeof SYNC_OPTIONS => {
+	return Object.keys( SYNC_OPTIONS ).includes( wpType );
+};
```
sejas, line 32: Is this function the same as [`isSyncOption`](https://github.com/Automattic/studio/pull/1544/files#diff-7b84b01e7fea0fc33c7baef3c96e620ac428988aed26df1052799bbfd2825e9cR16-R18)?

```
const isSyncOption = ( value: string ): value is SyncOption => {
	return Object.keys( SYNC_OPTIONS ).includes( value );
};
```

wojtekn:
> I wonder if we should include this PR and remove the other files and folders option. It would make the pull and push a bit more asymmetric but more accurate. In the next releases we can expand the push selection to all the other files and folders from the file system.

What if we dropped "Other files and directories" from Push completely then, and as a follow-up, worked on removing this option for Pull, replacing it with either all files and dirs available in filesystem, or `mu-plugins` and `fonts` options?

sejas:
It makes sense to me. Just to clarify. We'll leave the option in the Pull, and remove it in the Push.

sejas:
I created this issue STU-618.

katinthehatsite:
Not sure if this PR is in review or not as I am seeing some discussions from a couple of past months, but it is showing for me in review column so I decided to take a look 😅 

I read through everything and I am still unsure why would we want to include `mu-plugins` either in Push or Pull? `Fonts` of course makes sense but what is the reason the user would want to sync `mu-plugins`? Perhaps I am missing something obvious?

But then I also see this task was completed: STU-618 so I am unsure if we are still pursuing this PR?

bcotrim:
> Not sure if this PR is in review or not as I am seeing some discussions from a couple of past months, but it is showing for me in review column so I decided to take a look 😅
> 
> I read through everything and I am still unsure why would we want to include `mu-plugins` either in Push or Pull? `Fonts` of course makes sense but what is the reason the user would want to sync `mu-plugins`? Perhaps I am missing something obvious?
> 
> But then I also see this task was completed: STU-618 so I am unsure if we are still pursuing this PR?

Thanks for the comment and reminder @katinthehatsite 
I added STU-756 for us to follow-up on the issue. I am closing this PR based on the new issue.

wojtekn:
> I read through everything and I am still unsure why would we want to include mu-plugins either in Push or Pull? Fonts of course makes sense but what is the reason the user would want to sync mu-plugins? Perhaps I am missing something obvious?

@katinthehatsite see https://developer.wordpress.org/advanced-administration/plugins/mu-plugins/ - plugins from that directory can be used for different purposes. We use `mu-plugins/` for different Studio polyfills, WP Cloud and any other host can use it for their code, but most importantly, developer can also use it for their own required code and expect it will be available on each environment.

PR #1552, opened 2025-07-10
Remove other files and folders option from push

bcotrim:
## Related issues

- Fixes STU-618

## Proposed Changes

- Remove "other files and folders" from push operation

| Pull | Push |
|--------|--------|
| image | image | 

## Testing Instructions

- Open Studio and a site connected to WordPress.com
- Navigate to Sync and open the pull dialog, confirm the "other files and folders" exists and can be selected
- Complete the pull without any errors
- Navigate to Sync and open the push dialog, confirm the "other files and folders" does not exist.
- Complete the push and confirm that only the selected files are being pushed, unless "All files and folders" is selected

## Pre-merge Checklist

### DIFF
--- a/src/modules/sync/hooks/use-default-sync-tree.tsx
+++ b/src/modules/sync/hooks/use-default-sync-tree.tsx
@@ -3,7 +3,7 @@ import { useMemo } from 'react';
 import { TreeNode } from 'src/components/tree-view';
 import { SYNC_OPTIONS } from 'src/constants';
 
-export const useDefaultSyncTree = (): TreeNode[] => {
+export const useDefaultSyncTree = ( type: 'push' | 'pull' ): TreeNode[] => {
 	const { __ } = useI18n();
 
 	return useMemo( () => {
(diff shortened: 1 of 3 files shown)


src/modules/sync/hooks/use-default-sync-tree.tsx
```diff
@@ -3,7 +3,7 @@ import { useMemo } from 'react';
 import { TreeNode } from 'src/components/tree-view';
 import { SYNC_OPTIONS } from 'src/constants';
 
-export const useDefaultSyncTree = (): TreeNode[] => {
+export const useDefaultSyncTree = ( type: 'push' | 'pull' ): TreeNode[] => {
```
sejas, line 6: We need to add type to the useMemo dependencies array on line 74 `}, [ __, type ] );`

sejas (approved):
Works as expected. The Other files and folders option is not visible to users in the push dialog.
This is a temporal fix until we implement STU-584
Users won't be able to sync mu-plugins and fonts in the meantime unless they do a full sync by selecting Files and Database together.
Screenshot 2025-07-10 at 20 41 11

src/modules/sync/hooks/use-default-sync-tree.tsx
bcotrim: Nice catch! Thanks for fixing it

sejas: Thanks for creating the PR and your fast help.

PR #1566, opened 2025-07-14
add blueprints feature flag

bcotrim:
## Related issues

- Fixes STU-621

## Proposed Changes

- Adds feature flag for blueprints project

image

## Testing Instructions

- Code review
- Run `npm start` and `ENABLE_BLUEPRINTS=true npm start`
- Both should work and set the correct value for the feature flag (you can confirm on the app menu as shown above)

## Pre-merge Checklist

### DIFF
--- a/src/ipc-types.d.ts
+++ b/src/ipc-types.d.ts
@@ -84,7 +84,12 @@ type IpcApi = {
 	getPathForFile: ( file: File ) => string;
 };
 
-interface AppGlobals {
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+interface FeatureFlags {
+	enableBlueprints: boolean;
+}
+
+interface AppGlobals extends FeatureFlags {
 	platform: NodeJS.Platform;
 	appName: string;
 	appVersion: string;
--- a/src/lib/feature-flags.ts
+++ b/src/lib/feature-flags.ts
@@ -5,7 +5,7 @@ export interface FeatureFlagDefinition {
 	default: boolean;
 }
 
-export const FEATURE_FLAGS_DEFINITION = {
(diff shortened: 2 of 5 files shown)


src/lib/feature-flags.ts
```diff
@@ -5,10 +5,22 @@ export interface FeatureFlagDefinition {
 	default: boolean;
 }
 
-// eslint-disable-next-line @typescript-eslint/no-empty-object-type
-export interface FeatureFlags {}
+export const FEATURE_FLAGS_DEFINITION = {
```
bcotrim: Moving forward this and the ipc-types should be the only changes we need to add or remove feature flags

gcsecsey (approved):
Thanks for improving the feature flag handling @bcotrim! This looks good to me and is working great. :shipit: 

Image

src/ipc-types.d.ts
```diff
@@ -84,7 +84,12 @@ type IpcApi = {
 	getPathForFile: ( file: File ) => string;
 };
 
-interface AppGlobals {
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
```
sejas, line 87: Should we remove this eslint directive?
```suggestion
```

sejas (approved):
Works as expected.

**Tested `npm start`**

Screenshot 2025-07-14 at 17 55 07


**Tested `ENABLE_BLUEPRINTS=true npm start`**

Screenshot 2025-07-14 at 17 56 21

src/ipc-types.d.ts
bcotrim: I was considering keeping it, so we don't need to add/remove when the feature flags object is empty.
What do you think?

PR #1568, opened 2025-07-14
WordPress providers - wp-now implementation

bcotrim:
## Related issues

- Fixes STU-629

## Proposed Changes

The goal of this PR is to do some preparation for the development of Playground CLI integration with Studio.
By adding this provider layer we can:
1. Switch provider based on feature flag - good during development and opens up for a quick rollback if we need to
2. Easier to identify and implement Playground CLI operations needed to replace `wp-now`

Summary of changes:
  - Created WordPress provider abstraction layer to support multiple WordPress implementations
  - Refactored all `wp-now` dependencies to use provider interfaces in main process code
  - Updated child processes to use environment-based provider switching without Electron dependencies

## Testing Instructions

- Start Studio
- Please test as many operations as you can (create site, start, stop, import, preview, export, sync, ... )

## Pre-merge Checklist


vendor/wp-now/src/execute-wp-cli.ts
```diff
@@ -4,7 +4,7 @@ import { rootCertificates } from 'tls';
 import { createNodeFsMountHandler, loadNodeRuntime } from '@php-wasm/node';
 import { PHP, MountHandler, writeFiles, setPhpIniEntries } from '@php-wasm/universal';
 import { phpVar } from '@php-wasm/util';
-import { getSqliteCommandPath } from '../../../src/lib/sqlite-command-versions';
+import { getSqliteCommandPath } from '../../../src/lib/sqlite-command-path';
```
sejas:
```suggestion
import { getSqliteCommandPath } from '/src/lib/sqlite-command-path';
```

src/lib/sqlite-command-path.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+import os from 'os';
+import path from 'path';
+import fs from 'fs-extra';
+
+/**
+ * Get the path for SQLite command files
+ * This is a simplified version that doesn't depend on other modules
+ */
+export function getSqliteCommandPath() {
```
sejas, line 13: I see a function with the same name was declared in 
https://github.com/Automattic/studio/blob/5d320f02faac9712b9125d0b2da99c4695f45cd1/src/lib/sqlite-command-versions.ts#L21

Should we remove the old one?

src/ipc-handlers.ts
```diff
@@ -1426,3 +1430,26 @@ export async function listWpContentFolders(
...
+	defaultPhpVersion: string;
+	defaultWordPressVersion: string;
+	allowedPhpVersions: string[];
+} > {
+	return {
+		defaultPhpVersion: DEFAULT_PHP_VERSION(),
+		defaultWordPressVersion: DEFAULT_WORDPRESS_VERSION(),
+		allowedPhpVersions: ALLOWED_PHP_VERSIONS(),
+	};
+}
```
sejas: I couldn't find where this IPC function is used, except in tests. 🤔

src/ipc-handlers.ts
```diff
@@ -1426,3 +1430,26 @@ export async function listWpContentFolders(
...
+export function notifyProviderConstantsChanged( _event: IpcMainInvokeEvent ): void {
+	// Send updated constants to all renderer processes
+	const constants = {
+		defaultPhpVersion: DEFAULT_PHP_VERSION(),
+		defaultWordPressVersion: DEFAULT_WORDPRESS_VERSION(),
+		allowedPhpVersions: ALLOWED_PHP_VERSIONS(),
+	};
+
+	void sendIpcEventToRenderer( 'providerConstantsChanged', constants );
+}
```
sejas: The same goes for this function. I don't see it called anywhere. I would expect to see `getIpcApi().notifyProviderConstantsChanged` used somewhere.

src/lib/wordpress-provider/index.ts
```diff
@@ -0,0 +1,83 @@
...
+		// Ignore errors in case this is called from renderer process
+	}
+}
+
+// Constants as functions to allow runtime provider switching
+export const DEFAULT_PHP_VERSION = () => getWordPressProvider().DEFAULT_PHP_VERSION;
+export const DEFAULT_WORDPRESS_VERSION = () => getWordPressProvider().DEFAULT_WORDPRESS_VERSION;
+export const ALLOWED_PHP_VERSIONS = () => getWordPressProvider().ALLOWED_PHP_VERSIONS;
+export const SQLITE_FILENAME = () => getWordPressProvider().SQLITE_FILENAME;
+export const SQLITE_FILENAME_LEGACY = () => getWordPressProvider().SQLITE_FILENAME_LEGACY;
```
sejas: I'm not sure if we need these intermediary functions. I think we could use `getWordPressProvider` directly.
Let's avoid using capital letter variables for functions. We can use regular camelCase when we access the properties. 
So for example in `src/ipc-handlers.ts` we could call `getWordPressProvider().defaultPHPVersion`. What do you think?

```suggestion
```

src/lib/wordpress-provider/index.ts
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	// Notify renderer processes about provider constants changes
+	try {
+		const { sendIpcEventToRenderer } = require( 'src/ipc-utils' );
+		const constants = {
+			defaultPhpVersion: DEFAULT_PHP_VERSION(),
+			defaultWordPressVersion: DEFAULT_WORDPRESS_VERSION(),
+			allowedPhpVersions: ALLOWED_PHP_VERSIONS(),
+		};
```
sejas: Using the provider can simplify the logic:

```suggestion
		const constants = {
			defaultPhpVersion: provider.DEFAULT_PHP_VERSION,
			defaultWordPressVersion: provider.DEFAULT_WORDPRESS_VERSION,
			allowedPhpVersions: provider.ALLOWED_PHP_VERSIONS,
		};
```

vendor/wp-now/src/download.ts
```diff
@@ -4,7 +4,31 @@ import path from 'path';
...
+		return 'https://wordpress.org/nightly-builds/wordpress-latest.zip';
+	}
+
+	if ( ! isValidWordPressVersion( version ) ) {
+		throw new Error(
+			'Unrecognized WordPress version. Please use "latest" or numeric versions such as "6.2", "6.0.1", "6.2-beta1", or "6.2-RC1"'
+		);
+	}
+	return `https://wordpress.org/wordpress-${ version }.zip`;
+}
```
sejas: I see this code is inserted between imports. Let's move them after the imports.

src/lib/wordpress-version-utils.ts
```diff
@@ -1,5 +1,24 @@
-import { DEFAULT_WORDPRESS_VERSION } from 'vendor/wp-now/src/constants';
-import { isValidWordPressVersion } from 'vendor/wp-now/src/wp-playground-wordpress/is-valid-wordpress-version';
+const DEFAULT_WORDPRESS_VERSION = 'latest';
```
sejas: Let's remove this constant. We can pass the word `latest` in `getWordPressVersionUrl( version = 'latest' ) {`
```suggestion
```

Later we can also remove that default value.

sejas (commented):
@bcotrim, this looks great! It will allow us to work on the migration to Playground CLI without breaking any functionality ✨.

I tested Studio's main functionality and everything was working as expected. I was able to create, start, stop, and delete sites; create preview sites; import Jetpack backups; upload media; and change PHP and WP versions. Execute wp-cli command in the Assistant.

I left a few suggestions. My major concern is a couple of functions declared in IPC that I don't see being called anywhere.

src/lib/wordpress-version-utils.ts
sejas: - I created this short PR to fix  simplify this part https://github.com/Automattic/studio/pull/1582

src/lib/wordpress-provider/index.ts
wojtekn: > So for example in `src/ipc-handlers.ts` we could call `getWordPressProvider().defaultPHPVersion`. What do you think?

Actually, it should be `defaultPhpVersion` if we use camelCase.

src/lib/sqlite-command-path.ts
bcotrim: I had to create this minimal function to avoid import issues.

src/ipc-handlers.ts
bcotrim: Thanks for catching that, it was some leftover from different approaches I tried.

src/ipc-handlers.ts
bcotrim: I opted to call the `sendIpcEventToRenderer( 'providerConstantsChanged', constants );` directly in the provider, so we don't need this function. Thanks for catching it!

bcotrim:
@sejas @wojtekn thanks for the reviews!
I think I addressed all feedback, can I ask you to take another look, please? 🙏

src/site-server.ts
```diff
@@ -167,25 +130,17 @@ export class SiteServer {
...
-			php: this.details.phpVersion,
-			wp: this.meta.wpVersion,
+			phpVersion: this.details.phpVersion,
+			wpVersion: this.meta.wpVersion,
 			isWpAutoUpdating: this.details.isWpAutoUpdating,
+			absoluteUrl: getAbsoluteUrl( this.details ),
 		} );
 
-		options.absoluteUrl = getAbsoluteUrl( this.details );
-		options.siteLanguage = await getPreferredSiteLanguage( options.wordPressVersion );
```
sejas, line 181: Ha! I thought we forgot to pass this to startServer, but I can confirm it's not necessary. I tested creating a new site in Spanish and it worked as expected.

src/modules/preview-site/lib/version-comparison.ts
```diff
@@ -1,5 +1,5 @@
 import semver from 'semver';
-import { DEFAULT_PHP_VERSION } from 'vendor/wp-now/src/constants';
+import { DEFAULT_PHP_VERSION } from 'src/lib/wordpress-provider/constants';
```
sejas, line 2: We could consider using this constant from the provider, if possible.

src/stores/provider-constants-slice.ts
```diff
@@ -0,0 +1,38 @@
...
+	defaultWordPressVersion: string;
+	allowedPhpVersions: string[];
+}
+
+const initialState: ProviderConstantsState = {
+	// Set reasonable defaults until provider constants are loaded
+	defaultPhpVersion: '8.3',
+	defaultWordPressVersion: 'latest',
+	allowedPhpVersions: [ '8.4', '8.3', '8.2', '8.1', '8.0', '7.4' ],
+};
```
sejas: I think this initial state could use more "empty" values, as they will be overwritten when the `providerConstantsChanged` event is called.

src/constants.ts
```diff
@@ -76,6 +76,7 @@ export const IPC_VOID_HANDLERS = < const >[
 	'addSyncOperation',
 	'clearSyncOperation',
 	'logRendererMessage',
+	'notifyProviderConstantsChanged',
```
sejas: I think we can remove this constant. 

```suggestion
```

sejas (approved):
Thanks for the updates. I tested Studio creating sites, starting, stopping, exporting, importing. All worked great!

I noticed we are importing `DEFAULT_WORDPRESS_VERSION` directly from `src/lib/wordpress-provider/constants` in some tests and in `src/modules/preview-site/lib/version-comparison.ts`. Should we use the provider instead?

I wonder if we need to have the constants in the provider. I think having the ability to hot swap providers is just a nice-to-have and not a necessary goal. Requiring to restart Studio to swap providers is a reasonable expectation.

src/modules/preview-site/lib/version-comparison.ts
bcotrim: I think using the provider here will throw an error since this function is used in the renderer.

bcotrim:
> I wonder if we need to have the constants in the provider. I think having the ability to hot swap providers is just a nice-to-have and not a necessary goal. Requiring to restart Studio to swap providers is a reasonable expectation.

My main goal in allow hot swap is for us to be able to control the provider via API.
That opens up the possibility for a quick revert once the feature goes live if we start seeing some errors.

bcotrim:
@sejas I added some minor changes, please let me know if you have any additional feedback and I'll add a follow-up pr

src/lib/wordpress-provider/constants.ts
```diff
@@ -0,0 +1,8 @@
+export {
```
youknowriad, line 1: I don't really understand why this file exists? Why aren't we importing directly from the `vendor/wp-now/src/constants` where needed?

youknowriad: Ok I guess it's prep work for something else.

bcotrim: > Ok I guess it's prep work for something else.

Right on point! You can check the follow-up PR with the Playground CLI implementation: https://github.com/Automattic/studio/pull/1594

Feedback on either implementation is welcomed, I'm happy to follow up with a new PR. Note that we plan to remove the `wp-now` code once we feel the Playground CLI implementation is solid and we are confident in it.

PR #1579, opened 2025-07-17
Suppress punycode warnings for both CLI and the app

wojtekn:
## Related issues

- Fixes STU-639

## Proposed Changes

- I propose to suppress punycode warnings for both the CLI and the app. It doesn't cause any problems for the app, but it shows a confusing warning as part of the command output.

|**Before**|**After**|
|---|---|
|Screenshot 2025-07-17 at 14 23 51|Screenshot 2025-07-17 at 14 23 12|

Installing punycode as an explicit dependency didn't help - there is some dependency that loads deprecated punycode. 

## Testing Instructions

1. Rebuild command:
```
npm run cli:watch
```
3. Run command
```
node dist/cli/main.js
```
4. Confirm there is no punycode warning.

## Pre-merge Checklist

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1,5 +1,6 @@
 import path from 'node:path';
 import { __ } from '@wordpress/i18n';
+import { suppressPunycodeWarning } from 'common/lib/suppress-punycode-warning';
 import { StatsGroup, StatsMetric } from 'common/types/stats';
 import yargs from 'yargs';
 import { registerCommand as registerCreateCommand } from 'cli/commands/preview/create';
@@ -10,6 +11,8 @@ import { loadTranslations } from 'cli/lib/i18n';
 import { bumpAggregatedUniqueStat } from 'cli/lib/stats';
 import { version } from 'cli/package.json';
 import { StudioArgv } from 'cli/types';
+
+suppressPunycodeWarning();
 
 async function main() {
 	const locale = await loadTranslations();
--- a/common/lib/suppress-punycode-warning.ts
+++ b/common/lib/suppress-punycode-warning.ts
@@ -1,0 +1,19 @@
+export function suppressPunycodeWarning() {
+	// Save existing listeners
+	const originalListeners = process.listeners( 'warning' );
+
+	// Remove all current listeners
+	process.removeAllListeners( 'warning' );
+
+	// Add custom listener with filter
+	process.on( 'warning', ( warning: Error & { code?: string } ) => {
+		// Suppress only the punycode deprecation warning
+		if ( warning.name === 'DeprecationWarning' && warning.code === 'DEP0040' ) {
+			return;
+		}
+		// Otherwise, call original listeners (print warning as usual)
+		for ( const listener of originalListeners ) {
+			listener.call( process, warning );
+		}
+	} );
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,6 +17,7 @@ import {
 	REDUX_DEVTOOLS,
 } from 'electron-devtools-installer';
 import { PROTOCOL_PREFIX } from 'common/constants';
+import { suppressPunycodeWarning } from 'common/lib/suppress-punycode-warning';
 import { StatsGroup } from 'common/types/stats';
 import { IPC_VOID_HANDLERS } from 'src/constants';
 import * as ipcHandlers from 'src/ipc-handlers';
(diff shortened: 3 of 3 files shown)


gcsecsey (approved):
Changes LGTM and I can confirm that the warnings are suppressed. 👍 

| trunk | this branch |
|--------|--------|
| Image | Image |

sejas (commented):
UPDATE: It worked after repeating my steps xD.

~~I don't see the warning on trunk 😅 . Do I need to follow any specific steps?~~

~~I run:~~
- ~~`nvm use && npm install`~~
- ~~`npm run cli:watch`~~
- ~~`node dist/cli/main.js`~~

~~And here is the output:~~

```bash
❯ node dist/cli/main.js
WordPress Studio CLI

Commands:
  studio preview  Manage preview sites

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
  --path     Path to the WordPress files   [string] [default: Current directory]

You must provide a valid command
```

wojtekn:
@sejas thanks for confirming. I tried again, and reproduced.

sejas:
yep, all good, I was able to reproduce it.

sejas (approved):
I confirm this PR hides the punycode warning with code `DEP0040`.
It would be interesting to track an issue if upgrading some dependencies can fix the issue in the future.

**Before:**

```
❯ node dist/cli/main.js  --help
(node:31152) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
WordPress Studio CLI

Commands:
  studio preview  Manage preview sites

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
  --path     Path to the WordPress files   [string] [default: Current directory]

You must provide a valid command
```

**After**

```
❯ node dist/cli/main.js --help
WordPress Studio CLI

Commands:
  studio preview  Manage preview sites

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
  --path     Path to the WordPress files   [string] [default: Current directory]
`

**Comment by sejas** [comment]
Here are the dependencies that use punycode.

```
❯ npm ls punycode
./studio
├─┬ @wp-playground/blueprints@1.1.0
│ └─┬ ajv@8.12.0
│   └─┬ uri-js@4.4.1
│     └── punycode@2.3.1
└─┬ jest-environment-jsdom@29.7.0
  └─┬ jsdom@20.0.3
    ├─┬ tough-cookie@4.1.3
    │ └── punycode@2.3.1 deduped
    └─┬ whatwg-url@11.0.0
      └─┬ tr46@3.0.0
        └── punycode@2.3.1 deduped
```

I suspect that I didn't see the error at first because I was experimenting with the latest playground dependencies. We can revisit if we can remove this custom code after we update to Blueprints 1.2.3+.

**Comment by wojtekn** [comment]
@sejas the trick is that the warning doesn't come from NPM dependencies you showed in the `npm ls` output. The Punycode 2.3.1 module installed as an NPM dependency is functioning correctly.

The warning thrown from `punycode`, which is part of Node, and used by one of our dependencies, possibly this one:

```
% npm ls tr46                                              
studio@1.5.5 /Users/cyphelf/Sites/studio
├─┬ jest-environment-jsdom@29.7.0
│ └─┬ jsdom@20.0.3
│   └─┬ whatwg-url@11.0.0
│     └── tr46@3.0.0
└─┬ node-fetch@2.7.0
  └─┬ whatwg-url@5.0.0
    └── tr46@0.0.3
```

We could pursue it further, but even if we fix it, it can resurface if any other dependency starts using Node's punycode.

**Comment by sejas** [comment]
Yeah, I think it makes sense to merge this PR. But I also think that from time to time, if we upgrade a dependency that includes punycode, we should check if this warning is gone. My feeling is that `@wp-playground/blueprints` is the one pulling that dependency and causing the warning, but you're right that it could be tr46 or another package.

**Comment by epeicher** [comment]
@sejas, @wojtekn, also there is the `npm why` command that can be used to identify the chain of dependencies causing a package to be installed: https://docs.npmjs.com/cli/commands/npm-explain. 

For example, `npm why punycode`. The problem is that the output is a bit verbose.

Additionally, we can temporarily use `yarn && yarn why punycode` and the output is more concise:

CleanShot 2025-07-23 at 09 13 49@2x

Then delete the `yarn.lock` afterwards to clean leftovers.

**Comment by wojtekn** [comment]
@epeicher thanks for sharing those steps. 

As I mentioned above, it's not really about an NPM dependency that uses punycode as NPM dependency. It's about some dependency that loads NPM's punycode module which is deprecated since Node 22 - as far as I see, at least `tr46` package does that:

```
require("punycode")
```

See [the StackOverflow thread](https://stackoverflow.com/a/79371801/1067873).

PR #1594, opened 2025-07-21
Playground CLI Provider

bcotrim:
## Related issues

- Fixes STU-628 STU-630 STU-631 STU-632 STU-634 STU-635 STU-636 

## Proposed Changes

- Added Playground CLI Provider - New WordPress provider using `@wp-playground/cli@2.0.7`
- Implemented all server related operations
- Created Studio MU-plugins system - standard plugins + Studio-specific loader for plugin loading
- Updated build configuration - Webpack changes to support playground CLI child process

## Testing Instructions

- create a new site or use an existing one
- Start / Stop
- Edit the Site
- Create a new site with advanced settings
- Create a preview
- Sync the site with a remote site, and push / pull changes
- Import and Export
- Delete Site

Notes: 
- Please test as many different sites as you can (imported from files, synced from WordPress.com/Pressable)
- If you can test all the operations for sites created with `wp-now` and `playground-cli` providers
- If this PR https://github.com/Automattic/studio/pull/1630 hasn't been merged you need to change this condition https://github.com/Automattic/studio/blob/trunk/src/modules/add-site/index.tsx#L47 to always use the legacy modal

## Pre-merge Checklist


package.json
```diff
@@ -130,6 +130,7 @@
 		"@wordpress/i18n": "^5.26.0",
 		"@wordpress/icons": "^10.10.0",
 		"@wp-playground/blueprints": "^1.1.0",
+		"@wp-playground/cli": "^1.2.3",
```
sejas: Should we update all the playground dependencies together?

* @php-wasm/node
* @php-wasm/scopes
* @php-wasm/universal
* @wp-playground/blueprints
* @wp-playground/wordpress

```suggestion
		"@wp-playground/blueprints": "^1.2.3",
		"@wp-playground/cli": "^1.2.3",
```

bcotrim: Done!

bcotrim:
This PR should be ready for a final review after https://github.com/WordPress/wordpress-playground/pull/2440 is merged and we update the dependencies.

bcotrim:
@sejas I think this is finally ready for a final review.
I updated the Playground CLI versions, retested on macOS and Windows.
Can I ask you for another review? 🙏

src/lib/wordpress-provider/playground-cli/playground-server-process-child.ts
```diff
@@ -0,0 +1,259 @@
...
+
+async function stopServerFunc(): Promise< void > {
+	if ( ! server ) {
+		return;
+	}
+
+	try {
+		await server[ Symbol.asyncDispose ]();
+	} catch ( error ) {
+		// Don't re-throw the error, just continue cleanup
```
sejas: Should we at least console the error ?

src/lib/wordpress-provider/playground-cli/playground-server-process-child.ts
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// Intercept and prefix all console output from playground-cli
+const originalConsoleLog = console.log;
+const originalConsoleError = console.error;
+const originalConsoleWarn = console.warn;
+
+console.log = ( ...args: any[] ) => {
+	originalConsoleLog( '[playground-cli]', ...args );
+};
```
sejas, line 24: Nice!, I like prefixing the output for playground-cli

src/lib/wordpress-provider/playground-cli/playground-server-process.ts
```diff
@@ -0,0 +1,162 @@
...
+
+		// Create exit promise before starting the process
+		this.exitPromise = new Promise( ( resolve ) => {
+			this.exitResolve = resolve;
+		} );
+
+		this.process = utilityProcess.fork( PLAYGROUND_SERVER_PROCESS_MODULE_PATH );
+
+		this.process.on( 'message', ( message: unknown ) => {
+			const msg = message as { type?: string; id?: number; error?: string; result?: unknown };
```
sejas: Can we define the type of message and set it in the arguments?
In general I found a bit dangerous to use `unknown` types, and then forcing a type. I would suggest double checking if this improvement can be applied in other unkowuns.

src/site-server.ts
```diff
@@ -237,12 +237,50 @@ export class SiteServer {
...
+	}
+
 	async updateCachedThumbnail() {
 		if ( ! this.details.running ) {
 			console.warn( `Thumbnail update skipped: server ${ this.details.id } is not running.` );
 			return;
 		}
 
+		const isReady = await this.waitForWordPressReady();
+		if ( ! isReady ) {
```
sejas: Do you know why we didn't need this wait before? Are we adding it just to remove the admin bar from the thumbnail?

sejas (commented):
@bcotrim , I tested the providers and both worked well, I just caught a few improvements that can be handled in separate issues, so we can merge this PR.

* Playground CLI cannot start sites when the computer is offline
* A few mu-plugins appear in the directories, and therefore are also included in the export zip.
* "Open site" auto-logs in automatically. I don't think it's a big deal unless it conflicts with other processes, such as executing wp-cli commands or handling Jetpack and Woo external requests. If we keep the playground-cli auto-login, then we could create an issue to [remove our custom code](https://github.com/Automattic/studio/blob/76564beef38d4272b86adba685d83abd130af407/vendor/wp-now/src/start-server.ts#L50-L60).

**Starting a site when offline returns this error:**

Screenshot 2025-08-07 at 14 24 14

mu-plugins appear in mu-plugins folder:
Screenshot 2025-08-07 at 14 09 27

mu-plugins are included in exported zip:
Screenshot 2025-08-07 at 14 13 30

I tested this PR by returning always the [Legacy Add Site](https://github.com/Automattic/studio/blob/trunk/src/modules/add-site/index.tsx#L47) , and then creating a site, starting, stopping, exporting, importing, pulling a site, creating a preview site, logging in in a preview site, running wp-cli commands in the Assistant.

src/site-server.ts
bcotrim: Yes that was my goal.

sejas: Then, I would suggest removing the function `waitForWordPressReady`, and creating a low priority issue to remove the admin bar from the thumbnail. In the worst case scenario we can inject some CSS before taking the snapshot.

bcotrim:
@sejas thanks for review!
> Playground CLI cannot start sites when the computer is offline

Should be fixed now.

> A few mu-plugins appear in the directories, and therefore are also included in the export zip.

I added those plugins to the export list. I will still investigate why they were being added.

> "Open site" auto-logs in automatically. I don't think it's a big deal unless it conflicts with other processes, such as executing wp-cli commands or handling Jetpack and Woo external requests. If we keep the playground-cli auto-login, then we could create an issue to [remove our custom code](https://github.com/Automattic/studio/blob/76564beef38d4272b86adba685d83abd130af407/vendor/wp-now/src/start-server.ts#L50-L60).

I don't think it will conflict. We shouldn't need to remove it for now that since it's only used in `wp-now` and we should remove all of it soon.

src/site-server.ts
bcotrim: I removed it, and added the CSS injection. Let me know what you think.

sejas (approved):
Nice! Thanks for addressing the feedback.
Let's create one or more issues for the follow-ups, like removing the mu-plugins from the files directory.

PR #1599, opened 2025-07-22
Update crash screen styling and fix crash

wojtekn:
## Related issues

- Fixes STU-612

## Proposed Changes

- I propose to update crash screen styling and to fix the issue with another crash occurring when starting the site from a read-only location.

Screenshot 2025-07-22 at 13 05 41

### Test crash screen styling
1. Run Studio in dev mode (`npm start`)
2. Trigger renderer failure from Electron menu
3. Close the trace
4. Confirm "contact support" is styled correctly (blue, and no padding) and points to the 'Contact us' page 

Before the fix, there was additional padding around the "contact support", it was black, and pointing to another page.

### Test crash screen styling for RTL language
1. Run Studio in dev mode (`npm start`)
2. Change language to RTL
3. Trigger renderer failure from Electron menu
4. Close the trace
6. Confirm layout uses English in LTR mode
7. Confirm "contact support" is styled correctly and points to the 'Contact us' page

Before the fix, error screen was displayed in English but in RTL mode, with layout slightly broken. Additionally, the user couldn't close the trace with the mouse.

### Test running Studio from read-only container
1. Navigate to build in the Buildkite
2. Download DMG file for your system arch
3. Open DMG, start Studio from DMG
4. Start site if none of site automatically started
6. Confirm it starts and that Studio doesn't crash
7. Trigger Studio update
8. Confirm dialog with correct error is displayed

Before the fix, Studio running from read-only container was crashing when the site was started.

## Pre-merge Checklist

### DIFF
--- a/src/components/default-error-fallback.tsx
+++ b/src/components/default-error-fallback.tsx
@@ -3,7 +3,6 @@ import {
 	__experimentalHStack as HStack,
 } from '@wordpress/components';
 import { useI18n } from '@wordpress/react-i18n';
-import { useEffect } from 'react';
 import { DEFAULT_LOCALE } from 'common/lib/locale';
 import Button from 'src/components/button';
 import { DynamicStylesheet } from 'src/components/dynamic-stylesheet';
@@ -90,21 +89,11 @@ const RightPanel = () => {
 };
 
 export default function DefaultErrorFallback() {
-	const { isRTL } = useI18n();
-
-	useEffect( () => {
-		document.documentElement.dir = isRTL() ? 'rtl' : 'ltr';
-	}, [ isRTL ] );
-
 	return (
-		<>
+		<div dir={ 'ltr' }>
 			<DynamicStylesheet
 				id="wordpress-components-style"
-				href={
-					isRTL()
-						? '../main_window/styles/wordpress-components-style-rtl.css'
-						: '../main_window/styles/wordpress-components-style.css'
-				}
+				href={ '../main_window/styles/wordpress-components-style.css' }
 			/>
 			<VStack
 				className={ cx(
@@ -148,6 +137,6 @@ export default function DefaultErrorFallback() {
 					</div>
 				</HStack>
 			</VStack>
-		</>
+		</div>
 	);
 }
--- a/src/components/default-error-fallback.tsx
+++ b/src/components/default-error-fallback.tsx
@@ -3,6 +3,7 @@ import {
 	__experimentalHStack as HStack,
 } from '@wordpress/components';
 import { useI18n } from '@wordpress/react-i18n';
+import { useEffect } from 'react';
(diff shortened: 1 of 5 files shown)


src/components/default-error-fallback.tsx
```diff
@@ -88,48 +90,64 @@ const RightPanel = () => {
 };
 
 export default function DefaultErrorFallback() {
+	const { isRTL } = useI18n();
+
+	useEffect( () => {
+		document.documentElement.dir = isRTL() ? 'rtl' : 'ltr';
```
wojtekn: When Studio displays default error fallback, translations are not available, so it will always switch to LTR. We could even simplify this code and hardcode it to `ltr`. Any thoughts?

src/storage/user-data.ts
```diff
@@ -107,7 +107,7 @@ export async function saveUserData( data: UserData ): Promise< void > {
 	await writeFile( filePath, asString, 'utf-8' );
 }
 
-const LOCKFILE_PATH = nodePath.join( getResourcesPath(), LOCKFILE_NAME );
+const LOCKFILE_PATH = getUserDataLockFilePath();
```
wojtekn, line 110: I'm unsure why in https://github.com/Automattic/studio/pull/1319 we used two different locations for the app lock file. 🤔

sejas (approved):
Nice! Thanks for improving the crash window. I confirm I see the links correctly.

| Before | After |
|--------|--------|
|  Screenshot 2025-07-22 at 15 37 57 | Screenshot 2025-07-22 at 15 29 43 |

In RTL I see the english text as expected:

Screenshot 2025-07-22 at 15 31 09


I also see the expected error when trying to update Studio from a DMG image.

Screenshot 2025-07-22 at 15 42 59

src/components/default-error-fallback.tsx
sejas: I think it's better if we hard code it to LTR. That way we don't have "dead" code that won't work.

src/storage/user-data.ts
sejas: I like this improvement of using the same lock file function. But should it be a different PR?

wojtekn: It's not only an improvement, but a fix for one of the bugs reported in STU-612. @ndiego reported layout issues on crash screen, but the crash in steps he shared shouldn't occur in the first place. The lock file location fixes that crash.

It would make sense to have two separate issues, but I addressed both as part of one PR, as both problems were reported under one issue and are related - one reproduces the other.

src/components/default-error-fallback.tsx
wojtekn: Makes sense, done in e83f71c4104d742853931d0bce343d980c2f4317.

sejas (approved):
Thanks for clarifying. It makes sense having the fix together 👍

PR #1606, opened 2025-07-22
Playground CLI Provider - Create and Remove Site & Execute WP CLI

bcotrim:
## Related issues

- N/A

## Proposed Changes

- Implements site creation functionality for the playground-cli WordPress provider
- Adds WP-CLI command execution capability through isolated child processes
- Creates a dedicated worker process to prevent playground CLI's process.exit() calls from terminating Studio
- Uses blueprints to set up WordPress sites with basic configuration and execute WP-CLI commands

## Testing Instructions

- `npm start`
- Enable blueprints flag
- Create a new site with default options
- Create a new site with advanced options
- Remove site

## Pre-merge Checklist


forge.config.ts
```diff
@@ -28,7 +28,9 @@ const config: ForgeConfig = {
 		icon: './assets/studio-app-icon',
 		osxSign: {},
 	},
-	rebuildConfig: {},
+	rebuildConfig: {
+		ignoreModules: [ 'fs-ext' ],
```
sejas: We'll need to confirm Windows and Mac build versions work without rebuilding `fs-ext`.
It's a pity that new Playground dependencies require this rule.

src/lib/wordpress-provider/types.ts
```diff
@@ -42,6 +42,8 @@ export interface WordPressServerProcess {
 }
 
 export interface WordPressProvider {
+	readonly PROVIDER_TYPE: string;
```
sejas: Nit: we can consider renaming this property to NAME or PROVIDER_NAME as we don't expect to have more providers of the same type.

```suggestion
	readonly NAME: string;
```

It also aligns better with other constants of the code like `readonly PROVIDER_TYPE = PLAYGROUND_CLI_PROVIDER_NAME;`

sejas (commented):
@bcotrim , great work!
I tested creating new sites with default settings and they work great.
Creating new sites with different PHP and WP versions also work well.
Changing the versions of an existing site doesn't work for me. Maybe that's expected.

I noticed that the site image appears in the overview, and opening the site URL logs the user in automatically. We might need to consider creating an issue to disable this auto-login feature.

Screenshot 2025-07-25 at 12 24 48

Screenshot 2025-07-25 at 12 25 27

Screenshot 2025-07-25 at 12 26 03

I left a couple of comments. Not building the Windows versions in CI seems to be the biggest concern.

bcotrim:
@sejas I'm going to merge this one into https://github.com/Automattic/studio/pull/1594 for easier review.
I will address your feedback there

PR #1607, opened 2025-07-23
Change a way of generating dev build version

wojtekn:
## Related issues

- Related discussion: p1753275455765609-slack-C04GESRBWKW

## Proposed Changes

I propose to fix a way of generating dev build version: currently, it handles part that counts commits correctly, but it relies on version from `package.json` for the version part, so it uses incorrect dev version as soon as we release the first beta for the new release.

## Testing Instructions

1. Pretend you want to release new beta for incoming 1.5.6 and update it to 1.5.6-beta1 in `package.json` file.
2. Run `node ./scripts/prepare-dev-build-version.mjs`
3. Run `git diff` and see that the version was changed to 1.5.5-dev37 in the `package.json`

Before the fix, the generated version would be 1.5.6-dev37.

## Pre-merge Checklist

### DIFF
--- a/scripts/prepare-dev-build-version.mjs
+++ b/scripts/prepare-dev-build-version.mjs
@@ -23,10 +23,11 @@ const packageJsonPath = path.resolve( __dirname, '../package.json' );
 const packageJsonText = await fs.readFile( packageJsonPath, 'utf-8' );
 const packageJson = JSON.parse( packageJsonText );
 
-// Parse the version using semver to get just the core version numbers
-const parsedVersion = semver.parse( packageJson.version );
+// Use version from latestTag (strip leading 'v' if present)
+const tagVersion = latestTag.startsWith('v') ? latestTag.slice(1) : latestTag;
+const parsedVersion = semver.parse( tagVersion );
 if ( ! parsedVersion ) {
-	throw new Error( `Invalid version in package.json: ${ packageJson.version }` );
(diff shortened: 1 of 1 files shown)


sejas (approved):
Good catch!
I confirm that executing `node ./scripts/prepare-dev-build-version.mjs` updates the package version to the same `1.5.5-dev38` value, for both cases when we have `1.5.5` and `1.5.6-beta1` in package.json.
I also confirmed that `getLatestTag` won't return the 1.5.6-beta1 version as latest tag.
`git tag --sort=-v:refname'` returns all the versions, but the script filters out those with a `-`.
```
v1.5.5-beta6
v1.5.5-beta5
v1.5.5-beta4
v1.5.5-beta3
v1.5.5-beta2
v1.5.5-beta1
v1.5.5
v1.5.4-beta1
v1.5.4
```

https://github.com/Automattic/studio/blob/1e286f737e396ef6f1b1f68c6f9ae9f3192a2a8f/scripts/lib/git-utils.mjs#L17

sejas:
This PR won't fix the conflict immediately, but that's fine by me. We'll experience the same errors in dev builds until we release 1.5.6 or when the count exceeds the existing uploaded versions.

wojtekn:
> This PR won't fix the conflict immediately, but that's fine by me. We'll experience the same errors in dev builds until we release 1.5.6 or when the count exceeds the existing uploaded versions.

It will - we will remove the unnecessary 1.5.5-devXX versions from CDN.

PR #1608, opened 2025-07-23
[Tooling] Run Windows Setup before installing npm dependencies

iangmaia:
Fixes AINFRA-987

To fix the issue seen in https://github.com/Automattic/studio/pull/1594, this PR runs the Windows setup before installing the Node dependencies.
We also had to install Python and additional Visual Studio Build Tools components in both the Windows build and unit test jobs.

Depends on CI Toolkit Plugin PR https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/186

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -36,6 +36,8 @@ steps:
 
   - label: Windows Unit Tests
     command: |
+      & "prepare_windows_host_for_app_distribution.ps1"
+
       bash .buildkite/commands/install-node-dependencies.sh
       echo '--- :npm: Run Unit Tests'
       npm run test
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -36,7 +36,8 @@ steps:
 
   - label: Windows Unit Tests
     command: |
-      & "prepare_windows_host_for_app_distribution.ps1"
+      # prepare_windows_host_for_app_distribution.ps1 comes from CI Toolkit Plugin
+      & "prepare_windows_host_for_app_distribution.ps1" -InstallPython $true -InstallNativeCompilationTools $true
 
       bash .buildkite/commands/install-node-dependencies.sh
       echo '--- :npm: Run Unit Tests'
--- a/.buildkite/commands/build-for-windows.ps1
+++ b/.buildkite/commands/build-for-windows.ps1
@@ -18,7 +18,8 @@ if ($BuildType -notin $VALID_BUILD_TYPES) {
     Exit 1
 }
 
-& "prepare_windows_host_for_app_distribution.ps1" # via CI toolkit plugin
+# prepare_windows_host_for_app_distribution.ps1 comes from CI Toolkit Plugin
+& "prepare_windows_host_for_app_distribution.ps1" -InstallPython $true -InstallNativeCompilationTools $true
 If ($LastExitCode -ne 0) { Exit $LastExitCode }
 
 Write-Host "--- :npm: Installing Node dependencies"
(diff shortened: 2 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -36,6 +36,9 @@ steps:
 
   - label: Windows Unit Tests
     command: |
+      # prepare_windows_host_for_app_distribution.ps1 comes from CI Toolkit Plugin
+      & "prepare_windows_host_for_app_distribution.ps1" -InstallPython $true -InstallNativeCompilationTools $true
```
AliSoftware, line 40: I'm guessing this means you tested with the case of not installing the SDK (while just installing the NativeCompilationTools) and it wasn't enough?

iangmaia: Yeah, I tried to use
- `install_python.ps1` ([build](https://buildkite.com/automattic/studio/builds/7247)) directly, then 
- `"prepare_windows_host_for_app_distribution.ps1"` ([build](https://buildkite.com/automattic/studio/builds/7248)) and then
- `"prepare_windows_host_for_app_distribution.ps1" -InstallNativeCompilationTools` ([build](https://buildkite.com/automattic/studio/builds/7249)).

It only fully worked with `prepare_windows_host_for_app_distribution.ps1 comes from CI Toolkit Plugin
& "prepare_windows_host_for_app_distribution.ps1" -InstallPython  -InstallNativeCompilationTools` ([build](https://buildkite.com/automattic/studio/builds/7251#01983cce-cd16-465a-933f-03712d1ce31a)).

AliSoftware: Wait but none of those 3 cases cover the case I propose of installing the native compilation tools without installing the Windows 10 SDK though, do they? 🤔

iangmaia: Ah indeed, not that combination, I misunderstood your original comment.

At the moment `-InstallNativeCompilationTools` is an argument to the windows 10 SDK installation script, which always install the WindowsSDK. So passing both `-InstallNativeCompilationTools` and `-SkipWindows10SDKInstallation` will just skip the WindowsSDK installation.

Maybe we can rename the script `install_windows_10_sdk.ps1` to something like `install_visual_studio_build_tools.ps1` and have these parameters independent from each other.
The way it's currently built, with `SkipWindows10SDKInstallation` (so installed by default) is what makes me a bit hesitant to do that straight away (too many changes in a single toolkit update), though I think it would be a lot nicer.

Wdyt of adding those changes to a separate (likely breaking) version of the CI Toolkit and then update the apps accordingly?

AliSoftware: Yeah that goes with my comment in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/186#discussion_r2228794439 too about `-InstallNativeCompilationTools` being ignored if `-SkipWindows10SDKInstallation` is also passed.

I see you addressed this in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/186/commits/0f0228066e7621e7f7dd9b8b358c9640287ad9e5 and https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/186/commits/fbdc993462bf7728ed41ff7d3781850f2323faf8 👍 

> Maybe we can rename the script `install_windows_10_sdk.ps1` to something like `install_visual_studio_build_tools.ps1` and have these parameters independent from each other.
> […]
> Wdyt of adding that to a separate (likely breaking) version?

yep, makes sense, works for me 👍

iangmaia:
@AliSoftware  I re-run the build to make sure we're on the latest [a8c-ci-toolkit-buildkite-plugin](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin). Once we merge the branch over there and release a new version I'll update the reference here and we can finally merge this PR.

AliSoftware (approved):
Approving under the assumption you're gonna update the `CI_TOOLKIT_VERSION` once shipped before merging, like you mention above. 👍

PR #1609, opened 2025-07-25
[DO NOT MERGE] Test `SKIP_CERTIFICATE_DOWNLOAD`

iangmaia:
Testing `SKIP_CERTIFICATE_DOWNLOAD` variable.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -44,6 +44,8 @@ steps:
       npm run test
     agents:
       queue: windows
+    env:
+      SKIP_CERTIFICATE_DOWNLOAD: "true"
     plugins: [$CI_TOOLKIT_PLUGIN, $NVM_PLUGIN]
     notify:
       - github_commit_status:


iangmaia:
@AliSoftware re: [skipping certificates download](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/186#discussion_r2230761646). I've tested it here and it worked!

See the builds:
- [Skipping certificate download](https://buildkite.com/automattic/studio/builds/7263#019841d7-9c10-40b2-82b5-df19c47e6c52/373)
- [Downloading certificate](https://buildkite.com/automattic/studio/builds/7259#01984165-9e88-4ae9-a2b2-1ca3a5f5d38b/389)

Not downloading them still had the tests passing (🎉), though downloading them is very fast so we don't save a lot of time not downloading them. Perhaps it's worth it to still skip the download, but regardless it was a nice "TIL" moment 🙂

AliSoftware:
Nice! It's good to have confirmation they're not needed for tests 🙂 

I think at some point it'd make sense to separate the cert download in its own `.ps1`, and have it called from `prepare_windows_host…` unless `-SkipDownloadCertificate` switch is set etc… if only for consistency and clear separation of concerns/responsibilities in our toolkit scripts, and for the (admittedly very limited) slight improved security of not having such secrets on the machine if not necessary, more than for the ridiculous time it'd save.

But all that could also be a Linear issue for later.

iangmaia:
> I think at some point it'd make sense to separate the cert download in its own `.ps1`, and have it called from `prepare_windows_host…` unless `-SkipDownloadCertificate` switch is set etc… if only for consistency and clear separation of concerns/responsibilities in our toolkit scripts, and for the (admittedly very limited) slight improved security of not having such secrets on the machine if not necessary, more than for the ridiculous time it'd save.
 
Agreed 👍 I've created the issue AINFRA-1046 also referring the discussion about another script break down at https://github.com/Automattic/studio/pull/1608#discussion_r2229111592.

PR #1624, opened 2025-07-30
Update Onboarding to simplify flow

wojtekn:
## Related issues

- Fixes STU-622

## Proposed Changes

I propose to refresh onboarding:
- Display it only once for each user, enabling empty state that was unused
- Simplify it by removing import field
- Add additional subheading, center elements
- Change WordPress logo to Studio logo

|**Onboarding**|**New empty state**|
|---|---|
|Screenshot 2025-07-31 at 10 09 01|Screenshot 2025-07-31 at 10 09 37|

## Testing Instructions

1. Remove `onboardingCompleted` key from the `appdata-v1.json` file
2. Start Studio
3. Confirm that onboarding looks as expected
4. Proceed with adding a new site
5. Remove the site
6. Confirm that onboarding is not displayed again
7. Open the 'Add site' form and confirm that it's usable, that elements can be clicked
8. Restart Studio
9. Confirm that onboarding is not displayed again

## Pre-merge Checklist


src/components/site-form.tsx
```diff
@@ -338,166 +338,160 @@ export const SiteForm = ( {
 								error={ fileError }
 							/>
 						</div>
-						{ onSelectPath && (
```
wojtekn, line 341: The only change here is about removing the advanced settings parts from the `setFileForImport` check.

src/components/site-content-tabs.tsx
```diff
@@ -22,7 +22,7 @@ export function SiteContentTabs() {
 
 	if ( ! selectedSite ) {
 		return (
-			<div className="w-full h-full flex items-center justify-center">
+			<div className="w-full h-full flex items-center justify-center app-no-drag-region">
```
wojtekn, line 25: This is a fix for the 'Add site' form in the new empty state view.

wojtekn:
@bcotrim I skipped the feature flag as it doesn't seem required for the onboarding changes. 

Also, we could work on implementing empty state layout changes separately - the changes will be small in the end, limited to adding SVG and updating texts or CTA.

Let me know what you think about those.

src/components/onboarding.tsx
```diff
@@ -128,39 +109,34 @@ export default function Onboarding() {
...
 
-			<div
-				className="w-1/2 bg-white p-[50px] flex flex-col relative overflow-y-auto app-no-drag-region"
-				ref={ dropRef }
-			>
-				{ isDraggingOver && <DragAndDropOverlay /> }
-				<div className="flex flex-col justify-center items-start flex-[1_0_0%] gap-8">
-					<div className="flex flex-col items-start self-stretch gap-6">
-						<h1 className="font-normal text-xl leading-5">{ __( 'Add your first site' ) }</h1>
+			<div className="w-1/2 bg-white p-[50px] flex flex-col relative overflow-y-auto app-no-drag-region">
```
ivan-ottinger, line 131: Similarly as I [suggested in Bernardo's PR](https://github.com/Automattic/studio/pull/1625#pullrequestreview-3074548332), would we want to start using VStack, HStack and other WordPress core components more?

ivan-ottinger (commented):
The new onboarding works correctly. Nice!

Markup on 2025-07-31 at 11:07:28

It is only triggered when `onboardingCompleted` is not `true` and there are no sites.

The empty site state looks good as well and I did not observe any regressions.

Markup on 2025-07-31 at 11:08:09


I am sharing just one comment below. It also looks like one more test needs adjustment.

src/components/onboarding.tsx
wojtekn: I'm not sure if there is any benefit in this case, and it would probably make more sense for complex elements like a file tree or a grid. On top of that, VStack and HStack are marked as experimental and subject to heavy changes. 🤷‍♂️

https://developer.wordpress.org/block-editor/reference-guides/components/v-stack/

wojtekn:
> It also looks like one more test needs adjustment.

I fixed this e2e test.

src/components/onboarding.tsx
ivan-ottinger: Yes, they are marked as experimental, but we have been using them in Calypso a lot - also in the new dashboard. 🤔 

The same applies to components like Text and Heading. But yeah, I don't have strong opinion on using them in Studio.

ivan-ottinger (approved):

src/components/onboarding.tsx
bcotrim: There's seems to be a push to use the shared components as much as possible. I think it's most noticeable on more "complex" components (buttons, menus, etc... )
That being said, I also don't have a strong preference regarding `VStack` and `HStack`. I added them in the PR Ivan mentioned but I'm happy to adjust to what we decide.

bcotrim (approved):
LGTM 👍 

I agree with adding changes to the empty state in a follow-up PR

e2e/page-objects/onboarding.ts
```diff
@@ -25,7 +25,7 @@ export default class Onboarding {
 	}
 
 	get continueButton() {
-		return this.locator.getByRole( 'button', { name: 'Add site' } );
+		return this.locator.getByRole( 'button', { name: 'Continue' } );
```
bcotrim: This plan should work:
1. Adjust this condition to pass on both scenarios
2. Merge this PR
3. Open new PR to update the commit hash in `metrics-comparison.yml` to the trunk commit with this update change and adjust the condition back to `name: Continue`

What do you think?

```suggestion
		return this.locator.getByRole( 'button', { name: /Continue|Add site/ } );
```

PR #1625, opened 2025-07-30
Studio Blueprints: Add site - options modal

bcotrim:
## Related issues

- Fixes STU-654

## Proposed Changes

- Add new options modal for Add site workflow
- Fixes feature flag hook, to correctly get the defaults and update the flag values

image

Note: none of the options are working, we will implement each one of the options in separate follow-up PRs

## Testing Instructions

- Press "Add site"
- You should see the options modal as shared above
- Check if you can close the modal
- Disable blueprints flag from the Electron menu or restart Studio by running `npm start`
- Add site should work as previously

## Pre-merge Checklist


ivan-ottinger (commented):
Great work, Bernardo! I am still reviewing the code, but noticed two things, so wanted to share them with you already:

1. The icon of the second option is different than in the design:

Markup on 2025-07-31 at 10:10:14

2. The window dragging is interfering with the modal closing button and it can be tricky to close the modal:

https://github.com/user-attachments/assets/7e7933de-1557-407e-8c98-8490f6cf6de7

ivan-ottinger (commented):
The modal renders nicely and I did not observe any regressions regarding the legacy modal. 👍🏼 

Apart from my previous comment, here's one more thing we could consider:

I wonder if it would make sense to use HStack, VStack, Text and Heading components from the WordPress core instead of divs - since we are now moving to these components in Calypso as well.

nightnei:
>Check if you can close the modal

@bcotrim  I am continuing reviewing the code, but regarding UI - the close button is partially clickable:
https://github.com/user-attachments/assets/40ca2c3c-9753-4de5-b20e-0bfe696ab90d

nightnei:
> I wonder if it would make sense to use HStack, VStack, Text and Heading components from the WordPress core instead of divs - since we are now moving to these components in Calypso as well.

IMO - Studio is independent app, we should not make any dependency on dotcom, even regarding some rules, styleguides, etc.
Also, even already Studio uses npm instaed of yarn, we don't follow the same state-management libs and rules, so in general - I would not look at Calypso. And I woudl continue using divs.

src/components/add-site/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+	}, [] );
+
+	useIpcListener( 'add-site', () => {
+		if ( isAnySiteProcessing ) {
+			return;
+		}
+		openModal();
+	} );
+
+	// If blueprints is disabled, use the existing component
```
nightnei: I think code is self explanatory
```suggestion
```

src/components/add-site/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+		return (
+			<AddSiteLegacy
+				className={ className }
+				showModal={ showModal }
+				setShowModal={ setShowModal }
+			/>
+		);
+	}
+
+	// If blueprints is enabled, show modal with options
```
nightnei: I think code is self explanatory
```suggestion
```

ivan-ottinger:
> IMO - Studio is independent app, we should not make any dependency on dotcom, even regarding some rules, styleguides, etc.

I have seen VStack and HStack being used it some places in Studio already so that what made me think we could consider it further. Mostly from the perspective of unifying design and simplifying styling as I could see how using those core components reduces the need for custom styles for example.

src/components/modal.tsx
```diff
@@ -3,7 +3,11 @@ import { useI18n } from '@wordpress/react-i18n';
 import { cx } from 'src/lib/cx';
 import type { ComponentProps } from 'react';
 
-export default function Modal( { className, ...rest }: ComponentProps< typeof WPModal > ) {
+export default function Modal( {
+	className,
+	isFullScreen,
```
nightnei: NIT Typically prefix `is` isn't used. For example `disabled`. And I think it makes sense, since this has better readability `<Modal fullScreen onRequestClose={ closeModal }>` instead of `<Modal isFullScreen onRequestClose={ closeModal }>`.

```suggestion
	fullScreen,
```

src/hooks/use-feature-flags.tsx
```diff
@@ -26,15 +31,44 @@ interface FeatureFlagsProviderProps {
 }
 
 export const FeatureFlagsProvider: React.FC< FeatureFlagsProviderProps > = ( { children } ) => {
-	const [ featureFlags, setFeatureFlags ] = useState< FeatureFlagsContextType >( {
-		...defaultFeatureFlags,
+	const [ featureFlags, setFeatureFlags ] = useState< FeatureFlagsContextType >( () => {
```
nightnei, line 29: Could you please elaborate on all changes inside this module?
I didn't get why do we need them, also they look quite complicated and also we didn't have such changes before so I wonder - can we avoid them as we did it before?

nightnei (changes requested):
Also, let's move and initially create folder for it inside `src/modules`, e.g. `src/modules/add-site/*`

bcotrim:
@nightnei @ivan-ottinger thanks for the reviews!
I think I've addressed the feedback.

I decided to create a new modal for full-screen, the one from wordpress components isn't completely full-screen and my workaround was causing the strange behavior in the close button. Let me know what you think.

~As for the icon, I will follow-up with design and update in a follow-up PR if you agree.~

src/hooks/use-feature-flags.tsx
bcotrim: We had a bug with how feature flags were being updated/preserved, so changing the value in the menu wasn't being reflected in the app. But you are right, that logic was complex, and we didn't need all that.

The simplified implementation now properly handles the three sources of feature flags (defaults → appGlobals → API) in a cleaner way, and it correctly preserves API flags when the menu triggers an update.

Let me know if it is clear now.

src/hooks/use-feature-flags.tsx
```diff
@@ -26,15 +26,24 @@ interface FeatureFlagsProviderProps {
...
+		if ( window.appGlobals ) {
+			return {
+				...defaultFeatureFlags,
+				...window.appGlobals,
+			};
+		}
+		return { ...defaultFeatureFlags };
```
nightnei: Hm, can't we go this way? it should work:
Screenshot 2025-07-31 at 18 53 48

```suggestion
		return {
			...defaultFeatureFlags,
			...window.appGlobals,
		};
```

nightnei (approved):
Looks great 👍
Left only one small comment

src/modules/add-site/components/add-site-options.tsx
```diff
@@ -0,0 +1,85 @@
...
+			{ icon }
+			<VStack className="flex-1">
+				<Heading className="text-xl" weight="500">
+					{ title }
+				</Heading>
+				<Text className="text-base text-gray-500" weight="300">
+					{ description }
+				</Text>
+			</VStack>
+			<Icon className="mt-0.5" icon={ chevronRight } size={ 24 } fill="#949494" />
```
ivan-ottinger: We may want to use `chevronLeft` for RTL as currently it points the same as in LTR:

Markup on 2025-08-01 at 09:35:31

Also, I think the title and description should be starting on the right side when RTL locale is set.

ivan-ottinger (commented):
Thank you for making the updates, Bernardo!

I have noticed one more thing: each icon has a different size (36px vs 32px vs 24px).

It is mostly visible with the "Start from blueprint" button where the title and description is then shifted to the side a bit:

Markup on 2025-08-01 at 09:25:37

---

I left one more comment related to RTL below.

ivan-ottinger (approved):
Nice! Icons are now the same size and both LTR and RTL are rendered correctly!

| LTR | RTL |
|--------|--------|
| Markup on 2025-08-01 at 11:32:14 | Markup on 2025-08-01 at 11:32:44 |

PR #1630, opened 2025-08-01
Add site: "Create a site", "Start from a blueprint" and "Import from backup"

bcotrim:
## Related issues

- Fixes STU-623 STU-624 STU-625 STU-633

## Proposed Changes

- Add new UI for "Create a site"
- Add new UI for "Start from blueprint" 
- Add new UI for "Import from backup"
- Implemented navigation and stepper using WordPress components

| Options | Create a site | Start from blueprint | Import from backup
| image | image | image | image |

## Testing Instructions

- Click "Add Site"
- Create a site using "Create a site" option, all advanced settings should work as before
- Create a site using "Start from blueprint", all advanced settings should work as on "create a site"
- Create a site using "Import from backup", all advanced settings should work as on "create a site"
- If you are offline "Start from blueprint" should be disabled
- Blueprints are fetched from our WordPress API (you can confirm in the network tab details)
- Restart Studio with `npm start` or disabled the Blueprints flag in the Electron menu
- "Add site" functionality should work as before without any changes or regressions

## Pre-merge Checklist


sejas (commented):
Nice! Great start.  
The first thing I noticed is that the implementation looks larger than the design. I also see that the design is using different proportions than the default Studio size 1007x680px vs 900x600px.
Even though, the design has narrower rows and smaller font sizes.
Also, there are a few differences like the icons, images, and tags. Maybe I'm comparing a wrong figma page.
It would be great if the scroll would appear in the right , instead of over the buttons.
The bottom progress has ticks instead of numbers.

| Figma | Implementation |
|--------|--------|
|Screenshot 2025-08-04 at 23 07 25 | Screenshot 2025-08-04 at 23 08 30 |
| Screenshot 2025-08-04 at 23 24 04 | Screenshot 2025-08-04 at 23 13 01 |

src/components/site-form.tsx
```diff
@@ -338,166 +338,160 @@ export const SiteForm = ( {
...
+					<>
+						<div className="flex flex-row items-center mb-0">
+							<Button className="pl-0" onClick={ handleAdvancedSettingsClick }>
+								<Icon size={ 24 } icon={ chevronIcon } className={ error ? 'text-red-500' : '' } />
 								<div
-									className={ cx(
-										'transition-all duration-500 ease-in-out overflow-hidden flex flex-col gap-2 interpolate-size-allow-keywords',
-										isAdvancedSettingsVisible ? 'h-auto opacity-100' : 'h-0 opacity-0'
-									) }
+									className={ cx( 'text-[13px] leading-[16px] ml-2', error ? 'text-red-500' : '' ) }
```
sejas: In `cx`, we don't need to use a ternary condition.

```suggestion
									<div className={ cx( 'text-[13px] leading-[16px] ml-2', error && 'text-red-500' ) }>
```

src/index.ts
```diff
@@ -252,7 +252,7 @@ async function appBoot() {
 			const basePolicies = [
 				"default-src 'self'", // Allow resources from these domains
 				"script-src-attr 'none'",
-				"img-src 'self' https://*.gravatar.com https://*.wp.com data:",
+				"img-src 'self' https://*.gravatar.com https://*.wp.com https://blueprintlibrary.wordpress.com data:",
```
sejas, line 255: Nice addition to display the blueprint images.

src/stores/wpcom-api.ts
```diff
@@ -43,12 +43,35 @@ const snapshotUsageSchema = z
...
 };
 
+// Get or create a public WPCOM client for unauthenticated requests
+function getPublicWpcomClient(): WPCOM {
+	if ( ! publicWpcomClient ) {
+		publicWpcomClient = new WPCOM();
+	}
+	return publicWpcomClient;
+}
+
```
sejas: Why do we need the getPublicWpcomClient function?
Let's initialize the variable directly:

```suggestion
const publicWpcomClient = new WPCOM();

export const setWpcomClient = ( client: WPCOM | undefined ) => {
	wpcomClient = client;
};

```

And we can move it closer to its first usage.

src/modules/add-site/index.tsx
```diff
@@ -1,41 +1,284 @@
...
+		useCustomDomain,
+		setUseCustomDomain,
+		customDomain,
+		setCustomDomain,
+		customDomainError,
+		setCustomDomainError,
+		enableHttps,
+		setEnableHttps,
+		loadAllCustomDomains,
+		fileForImport,
```
sejas: This variable is not used. 

```suggestion
```

src/modules/add-site/index.tsx
```diff
@@ -1,41 +1,284 @@
...
+		setFileForImport,
+	} = useAddSite();
```
sejas: Could we group or simplify the logic for useAddSite? Destructuring all the elements and passing them as props makes the code a bit difficult to follow.

src/modules/add-site/index.tsx
```diff
@@ -1,41 +1,284 @@
...
+		setWpVersion,
+		latestStableVersion,
+		loadAllCustomDomains,
+	] );
+
+	useEffect( () => {
+		if ( showModal && ! nameSuggested && ! loadingSites ) {
+			void initializeForm();
+		}
+	}, [ showModal, nameSuggested, loadingSites, initializeForm ] );
```
sejas: Would be possible to call `initializeForm` programmatically, maybe inside `openModal`. That way we could remove the useEffect and and the useCallback.

src/modules/add-site/index.tsx
```diff
@@ -1,41 +1,284 @@
...
+			void refetch();
+		}
 		setShowModal( true );
-	}, [] );
+	}, [ refetch, isUninitialized ] );
 
 	const closeModal = useCallback( () => {
 		setShowModal( false );
-	}, [] );
+		resetForm();
```
sejas, line 274: Do we need to call resetForm when the modal closes, even if we call `initializeForm` when it opens?

sejas (commented):
@bcotrim , thanks for sharing the new Figma designs with me RToz6tIuQ7nlZrikBte4GU-fi-9247_41860
The implementation looks like the designs 👍 . I only found a small difference:
The breadcrumb circle should turn blue when the first step is completed:

| Before | After |
|--------|--------|
|  Screenshot 2025-08-06 at 13 23 20 | Screenshot 2025-08-06 at 13 22 38 |

I tested it and it works as expected. Maybe we can improve the selected state.

I also tested it making the window smaller:

Screenshot 2025-08-06 at 13 21 26
 
Screenshot 2025-08-06 at 13 21 34

And RTL:
Screenshot 2025-08-06 at 13 20 42
Screenshot 2025-08-06 at 13 20 27

src/modules/add-site/index.tsx
bcotrim: I looked into this, but the current pattern is actually necessary because initializeForm depends on sites being loaded, which might not be ready when openModal is called.

src/modules/add-site/components/import-backup.tsx
```diff
@@ -0,0 +1,190 @@
...
+	}, [] );
+
+	// Format file size
+	const formatFileSize = useCallback( ( bytes: number ) => {
+		if ( bytes === 0 ) return '0 Bytes';
+		const k = 1024;
+		const sizes = [ 'Bytes', 'KB', 'MB', 'GB' ];
+		const i = Math.floor( Math.log( bytes ) / Math.log( k ) );
+		return Math.round( ( bytes / Math.pow( k, i ) ) * 100 ) / 100 + ' ' + sizes[ i ];
+	}, [] );
```
sejas, line 108: these doesn't need to be useMemo or useCallback, they could be moved regular functions or hooks.

src/hooks/use-site-details.tsx
```diff
@@ -28,6 +29,7 @@ interface SiteDetailsContext {
 		wpVersion?: string,
 		customDomain?: string,
 		enableHttps?: boolean,
+		blueprint?: Blueprint | null,
```
sejas, line 32: Here we accept Blueprint, null or undefined, while ipc-handler accept only Blueprint or undefined. It would be great to consolidate types.

src/hooks/use-site-details.tsx
```diff
@@ -236,10 +260,12 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 					siteName,
 					wpVersion,
 					customDomain,
-					enableHttps
+					enableHttps,
+					undefined,
+					blueprint || undefined
```
sejas, line 265: We can consider moving the blueprint parameter to the first elements to avoid passing undefined in the middle, or making the parameters an object.

sejas (approved):
I was able to create a site with the selected blueprint.
I also tested importing a site, and creating a blank site when offline.

I noticed a few improvements:
1. The import screen shouldn't have a scroll with the default Studio site.
2. I suggest that when typing a site name, if the user press enter, it should go to the next step and create a site
3. When offline, the blueprints button is disabled with a not allow cursor, we can consider adding a tooltip explaining why it's disabled.
4. When the blueprint fails, we should display an error message or alert. Currently I selected a blueprint, and then deactivated the wifi, and a blank site was created without displaying any error.

Screenshot 2025-08-12 at 12 57 04


Success creating a Blueprint site:

https://github.com/user-attachments/assets/cc8b5cee-13bc-48bf-8366-a1b3644b176f

Success creating a site from an import.

https://github.com/user-attachments/assets/92133956-2531-4d01-8a47-6bf4cbe0b23d

Blueprint created when offline by enabling wifi, selecting the blueprint, and then deactivating wifi again, it results in a blank site.

https://github.com/user-attachments/assets/5f4ab1d5-854f-4ef8-874d-5e5bc04f342f

I think this is a big PR, and it's better to merge it and address any follow-ups in separate, smaller issues to reduce conflicts and share the work among the rest of the team.

sejas (commented):
I also tested it by creating a site and selecting a blueprint JSON file from my laptop.
Here are dozens of blueprint examples to use for testing: https://github.com/WordPress/blueprints/blob/trunk/GALLERY.md

https://github.com/user-attachments/assets/51673e8b-44bb-4fb4-b226-83f2e5f06b5a

sejas:
One last consideration is that the Figma designs  are using a window size of 1007x680px while Studio uses a default size of 900x600px.
Not sure if we should discuss increasing the default Studio size or "scale" down the sizes a bit.

Here is a comparison using the 1007x680px window size:

| Before | After |
|--------|--------|
|  design-start-menu | Screenshot 2025-08-12 at 13 45 34 |
| design-select-blueprint | Screenshot 2025-08-12 at 13 45 53 |

In general, I noticed that the fonts should be lighter, and a bit smaller in some cases.
I sued this Figma designs for comparison from July 25. RToz6tIuQ7nlZrikBte4GU-fi-9470_275077

nightnei:
I have also tested w/o the feature flag, to make sure that everything still works well.
Merging this PR, and will continue with smaller follow-up PRs, to address Antonio's feedback and mine.

PR #1642, opened 2025-08-06
Release v1.5.6

gcsecsey:
## Proposed Changes

Release v1.5.6

## Testing Instructions

Visual review should suffice


package.json
```diff
@@ -77,7 +77,7 @@
 		"cross-env": "^7.0.3",
 		"css-loader": "^6.0.0",
 		"ejs": "^3.1.10",
-		"electron": "35.0.0",
+		"electron": "^35.0.0",
```
sejas: Is this change intentional?
I think it makes sense to update Electron patch versions, but it would be better to do it in a beta release first.

sejas (commented):
The App starts correctly, and I tested basic functionality, create and start a site.
I also noticed that we are adding `^` to the `electron` version in package.json.

package.json
gcsecsey: Thanks for highlighting, this wasn't intentional, I made this change while debugging another issue. I agree, we can experiment with using the patch versions separately, for now I reverted this change.

sejas (approved):

PR #1645, opened 2025-08-06
Update environment types and related badge handling

ivan-ottinger:
## Related issues

- Resolves STU-652

## Proposed Changes

Pressable is no longer setting `sandbox` environment type for their sandbox sites. Instead, these sites are now using the `development` environment type. Because of that, we are adjusting the related environment badges in Studio to use `Development` badge for sites with `development` environment type set.

Related discussion: p1753471348570109-slack-C08AHBHGDEZ.

Together with this change we are also removing support for the `local` environment type (and related badge and Push / Pull modal copy). Only the following three environment types will be represented in badges: `Production`, `Staging` and `Development`. All other environments, including local will use `Production` as their badge.

We are also adjusting the logic of` getSiteEnvironment` to rely on `site.isStaging` for WordPress.com sites - instead of environment type. More context can be found [here](https://github.com/Automattic/studio/pull/1645#issuecomment-3167561891).

## Testing Instructions

**Testing with Pressable site**

2. Connect one of your Pressable sites to one of your Studio sites.
3. Through the Network Admin's Settings tab, change the `jetpack_callable_wp_get_environment_type` of your connected Pressable sits to any value (e.g. `local`, `staging`, `production`, `development`).
4. When reviewing the site badge in Studio:
  - if the `development` environment type is used, the `Development` badge should be rendered and use the blue color that was previously used for `Sandbox` badges
  - if the `local` environment type is used, the site should have the `Production` badge; if you click the Push / Pull button, the copy for the `production` site should be displayed in the modal (not for `local`)
  - `staging` and `production` environment types should be respected and related badges displayed in Studio UI

**Testing with WordPress.com site**

5. Connect one of your WordPress.com sites (that has staging site linked) to one of your Studio sites.
6. The `Production` and `Staging` badges should render correctly in Studio.

**General**

7. Badges displayed in the Sync modal should be consistent with badges in the list of already-connected sites on the Sync tab.

## Pre-merge Checklist


src/modules/sync/lib/environment-utils.ts
```diff
@@ -2,16 +2,14 @@ import { __ } from '@wordpress/i18n';
...
+	return connectedSite.environmentType ?? 'production';
 };
 
 export const getEnvironmentLabel = ( type: string ): string => {
 	const labels: Record< string, string > = {
 		staging: __( 'Staging' ),
-		sandbox: __( 'Sandbox' ),
+		development: __( 'Development' ),
+		local: __( 'Local' ),
 		production: __( 'Production' ),
```
ivan-ottinger: We are now supporting all four environment types provided by core: https://developer.wordpress.org/reference/functions/wp_get_environment_type/.

src/modules/sync/components/environment-badge.tsx
```diff
@@ -22,7 +22,7 @@ export function EnvironmentBadge( { type, selected, className }: EnvironmentBadg
 
 		const classes: Record< string, string > = {
 			production: 'bg-a8c-green-5 text-a8c-green-80',
-			sandbox: 'text-sandbox-text bg-sandbox-bg',
+			development: 'text-development-text bg-development-bg',
 			staging: 'text-a8c-yellow-80 bg-a8c-yellow-10',
 		};
```
ivan-ottinger: The `local` environment type will just use the default gray color, but we could customize it here as well.

src/modules/sync/lib/environment-utils.ts
```diff
@@ -2,16 +2,14 @@ import { __ } from '@wordpress/i18n';
...
-	return connectedSite.isStaging ? 'staging' : 'production';
+	return connectedSite.environmentType ?? 'production';
 };
 
 export const getEnvironmentLabel = ( type: string ): string => {
 	const labels: Record< string, string > = {
 		staging: __( 'Staging' ),
-		sandbox: __( 'Sandbox' ),
+		development: __( 'Development' ),
+		local: __( 'Local' ),
```
sejas: We don't expect to have any remote Local environment, but it's a valid core value that we should support as users or plugins could set that value. 👍

sejas (commented):
Great work, Ivan! The code looks clean and simple. I tested it, and it works well except for an edge case. If a WPcom user set a custom environment type, Studio displays different labels in the sites modal and in the connected sites screens.

I think we should be consistent and display always the environment type set by the user.

**Happy Path works great**

Screenshot 2025-08-06 at 15 08 28



**Inconsistent results for WPcom sites with custom labels**

Screenshot 2025-08-06 at 15 29 38

Screenshot 2025-08-06 at 15 29 42

It seems we should update the code in https://github.com/Automattic/studio/blob/669d1469ad4484f07440e18f6a98f95fdc50a322/src/modules/sync/components/sync-sites-modal-selector.tsx#L287-L297

src/modules/sync/lib/environment-utils.ts
ivan-ottinger: Yes, I just wanted to ensure it will get translated as one of the values supported by core.

ivan-ottinger:
Thank you for the review, Antonio!

> I tested it, and it works well except for an edge case. If a WPcom user set a custom environment type, Studio displays different labels in the sites modal and in the connected sites screens.

> I think we should be consistent and display always the environment type set by the user.

Indeed! That's a great catch! Thank you for noticing that. I will address it tomorrow morning. 🙂

wojtekn:
What happens if the user sets this variable to an unsupported environment type?

src/modules/sync/lib/environment-utils.ts
```diff
@@ -2,16 +2,14 @@ import { __ } from '@wordpress/i18n';
 import { SyncSite } from 'src/hooks/use-fetch-wpcom-sites/types';
 
 export const getSiteEnvironment = ( connectedSite: SyncSite ): string => {
-	if ( connectedSite.isPressable ) {
-		return connectedSite.environmentType ?? 'production';
-	}
-	return connectedSite.isStaging ? 'staging' : 'production';
+	return connectedSite.environmentType ?? 'production';
```
sejas: I modified this logic to support only the four environment types. See:

https://github.com/Automattic/studio/pull/1648/files#diff-d7212615886d109b1f34adab097f47d9877e00f7e088fbef173bc190cb37ab8a

sejas:
> What happens if the user sets this variable to an unsupported environment type?

If the environment type is not supported, I suggest that Studio will display the badge as Production.
I added the check in my PR https://github.com/Automattic/studio/pull/1648/files#diff-d7212615886d109b1f34adab097f47d9877e00f7e088fbef173bc190cb37ab8a
```tsx
const EnvironmentSchema = z.enum( [ 'production', 'staging', 'development', 'local' ] );
export type EnvironmentType = z.infer< typeof EnvironmentSchema >;

export const getSiteEnvironment = ( site: SyncSite ): EnvironmentType => {
	const parsed = EnvironmentSchema.safeParse( site.environmentType );
	return parsed.success ? parsed.data : 'production';
};
```

sejas:
@ivan-ottinger , I've merged https://github.com/Automattic/studio/pull/1648

ivan-ottinger:
> @ivan-ottinger , I've merged #1648

Thank you, Antonio. This PR is now ready for another review round. I have also updated the PR title, description and testing instructions.

src/modules/sync/components/sync-sites-modal-selector.tsx
```diff
@@ -212,6 +215,16 @@ function SiteItem( {
...
 
+	const renderStagingSiteBadge = () => {
+		const stagingSiteId = site.stagingSiteIds[ 0 ];
+		const stagingSite = allSites.find( ( s ) => s.id === stagingSiteId );
+
+		if ( ! stagingSite ) {
+			return null;
+		}
+		return <EnvironmentBadge type={ getSiteEnvironment( stagingSite ) } selected={ isSelected } />;
+	};
```
sejas: Nit: I think we can try to find the staging site in the main scope, and then render the badge if it exists. Something like this:

```diff
diff --git a/src/modules/sync/components/sync-sites-modal-selector.tsx b/src/modules/sync/components/sync-sites-modal-selector.tsx
index 05148bdb..1d4fb4c9 100644
--- a/src/modules/sync/components/sync-sites-modal-selector.tsx
+++ b/src/modules/sync/components/sync-sites-modal-selector.tsx
@@ -214,16 +214,7 @@ function SiteItem( {
 	const isUnsupported = site.syncSupport === 'unsupported';
 	const isPressable = site.isPressable;
 	const isDisabled = isDeleted || isUnsupported || needsUpgrade || isMissingPermissions;
-
-	const renderStagingSiteBadge = () => {
-		const stagingSiteId = site.stagingSiteIds[ 0 ];
-		const stagingSite = allSites.find( ( s ) => s.id === stagingSiteId );
-
-		if ( ! stagingSite ) {
-			return null;
-		}
-		return <EnvironmentBadge type={ getSiteEnvironment( stagingSite ) } selected={ isSelected } />;
-	};
+	const stagingSite = allSites.find( ( s ) => s.id === site.stagingSiteIds[ 0 ] );
 
 	return (
 		<div
@@ -297,14 +288,9 @@ function SiteItem( {
 			</div>
 			{ isSyncable && (
 				<div className="flex gap-2">
-					{ ! isPressable && (
-						<>
-							<EnvironmentBadge type={ getSiteEnvironment( site ) } selected={ isSelected } />
-							{ site.stagingSiteIds.length > 0 && renderStagingSiteBadge() }
-						</>
-					) }
-					{ isPressable && (
-						<EnvironmentBadge type={ getSiteEnvironment( site ) } selected={ isSelected } />
+					<EnvironmentBadge type={ getSiteEnvironment( site ) } selected={ isSelected } />
+					{ stagingSite && (
+						<EnvironmentBadge type={ getSiteEnvironment( stagingSite ) } selected={ isSelected } />
 					) }
 				</div>
 			) }
```

sejas (approved):
@ivan-ottinger , I tested the PR and works great. I see the real Environment type in the modal and when the site is connected.
I left a suggestion about removing `renderStagingSiteBadge`. It's not a strong opinion, so let me know what you think.


Screenshot 2025-08-07 at 15 17 48
Screenshot 2025-08-07 at 15 17 42

wojtekn:
@ivan-ottinger @sejas the part with supporting "local" as a type of remote environment seems a bit weird.

Should we diverge from core here and support only labels that make sense as labels used by WordPress.com or Pressable (production, staging, development), and ignore the rest? Or take a step back and only support labels that we show in the WordPress.com sites dashboard, so we would go back to using production and staging?

ivan-ottinger:
> [...] the part with supporting "local" as a type of remote environment seems a bit weird.

I see what you mean. 

> Should we diverge from core here and support only labels that make sense as labels used by WordPress.com or Pressable (production, staging, development), and ignore the rest?

If I am understand your point of view correctly, in the case of `local` environment type (or some other non-core value) we would not display any badge, right?

In that case we would need to rethink [the copy displayed on the sync modal](https://github.com/Automattic/studio/pull/1648#pullrequestreview-3095961757) as well though. Instead of `Push to Development` it could say `Push to {site-title / site-url}`

Curious to hear what Antonio thinks as well.

wojtekn:
@ivan-ottinger maybe we should follow what @sejas [proposed above](https://github.com/Automattic/studio/pull/1645#issuecomment-3160781158)? 

- production -> Production
- staging -> Staging
- development -> Development
- anything else -> Production

ivan-ottinger:
> @ivan-ottinger maybe we should follow what @sejas [proposed above](https://github.com/Automattic/studio/pull/1645#issuecomment-3160781158)?
> 
> * production -> Production
> * staging -> Staging
> * development -> Development
> * anything else -> Production

Yes, that should work. 👍🏼 

---

There's one issue I have observed with relying on the `environment_type` for WordPress.com sites: 

With the current staging site sync logic (on the WordPress.com side), the `environment_type` is synced as well. As a result we end up with both sites having the `Production` badge (as an example):

Markup on 2025-08-08 at 13:20:25

We could adjust the `getSiteEnvironment` change introduced in https://github.com/Automattic/studio/pull/1648/files#diff-d7212615886d109b1f34adab097f47d9877e00f7e088fbef173bc190cb37ab8aR8-R10. to still rely on `connectedSite.isStaging` for WordPress.com sites.

wojtekn:
Good catch @ivan-ottinger . It makes sense to have such a series of fallbacks indeed.

src/modules/sync/components/sync-sites-modal-selector.tsx
ivan-ottinger: Thanks, Antonio! This is a great idea! Much more simple. I have applied it in https://github.com/Automattic/studio/pull/1645/commits/30a3fab2e46a6bd82c8f303ee893c5534df60f8b.

sejas:
I see both solutions can be a bit confusing for some users, but **keeping the old logic that limited WPcom to only use production and staging seems the most reasonable.**
Just leaving here why the other approach also made sense. It seems that the environment type is displayed in Calypso overview for Agency sites. See the screenshots p1752755426181769/1752734657.162979-slack-C06DRMD6VPZ

ivan-ottinger:
> I see both solutions can be a bit confusing for some users, but keeping the old logic that limited WPcom to only use production and staging seems the most reasonable.

Yes, I agree that neither of the two solutions is 100% ideal.

I have updated the PR (including the title, description and testing instructions). Feel free to take another look if you like.

src/modules/sync/components/sync-sites-modal-selector.tsx
```diff
@@ -286,9 +290,12 @@ function SiteItem( {
...
 						<>
-							<EnvironmentBadge type="production" selected={ isSelected } />
-							{ site.stagingSiteIds.length > 0 && (
-								<EnvironmentBadge type="staging" selected={ isSelected } />
+							<EnvironmentBadge type={ getSiteEnvironment( site ) } selected={ isSelected } />
+							{ stagingSite && (
+								<EnvironmentBadge
+									type={ getSiteEnvironment( stagingSite ) }
+									selected={ isSelected }
+								/>
```
sejas: I see the older code can work too, and it would remove the need of passing `allSites`. However, the benefit of the new approach is that we centralize all the logic of identifying the environment inside a single function `getSiteEnvironment`.

sejas (approved):
@ivan-ottinger , great work! 🚢 
WPcom sites will display only Production and Staging based on isStaging.
Pressable sites will display any valid Studio environment type (Production, Staging, Development).

src/modules/sync/components/sync-sites-modal-selector.tsx
ivan-ottinger: Good catch! At the end I decided to restore the original code as I think it is more readable and simple: https://github.com/Automattic/studio/pull/1645/commits/dd7577d1d997d0db88bb909c741ec17a25788259.

PR #1651, opened 2025-08-07
Use only reachable tags to generate version

wojtekn:
## Related issues

- Related discussion: p1754569879531659/1752837980.840429-slack-CC7L49W13

## Proposed Changes

I propose to update the build script that generates the version to use only **reachable** tags to fix an edge case in which the build should run for an older tag, but reads a newer tag from the remote.

The chain of events looked as follows:
1. We merged the 583fcaf commit that still should read v1.5.5 as the latest stable tag
2. Within a minute, we merged the 4ffa1a2 commit and tagged it with v1.5.6
3. Build started for the 583fcaf but was able to fetch the latest tag as v1.5.6 at this point instead of 1.5.5.
4. The build failed as it tried to read 1.5.6.0 artifact but 1.5.5.0 was the valid one

## Testing Instructions

1. Switch to the commit:
```
gco 583fcaf4d2e4cb268d9670b0461717aa6d5a4d59
```
2. Apply changes from this branch
3. Run script:
```
node ./scripts/prepare-dev-build-version.mjs  
```
4. Confirm the version was generated as `1.5.5-dev65` and not `1.5.6-dev0`

## Pre-merge Checklist

### DIFF
--- a/scripts/lib/git-utils.mjs
+++ b/scripts/lib/git-utils.mjs
@@ -11,7 +11,7 @@ export const getLatestTag = () => {
 
 		// Get the latest reachable release tag that is not a beta tag
 		const latestReleaseTag = child_process
-			.execSync( 'git describe --tags --abbrev=0 --match \'v*\' --exclude \'*-beta*\'' )
+			.execSync( 'git describe --tags --abbrev=0 --match "v*" --exclude "*-beta*"' )
 			.toString()
 			.trim();
 
--- a/scripts/lib/git-utils.mjs
+++ b/scripts/lib/git-utils.mjs
@@ -8,7 +8,7 @@ export const getLatestTag = () => {
 	try {
 		// List all tags sorted by version, then filter for release tags only
 		const tags = child_process
-			.execSync( 'git tag --sort=-v:refname' )
+			.execSync( 'git tag --sort=-v:refname --merged=HEAD' )
 			.toString()
 			.trim()
 			.split( '\n' );
--- a/scripts/lib/git-utils.mjs
+++ b/scripts/lib/git-utils.mjs
@@ -6,6 +6,9 @@ import * as child_process from 'child_process';
  */
 export const getLatestTag = () => {
 	try {
+		// Fetch all tags to ensure they're available
+		child_process.execSync( 'git fetch --tags --force', { stdio: 'pipe' } );
+
 		// Get the latest reachable release tag that is not a beta tag
 		const latestReleaseTag = child_process
 			.execSync( 'git describe --tags --abbrev=0 --match \'v*\' --exclude \'*-beta*\'' )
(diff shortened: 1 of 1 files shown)


wojtekn:
Alternatively, we could use `git describe --tags --abbrev=0 --match 'v*' --exclude '*-beta*'` one liner and remove the whole `find` part, but I'm unsure if it looks cleaner. Thoughts?

Thanks @AliSoftware for the ideas!

AliSoftware:
One difference worth noting is that:
1. `git tag --sort=-v:refname --merged=HEAD` (then filtering out the ones containing `-` then keeping just the first) will return the reachable non-beta tag _with the highest version_
2. `git describe --tags --abbrev=0 --match 'v*' --exclude '*-beta*'` will return the _latest_ reachable non-beta tag 

This means that, in theory, if you create a tag `v1.7.0` and then later create a tag `v1.6.6`, then option 1 will return `v1.7.0` because has the higher value, while option 2 will return `v1.6.6` because it's the more recent tag in the git history.

Of course in practice such scenario is unlikely to happen, as you'd expect your tags for stable versions to always be increasing, so having `v1.6.6` be created later in the git history than `v1.7.0` seems unlikely.
But still, personally I feel it makes more sense semantically to ask git for "the latest reachable tag (that is not a beta tag)" rather than "amongst all the reachable git tags, the one that has the highest version value in its name", right?

Because in practice what you really want to achieve when computing that version to put in the `package.json` file is get the latest stable tag, and then count how many commit it has been since that last release, to derive the new version name to set in the `package.json`, right? 🤷

gcsecsey (commented):
Thanks for investigating this scenario further and creating a fix for it! When testing with the commit in the testing steps, I got a new `-dev65` version generated as described.

Image

IMO using `git describe --tags --abbrev=0 --match 'v*' --exclude '*-beta*'` would be cleaner both because it would let us avoid the extra `find`, and it's also semantically closer to what we'd like to achieve here as @AliSoftware pointed out.

I tested using this command as well and got the same `-dev65` version.

wojtekn:
Thanks for feedback @AliSoftware and @gcsecsey . I've updated PR to use `git describe`.

AliSoftware:
[CI failure](https://buildkite.com/automattic/studio/builds/7450/steps/canvas?jid=01988564-961b-450c-a75e-d83c0b761b21#01988564-961b-450c-a75e-d83c0b761b21/401-403)
```
fatal: No names found, cannot describe anything.
```
I wonder if you might need to `git fetch --tags --force` first to ensure all remote tags are fetched in the local working copy and you don't miss any? 🤔

wojtekn:
@AliSoftware, it looks like it was not about missing tags but escaped quotes, which were breaking the command on Windows. I changed those to double quotes what fixed the issue, but I left the tags pull as it still makes sense to have it.

AliSoftware (approved):

PR #1657, opened 2025-08-08
Keep export state to let user know what was completed

wojtekn:
## Related issues

- Fixes STU-667

## Proposed Changes

I propose to keep the export state to let the user know what was completed. This is consistent with Pull and Push actions now.

Screenshot 2025-08-08 at 15 58 23

Screenshot 2025-08-08 at 15 58 14

Screenshot 2025-08-08 at 15 58 07

## Testing Instructions

1. Create two Studio sites
2. Run the entire site export
3. Confirm that progress is displayed and when it completes, success message is preserved
4. Switch to another Studio site
5. Run the site database export
6. Confirm that progress is displayed and when it completes, success message is preserved
7. Switch to the previous site, confirm that the state is still preserved
8. Use Clear to remove the success state

## Pre-merge Checklist


src/components/content-tab-import-export.tsx
```diff
@@ -23,6 +25,34 @@ interface ContentTabImportExportProps {
 	selectedSite: SiteDetails;
 }
 
+const ExportClear = ( {
```
sejas: I see the markup of this component is the same as `SyncPullPushClear`. https://github.com/Automattic/studio/blob/28deea46e84483bb4c27b0bcd9c801ea7fb2eb72/src/modules/sync/components/sync-pull-push-clear.tsx#L7

What do you think if we rename and reuse the component?

sejas (approved):
Thanks for persisting the export success message.
It works as expected, I tested it by exporting database and full site in different sites and I confirm the success message is persisted, and clears correctly when the user clicks on clear.

We can consider combining `SyncPullPushClear` and `ExportClear` into a generic component.

https://github.com/user-attachments/assets/4cec1332-46e2-40ff-8aee-e9248d4bcc35

src/components/content-tab-import-export.tsx
wojtekn: Good catch, done in 15ae4ed9a9eed22de6a793518acd04a8c343dd14

sejas: @wojtekn, I think it would be better if we move the component to a separate file. Something like `src/components/clear-action.tsx`

sejas (approved):
@wojtekn , for some reason the `clear-action.tsx` file was missing. I just added it https://github.com/Automattic/studio/pull/1657/commits/6cd6972a87d63f4561fd5b955053be30adbcf8ab and tested the export and pull. Both work correctly. Thanks for making the changes ✨ .

wojtekn:
Thanks for adding the file @sejas . It seems it was missing as `git status` didn't show the untracked file and I missed it in the commit. I checked it more, and found out that I had `/src` in my global `.gitignore` for some reason.

PR #1668, opened 2025-08-12
Update wordpress i18n library to 6.1.0

wojtekn:
## Related issues

- Relates to https://github.com/Automattic/studio/pull/1662

## Proposed Changes

I propose updating @wordpress/i18n to 6.1.0. I opened a PR, as Dependabot won't be able to upgrade the package since it requires several changes to the translation function and sprintf usage.

## Testing Instructions

1. Test if translations work fine, especially those that use string or decimal placeholders.
2. Confirm that the build works fine.

## Pre-merge Checklist

### DIFF
--- a/src/modules/add-site/hooks/use-stepper.ts
+++ b/src/modules/add-site/hooks/use-stepper.ts
@@ -10,12 +10,12 @@ interface StepperStep {
 }
 
 interface StepperConfig {
-	onBlueprintContinue?: () => void;
-	onBackupContinue?: () => void;
-	onCreateSubmit?: ( event: FormEvent ) => void;
-	canSubmitBlueprint?: boolean;
-	canSubmitBackup?: boolean;
-	canSubmitCreate?: boolean;
+	onBlueprintContinue: () => void;
+	onBackupContinue: () => void;
+	onCreateSubmit: ( event: FormEvent ) => void;
+	canSubmitBlueprint: boolean;
+	canSubmitBackup: boolean;
+	canSubmitCreate: boolean;
 }
 
 interface StepperContext {
--- a/src/modules/add-site/hooks/use-stepper.ts
+++ b/src/modules/add-site/hooks/use-stepper.ts
@@ -19,7 +19,11 @@ interface StepperConfig {
 }
 
 interface StepperContext {
-	flow?: 'blueprint' | 'backup' | 'create';
+	flow: 'blueprint' | 'backup' | 'create';
+	steps: StepperStep[];
+}
+
+interface UseStepper {
 	steps: StepperStep[];
 	isVisible?: boolean;
 	actionButton?: {
@@ -30,7 +34,7 @@ interface StepperContext {
 	canSubmit?: boolean;
 }
 
-export function useStepper( config?: StepperConfig ): StepperContext {
+export function useStepper( config?: StepperConfig ): UseStepper {
 	const { __ } = useI18n();
 	const { location } = useNavigator();
 
--- a/src/modules/add-site/hooks/use-stepper.ts
+++ b/src/modules/add-site/hooks/use-stepper.ts
@@ -10,12 +10,12 @@ interface StepperStep {
 }
 
 interface StepperConfig {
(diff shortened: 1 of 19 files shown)


src/modules/add-site/hooks/use-stepper.ts
```diff
@@ -18,32 +19,33 @@ interface StepperConfig {
...
-	isVisible: boolean;
+	isVisible?: boolean;
 	actionButton?: {
 		label: string;
 		isVisible: boolean;
 	};
-	onSubmit: () => void;
-	canSubmit: boolean;
+	onSubmit?: () => void;
+	canSubmit?: boolean;
```
sejas: Let's avoid making this elementos optional, and let's use a different type for `stepperConfig`.
```suggestion
	onSubmit: () => void;
	canSubmit: boolean;
```

See my comment below.

src/modules/add-site/hooks/use-stepper.ts
```diff
@@ -18,32 +19,33 @@ interface StepperConfig {
...
 }
 
 export function useStepper( config?: StepperConfig ): StepperContext {
 	const { __ } = useI18n();
 	const { location } = useNavigator();
 
-	const stepperConfig = useMemo( () => {
-		const blueprintSteps = [
+	const stepperConfig = useMemo( (): StepperContext | null => {
```
sejas: `stepperConfig` always contains `flow` and `steps`. We can use Pick to extract those from the context type:

```suggestion
	const stepperConfig = useMemo( (): Pick< StepperContext, 'flow' | 'steps' > | null => {
```

src/stores/snapshot-slice.ts
```diff
@@ -461,18 +461,18 @@ window.ipcListener.subscribe( 'snapshot-success', ( event, payload ) => {
...
 			title: operation.snapshotName,
-			body: sprintf( __( "Preview site '%s' has been updated." ), operation.snapshotUrl ),
+			body: sprintf( __( "Preview site '%s' has been updated." ), operation.snapshotUrl || '' ),
 		} );
 	} else if ( operation.type === 'delete' ) {
 		if ( ! bulkOperation ) {
 			getIpcApi().showNotification( {
 				title: operation.snapshotName,
-				body: sprintf( __( "Preview site '%s' has been deleted." ), operation.snapshotUrl ),
+				body: sprintf( __( "Preview site '%s' has been deleted." ), operation.snapshotUrl || '' ),
```
sejas: Nit: We could declare snapshotUrl only once:
```ts
	const { snapshotUrl = '' } = operation;
	if ( operation.type === 'create' ) {
			title: operation.snapshotName,
			body: sprintf( __( "Preview site '%s' has been created." ), snapshotUrl ),
		} );
	} else if ( operation.type === 'update' ) {
			title: operation.snapshotName,
			body: sprintf( __( "Preview site '%s' has been updated." ), snapshotUrl ),
		} );
	} else if ( operation.type === 'delete' ) {
		if ( ! bulkOperation ) {
				title: operation.snapshotName,
				body: sprintf( __( "Preview site '%s' has been deleted." ), snapshotUrl ),
			} );
```

sejas (approved):
It works as expected. I tested a few strings with numbers.

I added a couple of suggestions: one about not changing the `StepperContext` type, and another minor nitpick.

It would be great to update the translations in GlotPress after merging this PR.

Screenshot 2025-08-13 at 18 05 25

src/modules/add-site/hooks/use-stepper.ts
wojtekn: Actually, I should fix this constant name, too, as it should be `stepperContext` here and not `stepperConfig`. The config is passed to this hook, and context is returned.

I untangled the types of context and hook return value - using the same type for both caused those needs to use a partial type.

src/stores/snapshot-slice.ts
wojtekn: Thanks, makes sense.

sejas (approved):
Thanks for applying the changes and untangling the types.

PR #1670, opened 2025-08-13
Ensure export error is not duplicated

wojtekn:
## Related issues

- Fixes STU-700

## Proposed Changes

I propose to change the condition to ensure that the export error is not duplicated.

## Testing Instructions

1. Create Studio site
2. Enable Blueprints feature flag (to trigger the export error)
3. Remove database constants from the `wp-config.php` field
4. Restart site (Playground CLI will re-add missing constant)
5. Export sites database
6. Confirm that the error message is displayed once

|**Before**|**After**|
|---|---|
| Screenshot 2025-08-13 at 10 57 01 | Screenshot 2025-08-13 at 10 53 24|

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-import-export.tsx
+++ b/src/components/content-tab-import-export.tsx
@@ -79,7 +79,7 @@ export const ExportSite = ( {
 							<div className="text-a8c-gray-70 a8c-body">{ currentProgress.statusMessage }</div>
 						</>
 					) }
-					{ isExportCompleted && (
+					{ isExportCompleted && ! isExportError && (
 						<ClearAction onClick={ handleClearExport }>
 							{ currentProgress.statusMessage }
 						</ClearAction>


gcsecsey (commented):
While testing this, I ran into another issue. I deleted all DB constants from `wp-config` because I was expecting [wp-now](https://github.com/Automattic/studio/blob/4136650226add4da4a295bdf5c5132ba68f6a2ba/vendor/wp-now/src/wp-now.ts#L540-L556) to add these back. However, when restarting the site, `wp-config` is also updated with the `DB_NAME` constant set to `wordpress`.

Image

This causes the export to fail with this alert:

Image

I can also reproduce this on trunk, when deleting just the DB_NAME constant from `wp-config`, so I'll add an issue to track this.

gcsecsey (approved):
The UI updates work as described for me, the export error is only rendered once:

| trunk | thus branch |
|--------|--------|
| Image | Image |

gcsecsey:
> I can also reproduce this on trunk, when deleting just the DB_NAME constant from `wp-config`, so I'll add an issue to track this.

I found [STU-691: Studio Blueprints: DB_NAME in wp-config](https://linear.app/a8c/issue/STU-691/studio-blueprints-db-name-in-wp-config) already tracking this 👍

PR #1672, opened 2025-08-13
Add tooltip when rewind_id is not available for selective sync pull dialog

sejas:
## Related issues

- Fixes STU-701
- Equivalent PR to https://github.com/Automattic/wp-calypso/pull/105158.

## Proposed Changes

- Add tooltip and disable checkboxes when rewind_id is not available.

## Testing Instructions

- Sandbox your API
- Set your sandbox in read only mode or add custom code to return an error in the `/studio-app/sync/get-latest-rewind-id` endpoint.
- Run `npm run start`
- Go to the Sync tab
- Connect a site
- Click pull
- Observe checkboxes are disabled and the tooltip appears in the Sync Modal 

Screenshot 2025-08-13 at 17 17 15

## Pre-merge Checklist

### DIFF
--- a/src/modules/sync/components/sync-dialog.tsx
+++ b/src/modules/sync/components/sync-dialog.tsx
@@ -8,9 +8,11 @@ import { ArrowIcon } from 'src/components/arrow-icon';
 import Button from 'src/components/button';
 import { RightArrowIcon } from 'src/components/icons/right-arrow';
 import Modal from 'src/components/modal';
+import { Tooltip } from 'src/components/tooltip';
 import { TreeView, TreeNode, updateNodeById } from 'src/components/tree-view';
 import { SYNC_OPTIONS } from 'src/constants';
 import { useContentFolders } from 'src/hooks/use-content-folders';
+import { cx } from 'src/lib/cx';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { getLocalizedLink } from 'src/lib/get-localized-link';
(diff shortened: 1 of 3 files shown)


src/stores/sync/sync-api.ts
```diff
@@ -81,7 +81,7 @@ export const syncApi = wpcomApi.injectEndpoints( {
 
 				throw new Error( 'Failed to fetch latest rewind ID' );
 			},
-			keepUnusedDataFor: 60, // Cache for 1 minute
+			keepUnusedDataFor: 0, // Avoid caching the response
```
nightnei, line 84: Oh, really 👍

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -148,6 +154,12 @@ export function SyncDialog( {
 		syncTo = localSiteName;
 		syncFromText = remoteSite.name;
 		syncToText = localSite.name;
+		tooltipNoRewindId = createInterpolateElement(
```
nightnei, line 157: Was going to propose to put it directly to the Tooltip initialization but then decided to think one more time and - indeed, it's better to keep it here 👍

nightnei (commented):
Looks great and works as expected 👍
Screenshot 2025-08-14 at 15 33 22

src/modules/sync/components/sync-dialog.tsx
sejas: Yes, we have it there so it only applies in the pull direction.

PR #1677, opened 2025-08-14
[Tooling] Tweak Unit Tests memory settings: follow-up

iangmaia:
## Related issues

Follow-up to https://github.com/Automattic/studio/pull/1675.

## Proposed Changes

- We've noticed one more failure after increasing the heap memory size with `--max-old-space-size`. This PR adds the `workerIdleMemoryLimit` jest worker setting to limit idle memory usage.

## Testing Instructions

- Make sure CI is 🟢 ; run the build a few more times to be sure it won't fail.

## Pre-merge Checklist

### DIFF
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -45,13 +45,6 @@ beforeEach( () => {
 	console.log = jest.fn();
 } );
 
-// Force garbage collection after each test
-afterAll(() => {
-	if (global.gc) {
-		global.gc();
-	}
-});
-
 if ( typeof window !== 'undefined' ) {
 	// The ipcListener global is usually defined in preload.ts
 	window.ipcListener = { subscribe: jest.fn() };
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -45,6 +45,13 @@ beforeEach( () => {
 	console.log = jest.fn();
 } );
 
+// Force garbage collection after each test
+afterAll(() => {
+	if (global.gc) {
+		global.gc();
+	}
+});
+
 if ( typeof window !== 'undefined' ) {
 	// The ipcListener global is usually defined in preload.ts
 	window.ipcListener = { subscribe: jest.fn() };
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -42,12 +42,16 @@ if (typeof window !== 'undefined') {
 
 // Silence console.log for all tests
 beforeEach( () => {
-	console.log = jest.fn();
+	// console.log = jest.fn();
 } );
 
 // Force garbage collection after each test
 afterAll(() => {
+	console.log('afterAll')
+	console.log(global.gc)
 	if (global.gc) {
+		console.log('---')
+		console.log('Running GC')
 		global.gc();
 	}
 });
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -42,16 +42,12 @@ if (typeof window !== 'undefined') {
 
 // Silence console.log for all tests
 beforeEach( () => {
-	// console.log = jest.fn();
+	console.log = jest.fn();
 } );
 
 // Force garbage collection after each test
 afterAll(() => {
-	console.log('afterAll')
-	console.log(global.gc)
 	if (global.gc) {
-		console.log('---')
-		console.log('Running GC')
 		global.gc();
 	}
 });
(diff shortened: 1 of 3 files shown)


jest-setup.ts
```diff
@@ -45,6 +45,13 @@ beforeEach( () => {
 	console.log = jest.fn();
 } );
 
+// Force garbage collection after each test
+afterAll(() => {
+	if (global.gc) {
```
twstokes: Did we want to garbage collect after all the tests or each test? The comment is saying each test but `afterAll` is being called here (rather than `afterEach`).

iangmaia: I've actually just tested it [here](https://buildkite.com/automattic/studio/builds/7587#0198bd0f-f663-41be-a7a9-2de3a05cbdb6/524-529) to be sure: the lifecycle worked as I expected (so it runs "after all unit tests in each file / suite", but indeed the comment was maybe ambiguous), but it turns out `global.gc` is undefined? 😓 I've then reverted that change, which wasn't the most elegant thing in the first place.

Anyway, hopefully the other measures (in this PR, `workerIdleMemoryLimit`) will already improve things and with the added monitoring in heap size and potential memory leaks we can further improve memory usage if needed.

twstokes: > but it turns out `global.gc` is undefined? 😓 I've then reverted that change, which wasn't the most elegant thing in the first place.

You may have spotted this already, but it looks like `--expose-gc` [would possibly need to be called](https://stackoverflow.com/a/30654451).

jest.config.ts
```diff
@@ -34,4 +35,5 @@ module.exports = {
 	globalSetup: '<rootDir>/jest-global-setup.ts',
 	setupFilesAfterEnv: [ '<rootDir>/jest-setup.ts' ],
 	watchPlugins: [ 'jest-watch-typeahead/filename', 'jest-watch-typeahead/testname' ],
+	logHeapUsage: true,
```
twstokes, line 38: Will it be beneficial to keep this enabled long term, or is this just for troubleshooting?

package.json
```diff
@@ -30,7 +30,7 @@
 		"cli:watch": "webpack --config webpack.cli.config.ts --watch",
 		"lint": "eslint --ext .ts,.tsx,.js,.jsx,.mjs {cli,common,src}",
 		"format": "prettier . --write",
-		"test": "cross-env NODE_OPTIONS='--no-deprecation --max-old-space-size=16384' jest",
+		"test": "cross-env NODE_OPTIONS='--no-deprecation --max-old-space-size=16384' jest --detectOpenHandles",
```
twstokes: I saw in the docs that [this isn't great for performance](https://jestjs.io/docs/cli#--detectopenhandles) – are we planning to keep it enabled long term?

jest-setup.ts
iangmaia: Ah, nice one, I hadn't seen it 👍 though I think it's probably a good idea to avoid the call if we can.
We have plenty of heap memory and these test workers should be as independent as possible, do not leak etc, so I'm thinking it's a better strategy to control and limit the memory usage in other ways.

package.json
iangmaia: I wanted to see at least one run with this flag but eventually missed removing it. Reverted it on a20fce760cf521d7dd87bd3334b7f44be77b0377 👍

iangmaia: Though after running without it, I still see this message:
> A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.

cc @wojtekn just in case.

jest.config.ts
iangmaia: In principle I was actually considering keeping this `true`.
It doesn't add a lot of noise, though one can still argue it adds _some_ noise when most of the times we don't care about it:
```console
...
PASS cli/commands/preview/tests/list.test.ts (6.817 s, 1269 MB heap size)
PASS src/components/tests/ai-clear-history-reminder.test.tsx (10.317 s, 574 MB heap size)
PASS src/about-menu/tests/open-about-menu.test.tsx (6.838 s, 539 MB heap size)
...
```

twstokes (approved):
Approving to unblock, makes sense to me. 👍

PR #1678, opened 2025-08-14
Submit site creation on Enter and add tooltip to disabled Blueprints

nightnei:
## Related issues

- Fixes STU-706

## Proposed Changes
It's follow-up to https://github.com/Automattic/studio/pull/1630

1. We add tooltip to disabled Blueprints button when it's offline
2. We add oportunity to initiate creation of a new site with the help of Enter button inside "Site name" field

## Testing Instructions
1. `ENABLE_BLUEPRINTS=true npm start`
4. Click on "Add site"
5. Disable wifi
6. Assert that `Start from a blueprint` button is disabled and hover on it shows tooltip
7. Click on "Create a site"
8. On the new screen immidiatelly click Enter (while cursor is inside "Site name" field)
9. Assert that creation of a new site is started

### DIFF
--- a/src/components/site-form.tsx
+++ b/src/components/site-form.tsx
@@ -301,7 +301,16 @@ export const SiteForm = ( {
 			<div className="flex flex-col">
 				<label className="flex flex-col gap-1.5 leading-4 mb-6">
 					<span className="font-semibold">{ __( 'Site name' ) }</span>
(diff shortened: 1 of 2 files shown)


ivan-ottinger (approved):
The changes look good and work as expected. 👌🏼 

Pressing the `Enter` when creates the site as expected and the tooltip in the offline work renders correctly:

Markup on 2025-08-15 at 16:04:37

PR #1696, opened 2025-08-20
Fix using localhost instead of 127.0.0.1

nightnei:
## Related issues

- Fixes STU-703

## Proposed Changes
After introducing playground CLI we started seeing 127.0.0.1 insterad iof localhost for all sites in browser. It was mistake, localhost shoudl be used, as previously with wp-now.

## Testing Instructions
1. Run `ENABLE_BLUEPRINTS=true npm start`
2. Create a site
3. Click on "Open site" or just open `http://localhost:8881` in browser
4. Assert that you wasn't redirected to `http://127.0.0.1:8881/`


ivan-ottinger (approved):
The proposed change looks good and works as expected. 👌🏼 

I can see the sites loading at `localhost` instead of `127.0.0.1`.

PR #1713, opened 2025-08-27
Limit number of blueprint categories displayed in UI

bcotrim:
## Related issues

- Fixes STU-722

## Proposed Changes

- Limits the number of categories displayed in the UI to 3
- Adds a "+X More" label with a tooltip for the remaining categories 

image

## Testing Instructions

Apply this patch

```typescript
  // In src/modules/add-site/components/blueprints.tsx around line 134
  const categories = ( item.blueprint.meta?.categories || [] ).filter(
      ( category ) => category !== 'Studio'
  ).concat( [
      'E-commerce', 'Portfolio', 'Blog', 'Business',
      'Photography', 'Restaurant', 'Real Estate',
      'Health & Fitness', 'Education', 'Technology',
      'Fashion', 'Travel', 'Music', 'Non-profit', 'Entertainment'
  ] );
  ```

- Navigate to "Add Site" > "Create from blueprint"
- Confirm the categories are limited to 3 with a '+X more' indicator

## Pre-merge Checklist

### DIFF
--- a/src/modules/add-site/components/blueprints.tsx
+++ b/src/modules/add-site/components/blueprints.tsx
@@ -163,8 +163,8 @@ export default function AddSiteBlueprint( {
 								</Text>
 							) ) }
 							{ remainingCount > 0 && (
-								<Tooltip 
-									text={ categories.slice( maxCategoriesToShow ).join( ', ' ) } 
+								<Tooltip
+									text={ categories.slice( maxCategoriesToShow ).join( ', ' ) }
 									delay={ 200 }
 									position="top right"
 									className="max-w-xs"
--- a/src/modules/add-site/components/blueprints.tsx
+++ b/src/modules/add-site/components/blueprints.tsx
@@ -352,7 +352,7 @@ export default function AddSiteBlueprint( {
 				) }
 			</HStack>
 
-			<div className="w-full px-3 [&_.dataviews-view-grid]:!grid [&_.dataviews-view-grid]:!grid-cols-3 [&_.dataviews-view-grid]:!gap-4 [&_.components-badge]:!bg-transparent [&_.components-badge]:!p-0">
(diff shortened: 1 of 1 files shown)


epeicher (approved):
Thanks for the changes Bernardo! I can confirm that I can see the rest of the categories with the tooltip as expected, but I have observed some missalignment when that extra chip is added. Maybe this can be solved with some top-align or some alternative?

CleanShot 2025-08-27 at 18 39 43@2x

Anyway, I don't think this is a blocker for this task, as it fixes the current situation, so it could be covered as a follow-up.

bcotrim:
@epeicher thanks for the review.
~I also noticed that issue. It already exists on `trunk` and I fixed it on this other PR: https://github.com/Automattic/studio/pull/1715/files#diff-0d0726eb32518342bd289e8d488379ad8ee5895ee9a14c1194dd9a847feef607L114~

It's actually another issue, let me look into it

ivan-ottinger (approved):
Nice! Apart from the alignment issue, that Roberto mentioned already, I did not notice anything else. The extra categories are displayed nicely in a tooltip.

ivan-ottinger:
I did this a quick test with the alignment fix in place and all is working well! 👌🏼 

Markup 2025-09-01 at 13 49 53

nightnei (changes requested):
~I have two concerns:~
~1. At the backend we have limit of 50 items, additionally we have limit here as 3 items, looks like 1 redundant limitation which just cimplicated codebase~
~2. We send via http up to 50 item, and anyway we slice them on frontend to 3 items - there is no sense to send more data over http, than we use on frontend~

~So, I think would be better to leverage that backend limitation to make it dynamic - backend expects optional parameter from frontend  - `number of items`, and if it's provided then we return just that number of items, otherwise 50.~

I realized that I mixed up things, my bad, sorry, the PR looks great 👍

nightnei (approved):

PR #1720, opened 2025-08-28
Fix handling invalid response of Blueprints

nightnei:
## Related issues

- Fixes STU-727

## Proposed Changes
With this PR we are handling and rendering erros of fetching Blueprints.
Also, I haven't found some way to handle errors from API and from `transformResponse` teh same way, types are different and they both has `data` as `unknown`. That's why we have `formatRtkError` which handles both types of errors and returns always `string`.

## Testing Instructions
Apply teh next:

``` diff
diff --git a/src/stores/wpcom-api.ts b/src/stores/wpcom-api.ts
index c89509c0..d8371aad 100644
--- a/src/stores/wpcom-api.ts
+++ b/src/stores/wpcom-api.ts
@@ -163,7 +163,7 @@ export const wpcomPublicApi = createApi( {
                        void
                >( {
                        query: () => ( {
-                               path: '/studio-app/blueprints',
+                               path: '/studio-app/blueprints1111',
                                apiNamespace: 'wpcom/v2',
                        } ),
                        transformResponse: ( response: unknown ) => {
```
1. Open Studio -> "Add site" -> From Blueprints
2. Assert that you see error `Error loading suggested blueprints: [500] No route was found matching the URL and request method.`

Apply the next:

``` diff
diff --git a/src/stores/wpcom-api.ts b/src/stores/wpcom-api.ts
index c89509c0..5195a04d 100644
--- a/src/stores/wpcom-api.ts
+++ b/src/stores/wpcom-api.ts
@@ -167,9 +167,9 @@ export const wpcomPublicApi = createApi( {
                                apiNamespace: 'wpcom/v2',
                        } ),
                        transformResponse: ( response: unknown ) => {
-                               if ( ! response || typeof response !== 'object' ) {
+                               // if ( ! response || typeof response !== 'object' ) {
                                        throw new Error( 'Invalid response format' );
-                               }
+                               // }
 
                                const responseObj = response as Record< string, unknown >;
                                const blueprints = responseObj.blueprints;
```
1. Open Studio -> "Add site" -> From Blueprints
2. Assert that you see error `Error loading suggested blueprints: Invalid response format`


epeicher (approved):
Thanks @nightnei! I have tested it, and it works as expected. I can see the errors being handled and with the expected error messages. 

|Invalid response | Invalid format|
|-------|------|
|CleanShot 2025-09-01 at 17 13 43@2x| CleanShot 2025-09-01 at 17 23 49@2x|

I would also keep an eye on real errors returned by the API as we might need to update the formatting errors, but this is a nice approach that might be cover most of them.

gcsecsey (approved):
This looks good and works well for me too. 👍 I can see the expected errors shown for the appropriate causes as described:

| loading error | invalid response |
|--------|--------|
| Image | Image |

wojtekn:
What if we switched to a more positive alert, similar to the ones Apple uses in its software?

For example, we could:
- drop red color
- say something simpler, 'Studio could not load blueprints. [Try again] or use your own blueprint'?

nightnei:
> What if we switched to a more positive alert, similar to the ones Apple uses in its software?
> 
> For example, we could:
> 
> * drop red color
> * say something simpler, 'Studio could not load blueprints. [Try again] or use your own blueprint'?

@wojtekn addressed [here](https://github.com/Automattic/studio/pull/1755)

PR #1725, opened 2025-08-29
Fix redirect to WP_SITEURL

nightnei:
## Related issues

- Fixes STU-742

## Proposed Changes
With this PR we prevent redirect from the same localhost:port to teh same localhost:port since new we have endless loop.

Why we didn't have it before - since wiuth wp-now we didn't have WP_SITEURL explicitly in DB when we used localhost. After migration to playground-cli we have it in DB so we got unexxpectedly endless loop.

## Testing Instructions
1. Create simple site
2. Create site with custom domain and SSL
3. Create site with custom domain and WITHOUT SSL
4. Assert that "Open site" and "WP admin" buttons work well for all 3 sites
5. Open `hosts` file to see which ports are used under the hood for 2 sites with custom domains: <br />Screenshot 2025-08-29 at 18 17 23
6. Open in browser `localhost:8882` and assert that you are redirected correctly to `with.wp.local`. Repeat the same for 8883.

### DIFF
--- a/src/lib/wordpress-provider/playground-cli/mu-plugins.ts
+++ b/src/lib/wordpress-provider/playground-cli/mu-plugins.ts
@@ -86,6 +86,17 @@ function getStandardMuPlugins( options: Partial< WordPressServerOptions > ): MuP
 			$current_host = $_SERVER['HTTP_HOST'] ?? '';
 
 			if ( preg_match( '/^localhost:\\d+$/', $current_host ) ) {
+				$wp_siteurl_host = parse_url( WP_SITEURL, PHP_URL_HOST );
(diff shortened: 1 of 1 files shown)


epeicher (approved):
Thanks @nightnei for the fix! I have tested the three scenarios and they all work as expected. I have also tested the redirection by using localhost and the port for the custom domains and they work well. Below there is one example with the one with SSL enabled. LGTM! :shipit: 

CleanShot 2025-08-29 at 19 29 11@2x

https://github.com/user-attachments/assets/eb63f19e-a1c3-45a6-9447-804305b06fc2

PR #1737, opened 2025-09-04
Studio: Fix php-wasm/node import error in worker threads for packaged builds

bcotrim:
## Related issues

- Fixes STU-758

## Proposed Changes

- Added @wp-playground/cli worker threads as webpack entries to bundle dependencies properly
- Used DefinePlugin to replace import.meta.dirname with ASAR-compatible path using process.resourcesPath
- Removed copying of worker thread files since they're now bundled with webpack
- Resolves critical issue where Studio DMG installations failed to start WordPress sites when installed in the Applications folder

## Testing Instructions

- `npm install`
- Build DMG package using `npm run make:dmg-arm64`
- Install Studio to Applications folder
- Create a new WordPress site and verify it starts successfully
- Verify no import errors occur in worker threads

## Pre-merge Checklist

### DIFF
--- a/webpack-loaders/fix-import-meta-dirname-loader.js
+++ b/webpack-loaders/fix-import-meta-dirname-loader.js
@@ -1,0 +1,47 @@
+/**
+ * Custom webpack loader to fix import.meta.dirname in specific packages
+ * 
+ * This loader targets @php-wasm/node and @wp-playground/cli worker threads
+ * to replace import.meta.dirname (which becomes undefined when bundled by webpack)
+ * with the correct path for ASAR archives.
+ * 
+ * This is necessary because @php-wasm/node uses import.meta.dirname to locate
+ * ICU data files (like icudt74l.dat), and when the app is packaged in an ASAR
+ * archive, these paths need to point to the correct location within the archive.
+ */
+
+const path = require('path');
+
+module.exports = function fixImportMetaDirnameLoader(source) {
+	// Only apply to specific files that need the fix
+	const shouldApplyFix = 
+		this.resourcePath.includes('@php-wasm/node') ||
+		this.resourcePath.includes('worker-thread-v1') ||
+		this.resourcePath.includes('worker-thread-v2');
+	
+	if (!shouldApplyFix) {
+		return source;
+	}
+
+	// Check if we're in production mode
+	const isProduction = this.mode === 'production';
+	
+	// Replace import.meta.dirname with the appropriate path
+	// In production: use process.resourcesPath to find the ASAR archive
+	// In development: use the absolute path to .webpack/main
+	const replacement = isProduction
+		? 'require("path").join(process.resourcesPath, "app.asar", ".webpack", "main")'
+		: JSON.stringify(path.resolve(__dirname, '..', '.webpack', 'main'));
+	
+	// Replace all occurrences of import.meta.dirname
+	const modifiedSource = source.replace(/import\.meta\.dirname/g, replacement);
+	
+	// Also handle import.meta.filename if present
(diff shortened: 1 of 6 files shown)


epeicher (commented):
Thanks @bcotrim! I have tested it by creating a new DMG and installed into my Applications folder 
CleanShot 2025-09-05 at 09 51 17@2x
Creating a site and starting it worked well.
Then I imported a backup I generated using Jetpack from a WPCOM site and when I tried to import it, I received the error that you can find in the log file [studio-20250905.0.log](https://github.com/user-attachments/files/22167512/studio-20250905.0.log)

|Error |
|-------|
|CleanShot 2025-09-05 at 09 55 18@2x|

I am not sure if it is related to this PR. I will try to investigate further, let me know if I can help with that.

webpack.main.config.ts
```diff
@@ -34,6 +34,17 @@ const extraEntries = [
...
 	},
+	// Add playground CLI worker threads as webpack entries to bundle dependencies
+	{
+		name: 'worker-thread-v1-BTJIbQLy',
+		path: './node_modules/@wp-playground/cli/worker-thread-v1-BTJIbQLy.js',
+		exportName: null,
+	},
+	{
+		name: 'worker-thread-v2-Pfv6UYF4',
+		path: './node_modules/@wp-playground/cli/worker-thread-v2-Pfv6UYF4.js',
```
gcsecsey: Nit: Are the identifiers at the end of the name/path needed? Do these serve some purpose?

gcsecsey (commented):
Thanks for working on this fix @bcotrim! The proposed solution looks good to me overall, and I could confirm it fixes the earlier errors when starting the existing servers when the app starts.

I see that the logs @epeicher attached also contain a seemingly relevant error at the bottom. 🤔 I tried importing the exported zip of [this playground](https://playground.wordpress.net/?blueprint-url=https%3A%2F%2Fblueprintlibrary.wordpress.com%3Fadd-extractor%3D1%26blueprint%3D532) (using the Quickstart blueprint) and this produced no issues. I'll try to double check with an export of a full site.

webpack-loaders/fix-import-meta-dirname-loader.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * Custom webpack loader to fix import.meta.dirname in specific packages
+ * 
+ * This loader targets @php-wasm/node and @wp-playground/cli worker threads
+ * to replace import.meta.dirname (which becomes undefined when bundled by webpack)
```
sejas, line 5: I wonder if we could populate `import.meta.dirname` to point to the right place.
If not, would it be possible to create a patch or fix it in the playground repositories somehow?

sejas (commented):
I tried it by running `npm i && npm run make:dmg-arm64`, and installing the app but I got an error when starting or creating a new site. Not sure if I'm missing some testing step.

| Starting | Creating |
|--------|--------|
| error-start | error-creating |

bcotrim:
@sejas @gcsecsey @epeicher thanks for the reviews.
I added a skip flag, so it wouldn't break performance tests. (performance tests will still build Studio the "old way"). If you all agree I would suggest merging this one and discuss/fix the build for all scenarios in a follow issue.

gcsecsey (approved):
Approving this so we can unblock the CfT. 🚢 We'll need to fix the performance tests separately.

wojtekn:
It's great that we found a fix, but it seems complex, and it can hit us in the future. Is there a way to simplify it, e.g., on the Playground side?

CC @Automattic/orbit

PR #1742, opened 2025-09-06
CLI: Add Sites list command

youknowriad:
Context about this PR pgle0O-QH-p2
See [this branch](https://github.com/Automattic/studio/commits/add/sites-list-command/) for the full CLI.

## Proposed Changes

This PR implements a very simple "list sites" command in the CLI tool. It's the first step towards a more capable and useful CLI tool.

## Testing Instructions

- Build the CLI tool `npm run cli:build`
- Run the command `node ./dist/cli/main.js sites list` 

It should list whatever sites you have in your Studio App in the terminal.

**Example**

Screenshot 2025-09-07 at 12 26 51 AM


youknowriad:
This is the smallest change possible. I'm considering defining a `commont/types/sites.ts` with the "Site" type and share it between CLI and App.

cli/commands/sites/list.ts
```diff
@@ -0,0 +1,94 @@
...
+import { Logger, LoggerError } from 'cli/logger';
+import { StudioArgv } from 'cli/types';
+
+interface SiteData {
+	id: string;
+	name: string;
+	path: string;
+}
+
+function getSitesCliTable( sites: SiteData[] ) {
```
bcotrim: If we remove `newSites`, we can display additional useful fields from appdata.sites:
  - Running status (running/stopped)
  - PHP version
  - WordPress version
  - Site URL (when running)

cli/index.ts
```diff
@@ -49,6 +50,10 @@ async function main() {
 			registerUpdateCommand( previewYargs );
 			previewYargs.demandCommand( 1, __( 'You must provide a valid command' ) );
 		} )
+		.command( 'sites', __( 'Manage local sites' ), ( sitesYargs ) => {
```
bcotrim: We should put these commands behind a feature flag for now.

  It would be better to ship the complete CLI sites functionality as a full release, with all subcommands (list, add, remove, start, stop), proper documentation, and coordinated marketing. This prevents
  users from discovering a partially-implemented feature.

What do you think?

bcotrim (commented):
Thanks for starting the CLI `sites` work @youknowriad 
This is a solid starting point!

I would recommend focusing solely on `appdata.sites` and excluding `newSites` from this implementation.

  `newSites` was created as a workaround for the CLI preview command to create sites, since the CLI lacked site management capabilities. Now that we're adding proper site management, we should update the preview command to use it and remove `newSites` entirely.

youknowriad:
> newSites was created as a workaround for the CLI preview command to create sites, since the CLI lacked site management capabilities. Now that we're adding proper site management, we should update the preview command to use it and remove newSites entirely.

This is good context for me. In my prototype I actually kind of avoided newSites entirely except for the list command. But I can exclude it from here and maybe ultimately get rid of it once we have the full "create commands" implemented..

cli/index.ts
youknowriad: I agree, do we have a precedent of using feature flags to hide CLI features? I know we have feature flags in the app but I'm unsure how these translate to the CLI.

bcotrim: AFAIK we haven't used feature flags in the CLI.
For now I would suggest using a env var to conditionally add the sites commands to the CLI in index.ts.
If you want I can take on that issue, and propose a solution on how we can manage feature flags across Studio and CLI.

youknowriad: @bcotrim Please go ahead I'd love that :)

cli/commands/sites/list.ts
youknowriad: I added more site informations. I'll work on unifying the SiteData and SiteDetails types separately from this PR.

cli/commands/sites/list.ts
```diff
@@ -0,0 +1,96 @@
...
+
+	return table;
+}
+
+function getSitesCliJson( sites: SiteData[] ): SiteData[] {
+	return sites.map( ( site ) => ( {
+		id: site.id,
+		name: site.name,
+		path: site.path,
+		running: site.running,
```
sejas: I think we should be consistent with properties for both JSON and table formats. Therefore, I recommend using the property `status` with the values `running` or `stopped`. This aligns with how WP-CLI usually works and will reduce the cognitive load for users who integrate the CLI tool.

Screenshot 2025-09-10 at 20 19 31

cli/lib/appdata.ts
```diff
@@ -15,6 +15,17 @@ import { validateAccessToken } from 'cli/lib/api';
 import { LoggerError } from 'cli/logger';
 
 const siteSchema = z
+	.object( {
+		id: z.string(),
+		path: z.string(),
+		name: z.string(),
+		phpVersion: z.string(),
+		running: z.boolean().optional(),
+		url: z.string().url().optional(),
```
sejas: `url` is not a property that comes from appData. We could infer the value using an expression similar to:
```
site.customDomain || `localhost:${site.port}`
``` 
Do you think it's worth transforming the appData site values to include the `site.url` property?

youknowriad:
how is this looking?

sejas (commented):
@youknowriad, thanks for pushing the CLI project forward, and splitting the functionality in small PRs.
I tested it by running `npm run cli:build` and then `node ./dist/cli/main.js sites list`, and it worked as expected.

I left a suggestion to use the same properties in both formats.
I wonder if we should add more data to the list, like the WordPress version of each site.

For the "running" status, I had the idea of checking if the site is up on the given port, or even building an endpoint at Express.js level that could return if the site is running. 

Screenshot 2025-09-10 at 20 02 43


I like how the subcommands are grouped under `studio sites` main command:

studio-sites-result

I see that Studio CLI adds the `--path` parameter as a global option. We might consider adding it at the command level instead or just removing it from the `sites` command, as I don't see it being useful for this command.

sejas:
> how is this looking?

It's looking great! ✨ I’ll be happy to help by submitting new PRs to add functionality.

sejas:
> I see that Studio CLI adds the --path parameter as a global option. We might consider adding it at the command level instead or just removing it from the sites command, as I don't see it being useful for this command.

- @youknowriad , I created this quick PR to remove `--path` from `studio sites` https://github.com/Automattic/studio/pull/1758

youknowriad:
> I wonder if we should add more data to the list, like the WordPress version of each site.

I looked at the type for `SiteDetails` in the app and I don't see the WP version there?

youknowriad:
> For the "running" status, I had the idea of checking if the site is up on the given port, or even building an endpoint at Express.js level that could return if the site is running.

Feels like just duplicate work (assuming the flag is persisted properly). We might need this later I give it to you but I prefer to keep things small for now.

youknowriad:
>I see that Studio CLI adds the --path parameter as a global option. We might consider adding it at the command level instead or just removing it from the sites command, as I don't see it being useful for this command.

Sounds good to me, so add it to all the preview commands right?

cli/commands/sites/list.ts
youknowriad: Makes sense :)

cli/lib/appdata.ts
youknowriad: No, I think it's better to leave this as they are in appData for now. I'll maybe just remove the url for now. We can improve and add more stuff later.

sejas:
> I looked at the type for `SiteDetails` in the app and I don't see the WP version there?

The WordPress version is not saved in appData and it's read directly from each sites's `wp-includes/version.php` [https://github.com/Automattic/studio/blob/66e76f8824495e86437b7e97956aa1cc81659223/src/ipc-handlers.ts#L906-L923](https://github.com/Automattic/studio/blob/66e76f8824495e86437b7e97956aa1cc81659223/src/ipc-handlers.ts#L906-L923)

youknowriad:
> The WordPress version is not saved in appData and it's read directly from each sites's wp-includes/version.php

Ok, I guess it's better to add a common library to do that and do it in its own PR. I created a dedicated issue for it STU-788

sejas:
> Sounds good to me, so add it to all the preview commands right?

Yesterday I created this PR that hides the --path option from the `studio sites` group of commands: Automattic/studio#1758

wojtekn:
@youknowriad thanks for the contribution - the list works well for me.

Could we change `sites` to `site` to follow the pattern established by WP CLI and adopted by the `preview` command?

youknowriad:
> Could we change sites to site to follow the pattern established by WP CLI and adopted by the preview command?

Works for me, but I'm not fully convinced. I think once we have site start/stop, we'll have to reshuffle the command names again. I think we'll have a better idea what works and what doesn't later.

wojtekn:
> Works for me, but I'm not fully convinced. I think once we have site start/stop, we'll have to reshuffle the command names again. I think we'll have a better idea what works and what doesn't later.

Why would we need to reshuffle it after adding start/stop? I think the singular names would make logical sense and would be consistent with WP CLI.

Studio currently has those:

```
studio preview create
studio preview list
studio preview delete
studio preview update
```

Now we would add:

```
studio site list
```

And later:

```
studio site create
studio site delete
studio site update
studio site start
studio site stop
```

In WP CLI we have:

```
wp plugin install
wp plugin list
wp plugin update
etc.
```

youknowriad:
For me if you put "start and stop" under "site", you should also put "preview" under "site". But I don't want to get into this discussion right now to be honest, there's no urgency. Let's build what we need and things might clarify later.

cli/index.ts
sejas: @youknowriad , @bcotrim , I've created a quick PR to add a feature flag using `ENABLE_CLI_V2` https://github.com/Automattic/studio/pull/1759
Let me know what you think.

cli/commands/site/list.ts
```diff
@@ -0,0 +1,102 @@
...
+		wordWrap: true,
+		wrapOnWordBoundary: false,
+	} );
+
+	sites.forEach( ( site ) => {
+		table.push( [
+			site.name,
+			site.path,
+			site.id,
+			site.running ? __( 'Running' ) : __( 'Stopped' ),
```
sejas: `site.running` will be always falsy, as it doesn't exist in AppData, so all sites will appear as Stopped until we implement a mechanism to detect whether a site is running or not.

wojtekn:
I see, thanks @youknowriad .

sejas (approved):
The code looks good to me, and now that it's behind a feature flag we can merge this PR. We can always add more fields in the future if necessary.
Currently, the status will be "Stopped" for all sites, but I think it's fine to keep it as a placeholder.
Merging this PR will make it easier to divide the work and add more site commands in different PRs.

| Table | JSON |
|--------|--------|
|  Screenshot 2025-09-11 at 12 37 58 | Screenshot 2025-09-11 at 12 37 48 |

cli/commands/site/list.ts
youknowriad: Oh really, where can I check exactly what things exists in app data or not? I think for now I'm just going to remove this information.

youknowriad: Do we have a "type" that defines what is saved into appData?

sejas: AppData is available in `~/Library/Application Support/Studio/appdata-v1.json` and here is an example of one of my sites:

```
    {
      "id": "e1014e86-ece3-4c1f-a1ac-632d7df5bae7",
      "name": "Antonio Sejas | <yay> | 123",
      "path": "/Users/macbookpro/Studio/my-blog",
      "adminPassword": "NWFake1230d0",
      "port": 8882,
      "phpVersion": "8.0",
      "isWpAutoUpdating": false,
      "customDomain": "antonio-sejas-yay-123.wp.local",
      "enableHttps": false,
      "autoStart": true,
      "themeDetails": {
        "name": "Twenty Twenty-Five/twentytwentyfive-trunk",
        "path": "/var/www/html/wp-content/themes/twentytwentyfive-trunk",
        "slug": "twentytwentyfive-trunk",
        "isBlockTheme": true,
        "supportsWidgets": false,
        "supportsMenus": false
      }
    }
```

The type that better represent it is `StoppedSiteDetails`:
https://github.com/Automattic/studio/blob/28ca9ea06a2bd89bd9d89f9acbf42ad3412a123e/src/ipc-types.d.ts#L14-L38

youknowriad: I'm thinking we might need to have that typed properly and shared with both CLI and app at some point to clarify expectations when reading from appData.

youknowriad: I was basing my info on the `SiteDetails` type and I was wrong haha.

sejas: Agreed, the AppData type should be common for CLI and the App 💯 .

sejas (approved):

youknowriad:
Thanks folks, more PRs coming :)

PR #1752, opened 2025-09-09
Make UI more responsive while site is being created

nightnei:
## Related issues

- Fixes STU-764

## Proposed Changes
We decided to go with more responsive messages while a site is being created

## Testing Instructions
1. Try to create a new site and as a second test create a new site from a backup
2. Assert in both cases that you see different messages of teh status of creatign a site
Screenshot 2025-09-09 at 18 10 58


bcotrim (commented):
@nightnei thanks for making the changes!
I think the new progress bar looks better, it gives a sense of constant progression compared to the "simulated steps" we had before.
However I think the feedback messages could be misleading. A user could think an issue is related to "Setting up WordPress tables…" when it actually isn't true. It could end up leading them on the wrong debug path, which can be very frustrating.
I would suggest something more generic like 'Setting up your site...' or 'Preparing your workspace...' without going into specific technical details.

We recently did something similar in calypso if you want to take a look: https://github.com/Automattic/wp-calypso/pull/104589

What do you think?

nightnei:
@bcotrim, I see you point, makes sense, what about this set? https://github.com/Automattic/studio/pull/1752/commits/0d3bfd6e0f73e5e13df3383d2e39db8801b523ca#diff-696f87d6f90b5ac111d50b4d8169341d30dccf1faf4179ba4774b7987f2aee9eR10-R19

bcotrim:
> @bcotrim, I see you point, makes sense, what about this set? [0d3bfd6#diff-696f87d6f90b5ac111d50b4d8169341d30dccf1faf4179ba4774b7987f2aee9eR10-R19](https://github.com/Automattic/studio/commit/0d3bfd6e0f73e5e13df3383d2e39db8801b523ca#diff-696f87d6f90b5ac111d50b4d8169341d30dccf1faf4179ba4774b7987f2aee9eR10-R19)

I think we still have the same issue, it makes it look like we are doing specific things (configuring, installing, optimizing, enhancing) which can be misleading, if we aren't actually doing those things. 
I was thinking about messages more like:
"Setting up your new site..."
"Getting everything ready..."
"Almost there..."
"Working on your site..."
"Making progress..."
"Nearly finished..."

My idea is to give the feedback that something is happening, without compromising on a wrong technical step, if we can't be sure about it. What do you think?

bcotrim (approved):
LGTM 👍

One thing I noticed is the timing between message updates feels quite fast, especially on complex blueprints where it reaches `Nearly finished` and stays there for a while. I've attached an example blueprint where this is noticeable.

That said, the progress bar changes already make this feel much more "alive," so this might not even be worth adjusting. What do you think?

[complex-blueprint.json](https://github.com/user-attachments/files/22257476/complex-blueprint.json)

nightnei:
@bcotrim this blueprint takes 40 seconds, we have 6 phrases, so I updated timeout to 5 secs, I think now it's good one. WDYT?

PR #1758, opened 2025-09-10
Remove --path option from studio sites list commands

sejas:
## Related issues

- https://github.com/Automattic/studio/pull/1742

## Proposed Changes

- I suggest removing the `--path` option from the `studio sites` group of commands.

## Testing Instructions

- Run `npm run cli:build`
- Run `node ./dist/cli/main.js sites list --help`
- Observe the `--path` parameter option is not offered as a possibility
- Run node ./dist/cli/main.js sites
- Observe the `--path` parameter option is not offered as a possibility
- Run `node dist/cli/main.js preview --help`
- Observe the `--path` parameter is still offered as an option: `--path     Path to the WordPress files   [string] [default: Current directory]`

| Before | After |
|--------|--------|
| studio-cli-paths-before | studio-cli-paths-after |

## Pre-merge Checklist

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -51,6 +51,9 @@ async function main() {
 			previewYargs.demandCommand( 1, __( 'You must provide a valid command' ) );
 		} )
 		.command( 'sites', __( 'Manage local sites' ), ( sitesYargs ) => {
+			sitesYargs.option( 'path', {
+				hidden: true,
+			} );
 			registerSitesListCommand( sitesYargs );
 			sitesYargs.demandCommand( 1, __( 'You must provide a valid command' ) );
 		} )


cli/index.ts
```diff
@@ -51,6 +51,9 @@ async function main() {
 			previewYargs.demandCommand( 1, __( 'You must provide a valid command' ) );
 		} )
 		.command( 'sites', __( 'Manage local sites' ), ( sitesYargs ) => {
+			sitesYargs.option( 'path', {
+				hidden: true,
+			} );
```
youknowriad, line 56: I wonder if it's better to actually just make this option explicit in the preview commands (and the newer commands that need it). But I'm hesitant. so this work for me :)

youknowriad (approved):

cli/index.ts
sejas: Yep! Either way will work.
I thought that `studio sync` might need the `--path` parameter, so I decided to take this approach.

PR #1760, opened 2025-09-11
Blueprints: Improve card selection in site creation

epeicher:
## Related issues

- Relates to STU-783

## Proposed Changes

- Adds an external css file scoped to the `blueprints.tsx` named `blueprints.css` so we can add style changes to the Dataviews component
- Adds a hover effect that adds a border to the whole card
- Adds a border to the whole card selected
- Adds a check icon to the bottom-right of the card to indicate that it is selected
- Adds a mock for css files so Unit Tests don't fail if they find `import './<file_name>.css';`

### Before

https://github.com/user-attachments/assets/57b02dd4-b4a2-474d-a064-9ed0c06b3454

### After

https://github.com/user-attachments/assets/266c2e55-c35a-4014-8392-5f80e23e0bff

## Testing Instructions

- Apply this branch and start Studio running `npm start`
- Add a site
- Select `Start from a blueprint`
- Check that when hovering over a blueprint, there is an indication that the whole card can be selected
- Click on a card (currently only the image and title are clickable)
- Check that there is an indication that the card is selected
- Click on another card
- Check that the selection changes
- Confirm there is no visual regressions of the card interaction

## Pre-merge Checklist


src/modules/add-site/components/blueprints.css
```diff
@@ -0,0 +1,33 @@
...
+    @apply !bg-transparent !p-0 !w-full;
+}
+
+.blueprints-container .components-badge .components-badge__content {
+    @apply !w-full;
+}
+
+.blueprints-container .components-badge > * {
+    @apply !w-full;
+}
```
ivan-ottinger: I wonder if we could combine these two.

ivan-ottinger (commented):
Nice improvement! 🙂 The changes work as expected and it is now clear which blueprints is selected! 👌🏼 

https://github.com/user-attachments/assets/416c6d19-1aea-499e-bbb2-4dfdff0d5ab9

What if we added a little bit more space between the blue borders when one of the blueprints is already selected?

And just noticed that we will need to make the checkbox display on the other side for RTL:

Markup on 2025-09-11 at 17:43:51

Also left one minor comment / suggestion below.

epeicher:
Thanks for your review @ivan-ottinger ! 

> What if we added a little bit more space between the blue borders when one of the blueprints is already selected?

> And just noticed that we will need to make the checkbox display on the other side for RTL:

Very good point about the RTL! I have added both suggestions as part of [2cbc0cd](https://github.com/Automattic/studio/pull/1760/commits/2cbc0cdadab84a645930101e07f6984dc15fabe4)

src/modules/add-site/components/blueprints.css
epeicher: I've combined them as part of [16ad04d](https://github.com/Automattic/studio/pull/1760/commits/16ad04db80c7dea92d8a775b3f5ac02672093143) because the `.components-badge__content` is a direct child of `.components-badge`, which means the `> *` selector would include it.

bcotrim (commented):
@epeicher thanks for this improvement! This component style is a lot easier to read and maintain now. 👍 

Some comments:
- Should all the card be clickable? It took me a little bit to understand only the image was clickable
- What do you think about adding a small animation when hovering the cards so the navigation feels a little bit smoother?

epeicher:
@bcotrim thanks for the review!
> * Should all the card be clickable? It took me a little bit to understand only the image was clickable

Yes, with these changes, it is now necessary, as the whole card is outlined. I identified that in the test steps and in the related task, but I will work on this as part of this PR to avoid the confussion.

> * What do you think about adding a small animation when hovering the cards so the navigation feels a little bit smoother?

That animation exist already, please see the `after` video in the [test steps](https://github.com/Automattic/studio/pull/1760#issue-3406129682). The values for the animation time were taken from the previous version, do you think we should make the animation more prominent?

src/modules/add-site/components/blueprints.css
ivan-ottinger: Nice! Thank you!

bcotrim:
> Yes, with these changes, it is now necessary, as the whole card is outlined. I identified that in the test steps and in the related task, but I will work on this as part of this PR to avoid the confussion.

That would be perfect, thanks 👍 

> That animation exist already, please see the `after` video in the [test steps](https://github.com/Automattic/studio/pull/1760#issue-3406129682). The values for the animation time were taken from the previous version, do you think we should make the animation more prominent?

After taking another look, I think it is good as is.

epeicher:
@bcotrim, I have found a way to handle the click item in the description and badge fields without modifying the `Dataviews` component, and I think it is a possible solution to allow selecting the whole card. I have added the changes as part of 9afafbd. That will make the whole card clickable, what do you think?

https://github.com/user-attachments/assets/1b305adf-b090-4fb5-9129-89984f9de9fe

katinthehatsite (commented):
For me functionality-wise, this is a great improvement from the current experience on trunk. I am able to select the whole card 👍 

One small thing is that I noticed there is small difference between the thinkness of the cards outline for on-hover and selected cards (see screenshot). I don't know if this is intended or not so just adding it here as an observation (see the second card is selected and then first one I hoevered over after selecting the second one):

Image

epeicher:
Latest changes in 1b594d6 include the whole title and the blank space in the footer next to the `Preview blueprint` link to the selection area, so the whole card can be selected.

epeicher:
Thanks @katinthehatsite for the review!

> One small thing is that I noticed there is small difference between the thinkness of the cards outline for on-hover and selected cards (see screenshot). I don't know if this is intended or not so just adding it here as an observation (see the second card is selected and then first one I hoevered over after selecting the second one):

I added that on purpose to differentiate between selection and hovering, I have asked @Marinatsu to review this but I am happy to update that if required.

epeicher:
@bcotrim, @Marinatsu, I have implemented most of the changes in the Figma design RToz6tIuQ7nlZrikBte4GU-fi-9666_237574  and this is ready for another review, please let me know if you find any issues as I have done quite a lot of small updates. This is the screenshot:

CleanShot 2025-09-16 at 17 51 45@2x

@ivan-ottinger, @katinthehatsite, you’re welcome to give it another round of review if you like 😄

epeicher:
~~I have found an issue with smaller heights, I am working on that~~
CleanShot 2025-09-16 at 17 59 59@2x

EDIT: This has been fixed as part of [3071cc2](https://github.com/Automattic/studio/pull/1760/commits/3071cc26d0c433296eab6b6cb00d6bfe5ac319f9)

CleanShot 2025-09-16 at 19 03 42@2x

bcotrim:
thanks for the adjustments @epeicher 

Some things I noticed:
1. Clicking in the white space below the description (but inside the card border) doesn't select the blueprint. Can we do that?
2. Based on the design update shared in Linear, the hovered state should not have a blue border but just "zoom" the thumbnail image.
image
3. If the user resizes the app to the min possible size it looks a bit broken. We don't have a design for that, but if we could at least break the title + preview link into 2 lines on that case, it would look a bit better in my opinion.
Screenshot 2025-09-16 at 18 58 50

What do you think about these changes?

wojtekn:
Looks great @epeicher !

> Clicking in the white space below the description (but inside the card border) doesn't select the blueprint. Can we do that?

Agreed, it would be nice to improve this part.

The one thing I'm unsure about is whether clicking "Preview blueprint" should select the blueprint. What do you think?

epeicher:
> thanks for the adjustments @epeicher
> 
> Some things I noticed:
> 
> 1. Clicking in the white space below the description (but inside the card border) doesn't select the blueprint. Can we do that?

Well spotted! That should work after 99eddea

> 2. Based on the design update shared in Linear, the hovered state should not have a blue border but just "zoom" the thumbnail image.

Thanks for that! I didn't realise of that looking at the desing, I've implemented that as part of [d9b552f](https://github.com/Automattic/studio/pull/1760/commits/d9b552f47f50e55c6e49f94617f604efea1d5216). Please note that I have also removed the shadow when hovering the cards as I don't see that in the design, but happy to add it if required.

3. If the user resizes the app to the min possible size it looks a bit broken. We don't have a design for that, but if we could at least break the title + preview link into 2 lines on that case, it would look a bit better in my opinion. 

Yep, I have done some amendments as part of [ef5d242](https://github.com/Automattic/studio/pull/1760/commits/ef5d2421796c40232c5f660fa846024075798537) and other commits. With those changes, the cards squeeze a bit but they don't look broken to me, please let me know if you find any issues

> What do you think about these changes?

All great suggetions @bcotrim , I've tried to implement all of them. Could you please give another review and let me know any issues?

epeicher:
> Looks great @epeicher !
> 
> > Clicking in the white space below the description (but inside the card border) doesn't select the blueprint. Can we do that?
> 
> Agreed, it would be nice to improve this part.
> 
> The one thing I'm unsure about is whether clicking "Preview blueprint" should select the blueprint. What do you think?

Thanks @wojtekn! I have added some code to avoid selecting the card if the user only clicks on `Preview blueprint` as I agree it might seem confusing. Done as part of [06141ef](https://github.com/Automattic/studio/pull/1760/commits/06141ef3e62c80855d155d61f18cf8fb72d70061)

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
bcotrim: Was this pushed by mistake?
If not, I'm up for a discussion on having some shared vscode settings, but I would suggest another PR or GitHub discussions.

src/modules/add-site/components/blueprints.css
```diff
@@ -0,0 +1,28 @@
+.blueprints-container .dataviews-title-field {
+	@apply px-5 w-full;
+}
+.blueprints-container .dataviews-view-grid__card {
+	@apply cursor-pointer relative justify-between outline outline-a8c-gray-5 rounded h-[279px] [@media(min-height:680px)]:h-[340px] min-w-32;
+}
+
+.blueprints-container .dataviews-view-grid__card:hover img {
+	@apply scale-110 transition-transform duration-150;
+}
```
bcotrim: This will apply the transition when hovering in and hovering out, making it look a little bit smoother.

```suggestion
.blueprints-container .dataviews-view-grid__card img {
	@apply transition-transform duration-150;
}

.blueprints-container .dataviews-view-grid__card:hover img {
	@apply scale-110;
}
```

bcotrim (approved):
LGTM 👍 

The final result looks great and follows the latest design.
Thanks for making all the adjustments @epeicher 

I left some minor suggestions, but feel free to merge whenever you feel it is ready

.vscode/settings.json
epeicher: That was a mistake 🤦 , I had different settings in my workspace and overrode them here

src/modules/add-site/components/blueprints.css
epeicher: Great point! I'm changing that

epeicher: [86b6525](https://github.com/Automattic/studio/pull/1760/commits/86b6525d4c5467c6fd4372054712aa8a5e16bbc6) thanks for the suggestion!

.vscode/settings.json
epeicher: [86b6525](https://github.com/Automattic/studio/pull/1760/commits/86b6525d4c5467c6fd4372054712aa8a5e16bbc6) removed 😄

PR #1762, opened 2025-09-11
Log out the user when the token is invalid

sejas:
## Related issues

- Fixes STU-669

## Proposed Changes

- Check every request and logout the user when the token is invalid

## Testing Instructions

- Log in with your WordPress.com account.
- Close Studio.
- Modify the `~/Library/Application Support/Studio/appdata-v1.json` file and change the accessToken text to make it invalid
- Open Studio.
- Observe a dialog that appears.
- Click OK.
- Observe that you are logged out.
- Log in again with your WordPress.com account.
- Observe that you have successfully logged in and can see the list of sites, create a preview site, or talk to the assistant.

https://github.com/user-attachments/assets/5b409277-287e-45b6-867f-fe9dbe4bd530

## Pre-merge Checklist

### DIFF
--- a/src/components/auth-provider.tsx
+++ b/src/components/auth-provider.tsx
@@ -198,12 +198,12 @@ function createWpcomClient(
 
 	return Object.assign( wpcom, {
 		request: function ( params: WpcomParams, callback: unknown ) {
-			const wrappedCallback = ( err: unknown, response: unknown ) => {
+			const wrappedCallback = ( err: unknown, response: unknown, headers: unknown ) => {
 				if ( err ) {
 					void handleInvalidTokenError( err );
 				}
 				if ( typeof callback === 'function' ) {
-					callback( err, response );
+					callback( err, response, headers );
 				}
 			};
 
--- a/src/components/auth-provider.tsx
+++ b/src/components/auth-provider.tsx
@@ -164,13 +164,17 @@ function createWpcomClient(
 ): WPCOM {
 	const wpcom = new WPCOM( token );
 
+	let isAuthErrorDialogOpen = false;
 	const handleInvalidTokenError = async ( response: unknown ) => {
-		if ( isInvalidTokenError( response ) && onInvalidToken ) {
-			getIpcApi().showErrorMessageBox( {
-				title: 'Session Expired',
-				message: 'Your session has expired. Please log in again.',
-			} );
+		if ( isInvalidTokenError( response ) && onInvalidToken && ! isAuthErrorDialogOpen ) {
+			isAuthErrorDialogOpen = true;
 			await onInvalidToken();
+			await getIpcApi().showMessageBox( {
+				type: 'error',
+				message: 'Session Expired',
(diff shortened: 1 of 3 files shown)


epeicher (commented):
Thanks @sejas! I can confirm that the session expired dialog appears and the reported issue is not happening anymore 🙌 . However, when I try the Assistant, I can see an error, and the following logged into the console 
CleanShot 2025-09-12 at 09 32 02@2x


I have tested in `trunk` and this error does not happen there

wojtekn:
Do we need the dialog to inform the user about that? What if we change the state to logged out? User will still see nudge in many places in the app like Previews, Sync or Assistant.

ivan-ottinger:
Great work on those changes! They work as expected and the dialog message is clear to me.

> However, when I try the Assistant, I can see an error, and the following logged into the console

That's a good catch! I did not observe it at first, but now can see it after trying submitting a request to the Assistant. Everything else is working correctly.

sejas:
> Do we need the dialog to inform the user about that? What if we change the state to logged out? User will still see nudge in many places in the app like Previews, Sync or Assistant.

I think it's better to inform the user when their logged-in or logged-out state has changed because the token has expired. The reason is that this is an action the user did not initiate, so in my opinion, it deserves a visual indication to notify the user.
Without any indication, users can be confused about why they are suddenly logged out.

sejas:
> However, when I try the Assistant, I can see an error, and the following logged into the console

@epeicher, Great catch !!! 🙏 , it seems I didn't pass the headers to the callback, and the Assistant check the headers to see if the user has reached the limit.

Fixed here: https://github.com/Automattic/studio/pull/1762/commits/8249a3899c398eb70937916f11a08a6701751b75
And found the wpcom code where the callback is defined: https://github.com/Automattic/wp-calypso/blob/a9d0b0e6d6ab051b43cf03e8617b51c9c6a1328f/packages/wpcom.js/src/lib/util/send-request.js#L72

epeicher (approved):
Thanks @sejas for addressing the comments! And thanks for adding the reference to the callback 👍 . I have tested the latest changes and they work as expected, I can see the dialog and after logging in again, I can update my preview site and use the AI assistant. LGTM! :shipit: 

Open Studio with invalid token|Preview | AI Assistant|
---|-------|------|
CleanShot 2025-09-12 at 17 26 06@2x| CleanShot 2025-09-12 at 17 23 04@2x| CleanShot 2025-09-12 at 17 23 24@2x|

sejas:
> Do we need the dialog to inform the user about that? What if we change the state to logged out? User will still see nudge in many places in the app like Previews, Sync or Assistant.

@wojtekn , do you think is ok to merge the PR with the dialog?, Should we add two buttons to the modal (`Ok`, `Log in`) one to close the alert and the other to trigger the login and open the OAuth page in the browser?

wojtekn:
@sejas, it's okay to merge. However, I'm unsure if we need to inform the user about being logged out if all places in the UI show that already.

PR #1763, opened 2025-09-11
Lib: Move port-finder to common folder to be used later by both CLI and App

youknowriad:
Fixes STU-770

This is just a tiny PR that moves the existing port-finder library without any change to the "common" folder. That way we should be able to use it for both site creation in app and site creation in CLI.

This is a first step towards removing the "newSites" workaround mentioned by @bcotrim


sejas (approved):
Code looks good.
I tested it by creating a new site and starting an existing site using the Studio App. Both flows started the site correctly.

PR #1766, opened 2025-09-11
FS Utils: Move all utils to a common lib

youknowriad:
## Related issues

Fixes STU-773

This is another small PR that doesn't really change anything code wise, it just moves code around. I noticed that we have two fs-utils lib files, one in "src" and one in "common". This PR unifies them and move the utils into a single file under "common".

On its own, this is a code quality improvement but also my motivation is that there are some utils that I need to move later on that depend on this (for site creation in CLI). So I prefer to do it in steps. This PR focus only on making "fs-utils" a common utility without any behavior change.

## Testing instructions

There should be really no impact here, it's just moving code around.


common/lib/fs-utils.ts
```diff
@@ -63,3 +64,44 @@ export function arePathsEqual( path1: string, path2: string ) {
 		return false;
 	}
 }
+
+export async function pathExists( path: string ): Promise< boolean > {
```
youknowriad, line 68: This is just the exact same code that was in src/lib/fs-utils without changes.

sejas (approved):
I tested the app by starting an existing site, creating a new one, and exporting a site.
Everything worked as expected.

It makes sense to share the fs-utils between in common folder. 👍

PR #1770, opened 2025-09-12
Move WordPress config constants into loader mu-plugin

ivan-ottinger:
## Related issues

- Fixes STU-760.

## Proposed Changes

- move WordPress config constants polyfill into loader mu-plugin to ensure all the requests get the WordPress constants defined

After looking into the issue with @bcotrim today, we noticed that when the related MailPoet code runs, it is already in a different context where the constants are not defined.

Moving the definition to an earlier mu-plugin does not solve the issue. Instead, the definitions need to be set in the mu-plugin loader.

## Testing Instructions

1. Check out the PR branch and build the app.
2. Try to reproduce the following issue: STU-760.
  - import a new site using the attached Jetpack backup
  - let the site start
3. The site should run correctly, without any errors.
5. Try to start and access other local sites. There should be no regressions.

## Pre-merge Checklist

### DIFF
--- a/src/lib/wordpress-provider/playground-cli/mu-plugins.ts
+++ b/src/lib/wordpress-provider/playground-cli/mu-plugins.ts
@@ -31,6 +31,31 @@ async function createLoaderMuPlugin(): Promise< string > {
 		 * Loads Studio-specific mu-plugins from /internal/studio/mu-plugins/
 		 */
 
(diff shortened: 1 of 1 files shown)


bcotrim (approved):
LGTM 👍 

I confirmed that the reported import issue no longer happens.
Nice catch!

PR #1781, opened 2025-09-15
Fix Eslint warning when importing cli-table3

sejas:
## Proposed Changes

- We have a eslint warning when importing Table from `cli-table3`. We followed the same approach in:
https://github.com/Automattic/studio/blob/a4fe0ab401aa7c20534d4cce6903fd9b9e77696d/cli/lib/snapshots.ts#L2-L3

## Testing Instructions

- Run `npm run lint`
- Observe no errors or warnings

## Pre-merge Checklist

### DIFF
--- a/cli/commands/site/list.ts
+++ b/cli/commands/site/list.ts
@@ -1,4 +1,5 @@
 import { __, _n, sprintf } from '@wordpress/i18n';
+// eslint-disable-next-line import/no-named-as-default
 import Table from 'cli-table3';
 import { PreviewCommandLoggerAction as LoggerAction } from 'common/logger-actions';
 import { readAppdata, type SiteData } from 'cli/lib/appdata';


nightnei (approved):
I tried other options and it seems that to avoid the comment we have only one option, but I am not sure that it's better, so let's go with proposed in the PR 👍

``` diff
diff --git a/cli/commands/site/list.ts b/cli/commands/site/list.ts
index cd4a3ff6..47371aaa 100644
--- a/cli/commands/site/list.ts
+++ b/cli/commands/site/list.ts
@@ -1,6 +1,5 @@
 import { __, _n, sprintf } from '@wordpress/i18n';
-// eslint-disable-next-line import/no-named-as-default
-import Table from 'cli-table3';
+import * as Table from 'cli-table3';
 import { PreviewCommandLoggerAction as LoggerAction } from 'common/logger-actions';
 import { readAppdata, type SiteData } from 'cli/lib/appdata';
 import { Logger, LoggerError } from 'cli/logger';
@@ -14,7 +13,7 @@ interface SiteTable {
 }
 
 function getSitesCliTable( sites: SiteData[] ) {
-       const table = new Table( {
+       const table = new Table.default( {
                head: [ __( 'Name' ), __( 'Path' ), __( 'ID' ), __( 'PHP' ) ],
                style: {
                        head: [ 'cyan' ],
```

sejas:
Thanks for sharing the other option 👌

PR #1714, opened 2025-08-27
Fix using custom domains with Playground-cli

nightnei:
## Related issues

- Fixes STU-726

## Proposed Changes
With this PR we are updating playground libs to have oiportunity to probvide site-url to playgropund-cli to fix custom domains in Studio.

## Testing Instructions
1. Run `ENABLE_BLUEPRINTS=true npm start`
2. Create Site with custom domain and w/o ssl
3. Assert that site and wp-admin work correctly
4. Create Site with custom domain and with ssl
5. Assert that site and wp-admin work correctly

PR #1783, opened 2025-09-16
Upgrade Electron Forge lib to 7.9.0 and unlock packager

wojtekn:
## Related issues

- Relates to #201 and #795

## Proposed Changes

I propose to remove the override for `@electron/packager` that locked it with the 18.1.3 version. We introduced that as higher versions of the packager have a change that prevents versions with alpha characters appearing in the fourth part of the version string from working (#1714). The related issue shows such a version as unsupported:

```
2024.07.0-hourly+56.pro14
```

And we used such a format, which also had the fourth element alphanumeric:

```
1.0.3-dev.f8140f6
```

We can remove the override now as we switched to a simpler version format:

```
1.5.6-dev104
```

As a part of this PR, I upgraded Electron Forge to the latest version.

## Testing Instructions

Check if builds go through.

## Pre-merge Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,13 +71,13 @@
 			"devDependencies": {
 				"@automattic/color-studio": "^4.1.0",
 				"@automattic/wp-babel-makepot": "^1.2.0",
-				"@electron-forge/cli": "^7.8.3",
-				"@electron-forge/maker-deb": "^7.8.3",
-				"@electron-forge/maker-dmg": "^7.8.3",
-				"@electron-forge/maker-squirrel": "^7.8.3",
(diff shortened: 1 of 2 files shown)


bcotrim (approved):
LGTM 👍 

I tested running the project locally and installing the built dev version from CI and both worked without any issues

Image

epeicher (approved):
Thanks for the upgrade @wojtekn! I have tested the following commands and they all work fine:
- `npm install`
- `npm run make`
- `npm run cli:build`
- `npm start`

Also, I've done a smoke test of Studio and I have not found any issues. LGTM! :shipit:

PR #1804, opened 2025-09-22
Optimize playground cli create site workflow

bcotrim:
## Related issues

- Fixes STU-724

## Proposed Changes

- Optimize the Playground CLI Provider create site workflow, to avoid running the CLI twice
- Start server now checks for a WordPress directory to assert if it should run the Playground CLI WordPress setup or skip it

## Testing Instructions

- Start Studio
- Create a site with and without blueprint
- Validate the Studio logs and confirm the `[playground-cli]` entries only appear once per site created/started

## Pre-merge Checklist

### DIFF
--- a/src/site-server.ts
+++ b/src/site-server.ts
@@ -173,7 +173,7 @@ export class SiteServer {
 		};
 
 		if ( this.meta.blueprint ) {
-			this.meta.blueprint = undefined;
+			delete this.meta.blueprint;
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


src/site-server.ts
```diff
@@ -20,7 +20,7 @@ import {
 import WpCliProcess, { MessageCanceled, WpCliResult } from 'src/lib/wp-cli-process';
 import { createScreenshotWindow } from 'src/screenshot-window';
 import { getSiteThumbnailPath } from 'src/storage/paths';
-import { Blueprint } from './stores/wpcom-api';
+import { Blueprint } from 'src/stores/wpcom-api';
```
epeicher: Nit: This helps a bit with readibility
```suggestion
import type { Blueprint } from 'src/stores/wpcom-api';
```

epeicher (approved):
Thanks @bcotrim, this is a smart optimization, I think it improves the code! I have teated it, and it works as expected. I can create and start a site with and without blueprints. And I only see log entries once. LGTM!

CleanShot 2025-09-22 at 16 03 41@2x

PR #1816, opened 2025-09-24
Studio: Clean up stack traces from import error dialogs

bcotrim:
## Related issues

- Fixes STU-814

## Proposed Changes

- Returns only first import error line to the UI
- Logs the full error message

## Testing Instructions

- Create a Studio site
- Create a `.sql` file with an incorrect query
- Import the SQL file into the site
- Verify that the error dialog shows a clean, user-friendly error message without PHP stack traces, file paths, or HTML markup
- Test with different types of errors to ensure the cleaning works across various error formats
- Confirm you can still see the full error in the Studio logs

## Pre-merge Checklist


epeicher (approved):
Thanks @bcotrim! I think for this change, ideally the playground-cli should return a pretty formatted or at least a summary of the error, but meanwhile this works as expected. I have tested with two different errors and the result is good, the SQL error is even better than before. LGTM!

|Before SQL error | After SQL error| After Import error
|-------|------|------
| Image| Image|Image|

wojtekn:
@bcotrim, the solution feels a bit complex for what it achieves. Would it be sufficient just to show the first line of the returned error?

Screenshot 2025-09-25 at 10 25 21

bcotrim:
@wojtekn @epeicher thanks for the reviews and feedback.
I've updated the solution to adjust only error message from imports. Can you please take another look?

epeicher:
> @wojtekn @epeicher thanks for the reviews and feedback. I've updated the solution to adjust only error message from imports. Can you please take another look?

This still works fine for the same errors [I tested above](https://github.com/Automattic/studio/pull/1816#pullrequestreview-3266427620):

|Incorrect SQL | Error importing|
|-------|------|
|  CleanShot 2025-09-25 at 12 21 23@2x| CleanShot 2025-09-25 at 12 21 06@2x|

wojtekn:
Looks good, thanks for simplifying that.

wojtekn (approved):

PR #1821, opened 2025-09-26
Studio: Clean up stack traces from .wpress import error dialogs

bcotrim:
## Related issues

- Fixes STU-814

## Proposed Changes

- Extended error handling in src/hooks/use-import-export.tsx to include .wpress files alongside .sql files
- Stack traces are now stripped from error messages for .wpress file imports, showing only the first line of the error

## Testing Instructions

- Import a corrupted or invalid .wpress file in Studio (you can find one in the Linear issue)
- Verify that the error dialog shows only a clean error message without stack traces
- Test that .sql file error handling still works as expected

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-import-export.tsx
+++ b/src/hooks/use-import-export.tsx
@@ -111,7 +111,10 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode 
 					} );
 				} else {
 					let errorToShow = error;
-					if ( file instanceof File && file.name.endsWith( '.sql' ) ) {
+					if (
+						file instanceof File &&
+						( file.name.endsWith( '.sql' ) || file.name.endsWith( '.wpress' ) )
+					) {
 						const errorMessage = ( error as Error )?.message;
 						if ( errorMessage ) {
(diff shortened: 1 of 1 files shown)


epeicher (approved):
Thanks @bcotrim! I was going to suggest what happens with other file types, but I guess there are no more, I have tested with corrupted `.zip`, `.wpress`, and `.sql` files, and they all work fine, so this LGTM!

| .wpress | .zip| .sql|
|-------|------|------|
|Image | Image |  Image|

src/hooks/use-import-export.tsx
```diff
@@ -111,7 +111,10 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
 					} );
 				} else {
 					let errorToShow = error;
-					if ( file instanceof File && file.name.endsWith( '.sql' ) ) {
+					if (
+						file instanceof File &&
+						( file.name.endsWith( '.sql' ) || file.name.endsWith( '.wpress' ) )
```
wojtekn: Why do we even limit this to specific file extensions?

PR #1824, opened 2025-09-29
Studio: Implement provider-specific minimum WordPress version requirements

bcotrim:
## Related issues

- Fixes STU-824

## Proposed Changes

- Added MINIMUM_WORDPRESS_VERSION constants to both providers (Playground CLI: 6.2.1, WP Now: 5.9.9)
- Updated WordPress versions API to accept minimumVersion parameter and use provider-specific minimums
- Modified WPVersionSelector component to use minimum version from provider constants
- Fixed dev version handling in playground-server-process-child.ts to pass "nightly" instead of specific dev version
- Updated TypeScript interfaces and Redux store to include minimum version support

## Testing Instructions

- Test WordPress version selection with Blueprint enabled (should show versions >= 6.2.1)
- Test WordPress version selection with Blueprint disabled (should show versions >= 5.9.9)
- Test creating a new site with nightly WordPress version
- Verify that the minimum version is respected in both provider scenarios
- Check that existing sites continue to work with their current WordPress versions

## Pre-merge Checklist


src/stores/wordpress-versions-api.ts
```diff
@@ -5,7 +5,8 @@ import { z } from 'zod';
 import { isWordPressDevVersion, isWordPressBetaVersion } from 'src/lib/wordpress-version-utils';
 import { withOfflineCheck } from 'src/stores/utils/with-offline-check';
 
-const MINIMUM_WORDPRESS_VERSION = '5.9.9';
+// Default minimum version for fallback
+const DEFAULT_MINIMUM_WORDPRESS_VERSION = '5.9.9';
```
wojtekn: Would it make sense to check all providers here and use the oldest version?

wojtekn (approved):
I tested it on macOS, and it worked fine - the custom nightly and oldest versions worked for me, they were created correctly, and were not auto-updated.

src/stores/wordpress-versions-api.ts
bcotrim: You mean the newest?
For example Playground is 6.2.1 and wp-now is 5.9.9, so 6.2.1 would be safe version to use. Did I understand your suggestion correctly?

src/lib/test-utils.tsx
```diff
@@ -40,6 +41,7 @@ export function createTestStore( options: TestStoreOptions = {} ) {
 				defaultPhpVersion: options.providerConstants.defaultPhpVersion || '',
 				defaultWordPressVersion: options.providerConstants.defaultWordPressVersion || '',
 				allowedPhpVersions: options.providerConstants.allowedPhpVersions || [],
+				minimumWordPressVersion: options.providerConstants.minimumWordPressVersion || '',
```
sejas: We could consider making minimumWordPressVersion option mandatory in `TestStoreOptions` too.

src/stores/wordpress-versions-api.ts
sejas: We could also remove this version and then make the  `{ minimumVersion?: string }` a mandatory parameter.

sejas (commented):
I tested it with blueprints feature flag enabled and disabled.
I confirm the WP versions when creating and editing site match the minimum version for each provider. I had to refresh (cmd+r) after enabling the feature flag.

One thing I've noticed is that sites created using wp-now with WP < v6.2.6 are returning a 500 error. In that case I wonder if we should just update `MINIMUM_WORDPRESS_VERSION` in https://github.com/Automattic/studio/blob/08874cec71163518ba23fa2c9e4cfd24d8aa7ca2/src/stores/wordpress-versions-api.ts#L8

ERROR 500 in WP < 6.2.6

```
[29-Sep-2025 16:54:00 UTC] PHP Fatal error:  Uncaught Error: Cannot access private property WP_SQLite_DB::$allow_unsafe_unquoted_parameters in /wordpress/wp-includes/wp-db.php:740
Stack trace:
#0 /wordpress/wp-includes/wp-db.php(740): wpdb->__get('allow_unsafe_un...')
#1 /wordpress/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php(386): wpdb->__get('allow_unsafe_un...')
#2 /wordpress/wp-includes/option.php(195): WP_SQLite_DB->prepare('SELECT option_v...', 'siteurl')
#3 /wordpress/wp-includes/default-constants.php(156): get_option('siteurl')
#4 /wordpress/wp-settings.php(343): wp_plugin_directory_constants()
#5 /wordpress/wp-config.php(96): require_once('/wordpress/wp-s...')
#6 /wordpress/wp-load.php(50): require_once('/wordpress/wp-c...')
#7 /wordpress/wp-admin/install.php(36): require_once('/wordpress/wp-l...')
#8 {main}
  thrown in /wordpress/wp-includes/wp-db.php on line 740
[29-Sep-2025 16:54:00 UTC] PHP Fatal error:  Uncaught Error: Cannot access private property WP_SQLite_DB::$allow_unsafe_unquoted_parameters in /wordpress/wp-includes/wp-db.php:740
Stack trace:
#0 /wordpress/wp-includes/wp-db.php(740): wpdb->__get('allow_unsafe_un...')
#1 /wordpress/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-db.php(386): wpdb->__get('allow_unsafe_un...')
#2 /wordpress/wp-includes/option.php(195): WP_SQLite_DB->prepare('SELECT option_v...', 'WPLANG')
#3 /wordpress/wp-includes/l10n.php(63): get_option('WPLANG')
#4 /wordpress/wp-includes/l10n.php(137): get_locale()
#5 /wordpress/wp-includes/l10n.php(835): determine_locale()
#6 /wordpress/wp-includes/class-wp-fatal-error-handler.php(46): load_default_textdomain()
#7 [internal function]: WP_Fatal_Error_Handler->handle()
#8 {main}
  thrown in /wordpress/wp-includes/wp-db.php on line 740
```

**With blueprints enabled:**

Screenshot 2025-09-29 at 17 35 14

**Without blueprints**

Screenshot 2025-09-29 at 17 37 09

src/stores/wordpress-versions-api.ts
wojtekn: > You mean the newest?
> For example Playground is 6.2.1 and wp-now is 5.9.9, so 6.2.1 would be safe version to use. Did I understand your suggestion correctly?

You used the oldest, so I proposed to use the oldest. But maybe we could remove it completely as @sejas proposed?

bcotrim: Just updating the min version constant to `6.2.6`?
That sounds good to me, less trouble to maintain both provider constants.

wojtekn: @bcotrim I think @sejas meant to remove `DEFAULT_MINIMUM_WORDPRESS_VERSION`, make the version a required param for the query, and always provide the version from the provider.

bcotrim: Got it now, thanks!

src/lib/wordpress-provider/playground-cli/playground-cli-provider.ts
```diff
@@ -36,13 +36,15 @@ export class PlaygroundCliProvider implements WordPressProvider {
 	static readonly DEFAULT_PHP_VERSION = RecommendedPHPVersion;
 	static readonly DEFAULT_WORDPRESS_VERSION = 'latest';
 	static readonly ALLOWED_PHP_VERSIONS = [ ...SupportedPHPVersions ];
+	static readonly MINIMUM_WORDPRESS_VERSION = '6.2.1'; // https://wordpress.github.io/wordpress-playground/blueprints/examples/#load-an-older-wordpress-version
```
sejas, line 39: Even though we set the version to 6.2.1 , the .org API only returns from 6.2.6. No need to make any change. I just wanted to explain the difference between the required minimum version and what's displayed in the dropdown.

Screenshot 2025-09-30 at 17 43 50
Screenshot 2025-09-30 at 17 43 43

src/lib/wordpress-provider/wp-now/wp-now-provider.ts
```diff
@@ -45,13 +45,15 @@ export class WpNowProvider implements WordPressProvider {
 	static readonly DEFAULT_PHP_VERSION = DEFAULT_PHP_VERSION;
 	static readonly DEFAULT_WORDPRESS_VERSION = DEFAULT_WORDPRESS_VERSION;
 	static readonly ALLOWED_PHP_VERSIONS = ALLOWED_PHP_VERSIONS;
+	static readonly MINIMUM_WORDPRESS_VERSION = '6.2.6';
```
sejas, line 48: It could probably work with 6.2.1 too, but I didn't test it as it's not an option in the dropdown.

sejas (approved):
It works as expected. Thanks for making the minimum WP version parameter mandatory. 👍 .
I tested it again by creating sites with 6.2.6 (minimum version available) and nightly.

PR #1848, opened 2025-10-06
fix admin password endpoint timing issue

bcotrim:
## Related issues

- N/A

## Proposed Changes

- Fixed admin password endpoint being redirected (302) instead of executing
- Moved endpoint check to run before WordPress routing by checking request at MU plugin load time
- Changed hook from `init` to `plugins_loaded` with priority 1 to intercept request early
- Added `escapePhpString` helper to properly escape special characters in passwords

## Testing Instructions

- Start a new site 
- Open wp-admin and logout
- Try to login with Studio provided credentials

## Pre-merge Checklist

### DIFF
--- a/src/lib/wordpress-provider/playground-cli/mu-plugins.ts
+++ b/src/lib/wordpress-provider/playground-cli/mu-plugins.ts
@@ -317,14 +317,15 @@ function getStandardMuPlugins( options: Partial< WordPressServerOptions > ): MuP
 		 * This endpoint should only be accessible locally.
 		 */
 
-		add_action( 'init', function() {
-			// Only handle requests to our endpoint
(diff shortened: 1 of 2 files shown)


sejas (approved):
Works as expected and I was able to logout and log in with the user/password from the site settings.


https://github.com/user-attachments/assets/24c53b1f-24a2-4e55-a711-43dc57d25450

epeicher (approved):
Thanks @bcotrim for the changes! I have tested this, and it works as described. I am able to log in to wp-admin after logging out. I have also tested with Preview sites and can log in as expected to wp-admin there. What was the previous issue? When was the redirecting happening? Anyway, I'm approving this as it is working as expected.

bcotrim:
> When was the redirecting happening?

During the create/start site, the endpoint would sometimes redirect and not work as expected.

PR #1851, opened 2025-10-07
Add electron agnostic isOnline utility

youknowriad:
## Related issues

- Fixes STU-775

This is a small refactoring PR that adds an isOnline common utility to be used by both CLI and Electron App since net.isOnline is an electron/node only utility.

I would like to use this utility later to download the resources necessary for the CLI to work properly (launch and create sites fully without the electron app). These resources include WP latest, the WPCLI phar and SQLite command. See this commit https://github.com/Automattic/studio/commit/3f180139042061a755aa63ff261f2136c8858878 which is going to be extracted to a follow-up PR to the current one.

## Testing instructions

Ideally, this shouldn't have any impact on the electron app.

### DIFF
--- a/common/lib/network-utils.ts
+++ b/common/lib/network-utils.ts
@@ -1,0 +1,20 @@
+import dns from 'dns/promises';
+
+/**
+ * Check if the system has internet connectivity by testing DNS resolution.
+ *
+ * @returns Promise that resolves to true if online, false if offline
+ */
+export async function isOnline(): Promise< boolean > {
+	try {
+		const timeoutPromise = new Promise( ( _, reject ) =>
+			setTimeout( () => reject( new Error( 'Timeout' ) ), 5000 )
+		);
+
+		await Promise.race( [ dns.resolve( 'google.com' ), timeoutPromise ] );
+
+		return true;
+	} catch {
+		return false;
+	}
+}
--- a/common/lib/network-utils.ts
+++ b/common/lib/network-utils.ts
@@ -11,7 +11,7 @@ export async function isOnline(): Promise< boolean > {
 			setTimeout( () => reject( new Error( 'Timeout' ) ), 5000 )
 		);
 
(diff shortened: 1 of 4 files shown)


common/lib/network-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @returns Promise that resolves to true if online, false if offline
+ */
+export async function isOnline(): Promise< boolean > {
+	try {
+		const timeoutPromise = new Promise( ( _, reject ) =>
+			setTimeout( () => reject( new Error( 'Timeout' ) ), 5000 )
+		);
+
+		await Promise.race( [ dns.resolve( 'google.com' ), timeoutPromise ] );
```
youknowriad: This was the simplest suggestion from our best friend Claudio. But happy to switch to any other context agnostic approach.

sejas: This approach looks good to me. I would suggest changing the domain to our public-api.

```suggestion
		await Promise.race( [ dns.resolve( 'public-api.wordpress.com' ), timeoutPromise ] );
```

sejas (approved):
Thanks for extracting the logic so it can be used by the CLI.

I confirm that the new `isOnline` function correctly identifies each case. I tested the PR by creating a new site while online and another while offline, and both worked correctly as before.

sejas (approved):

sejas:
@youknowriad, I noticed that the issue ID in the description, `STU-755`, is not correct.

youknowriad:
You're right sorry, I updated with the right issue.

youknowriad:
@sejas Is the performance metric failing just in this PR? or have you seen this elsewhere?

sejas:
@youknowriad, I think it's failing only on this PR. I see a few PRs that were merged today and passed the performance metric.

youknowriad:
Interesting 🤔

PR #1855, opened 2025-10-07
Post performance results as PR comment

wojtekn:
## Related issues

- N/A

## Proposed Changes

- I propose to post performance results as a PR comment, so they are easily available for the PR author and reviewers, without the need to dig into GitHub action logs or CodeVitals charts.

## Testing Instructions

- TBD

## Pre-merge Checklist

### DIFF
--- a/.github/workflows/metrics-comparison.yml
+++ b/.github/workflows/metrics-comparison.yml
@@ -17,6 +17,7 @@ jobs:
     runs-on: macos-15
     permissions:
       contents: read
+      pull-requests: write
     env:
       ARTIFACTS_PATH: ${{ github.workspace }}/artifacts
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


youknowriad:
I love this.  That said if the numbers posted above are not fake, then we have a problem in terms of stability of the CI job :) `42%` is huge as a margin of error.

wojtekn:
@youknowriad I changed rounds to 3, so now it will run each test three times and will use the median value.

It seems much better, and the only downside is that it takes ~30 minutes to complete the suite.

We could consider migrating it to Buildkite at some point.

youknowriad:
In Gutenberg we had started with 3 and reverted to 1 because we found that it didn't have much impact, now it could be different here :).

youknowriad (approved):

youknowriad:
When you have big diffs like 40% In most cases, I think the issue might be in the test itself, it could be not deterministic enough but it's very hard to find.

PR #1859, opened 2025-10-08
Avoid focus loss on switching tabs

katinthehatsite:
## Related issues

- Closes STU-53

## Proposed Changes

This PR ensures that we don't lose focus when we use the keyboard navigation when switching between tabs.

## Testing Instructions

* Start Studio
* Click on one of the tabs e.g. Overview
* Try using left and right arrows to move through tabs
* Confirm that you can move between tabs and that they do not lose focus
* Compare this behaviour to trunk and observe that keyboard navigation does not work correctly there and you can't move through tabs

## Pre-merge Checklist

### DIFF
--- a/src/components/site-content-tabs.tsx
+++ b/src/components/site-content-tabs.tsx
@@ -1,5 +1,6 @@
 import { TabPanel } from '@wordpress/components';
 import { useI18n } from '@wordpress/react-i18n';
+import { useEffect, useRef, useState } from 'react';
 import { ContentTabAssistant } from 'src/components/content-tab-assistant';
 import { ContentTabImportExport } from 'src/components/content-tab-import-export';
 import { ContentTabOverview } from 'src/components/content-tab-overview';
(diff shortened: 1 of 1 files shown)


bcotrim (commented):
Have you considered changing the `TabPanel` key to `selectedSite.id`? 
I think it would also fix the issue without needing the `isFirstRender` `lastChangeWasUser` logic.

Unrelated observation: While testing keyboard navigation, I noticed that navigating to the AI Assistant tab moves focus to the input. This is pre-existing behavior from `trunk`, but it feels unexpected when navigating via keyboard. Might be worth a separate issue?

katinthehatsite:
> Have you considered changing the TabPanel key to selectedSite.id?
I think it would also fix the issue without needing the isFirstRender lastChangeWasUser logic.

I did and this approach would be simpler but would it not affect setting the tabs like this, for example? https://github.com/Automattic/studio/blob/3348f903f1c49f00153b2d4e2995b750859c9bc8/src/hooks/use-site-details.tsx#L200

bcotrim:
> I did and this approach would be simpler but would it not affect setting the tabs like this, for example? 

You are right, on `deleteSite`it would work since it also changes selected site ID, but on `useListenDeepLinkConnection` it might break. Let's keep your original solution, thanks for the clarification!

bcotrim (approved):

gcsecsey (approved):
The changes LGTM. I was also thinking about potential simplifications, thanks for the explanation above Kat for why the useEffect and the refs are needed.

Functionally, this works great:

https://github.com/user-attachments/assets/3dce9079-5bd2-494f-8fb2-10d0d96715bf

katinthehatsite:
> The changes LGTM. I was also thinking about potential simplifications, thanks for the explanation above Kat for why the useEffect and the refs are needed.

Yeah I spent some time dancing around it yesterday but could not find something simpler to make sure there is no regressions 😅

PR #1865, opened 2025-10-09
Update wpcom library to 7.0.1

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/studio/security/dependabot/56

## Proposed Changes

This PR upgrades the `wpcom` package from 5.4.2 to 7.0.1 and addresses compatibility issues with the new version.

### Key Changes

1. **Package upgrades:**
   - Upgrade `wpcom` from 5.4.2 to 7.0.1
   - Add `wpcom-xhr-request` 1.3.0 as a new dependency (required by wpcom v7)

2. **CommonJS/ESM interop fixes:**
   - Created `src/lib/wpcom-factory.ts` - Normalizes wpcom imports for both ESM (Vite/browser) and CJS (Electron main process) contexts
   - Created `src/lib/wpcom-xhr-request-factory.ts` - Normalizes wpcom-xhr-request imports to handle the default export properly in Node.js/Electron CJS contexts

3. **Browser globals polyfill for CLI:**
   - Enhanced `cli/polyfills/browser-globals.js` to include `Blob` and `File` class stubs on the `window` object
   - Fixes "Right-hand side of 'instanceof' is not an object" error when wpcom-xhr-request checks for `window.Blob` in Node.js CLI context

4. **Updated wpcom usage across the codebase:**
   - Updated all wpcom initialization to use the new factory pattern: `wpcomFactory(token, wpcomXhrRequest)`
   - Updated imports to use normalized factories instead of direct package imports
   - Updated TypeScript type definitions for better type safety using `ReturnType<typeof wpcomFactory>`
   - Fixed type annotations in callbacks (e.g., chat-slice error handling)

5. **Updated type definitions:**
   - Enhanced `src/custom-package-definitions.d.ts` with proper wpcom type interfaces
   - Added type definition for wpcom-xhr-request module

## Testing Instructions

1. Start Studio
2. Test authentication flows:
   - Log in to WordPress.com
3. Test logged user actions:
   - Test AI assistant chat functionality
   - Test preview site creation
   - Test sync site push/pull operations
4. Test CLI operations:
   - Test preview site creation

## Pre-merge Checklist


src/lib/wpcom-factory.ts
```diff
@@ -0,0 +1,9 @@
+import wpcomModule from 'wpcom';
+
+// Normalize for both ESM (Vite/browser) and CJS (Electron IPC) contexts.
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const wpcomFactory: any =
```
wojtekn: Do you know of any better way to fix it? Or should we keep it until it's improved in the `wpcom` and `wpcom-xhr-request` libraries?

bcotrim: I think we could use `optimizeDeps` in electron.vite.config

wojtekn: Like adding `wpcom` and `wpcom-xhr-request` to `optimizeDeps` -> `include`? I tried, but it seems it works only for the renderer process?

sejas: I removed this `(wpcomModule as any).default` and everything seems to work fine. Why do we need this "normalization" in the first place?

See this PR: https://github.com/Automattic/studio/pull/1898

wojtekn: Without those factories, the authentication and preview sites don't work.

sejas (commented):
I tested this PR deeply and everything worked fine. I created preview sites, pulled and pushed sites, talked to the AI Assistant.

I'm not sure why we needed the ESM/CJS normalization, so I created a PR (https://github.com/Automattic/studio/pull/1898) to confirm that it seems we might not need it at all. I also took the opportunity to simplify the `wpcomFactory` by accepting only the token as a parameter, so that the `wpcomXhrRequest` will be transparent for consumers.

sejas (approved):
I tried multiple ways to avoid normalizing the export, but I kept running into the same issue where login and logout didn’t work because the packages are used in the renderer process: https://github.com/Automattic/studio/pull/1898

So, this is the most robust solution 👍

src/lib/wpcom-factory.ts
sejas: Thanks for clarifying. I see that the wpcom calls from a React component stopped working, and adding the two dependencies to `optimizeDeps` didn't fix the issue.

src/lib/wpcom-factory.ts
```diff
@@ -0,0 +1,9 @@
+import wpcomModule from 'wpcom';
+
+// Normalize for both ESM (Vite/browser) and CJS (Electron IPC) contexts.
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
sejas: @wojtekn, maybe we could avoid using this ESLint rule by relying on inferred types.

```diff
diff --git a/src/components/auth-provider.tsx b/src/components/auth-provider.tsx
index 2eb38472..ed56c141 100644
--- a/src/components/auth-provider.tsx
+++ b/src/components/auth-provider.tsx
@@ -5,13 +5,11 @@ import { useIpcListener } from 'src/hooks/use-ipc-listener';
 import { useOffline } from 'src/hooks/use-offline';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { isInvalidTokenError } from 'src/lib/is-invalid-oauth-token-error';
-import wpcomFactory from 'src/lib/wpcom-factory';
+import wpcomFactory, { type WPCOM } from 'src/lib/wpcom-factory';
 import wpcomXhrRequest from 'src/lib/wpcom-xhr-request-factory';
 import { useI18nLocale } from 'src/stores';
 import { setWpcomClient } from 'src/stores/wpcom-api';
 
-type WPCOM = ReturnType< typeof wpcomFactory >;
-
 export interface AuthContextType {
 	client: WPCOM | undefined;
 	isAuthenticated: boolean;
diff --git a/src/lib/wpcom-factory.ts b/src/lib/wpcom-factory.ts
index a1c9b316..42d38ede 100644
--- a/src/lib/wpcom-factory.ts
+++ b/src/lib/wpcom-factory.ts
@@ -1,9 +1,11 @@
 import wpcomModule from 'wpcom';
 
 // Normalize for both ESM (Vite/browser) and CJS (Electron IPC) contexts.
-/* eslint-disable @typescript-eslint/no-explicit-any */
-const wpcomFactory: any =
-	typeof wpcomModule === 'function' ? wpcomModule : ( wpcomModule as any ).default ?? wpcomModule;
+const wpcomFactory: typeof wpcomModule =
+	typeof wpcomModule === 'function'
+		? wpcomModule
+		: ( wpcomModule as { default: typeof wpcomModule } ).default;
+
+export type WPCOM = ReturnType< typeof wpcomFactory >;
 
 export default wpcomFactory;
-/* eslint-enable @typescript-eslint/no-explicit-any */
```

src/lib/wpcom-factory.ts
```diff
@@ -0,0 +1,9 @@
...
+const wpcomFactory: any =
+	typeof wpcomModule === 'function' ? wpcomModule : ( wpcomModule as any ).default ?? wpcomModule;
```
sejas: This second fallback could be omitted.

```suggestion
	typeof wpcomModule === 'function' ? wpcomModule : ( wpcomModule as any ).default;
```

src/lib/wpcom-factory.ts
wojtekn: Thanks, I updated it in 69aebe30d8362fa0331fa0aef9e6f694d97cc572

src/lib/wpcom-factory.ts
wojtekn: Removed in 5ef78a5802b0c0b3b0ee84627bcfb9c8ba58a83c

sejas (approved):

PR #1874, opened 2025-10-10
Studio: Fix auto-installation for sites created from existing WordPress folders

bcotrim:
## Related issues

- Fixes STU-863

## Proposed Changes

- Added `installWordPress` method to the WordPressProvider interface to check and install WordPress when needed
- Implemented the method in PlaygroundCliProvider to detect if WordPress is installed by checking .ht.sqlite database file size
- If WordPress is not installed, automatically adds a blueprint step to complete installation via /wp-admin/install.php
- Added no-op implementation in WpNowProvider since wp-now handles installation automatically
- Extracted installation blueprint step to `createInstallationStep` helper function to eliminate code duplication
- Updated ipc-handlers to use the new provider method for WordPress installation

## Testing Instructions

- Enable the blueprints feature flag in settings
- Download a fresh WordPress zip from wordpress.org and extract it to a folder
- In Studio, create a new site and select "Add existing site" with the WordPress folder
- Verify that the site is created successfully without showing the WordPress installation screen
- Confirm the site loads with WordPress already installed and configured
- Test creating sites with wp-now provider (blueprints disabled) to ensure no regression

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -261,7 +261,7 @@ export async function createSite(
 		}
 		if ( ! ( await pathExists( nodePath.join( path, 'wp-config.php' ) ) ) ) {
 			await installSqliteIntegration( path );
-			await getWordPressProvider().installWordPress(
+			await getWordPressProvider().installWordPressWhenNoWpConfig(
 				server,
 				siteName || nodePath.basename( path ),
 				details.adminPassword
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -261,6 +261,11 @@ export async function createSite(
 		}
 		if ( ! ( await pathExists( nodePath.join( path, 'wp-config.php' ) ) ) ) {
 			await installSqliteIntegration( path );
+			await getWordPressProvider().installWordPress(
+				server,
+				siteName || nodePath.basename( path ),
+				details.adminPassword
+			);
 		} else {
 			await updateSiteUrl( server, getSiteUrl( details ) );
 		}
--- a/src/lib/wordpress-provider/playground-cli/playground-cli-provider.ts
+++ b/src/lib/wordpress-provider/playground-cli/playground-cli-provider.ts
@@ -1,4 +1,3 @@
-import fs from 'fs';
 import nodePath from 'path';
 import { SupportedPHPVersions } from '@php-wasm/universal';
 import { Blueprint, StepDefinition } from '@wp-playground/blueprints';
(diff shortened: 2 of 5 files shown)


src/lib/wordpress-provider/playground-cli/playground-cli-provider.ts
```diff
@@ -218,6 +221,40 @@ export class PlaygroundCliProvider implements WordPressProvider {
...
+		let needsInstallation = true;
+
+		try {
+			const stats = await fs.promises.stat( dbPath );
+			// If database exists and is larger than 10KB, WordPress is already installed
+			needsInstallation = stats.size < 10 * 1024;
+		} catch {
+			// Database doesn't exist, needs installation
+			needsInstallation = true;
+		}
```
ivan-ottinger: This is clever. I was thinking if it could fail in some edge case, but couldn't think of any. Looking at a freshly-installed WordPress DB, it should always be larger than 10kB.

ivan-ottinger (approved):
Nice! The changes look good to me and they test well too. 👍🏼 

With the changes applied, I cannot reproduce the STU-863 issue anymore. When creating a new Studio site from existing directory, the original DB is preserved.

I have tested the experience with Blueprints enabled and also disabled.

src/lib/wordpress-provider/wp-now/wp-now-provider.ts
```diff
@@ -148,6 +148,15 @@ export class WpNowProvider implements WordPressProvider {
...
 	}
 
+	async installWordPress(
+		_server: SiteServer,
+		_siteName: string,
+		_adminPassword: string
+	): Promise< void > {
+		// wp-now handles WordPress installation automatically via installationSteps
+		// in vendor/wp-now/src/wp-now.ts, so no action needed here
+	}
```
sejas: Maybe we don't need to create an empty method?

src/lib/wordpress-provider/playground-cli/playground-cli-provider.ts
sejas: Let's just check if the database exists or not. If we really want to see whether the site is installed, we should check if a user exists or something similar.
It seems that a non installed site has a database of 94KB already.

src/ipc-handlers.ts
```diff
@@ -261,6 +261,11 @@ export async function createSite(
 		}
 		if ( ! ( await pathExists( nodePath.join( path, 'wp-config.php' ) ) ) ) {
 			await installSqliteIntegration( path );
+			await getWordPressProvider().installWordPress(
+				server,
+				siteName || nodePath.basename( path ),
+				details.adminPassword
+			);
```
sejas: This logic is executed only in cases where a user downloads a WordPress zip and manually adds that site folder to Studio. That folder won't have a wp-config.php file or a .ht.sqlite database.

sejas (commented):
This PR adds the installation step for when the WordPress files are downloaded by the user manually from https://wordpress.org/download/.  
I'll make some changes to ensure we don't run the installation step when wp-config.php already exists.

src/lib/wordpress-provider/playground-cli/playground-cli-provider.ts
sejas: I changed the condition here https://github.com/Automattic/studio/pull/1874/commits/c18bd534e75d7b32f0f5942723f927a47e6719e0

If the site has wp-config.php or a sqlite database we should not run the install step.

src/lib/wordpress-provider/wp-now/wp-now-provider.ts
sejas: I tried to refactor this, but it seems a rabbit hole. Let's leave it as it is. Eventually we'll remove wp-now provider.

src/lib/wordpress-provider/playground-cli/playground-cli-provider.ts
ivan-ottinger: > Let's just check if the database exists or not. If we really want to see whether the site is installed, we should check if a user exists or something similar.

> If the site has wp-config.php or a sqlite database we should not run the install step.

Good idea, this is cleaner indeed.

ivan-ottinger:
I have reviewed the latest changes and and retested the PR. The site creation is still working correctly. I did not observe any regressions.

sejas (approved):

PR #1891, opened 2025-10-15
"Add site" Stepper: Introduce click handler to allow clicks on the previous step that is already complete

ivan-ottinger:
## Related issues

- Resolves STU-852

## Proposed Changes

- introduce `handleStepClick` to handle clicks on the previous step that is already complete
- ensure the selected import file is remembered when navigating between steps
- adjust styling of the previous step label

https://github.com/user-attachments/assets/e77084c4-c405-45bf-9233-9d86c124023e

## Testing Instructions

2. Click the `Add site` button in the app sidebar and try to navigate through the steps of `Start from Blueprint` and `Import from backup` flows.
3. When you complete the first step and get to the second one by clicking the `Continue` button, it should be possible to get back to the first step by clicking the step label (not just the `Back` button).
4. When you are in the `Import from backup` flow and navigating back to the step where the import file is selected, it should be remembered.
5. When you review how the step labels are animated during transition between steps, there should be no odd movement / shift in those labels.

## Pre-merge Checklist

### DIFF
--- a/src/modules/add-site/components/stepper.tsx
+++ b/src/modules/add-site/components/stepper.tsx
@@ -1,8 +1,4 @@
-import {
-	__experimentalHStack as HStack,
-	__experimentalText as Text,
-	useNavigator,
-} from '@wordpress/components';
+import { __experimentalHStack as HStack, __experimentalText as Text } from '@wordpress/components';
 import { useI18n } from '@wordpress/react-i18n';
 import { FormEvent } from 'react';
 import Button from 'src/components/button';
@@ -31,8 +27,7 @@ export default function Stepper( {
 	canSubmitCreate,
 }: StepperProps ) {
 	const { __ } = useI18n();
-	const { goTo } = useNavigator();
-	const { steps, isVisible, actionButton, onSubmit, canSubmit } = useStepper( {
+	const { steps, isVisible, actionButton, onSubmit, canSubmit, goTo } = useStepper( {
 		onBlueprintContinue,
 		onBackupContinue,
 		onCreateSubmit,
--- a/src/modules/add-site/components/stepper.tsx
+++ b/src/modules/add-site/components/stepper.tsx
@@ -27,7 +27,7 @@ export default function Stepper( {
 	canSubmitCreate,
 }: StepperProps ) {
 	const { __ } = useI18n();
-	const { steps, isVisible, actionButton, onSubmit, canSubmit, goTo } = useStepper( {
+	const { steps, isVisible, actionButton, onSubmit, canSubmit } = useStepper( {
 		onBlueprintContinue,
 		onBackupContinue,
 		onCreateSubmit,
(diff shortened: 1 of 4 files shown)


src/modules/add-site/components/stepper.tsx
```diff
@@ -59,8 +81,8 @@ export default function Stepper( {
 							</div>
 							<Text
 								className={ cx(
-									`text-sm`,
-									isCurrent ? 'text-gray-900 font-medium' : 'text-gray-500'
+									`text-sm font-medium`,
```
ivan-ottinger, line 70: I am curios to hear what you think about this change.

Before (you can see slight shift / movement in labels):


https://github.com/user-attachments/assets/cf418630-fb05-4e1a-9dc0-37287082cba8

After (there's no shift in the labels; instead the unselected label has thinner font size):


https://github.com/user-attachments/assets/99580575-39a2-4bdc-a5d0-65038ee1d23f

bcotrim: Nice catch!

src/modules/add-site/components/stepper.tsx
```diff
@@ -27,6 +31,7 @@ export default function Stepper( {
 	canSubmitCreate,
 }: StepperProps ) {
 	const { __ } = useI18n();
+	const { goTo } = useNavigator();
```
bcotrim: What do you think about adding this function to the `useStepper` hook to keep the navigation logic together?

src/modules/add-site/components/stepper.tsx
```diff
@@ -45,10 +50,27 @@ export default function Stepper( {
 			<HStack spacing={ 6 } alignment="left">
 				{ steps.map( ( step, index ) => {
 					const isCurrent = step.status === 'current';
+					const isCompleted = step.status === 'completed';
+					const isClickable = isCompleted && step.path;
```
bcotrim: I would also move this logic to the `useStepper` hook.

bcotrim (approved):
LGTM 👍
Added some suggestions, but feel free to merge as is.

src/modules/add-site/components/stepper.tsx
ivan-ottinger: Good idea! I have made the change in https://github.com/Automattic/studio/pull/1891/commits/bfaafaa49545415da008195d00f0ce65187a0d73.

src/modules/add-site/components/stepper.tsx
ivan-ottinger: Thanks for the suggestions, Bernardo. I agree, it make sense to move these. I have moved the click handler as well: https://github.com/Automattic/studio/pull/1891/commits/ad724a706aaa395fdd05c36666781598c5d01e65.

gcsecsey (approved):
The changes LGTM and work well. I can navigate back using the labels, and the file selection in earlier step is also preserved. 👍

ivan-ottinger:
Thank you for your reviews, Bernardo and Gergely!

PR #1913, opened 2025-10-22
Fix plugin download timeouts by using low-speed timeout instead of hard timeout

bcotrim:
## Related issues

- Fixes STU-858

## Proposed Changes

- Replaced hard CURLOPT_TIMEOUT (30s) with low-speed timeout to allow large plugin downloads
- Set CURLOPT_LOW_SPEED_LIMIT to 1KB/s and CURLOPT_LOW_SPEED_TIME to 30s in the HTTP request timeout mu-plugin
- Kept overall timeout at 300s timeout
- Updated both src/lib/wordpress-provider/playground-cli/mu-plugins.ts and vendor/wp-now/src/wp-now.ts
- This allows large downloads (like Jetpack at 33MB) to complete while still failing fast on stalled connections

## Testing Instructions

- Install Network Link Conditioner (from Apple's Additional Tools for Xcode)
- Set throttling to a slow profile (e.g., "3G" or "Edge")
- In Studio, navigate to /wp-admin/plugin-install.php
- Try installing Jetpack plugin (33MB)
- Expected: Should complete successfully even on throttled connection (may take several minutes)
- With the old 30s timeout, this would fail with "Operation timed out after 30014 milliseconds"
- Optionally verify with normal internet by turning off Network Link Conditioner and installing Jetpack - should complete quickly without any timeout issues

## Pre-merge Checklist

### DIFF
--- a/src/lib/wordpress-provider/playground-cli/mu-plugins.ts
+++ b/src/lib/wordpress-provider/playground-cli/mu-plugins.ts
@@ -259,14 +259,23 @@ function getStandardMuPlugins( options: Partial< WordPressServerOptions > ): MuP
 	muPlugins.push( {
 		filename: '0-http-request-timeout.php',
 		content: `<?php
-		// Increase default timeouts to 30 seconds to accommodate slower network conditions and larger requests
+		// Use low-speed timeout instead of hard timeout to handle both large downloads and stalled connections
+		// - Allows large plugin downloads (e.g., Jetpack 33MB) to complete with reasonable internet speeds
+		// - Fails fast if connection stalls (speed drops below 1KB/s for 30 seconds)
+		// - Provides quick feedback for genuinely broken/unresponsive servers
+		// Match WordPress core's timeout for plugin downloads (300s)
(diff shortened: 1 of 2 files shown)


ivan-ottinger (approved):
Nice! The changes look good and I can confirm they fix the issue when installing Jetpack on DSL speeds. If I tested with slower connection (e.g. 3G or EDGE), it failed as with the current trunk.

> Install Network Link Conditioner (from Apple's Additional Tools for Xcode)

This took me a bit longer to figure our where to find the tool so I am linking it here - if anyone else would like to use it:
https://developer.apple.com/download/all/?q=Additional%20Tools

It is bundled inside Additional Tools and I found the tool quite buggy. For instance, my Preferences crashed multiple times and it was necessary to force-close it.

epeicher (approved):
Thanks @bcotrim for improving this! I have tested it with DSL connection and while it was failing in Production, it has worked using this branch. Changes also LGTM! :shipit: 

|Speed | Production version|This branch
|-------|------|-------
|Image|  Image|Image

PR #1918, opened 2025-10-22
Studio: Display a more specific error message when the push process fails

katinthehatsite:
## Related issues

Related to STU-749

## Proposed Changes

 This PR adds more detailed message for the user when the database import fails during the `Push` process, writes to logs the error and provides further direction for the user in the message:

Screenshot 2025-10-24 at 11 59 34 AM

Logs:

Screenshot 2025-10-24 at 11 57 00 AM

## Testing Instructions

* Use a Studio site with the broken database (you can break the database by, for example, putting a wrong `data_type` for one of the rows in one of the database tables so that SQLite to MySQL conversion breaks)
* Navigate to `Sync` tab
* Attempt a push process to one of your WP.com sites
* Confirm that you can see the error from the screenshot when the import fails
* For now, I left the wording like this but we can adjust it further if it is not clear enough

## Pre-merge Checklist


src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -137,16 +137,29 @@ export function useSyncPush( {
...
+					error_data: response.error_data,
+				} );
+				// If the impport dails due to a SQL import error, show a more specific message
+				const restoreMessage = response.error_data?.vp_restore_message || '';
+				const isSqlImportFailure = /importing sql dump/i.test( restoreMessage );
+				const message = isSqlImportFailure
+					? __(
+							'Database import failed on the remote site. Please review your database and try again or contact support and provide details from the logs below.'
+					  )
+					: response.error === 'Import timed out'
```
sejas: We could move this check into a variable, similar to `isSqlImportFailure`.

We can also consider using if/else instead of nested ternaries.

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -137,16 +137,29 @@ export function useSyncPush( {
 				} );
 			} else if ( response.success && response.status === 'failed' ) {
 				status = pushStatesProgressInfo.failed;
+				console.error( 'Push import failed:', {
+					remoteSiteId: syncPushState.remoteSiteId,
+					error: response.error,
+					error_data: response.error_data,
+				} );
+				// If the impport dails due to a SQL import error, show a more specific message
```
sejas:
```suggestion
				// If the import fails due to a SQL import error, show a more specific message
```

sejas (approved):
Code looks good. I just found a typo in the comments and added a suggestion about using if/else instead of nested ternaries.

src/hooks/sync-sites/use-sync-push.ts
katinthehatsite: Fixed in https://github.com/Automattic/studio/pull/1918/commits/1366eefdd7d6deffff33865960df9f509ce74057

src/hooks/sync-sites/use-sync-push.ts
katinthehatsite: I moved it into varible in https://github.com/Automattic/studio/pull/1918/commits/cff4efa7d467a2efd019c8d83d7a7e8ff7c52da7

PR #1921, opened 2025-10-23
Move Windows speed-up prompt from onboarding to main window load and update its trigger logic

ivan-ottinger:
## Related issues

- Resolves STU-883

## Proposed Changes

The goal of this PR is to make the "Windows speed-up" prompt display on app start after each version update, unless the user has already confirmed the prompt.

- change the type of `PromptWindowsSpeedUpResult` from `'yes' | 'no'` to
```
{
	response: 'yes' | 'no';
	appVersion: string;
	dontAskAgain: boolean;
}
```
  → this will ensure we can record on which app version the prompt has been triggered

- move the prompt trigger from onboarding to app load
- update the prompt logic to allow it to trigger on version update, unless the user already accepted the prompt
- add handling for legacy `promptWindowsSpeedUpResult` format to ensure the prompt won't trigger if the user already accepted it
- add checkbox to let users skip the prompt even after app version updates 
- add related unit tests

Markup on 2025-10-24 at 14:51:48

## Testing Instructions

On Windows:
1. Open the `appdata-v1.json` file for editing (e.g. by running `notepad "$env:APPDATA\Studio\appdata-v1.json"` in PowerShell).
2. Check out the PR branch, build the app and start it.
3. If the `Want to speed up site creation?` prompt displays, clicking:
  - the `Sounds good, do it for me` option should record the following in the `appdata-v1.json` :
  
```
    "promptWindowsSpeedUpResult": {
    "response": "yes",
    "appVersion": "1.6.1" // should match the current app version
  }
```
  
  - the `I'm to interested` option should record the following in the `appdata-v1.json` :
  
```
    "promptWindowsSpeedUpResult": {
    "response": "no",
    "appVersion": "1.6.1" // should match the current app version
    "dontAskAgain": true | false // depending on whether you have checked the box
  }
```

4. Test the following scenarios on app start (by triggering the prompt manually through the app menu (_Help → How can I make WordPress Studio faster?_) - or - by manually editing the `appdata-v1.json` file).

a) if the `appdata-v1.json` file does not include the `promptWindowsSpeedUpResult` record, the prompt should open automatically
b) if the `appdata-v1.json` file includes `promptWindowsSpeedUpResult` in the old format and:
   - the value is `"promptWindowsSpeedUpResult": 'no',`,  the prompt should open automatically
   - the value is `"promptWindowsSpeedUpResult": 'yes',`, the prompt should not open

c) if the `appdata-v1.json` file includes `promptWindowsSpeedUpResult` in the new format and:
   - the `response` value is `no` and the `appVersion` is different from the current app version, the prompt should open automatically
   - the `response` value is `no` and the `appVersion` is the same as the current app version, the prompt should not open automatically
   - the `response` value is `yes`, the prompt should not open regardless of the `appVersion`

5. Triggering the prompt from the app menu (_Help → How can I make WordPress Studio faster?_) should work regardless of the `promptWindowsSpeedUpResult` value in the `appdata-v1.json`. The `dontAskAgain` checkbox should not render in this case.

On macOS:
1. Check out the PR branch, build the app and start it.
2. There should be no regressions.

## Pre-merge Checklist

### DIFF
--- a/src/__mocks__/electron.ts
+++ b/src/__mocks__/electron.ts
@@ -64,6 +64,10 @@ BrowserWindow.prototype.emit = jest.fn( ( event: string, ...args: any[] ) => {
 	handlers.forEach( ( handler ) => handler( ...args ) );
 } );
 
+export const dialog = {
+	showMessageBox: jest.fn(),
+};
+
 export const Menu = {
 	buildFromTemplate: jest.fn(),
 	setApplicationMenu: jest.fn(),
--- a/src/__mocks__/electron.ts
+++ b/src/__mocks__/electron.ts
@@ -37,6 +37,7 @@ BrowserWindow.prototype.isFullScreen = jest.fn( () => false );
 
 const mockWebContents = {
 	on: jest.fn(),
+	once: jest.fn(),
 	send: jest.fn(),
 	isDestroyed: jest.fn( () => false ),
 };
--- a/src/components/onboarding.tsx
+++ b/src/components/onboarding.tsx
@@ -109,11 +109,6 @@ export default function Onboarding() {
 			event.preventDefault();
 			// Save current app version to prevent What's New from showing for new users
 			await saveLastSeenVersion( window.appGlobals.appVersion );
-			try {
-				await getIpcApi().promptWindowsSpeedUpSites( { skipIfAlreadyPrompted: true } );
-			} catch ( error ) {
-				console.error( error );
-			}
 
 			try {
 				await handleAddSiteClick();
--- a/src/lib/tests/windows-helpers.test.ts
+++ b/src/lib/tests/windows-helpers.test.ts
@@ -82,6 +82,7 @@ describe( 'promptWindowsSpeedUpSites', () => {
 				promptWindowsSpeedUpResult: {
 					response: 'no',
 					appVersion: currentVersion,
+					dontAskAgain: false,
 				},
 			} );
 
@@ -97,6 +98,7 @@ describe( 'promptWindowsSpeedUpSites', () => {
 				promptWindowsSpeedUpResult: {
 					response: 'no',
 					appVersion: '1.2.2', // Previous version
+					dontAskAgain: false,
 				},
 			} );
 			mockDialogShowMessageBox.mockResolvedValue( { response: 1, checkboxChecked: false } );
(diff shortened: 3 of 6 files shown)


bcotrim (approved):
LGTM and worked as described

I wonder if we should add a third option or a checkbox to "don't ask me again"
We don't have metrics for this, as far as I know, and I imagine most users would be willing to add that permission, but for the ones that don't want to, it can be annoying to get asked on every update (every 2 weeks approx.)

What do you think?

ivan-ottinger:
Thank you for the review, Bernardo!

> I wonder if we should add a third option or a checkbox to "don't ask me again"
We don't have metrics for this, as far as I know, and I imagine most users would be willing to add that permission, but for the ones that don't want to, it can be annoying to get asked on every update (every 2 weeks approx.)

> What do you think?

Yes, I think that's a good point. I will take a look at it.

ivan-ottinger:
> Yes, I think that's a good point. I will take a look at it.

I have made the changes and added related unit tests as well.

bcotrim (approved):
Thanks for addressing the changes and adding unit tests!
LGTM 👍

wojtekn:
> We don't have metrics for this, as far as I know, and I imagine most users would be willing to add that permission, but for the ones that don't want to, it can be annoying to get asked on every update (every 2 weeks approx.)

Wasn't the main motive to let the user dismiss it, but only until the app is updated to a newer version?

wojtekn:
Adding to the previous comment: Studio doesn't check whether the real-time protection exclusion rule exists, so without the "Don't Ask me again" option, it would also display to users who already added the exclusion, correct?

ivan-ottinger:
Thanks for the questions, Wojtek.

> Wasn't the main motive to let the user dismiss it, but only until the app is updated to a newer version?

The way I understood Bernardo here is that there can be users that wouldn't be happy to see the prompt on every version update. Those users can check the "don't ask me again" box and won't be prompted again anymore - even when the version updates and regardless if they responded "yes" or "no" on that prompt.

> Adding to the previous comment: Studio doesn't check whether the real-time protection exclusion rule exists, so without the "Don't Ask me again" option, it would also display to users who already added the exclusion, correct?

The logic introduced in this PR ensures that regardless of if the user responded "yes" or "no" to the prompt before, they won't receive the prompt again if they checked the "don't ask me again" box.

Similarly, if the user already responded "yes", they won't receive the prompt on future app updates anymore - even if they did not check the "don't ask me again" box.

PR #1925, opened 2025-10-24
Add ARM64 app download nudge for users running Intel build under emulation on Windows

ivan-ottinger:
## Related issues

Resolves STU-882

## Proposed Changes

- add ARM64 app download nudge for users running Intel build under emulation on Windows
- remove platform check from the logic (since we will now display the nudge on both macOS and Windows)

Screen Shot on 2025-10-31 at 14:34:08

## Testing Instructions

1. Check out the PR branch, build and run the app on Windows for x64 (`npm run make:windows-x64`) and arm64 (`npm run make:windows-arm64`).
2. The nudge should render only when running x64 on arm64 architecture.
3. There should be no change for the nudge on macOS.

## Pre-merge Checklist

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -137,7 +137,6 @@ void Promise.all( [ getIpcApi().getAppGlobals(), getIpcApi().getSentryUserId() ]
 
 		// Show warning if running an ARM64 translator
 		if (
-			( appGlobals.platform === 'darwin' || appGlobals.platform === 'win32' ) &&
 			appGlobals.arm64Translation &&
 			! localStorage.getItem( 'dontShowARM64Warning' )
 		) {
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -137,7 +137,7 @@ void Promise.all( [ getIpcApi().getAppGlobals(), getIpcApi().getSentryUserId() ]
 
 		// Show warning if running an ARM64 translator
 		if (
-			appGlobals.platform === 'darwin' &&
+			( appGlobals.platform === 'darwin' || appGlobals.platform === 'win32' ) &&
 			appGlobals.arm64Translation &&
 			! localStorage.getItem( 'dontShowARM64Warning' )
 		) {
@@ -149,6 +149,10 @@ void Promise.all( [ getIpcApi().getAppGlobals(), getIpcApi().getSentryUserId() ]
 						window.appGlobals.platform === 'darwin'
 							? __(
 									'Downloading the Mac with Apple Silicon Chip version of Studio will provide better performance.'
+							  )
+							: window.appGlobals.platform === 'win32'
+							? __(
+									'Downloading the ARM64 version of Studio will provide better performance.'
 							  )
 							: __(
 									'Downloading the optimized version of Studio will provide better performance.'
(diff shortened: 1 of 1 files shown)


src/renderer.ts
```diff
@@ -136,23 +136,23 @@ void Promise.all( [ getIpcApi().getAppGlobals(), getIpcApi().getSentryUserId() ]
...
 
 		// Show warning if running an ARM64 translator
-		if (
-			appGlobals.platform === 'darwin' &&
-			appGlobals.arm64Translation &&
-			! localStorage.getItem( 'dontShowARM64Warning' )
-		) {
+		if ( appGlobals.arm64Translation && ! localStorage.getItem( 'dontShowARM64Warning' ) ) {
 			const showARM64MessageBox = async () => {
+				const platformMessages: Record< string, string > = {
```
bcotrim, line 141: I think these need to match the option we have on the Studio site so it's to the user what they need to download. For macOS it looks good, for Windows is still unknown as far as I know.

Image

bcotrim (commented):
Changes LGTM but I think we need to wait for https://github.com/Automattic/studio/pull/1916 to be merged, to test with both versions.

Meanwhile I confirmed the warning when running the x64 version on my Windows VM.

Image

One thing to consider:
Would it make sense to have the "Download" option automatically download and install the correct version? I believe that should be possible following a similar approach to what we do in the auto-updater.

src/renderer.ts
ivan-ottinger: That's a good catch. I will update the text in this PR once we agree on it.

ivan-ottinger:
Thank you for the review, Bernardo!

> One thing to consider:
Would it make sense to have the "Download" option automatically download and install the correct version? I believe that should be possible following a similar approach to what we do in the auto-updater.

The auto-update process works with the currently running app architecture. I think it could be doable, but at the same time I suspect there could be potential issues if we try to run update from one app architecture to another. Maybe we could explore this once the ARM64 version is out.

> I think we need to wait for https://github.com/Automattic/studio/pull/1916 to be merged, to test with both versions.

Yes, indeed. I think we can wait until the ARM64 version is launched and published on the site and merge this PR for the later Studio release.

wojtekn (commented):
The code change looks good.

wojtekn:
@ivan-ottinger @bcotrim I've published ARM binary on landing page: https://developer.wordpress.com/studio/

What do you think about simplifying those texts, keeping the architecture name the same as on the landing page?

Mac:
> Downloading the Apple Silicon Chip version of Studio will provide better performance.

Windows:
> Downloading the ARM version of Studio will provide better performance.

ivan-ottinger:
> What do you think about simplifying those texts, keeping the architecture name the same as on the landing page?

Nice! That sounds good to me. I will make the changes.

src/renderer.ts
ivan-ottinger: Just noting that the copy has been updated in https://github.com/Automattic/studio/pull/1925/commits/137745a0c4e4940fd1e5e8d4e25bcacf98f3577e.

bcotrim (approved):
LGTM 👍

PR #1929, opened 2025-10-24
Untangling prod and staging sites

nightnei:
Fixes STU-907

## Proposed Changes
We want to display each site separately in the Sync dialog, so we no longer group production and staging together. We need this to be consistent with the "create site from wpcom" stepper.

Currently WPcom sites group production and staging sites under the same row. So with this PR we split them, so each row has only a site. That's the current behaviour for Pressable sites.

Also, I analyzed the code, and I don't see the necessity to have any migration. We just stopped using `stagingSiteIds`, and they will be cleaned up automatically over time.

## Testing Instructions
1. Open wordpress.com and create site and staging site
2. Checkout out to `trunk` (don't apply this patch)
4. Open Sync tab in studio and connect that site <br />Screenshot 2025-10-24 at 13 35 25
5. `git checkout add/untanglingProdAndStagingSites`
6. Open Sync tab and assert that now you see prod and staging sites as separate entities<br />Screenshot 2025-10-24 at 13 45 04
7. Click on "Connect another site" and assert that you see both sites as separate entities and both of them "Already connected" and disabled<br />Screenshot 2025-10-24 at 13 45 41
8. Try to push/pull and assert that it still works well after untangling
9. Open wordpress.com and delete staging site
10. Open Studio and assert that you see:<br />Screenshot 2025-10-24 at 13 58 21
11. Click on "Disconnect" button
12. Open wordpress.com and delete main site
13. Open wordpress.com and delete staging site:<br />Screenshot 2025-10-24 at 14 02 38
14. Open wordpress.com and create new site and staging site
15. Open Sync tab in studio, click on "Connect another site"
16. Assert that you see both prod and staging sites, created in 14-th step
17. Connect one and then the second one
18. Assert that everything went well
19. Try to push/pull and assert that it still works well


src/modules/sync/index.tsx
```diff
@@ -122,6 +122,7 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
 	const isModalOpen = useRootSelector( connectedSitesSelectors.selectIsModalOpen );
 	const { connectedSites } = useConnectedSitesData();
 	const { syncSites, isFetching, refetchSites } = useSyncSitesData();
+
```
katinthehatsite: Looks like unintended change :)

katinthehatsite (approved):
This worked well for me 👍 I did not spot any regressions. Also, the code looked good on my end and I don't have any further concerns:

Image

Two questions that I had were:

* is this change something that we want to ship behind the feature flag or should this be available right away to the user?
* I am finding the grouping of sites a bit messy in the modal to connect sites now that they are separate. I am wondering if we should improve how the sites are organized in there e.g. perhaps first showing all production sites and then all staging sites etc. It does not need to be handled as a part of this PR though, just something to consider further in this project 👍

nightnei:
@katinthehatsite 

> is this change something that we want to ship behind the feature flag or should this be available right away to the user?

Yep, we are going to ship it w/o feature flag.

> I am finding the grouping of sites a bit messy in the modal to connect sites now that they are separate. I am wondering if we should improve how the sites are organized in there e.g. perhaps first showing all production sites and then all staging sites etc. It does not need to be handled as a part of this PR though, just something to consider further in this project 👍

I was also thinking that we should reconsider the design, for example, we can make sections narrower, at least remove the gray line between the site name and the URL, since now it makes feeling like it's 2 different items. But as you menthioned and I agree - it should be a separate issue. This PR is preety big, so I would keep it focused on one thing - untangling.

katinthehatsite:
> But as you menthioned and I agree - it should be a separate issue. This PR is preety big, so I would keep it focused on one thing - untangling.

Sounds good, we can open a separate issue to handle this and to ask for design input 👍

nightnei:
@katinthehatsite created issue in Linear STU-918
@sejas when you get a chance, could you please also take a look at this PR and at the created issue

src/hooks/use-fetch-wpcom-sites/reconcile-connected-sites.tsx
```diff
@@ -1,24 +1,14 @@
...
- *   2. Find any staging sites that have been added to an already connected site
- *   3. Find any connected staging sites that have been deleted on WordPress.com
- *
- * We treat staging sites differently from production sites because users can't connect staging
- * sites separately from production sites (they're always connected together). So, while deleted
- * production sites are still rendered in the UI (with a "deleted" notice), we need to automatically
- * keep the list of staging sites up-to-date, which is where `stagingSitesToAdd` and
- * `stagingSitesToDelete` comes in.
+ * Generate updated site data to be stored in `appdata-v1.json`:
+ * -- Update the list of `connectedSites` with fresh data (name, URL, etc)
```
sejas, line 5: Wow! this simplified the logic a lot! 🥳

src/modules/sync/components/sync-sites-modal-selector.tsx
```diff
@@ -285,12 +282,7 @@ function SiteItem( {
...
-						<>
-							<EnvironmentBadge type={ getSiteEnvironment( site ) } selected={ isSelected } />
-							{ site.stagingSiteIds.length > 0 && (
-								<EnvironmentBadge type="staging" selected={ isSelected } />
-							) }
-						</>
+						<EnvironmentBadge type={ getSiteEnvironment( site ) } selected={ isSelected } />
 					) }
 					{ isPressable && (
 						<EnvironmentBadge type={ getSiteEnvironment( site ) } selected={ isSelected } />
```
sejas: Now Pressable and WPcom sites can use the same code. Let's unify the logic.

```tsx
				<div className="flex gap-2">
					<EnvironmentBadge type={ getSiteEnvironment( site ) } selected={ isSelected } />
				</div>
```

sejas (approved):
Great work! I confirm that the Staging and Production sites are now treated equally, and users can connect to them independently.
I confirm that existing connections were ported correctly to separate blocks, and connecting to and disconnecting sites work as expected.

I just left a suggestion to simplify the logic on `src/modules/sync/components/sync-sites-modal-selector.tsx:284`

| Before | After |
|--------|--------|
| Screenshot 2025-10-28 at 00 03 03 | Screenshot 2025-10-28 at 00 03 23 |

src/modules/sync/components/sync-sites-modal-selector.tsx
nightnei: Oh, yeah, missed it, thanks 👍

PR #1940, opened 2025-10-27
Studio: Use local tree for push

katinthehatsite:
## Related issues

Fixes STU-756

## Proposed Changes

This PR uses local tree of files from Studio sites for push operations instead of using hard-coded folders. For the `plugins` and `themes` folders, the depth is limited to individual plugins and themes and the user can not go further in. For folders like `uploads`, the user can go further into separate folders and individual uploads. Some files also remain hidden from the user such as SQLite integration database plugin:

Screenshot 2025-10-29 at 11 33 04 AM

## Testing Instructions

* Navigate to the `Sync` tab and connect a WPCOM site
* Click on the `Push` option
* Observe that the file structure reflects the local files (besides the files that should be hidden like `sqlite-integration` plugin)
* Confirm that in the `plugins` and `themes`, we can only go to the individual plugins and themes and can't get deeper
* Select some individual files
* Start the push process
* Confirm that it pushes selected files only (you can do this, for example, by removing the code that cleans up the zip to inspect the zip file that gets produced) 
* Confirm that push succeeds 

## Pre-merge Checklist


src/modules/sync/lib/tree-utils.ts
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	if ( SYNC_EXCLUSIONS.includes( itemName ) ) {
+		return true;
+	}
+
+	if ( relativePath.includes( 'mu-plugins' ) && itemName === 'sqlite-database-integration' ) {
+		return true;
+	}
+
```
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/e542b7022f833d3b8994476ca3ace25077c59f6e

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -12,8 +13,53 @@ type PushOptionsWithSelections = {
...
+
+	for ( const exclusion of SYNC_EXCLUSIONS ) {
+		if ( normalizedPath === exclusion || normalizedPath.startsWith( exclusion ) ) {
+			return true;
+		}
+	}
+
+	if ( normalizedPath.includes( 'mu-plugins/sqlite-database-integration' ) ) {
+		return true;
+	}
```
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/b67445fe810b44dde6f8454d63464b6d8c8b46f2

src/ipc-handlers.ts
```diff
@@ -1702,31 +1703,101 @@ export function comparePaths( event: IpcMainInvokeEvent, path1: string, path2: s
...
-			} );
+		const entries = await fs.promises.readdir( fullPath, { withFileTypes: true } );
+		const result = [];
+
+		for ( const entry of entries ) {
+			const isDirectory = entry.isDirectory();
+			const itemPath = path.endsWith( '/' )
+				? `${ path }${ entry.name }`
+				: `${ path }/${ entry.name }`;
+			const fullItemPath = itemPath.endsWith( '/' ) || ! isDirectory ? itemPath : `${ itemPath }/`;
```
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/20b694af4e4a521a87fdc7d0ae512108e1b979b6

src/modules/sync/hooks/use-selected-items-push-size.ts
```diff
@@ -45,9 +45,13 @@ export function useSelectedItemsPushSize(
...
 				for ( const child of wpContentNode.children ) {
 					if ( child.checked ) {
-						sizePromises.push(
-							getIpcApi().getDirectorySize( siteId, [ 'wp-content', child.name ] )
-						);
+						if ( child.type === 'file' ) {
+							sizePromises.push( getIpcApi().getFileSize( siteId, [ 'wp-content', child.name ] ) );
+						} else {
+							sizePromises.push(
+								getIpcApi().getDirectorySize( siteId, [ 'wp-content', child.name ] )
```
katinthehatsite, line 52: This is needed because before we only had directories in `wp-content` but now we can also have files such as `index.php`

src/ipc-handlers.ts
```diff
@@ -1710,31 +1711,96 @@ export function comparePaths( event: IpcMainInvokeEvent, path1: string, path2: s
...
 
-export async function listWpContentFolders(
-	_event: Electron.IpcMainInvokeEvent,
-	siteId: string,
-	subdir: GranularSyncFolders
-): Promise< { name: string; type: 'file' | 'folder' }[] > {
-	const server = SiteServer.get( siteId );
-	if ( ! server ) throw new Error( 'Site not found' );
-	const wpContentPath = nodePath.join( server.details.path, 'wp-content', subdir );
+async function listLocalFileTreeRecursive(
```
katinthehatsite: I am loading the full tree in this function instead of going node by node. I tried a different approach but it creates a weird scrolling effect in the UI otherwise

katinthehatsite:
Forgot to add to testing instructions that you can also add the `Restore` logs here https://mc.a8c.com/rewind/logs.php after `push`

src/ipc-handlers.ts
bcotrim: Some of the input/output of this function look like UI only related and shouldn't be included in the ipc-handler imo (parentChecked, expanded, hideExpandButton)
Ideally this function should only have the responsibility to load the file structure and return it, with a depth limit. The renderer should be responsible for mapping and keeping all the UI related state.
What do you think? Does it make sense?

bcotrim (commented):
Tests work as described and code LGTM outside of one suggestion.

src/ipc-handlers.ts
katinthehatsite: Good suggestion, I agree ❤️ 

I will refactor tomorrow and ping you again for review!

fredrikekelund:
We could probably remove `src/modules/sync/hooks/use-default-sync-tree.tsx` as part of this PR, right?

fredrikekelund:
I see now that we can't remove it altogether, given the current logic, but we could at least remove the `wp-content` children from `useDefaultSyncTree`

katinthehatsite:
> I see now that we can't remove it altogether, given the current logic, but we could at least remove the wp-content children from useDefaultSyncTree

I am going to take a look 👀 I am first going to refactor some code to address Bernardo's comment and then check `useDefaultSyncTree`

src/ipc-handlers.ts
katinthehatsite: I added some changes in https://github.com/Automattic/studio/pull/1940/commits/4a72cb2f87da94e9d8fb165763b823f686263ad1 to address this.

src/modules/sync/lib/tree-utils.ts
```diff
@@ -0,0 +1,79 @@
...
+	return false;
+};
+
+export const shouldLimitDepth = ( relativePath: string ): boolean => {
+	const normalizedPath = relativePath.replace( /^wp-content\//, '' );
+
+	// Match plugins/plugin-name or plugins/plugin-name/
```
katinthehatsite, line 20: Left these comments here for now but they could also be removed

bcotrim (approved):
Thanks for making the changes 
There are some comments from Copilot I think should be addressed.

One comment: should we hide empty folders? I think it doesn't make a lot of sense to allow the user to create a push with empty folders. What do you think?

src/stores/sync/sync-hooks.ts
```diff
@@ -71,3 +73,37 @@ export function useRemoteFileTree() {
...
+			setIsLoading( true );
+			setError( null );
+			try {
+				const rawNodes = await getIpcApi().listLocalFileTree( siteId, path, 3 );
+				return convertRawToTreeNodes( rawNodes, parentChecked );
+			} catch ( err ) {
+				const error = err instanceof Error ? err : new Error( String( err ) );
+				setError( error );
+				console.error( 'Failed to fetch local file tree:', error );
+				return null;
```
fredrikekelund: Could we return an empty array here?

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -87,33 +85,42 @@ const useDynamicTreeState = (
...
+				try {
+					const localTree = await fetchLocalChildren( localSiteId, 'wp-content', false );
+					if ( ! isCancelled && localTree ) {
+						setTreeState( ( treeState ) =>
+							updateNodeById( treeState, 'wp-content', { children: localTree || [] } )
+						);
+					}
+				} catch ( error ) {
+					console.error( 'Failed to load local file tree:', error );
+				}
```
fredrikekelund, line 99: It seems like neither `fetchLocalChildren` nor `updateNodeById` throws exceptions. Could we remove the try..catch statement here?

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -87,33 +85,42 @@ const useDynamicTreeState = (
...
+			};
```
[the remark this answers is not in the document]
fredrikekelund: If we change `fetchLocalChildren` to always return an array, we don't need this. But it's also true that it looks like we don't really display errors from `fetchLocalChildren` anywhere in the UI now, right? That seems like something we should look into.

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -168,30 +171,51 @@ export function SyncDialog( {
...
+
+	const treeViewProps = useMemo(
+		() => ( {
+			disabled: isErrorRewindId,
+			tree: treeState,
+			setTree: setTreeState,
+			onExpand: handleExpand,
+		} ),
+		[ isErrorRewindId, treeState, setTreeState, handleExpand ]
+	);
```
fredrikekelund: Why did we move these props to a memoized object?

src/modules/sync/lib/convert-tree-to-sync-options.ts
fredrikekelund, line 1: Windows-style paths need to be considered in this file, if I'm not mistaken.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -12,19 +13,124 @@ type PushOptionsWithSelections = {
...
+
+	return false;
 };
 
-const collectPathIds = ( nodes: TreeNode[], pathIds: string[] = [] ): string[] => {
+const getPathSyncOption = ( path: string ): SyncOption | null => {
+	if ( shouldExcludePathFromSync( path ) ) {
+		return null;
+	}
```
fredrikekelund: While there are no guarantees that the path passed to this function shouldn't be excluded, this check is redundant in practice now, since `collectSelectedPaths` already excludes relevant paths, and we use that function to filter the paths passed to `groupPathsBySyncOption`.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -37,45 +143,32 @@ const getCommonNodes = ( tree: TreeNode[] ) => {
...
-				}
+	if ( wpContent?.children?.length ) {
+		const selectedPaths = collectSelectedPaths( wpContent.children );
+		const syncOptionGroups = groupPathsBySyncOption( selectedPaths );
+
+		PATH_SYNC_MAP.forEach( ( { option } ) => {
+			if ( syncOptionGroups[ option ] ) {
+				optionsToSync.push( option );
 			}
 		} );
```
fredrikekelund:
```suggestion
		const selectedPaths = collectSelectedPaths( wpContent.children );

		PATH_SYNC_MAP.forEach( ( { option, pattern } ) => {
			const patternExistsAmongPaths = selectedPaths.some( ( path ) => {
				const normalizedPath = normalizePath( path );
				return pattern.test( normalizedPath );
			} );

			if ( patternExistsAmongPaths ) {
				optionsToSync.push( option );
			}
		} );
```
I found the logic in this file a little hard to follow at times, and I think we can simplify it along these lines. I haven't accounted for Unix/Windows-style paths here. The `normalizePath` function seems like a good place to do that. 

Making this change would allow us to remove the `getPathSyncOption` and `groupPathsBySyncOption` functions.

katinthehatsite:
> One comment: should we hide empty folders? I think it doesn't make a lot of sense to allow the user to create a push with empty folders. What do you think?

I don't have a strong preference. I will let @sejas to decide on this as he might have a preference. If I remember correctly, @sejas mentioned that the idea is for the user to see everything they have locally (besides some files we decide to hide) but yeah, it does not make much sense to allow the empty folders to be selected. Even if we allow them to be selected, then wew don't push any data in these folders so it is sort of pointless to have them selected 😅  

 I am going to merge this PR today to unblock two other issues I am working on and I will open a new issue to discuss this (since Antonio is AFK today). We can apply the changes in the follow up before the next release.

fredrikekelund (commented):
This is working pretty well in the app! Nice improvement 👍

I think there are some changes we should make to the code before landing this PR, though. I'm still reviewing, but figured I'd leave an initial batch of comments if you're currently working on this, @katinthehatsite.

src/ipc-handlers.ts
```diff
@@ -1711,27 +1712,61 @@ export function comparePaths( event: IpcMainInvokeEvent, path1: string, path2: s
...
+		for ( const entry of entries ) {
+			if ( shouldExcludeFromSync( entry.name ) ) {
+				continue;
+			}
+
+			const isDirectory = entry.isDirectory();
+			const itemPath = path.endsWith( '/' )
+				? `${ path }${ entry.name }`
+				: `${ path }/${ entry.name }`;
+			const fullItemPath = isDirectory ? `${ itemPath }/` : itemPath;
```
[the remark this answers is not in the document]
sejas: Agreed! This probably won't work in Windows.

src/modules/sync/lib/tree-utils.ts
```diff
@@ -0,0 +1,79 @@
...
+	if ( itemName.startsWith( '.' ) ) {
+		return true;
+	}
+
+	if ( SYNC_EXCLUSIONS.includes( itemName ) ) {
+		return true;
+	}
+
+	return false;
+};
```
fredrikekelund, line 15: There's a very similar function in `convert-tree-to-sync-options.ts`. Let's stick with a single one in both files.

src/modules/sync/components/sync-dialog.tsx
katinthehatsite: Ah yes, this one is a leftover. I tried a different approach initially instead of preloading the whole tree. It did not work well though with that approach. I will adjust it back 👍

src/modules/sync/lib/tree-utils.ts
```diff
@@ -0,0 +1,79 @@
...
+				} else if ( themeRegex.test( normalizedPath ) ) {
+					nodeType = 'theme';
+				}
+			}
+
+			const children = rawNode.children
+				? convertRawToTreeNodes( rawNode.children, parentChecked )
+				: rawNode.isDirectory
+				? []
+				: undefined;
```
fredrikekelund:
```suggestion
			let children: TreeNode[] | undefined;
			if ( rawNode.children ) {
				children = convertRawToTreeNodes( rawNode.children, parentChecked );
			} else if ( rawNode.isDirectory ) {
				children = [];
			} else {
				children = undefined;
			}
```
Let's unwrap the nested ternary here.

src/ipc-handlers.ts
```diff
@@ -1711,27 +1712,61 @@ export function comparePaths( event: IpcMainInvokeEvent, path1: string, path2: s
 	return arePathsEqual( path1, path2 );
 }
 
-export async function listWpContentFolders(
+export async function listLocalFileTree(
```
fredrikekelund, line 1760: Maybe this function is actually the most appropriate place to handle Unix/Window style paths. We could convert all paths to Unix style before returning. 

My apologies if this has already been considered, but it doesn't look like it at first glance.

src/modules/sync/lib/tree-utils.ts
```diff
@@ -0,0 +1,79 @@
...
+
+			return {
+				id: `local-${ rawNode.path.replace( pathCleanRegex, '-' ) }`,
+				name: rawNode.name,
+				label: rawNode.name,
+				checked: parentChecked,
```
fredrikekelund: `checked` is mostly relevant for pull operations, right? Do we even need the `parentChecked` argument for this function, or could we just always specify `false` here?

src/modules/sync/types.ts
```diff
@@ -1,3 +1,10 @@
 import { GRANULAR_SYNC_FOLDERS } from 'src/modules/sync/constants';
 
 export type GranularSyncFolders = ( typeof GRANULAR_SYNC_FOLDERS )[ number ];
+
+export type RawFileNode = {
```
fredrikekelund:
```suggestion
export type RawDirectoryEntry = {
```
It doesn't have to be this exact name, but I think `RawFileNode` is a little misleading when it can also represent a directory. The name I suggested is inspired by node's `fs.Dirent` type (see [docs](https://nodejs.org/api/fs.html#class-fsdirent))

sejas (commented):
@katinthehatsite thanks for creating the PR. The UI works as expected and I see the local files tree.
Regarding the backup produced, I think we are not including the elements in the root folder, and for some reason we are always including the mu-plugins folder.

For example, In this selection I found the following discrepancies:
- wp-content/mu-plugins shouldn't be added because I didn't select the folder.
- wp-content/uploads/2025/10 shouldn't be added because I didn't select the folder.
- wp-content/test-wp-content.php should be included and it wasn't

| Selected | Produced |
|--------|--------|
| Screenshot 2025-10-31 at 10 16 26 | Screenshot 2025-10-31 at 10 13 06 |

https://github.com/user-attachments/assets/31516dd1-8a09-45c4-8713-7a83b75ca01a

src/ipc-handlers.ts
```diff
@@ -1711,27 +1712,61 @@ export function comparePaths( event: IpcMainInvokeEvent, path1: string, path2: s
...
 	_event: Electron.IpcMainInvokeEvent,
 	siteId: string,
-	subdir: GranularSyncFolders
-): Promise< { name: string; type: 'file' | 'folder' }[] > {
+	path: string = 'wp-content',
```
```suggestion
	path: string,
```
Nit, but it seems reasonable to require a path argument for this function.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -12,19 +13,124 @@ type PushOptionsWithSelections = {
...
+	} );
+
+	Object.entries( collections ).forEach( ( [ key, set ] ) => {
+		if ( set.size > 0 ) {
+			specificSelections[ key as keyof typeof collections ] = Array.from( set );
+		}
+	} );
+
+	return Object.keys( specificSelections ).length > 0 ? specificSelections : undefined;
 };
```
fredrikekelund: If I select a file like `wp-content/example.jpg` in my sync push, it won't currently be included in the push archive. We should either change this behavior so that it's sent to the server, or update the file modal to not display files if they don't live inside the `plugins`, `themes` or `uploads` folders. 

EDIT: I see that @sejas also commented on this https://github.com/Automattic/studio/pull/1940#pullrequestreview-3403443147

sejas:
> I don't have a strong preference. I will let @sejas to decide on this as he might have a preference. If I remember correctly, @sejas mentioned that the idea is for the user to see everything they have locally (besides some files we decide to hide) but yeah, it does not make much sense to allow the empty folders to be selected. Even if we allow them to be selected, then wew don't push any data in these folders so it is sort of pointless to have them selected 😅

I think it's better to respect the filesystem as much as we can, so I vote to leave the empty folders there. In the same way that your computer doesn't hide empty folders from your file explorer, maybe a user wants to create a placeholder to upload things to, or a plugin requires that empty folder for some reason.

src/modules/sync/hooks/use-top-level-sync-tree.tsx
fredrikekelund, line 1: I think it'd make sense to rename this file and function to something like `useTopLevelSyncTree`. None of the tree entries in this file are ever replaced, so I think "default" is a little misleading for a name after we land this PR.

fredrikekelund (commented):
My final batch of comments.

src/modules/sync/lib/tree-utils.ts
```diff
@@ -0,0 +1,79 @@
...
+				children,
+				expanded: false,
+				hideExpandButton: shouldLimitDepth( rawNode.path ),
+			};
+		} )
+		.sort( ( a, b ) => {
+			if ( a.type !== b.type ) {
+				return a.type === 'folder' ? -1 : 1;
+			}
+			return a.name.toLowerCase().localeCompare( b.name.toLowerCase() );
```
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/db17cb22448e6ed256a15fe8caa5650ca09d9b01

src/modules/sync/constants.ts
```diff
@@ -1 +1,20 @@
...
+	'fonts',
+] as const;
+
+export const SYNC_EXCLUSIONS = [
+	'database',
+	'database/',
+	'db.php',
+	'debug.log',
+	'sqlite-database-integration',
+	'sqlite-database-integration/',
```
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/98a575271158bd041cab81f99f028619a00134f9

src/ipc-handlers.ts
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/ea1bd161efab48383298c7b16da9554c54916c5a

katinthehatsite:
> @katinthehatsite thanks for creating the PR. The UI works as expected and I see the local files tree. Regarding the backup produced, I think we are not including the elements in the root folder, and for some reason we are always including the mu-plugins folder.
> 
> For example, In this selection I found the following discrepancies:
> 
> * wp-content/mu-plugins shouldn't be added because I didn't select the folder.
> * wp-content/uploads/2025/10 shouldn't be added because I didn't select the folder.
> * wp-content/test-wp-content.php should be included and it wasn't
> 
> Selected	Produced
> Screenshot 2025-10-31 at 10 16 26	Screenshot 2025-10-31 at 10 13 06
>  push-selective-sync.mp4

I will take a look now, when I tested it, zip looked correct and the correct files were pushed but I have made a bunch of changes since then which probably affected something in the logic 👀

src/stores/sync/sync-hooks.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/369f7b874a3b9eac8c9edc43613eaa2d0045157d

src/modules/sync/lib/convert-tree-to-sync-options.ts
katinthehatsite: Nice catch, I made updates in https://github.com/Automattic/studio/pull/1940/commits/963af9ba5fab108e0094d509e6adce9f862bd6f3

src/modules/sync/lib/convert-tree-to-sync-options.ts
katinthehatsite: Nice simplification. I applied this suggestion and cleaned up the unused functions in https://github.com/Automattic/studio/pull/1940/commits/a9cfee8b228b145ac8b45c41b02ccb61a976b48b

src/ipc-handlers.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/30c8bc1d8bcedb6aa276d950c32d59bf4f848599

src/modules/sync/types.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/c42857775a5d6882f8aa24ee5b9f2866671073d9

src/modules/sync/hooks/use-top-level-sync-tree.tsx
katinthehatsite: Sounds good, updated in https://github.com/Automattic/studio/pull/1940/commits/d90828c9da74e962260b1591ccf53d9849273fd6

src/modules/sync/components/sync-dialog.tsx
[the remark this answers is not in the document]
katinthehatsite: > On line 94, localTree is already checked for truthiness on line 92, making the localTree || [] fallback redundant.

For now, just addressing this bit: https://github.com/Automattic/studio/pull/1940/commits/224bd8a042ef9f9727477f4f85993b65ce9f36ea

src/modules/sync/lib/tree-utils.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/2fa32c64770cf4a1e9d334d158baf2dbe4446c9d

src/modules/sync/lib/tree-utils.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/df2721d9dc6b7f0974002634238a65ace61c073b

src/ipc-handlers.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/f4e8b0f1a6c2644631e4db2324bc0f3ea9bfdf00

src/modules/sync/lib/convert-tree-to-sync-options.ts
katinthehatsite: I adjusted it in one of the commits: https://github.com/Automattic/studio/pull/1940/commits/f4e8b0f1a6c2644631e4db2324bc0f3ea9bfdf00

src/modules/sync/components/sync-dialog.tsx
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/747bece6eaef864b50a5c8bdd63aa0ef676456b1

katinthehatsite:
@sejas  the functionality should be working as expected now, I believe. I also double-checked the files that were pushed to WP.com and see the correct selection based on what I chose by connecting through SFTP: 

Screenshot 2025-11-03 at 11 20 56 AM

Would you be able to re-test again and let me know if it works for you better now? 🙇

src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -166,8 +166,30 @@ export class DefaultExporter extends EventEmitter implements Exporter {
...
+							( category === 'muPlugins' &&
+								( itemName.startsWith( 'languages/' ) || itemName === 'languages' ) ) ||
+							isRootFile;
+
+						if ( belongsToCategory ) {
+							itemPath = path.join( this.options.site.path, 'wp-content', itemName );
+							itemArchivePath = path.join( 'wp-content', itemName );
+						} else {
+							continue;
+						}
```
fredrikekelund: So if the file lives directly inside `wp-content`, we don't add it to the archive..? Shouldn't we exclude it from the modal tree view in the first place, then?

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -9,78 +9,136 @@ type PushOptionsWithSelections = {
...
+			if ( rel ) categories.uploads.add( rel );
+			return;
+		}
+
+		const contentsRoots = [ 'mu-plugins', 'fonts', 'languages' ];
+		const isContents =
+			contentsRoots.some( ( root ) => p === root || p.startsWith( `${ root }/` ) ) ||
+			! p.includes( '/' );
+
+		if ( isContents ) categories.contents.add( p );
```
fredrikekelund:
```suggestion
		if ( isContents ) {
			categories.contents.add( p );
		}
```
Let's use brackets for all control statements. This applies to several places in the code.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -9,78 +9,136 @@ type PushOptionsWithSelections = {
...
-		} else if ( node.indeterminate && node.children ) {
-			collectPathIds( node.children, pathIds );
+const categorizeSelectedPathsInPlace = (
+	nodes: TreeNode[] | undefined,
+	categories: Categories
+): void => {
+	traverseSelected( nodes, ( node ) => {
+		if ( ! node.path ) return;
+
+		const p = normalizePath( node.path );
```
fredrikekelund:
```suggestion
		const nodePath = normalizePath( node.path );
```
Use whatever name you like, but let's use something descriptive.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -9,78 +9,130 @@ type PushOptionsWithSelections = {
...
-const collectPathIds = ( nodes: TreeNode[], pathIds: string[] = [] ): string[] => {
-	nodes.forEach( ( node ) => {
-		if ( node.checked && node.pathId ) {
-			pathIds.push( node.pathId );
-		} else if ( node.indeterminate && node.children ) {
-			collectPathIds( node.children, pathIds );
+const categorizeSelectedPathsInPlace = (
+	nodes: TreeNode[] | undefined,
+	categories: Categories
+): void => {
```
fredrikekelund:
```suggestion
const convertTreeToSyncCategories = ( nodes: TreeNode[] | undefined ) => {
	const categories: Categories = {
		plugins: new Set< string >(),
		themes: new Set< string >(),
		uploads: new Set< string >(),
		contents: new Set< string >(),
	};
```
Let's construct the `categories` object inside this function instead. There's no reason to accept it as an argument. 

I also think there's an opportunity to rename this function for added clarity, but that's just a suggestion.

src/modules/sync/lib/convert-tree-to-sync-options.ts
fredrikekelund, line 1: I'd love some brief comments for the major functions in this file to help explain the logic to other devs.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -9,78 +9,130 @@ type PushOptionsWithSelections = {
...
+	{ prefix: 'themes/', key: 'themes' as const, option: SYNC_OPTIONS.themes },
+	{ prefix: 'uploads/', key: 'uploads' as const, option: SYNC_OPTIONS.uploads },
+] as const;
+
+const CONTENTS_ROOTS = [ 'mu-plugins', 'fonts', 'languages' ] as const;
+
+const normalizePath = ( path: string ): string =>
+	path.replace( /^\/?wp-content\//, '' ).replace( /^\/+|\/+$/g, '' );
+
+const traverseSelected = (
```
fredrikekelund:
```suggestion
const iterateOverCheckedNodes = (
```
Sorry for all the nitpicks. I'm making sense of the logic in this file and commenting on things as I go. In this case, I think we can rename the function for added clarity.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -9,78 +9,130 @@ type PushOptionsWithSelections = {
...
 		}
+
+		const isContents =
+			CONTENTS_ROOTS.some( ( root ) => p === root || p.startsWith( `${ root }/` ) ) ||
+			! p.includes( '/' );
+
+		if ( isContents ) categories.contents.add( p );
 	} );
-	return pathIds;
 };
```
fredrikekelund:
```suggestion
const convertTreeToSyncCategories = ( nodes: TreeNode[] | undefined ) => {
	const categories = {
		plugins: { files: new Set< string >(), option: SYNC_OPTIONS.plugins },
		themes: { files: new Set< string >(), option: SYNC_OPTIONS.themes },
		uploads: { files: new Set< string >(), option: SYNC_OPTIONS.uploads },
		contents: { files: new Set< string >(), option: SYNC_OPTIONS.contents },
	};

	iterateOverCheckedNodes( nodes, ( node ) => {
		if ( ! node.path ) return;

		const isPlugin = node.path.startsWith( '/wp-content/plugins/' );
		const isTheme = node.path.startsWith( '/wp-content/themes/' );
		const isUpload = node.path.startsWith( '/wp-content/uploads/' );
		const isContent =
			node.path.startsWith( '/wp-content/mu-plugins/' ) ||
			node.path.startsWith( '/wp-content/languages/' ) ||
			node.path.startsWith( '/wp-content/fonts/' ) ||
			! node.path.includes( '/' );

		if ( isPlugin ) {
			categories.plugins.files.add( node.path );
		} else if ( isTheme ) {
			categories.themes.files.add( node.path );
		} else if ( isUpload ) {
			categories.uploads.files.add( node.path );
		} else if ( isContent ) {
			categories.contents.files.add( node.path );
		}
	} );

	return Object.values( categories );
```
This is just a suggestion, but I believe we simplify the logic in this file in several ways by modifying the function along these lines. It lets us remove `STANDARD_CATEGORIES`, `CONTENTS_ROOTS`, and `normalizePath` altogether, and it groups the logic more clearly, IMO. 

In `convertTreeToPushOptions`, we can consume the function like so:

```ts
		const categories = categorizeSelectedPathsInPlace( wpContent.children );

		for ( const { files, option } of categories ) {
			if ( files.size ) {
				optionsToSync.push( option );
				specificSelections[ option ] = [ ...files ];
			}
		}
```

fredrikekelund (commented):
I left a couple of comments in the `convert-tree-to-sync-options.ts` with opportunities to simplify the code. They're not all blockers, but there's a lot of stuff happening between the sync modal and actually sending the archive to the server. I think the code would benefit from simplifications where possible.

sejas (commented):
@katinthehatsite , thanks for keep improving this PR.

I give it a try and plugins and themes worked as expected, and I found that mu-plugins and root files are not included in the backup file:

| Selected | Result |
|--------|--------|
|  select-root | result-root |
| select-wpcomsh | result-wpcomsh |

katinthehatsite:
Thanks for testing @sejas - I can't seem to get the same result as you 😅 This is what I am getting on the most recent version of this branch:

**mu-plugins**

Screenshot 2025-11-03 at 1 40 01 PM

**selection**

Screenshot 2025-11-03 at 1 39 44 PM

I also could not reproduce the issue with the root files as I am getting it exported and pushed successfully to remote site.

Let's coordinate on Slack.

src/modules/sync/lib/convert-tree-to-sync-options.ts
katinthehatsite: Sounds good to me, I will add those once I confirm with Antonio that everything works as expected 👌 and the major pieces are in place

src/modules/sync/lib/convert-tree-to-sync-options.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/76fc204d9de4852bbf5c0511a16b572367b48c7b

src/modules/sync/lib/convert-tree-to-sync-options.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/1ed60953698ef49ae7f4b2792b1c769ab1dfe1d1

katinthehatsite:
@sejas @fredrikekelund I made some more changes to the exporter 😅  so feel free to give it another test. I will work on fixing tests in the meantime

src/modules/sync/lib/convert-tree-to-sync-options.ts
katinthehatsite: This function was changed so marking as resolved for now until further review.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -6,81 +6,100 @@ import type { SyncOption } from 'src/types';
...
-		} else if ( node.indeterminate && node.children ) {
-			collectPathIds( node.children, pathIds );
+const convertTreeToSyncCategories = (
+	nodes: TreeNode[] | undefined
+): { paths: string[]; options: SyncOption[] } => {
+	const paths = new Set< string >();
+	const options = new Set< SyncOption >();
+
+	iterateOverCheckedNodes( nodes, ( node ) => {
+		if ( ! node.path ) return;
```
fredrikekelund: I don't know if this has been discussed, and I see that the ESLint `curly` [rule](https://eslint.org/docs/latest/rules/curly) isn't enabled, but I would strongly argue that all control statements should always have brackets.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -6,81 +6,100 @@ import type { SyncOption } from 'src/types';
...
+
+		const p = node.path.replace( /^\/?wp-content\//, '' );
```
fredrikekelund:
```suggestion
		const nodePath = node.path.replace( /^\/?wp-content\//, '' );
```
Nit, but there's no reason not to use descriptive variable names.

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -6,81 +6,100 @@ import type { SyncOption } from 'src/types';
...
+	return out;
 };
 
-const collectPathIds = ( nodes: TreeNode[], pathIds: string[] = [] ): string[] => {
-	nodes.forEach( ( node ) => {
-		if ( node.checked && node.pathId ) {
-			pathIds.push( node.pathId );
-		} else if ( node.indeterminate && node.children ) {
-			collectPathIds( node.children, pathIds );
+const convertTreeToSyncCategories = (
```
fredrikekelund, line 33: Nice work simplifying this file! It's a lot easier to follow the logic now 👍

src/lib/import-export/export/exporters/default-exporter.ts
fredrikekelund, line 1: Took me a minute to wrap my head around the changes in this file, but they make a lot of sense. Nice work 👍

src/lib/import-export/export/types.ts
```diff
@@ -8,21 +8,23 @@ export interface ExportOptions {
 	phpVersion: string;
 	splitDatabaseDumpByTable?: boolean;
 	specificSelections?: {
-		plugins?: string[];
-		themes?: string[];
-		uploads?: string[];
+		paths?: string[];
 	};
```
fredrikekelund:
```suggestion
specificSelectionPaths?: string[];
```
Could we simplify this even further, like so?

src/ipc-handlers.ts
```diff
@@ -1711,27 +1712,59 @@ export function comparePaths( event: IpcMainInvokeEvent, path1: string, path2: s
...
+		const result = [];
+
+		for ( const entry of entries ) {
+			if ( shouldExcludeFromSync( entry.name ) ) {
+				continue;
+			}
+
+			const isDirectory = entry.isDirectory();
+			const itemPath = nodePath.join( path, entry.name ).replace( /\\/g, '/' );
+			const fullItemPath = isDirectory ? itemPath + '/' : itemPath;
```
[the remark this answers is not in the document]
fredrikekelund: This stood out to me when reviewing the code just now, too. What's the significance of the trailing slash? It seems to almost just add more requirements to the `shouldLimitDepth` function.

src/ipc-handlers.ts
```diff
@@ -1711,27 +1712,59 @@ export function comparePaths( event: IpcMainInvokeEvent, path1: string, path2: s
...
+
+			const fileNode: RawDirectoryEntry = {
```
```suggestion
			const directoryEntry: RawDirectoryEntry = {
```
Nit, but let's rename the variable, too, if it can represent both files and directories.

src/modules/sync/constants.ts
```diff
@@ -1 +1,10 @@
...
+export const SYNC_EXCLUSIONS = [
+	'database',
+	'db.php',
+	'debug.log',
+	'sqlite-database-integration',
+	'.DS_Store',
+	'Thumbs.db',
+];
+
+export type SyncExclusion = ( typeof SYNC_EXCLUSIONS )[ number ];
```
fredrikekelund:
```suggestion
```
This type appears to be unused.

src/modules/sync/constants.ts
```diff
@@ -1 +1,10 @@
-export const GRANULAR_SYNC_FOLDERS = [ 'plugins', 'themes', 'uploads' ] as const;
+export const SYNC_EXCLUSIONS = [
```
fredrikekelund, line 1: How about including `.git` and `node_modules` here, too? It would clarify expectations for users, since we ultimately exclude those folders in the exporter, anyway. 

Also, maybe `wp-content/cache` ought to be included. Seems like something users typically don't want to sync.

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -168,29 +171,45 @@ export function SyncDialog( {
...
+		},
+		[ type, rewindId, remoteSite.id, fetchChildren ]
+	);
+
+	const treeViewProps = {
+		disabled: isErrorRewindId,
+		tree: treeState,
+		setTree: setTreeState,
+		onExpand: handleExpand,
 	};
```
fredrikekelund: Wasn't the plain to move these props back to L299?

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -87,33 +85,42 @@ const useDynamicTreeState = (
...
-					} );
-				} );
-
-				return newState;
-			} );
+			let isCancelled = false;
+			const loadLocalTree = async () => {
+				try {
+					const localTree = await fetchLocalChildren( localSiteId, 'wp-content' );
+					if ( ! isCancelled && localTree ) {
```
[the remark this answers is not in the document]
fredrikekelund: Plus one to this.

fredrikekelund (commented):
This is shaping up nicely! 👍 I can confirm that root-level files in wp-content are included in the push (still), and the logic is easier to follow now. 

I left some comments that would be nice to address before landing, but I think we're on the home stretch now.

src/lib/import-export/export/types.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/b3db695fe4a22b92160a0120befb1764207e738f

src/ipc-handlers.ts
katinthehatsite: Sounds good, adjusted in https://github.com/Automattic/studio/pull/1940/commits/b3db695fe4a22b92160a0120befb1764207e738f

src/modules/sync/constants.ts
katinthehatsite: Applied in https://github.com/Automattic/studio/pull/1940/commits/6faa332eb6e7c2ad769c81ae55abcd0f56a5b976 

I initially used this for `ipc-handlers` but since then I refactored the function so it is no longer needed.

src/modules/sync/constants.ts
katinthehatsite: Sounds good, adjusted in https://github.com/Automattic/studio/pull/1940/commits/e6170b736328ad3c8dd90afcf0ceb9e501fa2696

src/modules/sync/components/sync-dialog.tsx
katinthehatsite: I moved them initially I think but then refactored more stuff and they ended up outside 😅 In any case, adjusted in https://github.com/Automattic/studio/pull/1940/commits/85aff5d5bdbb2bd2d8425148a4c0774026635232

src/modules/sync/components/sync-dialog.tsx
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/9a5ea618635758b9357db9bab5abe6e4cd229abd

src/modules/sync/lib/convert-tree-to-sync-options.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/5db98b7b459df0d147a4c0d3a4bc1bab6ee4987a

src/modules/sync/lib/convert-tree-to-sync-options.ts
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/0b3ad02f3f4927c40e5f82492a6f5ac0b48879f7

src/ipc-handlers.ts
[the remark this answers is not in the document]
katinthehatsite: Cleaned up in: https://github.com/Automattic/studio/pull/1940/commits/0b3ad02f3f4927c40e5f82492a6f5ac0b48879f7

I tested the functionality and it seems to be working well without this 👍

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -6,81 +6,100 @@ import type { SyncOption } from 'src/types';
...
+	}
+};
+
+const collectPathIds = ( nodes: TreeNode[] | undefined ): string[] => {
+	const out: string[] = [];
+	iterateOverCheckedNodes( nodes, ( node ) => {
+		if ( node.pathId ) out.push( node.pathId );
+	} );
+	return out;
 };
```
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/3fd8dce773e927dc358bd0266747ad4403f8853b

src/modules/sync/lib/convert-tree-to-sync-options.ts
```diff
@@ -12,19 +13,124 @@ type PushOptionsWithSelections = {
...
+	const normalizedPath = normalizePath( path );
+
+	for ( const exclusion of SYNC_EXCLUSIONS ) {
+		if ( normalizedPath === exclusion || normalizedPath.startsWith( exclusion ) ) {
+			return true;
+		}
+	}
+
+	return false;
 };
```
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/d3ae5c4d20ff8ce0a3391461158d0e8f6084d9a5

src/modules/sync/lib/tree-utils.ts
```diff
@@ -0,0 +1,80 @@
...
+			if ( rawNode.children ) {
+				children = convertRawToTreeNodes( rawNode.children );
+			} else if ( rawNode.isDirectory ) {
+				children = [];
+			} else {
+				children = undefined;
+			}
+
+			return {
+				id: `local-${ rawNode.path.replace( pathCleanRegex, '-' ) }`,
```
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/d3ae5c4d20ff8ce0a3391461158d0e8f6084d9a5

src/modules/sync/lib/convert-tree-to-sync-options.ts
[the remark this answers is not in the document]
katinthehatsite: https://github.com/Automattic/studio/pull/1940/commits/3fd8dce773e927dc358bd0266747ad4403f8853b

src/modules/sync/components/sync-dialog.tsx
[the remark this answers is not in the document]
katinthehatsite: > But it's also true that it looks like we don't really display errors from fetchLocalChildren anywhere in the UI now, right? 

I will create a separate issue for this so that we can get some clarity on design for this use case.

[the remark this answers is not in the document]
katinthehatsite: https://linear.app/a8c/issue/STU-966/studio-add-ui-for-when-local-tree-for-push-has-errors

src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -154,72 +153,58 @@ export class DefaultExporter extends EventEmitter implements Exporter {
...
+							isExcluded ||
+							entry.name.includes( '.git' ) ||
+							entry.name.includes( 'node_modules' ) ||
+							entry.name.includes( 'cache' )
+						) {
+							return false;
 						}
-					}
+						return entry;
+					} );
```
sejas, line 202: We have the opportunity to unify this if-else logic related to full vs partial export.
For example, currently the push doesn't exclude `sqlite-database-integration`, and the export site doesn't include other extra folders or files inside the wp-content folder.
We can tackle that as a follow-up.

sejas (approved):
Great work Kat!
I tested many cases and they work pretty well. The only issue I found is that selecting wp-content (or just mu-plugins) also includes the `sqlite-database-integration`, which should be excluded.

Screenshot 2025-11-05 at 00 14 07

And selecting all (folders and database) it correctly excludes the `sqlite-database-integration`, however doesn't include the extra folders or files like `extra-folder` or `test-wp-content.php`.

Screenshot 2025-11-05 at 00 13 33

We could create a follow-up issue for that, but we should probably fix it before the next release.

src/lib/import-export/export/exporters/default-exporter.ts
sejas: @katinthehatsite , I'll create a follow-up PR.

katinthehatsite: > For example, currently the push doesn't exclude sqlite-database-integration, and the export site doesn't include other extra folders or files inside the wp-content folder.

Yeah I saw the bit with export, we could unify it. I left it as is for now to preserve the existing functionality.

For the `sqlite-database-integration`, let's fix it in this PR. It must be something small because this was working before. I will take a look in a bit

katinthehatsite: To close the loop: Since we discussed this on Slack, we will solve both issues in https://github.com/Automattic/studio/pull/2004 and proceed with merging this one.

sejas: - Follow-up PR: https://github.com/Automattic/studio/pull/2006

PR #1942, opened 2025-10-27
Update deleted sync site UI

nightnei:
## Proposed Changes
When a site was deleted in wpcom, we render not specific message, so I propose:
1. Say that the site was most likely deleted
2. Render site url, since now we have only site name. Also make site url clickable, so that user can open and quickly access that site to check what is wrong.

## Testing Instructions
1. Create wpcom site
2. In Studio, open Sync tab and connect it
3. Delete that site in wpcom
4. Open Sync tab again
5. Assert that you see the new message<br />
<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2025-10-27 at 17 07 46
<td>Screenshot 2025-10-27 at 17 06 16
</tr>
</table>

7. Click on site name
8. Assert that you we redirected to that site and see the message:<br />Screenshot 2025-10-27 at 17 08 36


epeicher (approved):
Thanks @nightnei for improving this! I have tested it, and I can see the URL and the new message, which I agree is more informative. 

One small improvement I would try is to style the URL link with the same blue than the `Get help` to indicate that is a link, as I didn't realize that it was clickable until hovering over it. But apart from that, this LGTM! 

|Studio | Calypso|
|-------|------|
| Image|Image|

nightnei:
> One small improvement I would try is to style the URL link with the same blue than the Get help to indicate that is a link, as I didn't realize that it was clickable until hovering over it.

@epeicher great point 👍 Adjusted

PR #1943, opened 2025-10-27
Push: Check existing progress of push on start

epeicher:
## Related issues

- Related to STU-751

## Proposed Changes
- Check if there is any Push in progress when starting Studio and updates the progress accordingly
- Adds a `useEffect` to `sync-sites-context.tsx` that only runs once to check if there is any connected site with a push operation in progress
- Update the `pushStates` of the sites if there is any status in progress

https://github.com/user-attachments/assets/d5f40178-f944-47cc-b68b-0976f6b20934

## Testing Instructions

* Apply these changes and run `npm start` 
* Navigate to the `Sync` tab and connect a site or use any existing connected site
* Select Push
* Wait for the status to be progressed to `Backing up remote site…`
* Press `Cmd+Q` to quit Studio
* Check that the dialog appears and notify that the operation will continue. Accept and quit Studio
* Open Studio again
* Navigate to the Sync tab
* Check the progress continues

## Pre-merge Checklist


src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -99,6 +108,79 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
...
+					setPushStates( restoredStates );
+				}
+			} catch ( error ) {
+				// Initialization is not critical to app functionality, but log the error
+				console.error( 'Failed to initialize push states from server:', error );
+			}
+		};
+
+		void initializePushStates();
+	}, [ client, pushStatesProgressInfo, connectedSites ] );
```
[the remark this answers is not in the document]
epeicher: `connectedSites` is memoized as it uses a Redux selector, additionally, the `hasInitialized` ref prevents the effect from running multiple times, so this should not be a concern

src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -99,6 +108,79 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
...
+						}
+					} catch ( error ) {
+						// Continue checking other sites even if one fails
+						console.error( `Failed to check push progress for site ${ connectedSite.id }:`, error );
+					}
+				}
+
+				// Update the push states with all restored operations
+				if ( Object.keys( restoredStates ).length > 0 ) {
+					setPushStates( restoredStates );
```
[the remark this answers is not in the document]
epeicher: This should be executed during initialization so there should not be any previous states.

src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -99,6 +108,79 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
 	const { syncSites, isFetching, refetchSites } = useSyncSitesData();
 	useListenDeepLinkConnection( { refetchSites } );
 
+	const { client } = useAuth();
+	const { pushStatesProgressInfo } = useSyncStatesProgressInfo();
+	const hasInitialized = useRef( false );
```
sejas, line 113: What if we use a regular variable instead of `useRef`?

src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -99,6 +108,79 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
...
+			try {
+				// Get all local sites to map site IDs to SiteDetails
+				const allSites = await getIpcApi().getSiteDetails();
+
+				const restoredStates: PushStates = {};
+
+				// Query the server for each connected site to check for in-progress operations
+				for ( const connectedSite of connectedSites ) {
+					try {
+						// Find the local site
```
sejas: Let's remove these comments when they are unnecessary.

sejas (commented):
Feature-wise, it works as described. I confirm that it continues polling push information from the backend after restarting the app. ✨

https://github.com/user-attachments/assets/1c1d5605-47a2-4941-8414-2f7774b7ff0e

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -64,6 +64,33 @@ export type UseSyncPush = {
...
 };
 
+/**
+ * Maps an ImportResponse status to a PushStateProgressInfo object.
+ * Returns null if the operation is finished or failed (should not be restored).
+ */
+export function mapImportResponseToPushState(
+	response: ImportResponse,
+	pushStatesProgressInfo: Record< PushStateProgressInfo[ 'key' ], PushStateProgressInfo >
+): PushStateProgressInfo | null {
```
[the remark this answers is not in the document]
epeicher: Done as part of [449d839](https://github.com/Automattic/studio/pull/1943/commits/449d83917eaca3a7ce4f86702abafeb8c4c083a3)

src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -99,6 +108,79 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
...
+							continue;
+						}
+
+						// Query the server for the current import status
+						const response = ( await client.req.get(
+							`/sites/${ connectedSite.id }/studio-app/sync/import`,
+							{
+								apiNamespace: 'wpcom/v2',
+							}
+						) ) as ImportResponse;
```
fredrikekelund:
```suggestion
						const response = await client.req.get< ImportResponse >(
							`/sites/${ connectedSite.id }/studio-app/sync/import`,
							{ apiNamespace: 'wpcom/v2' }
						);

```
The wpcom.js request functions support generics for the response body. This way, we avoid short-circuiting the type checker with the `as` operator.

src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -99,6 +108,79 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
...
+	useEffect( () => {
+		// Only run once, when client and connectedSites are both available
+		if ( hasInitialized.current || ! client || connectedSites.length === 0 ) {
+			return;
+		}
+
+		hasInitialized.current = true;
+
+		const initializePushStates = async () => {
+			try {
```
fredrikekelund: This try-catch basically only catches exceptions from `getIpcApi().getSiteDetails()`, right? I'm not sure that's really necessary, but if we keep it, then maybe we can wrap just that row instead of the entire function body to reduce indentation.

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -64,6 +64,29 @@ export type UseSyncPush = {
 	cancelPush: CancelPush;
 };
 
+/**
+ * Maps an ImportResponse status to a PushStateProgressInfo object.
+ * Returns null if the operation is not in progress or unknown.
+ */
+export function mapImportResponseToPushState(
```
fredrikekelund, line 71: Could we use this function in `getPushProgressInfo`, too? It'd be nice to map the API status to a local value in a single place.

src/hooks/sync-sites/sync-sites-context.tsx
fredrikekelund: I don't believe that would persist the value.

src/hooks/sync-sites/sync-sites-context.tsx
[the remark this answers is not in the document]
fredrikekelund: In practice, you're probably right, @epeicher. Still, this was something I reacted to, as well. Seems like an unnecessary risk to take when merging with the existing state, instead of overwriting the entire object, is so simple. Call it future proofing.

src/hooks/sync-sites/sync-sites-context.tsx
epeicher: If we use a regular variable here, its value will be reset on each re-render, but this ref is used to keep the value between re-renders. Alternatively, we could use a global variable outside of the component, but that will be reused by all instances of this component, which does not seem better. 

Let me know if you see a better approach or if you think we should explore alternatives!

src/hooks/sync-sites/sync-sites-context.tsx
epeicher: Sure, I have removed and re-worded some of them here [5b5f0a1](https://github.com/Automattic/studio/pull/1943/commits/5b5f0a1764948638f438bf5bea6e4c8b8657eaa9) but please suggest if other comments feel also unnecessary, happy to remove them 😄

fredrikekelund (commented):
This is a great addition 👍 I haven't finished my review, but since I see you're actively working on this PR, @epeicher, I'll submit some of the feedback I have so far.

fredrikekelund (approved):
I wish we had untangled some of the push/pull React context and moved it to Redux slices, but that's beyond the scope of this PR. 

Again, great addition 👍 Functionality-wise, this works as described. The code looks good overall, but I'd prefer we address the comments from my previous review (especially [this](https://github.com/Automattic/studio/pull/1943/files#r2472285443)). Approving now to unblock you later, @epeicher.

src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -99,6 +108,72 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
...
+
+				const restoredStates: PushStates = {};
+
+				for ( const connectedSite of connectedSites ) {
+					try {
+						const selectedSite = allSites.find( ( site ) => site.id === connectedSite.localSiteId );
+
+						if ( ! selectedSite ) {
+							continue;
+						}
```
sejas: I think we shouldn't limit the code for selected sites, and we should try to "resume" all the connected sites.

For example a user could start a push, then change the site, close the app, open Studio and currently the push is not "resumed", because the selected site is different. So I would expect that when a user goes back to the previous site, the push continues.

https://github.com/user-attachments/assets/aa68a177-7b8e-4ff9-8b1b-887a56a7539c

src/hooks/sync-sites/sync-sites-context.tsx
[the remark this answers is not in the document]
epeicher: I agree, simplicity and future proofing are win-win 😄 , I am updating that

src/hooks/sync-sites/sync-sites-context.tsx
epeicher: This could also catch potential exceptions thrown by [this request](https://github.com/Automattic/studio/pull/1943/files#diff-ecf9c95b9648c5c173a02f6de0bd65dc125e05eee6e9104f9e9bbb4a87368413R138), for example, being offline, so I think it's good that it covers both

fredrikekelund: There's already an inner try-catch statement to capture exceptions from that API call, though. That's why I figured we could make this statement wrap fewer lines.

src/hooks/sync-sites/sync-sites-context.tsx
sejas: If we want to optimize the code and check only those sites that have started pushing, we could initialize `lastPushTimestamp` when the push starts and then use that variable to filter for connected sites that have initiated any push. However, this could be addressed in a follow-up issue.

My reasoning is that we are moving towards allowing users to start a push, complete the upload, and then let them close Studio. In that case, `lastPushTimestamp` will never be set, even if Studio starts after the push has finished.

src/hooks/sync-sites/sync-sites-context.tsx
epeicher: Thanks for the suggestion, much better now, added as part of 46bda48a

src/hooks/sync-sites/sync-sites-context.tsx
sejas: Thanks for the explanation. I think it's a good approach. I'm extra suspicious of states being saved in useRef, but it makes sense for this use case.

src/hooks/sync-sites/sync-sites-context.tsx
fredrikekelund: Is that really what happens there, though? The way I interpret the code, this is just to grab a `SiteDetails` reference to be used in the `SyncPushState` object. It's not actually related to which site is selected in the UI. 

In your video, maybe the push just finished before you returned to the Sync tab for the relevant site?

src/hooks/sync-sites/sync-sites-context.tsx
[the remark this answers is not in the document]
epeicher: Done as part of fd67b238

src/hooks/sync-sites/sync-sites-context.tsx
epeicher: Ah, you're right, I mixed it with the inner one. 

I could reduce the indentation by narrowing the scope, but it would require extracting the `allSites` variable outside of the try-catch and checking if it's null before continuing after the try-catch, which could be less readable. Alternatively, I could remove this try-catch as it could be unnecessary, what option do you prefer?

src/hooks/sync-sites/sync-sites-context.tsx
sejas: I double-checked and I'm still thinking that the resume push only works for the selected site. Here is a short video to confirm it. I start pushing a site, select a different site and then close the app, which didn’t resume the push, and then restart it again, this time selecting the correct site first, which successfully resumes the push.

https://github.com/user-attachments/assets/4955d4e1-f54e-400f-97fb-2b876b0fba23

fredrikekelund: Oh, wait. The problem is that the `connectedSites` array only contains sites for the site that's currently selected in the UI. So the problem isn't on these lines, but you're definitely right that this is an issue, @sejas. Nice catch.

sejas: So, we were both right! Teamwork! 🙌

epeicher: Great catch indeed! I am going to remove the selected site limitation and check for all connectedSites instead. Thanks for the suggestion!

src/hooks/sync-sites/use-sync-push.ts
epeicher: I looked into refactoring this, but `getPushProgressInfo` does more than mapping. We could reuse it in lines 179 and 181, but we’d then need extra null checks, so there’s no real improvement.

I also explored adding `apiStatus` to the `pushStatesProgressInfo` object, but the result felt less clear than the current approach.

I’m definitely open to alternatives if you have a different idea!

src/hooks/sync-sites/sync-sites-context.tsx
epeicher: I have fixed this as part of be16976f (and 0201f628). I have also renamed `selectedSite` to `localSite` as it better reflects what the code is doing, as the `localSite` does not necessarily is _selected_. 

Could you please review it and let me know your view on those changes?

src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -99,6 +108,73 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
 	const { syncSites, isFetching, refetchSites } = useSyncSitesData();
 	useListenDeepLinkConnection( { refetchSites } );
 
+	const { client } = useAuth();
+	const { pushStatesProgressInfo } = useSyncStatesProgressInfo();
+	const hasInitialized = useRef( false );
+
+	// Initialize push states from in-progress server operations on mount
+	useEffect( () => {
+		// Only run once, when client and connectedSites are both available
```
fredrikekelund:
```suggestion
```
Nit, but this comment is stale, and I'd argue it's superfluous in the first place. Let's go ahead and remove it.

src/hooks/sync-sites/sync-sites-context.tsx
fredrikekelund: How about removing this clause and wrapping the `initializePushStates` call in a try..catch instead?

src/hooks/sync-sites/sync-sites-context.tsx
fredrikekelund: Nice, I can confirm that it works as expected now 👍

fredrikekelund (commented):
Some final minor comments

src/hooks/sync-sites/sync-sites-context.tsx
epeicher: I have tested it and in that case, because we are not awaiting the promise because of the `void` operator, an error in any of the `await getIpcApi()...` will be propagated and we'll get the following error in the UI
CleanShot 2025-10-29 at 14 54 19@2x

fredrikekelund: You could add a `.catch` handler to the `initializePushStates` call and remove the `void` operator.

epeicher: Yes, of course, I have tested that and it works. Do you think that is more readable than the try-catch in the `initializePushStates`? I honestly think we would be mixing promise handling (the `.catch`) with async-await, but if you have a strong opinion about it, I am happy to change that

src/hooks/sync-sites/sync-sites-context.tsx
epeicher: Agreed it does not add much value and it became stale after latest change, removed [5938e04](https://github.com/Automattic/studio/pull/1943/commits/5938e042a4538be9e60f8b1ff01514351b0236f0)

sejas (approved):
Works as expected. I confirm all ongoing pushes are resumed after restarting Studio.
Thanks for fixing the issue I caught by mistake, and renaming the localSite variable.

src/hooks/sync-sites/sync-sites-context.tsx
fredrikekelund: Principally, I don't think it's wrong to mix "classic" promise handling with async/await. The root problem here is essentially that `useEffect` can't handle async callbacks, but we still want _some_ way to run async code inside that function. To that end, it's more about finding the least clunky solution.

However, this specific issue is really mostly cosmetic and not worth getting hung up on, IMO. Feel free to make a call and proceed as you see fit, @epeicher 👍

epeicher: > The root problem here is essentially that useEffect can't handle async callbacks, but we still want some way to run async code inside that function. 

💯 I agree, that's the root issue

> it's more about finding the least clunky solution.

I didn't find clunky to have the outer try-catch, but you and two different agents think the `.catch` is less clunky so [e5fb0f8](https://github.com/Automattic/studio/pull/1943/commits/e5fb0f8adce3e5e70fb20ddf8e5eaa790b4040aa) 😄

wojtekn:
@epeicher This is a great improvement to sync UX 🥳

PR #1946, opened 2025-10-28
Add support for appx for arm64 architecture

wojtekn:
## Related issues

- STU-59

## Proposed Changes

AppX packages were hardcoded with `ProcessorArchitecture="x64"` for all builds, causing Microsoft Store to reject ARM64 packages as duplicates.

I propose patching `electron2appx` via `patches/electron2appx+2.1.2.patch` to respect the `FILE_ARCHITECTURE` environment variable when generating manifests. The script now uses architecture-specific input directories and Windows Kit paths.

I considered those alternatives:
- Manual manifest creation before creating appx - too much code duplication
- Polling to fix manifest just after it is generated -  possible race condition concerns 
- Migration to Electron Builder - migration complexity
- Migration to Electron Forge maker-appx - it also uses the same lib, so doesn't solve the problem

Separately, I can contribute the patch upstream to the `electron2appx` repository.

## Testing Instructions

- Confirm if the builds work fine
- Confirm that Microsoft Store accepts two packages: x64 and arm64 and doesn't trigger duplicate alert

## Pre-merge Checklist

### DIFF
--- a/scripts/package-appx.mjs
+++ b/scripts/package-appx.mjs
@@ -13,6 +13,13 @@ if ( ! process.env.WINDOWS_CODE_SIGNING_CERT_PASSWORD ) {
 
 const __dirname = path.dirname( fileURLToPath( import.meta.url ) );
 
+// Get architecture from environment variable, default to x64 for backward compatibility
+const architecture = process.env.FILE_ARCHITECTURE || 'x64';
+if ( architecture !== 'x64' && architecture !== 'arm64' ) {
+	console.error( `Invalid architecture: ${ architecture }. Must be 'x64' or 'arm64'.` );
+	process.exit( 1 );
+}
+
 const windows10SDKVersionPath = path.resolve( __dirname, '..', '.windows-10-sdk-version' );
 try {
 	await fs.access( windows10SDKVersionPath );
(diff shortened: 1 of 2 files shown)


wojtekn:
I tested the `.appx` files downloaded from the build. The manifest for both files includes correct architectures:

|**arm64**|**x64**|
|---|---|
|Screenshot 2025-10-29 at 08 15 33|Screenshot 2025-10-29 at 08 15 43|

I checked `.exe` files and they are also correct - one is arm64 and another one is x64.

Microsoft Store now sees it as an unique package with correct architecture:

Screenshot 2025-10-29 at 08 40 33

scripts/package-appx.mjs
```diff
@@ -22,6 +29,8 @@ try {
 }
 const windows10SDKVersionContent = await fs.readFile( windows10SDKVersionPath );
 const windows10SDKVersion = windows10SDKVersionContent.toString().trim();
+// Windows SDK tools (makeappx.exe, signtool.exe) are always in the x64 directory,
+// regardless of the target architecture. The architecture only affects the manifest.
```
AliSoftware, line 33: Thanks for commenting that explicitly 👍

AliSoftware (approved):
Approving to unblock and because the patch looks reasonable to me.

I haven't tested it myself, but since you did validate it on your end I think we should be good.

---

cc @mokagio because I remember him working on this in the past (internal ref: paaHJt-8gi-p2 — see "AppX arm64 gotchas" section).

Though his work on that was on Electron projects other than studio, with those other projects potentially using either Electron Builder or Electron Forge and thus not being in the exact same context than your case for Studio here (and with the limitations you mentions in your PR description about migrating to those).

But maybe Gio will still have some insights in how he solved that problem for those other projects back then?

wojtekn:
Thanks @AliSoftware , I answered there as well.

PR #1947, opened 2025-10-28
Update wpcom.js and clean types

wojtekn:
## Related issues

- N/A

## Proposed Changes

- I propose updating wpcom.js and cleaning types following the type definition added in the dependent library (https://github.com/Automattic/wp-calypso/pull/106576).

## Testing Instructions

- Confirm if builds go through
- Confirm authentication flow and other WP.com API requests work fine

## Pre-merge Checklist


src/components/auth-provider.tsx
```diff
@@ -87,21 +87,16 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 	const logout = useCallback( async () => {
 		try {
 			if ( ! isOffline && client ) {
-				client.request(
-					{
+				try {
+					await client.req.del( {
```
wojtekn: Let's remove direct use of request handler.

katinthehatsite (approved):
I did not spot any regressions 👍

src/components/auth-provider.tsx
sejas: Using the `client.request` seems necessary. The new implementation doesn't use the DELETE method and instead uses POST. Although, we could update our endpoint `https://public-api.wordpress.com/wpcom/v2/studio-app/token` to also accept the POST method.

Screenshot 2025-10-28 at 19 21 13

Testing this behaviour I found that on trunk the endpoint returns a 401 error 😅 .

Screenshot 2025-10-28 at 19 23 19

sejas (commented):
The code looks good. I tested the PR by creating a preview site, logging out and logging in, pulling a remote site, and interacting with the assistant. Everything worked as expected except for logging out, which returned a 404 error.

katinthehatsite:
> Everything worked as expected except for logging out, which returned a 404 error.

Huh, interesting! It still works for me for logging out correctly but I do see the same error that you mentioned in console.

sejas:
Yeah, logging out still works because it simply removes the token from AppData. However, the API call fails to invalidate the existing token. That error is not visible for users.

wojtekn:
Good catch @sejas ! I checked the code more, and it seems it's done that way because all v1 APIs use POST for delete methods (See https://developer.wordpress.com/docs/api/rest-api-reference/), and v2 uses proper DELETE requests. I think it's better to leave the endpoint part strict here. I set the method explicitly in f46ba669a63cd3754a9c8e468a2a9ec2342ec424 and it fixes the issue.

fredrikekelund (commented):
I see we removed a lot of Typescript generics in #1865 and opted to use `as` statements instead for the return types from wpcom.js req functions.

The new wpcom.js types still support generics, and while they are functionally identical to how we use the `as` operator, it is still safer to use generics to accommodate future changes to the wpcom.js API signatures. 

I recommend using this PR to reinstate use of Typescript generics. 

To clarify, I mean doing this:


```ts
const response = await client.req.get< SyncBackupResponse >(
	`/sites/${ remoteSiteId }/studio-app/sync/backup`,
	{
		apiNamespace: 'wpcom/v2',
		backup_id: backupId,
	}
);
```

instead of this

```ts
const response = ( await client.req.get(
	`/sites/${ remoteSiteId }/studio-app/sync/backup`,
	{
		apiNamespace: 'wpcom/v2',
		backup_id: backupId,
	}
) ) as SyncBackupResponse;

**Comment by wojtekn** [comment]
@fredrikekelund thanks, changed in 4a0489035dc7f8a1c90f708b7f5985f9be3ae6d9

File: src/components/auth-provider.tsx [hunk 2]
```diff
@@ -87,21 +87,18 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 	const logout = useCallback( async () => {
 		try {
 			if ( ! isOffline && client ) {
-				client.request(
-					{
+				try {
```
fredrikekelund: Do we really need the inner try..catch statement here?

src/components/auth-provider.tsx
```diff
@@ -199,20 +192,26 @@ function createWpcomClient(
...
+			}
+		};
 
-			return originalRequestHandler( addLocaleToParams( params ), wrappedCallback );
-		},
-	} );
+		return wpcomXhrRequest( modifiedParams, wrappedCallback );
+	};
+
+	const wpcom = wpcomFactory( token, wrappedRequestHandler as typeof wpcomXhrRequest );
```
fredrikekelund:
```suggestion
	const wpcom = wpcomFactory( token, wrappedRequestHandler );
```
Doesn't look like we need the `as` operator here.

src/components/auth-provider.tsx
```diff
@@ -181,8 +176,6 @@ function createWpcomClient(
 		}
 	};
 
-	const originalRequestHandler = wpcom.request.bind( wpcom );
-
 	const addLocaleToParams = ( params: WpcomParams ) => {
```
fredrikekelund, line 178:
```suggestion
	const addLocaleToParams = ( params: object ) => {
```
Seems like we can just use the `object` type here.

src/components/auth-provider.tsx
```diff
@@ -199,20 +192,26 @@ function createWpcomClient(
...
-				if ( typeof callback === 'function' ) {
-					callback( err, response, headers );
-				}
-			};
+	// Wrap the request handler to add locale and error handling before passing to wpcomFactory
+	const wrappedRequestHandler = (
+		params: object,
+		callback: ( err: unknown, response?: unknown, headers?: unknown ) => void
+	) => {
+		const modifiedParams = addLocaleToParams( params as WpcomParams );
```
fredrikekelund, line 199:
```suggestion
		const modifiedParams = addLocaleToParams( params );
```
Per my other suggestion of sticking to the `object` type.

src/custom-package-definitions.d.ts
```diff
@@ -48,67 +48,4 @@ declare module '@timfish/forge-externals-plugin' {
...
-
-		( token?: string | WPCOMRequestHandler, reqHandler?: WPCOMRequestHandler ): WPCOMInstance;
-	}
-
-	const WPCOM: WPCOMConstructor;
-	export default WPCOM;
-	/* eslint-enable @typescript-eslint/no-explicit-any */
-}
-
 declare module 'wpcom-xhr-request';
```
fredrikekelund, line 51: Would be nice to add a definition file to `wpcom-xhr-request` next, too

cli/lib/tests/api.test.ts
```diff
@@ -33,7 +36,7 @@ describe( 'API Module', () => {
 					post: jest.fn().mockResolvedValue( mockResponse ),
 				},
 			};
-			( wpcomFactory as jest.Mock ).mockReturnValue( mockWpcom );
+			( wpcomFactory as unknown as jest.Mock ).mockReturnValue( mockWpcom );
```
jsnajdr: Would `jest.mocked( wpcomFactory ).mockReturnValue` work here? `jest.mocked` is a Jest helper for convenient typing of mocked functions and objects.

Some time ago I did a [refactoring PR in Calypso](https://github.com/Automattic/wp-calypso/pull/103802) for `jest.mocked`, the same idea could also work here.

fredrikekelund (approved):
Works as expected 👍 I left a couple of minor comments, but nothing blocking.

src/components/auth-provider.tsx
wojtekn: Yes, we need that to continue with a local logout if token removal fails.

src/components/auth-provider.tsx
wojtekn: It seems the type was added intentionally in https://github.com/Automattic/studio/commit/1f8cb5983ec9fee912cb45aafdaa8331d09d330e, so I would leave it.

src/components/auth-provider.tsx
wojtekn: Same as above.

src/components/auth-provider.tsx
wojtekn: Thanks, removed.

src/custom-package-definitions.d.ts
wojtekn: Yes, I may work on that at some point, too.

cli/lib/tests/api.test.ts
wojtekn: Makes sense, I've changed the code to use a helper. I added casting for the param to avoid having to mock the whole object, though.

src/components/auth-provider.tsx
fredrikekelund: Right. We could still move this entire block (the `if ( ! isOffline && client ) {` statement) outside the surrounding try..catch. This is all just to reduce indentation and make the code easier to read, though, so no biggie.

wojtekn: Makes sense, done in 31e25f25a961fc531599c332d91a9902a37c87f7

PR #1949, opened 2025-10-28
Download Python via PowerShell

twstokes:
## Related issues

- Closes AINFRA-1449

## Proposed Changes

- Downloads Python directly from our CDN to work around Chocolatey rate limiting until the dependency is baked into the AMI

## Testing Instructions

1. Confirm that CI is green
2. Smoke test the Studio artifacts on WIndows

## Pre-merge Checklist

### DIFF
--- a/.buildkite/commands/build-for-windows.ps1
+++ b/.buildkite/commands/build-for-windows.ps1
@@ -28,8 +28,33 @@ if ($Architecture -notin $VALID_ARCHITECTURES) {
     Exit 1
 }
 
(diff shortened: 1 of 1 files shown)


twstokes:
I ran the Squirrel setup executable on Windows and smoke tests passed.

twstokes:
It may be interesting to see the results of https://github.com/Automattic/studio/pull/1954 before merging this, though, just in case we find Python isn't needed. I also would lean on the Studio team to know if the release builds might require Python, since only dev builds will be done in this PR.

Update: That PR passed, so we'll likely cancel this one in favor of it.

PR #1950, opened 2025-10-28
Preview Sites: Delete all preview sites using `/delete/all` API endpoint

gavande1:
## Related issues

- Fixes STU-851

## Proposed Changes

- Add a new mutation endpoint for deleting all preview sites for the user 
- Use the API endpoint to delete preview sites in the preference modal

## Testing Instructions

1. Apply the patch 194679-ghe-Automattic/wpcom on your sandbox.
2. Run Studio using this branch.
3. Start the application with:

   ```bash
   npm start
   ```
4. Open Developer Tools and keep the Network tab open.
5. Navigate to the Preferences tab in Studio and log in (if not already).
6. Go to the Usage tab and click Delete all preview sites.
7. Verify that a request to

   ```
   /jurassic-ninja/delete/all/delete/all
   ```

   was sent and returned a successful response
8. Confirm that the usage numbers are reset.
9. Close and reopen the Usage tab.
10. Ensure the updated usage numbers are still reflected.
11. Open the following file on your system:

    ```
    ~/Library/Application Support/Studio/appdata-v1.json
    ```
12. Verify that the snapshots array is cleared and shows as `[]`.

## Pre-merge Checklist

### DIFF
--- a/src/stores/utils/with-offline-check.ts
+++ b/src/stores/utils/with-offline-check.ts
@@ -23,7 +23,9 @@ export function withOfflineCheckMutation< TResult, TArg, TBaseQuery extends Base
 
 		const wrappedTrigger = ( ( ...args: Parameters< typeof trigger > ) => {
 			if ( isOffline ) {
-				return Promise.reject( new Error( 'Cannot perform mutation while offline' ) ) as ReturnType< typeof trigger >;
+				return Promise.reject( new Error( 'Cannot perform mutation while offline' ) ) as ReturnType<
+					typeof trigger
+				>;
 			}
 			return trigger( ...args );
 		} ) as typeof trigger;
--- a/src/stores/utils/with-offline-check.ts
+++ b/src/stores/utils/with-offline-check.ts
@@ -1,6 +1,6 @@
 import { useOffline } from 'src/hooks/use-offline';
 import type { BaseQueryFn } from '@reduxjs/toolkit/query';
-import type { TypedUseQuery } from '@reduxjs/toolkit/query/react';
+import type { TypedUseQuery, TypedUseMutation } from '@reduxjs/toolkit/query/react';
 
 export function withOfflineCheck< TResult, TArg, TBaseQuery extends BaseQueryFn >(
 	useQueryHook: TypedUseQuery< TResult, TArg, TBaseQuery >
@@ -11,5 +11,23 @@ export function withOfflineCheck< TResult, TArg, TBaseQuery extends BaseQueryFn 
 			...options,
 			skip: isOffline || options?.skip,
 		} );
+	};
+}
+
+export function withOfflineCheckMutation< TResult, TArg, TBaseQuery extends BaseQueryFn >(
+	useMutationHook: TypedUseMutation< TResult, TArg, TBaseQuery >
+): TypedUseMutation< TResult, TArg, TBaseQuery > {
+	return ( options = {} ) => {
+		const isOffline = useOffline();
+		const [ trigger, result ] = useMutationHook( options );
+
+		const wrappedTrigger = ( ( ...args: Parameters< typeof trigger > ) => {
+			if ( isOffline ) {
+				return Promise.reject( new Error( 'Cannot perform mutation while offline' ) ) as ReturnType< typeof trigger >;
+			}
+			return trigger( ...args );
+		} ) as typeof trigger;
+
+		return [ wrappedTrigger, result ] as const;
 	};
 }
--- a/src/modules/user-settings/components/snapshot-info.tsx
+++ b/src/modules/user-settings/components/snapshot-info.tsx
@@ -25,7 +25,7 @@ export const SnapshotInfo = ( {
 	const { __ } = useI18n();
 	const { data: snapshotUsage } = useGetSnapshotUsage( undefined, {
 		refetchOnMountOrArgChange: true,
-	});
+	} );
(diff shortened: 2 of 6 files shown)


wojtekn:
I was wondering if it will be clear to the user that it removes all preview sites on the account, and not only those created from the given Studio instance, but it seems that the dialog already makes it clear:

> All preview sites that exist for your WordPress.com account, along with all posts, pages, comments, and media, will be lost.

src/stores/utils/with-offline-check.ts
```diff
@@ -13,3 +13,23 @@ export function withOfflineCheck< TResult, TArg, TBaseQuery extends BaseQueryFn
 		} );
 	};
 }
+
+export function withOfflineCheckMutation< TResult, TArg, TBaseQuery extends BaseQueryFn >(
```
katinthehatsite, line 17: We already disable the button to delete preview sites when the user isOffline, so I am wondering if we need this? 

Image

gavande1: That’s a great point. I agree it’s not needed right now, but I added it just in case we need the offline check elsewhere while using another API. No strong opinion here, though I would slightly prefer keeping it since `withOfflineCheckMutation` could be useful later.

gavande1:
> I was wondering if it will be clear to the user that it removes all preview sites on the account, and not only those created from the given Studio instance, but it seems that the dialog already makes it clear:

This is what I thought as well. That's why I didn't alter the message.

ivan-ottinger (commented):
The changes look good and work correctly. The new endpoint is being used:

Screen Shot on 2025-10-29 at 11:02:48

and the logic clears up the `snapshots` in the `appdata-v1.json` as well:
Screen Shot on 2025-10-29 at 11:04:47

One thing I noticed is that the `Delete all preview sites` button does not get disabled when there are no preview sites left:

Screen Shot on 2025-10-29 at 11:02:01

gavande1:
> One thing I noticed is that the Delete all preview sites button does not get disabled when there are no preview sites left:

@ivan-ottinger, Great question. This is the intended behavior for cases where two separate instances of Studio are out of sync. While disabling the button might seem reasonable, there are several scenarios where keeping it enabled is actually helpful.

ivan-ottinger:
> @ivan-ottinger, Great question. This is the intended behavior for cases where two separate instances of Studio are out of sync. While disabling the button might seem reasonable, there are several scenarios where keeping it enabled is actually helpful.

I see. Are you referring to a case where the `snapshots`  are still recorded on some other app instance? Is the intention to clear that leftover `snapshosts` array with the button?

wojtekn:
> I see. Are you referring to a case where the snapshots are still recorded on some other app instance? Is the intention to clear that leftover snapshosts array with the button?

This is the goal, with the new implementation, the button should always be available to let the user clean preview sites from their account.

ivan-ottinger:
> This is the goal, with the new implementation, the button should always be available to let the user clean preview sites from their account.

Makes sense. 👍🏼

ivan-ottinger (approved):

katinthehatsite:
> This is the goal, with the new implementation, the button should always be available to let the user clean preview sites from their account.

I am finding this a bit of a strange UI but I guess there are not many ways around it.

src/stores/utils/with-offline-check.ts
katinthehatsite: Okay, we can keep it then 👍 If it is not something that we use often, we can clean this up later.

katinthehatsite (approved):
👍

ivan-ottinger:
> I am finding this a bit of a strange UI but I guess there are not many ways around it.

Yeah. likewise. I thought about it yesterday and maybe I am missing something, but:

Cannot we do the following?
- disable the `Delete all preview sites` button when there are no preview site linked to the user's account
- let the `Clear` button on the Previews tab remove those already-deleted sites from `appdata-v1.json` on another Studio instance

The `Delete all preview sites` button would:
- trigger the endpoint to clear all preview sites on the server
- clear preview sites from `appdata-v1.json`  on the local Studio instance

The `Clear` button on the Previews tab would only clear the leftover preview sites from the `appdata-v1.json` .

gavande1:
> Cannot we do the following?
> - disable the Delete all preview sites button when there are no preview site linked to the user's account
> - let the Clear button on the Previews tab remove those already-deleted sites from appdata-v1.json on another Studio instance

Since we are already [querying usage](https://github.com/Automattic/studio/blob/trunk/src/modules/user-settings/components/snapshot-info.tsx#L26) on the component. I think it makes sense to do a quick check on the user's account. I will update the code to disable the button when there are no preview sites on the user's account. cc @wojtekn 

The remaining suggested behaviours have already been implemented in #1950.

gavande1:
@ivan-ottinger Thanks for the feedback so far. I have addressed your concerns in [e45a445](https://github.com/Automattic/studio/pull/1950/commits/e45a445ae217073ad5eeceb5d20b533613366456). Could you review it again?

ivan-ottinger:
> @ivan-ottinger Thanks for the feedback so far. I have addressed your concerns in [e45a445](https://github.com/Automattic/studio/pull/1950/commits/e45a445ae217073ad5eeceb5d20b533613366456). Could you review it again?

Nice! Thank you for the updates. The changes look good and I can see that if there are no remote preview sites, the button gets disabled:

Screen Shot on 2025-10-30 at 12:55:58

PR #1952, opened 2025-10-28
Studio: Optimize import workflow by removing unnecessary server start/stop

bcotrim:
## Related issues

- Fixes STU-744

## Proposed Changes

- Added `setupWordPressFilesOnly()` method to PlaygroundCliProvider that copies WordPress files directly without starting the serve
- Modified import workflow in ipc-handlers.ts to use the new method instead of the start/stop cycle
- Added blueprint clearing after import to prevent imported data from being overwritten on first site start

## Testing Instructions

- Create a new site in Studio using a import file
- Import a backup site into an existing Studio site

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -172,9 +172,7 @@ export async function importSite(
 	}
 	try {
 		if ( ! isWordPressDirectory( site.details.path ) ) {
-			// Workaround to have the necessary WordPress files to run the import - STU-744
-			await site.start();
-			await site.stop();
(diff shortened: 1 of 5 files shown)


katinthehatsite (approved):
The PR worked as described. I did not encounter any further issues 👍

epeicher (approved):
Thanks @bcotrim for improving this. I have tested it, and it works as described. I have been able to create a site from an import file and import a file into an existing site without issues. LGTM! :shipit:

PR #1955, opened 2025-10-29
Preview Sites: Add preview site status check when visiting preview tab

gavande1:
## Related issues

- Fixes STU-209

## Proposed Changes

- Add a check for getting the preview site status by using the API endpoint
- Add RTK query for getting preview site status

## Testing Instructions

- Check out 194679-ghe-Automattic/wpcom on your sandbox
- Check out this branch on Studio
- Run Studio using `npm start`
- Create new preview site if one doesn't exists
- Go to https://developer.wordpress.com/docs/api/console/
- Select `WP REST API` -> `wpcom/v2`
- Search for `/jurassic-ninja/delete/all` route and select it. 
- Send the request and wait for a couple of seconds.
- Come back to Studio, navigate between the preview tab & other tabs.
- Confirm the preview site is automatically marked as deleted. 

CleanShot 2025-10-29 at 14 45 47@2x

## Pre-merge Checklist

### DIFF
--- a/src/modules/preview-site/components/preview-site-row.tsx
+++ b/src/modules/preview-site/components/preview-site-row.tsx
@@ -16,6 +16,7 @@ import { DeleteProgressRow } from 'src/modules/preview-site/components/delete-pr
 import { PreviewActionButtonsMenu } from 'src/modules/preview-site/components/preview-action-buttons-menu';
 import { useAppDispatch, useRootSelector } from 'src/stores';
 import { snapshotActions, snapshotSelectors } from 'src/stores/snapshot-slice';
+import { useGetSnapshotStatus } from 'src/stores/wpcom-api';
 
 interface PreviewSiteRowProps {
 	snapshot: Snapshot;
@@ -42,9 +43,12 @@ export function PreviewSiteRow( {
 	const deleteOperation = useRootSelector( ( state ) =>
 		snapshotSelectors.selectDeleteOperationForSnapshot( state, snapshot.url )
 	);
+	const { data: snapshotStatus, refetch: refetchSnapshotStatus } = useGetSnapshotStatus( snapshot.atomicSiteId );
 	const { formatRelativeTime } = useFormatLocalizedTimestamps();
 	const [ showUpdatedMessage, setShowUpdatedMessage ] = useState( false );
 	const wasUpdating = useRef( false );
+	const isDeleted = snapshotStatus?.isDeleted;
+	const isSiteInactive = isExpired || isDeleted;
 
 	useEffect( () => {
 		if ( ! updateOperation ) {
@@ -72,6 +76,11 @@ export function PreviewSiteRow( {
 
 		return () => clearTimeout( timeoutId );
 	}, [ updateOperation ] );
+
+	// Refetch snapshot status to check if the site has been deleted
+	useEffect( () => {
+		refetchSnapshotStatus();
+	}, [ refetchSnapshotStatus ] );
 
 	const getLastUpdateTimeText = () => {
 		if ( ! date ) {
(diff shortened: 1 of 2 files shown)


src/modules/preview-site/components/preview-site-row.tsx
```diff
@@ -108,31 +119,41 @@ export function PreviewSiteRow( {
...
-					</div>
-					<Button
-						variant="link"
-						disabled={ isExpired }
-						className={ cx(
-							'!text-a8c-gray-700 max-w-full',
-							isExpired ? 'pointer-events-none' : 'hover:!text-a8c-blue-50'
+					<Tooltip
+						text={ __(
+							'This preview site has been marked as deleted. You can no longer access it. You can clear it from you list by clicking the Clear button.'
```
gavande1: I need suggestions for the message. I feel it's not very good.

src/modules/preview-site/components/preview-site-row.tsx
```diff
@@ -108,31 +114,41 @@ export function PreviewSiteRow( {
...
 						) }
-						onClick={ () => getIpcApi().openURL( `https://${ url }` ) }
+						disabled={ ! isDeleted }
 					>
-						<span className={ cx( 'truncate', isExpired && 'line-through text-a8c-gray-700' ) }>
-							{ url }
-						</span>
-						{ ! isExpired && <ArrowIcon /> }
-					</Button>
+						<div className="flex flex-col">
```
fredrikekelund: We could make this a bit more compact like so:

```tsx
				<Tooltip
					className="grow overflow-hidden pe-4"
					text={ __(
						'This preview site has been marked as deleted. You can no longer access it. You can clear it from you list by clicking the Clear button.'
					) }
					disabled={ ! isDeleted }
				>
```

`w-[51%]` seems a bit random. Replacing it with `grow` seems reasonable.

src/modules/preview-site/components/preview-site-row.tsx
```diff
@@ -108,31 +114,41 @@ export function PreviewSiteRow( {
...
-					</div>
-					<Button
-						variant="link"
-						disabled={ isExpired }
-						className={ cx(
-							'!text-a8c-gray-700 max-w-full',
-							isExpired ? 'pointer-events-none' : 'hover:!text-a8c-blue-50'
+					<Tooltip
+						text={ __(
+							'This preview site has been marked as deleted. You can no longer access it. You can clear it from you list by clicking the Clear button.'
```
fredrikekelund:
```suggestion
							'This preview site has been deleted on the server. Remove it from the list by clicking the Clear button.'
```
Just a suggestion. Making it briefer seems like an improvement.

src/stores/wpcom-api.ts
```diff
@@ -152,6 +168,14 @@ export const wpcomApi = createApi( {
 			keepUnusedDataFor: 60 * 60,
 			providesTags: [ 'SnapshotUsage' ],
 		} ),
+		getSnapshotStatus: builder.query< LiveSnapshotStatus, number >( {
```
fredrikekelund:
```suggestion
		getSnapshotStatus: builder.query< z.infer< typeof snapshotStatusSchema >, number >( {
```
Sorry for the nitpick, but we might as well use `z.infer` here, since we don't reuse `LiveSnapshotStatus` anywhere else.

fredrikekelund (approved):
Nice and neat change that works as expected 👍 I left a couple of suggestions that would be nice to address before merging, but there are no blockers.

fredrikekelund:
One question struck me: what would `useGetSnapshotStatus` return when a preview site has expired? Will all expired sites be marked as deleted from now on..?

src/stores/wpcom-api.ts
gavande1: Thanks for the suggestion. I have updated it.

src/modules/preview-site/components/preview-site-row.tsx
gavande1: Makes sense. I have updated the code and position of the tooltip.

src/modules/preview-site/components/preview-site-row.tsx
gavande1: Looks good to me. I have updated the message.

gavande1:
> One question struck me: what would `useGetSnapshotStatus` return when a preview site has expired? Will all expired sites be marked as deleted from now on..?

Great questions. `useGetSnapshotStatus` would return `is_deleted=true` when the site has expired. I have modified conditions to keep the existing behaviour of the expired site.

src/modules/preview-site/components/preview-site-row.tsx
```diff
@@ -142,18 +159,18 @@ export function PreviewSiteRow( {
...
-							<Tooltip text={ dateString } disabled={ ! date }>
+							<Tooltip text={ dateString } disabled={ ! date || isSiteInactive }>
 								{ getLastUpdateTimeText() }
 							</Tooltip>
 						) }
 					</div>
 					<div className="flex items-center">
-						<Tooltip text={ expireDateString } disabled={ isExpired }>
+						<Tooltip text={ expireDateString } disabled={ isSiteInactive }>
 							<div className="w-[150px] text-a8c-gray-700 pl-4">{ countDown }</div>
```
fredrikekelund: Can we render `Deleted` on this row if the site has been deleted?

fredrikekelund (approved):
Works as expected 👍 Thanks for those additional changes, @gavande1. I left one final suggestion that I think we should address before landing. Approving now to unblock you, though.

src/modules/preview-site/components/preview-site-row.tsx
gavande1: It makes sense to me. I have included it; here is the final look. 

Image

src/modules/preview-site/components/preview-site-row.tsx
```diff
@@ -107,32 +113,43 @@ export function PreviewSiteRow( {
...
-						variant="link"
-						disabled={ isExpired }
-						className={ cx(
-							'!text-a8c-gray-700 max-w-full',
-							isExpired ? 'pointer-events-none' : 'hover:!text-a8c-blue-50'
+				<div className="overflow-hidden pe-4">
+					<Tooltip
+						placement="top-start"
+						text={ __(
+							'This preview site has been deleted on the server. Remove it from the list by clicking the Clear button.'
```
katinthehatsite:
```suggestion
							'This preview site has been deleted from the server. You can remove it from the list by clicking Clear button.'
```

src/modules/preview-site/components/preview-site-row.tsx
katinthehatsite: Added a suggestion but mostly to tweak the grammar a bit 👍

katinthehatsite (approved):
Image

The functionality looks good to me, I did not spot any regressions. Thanks for improving user experience 👍 

I just left a minor suggestion for the message copy.

src/modules/preview-site/components/preview-site-row.tsx
gavande1: Thanks, @katinthehatsite, for the suggestion. It looks better. I have added in [c6e029e](https://github.com/Automattic/studio/pull/1955/commits/c6e029e024678e75bfc8c1c75ee25f17ee9522aa).

PR #1966, opened 2025-10-29
Sync: Fix right padding on Windows for select files dropdown

epeicher:
## Related issues

- Fixes STU-843

## Proposed Changes

- Applies a patch that removes only for Windows the [`field-sizing: content` ](https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing) CSS from the SelectControl component as it does not work well on Windows

|Windows|Before | After|
|-------|-------|------|
| Collapsed | CleanShot 2025-10-30 at 12 51 27@2x| CleanShot 2025-10-30 at 13 00 37@2x| 
| Expanded| CleanShot 2025-10-30 at 12 51 32@2x| CleanShot 2025-10-30 at 13 00 42@2x|

## Testing Instructions

- On Windows
- Apply this branch and run `npm install` and `npm start`
- On Studio, go to Sync tab and connect a site 
- Click on Push or Pull and on the modal, expand the `All files and folders` dropdown
- Check that there is some padding on the right of both options
- Repeat the steps on Mac
- Check theres is no changes with production

> [!NOTE]
> I have tried to also provide some extra padding to the _bottom_ of the `Specific files and folders` option, but I have not found a simple way to do that (I have discarded using CSS selectors), and the impact is not as big as the right padding that this issue fixes 

## Pre-merge Checklist


bcotrim (approved):
LGTM 👍

katinthehatsite (approved):
The changes look good 👍 

One question that I have is that when I was reviewing the issue, I noticed https://github.com/WordPress/gutenberg/issues/72424 - do you think we should create a patch to the component instead and apply it in Studio so that we can fix all instances of the component usage? As a sidenote, I have not researched how much we are using this component, just asking this as a general question.

epeicher:
> One question that I have is that when I was reviewing the issue, I noticed https://github.com/WordPress/gutenberg/issues/72424 - do you think we should create a patch to the component instead and apply it in Studio so that we can fix all instances of the component usage? As a sidenote, I have not researched how much we are using this component, just asking this as a general question.

That's a good point @katinthehatsite, and I think is what @wojtekn suggests in the linked task STU-843, that way, if there is a fix in the Gutenberg library, we just need to remove the patch and it will be automatically applied. 

@wojtekn, do you think I should try to patch the component instead of the current fix that uses the `isWindows()` flag? I didn't consider the patch option when working on this yesterday.

wojtekn:
@epeicher, if it's getting fixed in Gutenberg, an NPM patch may be a good way. With that approach, we won't miss the change in mainstream as our patch will stop applying and will need to be removed.

epeicher:
Hi @katinthehatsite, @wojtekn, I have applied the patch to the `@wordpress/components` npm package as part of [d92d1a5](https://github.com/Automattic/studio/pull/1966/commits/d92d1a52215581859b675e940ac5fbbebf805103) but that adds some extra padding to the arrow on the collapsed component on **both** Mac and Windows, please see the screenshots on the Description of the PR. 

Could you please review them and let me know your view? I would be happy to revert to the previous version if we don't think the extra padding should be there on Mac.

CleanShot 2025-10-30 at 12 32 22@2x

katinthehatsite:
> Could you please review them and let me know your view? I would be happy to revert to the previous version if we don't think the extra padding should be there on Mac.

Hmm, if I were to be honest, I think it looked better before. Although I prefer the patch approach, I think user experience should be a priority and that gap is a bit 🫠

epeicher:
> Hmm, if I were to be honest, I think it looked better before. Although I prefer the patch approach, I think user experience should be a priority and that gap is a bit 🫠

@katinthehatsite, I totally agree. I managed to cover both requirements using a patch applied only on the affected platform, so it doesn’t impact the user experience elsewhere. The change is here [2d69ae5](https://github.com/Automattic/studio/pull/1966/commits/2d69ae52c5f8fea88da83bacede0063e9f6c205f). I have updated the testing instructions and description to reflect this.

patches/@wordpress+components+30.6.0.patch
```diff
@@ -41,3 +41,45 @@ index 24af829..334bb6f 100644
...
...HQmJiYge1xuXHRcdGFwcGVhcmFuY2U6IG5vbmU7XG5cdFx0YmFja2dyb3VuZDogdHJhbnNwYXJlbnQ7XG5cdFx0Ym94LXNpemluZzogYm9yZGVyLWJveDtcblx0XHRib3JkZXI6IG5vbmU7XG5cdFx0Ym94LXNoYWRvdzogbm9uZSAhaW1wb3J0YW50O1xuXHRcdGNvbG9yOiBjdXJyZW50Q29sb3I7IC8vIE92ZXJyaWRlcyBob3Zlci9mb2N1cyBzdHlsZXMgaW4gZm9ybXMuY3NzXG5cdFx0Y3Vyc29yOiBpbmhlcml0O1xuXHRcdGRpc3BsYXk6IGJsb2NrO1xuXHRcdGZvbnQtZmFtaWx5OiBpbmhlcml0O1xuXHRcdG1hcmdpbjogMDtcblx0XHR3aWR0aDogMTAwJTtcblx0XHRtYXgtd2lkdGg6IG5vbmU7XG5cdFx0d2hpdGUtc3BhY2U6IG5vd3JhcDtcblx0XHR0ZXh0LW92ZXJmbG93OiBlbGxpcHNpcztcblxuXHRcdCR7IGZvbnRTaXplU3R5bGVzIH07XG5cdFx0JHsgc2l6ZVN0eWxlcyB9O1xuXHRcdCR7IHNpemVQYWRkaW5ncyB9O1xuXHRcdCR7IG92ZXJmbG93U3R5bGVzIH1cblx0XHQkeyB2YXJpYW50U3R5bGVzIH1cblx0fVxuYDtcblxuZXhwb3J0IGNvbnN0IERvd25BcnJvd1dyYXBwZXIgPSBzdHlsZWQuZGl2YFxuXHRtYXJnaW4taW5saW5lLWVuZDogJHsgc3BhY2UoIC0xICkgfTsgLy8gb3B0aWNhbGx5IGFkanVzdCB0aGUgaWNvblxuXHRsaW5lLWhlaWdodDogMDtcblxuXHRwYXRoIHtcblx0XHRmaWxsOiBjdXJyZW50Q29sb3I7XG5cdH1cbmA7XG5cbmV4cG9ydCBjb25zdCBJbnB1dENvbnRyb2xTdWZmaXhXcmFwcGVyV2l0aENsaWNrVGhyb3VnaCA9IHN0eWxlZChcblx0SW5wdXRDb250cm9sU3VmZml4V3JhcHBlclxuKWBcblx0cG9zaXRpb246IGFic29sdXRlO1xuXHRwb2ludGVyLWV2ZW50czogbm9uZTtcblxuXHQkeyBydGwoIHsgcmlnaHQ6IDAgfSApIH1cbmA7XG4iXX0= */",
```
wojtekn, line 62: Any reason to add a source map with this patch?

epeicher: That's done automatically by `patch-package` utility when generating the patch, I have tried to remove it manually but the patch fails to be applied in that case

PR #1972, opened 2025-10-30
Execute `/studio-app/sync/import` network request from node process

fredrikekelund:
## Related issues

- Related to STU-942

## Proposed Changes

Previously, the `/sites/${ remoteSiteId }/studio-app/sync/import` API request was made from the renderer process. This required the entire site archive to be stored in the renderer process's memory. This isn't necessary, and it's much more efficient (especially for large archives) to stream the file contents from disk instead. This PR implements a `pushArchive` IPC handler that does just that. 

## Testing Instructions

1. Have a Studio site connected to a remote site
2. Try the push functionality with a few different configurations (all files/some files/with db/without) and verify that it works as expected

## Pre-merge Checklist


src/ipc-handlers.ts
```diff
@@ -778,22 +781,62 @@ export async function exportSiteToPush(
...
+	const rawResponse = await wpcom.req.post( {
+		path: `/sites/${ remoteSiteId }/studio-app/sync/import`,
+		apiNamespace: 'wpcom/v2',
+		formData,
+	} );
+
+	return pushArchiveResponseSchema.parse( rawResponse );
+}
+
+export function removeTemporaryFile( event: IpcMainInvokeEvent, path: string ) {
```
fredrikekelund, line 837: Renaming this function is a secondary function of this PR. I did it because "temporal" seems like a misnomer in this context.

src/ipc-handlers.ts
```diff
@@ -712,7 +715,7 @@ export async function archiveSite( event: IpcMainInvokeEvent, id: string, format
 	return { archivePath, archiveSizeInBytes: stats.size };
 }
 
-export async function exportSiteToPush(
+export async function exportSiteForPush(
```
fredrikekelund, line 718: Renaming this function is also a secondary function of this PR. I did it because I think this clarifies the relationship better between this function and `pushArchive`.

fredrikekelund:
I ran a quick test with a 767MB archive to compare this branch with the latest stable release. The results show significant promise:

| Time | v1.6.2 stable | this branch |
| - | - | - |
| Total | 14:42 | 10:45 |
| Upload | 5min | 39s  |

src/ipc-handlers.ts
```diff
@@ -778,22 +781,68 @@ export async function exportSiteToPush(
...
+
+		return { success: true };
+	} catch ( error ) {
+		const parseResult = z.object( { error: z.string() } ).safeParse( error );
+
+		if ( parseResult.success ) {
+			return { success: false, error: parseResult.data.error };
+		}
+
+		return { success: false, error: 'Unknown error' };
```
fredrikekelund, line 833: IPC functions that throw in the node process will also throw in the renderer process. The crux is that the error object thrown by the renderer has a different structure, so the most reliable way to pass data back to the renderer is to simply return.

src/ipc-handlers.ts
```diff
@@ -778,22 +781,68 @@ export async function exportSiteToPush(
...
+			},
+		],
+	];
+
+	if ( optionsToSync ) {
+		formData.push( [ 'options', optionsToSync.join( ',' ) ] );
+	}
+
+	try {
+		await wpcom.req.post( {
```
epeicher, line 819: Is the response of this either a success or an exception in all cases? I was considering getting the response and check the response.success value from it

epeicher (approved):
Thanks @fredrikekelund for improving this! I have tested the Push with different scenarios using both partial sync and full sync, and they all have worked as expected. LGTM! 

I have just left a non-blocking question out of curiosity

src/ipc-handlers.ts
fredrikekelund: `wpcom.req` throws if the request fails or if the response HTTP status code isn't in the 200-range. Since we also don't care about the response body, I don't think we gain anything by validating it.

fredrikekelund: The `wpcom.js` and `wpcom-xhr-request` API is a little arcane, but the network requests are ultimately based on superagent, so I looked at the [source code](https://github.com/forwardemail/superagent) to double-check my assumption here.

epeicher: > wpcom.req throws if the request fails or if the response HTTP status code isn't in the 200-range. Since we also don't care about the response body, I don't think we gain anything by validating it.

Agreed, thanks for clarifying!

PR #1980, opened 2025-10-30
Sync: Remove unnecesary ref when initializing Push states

epeicher:
## Related issues

- Related to https://github.com/Automattic/studio/pull/1943

## Proposed Changes

- Removes the ref that prevented the `useEffect` to run multiple times as it is not necessary anymore after https://github.com/Automattic/studio/pull/1943/commits/be16976f40b9faaad6c8b3efeb26778b88e858f9
- Removes the `void` operator as the promise is handled by the `.catch`

## Testing Instructions

- Apply these changes and run npm start
- Navigate to the Sync tab and connect a site or use any existing connected site
- Select Push
- Wait for the status to be progressed to Backing up remote site…
- Select another site before leaving Studio
- Press Cmd+Q to quit Studio
- Check that the dialog appears and notify that the operation will continue. Accept and quit Studio
- Open Studio again
- Select the site that has a Push in progress
- Navigate to the Sync tab
- Check the progress continues
- Check in the network tab the `import` endpoint is checked only once for the connected sites initially, and it is not executed when selecting the Sync tab of different sites every time. Please note that if there is an operation in progress, the `import` endpoint will be checked periodically.
 
CleanShot 2025-10-30 at 18 48 01@2x

## Pre-merge Checklist

### DIFF
--- a/src/hooks/sync-sites/sync-sites-context.tsx
+++ b/src/hooks/sync-sites/sync-sites-context.tsx
@@ -110,15 +110,12 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
 
 	const { client } = useAuth();
 	const { pushStatesProgressInfo } = useSyncStatesProgressInfo();
-	const hasInitialized = useRef( false );
 
 	// Initialize push states from in-progress server operations on mount
 	useEffect( () => {
-		if ( hasInitialized.current || ! client ) {
+		if ( ! client ) {
 			return;
 		}
-
-		hasInitialized.current = true;
 
 		const initializePushStates = async () => {
 			const allSites = await getIpcApi().getSiteDetails();
@@ -166,7 +163,7 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
 			}
 		};
 
-		void initializePushStates().catch( ( error ) => {
+		initializePushStates().catch( ( error ) => {
 			// Initialization is not critical to app functionality, but log the error
 			console.error( 'Failed to initialize push states from server:', error );
 		} );


src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -110,16 +110,13 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
 
 	const { client } = useAuth();
 	const { pushStatesProgressInfo } = useSyncStatesProgressInfo();
-	const hasInitialized = useRef( false );
```
fredrikekelund, line 113: At first, I was going to say that I think it's a good idea to make our intention explicit, only to run this effect once (i.e., to preserve the ref). However, after thinking about it, I realized that re-running it whenever `client` changes actually makes a lot of sense, since it will rerun when the user authenticates with WordPress.com.

src/hooks/sync-sites/sync-sites-context.tsx
```diff
@@ -166,7 +163,7 @@ export function SyncSitesProvider( { children }: { children: React.ReactNode } )
 			}
 		};
 
-		void initializePushStates().catch( ( error ) => {
+		initializePushStates().catch( ( error ) => {
```
fredrikekelund, line 166: 👍

fredrikekelund (approved):
LGTM 👍 I haven't tested, but I feel confident about the code changes.

src/hooks/sync-sites/sync-sites-context.tsx
epeicher: Yes, indeed, I didn't think of that use case, but I have tested it, and it works perfectly. The progress is displayed after authentication 👍

PR #1981, opened 2025-10-31
Adopt custom Windows image in CI

mokagio:
## Related issues

- https://linear.app/a8c/issue/AINFRA-1471/adopt-custom-windows-ami-in-studio

## Proposed Changes

- Use our new custom Windows image in CI to save some time during builds, avoid the recent out-of-space issue, and unlock future improvements

## Testing Instructions

- Verify CI is green, especially the Windows steps

## Pre-merge Checklist

### DIFF
--- a/.buildkite/shared-pipeline-vars
+++ b/.buildkite/shared-pipeline-vars
@@ -5,7 +5,7 @@
 
 # The ~> modifier is not currently used, but we check for it just in case
 XCODE_VERSION=$(sed -E -n 's/^(~> )?(.*)/xcode-\2/p' .xcode-version)
-CI_TOOLKIT_VERSION='5.5.0'
+CI_TOOLKIT_VERSION='5.7.0'
 NVM_PLUGIN_VERSION='0.6.0'
 
 export IMAGE_ID="$XCODE_VERSION"
--- a/.buildkite/shared-pipeline-vars
+++ b/.buildkite/shared-pipeline-vars
@@ -11,3 +11,4 @@ NVM_PLUGIN_VERSION='0.6.0'
 export IMAGE_ID="$XCODE_VERSION"
 export CI_TOOLKIT_PLUGIN="automattic/a8c-ci-toolkit#$CI_TOOLKIT_VERSION"
 export NVM_PLUGIN="automattic/nvm#$NVM_PLUGIN_VERSION"
+export WINDOWS_QUEUE='windows'
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -29,7 +29,7 @@ steps:
       queue: "{{matrix}}"
     matrix:
       - mac
-      - windows
+      - $WINDOWS_QUEUE
     notify:
       - github_commit_status:
           context: Unit Tests
@@ -114,7 +114,7 @@ steps:
     steps:
       - label: 🔨 Windows Dev Build - {{matrix}}
         agents:
-          queue: windows
+          queue: $WINDOWS_QUEUE
         command: powershell -File .buildkite/commands/build-for-windows.ps1 -BuildType dev -Architecture {{matrix}}
         artifact_paths:
           - out\**\studio-setup.exe
@@ -222,7 +222,7 @@ steps:
     steps:
       - label: 🔨 Windows Release Build - {{matrix}}
         agents:
-          queue: windows
+          queue: $WINDOWS_QUEUE
         command: powershell -File .buildkite/commands/build-for-windows.ps1 -BuildType release -Architecture {{matrix}}
         artifact_paths:
           - out\**\studio-setup.exe


.buildkite/shared-pipeline-vars
```diff
@@ -5,9 +5,10 @@
...
 # The ~> modifier is not currently used, but we check for it just in case
 XCODE_VERSION=$(sed -E -n 's/^(~> )?(.*)/xcode-\2/p' .xcode-version)
-CI_TOOLKIT_VERSION='5.5.0'
+CI_TOOLKIT_VERSION='5.7.0'
 NVM_PLUGIN_VERSION='0.6.0'
 
 export IMAGE_ID="$XCODE_VERSION"
 export CI_TOOLKIT_PLUGIN="automattic/a8c-ci-toolkit#$CI_TOOLKIT_VERSION"
 export NVM_PLUGIN="automattic/nvm#$NVM_PLUGIN_VERSION"
+export WINDOWS_QUEUE='windows'
```
mokagio, line 14: This DRY seemed like a good idea to make it faster and less error-prone (change one but not the other) to update the queue for testing, as I expect we'll iterate on it a few more times in the near future.

wojtekn:
@mokagio, where exactly change to a custom image? Is it part of the toolkit update?

.buildkite/shared-pipeline-vars
AliSoftware: While we might indeed iterate a couple of times in the upcoming weeks while improving the AMI,
 - I don't expect us to merge any change of the queue to `windows-staging` (I'd instead expect we point to `windows-staging` in a draft PR that we'd never merge)
 - Once things are settled and we're happy with a stable and more complete AMI, I don't expect we'll change the queue anymore for a very long time

So I'm not super convinced this is _that_ useful 🤷  (Besides, if we start doing it for the `windows` queue, why not for `mac` as well?)

That being said, it's not critical, and if you feel that's really useful that shouldn't be a blocker. Up to you.

AliSoftware:
> @mokagio, where exactly change to a custom image? Is it part of the toolkit update?

@wojtekn  In practice nothing in this PR is responsible for the change to using the custom AMI.

Instead, the custom AMI has been deployed in our CloudFormation infrastructure via https://github.com/Automattic/buildkite-ci/pull/743, and as soon as that was deployed, any new CI job targeting the `windows` queue started using an EC2 instance based on that custom AMI (verified on [this job for `studio`](https://buildkite.com/automattic/studio/builds/8648/steps/canvas?jid=019a481a-c261-46b5-a47b-05bf2fe919cb))

FYI:
 - Before deploying the custom AMI to our `windows` CloudFormation stack, we deployed it to our `windows-staging` stack/queue to validate that `studio` CI jobs would still work when using that custom AMI. Only then did we deploy the custom AMI to the `windows` queue too via https://github.com/Automattic/buildkite-ci/pull/743
 - In order to make the transition smooth, [we've set an environment variable when provisioning the custom AMI](https://github.com/Automattic/buildkite-ci/blob/b242fc800b203a6af92398bdb38b72acf96853d5/src/agents/windows/windows-build-image.pkr.hcl#L123) to allow us to test during job scripts which version of the AMI the EC2 instance is running. We've then been [using that env var in the latest version of the `ci-toolkit` plugin to skip the installation of the tooling](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/prepare_windows_host_for_app_distribution.ps1#L30-L36) (Windows SDK, MSVC compilers, python) when calling `prepare_windows_host_for_distribution.ps1` if the value of that env var is >= the AMI version in which we started pre-provisioning those tools.
 - In the long run, `studio` will be able to [remove all the calls to `prepare_windows_host_for_distribution.ps1`](https://linear.app/a8c/issue/AINFRA-1465/remove-env-setup-logic-now-in-ami-from-ci-toolkit-once-all-projects) from the commands called by your CI pipelines (you can probably already start doing that), and instead only call `& setup_windows_code_signing.ps1` in pipelines that really require the distribution certificate to be installed (e.g. the ones doing Release Builds and needing release code signing), while other CI jobs would not need to call any tooling install script from `ci-toolkit` anymore. But in the meantime, thanks to [our change in `prepare_windows_host_for_distribution.ps1`](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/prepare_windows_host_for_app_distribution.ps1#L30-L36) this is already what gets ultimately called anyway (just more indirectly), making the transition to the final setup smoother.
 - In the meantime, we plan to iterate on the custom `windows` AMI to add more and more tools pre-bundled into it (e.g. Yarn, Node, Rust, Go, …) to minimize even more the tooling that some other projects have been needing and installing ad-hoc up until now

AliSoftware (approved):
Not convinced by the need for `WINDOWS_QUEUE=windows`, but that's not a blocker either.

And we need to migrate to the latest `ci-toolkit` sooner than later so that the install of [the custom tooling gets skipped when it detects we're using the custom AMI >=v0.2](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/prepare_windows_host_for_app_distribution.ps1#L30-L36), so approving to unblock.

wojtekn (approved):

wojtekn:
Thanks for the detailed explanation, @AliSoftware . I've opened https://github.com/Automattic/studio/pull/1996 to start applying those changes.

PR #1982, opened 2025-10-31
Studio: Fix errors on cancelSyncOperation

katinthehatsite:
## Related issues

N/A 

I was working on unrelated PR and noticed this error in `preload.ts`:

Screenshot 2025-10-31 at 10 10 03 AM

## Proposed Changes

This PR fixes the error on the screenshot above. 

## Testing Instructions

* Navigate to `preload.ts`
* Confirm that the error is no longer present

## Pre-merge Checklist


src/preload.ts
```diff
@@ -106,7 +106,7 @@ const api: IpcApi = {
 		ipcRendererInvoke( 'getConnectedWpcomSites', localSiteId ),
 	addSyncOperation: ( id, status ) => ipcRendererSend( 'addSyncOperation', id, status ),
 	clearSyncOperation: ( id ) => ipcRendererSend( 'clearSyncOperation', id ),
-	cancelSyncOperation: ( id ) => ipcRendererSend( 'cancelSyncOperation', id ),
+	cancelSyncOperation: ( id ) => ipcRendererInvoke( 'cancelSyncOperation', id ),
```
katinthehatsite: I see that we await for this function when we use it so I switched it to `ipcRendererInvoke`.

epeicher (commented):
Thanks for improving this @katinthehatsite! I was looking at the types, and something was not right, as the `cancelSyncOperation` does not return a Promise, so it does not need to be awaited. But the linting was complaining about that function not being awaited. Eventually, I have found that the `cancelSyncOperation` must be added to [IPC_VOID_HANDLERS](https://github.com/Automattic/studio/blob/3853f51ca74f46c8c6d679cbb41634c0c43c21de/src/constants.ts#L84), and the await can be removed. That way all the types make sense 😇

katinthehatsite:
> Thanks for improving this @katinthehatsite! I was looking at the types, and something was not right, as the cancelSyncOperation does not return a Promise, so it does not need to be awaited. But the linting was complaining about that function not being awaited. Eventually, I have found that the cancelSyncOperation must be added to [IPC_VOID_HANDLERS](https://github.com/Automattic/studio/blob/3853f51ca74f46c8c6d679cbb41634c0c43c21de/src/constants.ts#L84), and the await can be removed. That way all the types make sense 😇

I did that as well initially 😅  but then I started wondering if it makes sense to await. If we don't, then this could lead to race conditions where the code updates the UI state before the cancellation is fully processed in the main process. Perhaps I am overthinking this though because all the actions seem to be synchronous there 🫠 

What do you think?

epeicher:
> I did that as well initially 😅 but then I started wondering if it makes sense to await. If we don't, then this could lead to race conditions where the code updates the UI state before the cancellation is fully processed in the main process. Perhaps I am overthinking this though because all the actions seem to be synchronous there 🫠

What do you think?

~~We cannot await there as the [original function does not return a promise](https://github.com/Automattic/studio/blob/cb4422de7e9d7645d6e6e9b1eb454b4c5010fa2d/src/ipc-handlers.ts#L1405), it just returns undefined, so `await` does not have any effect there, this is the warning if we add the function to the IPC_VOID_HANDLERS constant
CleanShot 2025-10-31 at 16 17 57@2x so I would update the types 😄 Alternatively, we could add a return value to the `cancelSyncOperation` if we want to wait for the call 🤔~~

EDIT: I have edited the comment above as it is not correct, if we change to `ipcRendererInvoke` a Promise is returned by [invoke](https://www.electronjs.org/docs/latest/api/ipc-renderer#ipcrendererinvokechannel-args) and we can await for it (sorry for the confusion 🙏 ). 

I have tested it, and I can see an error in the UI when canceling the operation. It seems the `exportSiteToPush` throws an error when aborting the operation that is now captured.

katinthehatsite:
> I have tested it, and I can see an error in the UI when canceling the operation. It seems the exportSiteToPush throws an error when aborting the operation that is now captured.

Huh, interesting! I see the same thing. Okay, let's switch to the approach with `IPC_VOID_HANDLERS` and skip awaiting.

PR #1983, opened 2025-10-31
Add platform-specific terminal app checking

ivan-ottinger:
## Related issues

Related to STU-931

## Proposed Changes

- ensure that only the terminal apps available on specific OS are listed
  - this removes iTerm and Ghostty from the Studio app on Windows 

| Before | After |
|--------|--------|
| Screen Shot on 2025-10-31 at 16:47:25 | Screen Shot on 2025-10-31 at 16:46:35 | 

~~On Monday I am planning to propose adding PowerShell as another terminal option on Windows in another PR.~~

→ Update: After further consideration I don't think it makes sense to add PowerShell. In the future we could consider some additional options though.

## Testing Instructions

1. Check out the PR branch and build the app with `npm install && npm start` on Windows.
2. Head over to the app settings modal and take a look at the available Terminal options.
3. There should be only `Command Prompt` and `Warp` app options available.
4. There should be no change on the macOS Studio version.

## Pre-merge Checklist

### DIFF
--- a/src/stores/installed-apps-api.ts
+++ b/src/stores/installed-apps-api.ts
@@ -10,7 +10,7 @@ import {
 import {
 	SupportedTerminal,
 	terminalConfig,
-	isTerminalSupportedOnPlatform,
+	getTerminalsSupportedOnPlatform,
 } from 'src/modules/user-settings/lib/terminal';
 
 const getFirstInstalledEditor = async (): Promise< SupportedEditor | null > => {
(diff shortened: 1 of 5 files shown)


src/stores/installed-apps-api.ts
```diff
@@ -7,7 +7,11 @@ import {
 	supportedEditorConfig,
 	SUPPORTED_EDITORS,
 } from 'src/modules/user-settings/lib/editor';
-import { SupportedTerminal, supportedTerminalNames } from 'src/modules/user-settings/lib/terminal';
+import {
+	SupportedTerminal,
+	terminalConfig,
+	isTerminalSupportedOnPlatform,
```
epeicher: Non-blocking suggestion: What do you think about adding a method that returns directly the terminal supported by platform? Something like `getTerminalsSupportedOnPlatform` so the `lib/terminal` is in charge of doing the filtering? I think this leaves the filtering responsibility to the terminal lib and makes the code more maintainable.

epeicher (approved):
Thanks for improving this @ivan-ottinger! I have tested on both, Mac and Windows, and they display the correct Terminal applications. LGTM! :shipit: 

|Mac | Windows|
|-------|------|
| Image|  Image|

I have left a minor suggestion of refactor to consider, but feel free to progress with this version if you prefer it

src/stores/installed-apps-api.ts
ivan-ottinger: Thank you for the suggestion, Roberto! I have made the change in https://github.com/Automattic/studio/pull/1983/commits/6d22cf7dd44f7ac5088b8c1729146a47fe7e7e05.

I agree it is more maintainable and easier to read as well. 🙂

epeicher: Thanks @ivan-ottinger for addressing the comment! I’ve tested it again on Mac and Windows, and it works perfectly 🙌

PR #1989, opened 2025-11-03
Blueprints: Add a deeplink to create a new site from the blueprint URL

gavande1:
## Related issues

- Fixes STU-920

## Proposed Changes

- Add new deeplink route `wpcom-local-dev://add-site` in studio to handle site creation using deeplink using blueprint urls. 

### Notes

> [!IMPORTANT]
> I have removed the following screen after the suggestion from Gergely.  

I have temporarily added the following error screen to display an error with the blueprint URL. I suggest adding a new screen with blueprint details when the user clicks on the deeplink. 

CleanShot 2025-11-03 at 13 14 12@2x

Currently, it displays the following screen when the user clicks on the deeplink with valid blueprint URL:  

CleanShot 2025-11-03 at 13 15 27@2x

## Testing Instructions

- Check out this branch
- Visit https://jsbin.com/levaburate/2. I have different cases of deep links configured. 
- Click on each link and check the behaviour. 
- Try to create a new site using a valid blueprint. 
- Make sure the site is created normally. 
- Check for any regression in existing flows. 

## Pre-merge Checklist


src/lib/deeplink/handlers/add-site-blueprint-with-url.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Handles the add-site deeplink callback.
+ * This function is called when a user clicks a deeplink like:
+ * wpcom-local-dev://add-site?blueprint_url=<encoded-url>
+ *
+ * It downloads the blueprint from the URL, saves it locally, and opens the Add Site modal
+ * with the blueprint pre-filled.
+ */
+export async function handleAddSiteBlueprintWithUrl( urlObject: URL ): Promise< void > {
```
ivan-ottinger: I think we could consider making the `add-site` deeplink handler name more universal. In a way that it would semantically cover both cases:
- blueprint from linked JSON URL
- blueprint that's a part of URL

And also keep the `Deeplink` in the name.

For example, it could be called `handleAddSiteDeeplink` and then we would just distinguish two parameters in its body (`blueprint_url` and `blueprint`).

src/modules/add-site/components/blueprint-error.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Icon, warning } from '@wordpress/icons';
+import { useI18n } from '@wordpress/react-i18n';
+import Button from 'src/components/button';
+
+interface BlueprintErrorProps {
+	errorMessage: string;
+	onBack: () => void;
+}
+
+export default function BlueprintError( { errorMessage, onBack }: BlueprintErrorProps ) {
```
ivan-ottinger: I like the new error screen and the flexibility it adds when we are presenting the information to the user.

At the same time I am just thinking that maybe we could display the errors in the standard error message and leave the user where they were (in the Studio app) before they clicked the deeplink. 🤔

ivan-ottinger (commented):
Nice! 🚀 

I see this PR is still in the draft, but I wanted to share some thoughts that came to my mind while reviewing it. 🙂 

> I have temporarily added the following error screen to display an error with the blueprint URL. I suggest adding a new screen with blueprint details when the user clicks on the deeplink.

> Currently, it displays the following screen when the user clicks on the deeplink with valid blueprint URL:

What if we used customized `Add a site` step?

Image

When the user lands on the `Add a site` step after opening a deeplink, we could display blueprint information there, the user can customize the site name and other settings. And then they can create the site.

I would say that we could even hide these to prevent any confusion:

Image

and use `Cancel` instead of `Back` button. At the end we could have a step like this:

Image

I think it would be great if we could discuss this with @gcsecsey and @Marinatsu.

src/modules/add-site/components/blueprint-error.tsx
gavande1: Absolutely. We could use the standard error messages, and I prefer that approach. However, showing the actual error might help developers debug the issue more easily. I was thinking from the perspective of developers who are building these blueprints. I would still wait for Marina’s input on the designs.

src/lib/deeplink/handlers/add-site-blueprint-with-url.ts
gavande1: Great suggestion. I will refactor the function name to be flexible, allowing for different types of entry points.

gavande1:
I loved your suggestions, @ivan-ottinger. Your suggestions closely align with my thoughts on showing blueprint information before creating a site. Let's wait for Marina's and Gergely's suggestions.

src/lib/deeplink/handlers/add-site-blueprint-with-url.ts
gavande1: I have updated the function in [45b1469](https://github.com/Automattic/studio/pull/1989/commits/45b14694858a32154908b28bb3014569b59a8d99).

src/modules/add-site/components/blueprint-error.tsx
gcsecsey: > At the same time I am just thinking that maybe we could display the errors in the standard error message and leave the user where they were (in the Studio app) before they clicked the deeplink.

I really like this idea too, I think it'd be great to discuss this in a follow-up. Added STU-967 to track it.

gcsecsey (commented):
Thanks for tackling this issue @gavande1, even with the undecided UI, it's great to see the underlying mechanism coming along! 🙌 

> What if we used customized Add a site step?

We do have a similar plan [in a later issue](https://linear.app/a8c/issue/STU-922), where we'd add a step that contains the current file selector, as well as the UI for importing from a URL. See zJhs43GtRkZmSuiuZKGLp0-fi-0_1 for sketches on this.

> I have temporarily added the following error screen to display an error with the blueprint URL. I suggest adding a new screen with blueprint details when the user clicks on the deeplink.

On the long term, I agree that we'll need a new screen, and as mentioned above we do have plans to implement a new step later. For the scope of this issue, I think we could reuse the error message template we use for invalid templates that are selected from the file system: 

Image

gavande1:
> On the long term, I agree that we'll need a new screen, and as mentioned above we do have plans to implement a new step later. For the scope of this issue, I think we could reuse the error message template we use for invalid templates that are selected from the file system:

Thanks @gcsecsey. I have updated the code to use the existing way of displaying an error message.

gcsecsey (approved):
Thanks for the updates @gavande1! 🙌 

Functionally this works well now, and we can merge this and iterate on the UX later.

| Valid Blueprint | Invalid Blueprint URL | Invalid Blueprint file |
|--------|--------|---|
| Image | Image | Image |

Note that we'll have to double-check the URL scheme and the references after https://github.com/Automattic/studio/pull/2029 is merged.

gavande1:
@gcsecsey, I am deploying this as you suggested. We can iterate on designs as soon as they are available.

PR #1997, opened 2025-11-03
Add flow to create a new site from remote site

sejas:
## Related issues

- Fixes STU-856

## Proposed Changes

- Add a new flow to create a new site and start pulling immediately.

## Testing Instructions

- Enable the flag `streamlineOnboarding` from the UI or from CLI env.
- Click on `Add site`
- Notice that there is a new button to Create a new site from remote.
- Select a remote site and continue with the process
- Observe that the site is created, goes to the sync tab, it has the remote site connected and starts pulling immediately without user interaction.

https://github.com/user-attachments/assets/e337454b-31cc-43e6-af20-a946329e2761

## Pre-merge Checklist


src/modules/add-site/hooks/use-stepper.ts
```diff
@@ -16,14 +16,16 @@ interface StepperStep {
...
 	onCreateSubmit?: ( event: FormEvent ) => void;
 	canSubmitBlueprint?: boolean;
 	canSubmitBackup?: boolean;
+	canSubmitPullRemote?: boolean;
 	canSubmitCreate?: boolean;
 }
 
 interface StepperContext {
-	flow: 'blueprint' | 'backup' | 'create';
+	flow: 'blueprint' | 'backup' | 'create' | 'pullRemote';
```
epeicher: Nit: If they are the same values, we could reuse the [new type here](https://github.com/Automattic/studio/pull/1997/files#diff-da90fcaab90dad62a8cd25b520190037033dea8564e7c99f965185387bb22592R15), similar to [this change](https://github.com/Automattic/studio/pull/1997/files#diff-6ad9bc3c6af5cb53c82533929d1f4630964c1d168d0334906ee27553a70914d4R91)

epeicher (approved):
Thanks @sejas, this is a great functionality! I have tested it, and it works as expected. I can create a site from a WP.com site, and the `Pull` starts automatically. 

One suggestion to consider for a follow-up is to add the search box to the list of sites to connect. I assume that is not a common scenario to have that many sites, but I think it would be helpful. 

Another suggestion to avoid errors during the import when opening the site or wp-admin is to disable the links, but again, this could be a follow-up only if we think it's helpful.

I have left a minor nit comment, but I think this is ready for merging 🙌 . LGTM! :shipit: 

|Select site | Site creating| Site importing | Site Imported 🎉 |
|-------|------|------|------|
| Image|Image |ImageImage |Image

src/modules/add-site/hooks/use-stepper.ts
sejas: Great suggestion ! Done it here https://github.com/Automattic/studio/pull/1997/commits/0fff4070772d714eabd8d2e6ce7073cc754729a3

PR #1998, opened 2025-11-03
Fix RTL styling on WordPress Core components

bcotrim:
## Related issues

- Fixes STU-885

## Proposed Changes

- Created WordPressStyles component using DynamicStylesheet to dynamically load WordPress component styles based on RTL/LTR direction
- Updated DynamicStylesheet to insert styles at the beginning of head for proper CSS loading order
- Fixed production CSS file paths to use relative paths from renderer HTML
- Removed static WordPress CSS import from index.css in favor of dynamic loading

## Testing Instructions

- Run `npm start`
- Switch language to RTL in settings
- Verify WordPress components (modals, checkboxes) display correctly in RTL
- Test in production build: Run `npm run make` and verify styles load correctly

## Pre-merge Checklist

### DIFF
--- a/src/components/default-error-fallback.tsx
+++ b/src/components/default-error-fallback.tsx
@@ -89,12 +89,16 @@ const RightPanel = () => {
 };
 
 export default function DefaultErrorFallback() {
+	const { isRTL } = useI18n();
+	const isRtl = isRTL();
+
+	const href = isRtl
+		? './main_window/styles/wordpress-components-style-rtl.css'
+		: './main_window/styles/wordpress-components-style.css';
+
 	return (
-		<div dir={ 'ltr' }>
-			<DynamicStylesheet
-				id="wordpress-components-style-error-fallback"
-				href={ './main_window/styles/wordpress-components-style.css' }
-			/>
+		<div dir={ isRtl ? 'rtl' : 'ltr' }>
+			<DynamicStylesheet id="wordpress-components-style-error-fallback" href={ href } />
 			<VStack
 				className={ cx(
 					'h-screen bg-chrome backdrop-blur-3xl pr-chrome app-drag-region',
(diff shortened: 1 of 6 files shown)


ivan-ottinger (approved):
Thank you for fixing the issue with the broken RTL styles, Bernardo.

The proposed code looks good to me and works correctly. When a RTL local is set, the related WordPress Core components styles are being loaded as expected.

I have tested it in the dev and also built app.

The paddings around `X` button in the modal and around checkboxes is now correct.

Image

Image

Just the unit tests need adjustment.

src/components/default-error-fallback.tsx
```diff
@@ -92,8 +92,8 @@ export default function DefaultErrorFallback() {
 	return (
 		<div dir={ 'ltr' }>
 			<DynamicStylesheet
-				id="wordpress-components-style"
-				href={ '../main_window/styles/wordpress-components-style.css' }
+				id="wordpress-components-style-error-fallback"
+				href={ './main_window/styles/wordpress-components-style.css' }
```
epeicher: Nit: Should we apply the same logic for RTL for the errors than here https://github.com/Automattic/studio/pull/1998/files#diff-d47d7dfb85379fc920c32c6ccd0a9f5817e5104a1db25afd75a39b779fc5af70R22 ? 🤔 This is what I get 

Image

epeicher (approved):
Thanks @bcotrim for solving this! I have tested it in dev and production mode and it works as expected. I left a minor non-blocking comment. Code also LGTM! :shipit: 

Image

katinthehatsite (approved):
This works for me, I did not spot any regressions on my end 👌

src/components/default-error-fallback.tsx
bcotrim: Nice catch! Addressed in https://github.com/Automattic/studio/pull/1998/commits/dbd0a597f608321a7a690f0cbb7f27f97260512d

sejas (commented):
Nice! Thanks for fixing the RTL CSS issue! ✨

PR #2000, opened 2025-11-04
Studio: Resolve warnings about memory leaks related to listeners

katinthehatsite:
## Related issues

Closes STU-937

## Proposed Changes

I propose to use one listener for in the parent component for `site-context-menu-action` and find the relevant site after, instead of using an individual listener for each site item. This is how it works: 

  **Before (multiple listeners):**

*  Each `SiteItem` component subscribed to `site-context-menu-action`
 * When the event fired, all 11+ listeners would execute
*  Each listener checked if `(data.siteId === site.id)` to see if it should
  handle the event
*  Only the matching site item would process the action

  **After (single listener):**

* Only the `SiteMenu` parent subscribes once to `site-context-menu-action`
* When the event fires, only 1 listener executes
* It finds the matching site
* It processes the action for that site

## Testing Instructions

 * Open the developer tools
 * Navigate through the app e.g. switching between sites and tabs
 * Confirm that you don't see the React warning as below:

```
VM4 sandbox_bundle:2 MaxListenersExceededWarning: Possible 
EventEmitter memory leak detected. 11 site-context-menu-action listeners 
added. Use emitter.setMaxListeners() to increase limit
    at _addListener (VM4 sandbox_bundle:2:43268)
    at IpcRenderer.addListener (VM4 sandbox_bundle:2:46156)
    at Object.subscribe (<anonymous>:139:24) 
```

## Pre-merge Checklist

### DIFF
--- a/src/components/site-menu.tsx
+++ b/src/components/site-menu.tsx
@@ -199,7 +199,7 @@ export default function SiteMenu( { className }: SiteMenuProps ) {
 		const unsubscribe = window.ipcListener.subscribe(
 			'site-context-menu-action',
 			async ( _, actionData: { action: string; siteId: string } ) => {
-				const site = data.find( ( s ) => s.id === actionData.siteId );
+				const site = data.find( ( site ) => site.id === actionData.siteId );
 				if ( ! site ) {
 					return;
 				}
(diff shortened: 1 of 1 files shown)


epeicher (approved):
Thanks for solving this @katinthehatsite! I have tested it, and it works as advertised. I was able to reproduce the error in `trunk` with 11 sites and the issue is not reproduced anymore in this branch 🙌 . 

Also, I have tested the context menu tasks, and they work as expected. Code changes also LGTM! :shipit: 

|Before | After|
|-------|------|
| Image| Image|

ivan-ottinger (approved):
Nice fix! I can confirm the issue is resolved. There are no more warnings logged in the console and the site context menu is still working as expected. 👌🏼

PR #2002, opened 2025-11-04
Implement new site launch and import buttons

epeicher:
## Related issues

- Closes STU-910

https://github.com/user-attachments/assets/4ef84ce6-1529-40c0-ac45-de07d64f1085

Design RToz6tIuQ7nlZrikBte4GU-fi-10114_133424 

## Proposed Changes

- Implemented new `Publish site` and `Pull site` buttons to replace the single `Connect site` button
- Added some state to keep track of the different modes  (Push, Pull or Connect) and the selected site when required
- Added the SyncDialog component to the ContentTabSync
- Added new handlers to open or not the new Dialog
- Updated UI copy and wording throughout sync components for clarity
- Updated modal titles and button text to reflect the operation mode
- Added custom tooltip text support to ConnectButton component
- Updated tests

## Testing Instructions

1. Start Studio and select a local site
2. Navigate to the Sync tab
3. Log in to WordPress.com if not already authenticated
4. **Test Publish Flow:**
   - Click `Publish site` button
   - Verify the site selector modal opens with title `Publish your site`
   - Select a site from the list
   - Verify the sync dialog opens showing `Push to [Environment]`
   - Select `Push` and check there are no errors

5. **Test Pull Flow:**
   - Click `Pull site` button
   - Verify the site selector modal opens with title `Select a site to import`
   - Select a site from the list
   - Verify the sync dialog opens showing `Pull from [Environment]`
   - Click on `Pull` and check the pull is success

6. **Test Connect Flow (when sites already exist):**
   - Click `Connect another site` button
   - Verify the modal opens with title `Connect your site`
   - Select a site and verify it connects without opening the sync dialog
   -
7. **Test Offline Mode:**
   - Disconnect from the internet
   - Verify `Publish site` and `Pull site` shows tooltip with a meaningful message
   - Verify buttons are disabled when offline

## Pre-merge Checklist


src/modules/sync/index.tsx
```diff
@@ -168,45 +235,73 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
...
+							selectedRemoteSite: null,
+						} );
+					} }
+					onRequestClose={ () => {
+						setModalState( {
+							mode: null,
+							selectedRemoteSite: null,
+						} );
+					} }
+				/>
```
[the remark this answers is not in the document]
epeicher: Fixed in [e276491](https://github.com/Automattic/studio/pull/2002/commits/e27649198fb85065e0792f39cb8db0d352839185)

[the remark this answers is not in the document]
epeicher: Solved in [e276491](https://github.com/Automattic/studio/pull/2002/commits/e27649198fb85065e0792f39cb8db0d352839185)

src/modules/sync/index.tsx
```diff
@@ -147,6 +169,51 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
...
+			const connectedSite = await handleConnect( selectedSiteFromList );
+			if ( ! connectedSite ) {
+				return;
+			}
+		}
+		const siteToUse = connectedSites.find( ( site ) => site.id === siteId ) || selectedSiteFromList;
+
+		dispatch( connectedSitesActions.closeModal() );
+
+		if ( mode == 'push' || mode == 'pull' ) {
```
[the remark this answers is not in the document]
epeicher: 👍 done in [53b7a23](https://github.com/Automattic/studio/pull/2002/commits/53b7a23e1a785e5c0181da5c2af9427de8f68281)

src/modules/sync/components/sync-sites-modal-selector.tsx
```diff
@@ -335,14 +351,28 @@ function Footer( {
...
+	mode?: SyncModalMode;
 } ) {
 	const { __ } = useI18n();
 
+	const getButtonText = () => {
+		switch ( mode ) {
+			case 'push':
+				return __( 'Publish' );
+			case 'pull':
+				return __( 'Import' );
```
sejas:
```suggestion
			case 'push':
			case 'pull':
				return __( 'Next' );
```

src/modules/sync/index.tsx
```diff
@@ -34,13 +42,13 @@ function SiteSyncDescription( { children }: PropsWithChildren ) {
 				</div>
 				<div className="max-w-[40ch] text-a8c-gray-70 a8c-body">
 					{ __(
-						'Connect your existing WordPress.com or Pressable sites with Jetpack activated, or create a new one. Then share your work with the world.'
+						'Launch your existing WordPress.com or Jetpack-activated Pressable sites, or import an exisiting one. Then, share your work with the world.'
```
sejas: There is a typo here:

```suggestion
						'Launch your existing WordPress.com or Jetpack-activated Pressable sites, or import an existing one. Then, share your work with the world.'
```

src/modules/sync/index.tsx
```diff
@@ -147,6 +169,54 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
 		}
 	};
 
+	const handleLaunchSite = () => {
+		setModalState( ( prev ) => ( { ...prev, mode: 'push' } ) );
+		dispatch( connectedSitesActions.openModal() );
```
sejas: I wonder if the modalState could be part of the Redux slice and the openModal could receive the `mode` as parameter.

src/modules/sync/index.tsx
```diff
@@ -168,45 +238,73 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
...
 			) }
 
 			{ isModalOpen && (
 				<SyncSitesModalSelector
+					mode={ modalState.mode || 'connect' }
 					isLoading={ isFetching }
-					onRequestClose={ () => dispatch( connectedSitesActions.closeModal() ) }
+					onRequestClose={ () => {
+						dispatch( connectedSitesActions.closeModal() );
+						setModalState( ( prev ) => ( { ...prev, mode: null } ) );
```
sejas: If the modalState would live inside the Redux slice, then we could clear the mode inside closeModal.

sejas (approved):
Cool! I confirm that clicking on Publish site let me choose the site and then opens the Selective sync modal. I think works as a first iteration, but I would prefer if the modal is the "same" and it doesn't blink.
Also, I wonder if the connection should only happen after the user actually clicks on the final `Push` or `Pull`. If the user selects a remote site and then cancels the second modal, I would expect that no sites have been connected yet. That's my personal opinion, so I'm open to other opinions.

As I said, I think this is a good first iteration and I think we can merge it after 1.6.3 is out, and create a follow-up to polish the UX.

I also left some suggestions, like renaming the first modal button to `Next` and maybe moving the modalState to the Redux slice.

https://github.com/user-attachments/assets/ae4b9ae3-565b-4c1e-a344-25f88ae7b8ad

src/modules/sync/index.tsx
epeicher: That is a great suggestion! I have actually needed to do it on https://github.com/Automattic/studio/pull/2025, I can do it in this PR though so we test with the final code

src/modules/sync/components/sync-sites-modal-selector.tsx
epeicher: 👍 Done in  [c5bca9b](https://github.com/Automattic/studio/pull/2002/commits/c5bca9b98656bfd7841ca5061f4dde14a8912821)

src/modules/sync/index.tsx
epeicher: Well spotted! Fixed in [bb66758](https://github.com/Automattic/studio/pull/2002/commits/bb6675808d6b5569851152d037574f93002b4e80)

src/modules/sync/index.tsx
epeicher: I've gone ahead and moved to Redux, it simplifies the code a lot [38d8ae4](https://github.com/Automattic/studio/pull/2002/commits/38d8ae4e054f2318108317d38650ab0d1174e2f2)

src/modules/sync/index.tsx
epeicher: That's absolutely right, done as part of 38d8ae4

src/modules/sync/components/sync-sites-modal-selector.tsx
sejas: Thank you!

epeicher:
Thanks @sejas  for your review!

> Also, I wonder if the connection should only happen after the user actually clicks on the final `Push` or `Pull`. If the user selects a remote site and then cancels the second modal, I would expect that no sites have been connected yet. That's my personal opinion, so I'm open to other opinions.

That's a good point, I was hesitant to change that or not after initial implementation, but I agree it can be _surprising_ to users if the site is connected even if they click cancel. I have implemented the change [953be86](https://github.com/Automattic/studio/pull/2002/commits/953be86ed6b9b0611449b8c21e2c4778de7d4d9b). Now, there is another blink when pulling/pushing and listing the site, but that can also be tackled as a follow-up. If we decide to go with the Connect always approach, the commit can be reverted easily.

Additionally, I have implemented all your suggestions, so I would appreciate another quick look if possible 🙏

sejas (approved):
@epeicher , great work. Thanks addressing the feedback. Connecting the site in the last step is a huge improvement. I'll create a low priority issue to fix the blink between modals. Let's merge the PR once the V1.6.3 is out and the channel is green.
https://github.com/user-attachments/assets/5a7a1548-bfa4-4d6e-8daa-c18a44108a5b

src/modules/sync/index.tsx
```diff
@@ -168,45 +214,64 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
...
+							{ __( 'Publish site' ) }
+						</ConnectButton>
+						<ConnectButton
+							variant="secondary"
+							connectSite={ handleImportSite }
 							disableConnectButtonStyle={ true }
+							tooltipText={ __( 'Importing a remote site requires an internet connection.' ) }
 						>
-							{ __( 'Connect site' ) }
+							{ __( 'Pull site' ) }
```
katinthehatsite, line 239: I think we should have a blue border on this button according to the design:

Image

And I am currently seeing the black one:

Image

src/modules/sync/index.tsx
```diff
@@ -34,13 +41,13 @@ function SiteSyncDescription( { children }: PropsWithChildren ) {
...
-						'Connect your existing WordPress.com or Pressable sites with Jetpack activated, or create a new one. Then share your work with the world.'
+						'Launch your existing WordPress.com or Jetpack-activated Pressable sites, or import an existing one. Then, share your work with the world.'
 					) }
 				</div>
 				<div className="mt-6">
 					{ [
 						__( 'Push and pull changes from your live site.' ),
-						__( 'Connect multiple environments.' ),
+						__( 'Supports staging and production sites.' ),
 						__( 'Sync database and file changes.' ),
```
katinthehatsite, line 51: I am finding `file changes` in there to sounding a bit strange, I would change it to something simpler like:


```suggestion
						__( 'Sync files and database.' ),
```

katinthehatsite (commented):
Nice work, everything seems to be working as expected for me ❤️ 

I found an edge case where I was able to start push process for two different sites at the same time (it seems that it is not intentional). This is how I did it:

* On a Studio site A, that does not have any sites connected, navigate to `Sync` tab
* Start either `Pull` or `Push` push from the modal
* Switch to a Studio site B that **does not have any connections**
* Start another push or pull processes from the modal
* Observe two push/ pull processes running at the same time
* If Studio site B has sites connected already, then the `Push` and `Pull` buttons are correctly disabled

katinthehatsite (commented):
Also, this is maybe a suggestion, I was thinking that it would be cool to have a `Back ` button in the connection screen so that you can navigate back from the Sync modal to connection modal if you change your mind about the site you want to connect.

This definitely does not need to be implemented in this PR, we can just add this for consideration for @Marinatsu

src/modules/sync/index.tsx
epeicher: No problem changing it! It was already like that, though, and it’s in the design, so maybe worth a quick check with others? What do you think @sejas or @wojtekn?

src/modules/sync/index.tsx
epeicher: Well spotted! Changed as part of 51f1e67 in a similar way than the `Export database` button. Ideally, we should not need to override the style buttons, but I would tackle that as a follow-up, as it implies checking all the Buttons used to avoid regressions

epeicher:
Thanks @katinthehatsite for your review!

> I found an edge case where I was able to start push process for two different sites at the same time 

Well spotted! I think we should apply the same logic than to Pull and Push buttons, and disabled them when there is another site syncing, what do you think @sejas ? This would be the result:

CleanShot 2025-11-07 at 13 10 15@2x

katinthehatsite:
> I think we should apply the same logic than to Pull and Push buttons, and disabled them when there is another site syncing, what do you think @sejas ? This would be the result:

This makes sense to me 👍

sejas:
Great addition! Thanks for making it consistent ✨

src/modules/sync/components/connect-button.tsx
```diff
@@ -3,40 +3,40 @@ import Button, { ButtonVariant } from 'src/components/button';
 import offlineIcon from 'src/components/offline-icon';
 import { Tooltip } from 'src/components/tooltip';
 import { useOffline } from 'src/hooks/use-offline';
-import { cx } from 'src/lib/cx';
 
 interface ConnectButtonProps {
 	variant: ButtonVariant;
 	connectSite?: () => void;
-	disableConnectButtonStyle?: boolean;
```
epeicher, line 11: This was always used as true, so it has been removed, and the functionality updated

sejas (approved):

PR #2005, opened 2025-11-05
Upgrade ESLint to v9

fredrikekelund:
## Related issues

- Fixes STU-968

## Proposed Changes

Due to some issues with the ESLint language server in Zed, I decided to tackle the ESLint v9 upgrade. Per the [migration guide](https://eslint.org/docs/latest/use/migrate-to-9.0.0), the biggest change is that the old `eslintrc` config format is now deprecated. "Flat config" is now the default config format. 

Here are the changes in this PR:
- Upgraded the core `eslint` dependency and related dependencies to the latest versions
- Migrated the config to the new flat config format in `eslint.config.mjs`
- Autofixed some minor issues
- Manually fixed some issues related to React hooks

## Testing Instructions

1. Run `npm run lint` and ensure there are no warnings or errors
2. Smoke test the Assistant tab and ensure there are no regressions (this is where I had to make manual changes to fix the linter errors)

## Pre-merge Checklist


src/hooks/use-ai-icon.ts
```diff
@@ -23,33 +23,33 @@ const useAiIcon = () => {
...
+			} else if ( stateName === 'thinking' && thinkingInput ) {
+				// eslint-disable-next-line react-hooks/immutability
+				thinkingInput.value = value;
+			} else if ( stateName === 'typing' && typingInput ) {
+				// eslint-disable-next-line react-hooks/immutability
+				typingInput.value = value;
+			}
+		},
+		[ inactiveInput, thinkingInput, typingInput ]
+	);
```
fredrikekelund, line 47: We could have made this change in `src/components/ai-input.tsx`, where this hook is consumed. I just thought this was a more elegant API.

src/components/content-tab-assistant.tsx
```diff
@@ -200,62 +269,6 @@ const AuthenticatedView = memo(
 			[ submitPrompt, siteId, instanceId ]
 		);
 
-		const RenderLastMessage = useCallback(
```
fredrikekelund, line 203: Defining a React component this way is a real code smell. I believe my change preserves the previous behavior, but this is an area that could use an extra pair of eyes.

packages/eslint-plugin-studio/src/rules/require-lock-before-save.js
fredrikekelund, line 1: The changes in this file are all just auto-formatting.

packages/eslint-plugin-studio/package.json
```diff
@@ -1,13 +1,14 @@
 {
-	"name": "@studio/eslint-plugin",
+	"name": "eslint-plugin-studio",
+	"private": true,
```
fredrikekelund, line 3: The previous package name triggered some validation logic in ESLint around the plugin name, so I opted for this simpler version. I also added the `private` field to indicate that we don't intend to publish this code to npm.

eslint.config.mjs
fredrikekelund, line 1: This file was initially generated with `npx @eslint/migrate-config .eslintrc.json` and then tweaked through several iterations. The flat config file format doesn't bring about any huge changes for us. It's really mostly formatting. 

If you're interested in the motivation behind ESLint's change, see the original [RFC](https://github.com/eslint/rfcs/tree/main/designs/2019-config-simplification) and [this blog post](https://eslint.org/blog/2023/10/flat-config-rollout-plans/).

package.json
```diff
@@ -109,6 +108,7 @@
 		"ts-loader": "^9.5.4",
 		"ts-node": "^10.0.0",
 		"typescript": "~5.9.3",
+		"typescript-eslint": "^8.46.3",
```
wojtekn, line 110: Let's update https://github.com/Automattic/studio/blob/843bd4328c1051c20f730a41a55db079354c0c07/.github/dependabot.yml to account for that change.

package.json
```diff
@@ -45,10 +45,10 @@
 		"@electron-forge/maker-squirrel": "^7.10.2",
 		"@electron-forge/maker-zip": "^7.10.2",
 		"@electron-forge/plugin-auto-unpack-natives": "^7.10.2",
+		"@eslint/js": "^9.39.1",
```
wojtekn, line 48: Let's add this namespace to https://github.com/Automattic/studio/blob/843bd4328c1051c20f730a41a55db079354c0c07/.github/dependabot.yml

src/components/ai-input.tsx
```diff
@@ -47,28 +46,16 @@ const UnforwardedAIInput = (
 	}, [ disabled, inputRef ] );
 
 	useEffect( () => {
-		startStateMachine();
+		setInputState( 'inactive', disabled );
```
wojtekn, line 49: It seems tests fail on that - should we mock them?

wojtekn (commented):
Thanks for updating eslint, looks good.

eslint.config.mjs
```diff
@@ -0,0 +1,74 @@
...
+	{
+		plugins: {
+			studio: studio,
+		},
+		languageOptions: {
+			globals: {
+				...globals.browser,
+				...globals.node,
+			},
+			ecmaVersion: 5,
```
epeicher: Nit: I don't think we need to support `ecmaVersion: 5`, so we could remove this and use the default "latest" as [recommended in the documentation](https://eslint.org/docs/latest/use/configure/language-options#specifying-javascript-options)

epeicher (commented):
Thanks @fredrikekelund for upgrading this! I have tested it, and it works as expected. I have executed the `npm run lint` without any issues, and I can see linting errors if in my IDE (if I add some 😄). 

I have also tested the Assistant tab, and I have not found any regressions. Once we address comments by @wojtekn  and fix the Unit Test, this could progress.

package.json
fredrikekelund: Thanks for highlighting that. I've made the required changes in this file 👍

eslint.config.mjs
fredrikekelund: I believe you're right. Vite should handle any transpiling needs with esbuild.

fredrikekelund:
Thanks for the reviews. I went over your comments, @wojtekn, and fixed the failing tests (I had overlooked that there was a global mock call for `src/hooks/use-ai-icon` in `jest-setup.ts`).

While going over our ESLint dependencies in `package.json`, I also noticed we depended on `eslint-plugin-jest-dom` without actually using it. I've enabled that plugin and auto-fixed the issues it pointed out.

fredrikekelund:
I'd appreciate an approval, @epeicher or @wojtekn 🙂

epeicher (approved):
Thanks for addressing the comments. All tests pass now. This LGTM! :shipit:

wojtekn:
> While going over our ESLint dependencies in package.json, I also noticed we depended on eslint-plugin-jest-dom without actually using it. I've enabled that plugin and auto-fixed the issues it pointed out.

Thanks for doing that! 💪

PR #2007, opened 2025-11-05
Studio: Be more specific in push success email by listing individual sync elements

katinthehatsite:
## Related issues

Closes STU-759

## Proposed Changes

* Please refer to description and details in 195447-ghe-Automattic/wpcom

## Testing Instructions

*  Please refer to the testing instructions in 195447-ghe-Automattic/wpcom

## Pre-merge Checklist


gavande1 (approved):
I have tested this. LGTM 👍

src/ipc-handlers.ts
```diff
@@ -824,7 +825,13 @@ export async function pushArchive(
...
 	];
 
-	if ( optionsToSync ) {
+	if ( specificSelectionPaths && specificSelectionPaths.length > 0 ) {
+		const options = optionsToSync ? [ ...optionsToSync, 'paths' ] : [ 'paths' ];
+		formData.push( [ 'options', options.join( ',' ) ] );
+
+		const joinedPaths = specificSelectionPaths.join( ',' );
+		formData.push( [ 'include_path_list', joinedPaths ] );
+	} else if ( optionsToSync ) {
```
sejas: @katinthehatsite , Is there any specific reason we are sending the option `paths` ?
I think that option is not necessary. For the same reason I would suggest renaming `include_path_list` var name. Maybe `push_selection_paths`? 🤔 .

```suggestion
	if ( specificSelectionPaths && specificSelectionPaths.length > 0 ) {
		const joinedPaths = specificSelectionPaths.join( ',' );
		formData.push( [ 'push_selection_paths', joinedPaths ] );
	}

	if ( optionsToSync ) {
```

sejas (changes requested):
@katinthehatsite , it works great! I tested it the PR with the backend PR 195447-ghe-Automattic/wpcom . However I noticed that we are sending `paths` option, which I believe is not necessary. In that case I also suggest renaming the `include_path_list` param to avoid confusion with a valid param that uses VaultPress.

src/ipc-handlers.ts
katinthehatsite: > @katinthehatsite , Is there any specific reason we are sending the option paths ?

Don't we need to send SQL which does not come as a part of `wp-content` so that we can display information regarding the database being pushed to the user?

katinthehatsite: For example, this is what we get when choosing database: 

```
optionsToSync [ 'sqls', 'plugins' ]
specificSelectionPaths [ 'plugins/akismet' ]
```
And this is what we get when choosing all the files: 

```

This is what happens when we select everything:

```optionsToSync [ 'all' ]
specificSelectionPaths undefined
```

And this is what happens when we select just database:

```
optionsToSync [ 'sqls' ]
specificSelectionPaths undefined
```

sejas: My comment is specific about appending `pahts` to the options we send to the backend:
`const options = optionsToSync ? [ ...optionsToSync, 'paths' ] : [ 'paths' ];`

Studio was not using that option before, and AFAIK that option is not necessary.

sejas: I was suspicious of the `paths` option because the official parameter `include_path_list` that VaultPress requires needs to be in base64. It seems to me that we are sending both the `paths` option and the `include_path_list` to the backend, and then from the backend to VaultPress. However, the `include_path_list` will actually be ignored because it's not in base64. 

> Don't we need to send SQL which does not come as a part of wp-content so that we can display information regarding the database being pushed to the user?

Yes, sqls is a valid option, that we need to send when database is checked.

> For example, this is what we get when choosing database: ....

Those are correct values ! 👌

katinthehatsite: Made some changes based on your suggestions, let me know what you think

sejas (approved):
Thanks for removing the unused `paths` option and renaming the param from `include_path_list` to `list_sync_items`.

I tested it and it works as expected. I received the email displaying the selected items.

While testing the PR, I found an issue, but it's also present on trunk, so let's fix it separately in STU-1003.  

Let's rebase onto trunk and make sure the CI passes.

PR #2018, opened 2025-11-05
Make PR builds optional

wojtekn:
## Related issues

- N/A

## Proposed Changes

- I propose to make PR builds optional. They are not used for anything, and we can save processing time and make builds quicker.

## Testing Instructions

- Confirm if build goes through, and that steps can be still triggered manually in Buildkite

## Pre-merge Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -58,14 +58,14 @@ steps:
       - github_commit_status:
           context: E2E Tests
 
-  - block: "🚦 Build for Mac?"
+  - input: "🚦 Build for Mac?"
     prompt: "Do you want to build Mac dev binaries for this PR?"
-    key: block-dev-mac
+    key: input-dev-mac
     if: build.tag !~ /^v[0-9]+/ && build.branch != 'trunk'
 
   - group: 📦 Build for Mac
     key: dev-mac
-    depends_on: block-dev-mac
+    depends_on: input-dev-mac
     if: build.tag !~ /^v[0-9]+/
     steps:
       - label: 🔨 Mac Dev Build - {{matrix}}
@@ -112,14 +112,14 @@ steps:
           - x64
           - arm64
 
-  - block: "🚦 Build for Windows?"
+  - input: "🚦 Build for Windows?"
     prompt: "Do you want to build Windows dev binaries for this PR?"
-    key: block-dev-windows
+    key: input-dev-windows
     if: build.tag !~ /^v[0-9]+/ && build.branch != 'trunk'
 
   - group: 📦 Build for Windows
     key: dev-windows
-    depends_on: block-dev-windows
+    depends_on: input-dev-windows
     if: build.tag !~ /^v[0-9]+/
     steps:
       - label: 🔨 Windows Dev Build - {{matrix}}
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -109,9 +109,6 @@ steps:
         matrix:
           - x64
           - arm64
-        notify:
-          - github_commit_status:
-              context: All Mac Dev Builds
     depends_on: block-dev-mac
     allow_dependency_failure: true
     if: build.tag !~ /^v[0-9]+/
@@ -137,9 +134,6 @@ steps:
         matrix:
           - x64
           - arm64
-        notify:
-          - github_commit_status:
-              context: All Windows Dev Builds
     depends_on: block-dev-windows
     allow_dependency_failure: true
     if: build.tag !~ /^v[0-9]+/
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -58,6 +58,11 @@ steps:
       - github_commit_status:
           context: E2E Tests
 
+  - block: "🚦 Build for Mac?"
+    prompt: "Do you want to build Mac dev binaries for this PR?"
+    key: block-dev-mac
+    if: build.tag !~ /^v[0-9]+/ && build.branch != 'trunk'
+
   - group: 📦 Build for Mac
     key: dev-mac
     steps:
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -125,9 +134,8 @@ steps:
         matrix:
           - x64
           - arm64
-        notify:
-          - github_commit_status:
-              context: All Windows Dev Builds
+    depends_on: block-dev-windows
```
wojtekn: Removing `notify` to ensure that there won't be pending checks under PR, which would cause the whole PR to be marked as pending.

.buildkite/pipeline.yml
```diff
@@ -104,11 +109,15 @@ steps:
         matrix:
           - x64
           - arm64
-        notify:
-          - github_commit_status:
-              context: All Mac Dev Builds
+    depends_on: block-dev-mac
+    allow_dependency_failure: true
```
AliSoftware: This shouldn't be necessary here—especially given a `block` step like the one you depend on here is not av step that actually runs a job in an agent and this will never have a "failure" state per se.

```suggestion
```

.buildkite/pipeline.yml
```diff
@@ -125,9 +134,8 @@ steps:
         matrix:
           - x64
           - arm64
-        notify:
-          - github_commit_status:
-              context: All Windows Dev Builds
+    depends_on: block-dev-windows
+    allow_dependency_failure: true
```
AliSoftware:
```suggestion
```

.buildkite/pipeline.yml
```diff
@@ -104,11 +109,15 @@ steps:
         matrix:
           - x64
           - arm64
-        notify:
-          - github_commit_status:
-              context: All Mac Dev Builds
+    depends_on: block-dev-mac
```
AliSoftware: 💄 💡 Totally opinionated, but I think it would be more readable to move that `depends_on:` attribute in earlier lines in this YAML node, eg on line 70 right after the `label:` attribute.

That would make the dependency more visible (and also closer to the corresponding `key:` of the previous `block` step on line 63), which arguably will make the connection/logic between those 2 steps easier to understand?

(Same suggestion for the Windows one down below obviously)

.buildkite/pipeline.yml
wojtekn: Isn't it needed for production/dev builds on trunk, which won't have `block-dev-mac` dependency, and for which it would return `false`?

.buildkite/pipeline.yml
wojtekn: Makes sense. Should it be placed after `key` rather than `steps -> label` to target the whole group?

.buildkite/pipeline.yml
AliSoftware: If a `block` step is skipped due to an `if:` condition, Buildkite automatically considers it a fulfilled dependency 🙂 

See https://buildkite.com/docs/pipelines/configure/dependencies#order-of-operations:

> However, if the step you're dependent on is excluded from the build due to an `if` condition, the dependency will be ignored and the step that depends on it will run once any other dependencies are satisfied.

.buildkite/pipeline.yml
AliSoftware: Ah, good catch! Yes indeed.

I misread which indentation level this `depends_on` was on and thus which YAML node it was associated to at first. Which kinda proves the point that having those so far below makes it harder to read… also to which parent node they're associated 😉 

Given this, I'd actually probably also move the `if:` from line L110/R114 higher and closer to the `group` it belongs to too?

wojtekn: Thanks, I moved them up.

However, there are still some steps that have `if` as a trailing item, and steps with long `command` that have `depends_on`, `if`, and other control nodes below.

I would skip more ordering changes under this PR, and we could reorganize them separately if we want.

.buildkite/pipeline.yml
wojtekn: Thanks, removed.

.buildkite/pipeline.yml
wojtekn: Thanks, I've got it wrong then. Let's try.

.buildkite/pipeline.yml
AliSoftware: Fair enough.
It's only an opinionated readability matter anyway, nothing critical nor blocking 👍

AliSoftware (dismissed):

.buildkite/pipeline.yml
```diff
@@ -58,8 +58,15 @@ steps:
       - github_commit_status:
           context: E2E Tests
 
+  - block: "🚦 Build for Mac?"
```
AliSoftware:
```suggestion
  - input: "🚦 Build for Mac?"
```

.buildkite/pipeline.yml
```diff
@@ -104,13 +111,16 @@ steps:
         matrix:
           - x64
           - arm64
-        notify:
-          - github_commit_status:
-              context: All Mac Dev Builds
-    if: build.tag !~ /^v[0-9]+/
+
+  - block: "🚦 Build for Windows?"
```
AliSoftware:
```suggestion
  - input: "🚦 Build for Windows?"
```
`block` steps are same as `input` except that they also implicitly add dependencies to all steps declared before it, and make all steps declared after it be dependent on it too. Which is not what we want here.
Instead we want `input` and declare our dependencies explicitly with `key`+`depends_on` (which we already do)

AliSoftware (commented):
Sorry I missed this important issue between block vs input in my previous review

.buildkite/pipeline.yml
wojtekn: Thanks, updated. It seems that the chart shows the chain of dependencies correctly now.

AliSoftware (approved):

PR #2033, opened 2025-11-06
Add pm2 cli commands

bcotrim:
## Related issues

- Related to STU-950

## Proposed Changes

- Added PM2 daemon to manage Studio related sub-processes
- Added Studio HTTP proxy for custom domains with routing logic
- Added SSL certificate generation for HTTPS custom domains
- Added hosts file management 
- Infrastructure setup for CLI-based site management

  **Note:** The actual WordPress site start logic is not implemented in this PR and will be added in a follow-up PR. This PR focuses on establishing the infrastructure (PM2 daemon, proxy server, certificate management, hosts file management).

## Testing Instructions

- Code review
- CI checks

## Pre-merge Checklist


cli/package.json
fredrikekelund, line 1: Do you think we should reconsider how CLI dependencies are handled now and move everything that only the CLI depends on to this `package.json`? Or is pm2 still an edge case?

EDIT: I realize now that Vite copies the entire `node_modules` folder to the build directory, so we should probably NOT move all CLI dependencies here, since that would make the build bundle unnecessarily large.

cli/lib/pm2-manager.ts
```diff
@@ -0,0 +1,381 @@
...
+		return require( 'pm2' );
+	} catch ( error ) {
+		const possiblePaths: string[] = [
+			path.join( __dirname, 'node_modules', 'pm2' ),
+			path.join( path.dirname( __dirname ), 'node_modules', 'pm2' ),
+			path.join( __dirname, '..', 'node_modules', 'pm2' ),
+			path.join( __dirname, '..', '..', 'node_modules', 'pm2' ),
+			path.join( __dirname, '..', '..', '..', 'node_modules', 'pm2' ),
+			path.resolve( process.cwd(), 'node_modules', 'pm2' ),
+		];
```
fredrikekelund, line 32: Plenty of variation here 🤔 At first, I thought that it would be convenient if we could continue shipping the CLI as a single JS file to be executed by Electron's bundled node.js runtime, but then I realized that the PHP-WASM binaries will need to be handled similarly anyway, so maybe this is something we can live with after all…

cli/lib/pm2-manager.ts
```diff
@@ -0,0 +1,381 @@
...
+
+/**
+ * Start the proxy server via PM2
+ * This launches the CLI with `proxy start --managed` which runs the proxy servers
+ */
+export async function startProxyProcess( cliPath: string ): Promise< ProcessDescription > {
+	await ensureDaemonRunning();
+
+	return new Promise( ( resolve, reject ) => {
+		const processConfig: pm2.StartOptions = {
```
```suggestion
		const processConfig: Pm2StartOptions = {
```
Doing this and adding an import statement like so at the top seems to work. 

```ts
import { StartOptions as Pm2StartOptions } from 'pm2';
```

fredrikekelund (commented):
I collected some thoughts and questions as I went over the PR.

The biggest question, though, is whether we take this pm2 approach or something more homegrown. I have some thoughts there, but I'd like to hear yours, too, after working on this change. Let's discuss this in a huddle later 🙂

fredrikekelund (approved):
Bernardo and I worked on this together at our meetup last week. This is the first step towards running sites with the Studio CLI. The next step will be to move the Playground PHP-WASM dependencies to the CLI. 

I confirm that this PR works as intended for implementing basic pm2 management features, and for starting the custom domain proxy 👍

Please note that this PR does not affect the CLI build yet, because we deliberately didn't include a call to `registerCommand` from `cli/commands/site/start`.

PR #2038, opened 2025-11-07
Studio: Refresh the CTA for users without sites

katinthehatsite:
## Related issues

Closes STU-912

## Proposed Changes

This PR adds a modal with call to action for users that don't have any WP.com sites:

Screenshot 2025-11-07 at 4 08 17 PM

Original design: RToz6tIuQ7nlZrikBte4GU-fi-10114_134320 (minus the image on the top as we are not supposed to have the image there.

I made a small change to the design with the modal title:

* firstly, we should not really use the logo in the title, it does not work well as it is too large to use with the default title size and it does not work with that type
* we could do the logo in the title but we need to do some CSS to achieve that look (which I can do if we really want that)
* I was trying to find a simpler solution that features the logo but does not require a whole lot of customizations (this is open for discussion along with the copy)

## Testing Instructions

* Apply the following diff

```diff --git a/src/hooks/use-fetch-wpcom-sites/index.tsx b/src/hooks/use-fetch-wpcom-sites/index.tsx
index cc226545..5e39a4cd 100644
--- a/src/hooks/use-fetch-wpcom-sites/index.tsx
+++ b/src/hooks/use-fetch-wpcom-sites/index.tsx
@@ -233,8 +233,9 @@ export const useFetchWpComSites = ( connectedSiteIdsOnlyForSelectedSite: number[
                [ rawSyncSites, memoizedConnectedSiteIds ]
        );
        return {
-               syncSites: syncSitesWithSyncSupportForSelectedSite,
+               syncSites: [],
                isFetching: isFetchingSites.current,
                refetchSites: fetchSites,
        };
```
* Navigate to `Sync` tab
* Click on `Publish site` or `Pull site`
* See the loading state on either of the buttons (this is new because instead of fetching the WP.com sites in the sync modal and seeing the loading state there right away, we are fetching them first and then deciding which modal to fetch)
* Confirm that you can see the modal for no sites appear
* Remove the diff that you applied
* Click on the button again
* Confirm that you see the sync modal
* Finally, navigate to a Studio site which already has some sites connected
* Click on `Connect another site` button
* Confirm that you can see the sync modal

## Pre-merge Checklist


katinthehatsite:
@sejas I am leaving this as draft for now, just wanted to get your thoughts on design and copy for now. What do you think?

sejas:
Sounds good! Thanks for working in this issue. I'll give it an early review.

katinthehatsite:
@sejas One more question - I guess with the new design, we should now display this modal when the user clicks on either of those buttons, right?

Screenshot 2025-11-07 at 5 15 28 PM

sejas:
Yes! both buttons will trigger the same modal if the user doesn't have any WPcom site.

sejas (commented):
> firstly, we should not really use the logo in the title, it does not work well as it is too large to use with the default title size and it does not work with that type

@katinthehatsite , I think the default title is fine without any customizations, and we can move the WPcom logo to the description. 👍

Could you rebase the branch and move it to ready for review?

katinthehatsite:
> @katinthehatsite , I think the default title is fine without any customizations, and we can move the WPcom logo to the description. 👍
Could you rebase the branch and move it to ready for review?

Awesome, on it 👍

katinthehatsite:
@sejas Feel free to review, I will test closer tomorrow once again before opening it for the team testing.

src/modules/sync/index.tsx
```diff
@@ -218,9 +230,12 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
 						<ConnectButton
 							variant="primary"
 							connectSite={ handleLaunchSite }
-							disabled={ isAnySiteSyncing }
+							disabled={ isAnySiteSyncing || pendingModalMode !== null }
+							isBusy={ pendingModalMode === 'push' }
 							tooltipText={
-								isAnySiteSyncing
+								pendingModalMode === 'pull'
+									? __( 'Please wait for the current operation to finish.' )
```
katinthehatsite, line 237: Technically, we don't need this tooltip here but I added it in case it is not clear for the user. We can remove it if not needed

src/modules/sync/components/sync-sites-modal-selector.tsx
```diff
@@ -89,11 +89,9 @@ export function SyncSitesModalSelector( {
 						</div>
 					) }
 
-					{ ! isLoading && isEmpty && (
+					{ ! isLoading && isEmpty && searchQuery && (
 						<div className="flex justify-center items-center h-full">
-							{ searchQuery
-								? sprintf( __( 'No sites found for "%s"' ), searchQuery )
-								: __( 'No sites found' ) }
```
katinthehatsite, line 96: I removed this fallback since if there are no sites found without searchQuery, we should always get the modal with `NoWpcomSitesModal`

src/modules/sync/components/create-button.tsx
```diff
@@ -27,15 +29,17 @@ export const CreateButton = ( {
 			text={ __( 'Creating a site requires an internet connection.' ) }
 			icon={ offlineIcon }
 			placement="top-start"
+			className={ className }
```
nightnei: We have className here and below. Do we need it in both places? Can we have className and tooltipClassName for example?

nightnei (approved):
Left just 1 small comment 

LGTM and works as expected 👍


Screenshot 2025-11-19 at 14 11 40

src/modules/sync/components/create-button.tsx
katinthehatsite: I removed it in https://github.com/Automattic/studio/pull/2038/commits/98b7e842806c1b40a90cc8ddd924e98018ba1246

Thanks for testing! I added this unintentionally while checking adjusting the one below 😅

PR #2041, opened 2025-11-07
Add toggle to settings modal to un-/install CLI

fredrikekelund:
## Related issues

- Fixes STU-664

## Proposed Changes

This PR makes the option to install the Studio CLI more easily available by exposing a toggle in the settings modal:

Screenshot 2025-11-07 at 15 29 31

From a technical perspective, this PR does the following:

- Implements the new UI
- Adds RTK Query hooks for querying whether the CLI is installed
- Adds RTK Query mutations for installing and uninstalling the CLI
- Adds a procedure for uninstalling the CLI on macOS and Windows (we didn't have this previously)
- On Windows, we no longer automatically install the CLI when the app launches
- Tweaks the script that executes Studio CLI to also support running in development node (using the system level `node` executable and a different location for the CLI JS bundle)

## Testing Instructions

1. Open the settings modal
2. Ensure that the `Save` button is only enabled when you've changed the value of any of the settings
3. Change the install toggle for Studio CLI
4. Save the settings
5. Open your terminal and ensure that the `studio` command is available of unavailable (depending on which action you just took)
6. Open the settings modal again and flip the toggle back
7. Open your terminal and ensure that the `studio` command is available of unavailable (depending on which action you just took)

## Pre-merge Checklist


src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -1,63 +1,82 @@
...
-	};
+	const localeSelection = dirtyLocale ?? savedLocale ?? 'en';
+	const editorSelection = dirtyEditor ?? editor ?? 'vscode';
+	const terminalSelection = dirtyTerminal ?? terminal ?? 'terminal';
+	const isCliInstalledSelection = dirtyIsCliInstalled ?? isCliInstalled ?? false;
 
-	const hasChanges =
-		locale !== savedLocale || currentEditor !== editor || currentTerminal !== terminal;
+	const hasChanges = [
+		[ dirtyLocale, savedLocale ],
```
nightnei, line 62: I haven't seen prefix `dirty` for such cases. Just wonder - did you noticed that it's used somewhere outside YOLO/A8C? We can make it an agreement for such cases.

src/modules/cli/lib/installation/windows.ts
```diff
@@ -91,13 +92,47 @@ const installProxyBatFile = async () => {
...
+		const currentPath = await getPathFromRegistry();
+		const newPath = currentPath
+			.split( ';' )
+			.filter( ( item ) => item.trim().toLowerCase() !== unversionedBinDirPath.toLowerCase() )
+			.join( ';' );
+
+		await setPathInRegistry( newPath );
+	} catch ( error ) {
+		Sentry.captureException( error );
+		console.error( 'Failed to uninstall CLI', error );
```
nightnei: Not critical, but WDYT about throwing error outside to render it instead of console.log?

src/modules/cli/lib/ipc-handlers.ts
```diff
@@ -0,0 +1,70 @@
+import { dialog } from 'electron';
```
nightnei, line 1: Looking at `installation` folder, it seems that we could go with factory class and automatically create instalce of it depending on OS, as a result we coudl avoid such things as [if ( process.platform !== 'win32' ) {](https://github.com/Automattic/studio/pull/2041/files#diff-50150ccd68701679f16e4bc8a9b4024e5b4d9cc9a8fb7f077e55600ee978f3caR132) here, unify function names, less conditions, etc. I see that we export only `install/uninstall/is` functions from there, so we won't have Liskov substitution principle issues and it will be more structured and easy in the future to add other OS if we need.

No action needed, just wondering your opinion as you know the codebase better and I would like to understand :)

nightnei (approved):
Looks great and works as expected 👍 Just left a few comments which I wonder.

wojtekn:
@fredrikekelund I see it follows the design, but shouldn't the radio button be placed on the left in LTR language? It's a bit weird UI now.

bcotrim (approved):
Works as described
LGTM 👍

fredrikekelund:
> I see it follows the design, but shouldn't the radio button be placed on the left in LTR language? It's a bit weird UI now.

My instinct would be to place the toggle on the right hand side for LTR languages (and flipped in RTL languages). Given that this also matches the design from Marina, I will land this PR as is, and we can iterate on it later if needed.

wojtekn:
> My instinct would be to place the toggle on the right hand side for LTR languages (and flipped in RTL languages). Given that this also matches the design from Marina, I will land this PR as is, and we can iterate on it later if needed.

Yes, but design is not always correct. Quickly looking, I haven't found any place in WordPress core or WordPress.com that shows UI elements like checkbox, radio button or toggle on the right in LTR language.

Screenshot 2025-11-21 at 14 02 55

Screenshot 2025-11-21 at 14 02 35

Screenshot 2025-11-21 at 14 02 11

Screenshot 2025-11-21 at 14 01 56

wojtekn:
Update: I found some, as Jetpack uses that approach for toggles that are displayed in the box heading:

Screenshot 2025-11-21 at 14 05 32

However, in forms it follows the standard convention:

Screenshot 2025-11-21 at 14 06 35

fredrikekelund:
I don't think the wp-admin design language should be our guide, but seeing your examples from the hosting dashboard, I take your point, @wojtekn. I'll make that change.

bin/studio-cli.sh
```diff
@@ -1,13 +1,19 @@
 #!/bin/sh
 
-# This script is assumed to live in `/Applications/Studio.app/Contents/Resources/bin/studio-cli.sh`
+# The default assumption is that this script lives in `/Applications/Studio.app/Contents/Resources/bin/studio-cli.sh`
 CONTENTS_DIR=$(dirname "$(dirname "$(dirname "$(realpath "$0")")")")
 ELECTRON_EXECUTABLE="$CONTENTS_DIR/MacOS/Studio"
 CLI_SCRIPT="$CONTENTS_DIR/Resources/cli/main.js"
```
wojtekn, line 6: It's a bit out of scope here, but as you are making more improvements around installing/uninstalling, I would like ask - what do you think about unsetting NODE_OPTIONS in the wrapper to fix the warning that can appear in console when it's set for the terminal session?

src/modules/cli/lib/installation/macos.ts
```diff
@@ -65,9 +66,14 @@ export async function installCLIOnMacOSWithConfirmation() {
 	}
 }
 
+export async function isCliInstalled() {
+	const currentSymlinkDestination = await getCurrentSymlinkDestination();
```
wojtekn: This check causes marking CLI as not installed when I install it from the built app and run the app in dev mode.

Should we adjust it to treat it as installed in both cases? Or at least, always treat the one from production build as installed, and dev link as uninstalled, so user running production build could install it again, overwriting any existing dev mode link?

fredrikekelund: There's some nuance to this issue. For context, installing the CLI from the Studio production build yields different results than installing it from the development build. This is why I added the warning modals in `src/modules/cli/lib/installation/index.ts` when running the app in development mode.

So, why is there a difference to begin with? This is because the symlink in the `$PATH` directory points to a location determined by Electron (via `app.getPath( 'exe' )`), which returns different values depending on whether the app is packaged. 

We could indeed add some custom logic that checks the typical production return value of `app.getPath( 'exe' )` in the development version of the app.

After mulling it over, I think this is a good idea. It doesn't affect end users, but it reduces the likelihood that developers will unintentionally affect their environment.

bin/studio-cli.sh
fredrikekelund: This is fair 👍 I've made the change

wojtekn: Thanks!

src/modules/cli/lib/installation/macos.ts
wojtekn: Thanks for considering this. I think it will lead to better dev UX, since the UI won't have different states in dev and production modes.

The warning displayed when trying to save the option when running dev mode is clear.

Small nit: when the user decides to cancel the save, the settings dialog becomes closed. It could be a better UX to keep it open.

wojtekn (approved):
One last thought - would it make sense to have "Learn more" pointing to https://developer.wordpress.com/docs/developer-tools/studio/cli/ in the end of the field description? CC @ndiego

src/modules/cli/lib/ipc-handlers.ts
fredrikekelund: I take your point to be "let's use an interface" and have the platform-specific files export a class that satisfies the interface. Sure, I've made that change 👍

Please note, however, that the platform-specific classes still have the equivalent `if ( process.platform !== 'win32' ) {` as before. I find this appropriate, since they could still theoretically be consumed on the wrong platform.

fredrikekelund:
> would it make sense to have "Learn more" pointing

I think so 👍 Added

bin/studio-cli.sh
fredrikekelund: For future reference, Wojtek was referring to the warning described in https://github.com/Automattic/studio/pull/2128

PR #2058, opened 2025-11-11
Add multi-worker support to the beta features menu.

lynnbgriffith:
This exposes the experimental multi-worker support in Playground via the beta features menu. You should notice a significant speedup complex pages like the site editor with it enabled. Currently, I'm manually getting the CPU core count and creating n - 1 workers. I believe Playground is supposed to do that detection itself, but when I just pass `experimentalMultiWorker`, it runs with `true` as the number of workers and doesn't actually create multiple workers. Happy to drop the detection of core count in Studio if that can be done in Playground CLI itself, though.

@brandonpayton


fredrikekelund (approved):
This works as expected 👍 The performance gains are massive. This will be a real treat for our users!

I would like us to discuss a strategy for enabling multi-worker support by default, but we can do that after landing this PR.

@griffbrad, I will land this PR once `trunk` is unblocked (we are currently waiting on the 1.6.4 release)

lynnbgriffith:
Sounds good! This is safe to merge as-is, but I wanted to point out that @brandonpayton merged this PR in Playground which should make my own CPU core count in this PR unnecessary:
https://github.com/WordPress/wordpress-playground/pull/2883

We'd just need to ensure Studio is using a release with that change and then this PR could be simplified a bit more.

fredrikekelund:
@griffbrad, thanks for pointing that out. The changes in https://github.com/WordPress/wordpress-playground/pull/2883 might support default values for `--experimental-worker-count` in some cases where the Playground CLI is called directly from the terminal (I say "might support" because I think we should be using the `default` option instead of `coerce` with yargs to make it effective). However, Studio imports and calls the `runCLI` function directly, and in this case, Playground doesn't provide a default config value.

cc @brandonpayton

PR #2065, opened 2025-11-13
Update lock file to fix vulnerabilities

nightnei:
## Proposed Changes
We are updating lock file to fix vulnerabilities which are not a breaking changes.

<table>
<tr>
<td>BEFORE
<td>Screenshot 2025-11-13 at 17 04 21
</tr>
<tr>
<td>AFTER
<td>Screenshot 2025-11-13 at 17 04 53
</tr>
</table>

## Testing Instructions
1. Run Studio
2. Assert that it works

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2355,12 +2355,10 @@
 			}
 		},
 		"node_modules/@babel/runtime": {
-			"version": "7.25.7",
-			"resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.25.7.tgz",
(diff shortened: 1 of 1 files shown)


wojtekn (approved):
Could we update corresponding entries in the `package.json` to make the "required" version consistent with "installed"?

nightnei:
> Could we update corresponding entries in the package.json to make the "required" version consistent with "installed"?

Theoretically - yes, but I updated it with npm audit, which didn't update package.json, so I would keep it as is, w/o manual additional changes.

PR #2083, opened 2025-11-18
Fix connected sites when logging in

epeicher:
## Related issues

- Fixes STU-944

## Proposed Changes

- Adds an event listener to listen for the `user-data-updated` event. 
- Updates the connected sites when the user is logged in.

## Testing Instructions

1. Apply this branch and run `npm start`
2. Select a site with connected sites or connect a few sites
3. Log out
5. Restart the app or press `cmd + r` if you are in dev mode
7. Log in again
9. Verify that connected sites are displayed

## Pre-merge Checklist


katinthehatsite (approved):
This is so bizarre, I was able to reproduce this on trunk. 

The fix worked for me but perhaps we can consider the suggestion from Copilot.

epeicher:
> The fix worked for me but perhaps we can consider the suggestion from Copilot.

Yes, I already considered that, and discarded it because it seemed unnecessary, but I am going to reconsider to avoid excessive reloadings as suggested

nightnei (approved):
LGTM and I confirm that in trunk we had the issue and it's fixed with the changes 👍

fredrikekelund:
I think we should take a different approach here. 

The user config watcher was implemented to notify the Studio app of changes made by the CLI. The Studio app already has all state in memory, so unless we are truly looking for changes made by another program (the CLI) to the file on disk, resorting to the user config watcher is an indication we need to rethink the state structure, IMO. 

The ideal solution here would be to implement `src/stores/sync/connected-sites-slice.ts` with RTK Query instead (using the [createApi](https://redux-toolkit.js.org/rtk-query/api/createApi) API). This would allow us to pass `authToken.id` as a query dependency and automatically revalidate the data whenever the auth details change. I'll take a stab at this and submit a PR to show what I have in mind.

epeicher:
> I think we should take a different approach here.
> 
> The user config watcher was implemented to notify the Studio app of changes made by the CLI. The Studio app already has all state in memory, so unless we are truly looking for changes made by another program (the CLI) to the file on disk, resorting to the user config watcher is an indication we need to rethink the state structure, IMO.
> 
> The ideal solution here would be to implement `src/stores/sync/connected-sites-slice.ts` with RTK Query instead (using the [createApi](https://redux-toolkit.js.org/rtk-query/api/createApi) API). This would allow us to pass `authToken.id` as a query dependency and automatically revalidate the data whenever the auth details change. I'll take a stab at this and submit a PR to show what I have in mind.

Thanks @fredrikekelund for the additional context. The changes followed similar patterns used by other components like [new-sites](https://github.com/Automattic/studio/blob/65b8969bccbb6a7a3f5c1afcc9eba7d91dfe3409/src/stores/new-sites-slice.ts#L55) and [snapshots](https://github.com/Automattic/studio/blob/65b8969bccbb6a7a3f5c1afcc9eba7d91dfe3409/src/stores/snapshot-slice.ts#L297), but I didn't realize they had some other reasons for that. Your approach make sense to me, I will review your changes with interest when they’re ready 😄

PR #2029, opened 2025-11-06
Studio: Update Studio deeplink URL scheme from `wpcom-local-dev` to `wp-studio`

gavande1:
## Related issues

- Fixes STU-936

## Proposed Changes

- Update Studio deeplink URL scheme from `wpcom-local-dev://` to `wp-studio://`. 
- Update docs

## Testing Instructions

- Check out this branch in Studio
- Run Studio. `npm start`
- Try to log in to the WordPress.com account
- Make sure you are logged in without any regression. 

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2093, opened 2025-11-19
Blueprints: Update validation

ivan-ottinger:
## Related issues

- Resolves STU-1007
- Related to STU-1008

## Proposed Changes

- extract out the blueprint validation logic into utility `validateBlueprintData` function - so it can be reused easily
- display validation error message in a dialog when incorrect blueprint is passed in a deeplink
- remove unnecessary second deeplink blueprint validation from the renderer process
- update and add related unit tests
- ensure the deeplink that opens blueprint from a linked URL gets the user to the "Add a site" step directly (instead of the previous "Start from a Blueprint" step)

ℹ️ Please note that this PR does not add validation to the handling of deeplinks that include the base64-encoded blueprint. This is because we will have another PR that will refactor and unify the logic that handles both deeplink types: STU-1008.

## Testing Instructions

1. Check out the PR and build the app with `npm install && npm start`.
2. Try to open the blueprint deeplinks provided here: https://jsbin.com/levaburate/2/edit?html,css,output.
  - First two should result in a dialog error message.
  - The third one should get you to the "Add a site" step and allow you to create a new site with it.
3. Try to manually create sites using blueprint files from 39a4e-pb. The results should be as expected - there should be no regressions.

## Pre-merge Checklist

### DIFF
--- a/src/lib/deeplink/handlers/add-site-blueprint-with-url.ts
+++ b/src/lib/deeplink/handlers/add-site-blueprint-with-url.ts
@@ -4,6 +4,7 @@ import * as Sentry from '@sentry/electron/main';
 import { __ } from '@wordpress/i18n';
 import fs from 'fs-extra';
 import { sendIpcEventToRenderer } from 'src/ipc-utils';
+import { validateBlueprintData } from 'src/lib/blueprint-features';
 import { download } from 'src/lib/download';
 import { getMainWindow } from 'src/main-window';
 
@@ -71,6 +72,24 @@ export async function handleAddSiteWithBlueprint( urlObject: URL ): Promise< voi
 
 	try {
 		await download( decodedUrl, blueprintPath, false, 'blueprint' );
+
+		const blueprintJson = await fs.readJson( blueprintPath );
+		const validation = await validateBlueprintData( blueprintJson );
+
+		if ( ! validation.valid ) {
+			await fs.remove( blueprintPath ).catch( () => {
+				// Ignore cleanup errors
+			} );
+
+			await dialog.showMessageBox( mainWindow, {
+				type: 'error',
+				message: __( 'Blueprint validation failed' ),
+				detail: validation.error || __( 'Invalid Blueprint format' ),
+				buttons: [ __( 'OK' ) ],
+			} );
+			return;
+		}
+
 		await sendIpcEventToRenderer( 'add-site-blueprint-from-url', { blueprintPath } );
 	} catch ( error ) {
 		console.error( 'Failed to download blueprint from deeplink:', error );
--- a/src/lib/deeplink/handlers/add-site-blueprint-with-url.ts
+++ b/src/lib/deeplink/handlers/add-site-blueprint-with-url.ts
@@ -11,8 +11,8 @@ import { getMainWindow } from 'src/main-window';
 /**
  * Handles the add-site deeplink callback.
  * This function is called when a user clicks a deeplink like:
- * - wpcom-local-dev://add-site?blueprint_url=<encoded-url>
- * - wpcom-local-dev://add-site?blueprint=<base64-encoded-json>
+ * - wp-studio://add-site?blueprint_url=<encoded-url>
+ * - wp-studio://add-site?blueprint=<base64-encoded-json>
  *
  * It either downloads the blueprint from the URL or decodes the base64 blueprint,
  * and opens the Add Site modal with the blueprint pre-filled.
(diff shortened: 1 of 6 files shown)


src/lib/deeplink/handlers/add-site-blueprint-with-url.ts
```diff
@@ -71,6 +72,24 @@ export async function handleAddSiteWithBlueprint( urlObject: URL ): Promise< voi
 
 	try {
 		await download( decodedUrl, blueprintPath, false, 'blueprint' );
+
+		const blueprintJson = await fs.readJson( blueprintPath );
+		const validation = await validateBlueprintData( blueprintJson );
+
+		if ( ! validation.valid ) {
+			await fs.remove( blueprintPath ).catch( () => {
+				// Ignore cleanup errors
```
gcsecsey, line 81: I was thinking maybe we can log this, but as this is a temporary file, I think this is fine to skip. 👍

gcsecsey (approved):
Thanks for refactoring the validation @ivan-ottinger, and for updating the tests! I tested both blueprints from the drive and a couple from the [blueprints gallery](https://github.com/WordPress/blueprints/tree/trunk/blueprints), and I'm getting the same validation issues reported as on trunk.

| Invalid URL | Invalid Blueprint | Valid Blueprint |
|--------|--------|--------|
| Image | Image | Image |

gavande1 (commented):
@ivan-ottinger, I believe we are validating the blueprint twice as per the suggested changes in this PR. Is it intended, or am I missing something? Do you think we should refactor that? 

https://github.com/Automattic/studio/blob/0ea9e3098f2f2555fd1f59fc747fa83797004d18/src/modules/add-site/hooks/use-blueprint-deeplink.ts#L143-L150

gavande1 (approved):
I have tested this and it works as expected. I have left a comment about duplicate validation. Otherwise, the changes look good to me.

Image

ivan-ottinger:
Thank you for your reviews and testing, Gergely and Rahul!

> @ivan-ottinger, I believe we are validating the blueprint twice as per the suggested changes in this PR. Is it intended, or am I missing something? Do you think we should refactor that?

That's a good catch, @gavande1! This second validation is no longer needed - since we validate earlier in the flow, before even the modal has a chance to open.

I have addressed that in https://github.com/Automattic/studio/pull/2093/commits/6560c930e0ca670adb66a1e846d1f21750a6b0ab and the PR is ready for another review round. 🙂 

I am looking forward to the next simplifications related to STU-1008 in the next PR.

src/modules/add-site/hooks/use-blueprint-deeplink.ts
```diff
@@ -94,11 +90,33 @@ export function useBlueprintDeeplink(
 			if ( isAnySiteProcessing ) {
 				return;
 			}
-			setPendingBlueprintPath( blueprintPath );
-			setInitialNavigatorPath( '/blueprint' );
-			openModal();
+			try {
```
ivan-ottinger, line 93: I have moved the blueprint processing logic from the original `useEffect` here (minus the extra validation that is not needed).

It would all work with the `useEffect` in place, but since it is not necessary, I would like to simplify the code and reduce the complexity.

gcsecsey (approved):
Thanks for removing the extra validation, the updated version LGTM and still works well in all scenarios. 🙌

src/modules/add-site/hooks/use-blueprint-deeplink.ts
gavande1: Great. Thanks for the explanation.

src/lib/deeplink/tests/add-site.test.ts
```diff
@@ -30,8 +34,14 @@ describe( 'handleAddSiteWithBlueprint', () => {
 		focus: jest.fn(),
 	} as unknown as BrowserWindow;
 
+	const createBlueprintUrl = ( blueprintUrl: string ) => {
+		const encodedUrl = encodeURIComponent( blueprintUrl );
+		return new URL( `wpcom-local-dev://add-site?blueprint_url=${ encodedUrl }` );
```
gavande1: @ivan-ottinger, I merged #2029, so we need to rebase this branch.

ivan-ottinger: Thanks for the heads-up, Rahul. I have rebased the branch and updated remaining URLs in https://github.com/Automattic/studio/pull/2093/commits/ff51c4672431714a1c0d2dd38878e9b983839c19.

gavande1 (approved):
Thanks @ivan-ottinger for addressing the feedback. LGTM 👍

PR #2076, opened 2025-11-17
Clean up unused dependencies in an attempt to speed up Windows CI runs

fredrikekelund:
## Related issues

Fixes STU-1006

## Proposed Changes

This PR contains a number of attempts at speeding up the `install-node-dependencies` step in Windows CI runs. Ultimately, I wasn't able to have a big impact, but the changes still make sense from a quality perspective, and I recommend landing them for that reason. Here's what's changed:

1. Add a `package-lock.json` to `./cli` and run `npm ci` in that folder instead of `npm install --prefix cli` in the `postinstall` script
2. Don't install the `scripts/compare-perf` dependencies on Buildkite, since we don't need to (we run that workflow with GitHub Actions)
3. Remove unused or unnecessary dependencies

<del>In my initial testing, this reduced the time it took to run unit tests on Windows by ~2,5min 👍</del><ins>These results were not reliable</ins>

## Testing Instructions

CI should pass. 

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2095, opened 2025-11-19
Implement RTK Query approach for connectedSites

fredrikekelund:
## Related issues

- Related to https://github.com/Automattic/studio/pull/2083#issuecomment-3552639410
- Related to STU-944

## Proposed Changes

In #2083, we implemented a manual refetch of the list of connected sites when users log in to Studio. This prompted me to look into refactoring the state tree to use RTK Query instead, because with RTK Query, queries are automatically re-evaluated whenever a dependency changes (e.g., the currently logged-in user). 

This PR does the following:
- Adds `src/stores/sync/connected-sites.ts` with an RTK Query API (query hooks and mutation hooks)
- Removes `src/stores/sync/connected-sites-hooks.ts` and `src/stores/sync/connected-sites-slice.ts`
- Updates all the consumers of those files to use the new approach instead

## Testing Instructions

Do a full smoke test of the sync tab, with and without the "Streamline Onboarding" feature flag enabled. 

Also try logging out and then logging back in to WordPress.com and ensure that the list of connected sites refreshes as expected. 

## Pre-merge Checklist


src/index.html
epeicher, line 1: Just curious, this was not required originally, right?

src/modules/sync/index.tsx
```diff
@@ -125,32 +123,30 @@ export type OpenSitesSyncSelector = ( options?: { disconnectSiteId?: number } )
...
 export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails } ) {
 	const { __ } = useI18n();
-	const dispatch = useAppDispatch();
-	const isModalOpen = useRootSelector( connectedSitesSelectors.selectIsModalOpen );
-	const reduxModalMode = useRootSelector( connectedSitesSelectors.selectModalMode );
-	const { connectedSites } = useConnectedSitesData();
-	const { syncSites, isFetching, refetchSites } = useSyncSitesData();
-	const { connectSite, disconnectSite } = useConnectedSitesOperations();
+	const [ isModalOpen, setIsModalOpen ] = useState( false );
+	const [ modalMode, setModalMode ] = useState< SyncModalMode | null >( null );
```
epeicher: Nit: Have you considered using an object for the two states, since they are directly related?

epeicher (approved):
Thanks for working on this @fredrikekelund! I really like this refactoring, as the api hooks should "do their work" and refresh the data, so this is 💯 

I have tested it, and it works as expected. I have tested the connected sites and the disabling of the import/export buttons, and I have not found any issues.

I have left one question and one minor comment, but accepting this as they are non-blocking. Please remember to merge it with trunk as there has been some changes in the Sync tab. LGTM! :shipit:

src/modules/sync/index.tsx
epeicher: Adding here that this local state was added to Redux to allow for these changes https://github.com/Automattic/studio/pull/2025, please let me know  if you have any questions about it after the merge

fredrikekelund: I moved the modal open state (and mode) back to Redux after seeing your changes in #2025, @epeicher 👍

src/index.html
fredrikekelund: Precisely 👍 I should have left a comment about this straight away, but yes, this is more of a follow-up to #2076, because I realized that the file we actually use is `index.html` in the root directory

src/components/tests/content-tab-assistant.test.tsx
fredrikekelund, line 1: The changes to this file are unrelated to the core purpose of this PR. I just realized that this test calls `testActions.resetState` in a `beforeEach` function, and this action has no effect unless we also apply `testReducer`

src/modules/sync/index.tsx
```diff
@@ -199,11 +184,10 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
 		}
 
 		if ( mode === 'push' || mode === 'pull' ) {
-			dispatch( connectedSitesActions.setModalMode( mode ) );
```
fredrikekelund, line 202: This is the only place I could identify where `setModalMode` is called with a non-null argument. I went ahead and replaced it with `openModal`, because I believe that will have the same effect. Would appreciate if you could confirm, @epeicher

electron.vite.config.ts
```diff
@@ -1,5 +1,4 @@
 import { resolve } from 'path';
-import { pathToFileURL } from 'url';
```
fredrikekelund, line 2: Unrelated to the core changes in this PR. This function was unused.

fredrikekelund:
Thanks for the review, @epeicher 🙏 I've merged trunk, resolved the conflicts, and opted to move the modal open state back to Redux to support your changes in #2025. I'm relatively confident that these changes shouldn't have an impact on the flows you're currently working on, but I would appreciate if you could take it for another test run to confirm.

epeicher (approved):
Thanks @fredrikekelund for merging with the latest changes. I have added the commit 0931b37 to fix a small issue with the mode being reset. I have also logged a task, STU-1035, that I discovered while testing this, but it is unrelated to these changes. All in all, I have tested it by enabling the feature flag, and it works as expected, nice refactor. LGTM! :shipit: 

#### Publish site header button
https://github.com/user-attachments/assets/d3e03004-67f6-4490-b299-e77f827b52a5

#### Publish site on Sync tab

https://github.com/user-attachments/assets/5e394ace-c426-4133-a402-95b1b1555b63

sejas (approved):
Great refactor, thanks for simplifying the logic and also removing the `user-data-updated` listener.

I confirm the connected sites work correctly when logging out and logging in.

https://github.com/user-attachments/assets/662a5557-4715-41fa-8c91-21e3cfcf59c5

The connected sites also work 

https://github.com/user-attachments/assets/56cd1afc-12b5-4908-a038-2117625fb62b

PR #2096, opened 2025-11-19
CLI: Studio start site

bcotrim:
## Related issues

- Fixes STU-950

## Proposed Changes

- `studio site start --path <site-path>` command to start local WordPress sites
- `--skip-browser` flag to skip opening browser after starting

## Testing Instructions

1. Add a site to Studio (via Studio app)
2. Run `npm run cli:build && node dist/cli/main.js site start --path /path/to/site`
3. Verify the site starts and browser opens to wp-admin with auto-login
4. Test `--skip-browser` flag: `node dist/cli/main.js site start --path /path/to/site --skip-browser`
5. Verify browser does not open but site is still running
6. Test with custom domain site
7. Verify proxy starts, SSL certs generate, and hosts file is updated

## Pre-merge Checklist


cli/commands/site/start.ts
```diff
@@ -23,7 +25,20 @@ async function startProxyIfNeeded( logger: Logger< LoggerAction > ) {
 	}
 }
 
-export async function runCommand( siteFolder: string ): Promise< void > {
+async function openSiteInBrowser( site: SiteData ) {
+	const siteUrl = getSiteUrl( site );
+	try {
+		const wpAdminUrl = `${ siteUrl }/wp-admin/`;
+		const adminUrl = `${ siteUrl }/studio-auto-login?redirect_to=${ encodeURIComponent(
```
fredrikekelund:
```suggestion
		const autoLoginUrl = `${ siteUrl }/studio-auto-login?redirect_to=${ encodeURIComponent(
```

cli/commands/site/start.ts
fredrikekelund, line 1: The changes to this file are super clear 👍

cli/lib/pm2-manager.ts
```diff
@@ -113,61 +132,125 @@ function getProxyProcessEnv() {
...
+				return;
+			}
+
+			resolve( {
+				name: process.name,
+				pmId: process.pm_id,
+				status: process.pm2_env.status,
+			} );
+		} );
+	} );
```
fredrikekelund: Looks like this function is unused, so we can remove it.

cli/lib/pm2-manager.ts
fredrikekelund, line 1: The changes to this file are clear 👍

cli/wordpress-server-child.ts
fredrikekelund, line 1: The logic in this file is mostly clear, but I have two comments:
1. Let's use a different name. "daemon" implies that there's a single instance, IMO. I would call it `wordpress-server-child` or something along those lines. 
2. The IPC stuff in this file is relatively clear, if I understand it correctly that it only receives messages from the parent. Two questions on this topic:
  2.1 Should we use the `pm2` APIs to receive and send messages from here, or does that complicate things without any clear benefit?
  2.2 Given how we process a single IPC message type, this seems like a great place to start adding zod validation to the IPC messages.

cli/package.json
```diff
@@ -8,8 +8,14 @@
...
-		"pm2": "^5.3.0",
-		"http-proxy": "^1.18.1"
+		"pm2": "^6.0.13",
+		"http-proxy": "^1.18.1",
+		"@php-wasm/node": "^3.0.22",
+		"@php-wasm/scopes": "^3.0.22",
+		"@php-wasm/universal": "^3.0.22",
+		"@wp-playground/cli": "^3.0.22",
+		"@wp-playground/blueprints": "^3.0.22",
+		"@wp-playground/wordpress": "^3.0.22"
```
fredrikekelund: Do we really need to add these to this `package.json`..? Will the CLI build work if we just traverse the file system and find these modules in the root project directory instead?

cli/lib/wordpress-server-manager.ts
```diff
@@ -0,0 +1,297 @@
...
+	const processName = getProcessName( siteId );
+	return stopProcess( processName );
+}
+
+/**
+ * Get status of a WordPress server
+ */
+export async function getServerStatus( siteId: string ): Promise< ProcessDescription | null > {
+	const processName = getProcessName( siteId );
+	return getProcessStatus( processName );
```
fredrikekelund:
```suggestion
	return isProcessRunning( processName );
```

cli/lib/wordpress-server-manager.ts
fredrikekelund, line 1: This file is mostly clear and easy to follow, except I don't understand which parts of the IPC communication are used and which are not. Let's remove any unused components, then use Zod schemas to validate the structure of the message objects. This should make the code easier to reason about.

cli/lib/wordpress-server-manager.ts
```diff
@@ -0,0 +1,297 @@
...
+					callback( packet.raw.message );
+				}
+			};
+
+			bus.on( 'process:msg', consoleHandler );
+		} )
+		.catch( ( error ) => {
+			console.error( 'Failed to set up console message handler:', error );
+		} );
+}
```
fredrikekelund:
```suggestion
async function setupConsoleMessageHandler(
	pmId: number,
	callback: ( message: string ) => void
): Promise< void > {
	try {
		const bus = await getPm2Bus();

		const consoleHandler = ( packet: any ) => {
			if (
				packet?.process?.pm_id === pmId &&
				packet?.raw?.type === 'console-message' &&
				packet?.raw?.message
			) {
				callback( packet.raw.message );
			}
		};

		bus.on( 'process:msg', consoleHandler );
	} catch ( error ) {
		console.error( 'Failed to set up console message handler:', error );
	}
}
```

fredrikekelund (commented):
Stellar work, @bcotrim! 👍 We need to over this and tweak or clarify a couple of things, but the fundamental flow is there and works as expected. Let's chat more about the details in a huddle 🙂

cli/wordpress-server-child.ts
```diff
@@ -0,0 +1,187 @@
...
+			const validationResult = MessageSchema.safeParse( message );
+			if ( ! validationResult.success ) {
+				throw new Error(
+					`Invalid message format: ${ validationResult.error.errors
+						.map( ( e ) => e.message )
+						.join( ', ' ) }`
+				);
+			}
+
+			const validMessage = validationResult.data;
```
fredrikekelund:
```suggestion
	process.on( 'message', async ( packet: unknown ) => {
		try {
			const validMessage = MessageSchema.parse( packet );
```
We can simplify this now that we have zod validation.

cli/wordpress-server-child.ts
```diff
@@ -0,0 +1,187 @@
...
+					throw new Error( `Unknown message type: ${ validMessage.type }` );
+			}
+
+			if ( process.send && validMessage.id !== undefined ) {
+				const response = { id: validMessage.id, type: validMessage.type, result };
+				const responseValidation = MessageSchema.safeParse( response );
+				if ( responseValidation.success ) {
+					process.send( response );
+				}
+			}
```
fredrikekelund:
```suggestion
			if ( validMessage.id !== undefined ) {
				const response: ResponseMessageResult = {
					id: validMessage.id,
					topic: validMessage.topic,
					result,
				};
				process.send!( response );
			}
```
No need to parse with zod when TS can statically analyze types. Also, we know that `process.send` is defined in this context because we wrap the entire event handler in an `if ( process.send ) {` statement. 

Also, see my comment in `cli/lib/types/wordpress-server.ts` for the `ResponseMessageResult` definition.

cli/wordpress-server-child.ts
```diff
@@ -0,0 +1,187 @@
...
+		} catch ( error ) {
+			const messageId =
+				packet.type === 'process:msg' && packet.data?.id !== undefined ? packet.data.id : packet.id;
```
fredrikekelund: In which cases does the packet ID come in the `packet.id` prop and `packet.data.id` respectively?

cli/wordpress-server-child.ts
```diff
@@ -0,0 +1,187 @@
...
+					id: messageId,
+					type: 'error',
+					error: error instanceof Error ? error.message : String( error ),
+					errorStack: error instanceof Error ? error.stack : undefined,
+				};
+				const errorValidation = MessageSchema.safeParse( errorResponse );
+				if ( errorValidation.success ) {
+					process.send( errorResponse );
+				}
+			}
```
fredrikekelund:
```suggestion
			console.log( '[WordPress Server Child] Error:', error );

			const minimalMessageSchema = z.object( { id: z.number() } );
			const message = minimalMessageSchema.safeParse( packet );

			if ( message.success ) {
				const errorResponse: ResponseMessageError = {
					id: message.data.id,
					topic: 'error',
					error: error instanceof Error ? error.message : String( error ),
					errorStack: error instanceof Error ? error.stack : undefined,
				};
				process.send!( errorResponse );
			}
```
I suggest doing something like this:
- Logging the error as standard seems sensible
- We might as well make use of zod
- See my comment in `cli/lib/types/wordpress-server.ts` for the `ResponseMessageError` definition.

cli/lib/types/wordpress-server.ts
```diff
@@ -0,0 +1,99 @@
...
+	type: z.string(),
+	data: z
+		.object( {
+			config: ServerConfigSchema.optional(),
+		} )
+		.optional(),
+	result: z.unknown().optional(),
+	error: z.string().optional(),
+	errorStack: z.string().optional(),
+} );
```
fredrikekelund:
```suggestion
const messageBaseSchema = z.object( {
	id: z.number(),
} );

const serverConfigSchema = z.object( {
	siteId: z.string(),
	sitePath: z.string(),
	port: z.number(),
	phpVersion: z.string().optional(),
	wpVersion: z.string().optional(),
	absoluteUrl: z.string().optional(),
	adminPassword: z.string().optional(),
	siteTitle: z.string().optional(),
	siteLanguage: z.string().optional(),
	isWpAutoUpdating: z.boolean().optional(),
	blueprint: z.any().optional(), // Blueprint type is complex, allow any for now
} );

export type ServerConfig = z.infer< typeof serverConfigSchema >;

const startServerMessage = messageBaseSchema.extend( {
	topic: z.literal( 'start-server' ),
	data: z.object( {
		config: serverConfigSchema,
	} ),
} );

export const messageSchema = z.discriminatedUnion( 'topic', [ startServerMessage ] );
```
Here's how I think we should be validating incoming messages:
- One schema per message type
- Use a discriminated union based on the `type` prop. This means we can be confident about the message payload types. 
- I know this `ServerConfig` type doesn't match the one already defined in this file 1:1, because the `blueprint` type differs. This approach has the benefit of using a single source of truth for the type definition, though.
- Nit, but I also suggest using lowercase names for zod schemas. TS type names typically start with uppercase, and I think this is an easy way to distinguish between schemas and types.

cli/lib/types/wordpress-server.ts
```diff
@@ -0,0 +1,99 @@
...
+export interface Message {
+	id?: number;
+	type: string;
+	data?: {
+		config?: ServerConfig;
+	};
+	result?: unknown;
+	error?: string;
+	errorStack?: string;
+}
```
fredrikekelund:
```suggestion
type Message = {
	id: number;
};

export type RequestMessage = Message & {
	topic: string;
	data: unknown;
};

export type ResponseMessageError = Message & {
	topic: 'error';
	error: string;
	errorStack?: string;
};

export type ResponseMessageResult = Message & {
	topic: string;
	result: unknown;
};
```
Let's be more granular here. Something along these lines.

cli/wordpress-server-child.ts
```diff
@@ -0,0 +1,187 @@
...
+	} catch ( error ) {
+		server = null;
+		console.error( `[WordPress Server Child] Failed to start server:`, error );
+		throw error;
+	}
+}
+
+if ( process.send ) {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	process.on( 'message', async ( packet: any ) => {
```
fredrikekelund: Could we move this function definition outside the if statement to reduce indentation and make the logic easier to follow?

cli/lib/types/wordpress-server.ts
```diff
@@ -0,0 +1,99 @@
...
+		// Response message with id
+		z.object( {
+			type: z.string(),
+			id: z.number(),
+			result: z.unknown().optional(),
+			error: z.string().optional(),
+			errorStack: z.string().optional(),
+		} ),
+	] ),
+} );
```
fredrikekelund:
```suggestion
const readyPacketSchema = z.object( {
	topic: z.literal( 'ready' ),
} );

const activityPacketSchema = z.object( {
	topic: z.literal( 'activity' ),
} );

const resultPacketSchema = z.object( {
	topic: z.literal( 'result' ),
	id: z.number(),
	result: z.unknown(),
} );

const errorPacketSchema = z.object( {
	topic: z.literal( 'error' ),
	id: z.number(),
	error: z.string(),
	errorStack: z.string().optional(),
} );

export const packetSchema = z.object( {
	process: z.object( {
		pm_id: z.number(),
	} ),
	raw: z.discriminatedUnion( 'topic', [
		readyPacketSchema,
		activityPacketSchema,
		resultPacketSchema,
		errorPacketSchema,
	] ),
} );
```
This way we can make much better use of zod and be confident about the data types depending on which `type` prop the packet has.

cli/lib/wordpress-server-manager.ts
```diff
@@ -0,0 +1,252 @@
...
+	const bus = await getPm2Bus();
+
+	return new Promise( ( resolve, reject ) => {
+		const timeout = setTimeout( () => {
+			bus.off( 'process:msg', readyHandler );
+			reject( new Error( 'Timeout waiting for ready message from WordPress server child' ) );
+		}, PLAYGROUND_CLI_INACTIVITY_TIMEOUT );
+
+		const readyHandler = ( packet: unknown ) => {
+			const validationResult = PacketSchema.safeParse( packet );
```
fredrikekelund:
```suggestion
			const validationResult = packetSchema.safeParse( packet );
```

cli/lib/wordpress-server-manager.ts
```diff
@@ -0,0 +1,252 @@
...
+					};
+					if ( 'errorStack' in validPacket.raw && validPacket.raw.errorStack ) {
+						error.stack = validPacket.raw.errorStack;
+					}
+					reject( error );
+				} else {
+					resolve( 'result' in validPacket.raw ? validPacket.raw.result : undefined );
+				}
+			}
+		};
```
fredrikekelund, line 207:
```suggestion
		const responseHandler = ( packet: unknown ) => {
			const validationResult = packetSchema.safeParse( packet );
			if ( ! validationResult.success ) {
				cleanup();
				reject( validationResult.error );
				return;
			}

			const validPacket = validationResult.data;

			if ( validPacket.raw.topic === 'activity' ) {
				lastActivityTimestamp = Date.now();
				const tracker = activityTrackers.get( id );
				if ( tracker ) {
					tracker.lastActivityTimestamp = lastActivityTimestamp;
				}
				return;
			}

			if ( validPacket.raw.topic === 'error' ) {
				const error = new Error( validPacket.raw.error );
				if ( validPacket.raw.errorStack ) {
					error.stack = validPacket.raw.errorStack;
				}
				cleanup();
				reject( error );
			} else if ( validPacket.raw.topic === 'result' && validPacket.raw.id === id ) {
				cleanup();
				resolve( validPacket.raw.result );
			}
		};
```
A couple of things going on here:
- We consume the zod schema in a way where we are confident about the packet payload based on the `type` prop. 
- Packets with the `result` topic should also be verified by the message ID before resolving the promise. 
- Validation errors cause the promise to reject. I'm not sure if this is an ideal solution, but I think it's a good start. Without this, there will be a lot of manual inspection of logs and killing hanging processes during development.

cli/lib/wordpress-server-manager.ts
```diff
@@ -0,0 +1,252 @@
...
+async function sendMessage(
+	pmId: number,
+	type: string,
+	data: Message[ 'data' ]
+): Promise< unknown > {
+	const bus = await getPm2Bus();
+
+	return new Promise( ( resolve, reject ) => {
+		const id = nextMessageId++;
+		const message: Message = { id, type, data };
```
fredrikekelund:
```suggestion
async function sendMessage( pmId: number, topic: string, data: unknown ): Promise< unknown > {
	const bus = await getPm2Bus();

	return new Promise( ( resolve, reject ) => {
		const id = nextMessageId++;
		const message: RequestMessage = { topic, id, data };
```

cli/lib/pm2-manager.ts
```diff
@@ -79,8 +89,6 @@ async function connect(): Promise< void > {
 
 export async function startDaemon(): Promise< void > {
 	await connect();
-	// Keep connection open - subsequent operations will reuse it
-	// The isConnected flag prevents duplicate connections
 }
```
fredrikekelund: Nit, but since we already export `disconnect` directly, I suggest we do the same with `connect` and remove the `startDaemon`. Seems redundant right now.

cli/lib/wordpress-server-manager.ts
```diff
@@ -0,0 +1,252 @@
...
+				data: message,
+				topic: true,
+			},
+			( error ) => {
+				if ( error ) {
+					cleanup();
+					reject( error );
+				}
+			}
+		);
```
fredrikekelund:
```suggestion
		pm2.sendDataToProcessId( pmId, message, ( error ) => {
			if ( error ) {
				cleanup();
				reject( error );
			}
		} );
```
Let's use the `topic` property. pm2 requires this prop to be defined.

cli/wordpress-server-child.ts
```diff
@@ -0,0 +1,187 @@
...
+				};
+				const errorValidation = MessageSchema.safeParse( errorResponse );
+				if ( errorValidation.success ) {
+					process.send( errorResponse );
+				}
+			}
+		}
+	} );
+
+	process.send( { type: 'ready' } );
```
fredrikekelund:
```suggestion
	process.send( { topic: 'ready' } );
```
pm2 requires the `topic` prop when sending messages. Makes sense to use that prop in response messages, too.

cli/wordpress-server-child.ts
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		if ( config.adminPassword ) {
+			await setAdminPassword( server, config.adminPassword );
+		}
+	} catch ( error ) {
+		server = null;
+		console.error( `[WordPress Server Child] Failed to start server:`, error );
+		throw error;
+	}
```
fredrikekelund, line 138: It seems to me that we should explicitly kill the process if this happens, no? Maybe we should send an `error` response first, but leaving the process hanging seems like a bad idea.

cli/lib/pm2-manager.ts
```diff
@@ -113,61 +129,84 @@ function getProxyProcessEnv() {
...
 	return new Promise( ( resolve, reject ) => {
 		const processConfig: StartOptions = {
-			name: PROXY_PROCESS_NAME,
+			name: processName,
 			interpreter: process.execPath,
-			script: proxyDaemonPath,
+			script: scriptPath,
 			exec_mode: 'fork',
 			autorestart: true,
 			max_restarts: 5,
```
```suggestion
autorestart: false,
```
Let's change this. If a `wordpress-server-child` process fails to start the Playground server, the process should die. However, this config will cause pm2 to restart the process and leave it hanging indefinitely, since it needs the `start-server` message to actually initiate Playground.

fredrikekelund (commented):
Big improvement! 👍 I left several suggestions. I've also made those changes locally, and given the number of changes, it might actually be easiest if I push what I have.

fredrikekelund:
Recapping @bcotrim's and my discussion from Slack. Before landing this PR, Bernardo will:
- Make it so we log additional details to the console after a site has started – Site URL, admin username, admin password. We choose to log the admin password because the auto-login URL is so long that it will be cumbersome to print in the terminal, and if we didn't make the password easily accessible, there's a greater chance that users would change the password. Being a local site, they might pick a very weak password. The password continues to work on the preview site if they ever publish one of those, though, and that's why we think it's more of a security risk to not print the password in the terminal. 
- Store a `latestCliPid` prop for each site in the appdata file. We will later use this as a mechanism for querying the site's running state from Studio. 
- Add unit tests

fredrikekelund (approved):
Great work, @bcotrim 👍 The changes we discussed on Friday (https://github.com/Automattic/studio/pull/2096#issuecomment-3563533009) work as expected.

Let's remove the `registerSiteStartCommand` call from `cli/index.ts` before landing this PR, so that we don't ship unfinished work to end users. 

Also, cc @youknowriad for awareness of the PR.

PR #2103, opened 2025-11-20
Blueprints: Refactor deeplink handling to use consistent file-based approach; unused code removal

ivan-ottinger:
## Related issues

Resolves STU-1008

## Proposed Changes

The goal of this PR is to unify blueprint deeplinks handling and remove unnecessary code:

- remove `handleBlueprintFromBase64` callback → instead use `handleBlueprintFromUrl` for both blueprint deeplink types
- unify `blueprintPath` logic to use current timestamp for temporary json file name
- remove unused error state and related notice (we used it before to display errors in the renderer process, now the errors are caught earlier)
- remove unnecessary `initialNavigatorPath` state (we now directly jump to the "Add a site" step so this state is not needed anymore)
- rename `add-site-blueprint-from-url` to `add-site-with-blueprint`
- clean up related unit tests

## Testing Instructions

1. Check out the PR and build the app with `npm install && npm start`.
2. Try to open the blueprint deeplinks provided here: https://jsbin.com/levaburate/2/edit?html,css,output.
  - First two should result in a dialog error message.
  - The third one should get you to the "Add a site" step and allow you to create a new site with it.
3. Try to open the following Base64-encoded blueprint links:
 
This one should result in an error message:

```
wp-studio://add-site?blueprint=ewogICJuYW1lIjogIk1lb3dzeSIsCiAgInNwZWNpZXMiOiAiY2F0IiwKICAiaW52YWxpZCI6ICJ0aGlzIGlzIG5vdCBhIHZhbGlkIGJsdWVwcmludCIKICAibWlzc2luZyI6ICJjb21tYSBhYm92ZSIKfQo=
```
 
This one should work correctly and let you create a site with it (installs Astra theme):

```
[...]

PR #2104, opened 2025-11-20
Studio Assistant: Add "Open in Studio" button to Blueprint output

gcsecsey:
## Related issues

- Fixes STU-995

## Proposed Changes

- Add a utility to detect if the generated JSON is a Blueprint
- Show an "Open in Studio" button for Blueprints
- Redirect to the deeplink with the Blueprint content when the button is clicked

## Testing Instructions

- Open Studio and go to the Assistant tab on one of the sites
- Prompt it to generate a Blueprint to be used for a new site
- Try prompting for both simpler and more complicated setups, to test the Blueprint parsing at the later step:
  
  - Installing one plugin: 
    ```
    Help me start a site using Blueprints that installs WooCommerce
    ```
  - Installing a plugin, a theme, and executing some code:
    ```
    Help me start a site using Blueprints that:
     - Installs WooCommerce
     - Installs the Twenty Twenty-Three theme from WP.org
     - Adds 5 example products to WooCommerce
    ```
> [!NOTE]
> The Assistant can potentially generate invalid blueprints, where the plugins array is not flat, or the theme install step is invalid. It can take a couple of tries or re-prompts to get a semantically correct Blueprint generated.

- Scroll down to the bottom of the generated JSON
- Check that there is a new "Open in Studio" button rendered, and click it
- Check that you're dropped in the "Add a site" step with the blueprint loaded
- Check that the site creation proceeds without errors

| trunk | this branch |
|--------|--------|
| image | image |

Opening the Blueprint:
https://github.com/user-attachments/assets/e36ae1c6-8614-4e40-b788-d56228f1501a

## Pre-merge Checklist


src/lib/deeplink/handlers/add-site-blueprint-with-url.ts
```diff
@@ -32,7 +32,9 @@ export async function handleAddSiteWithBlueprint( urlObject: URL ): Promise< voi
 	if ( blueprintBase64 ) {
 		try {
 			const blueprintJson = Buffer.from( blueprintBase64, 'base64' ).toString( 'utf-8' );
-			JSON.parse( blueprintJson );
+			const parsedBlueprint = JSON.parse( blueprintJson );
+			await validateBlueprintData( parsedBlueprint );
+
```
gcsecsey: This can be reverted after https://github.com/Automattic/studio/pull/2103 is merged.

src/components/assistant-code-block.tsx
```diff
@@ -48,6 +49,24 @@ const LanguageBlock = ( props: ContextProps & CodeBlockProps ) => {
 
 	const { selectedSite } = useSiteDetails();
 
+	const isBlueprint = useMemo( () => {
```
bcotrim: Have you considered using the `validateBlueprintData` function?
  In my opinion it would be a good check to:
  1. Provide a reliable way to assert if a JSON is a blueprint
  2. Prevent our users from clicking "Open in Studio" for a blueprint that isn't valid, which is a bad user experience.

src/components/assistant-code-block.tsx
```diff
@@ -70,6 +89,27 @@ const LanguageBlock = ( props: ContextProps & CodeBlockProps ) => {
...
+								''
+							);
+							const base64 = btoa( binString );
+							const url = `${ PROTOCOL_PREFIX }://add-site?blueprint=${ encodeURIComponent(
+								base64
+							) }`;
+							getIpcApi().openURL( url );
+						} }
+					>
+						{ __( 'Open in Studio' ) }
```
ivan-ottinger: Since the user is already in the Studio app in this case, what if we used something like `Use to create a new site` instead?

src/components/assistant-code-block.tsx
```diff
@@ -70,6 +89,27 @@ const LanguageBlock = ( props: ContextProps & CodeBlockProps ) => {
 						} );
 					} }
 				></CopyTextButton>
+				{ isBlueprint && (
+					<Button
+						icon={ external }
```
ivan-ottinger: What if we used `create` (`+`) icon for the button - since clicking it will lead to new site creation?

src/components/assistant-code-block.tsx
gcsecsey: Thanks @bcotrim this is a really good point! I added a small hook to run the validation and updated the condition to use this in [7890264](https://github.com/Automattic/studio/pull/2104/commits/78902649efd288b66135a9b57f1059f3d819c94d).

src/components/assistant-code-block.tsx
gcsecsey: Great idea, thanks @ivan-ottinger! I added it [d3b7d4b](https://github.com/Automattic/studio/pull/2104/commits/d3b7d4b569e55f25675963b987ab78ca2cc4b1d5).

src/components/assistant-code-block.tsx
gcsecsey: I added it [d3b7d4b](https://github.com/Automattic/studio/pull/2104/commits/d3b7d4b569e55f25675963b987ab78ca2cc4b1d5).

src/components/open-blueprint-button.tsx
```diff
@@ -0,0 +1,33 @@
...
+}: OpenBlueprintButtonProps ) {
+	return (
+		<Button
+			{ ...props }
+			icon={ icon }
+			onClick={ () => {
+				const bytes = new TextEncoder().encode( content );
+				const binString = Array.from( bytes, ( byte ) => String.fromCodePoint( byte ) ).join( '' );
+				const base64 = btoa( binString );
+				const url = `${ PROTOCOL_PREFIX }://add-site?blueprint=${ encodeURIComponent( base64 ) }`;
```
gavande1: I think we should extract deeplink URL construction to its dedicate function instead of doing it inline.

src/components/assistant-code-block.tsx
```diff
@@ -70,6 +73,14 @@ const LanguageBlock = ( props: ContextProps & CodeBlockProps ) => {
 						} );
 					} }
 				></CopyTextButton>
+				{ isValidBlueprint && (
+					<OpenBlueprintButton
```
gavande1: Do you also think the name of button component is not accurate to its job? I think `CreateSiteWithBlueprintButton` or `AddSiteWithBlueprintButton` looks more appropriate.

src/hooks/use-is-valid-blueprint.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { Blueprint } from 'src/stores/wpcom-api';
+
+export function useIsValidBlueprint( className?: string, content?: string ) {
+	const blueprintJson = useMemo< Blueprint[ 'blueprint' ] | null >( () => {
+		if ( ! className?.includes( 'language-json' ) ) {
+			return null;
+		}
+		if ( ! content ) {
+			return null;
+		}
```
gavande1:
```suggestion
		if ( ! className?.includes( 'language-json' ) || ! content ) {
			return null;
		}
```
We can simplify a logic by the moving condition.

src/hooks/use-is-valid-blueprint.ts
```diff
@@ -0,0 +1,42 @@
...
+		} catch {
+			return null;
+		}
+	}, [ className, content ] );
+
+	const [ isBlueprintValid, setIsBlueprintValid ] = useState( false );
+
+	useEffect( () => {
+		const validate = async () => {
+			setIsBlueprintValid( false );
```
gavande1: I am wondering do we need to set state here again as default value is already been set above?

src/hooks/use-is-valid-blueprint.ts
```diff
@@ -0,0 +1,42 @@
...
+			if ( ! blueprintJson ) {
+				return;
+			}
+
+			try {
+				const validation = await getIpcApi().validateBlueprint( blueprintJson );
+				setIsBlueprintValid( Boolean( validation.valid ) );
+			} catch {
+				setIsBlueprintValid( false );
```
gavande1: Do we need to set false value here again? The default state is already false.

nightnei: Maybe we could print console.error here to not mute it

nightnei (approved):
LGTM and works as expected 👍
However, it seems that `blogname` was ignored, but it's definitelly not connected to this PR.
Screenshot 2025-11-25 at 14 40 35
Screenshot 2025-11-25 at 14 40 43

src/hooks/use-is-valid-blueprint.ts
gcsecsey: Thanks, I applied the suggestion. 👍

src/hooks/use-is-valid-blueprint.ts
gcsecsey: Since `blueprintJson` isn't changing, I agree it's not needed here. Updated in f5c8eca1714bc564d43a8f240c90cb69a16950a9.

src/hooks/use-is-valid-blueprint.ts
gcsecsey: Good points, thanks, I updated it in f5c8eca1714bc564d43a8f240c90cb69a16950a9.

src/components/open-blueprint-button.tsx
gcsecsey: Done in 2284f05dcf52ba81c7f97b28096380408c0b5d48.

src/components/assistant-code-block.tsx
gcsecsey: Good point, thanks! I like `AddSiteWithBlueprintButton`, I updated it in e530c58dc63c0e87befe15de5fccee1d6f80945f.

gavande1 (approved):
I have tested this and it works great. It took couple of retries to get the correct blueprint. Maybe we could open another issue to refine blueprint generation. But that's not related to this issue. This works well when valid blueprint is generated. LGTM 👍 

CleanShot 2025-11-26 at 11 40 56@2x
  
Image

bcotrim (approved):
Thanks for adding the suggestion @gcsecsey 
This looks great 👍

gcsecsey:
> I have tested this and it works great. It took couple of retries to get the correct blueprint. Maybe we could open another issue to refine blueprint generation. But that's not related to this issue. This works well when valid blueprint is generated. LGTM 👍
> 
> CleanShot 2025-11-26 at 11 40 56@2x Image

That's a good point, thanks @gavande1. I added STU-1053 to track this.

gavande1:
> That's a good point, thanks @gavande1. I added [STU-1053](https://linear.app/a8c/issue/STU-1053) to track this.

Lovely. Thanks!

PR #2114, opened 2025-11-24
Stats: Unify `bumpStat` function and validate its parameters length

ivan-ottinger:
## Related issues

Resolves STU-694

## Proposed Changes

- unify two `bumpStat` functions into one
  - validate its parameter length to ensure requirements are met (PCYsg-G3-p2)
  - add related unit tests

## Testing Instructions

Testing long stat group:

1. Check out the branch and apply a change like this one:
```diff
diff --git a/common/types/stats.ts b/common/types/stats.ts
index 286132de..a1495a41 100644
--- a/common/types/stats.ts
+++ b/common/types/stats.ts
@@ -4,7 +4,7 @@ export enum StatsGroup {
 	STUDIO_APP_LAUNCH_TOTAL = 'studio-app-launch-total',
 	STUDIO_APP_LAUNCH_UNIQUE = 'studio-app-launch-uniques',
 	STUDIO_APP_LAUNCH_UNIQUE_MONTHLY = 'studio-app-launch-uniqs-mon',
-	STUDIO_SITE_CREATE = 'studio-app-site-create',
+	STUDIO_SITE_CREATE = 'studio-app-site-create-very-long-group-name',
 	STUDIO_IMPORT = 'studio-app-import',
 	STUDIO_EXPORT = 'studio-app-export',
 	// Studio CLI
```
2. Build the app with `npm install && npm start`.
3. Create a new site.
4. When you look into the terminal, there should be a similar error message logged:

CleanShot 2025-11-24 at 12 57 30@2x

5. Without the diff from the step 1 applied, the logs in the terminal should include stat bumps:

CleanShot 2025-11-24 at 15 19 45@2x

## Pre-merge Checklist

### DIFF
--- a/cli/lib/stats.ts
+++ b/cli/lib/stats.ts
@@ -1,3 +1,4 @@
+import { bumpStat } from 'common/lib/bump-stat';
 import { AggregateInterval, StatsGroup, StatsMetric } from 'common/types/stats';
 import { isSameDay, isSameMonth, isSameWeek } from 'date-fns';
 import { lockAppdata, readAppdata, saveAppdata, unlockAppdata } from 'cli/lib/appdata';
@@ -40,26 +41,7 @@ export async function bumpAggregatedUniqueStat(
 	}
 }
 
-// Returns true if we attempted to bump the stat
-export function bumpStat( group: StatsGroup, stat: StatsMetric, bumpInDev = false ) {
-	if ( process.env.NODE_ENV === 'development' && ! bumpInDev ) {
-		console.info( `Would have bumped stat: ${ group }=${ stat }` );
-		return false;
-	}
-
-	// Fire and forget POST request
-	fetch( 'https://public-api.wordpress.com/wpcom/v2/studio-app/bump-stat', {
-		method: 'POST',
-		headers: {
-			'Content-Type': 'application/json',
-		},
-		body: JSON.stringify( { group, stat } ),
-	} ).catch( () => {
-		// A failed request typically indicates a network issue, which we don't need to report
-	} );
-
-	return true;
-}
+export { bumpStat } from 'common/lib/bump-stat';
 
 // Returns UTC timestamp of the last time the stat was bumped, or null if it has never been bumped.
 async function getLastBump( group: StatsGroup, stat: StatsMetric ): Promise< number | null > {
--- a/cli/lib/stats.ts
+++ b/cli/lib/stats.ts
@@ -41,8 +41,6 @@ export async function bumpAggregatedUniqueStat(
 	}
 }
 
-export { bumpStat } from 'common/lib/bump-stat';
-
 // Returns UTC timestamp of the last time the stat was bumped, or null if it has never been bumped.
 async function getLastBump( group: StatsGroup, stat: StatsMetric ): Promise< number | null > {
 	const { lastBumpStats } = await readAppdata();
--- a/common/lib/bump-stat.ts
+++ b/common/lib/bump-stat.ts
@@ -1,0 +1,22 @@
+import { StatsGroup, StatsMetric } from 'common/types/stats';
+
+// Returns true if we attempted to bump the stat
+export function bumpStat( group: StatsGroup, stat: StatsMetric, bumpInDev = false ) {
+	if ( process.env.E2E || ( process.env.NODE_ENV === 'development' && ! bumpInDev ) ) {
+		console.info( `Would have bumped stat: ${ group }=${ stat }` );
+		return false;
+	}
+
+	// Fire and forget POST request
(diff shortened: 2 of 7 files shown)


common/lib/bump-stat.ts
```diff
@@ -0,0 +1,41 @@
...
+		);
+		return false;
+	}
+
+	if ( stat.length > MAX_STAT_LENGTH ) {
+		console.error(
+			`Stat name "${ stat }" exceeds maximum length of ${ MAX_STAT_LENGTH } characters (actual: ${ stat.length }). Stat will not be bumped.`
+		);
+		return false;
+	}
```
ivan-ottinger, line 22: This is the new part of the function to validate the length of parameters. The rest is the same.

cli/lib/stats.ts
```diff
@@ -40,26 +41,7 @@ export async function bumpAggregatedUniqueStat(
...
-			'Content-Type': 'application/json',
-		},
-		body: JSON.stringify( { group, stat } ),
-	} ).catch( () => {
-		// A failed request typically indicates a network issue, which we don't need to report
-	} );
-
-	return true;
-}
+export { bumpStat } from 'common/lib/bump-stat';
```
fredrikekelund: Let's not re-export the function. All consumers of the function should just import directly from `common/lib/bump-stat`

src/lib/bump-stats/index.ts
```diff
@@ -46,26 +47,7 @@ export function bumpAggregatedUniqueStat(
...
-			'Content-Type': 'application/json',
-		},
-		body: JSON.stringify( { group, stat } ),
-	} ).catch( () => {
-		// A failed request typically indicates a network issue, which we don't need to report
-	} );
-
-	return true;
-}
+export { bumpStat } from 'common/lib/bump-stat';
```
fredrikekelund: Same thing here about not re-exporting.

fredrikekelund (commented):
Nice streamlining, and the length checks are great 👍 I would approve this straight away, but the re-exporting of `bumpStat` sticks out to me as something we should change.

cli/lib/stats.ts
ivan-ottinger: Thank you for the review and feedback, Fredrik! 

> Let's not re-export the function. All consumers of the function should just import directly from common/lib/bump-stat

This is something I have been considering and at the end decided for the reexport, but happy to switch to let all the consumers import it directly. 🙂

epeicher (approved):
Thanks @ivan-ottinger for improving this! I have tested it, and it works as expected. I can see the validation for long names and the correct stat being sent. I see you agree with updating the consumers instead of re-exporting the function so once those changes are done, this could be merged. LGTM! 


|Long name stat | Valid name stat|
|-------|------|
|  Image|  Image|

katinthehatsite (approved):
Nice, thank you for this improvement. The changes look good to me 👍

ivan-ottinger:
Thank you for your reviews and testing, everyone! 🙂 I have replaced the re-exports with direct imports in each consumer in https://github.com/Automattic/studio/pull/2114/commits/5525062f07f9adfc726e614373775a64b54d8f05 and will merge the PR once tests pass and the code freeze is lifted.

src/lib/bump-stats/index.ts
ivan-ottinger: Resolved in https://github.com/Automattic/studio/pull/2114/commits/5525062f07f9adfc726e614373775a64b54d8f05.

cli/lib/stats.ts
ivan-ottinger: Resolved in https://github.com/Automattic/studio/pull/2114/commits/5525062f07f9adfc726e614373775a64b54d8f05.

common/lib/bump-stat.ts
fredrikekelund, line 1: Could we move `bumpAggregatedUniqueStat` here, too? The Sentry error-catching we do in `src/lib/bump-stats/index.ts` could be moved to the consumers.

ivan-ottinger: Thanks, Fredrik. If I am understanding correctly, you would like to unify both `bumpAggregatedUniqueStat` and move them to `common/lib/bump-stat.ts`, right?

I have looked at the changes needed and I believe it would be better to handle that change in a separate task / PR - since we already have this one PR already approved and tested.

Here's a new task where we can tackle the changes: STU-1052.

fredrikekelund: After looking more closely at that function, I understand the difficulties. We would need to consolidate CRUD operations for the appdata file (or pass an appdata provider to `bumpAggregatedUniqueStat`) to make this work. That's definitely an option, since the only big difference between the app and the CLI is that the CLI uses a hardcoded path to appdata. In contrast, Studio uses a dynamically generated one (based on a path returned by Electron).

PR #2120, opened 2025-11-25
Make `build-for-windows.ps1` fail if `install-node-dependencies.sh` fails

fredrikekelund:
## Related issues

- Fixes STU-1044

## Proposed Changes

The Windows production build was borked in the recent 1.6.4 release because the `postinstall` script in `package.json` failed (specifically, `scripts/download-wp-server-files.ts` failed due to GitHub rate limits on the SQLite plugin download). This _should_ have caused the entire build to fail, but it didn't. 

The reason it didn't is that we don't call `set -e` in `install-node-dependencies.sh`. Instead, we use the shebang line at the top of that script to set the `-e` flag. However, because we specify that the `bash` executable should be used to run the script in `build-for-windows.ps1`, the shebang line is ignored, and the `-e` flag is never set. 

This PR fixes the problem by setting the `-eu` flags in `install-node-dependencies.sh` with an actual `set` command, and not in the shebang comment. 

## Testing Instructions

1. The `Build for Windows` step should pass on CI

## Pre-merge Checklist

### DIFF
--- a/.buildkite/commands/install-node-dependencies.sh
+++ b/.buildkite/commands/install-node-dependencies.sh
@@ -1,4 +1,6 @@
-#!/bin/bash -eu
+#!/bin/bash
+
+set -eu
 
 PLATFORM=$(uname -s)
 ARCHITECTURE=$(uname -m)
--- a/.buildkite/commands/install-node-dependencies.sh
+++ b/.buildkite/commands/install-node-dependencies.sh
@@ -42,12 +42,13 @@ npm ci \
   --maxsockets "$MAX_SOCKETS" \
   "$@"
 
-cd cli; npm ci \
+cd cli
+npm ci \
   --prefer-offline \
   --no-audit \
   --no-progress \
-  --maxsockets "$MAX_SOCKETS"; \
-  cd -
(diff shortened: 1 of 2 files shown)


.buildkite/commands/install-node-dependencies.sh
fredrikekelund: The changes to this file are only cosmetic. I made them during an earlier investigation path, but left them in simply because I thought it made the script slightly clearer.

AliSoftware (commented):
What do you think about instead removing the `-eu` from the shebang of the `install-node-dependencies.sh` and replacing it with an explicit call to `set -eu` in that same file right after that shebang?

That way the onus would still be on the `install-node-dependencies.sh` to set those flags itself, rather than expecting the caller (that one `.ps1` script for the time being but maybe more scripts would call it in the future) to remember to use `bash -eu` when calling it.

fredrikekelund:
@AliSoftware, good call. I've made that change 👍

AliSoftware (approved):

wojtekn (approved):

gcsecsey (approved):
LGTM 👍

PR #2118, opened 2025-11-25
remove studio start from app bundle

bcotrim:
## Related issues

- N/A

## Proposed Changes

- Remove studio start command from CLI index
- Remove some dependencies from Studio CLI build

## Testing Instructions

- CI checks
- `npm run cli:build`
- `npm run make`

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2121, opened 2025-11-25
Revert "remove studio start from app bundle"

bcotrim:
Reverts Automattic/studio#2118


fredrikekelund (approved):
Reverting this to ensure that we can work on the CLI more conveniently. This will likely be reverted again before the next stable Studio release

PR #2126, opened 2025-11-25
Adjust sync sites after untangling

nightnei:
## Related issues

- Fixes STU-918

## Proposed Changes
After untangling production and staging sites in Sync tab - there is more reasons to adjust the design to make items more gathered

## Testing Instructions
1. Run Studio
2. Open Sync tab
3. Connnect a few sites
4. Assert that everything looks good

<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2025-11-25 at 13 40 27
<td>Screenshot 2025-11-25 at 13 39 52
</tr>
</table>

### DIFF
--- a/src/modules/sync/components/sync-connected-sites.tsx
+++ b/src/modules/sync/components/sync-connected-sites.tsx
@@ -214,7 +214,7 @@ const SyncConnectedSitesSectionItem = ( {
 	return (
 		<div className="grid grid-cols-[max-content_1fr_max-content]">
 			<div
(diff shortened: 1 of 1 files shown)


katinthehatsite (approved):
I think this is a good start 👍 

I would have preferred the spacing here to be smaller as it visually looks bigger compared to other rows but I understand that it is coming from the `Reconnect` button so there is not much we could do:

Image

sejas (approved):
Changes look good! I tested it with connected and deleted sites.
In the future, we could redesign the entire screen and use data views with action buttons.

| Before | After |
|--------|--------|
|  Image | Image |

PR #2127, opened 2025-11-25
Update `studio site list` CLI command

fredrikekelund:
## Related issues

- Fixes STU-952

## Proposed Changes

@youknowriad originally implemented the `studio site list` command in https://github.com/Automattic/studio/pull/1742. This PR updates that command with relevant site details now that https://github.com/Automattic/studio/pull/2096 has been merged. We now include:
- Site status (online/offline)
- Site name
- Site path (prettified to make many paths shorter)
- Site URL
- PHP version

## Testing Instructions

1. Run `npm run cli:build`
4. Run `node dist/cli/main.js site list`
5. Ensure that the output is good
6. Run `node dist/cli/main.js site start --path PATH_TO_SITE` (where `PATH_TO_SITE` is the path to a Studio site)
7. Run `node dist/cli/main.js site list` again
8. Ensure that the site is now marked as "online" in the output

## Pre-merge Checklist


cli/commands/site/start.ts
```diff
@@ -112,7 +112,6 @@ export async function runCommand( siteFolder: string, skipBrowser = false ): Pro
 			const loggerError = new LoggerError( __( 'Failed to start site infrastructure' ), error );
 			logger.reportError( loggerError );
 		}
-		process.exit( 1 );
```
fredrikekelund, line 64: We already set the exit code in `cli/logger.ts`. No need to explicitly exit the process in this way.

nightnei:
@fredrikekelund after `Run node dist/cli/main.js site list` I got:
Screenshot 2025-11-25 at 14 17 39

Maybe something is wrong with my environment. Could you please double-check, and if it indeed works for you, then I will take a look deeper into why it fails for me.

cli/commands/site/list.ts
```diff
@@ -1,69 +1,98 @@
...
+					border: [],
+				},
+			} );
+
+			table.push(
+				...sitesData.map( ( site ) => [
+					site.status,
+					site.name,
+					site.path,
+					{ href: new URL( site.url ).toString(), content: site.url },
```
fredrikekelund, line 86: @nightnei, I believe the error you mentioned https://github.com/Automattic/studio/pull/2127#issuecomment-3575882860 happens here. Could you please add a simple `console.log(site.url)` call right above the `table.push` call here to see what might be causing the error here?

nightnei (dismissed):
LGTM and everything works as expected:
Screenshot 2025-11-25 at 15 29 07
Screenshot 2025-11-25 at 15 32 16
Screenshot 2025-11-25 at 15 32 31

However, I have a few questions:
1. If Studio is running and we have started sites - they are not reflected in CLI, is it expected?
2. After running a site via CLI, then running Studio (with previously started sites, so it's expected that they will be started after running Studio again) - I get the next error: 
Screenshot 2025-11-25 at 15 35 10

Marking as "requested changes", just not to miss your answer :)

fredrikekelund:
> If Studio is running and we have started sites - they are not reflected in CLI, is it expected?

It's expected for now. We intend to make the Studio CLI the "back-end" of Studio, and use it to start/stop/create sites. At that point, the CLI "running status" will be the single source of truth. For now, though, a site's running status isn't shared between the CLI and the app. 

> After running a site via CLI, then running Studio (with previously started sites, so it's expected that they will be started after running Studio again) - I get the next error:

Same thing: this is expected for now, but it will be resolved when the Studio CLI is responsible for starting/stopping/creating sites.

cli/commands/site/list.ts
```diff
@@ -1,69 +1,98 @@
...
-	sites.forEach( ( site ) => {
-		table.push( [ site.name, site.path, site.id, site.phpVersion ] );
-	} );
+async function getSiteListData( sites: SiteData[] ) {
+	const result: SiteListEntry[] = [];
 
-	return table;
-}
+	for await ( const site of sites ) {
+		const isOnline = await isServerRunning( site.id );
```
bcotrim, line 27: `isServerRunning` will eventually call `listProcesses` I think we could do some optimization here to only fetch the list of running processes once.

bcotrim (changes requested):

Image

Should we also display the WordPress version, or just PHP?
Edit: I just realized we don't store wp version in appdata, so we could address this in a follow-up.
We show this in Studio settings using `common/lib/get-wordpress-version.ts`
image

fredrikekelund:
> Should we also display the WordPress version, or just PHP?

I pondered this and ultimately decided to remove the PHP version. There are three truly relevant properties that this command should return, IMO: running status, path, and URL.

The name is largely irrelevant in the CLI context, but still makes sense to keep for consistency. Rather than printing more properties with this command (PHP version _and_ WP version), users can turn to the `studio site status` command for comprehensive details about a single site. 

> `isServerRunning` will eventually call `listProcesses` I think we could do some optimization here to only fetch the list of running processes once. 

I implemented a `cacheFunctionTTL` higher-order function that caches the return value of `listProcesses` for a short duration. This didn't make a big difference with the 10 sites I have (20 milliseconds), but I still decided to leave it in for now.

common/lib/cache-function-ttl.ts
```diff
@@ -0,0 +1,16 @@
...
+	return async ( ...args: Args ) => {
+		const cachedValue = cache.get( fn );
+		if ( cachedValue && Date.now() - cachedValue.timestamp < ttl ) {
+			return cachedValue.result as Return;
+		}
+		const value = await fn( ...args );
+		cache.set( fn, { result: value, timestamp: Date.now() } );
+		return value;
+	};
+}
```
bcotrim:
```suggestion
export function cacheFunctionTTL< Args extends unknown[], Return >(
	fn: ( ...args: Args ) => Promise< Return >,
	ttl = 1000
) {
	const cache = new Map< string, { result: Return; timestamp: number } >();

	return async ( ...args: Args ) => {
		const key = JSON.stringify( args );
		const cachedValue = cache.get( key );
		if ( cachedValue && Date.now() - cachedValue.timestamp < ttl ) {
			return cachedValue.result;
		}
		const value = await fn( ...args );
		cache.set( key, { result: value, timestamp: Date.now() } );
		return value;
	};
}
```

Currently this function has a bug for calls with different arguments.
A test like this would fail:

```typescript
it( 'should return different results for different arguments', async () => {
		const fetchUser = jest.fn( async ( userId: number ) => {
			return { id: userId, name: `User ${ userId }` };
		} );

		const cachedFetchUser = cacheFunctionTTL( fetchUser, 1000 );

		const user1 = await cachedFetchUser( 1 );
		const user2 = await cachedFetchUser( 2 );

		expect( user1 ).toEqual( { id: 1, name: 'User 1' } );
		expect( user2 ).toEqual( { id: 2, name: 'User 2' } );
	} );
```

fredrikekelund: Good catch 👍 I've updated `cacheFunctionTTL` to account for function arguments, too (using `fast-deep-equal` to compare function arguments)

bcotrim (approved):
Thanks for making the change to the cache function. I would still go for the simple suggestion but your implementation is very solid and well tested.
LGTM 👍 

Image

PR #2128, opened 2025-11-25
Fix env var cli warning

wojtekn:
## Related issues

- N/A

## Proposed Changes

This error occurs when Electron is launched as a subprocess by another application, and for security reasons, it disables Node.js environment variables in this context:

```
[1125/145742.696724:ERROR:electron/shell/app/node_main.cc:146] Node.js environment variables are disabled because this process is invoked by other apps.
```

I propose to pass the environment variable to the process being spawned.

## Testing Instructions

- Confirm there is no warning displayed when `studio` runs.

## Pre-merge Checklist

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1,8 +1,7 @@
 import 'cli/polyfills/browser-globals.js';
 import path from 'node:path';
 import { __ } from '@wordpress/i18n';
-import { suppressElectronEnvWarning } from 'common/lib/suppress-electron-env-warning';
-import { suppressPunycodeWarning } from 'common/lib/suppress-punycode-warning';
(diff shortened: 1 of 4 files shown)


wojtekn:
It doesn't fix the warning for me.

wojtekn:
Suppressing error doesn't change anything here, too.

Kudos to @fredrikekelund for finding the thread: https://github.com/microsoft/vscode/issues/204005

In my case, it was caused by the wp-calypso memory fix I had set in `~/.zshrc` using NODE_OPTIONS. It seems that adding the following to the `/usr/local/bin/studio` file helps:

```
# Unset NODE_* variables to avoid warning
unset NODE_OPTIONS
```

I will close this issue. We may consider adding this to the file shipped with Studio to fix it for everyone.

PR #2130, opened 2025-11-25
Blueprints: Add dedicated `blueprintDeeplink` step flow to handle blueprint deeplinks

ivan-ottinger:
## Related issues

- Resolves STU-1039

## Proposed Changes

- introduce a new "Start from a Blueprint" step / flow for blueprint deeplinks

CleanShot 2025-11-26 at 11 17 59@2x

## Testing Instructions

2. Try to open the blueprint deeplinks provided here: https://jsbin.com/levaburate/2/edit?html,css,output.
  - First two should result in a dialog error message.
  - The third one should get you to the new "Start from a Blueprint" step that displays blueprint title and description. It should also allow you to create a new site with it.
3. Try to open the following Base64-encoded blueprint links:
 
This one should result in an error message:

```
wp-studio://add-site?blueprint=ewogICJuYW1lIjogIk1lb3dzeSIsCiAgInNwZWNpZXMiOiAiY2F0IiwKICAiaW52YWxpZCI6ICJ0aGlzIGlzIG5vdCBhIHZhbGlkIGJsdWVwcmludCIKICAibWlzc2luZyI6ICJjb21tYSBhYm92ZSIKfQo=
```
 
This one should work correctly, display blueprint title and description and let you create a new site with it:

```
wp-studio://add-site?blueprint=eyJtZXRhIjp7InRpdGxlIjoiVGhpcyBpcyBhIHRlc3QgYmx1ZXByaW50IiwiZGVzY3JpcHRpb24iOiJJdCBpbnN0YWxscyBBc3RyYSB0aGVtZSBhbmQgWW9hc3QgU0VPIHBsdWdpbiIsImF1dGhvciI6Iml2YW4iLCJjYXRlZ29yaWVzIjpbInJzcyIsInNvY2lhbCB3ZWIiXX0sInByZWZlcnJlZFZlcnNpb25zIjp7InBocCI6IjguMyIsIndwIjoibGF0ZXN0In0sInN0ZXBzIjpbeyJzdGVwIjoiaW5zdGFsbFRoZW1lIiwidGhlbWVEYXRhIjp7InJlc291cmNlIjoid29yZHByZXNzLm9yZy90aGVtZXMiLCJzbHVnIjoiYXN0cmEifSwib3B0aW9ucyI6eyJhY3RpdmF0ZSI6dHJ1ZX19LHsic3RlcCI6Imluc3RhbGxQbHVnaW4iLCJwbHVnaW5EYXRhIjp7InJlc291cmNlIjoid29yZHByZXNzLm9yZy9wbHVnaW5zIiwic2x1ZyI6IndvcmRwcmVzcy1zZW8ifSwib3B0aW9ucyI6eyJhY3RpdmF0ZSI6dHJ1ZX19XX0=
```

4. Try to manually create sites using blueprint files from 39a4e-pb. The results should be as expected - there should be no regressions.

5. If you like, you can also test with your own blueprint files and blueprint deeplinks. 🙂 

## Pre-merge Checklist

### DIFF
--- a/src/modules/add-site/components/blueprint-deeplink.tsx
+++ b/src/modules/add-site/components/blueprint-deeplink.tsx
@@ -33,11 +33,7 @@ export default function BlueprintDeeplink( { selectedBlueprint }: BlueprintDeepl
 							{ __( 'Blueprint selected' ) }
 						</Text>
 					</HStack>
-					<Text
-						className="text-base font-medium text-gray-900 max-w-md px-4"
-						weight={ 400 }
-						title={ blueprintTitle }
-					>
+					<Text className="text-base font-medium text-gray-900 max-w-md px-4" weight={ 400 }>
 						{ blueprintTitle }
 					</Text>
 					{ blueprintDescription && (
--- a/src/modules/add-site/components/blueprint-deeplink.tsx
+++ b/src/modules/add-site/components/blueprint-deeplink.tsx
@@ -1,0 +1,52 @@
+import {
+	__experimentalVStack as VStack,
+	__experimentalHStack as HStack,
+	__experimentalHeading as Heading,
+	__experimentalText as Text,
+	Icon,
+} from '@wordpress/components';
+import { check } from '@wordpress/icons';
+import { useI18n } from '@wordpress/react-i18n';
+import { Blueprint } from 'src/stores/wpcom-api';
+
+interface BlueprintDeeplinkProps {
+	selectedBlueprint?: Blueprint;
+}
+
+export default function BlueprintDeeplink( { selectedBlueprint }: BlueprintDeeplinkProps ) {
+	const { __ } = useI18n();
+
+	const blueprintTitle = selectedBlueprint?.title || __( 'Blueprint' );
+	const blueprintDescription = selectedBlueprint?.excerpt || '';
+
+	return (
+		<VStack className="text-center w-full" alignment="top" spacing={ 0 }>
+			<Heading className="text-center text-[32px] text-gray-900 mb-[59px]" weight={ 500 }>
+				{ __( 'Start from a Blueprint' ) }
+			</Heading>
+
+			<div className="w-full max-w-[400px] h-[200px] mx-auto p-12 border-2 rounded-xl border-gray-300 bg-gray-50">
+				<VStack className="items-center justify-center h-full" spacing={ 2 }>
+					<HStack spacing={ 2 } alignment="center" className="text-green-600">
+						<Icon icon={ check } size={ 24 } />
+						<Text className="text-base font-medium text-gray-900">
+							{ __( 'Blueprint selected' ) }
+						</Text>
+					</HStack>
+					<Text
+						className="text-base font-medium text-gray-900 max-w-md px-4"
+						weight={ 400 }
+						title={ blueprintTitle }
+					>
+						{ blueprintTitle }
+					</Text>
+					{ blueprintDescription && (
+						<Text className="text-sm text-gray-600 max-w-md px-4 line-clamp-2">
+							{ blueprintDescription }
+						</Text>
+					) }
+				</VStack>
+			</div>
+		</VStack>
+	);
+}
--- a/src/modules/add-site/components/blueprint-deeplink.tsx
+++ b/src/modules/add-site/components/blueprint-deeplink.tsx
@@ -24,7 +24,7 @@ export default function BlueprintDeeplink( { selectedBlueprint }: BlueprintDeepl
 				{ __( 'Start from a Blueprint' ) }
 			</Heading>
 
-			<div className="w-full max-w-[400px] h-[200px] mx-auto p-12 border-2 rounded-xl border-gray-300 bg-gray-50">
+			<div className="w-full max-w-[400px] h-[250px] mx-auto p-12 border-2 rounded-xl border-gray-300 bg-gray-50">
 				<VStack className="items-center justify-center h-full" spacing={ 2 }>
 					<Icon icon={ check } size={ 24 } className="text-green-600" />
 					<Text className="text-xl font-medium text-gray-900">{ __( 'Blueprint selected' ) }</Text>
--- a/src/modules/add-site/components/blueprint-deeplink.tsx
+++ b/src/modules/add-site/components/blueprint-deeplink.tsx
@@ -1,6 +1,5 @@
 import {
 	__experimentalVStack as VStack,
-	__experimentalHStack as HStack,
 	__experimentalHeading as Heading,
 	__experimentalText as Text,
 	Icon,
@@ -27,12 +26,8 @@ export default function BlueprintDeeplink( { selectedBlueprint }: BlueprintDeepl
 
 			<div className="w-full max-w-[400px] h-[200px] mx-auto p-12 border-2 rounded-xl border-gray-300 bg-gray-50">
 				<VStack className="items-center justify-center h-full" spacing={ 2 }>
-					<HStack spacing={ 2 } alignment="center" className="text-green-600">
-						<Icon icon={ check } size={ 24 } />
-						<Text className="text-base font-medium text-gray-900">
-							{ __( 'Blueprint selected' ) }
-						</Text>
-					</HStack>
+					<Icon icon={ check } size={ 24 } className="text-green-600" />
+					<Text className="text-xl font-medium text-gray-900">{ __( 'Blueprint selected' ) }</Text>
 					<Text className="text-base font-medium text-gray-900 max-w-md px-4" weight={ 400 }>
 						{ blueprintTitle }
 					</Text>
--- a/src/modules/add-site/hooks/use-stepper.ts
+++ b/src/modules/add-site/hooks/use-stepper.ts
@@ -16,10 +16,12 @@ interface StepperStep {
 
 interface StepperConfig {
 	onBlueprintContinue?: () => void;
+	onBlueprintDeeplinkContinue?: () => void;
 	onBackupContinue?: () => void;
 	onPullRemoteContinue?: () => void;
 	onCreateSubmit?: ( event: FormEvent ) => void;
 	canSubmitBlueprint?: boolean;
+	canSubmitBlueprintDeeplink?: boolean;
 	canSubmitBackup?: boolean;
 	canSubmitPullRemote?: boolean;
 	canSubmitCreate?: boolean;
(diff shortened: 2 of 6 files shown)


src/modules/add-site/index.tsx
```diff
@@ -379,7 +399,7 @@ export default function AddSite( { className, variant = 'outlined' }: AddSitePro
 		setBlueprintPreferredVersions,
 	} );
 
-	const initialNavigatorPath = selectedBlueprint ? '/blueprint/create' : '/';
+	const initialNavigatorPath = selectedBlueprint ? '/blueprintDeeplink' : '/';
```
gavande1: We could use better name for route as `blueprintDeeplink` doesn't sounds right. How about `/blueprint/deeplink`?

src/modules/add-site/components/blueprint-deeplink.tsx
```diff
@@ -0,0 +1,48 @@
...
+export default function BlueprintDeeplink( { selectedBlueprint }: BlueprintDeeplinkProps ) {
+	const { __ } = useI18n();
+
+	const blueprintTitle = selectedBlueprint?.title || __( 'Blueprint' );
+	const blueprintDescription = selectedBlueprint?.excerpt || '';
+
+	return (
+		<VStack className="text-center w-full" alignment="top" spacing={ 0 }>
+			<Heading className="text-center text-[32px] text-gray-900 mb-[59px]" weight={ 500 }>
+				{ __( 'Start from a Blueprint' ) }
```
gavande1, line 24:
```suggestion
				{ __( 'Start with Blueprint' ) }
```
This sounds more grammatically correct to me as Blueprint is feature. Let me know what do you think.

src/modules/add-site/components/blueprint-deeplink.tsx
```diff
@@ -0,0 +1,48 @@
...
+			</Heading>
+
+			<div className="w-full max-w-[400px] h-[200px] mx-auto p-12 border-2 rounded-xl border-gray-300 bg-gray-50">
+				<VStack className="items-center justify-center h-full" spacing={ 2 }>
+					<HStack spacing={ 2 } alignment="center" className="text-green-600">
+						<Icon icon={ check } size={ 24 } />
```
gavande1: I would put Icon on top of title, not to the left. Also I would raise font size of title.

gavande1 (commented):
I have tested this PR, and it works great. Even without the designs, this looks solid. Thanks for working on it. I left some minor comments for you to review. 
CleanShot 2025-11-26 at 11 54 31@2x

src/modules/add-site/hooks/use-stepper.ts
```diff
@@ -66,6 +68,18 @@ export function useStepper( config?: StepperConfig ): UseStepper {
 			{ id: 'site-details', label: __( 'Site name & details' ), path: '/pullRemote/create' },
 		];
 
+		const blueprintDeeplinkSteps: StepperStep[] = [
+			{ id: 'blueprint-selected', label: __( 'Blueprint selected' ), path: '/blueprintDeeplink' },
```
gavande1:
```suggestion
			{ id: 'blueprint-selected', label: __( 'Blueprint details' ), path: '/blueprintDeeplink' },
```
In this case user didn't select blueprint so showing "Blueprint selected" doesn't look right. I would suggest to change it to "Blueprint details" or something else to reflect text that is more closer to user action.

ivan-ottinger: Good point! I have changed the step label to `Blueprint details` in https://github.com/Automattic/studio/pull/2130/commits/34aa07c89a562daefc1f7821957f52e54bf62a22.

src/modules/add-site/components/blueprint-deeplink.tsx
ivan-ottinger: > This sounds more grammatically correct to me as Blueprint is feature. Let me know what do you think.

Thanks for the suggestion. The reason I picked `Start from a Blueprint` is because we already use it for the standard Blueprint flow:

CleanShot 2025-11-26 at 10 51 17@2x

src/modules/add-site/index.tsx
ivan-ottinger: I see what you mean. Thanks for the suggestion.

Yesterday I was considering different paths, but then picked `blueprintDeeplink`. If we decided for `/blueprint/deeplink` instead, we would then also need to consider:

- change of `/blueprintDeeplink/create` to `/blueprint/deeplink/create`
- need to ensure correct order of path checks in `src/modules/add-site/hooks/use-stepper.ts` (since we have already rely on `.startsWith( '/blueprint' )` for the normal Blueprint flow):

```js
		if ( location.path?.startsWith( '/blueprintDeeplink' ) ) {
			return {
				flow: 'blueprintDeeplink',
				steps: blueprintDeeplinkSteps,
			};
		}

		if ( location.path?.startsWith( '/blueprint' ) ) {
			return {
				flow: 'blueprint',
				steps: blueprintSteps,
			};
		}
```

But I am open to these changes if we agree on them.

src/modules/add-site/components/blueprint-deeplink.tsx
ivan-ottinger: Sounds good to me. 👌🏼 I have made the change in https://github.com/Automattic/studio/pull/2130/commits/bcfb9c6efabad885956c034a2af50112865b473b.

ivan-ottinger:
Thank you for the review and testing of the draft PR, Rahul! 🙂 I have applied some of the changes and and responded to your comments.

I have also applied the following change to handle a case where the modal is already opened when the user uses a deeplink: https://github.com/Automattic/studio/pull/2130/commits/49c68ef09ff3c86fa6fff55ed0e2c0d9d8b248a6.

wojtekn:
@ivan-ottinger I think UX with that step is great - it's clear for the user what happened after navigating deeplink to Studio, and it's clear what will happen when the user clicks "Continue".

I noticed a small layout issue - in the smallest possible Studio window, the text is cut:

Screenshot 2025-11-26 at 13 33 50

Also, I think we might look for a better way to display errors - the dialog seems a bit obtrusive, and doesn't allow user to cancel the flow using the keyboard:

Screenshot 2025-11-26 at 13 33 05

What if we display the error as red text with an icon below the field, similar to what we do in other forms? Or maybe we could reuse the warning we have in "Start from Blueprint" which shows validation messages informing about unsupported features included in the Blueprint?

ivan-ottinger:
Thank you for testing, Wojtek!

> I noticed a small layout issue - in the smallest possible Studio window, the text is cut:

Good catch! I have fixed this in https://github.com/Automattic/studio/pull/2130/commits/f41261d05e24ad06be7e5276121194e76b3eb4d6.

> Also, I think we might look for a better way to display errors - the dialog seems a bit obtrusive, and doesn't allow user to cancel the flow using the keyboard:

From what I see in your screenshot, the dialog window with error there is displayed on top of the flow that had been previously opened with a correct deeplink. When a deeplink is broken, it does not trigger the flow at all and just displays the error message in the dialog. 🙂 

When it comes to the validation warnings, we have a separate task (STU-1047) to add them in the same way as it is currently with the standard Blueprint flow:

CleanShot 2025-11-25 at 16 57 34@2x

src/modules/add-site/index.tsx
gavande1: If it's too big of a change, then we could just keep things as they are right now.

src/modules/add-site/components/blueprint-deeplink.tsx
gavande1: I agree. However, I feel this action is a bit different than the previous one. I would still choose "Start with a Blueprint" over "Start from a Blueprint." I think it's much clearer. However, I leave the final decision to you.

gavande1:
> I think we might look for a better way to display errors - the dialog seems a bit obtrusive, and doesn't allow user to cancel the flow using the keyboard:

+1 to that. I feel that current error flow is not so great in terms of UX.

ivan-ottinger:
> I think we might look for a better way to display errors - the dialog seems a bit obtrusive, and doesn't allow user to cancel the flow using the keyboard:

> +1 to that. I feel that current error flow is not so great in terms of UX.

With the currently-proposed implementation, if the deeplink is broken, it does not start the flow at all, so the user doesn't need to cancel the flow. If we displayed the errors in the flow instead of the dialog, the user would be taken away from where they are at the moment to the flow where they could just read the same error message. We wouldn't be able to display any blueprint description / title - since the JSON (or the deeplink) is broken in some way.

That's why I believe it makes sense to keep the error in the dialog. On the other hand, if the link is correct and just have some unsupported features, then yes, I think it makes sense to start the flow and display the warnings to the user - since in that case, the user can take action and continue creating their site.

src/modules/add-site/components/blueprint-deeplink.tsx
gcsecsey: I think it makes sense to use `Start from a Blueprint` as the flow is very similar, only the way we get the Blueprint is different. 🙂

If we wanted to differentiate, I could also imagine `Open Blueprint` or `Import Blueprint` as the headings for this screen.

gcsecsey (approved):
Thanks for wrangling this Ivan! I think the new flow already has great UX, I feel well informed about the import process.

I tested with the different scenarios and it's working great in all cases. I also generated a Blueprint that doesn't have any metadata:

```
wp-studio://add-site?blueprint=ewogICIkc2NoZW1hIjogImh0dHBzOi8vcGxheWdyb3VuZC53b3JkcHJlc3MubmV0L2JsdWVwcmludC1zY2hlbWEuanNvbiIsCiAgImxvZ2luIjogdHJ1ZSwKICAic2l0ZU9wdGlvbnMiOiB7CiAgICAiYmxvZ25hbWUiOiAiU3Vic3RyYXRhIgogIH0sCiAgInN0ZXBzIjogWwogICAgewogICAgICAic3RlcCI6ICJyZXNldERhdGEiCiAgICB9CiAgXSwKICAicHJlZmVycmVkVmVyc2lvbnMiOiB7CiAgICAicGhwIjogIjguMyIsCiAgICAid3AiOiAibGF0ZXN0IgogIH0sCiAgImZlYXR1cmVzIjoge30KfQo=
```

I think the copy in this case is also solid:

Image

src/modules/add-site/components/blueprint-deeplink.tsx
ivan-ottinger: Thank you for your feedback Rahul and Gergely. I will leave the title as-is for now, but we can indeed revisit it later, e.g. once we get the design input on this step from Marina.

src/modules/add-site/index.tsx
ivan-ottinger: > If it's too big of a change, then we could just keep things as they are right now.

I was mostly worried about fragility of the `.startsWith` path checks in combination with paths that share the same string (and relying on the order of those checks only).

Looking at it from a different angle today, I changed the path of the flow where user selects blueprint themselves from `/blueprint` to `/blueprint/select` as well (in https://github.com/Automattic/studio/pull/2130/commits/a0d3cae9fe151dd145496889fb55f3b507c3af9e).

This way the paths are distinct and we could also get rid of that camel case. 🙂

ivan-ottinger:
Thank you for your feedback and testing everyone!

PR #2132, opened 2025-11-25
Implement RTK Query approach for fetching wpcom sites

epeicher:
## Related issues

- Related STU-1035
- Related STU-710

### Before

https://github.com/user-attachments/assets/fa96ab70-c0ff-45ab-ace3-69e546c95640

### After 

https://github.com/user-attachments/assets/063b269f-f061-412d-8073-194e63c4cbf4

## Proposed Changes

- Migrated from custom `useFetchWpComSites` hook to RTK Query with `useGetWpComSitesQuery`. The new query uses the userId to invalidate the cache
- Improved state management with automatic caching, background refetching, and request deduplication
- Reorganized sync utilities:
  - Moved sync-related utilities from `src/hooks/use-fetch-wpcom-sites/` to `src/modules/sync/lib/`
  - Consolidated types into `src/modules/sync/types.ts`
  - Created shared `sync-support.ts` module for `getSyncSupport` and helper functions
- Updated all imports across 20+ files to reflect new structure
- Removed deprecated hook directory
- Enables by default the `Streamline Onboarding` feature flag

## Testing Instructions

- Use or create a site that does not have any connected sites
- Click on `Publish site` button and observe that the loading experience is fine (there are no multiple loading indicators)
- On the Sync tab, click on `Publish site` and check that the loading is minimal as the sites are cached
- Test also the `Pull site` button
- Delete or create a new site on WordPress.com
- Click on the `Publish site` button, check that the deleted or created site is displayed (there could be a delay)
- Log out and log in with a user with no sites
- Click on the `Publish site` button
- Check that the `Find a perfect plan` modal is displayed
- Check different use cases and confirm that the sites are correctly loaded

## Pre-merge Checklist


src/modules/sync/index.tsx
```diff
@@ -136,27 +135,52 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
...
+		}
+	}, [ refetchWpComSites, isUninitializedSyncSites ] );
+
+	// Refetch sites when modal opens to check for newly created sites
 	useEffect( () => {
-		if ( pendingModalMode && ! isFetchingSyncSites ) {
-			dispatch( connectedSitesActions.openModal( pendingModalMode ) );
-			setPendingModalMode( null );
+		if ( isModalOpen && isAuthenticated && ! isUninitializedSyncSites ) {
+			refetchWpComSites().catch( ( error ) => {
```
sejas: Would be possible to refetch wpcom sites inside openModal action or call it at the same time?

sejas (commented):
Great work! Thanks for moving the wpcom sites request to RTK. Now it doesn't block the user every time we click on the buttons.
I just noticed that we display a loading state in the push site button every time we change tabs.

https://github.com/user-attachments/assets/c76cc022-3a40-4c46-8eb4-6480e784bd52

src/modules/sync/index.tsx
epeicher: Absolutely, I have done the changes on [d7113b1](https://github.com/Automattic/studio/pull/2132/commits/d7113b16455a7c9a5a2ce7718508f73120544642), I think that is better. 

I have also fixed the loading indicator in the button as part of [9a6f74f](https://github.com/Automattic/studio/pull/2132/commits/9a6f74f7f8ff62ba163e2795473116d0e61cfe5b)

common/lib/cache-function-ttl.ts
```diff
@@ -31,7 +31,7 @@ export function cacheFunctionTTL< Args extends unknown[], Return >(
 }
 
 function pruneCache(): void {
-	for ( const [ fn, cachedResults ] of cache ) {
+	for ( const [ _, cachedResults ] of cache ) {
```
epeicher, line 34: This is a lint fix, unrelated to this PR

src/stores/sync/wpcom-sites.ts
```diff
@@ -0,0 +1,194 @@
...
+	type SitesEndpointSite,
+} from 'src/modules/sync/lib/sync-support';
+import { withOfflineCheck } from 'src/stores/utils/with-offline-check';
+import { getWpcomClient } from 'src/stores/wpcom-api';
+import type { SyncSite, SyncSupport } from 'src/modules/sync/types';
+
+// Schema for WordPress.com sites endpoint
+const sitesEndpointSiteSchema = z.object( {
+	ID: z.number(),
+	is_wpcom_atomic: z.boolean(),
```
sejas, line 14: We have parts of this type duplicated  in `sync-support.ts`. Can we unify them?
https://github.com/Automattic/studio/blob/b09206b3ed665d51226078da451a9875e0158e21/src/modules/sync/lib/sync-support.ts#L3-L8

src/modules/sync/index.tsx
```diff
@@ -136,27 +135,53 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
...
+			return [];
+		}
+		try {
+			const result = await refetchWpComSites();
+			return result.data ?? [];
+		} catch ( error ) {
+			// Query might not be ready to refetch yet (e.g., was skipped due to offline)
+			console.warn( 'Failed to refetch sites:', error );
+			return [];
+		}
```
sejas: I see this pattern of catching the refetchWpComSite error repeated a few times. I wonder if makes sense to move the try catch inside the query and always return an empty array if it fails.

sejas (commented):
@epeicher, thanks for making the adjustments. The PR looks good. I only found that Studio throws a warning when clicking `Publish site` from a different tab, like overview.

`Failed to refetch sites: Error: Cannot refetch a query that has not been started yet.`

Image

src/stores/sync/wpcom-sites.ts
epeicher: Definitely, this type can be inferred from the schema, I will update this

src/modules/sync/index.tsx
epeicher: Yes, that's right, the `refetchWpComSite` is not always ready, so it can throw errors, I will encapsulate this on a function 👍

epeicher: This has been simplified as part of [63cc6b8](https://github.com/Automattic/studio/pull/2132/commits/63cc6b85204308c49d61f9455d34c54d131363b8), as the effect is not required anymore after refetching the sites directly. There are not any errors now, but I have added the `void` operator to refetch the sites on the background (i.e. do not use `await` every time) and ignore any potential errors during the refetch, please let me know if you think it would be better to handle and maybe log them.

src/stores/sync/wpcom-sites.ts
epeicher: Done as part of [427d3ab](https://github.com/Automattic/studio/pull/2132/commits/427d3ab0710e5e97afed3d5c17cce2cc22dc8705), duplication removed, great catch!

src/modules/sync/index.tsx
epeicher: Additionally, I have consolidated both buttons into one [with a condition](https://github.com/Automattic/studio/pull/2132/commits/7e1bbdbe6c10a200f90b7053b2613f0faa457f82#diff-9c7af3248cfe05390c0339be78082480629b738a1c274c90e4586c81b15c0128R40) that redirects to the Sync tab only when required in the commit [7e1bbdb](https://github.com/Automattic/studio/pull/2132/commits/7e1bbdbe6c10a200f90b7053b2613f0faa457f82)

katinthehatsite (commented):
First, nice work 👍 This is a huge PR and lots of work so well done!

I noticed one thing that when the sync modal is already open, I see the loading state on the `Publish site` button in the background:

Image

I also see it for the case when I click on the `Publish site` in the top right corner and then close the modal, the loading state persists in that case instead of going away.

src/components/publish-site-button.tsx
```diff
@@ -1,53 +1,79 @@
...
 	const isAnySiteSyncing = isAnySitePulling || isAnySitePushing;
-	const handlePublishClick = () => {
-		setSelectedTab( 'sync' );
+
+	const handlePublishClick = useCallback( () => {
+		if ( redirectToSync ) {
+			setSelectedTab( 'sync' );
+		}
+		if ( isAuthenticated && ! isUninitializedSyncSites ) {
+			// Refetch sites on the background but ignore errors
```
katinthehatsite:
```suggestion
			// Refetch sites in the background but ignore errors
```

epeicher:
Thanks for the review @katinthehatsite!

> I noticed one thing that when the sync modal is already open, I see the loading state on the Publish site button in the background

Yes, that was expected as I included `isBusy` for the button while the sites are being fetched in the background, but I understand that seems confusing so I have removed it as part of [b134968](https://github.com/Automattic/studio/pull/2132/commits/b134968db13567effc158484ad1700f5d5cd69e4)

src/components/publish-site-button.tsx
```diff
@@ -1,53 +1,75 @@
...
+	const handlePublishClick = useCallback( () => {
+		if ( redirectToSync ) {
+			setSelectedTab( 'sync' );
+		}
+		if ( isAuthenticated && ! isUninitializedSyncSites ) {
+			// Refetch sites in the background but ignore errors
+			void refetchWpComSites();
+		}
 		dispatch( connectedSitesActions.openModal( 'push' ) );
-	};
```
fredrikekelund, line 30: Why do we need to call `refetchWpComSites` here? If the goal is to refetch whenever the sync modal opens, would it not make more sense to call this from within the sync modal?

We could combine it with the `refetchOnMountOrArgChange` option to refetch the hook data when mounting.

src/hooks/sync-sites/use-listen-deep-link-connection.ts
fredrikekelund, line 1: I suggest moving more of this logic to `src/stores/sync/connected-sites.ts` by updating the `connectSite` mutation like so:

```ts
connectSite: builder.mutation<
	SyncSite[],
	{ remoteSiteId: number; localSiteId: string; userId?: number }
>( {
	queryFn: async ( { remoteSiteId, localSiteId, userId }, api ) => {
		const connectedSites = await getIpcApi().getConnectedWpcomSites( localSiteId );
		const { data: remoteSites = [] } = await api.dispatch(
			wpcomSitesApi.endpoints.getWpComSites.initiate( {
				connectedSiteIds: connectedSites.map( ( site ) => site.id ),
				userId,
			} )
		);
		const siteToConnect = remoteSites.find( ( site ) => site.id === remoteSiteId );

		if ( ! siteToConnect ) {
			return {
				error: { status: 'CUSTOM_ERROR', error: 'Site not found in WordPress.com sites' },
			};
		}

		await getIpcApi().connectWpcomSites( [
			{
				sites: [ siteToConnect ],
				localSiteId,
			},
		] );
	
		const actualConnectedSites = await getIpcApi().getConnectedWpcomSites( localSiteId );
	
		return { data: actualConnectedSites };
	},

```

This would allow us to shorten this hook significantly:

```ts
export function useListenDeepLinkConnection() {
	const [ connectSite ] = useConnectSiteMutation();
	const { selectedSite, setSelectedSiteId } = useSiteDetails();
	const { setSelectedTab, selectedTab } = useContentTabs();
	const { user } = useAuth();

	useIpcListener( 'sync-connect-site', async ( _event, { remoteSiteId, studioSiteId } ) => {
		if ( selectedSite?.id && selectedSite.id !== studioSiteId ) {
			// Select studio site that started the sync
			setSelectedSiteId( studioSiteId );
		}
		await connectSite( { remoteSiteId, localSiteId: studioSiteId, userId: user?.id } );
		if ( selectedTab !== 'sync' ) {
			// Switch to sync tab
			setSelectedTab( 'sync' );
		}
	} );
}
```

src/modules/add-site/components/pull-remote-site.tsx
```diff
@@ -118,9 +120,20 @@ export function PullRemoteSite( {
...
+	const { selectedSite } = useSiteDetails();
+	const { data: connectedSites = [] } = useGetConnectedSitesForLocalSiteQuery( {
+		localSiteId: selectedSite?.id,
+		userId: user?.id,
+	} );
+	const connectedSiteIds = connectedSites.map( ( { id } ) => id );
+	const { data: syncSites = [], refetch: refetchSites } = useGetWpComSitesQuery( {
+		connectedSiteIds,
+		userId: user?.id,
+	} );
```
fredrikekelund:
```suggestion
	const { data: syncSites = [] } = useGetWpComSitesQuery(
		{ connectedSiteIds, userId: user?.id },
		{ refetchOnMountOrArgChange: true }
	);
```
I believe we could remove the `refetchSites` call by passing the `refetchOnMountOrArgChange` flag here.

src/modules/sync/index.tsx
```diff
@@ -136,27 +136,35 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
...
+		refetch: refetchWpComSites,
+	} = useGetWpComSitesQuery( { connectedSiteIds, userId: user?.id } );
+
+	// Helper function to open modal and refetch sites to check for newly created sites
+	const handleOpenModal = useCallback(
+		( mode: SyncModalMode ) => {
+			if ( isAuthenticated && ! isUninitializedSyncSites ) {
+				// Refetch sites on the background but ignore errors
+				void refetchWpComSites();
+			}
```
fredrikekelund: Instead of explicitly refetching in the `handleOpenModal` handler, I suggest adding another `useGetWpComSitesQuery` call in `SyncSitesModalSelector`. See my comment in that file for more details.

src/modules/sync/components/sync-sites-modal-selector.tsx
fredrikekelund, line 1: I suggest making it so that the `useGetWpComSitesQuery` hook is declaratively refetched by removing the `syncSites` prop from this component, and adding another call to the `useGetWpComSitesQuery` hook like so:

```ts
	const { user } = useAuth();
	const { data: connectedSites = [] } = useGetConnectedSitesForLocalSiteQuery( {
		localSiteId: selectedSite.id,
		userId: user?.id,
	} );
	const connectedSiteIds = connectedSites.map( ( { id } ) => id );
	const { data: syncSites = [] } = useGetWpComSitesQuery(
		{ connectedSiteIds, userId: user?.id },
		{ refetchOnMountOrArgChange: true }
	);
```

The `refetchOnMountOrArgChange` option should accomplish the same thing as the refetch call in `src/modules/sync/index.tsx`

src/stores/sync/wpcom-sites.ts
```diff
@@ -0,0 +1,192 @@
...
+					);
+
+					// whenever array of syncSites changes, we need to update connectedSites to keep them updated with wordpress.com
+					const { updatedConnectedSites } = reconcileConnectedSites(
+						allConnectedSites,
+						syncSitesForReconciliation
+					);
+					await getIpcApi().updateConnectedWpcomSites( updatedConnectedSites );
+
+					// Second transformation using connectedSiteIds parameter (for syncSupport calculation for selected site)
```
fredrikekelund:
```suggestion
					const syncSitesForReconciliation = transformSitesResponse(
						parsedResponse.sites,
						allConnectedSites.map( ( { id } ) => id )
					);

					const { updatedConnectedSites } = reconcileConnectedSites(
						allConnectedSites,
						syncSitesForReconciliation
					);
					await getIpcApi().updateConnectedWpcomSites( updatedConnectedSites );

```
Nit, but the variable names are clear enough here that I would argue we could skip the line comments.

src/stores/sync/wpcom-sites.ts
```diff
@@ -0,0 +1,192 @@
...
+		isStaging,
+		isPressable: isPressableSite( site ),
+		environmentType: site.environment_type,
+		syncSupport,
+		lastPullTimestamp: null,
+		lastPushTimestamp: null,
+	};
+}
+
+function transformSitesResponse( sites: unknown[], connectedSiteIds: number[] ): SyncSite[] {
```
fredrikekelund, line 86: I know this function is the same as before, but I think the `connectedSiteIds` deserves a short explainer in a comment.

fredrikekelund (commented):
Great refactoring, I love to see it 👍

I left a couple of comments, mostly with a focus on how we can make the code more declarative and rely on RTK Query internals more, or move logic from components inside mutations. 

I haven't tested extensively yet, but happy to do so in a second review later.

src/components/publish-site-button.tsx
epeicher: I would say that's a matter of preference, I updated that [motivated by this comment](https://github.com/Automattic/studio/pull/2132#discussion_r2566701101) but I will investigate the result of using it when opening the sync modal and combining it with `refetchOnMountOrArgChange`

src/hooks/sync-sites/use-listen-deep-link-connection.ts
epeicher: Thanks for the suggestion, I will work on that

src/modules/add-site/components/pull-remote-site.tsx
epeicher: OK, I will have a look at it, I had that initially, but it was refetching the sites more than expected. I will check it again after latest updates.

src/stores/sync/wpcom-sites.ts
epeicher: Hmm… ok

src/components/publish-site-button.tsx
epeicher: After doing the changes, it's clear that it was better to do it in the component, thanks for the suggestion! I have done it as part of [5079edf](https://github.com/Automattic/studio/pull/2132/commits/5079edfa49fdd135682b17c5e4e2308657ee7e33)

src/modules/sync/index.tsx
epeicher: Yes, that's a better approach, I have implemented it as part of [5079edf](https://github.com/Automattic/studio/pull/2132/commits/5079edfa49fdd135682b17c5e4e2308657ee7e33)

src/modules/sync/components/sync-sites-modal-selector.tsx
epeicher: Agreed 😄 , I have done it as part of [5079edf](https://github.com/Automattic/studio/pull/2132/commits/5079edfa49fdd135682b17c5e4e2308657ee7e33)

src/stores/sync/wpcom-sites.ts
epeicher: Done as part of [99a6b55](https://github.com/Automattic/studio/pull/2132/commits/99a6b5566847f6ce7851a36c476622bede2d91cf)

src/stores/sync/wpcom-sites.ts
epeicher: Done as part of [99a6b55](https://github.com/Automattic/studio/pull/2132/commits/99a6b5566847f6ce7851a36c476622bede2d91cf)

epeicher:
Thanks @fredrikekelund for the review and the suggestions! I have implemented most of them and I still want to look at [this one](https://github.com/Automattic/studio/pull/2132#discussion_r2580931078) and [this one](https://github.com/Automattic/studio/pull/2132#discussion_r2580945186). I will let you know when they are addressed and we can do another review 🙏

src/hooks/sync-sites/use-listen-deep-link-connection.ts
epeicher: I have been working on this, but I have found that this will trigger an additional call to the `/me/sites` endpoint when the `connectSite` mutation is used from [the add-site](https://github.com/Automattic/studio/blob/99a6b5566847f6ce7851a36c476622bede2d91cf/src/hooks/use-add-site.ts#L144) and from [the ContentTabSync handleConnect](https://github.com/Automattic/studio/blob/99a6b5566847f6ce7851a36c476622bede2d91cf/src/modules/sync/index.tsx#L159), where they already have the site, so I am not sure we are improving here.

What do you think about tackling this specific refactor as a follow-up and progressing with the current changes?

src/modules/add-site/components/pull-remote-site.tsx
epeicher: I have done the changes as part of 61e8a4bc, they do the same refetch on mount and we avoid the `useEffect` 👍

epeicher:
Hi @fredrikekelund, I have implemented most of your suggestions, and replied to [one of the suggestions here](https://github.com/Automattic/studio/pull/2132#discussion_r2584112136). I have tested it again, but this will benefit from another testing and some review after latest changes. 

Could you please give another round of review and let me know if you find any issues? 🙏

fredrikekelund:
Taking a look now 👀

src/hooks/sync-sites/use-listen-deep-link-connection.ts
fredrikekelund: Sounds reasonable to tackle this in a follow-up. For context, my main goal here was to consolidate "business logic" in the Redux slices as much as possible, and make custom hooks and components passive consumers.

In theory, the only reason this needs to be a hook rather than an event listener in the Redux slice is that `selectedTab` and `selectedSiteId` state live in React contexts, so we can't access them from the Redux slice. If we could, it would make more sense to have components that simply say "connect this site ID", and the Redux mutation would take care of everything behind the scenes.

src/components/publish-site-button.tsx
```diff
@@ -1,18 +1,22 @@
...
+import { ConnectButton } from 'src/modules/sync/components/connect-button';
 import { useAppDispatch } from 'src/stores';
 import {
 	connectedSitesActions,
 	useGetConnectedSitesForLocalSiteQuery,
 } from 'src/stores/sync/connected-sites';
 
-export const PublishSiteButton = () => {
+export const PublishSiteButton = ( {
+	redirectToSync = true,
```
fredrikekelund: Just curious: it looks like the only place we pass `false` for this prop is on the sync tab. Do we really need it if that's the case?

src/components/publish-site-button.tsx
```diff
@@ -24,30 +28,33 @@ export const PublishSiteButton = () => {
...
+			icon={ cloudUpload }
+			connectSite={ handlePublishClick }
+			disabled={ isAnySiteSyncing }
+			tooltipText={
+				isAnySiteSyncing
+					? __(
+							'Another site is syncing. Please wait for the sync to finish before you publish your site.'
+					  )
+					: __( 'Publishing your site requires an internet connection.' )
+			}
```
fredrikekelund, line 47:
```suggestion
			disabledTooltipText={ __(
				'Another site is syncing. Please wait for the sync to finish before you publish your site.'
			) }
			offlineTooltipText={ __( 'Publishing your site requires an internet connection.' ) }
```
This is mostly a suggestion, but I think this would clarify how and when which tooltip text is displayed in `ConnectButton`.

src/modules/sync/index.tsx
```diff
@@ -267,44 +239,16 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
...
+				<SyncSitesModalSelector
+					mode={ reduxModalMode || 'connect' }
+					onRequestClose={ () => {
+						dispatch( connectedSitesActions.closeModal() );
+					} }
+					onConnect={ async ( siteId: number ) => {
+						await handleSiteSelection( siteId );
+					} }
+					selectedSite={ selectedSite }
+				/>
```
fredrikekelund, line 225: This is great 👍

src/modules/sync/tests/index.test.tsx
```diff
@@ -115,10 +118,11 @@ describe( 'ContentTabSync', () => {
...
 		currentMock.getConnectedWpcomSites.mockResolvedValue( connectedSites );
 
-		( useFetchWpComSites as jest.Mock ).mockReturnValue( {
-			syncSites,
+		( useGetWpComSitesQuery as jest.Mock ).mockReturnValue( {
+			data: syncSites,
+			isLoading: false,
 			isFetching: false,
-			refetchSites: jest.fn(),
+			refetch: jest.fn().mockResolvedValue( { data: syncSites } ),
```
fredrikekelund: Nit, we could remove these lines, since we only consume `data` in the component

src/modules/sync/tests/index.test.tsx
```diff
@@ -179,10 +183,11 @@ describe( 'ContentTabSync', () => {
...
 		} );
 
-		( useFetchWpComSites as jest.Mock ).mockReturnValue( {
-			syncSites: [],
+		( useGetWpComSitesQuery as jest.Mock ).mockReturnValue( {
+			data: [],
+			isLoading: false,
 			isFetching: false,
-			refetchSites: jest.fn(),
+			refetch: jest.fn().mockResolvedValue( { data: [] } ),
```
fredrikekelund: Same thing here about removing

src/stores/sync/wpcom-sites.ts
```diff
@@ -0,0 +1,197 @@
...
+					return {
+						error: {
+							status: 500,
+							data: error,
+						},
+					};
+				}
+			},
+			providesTags: ( _result, _error, arg ) => [ { type: 'WpComSites', userId: arg.userId } ],
+			keepUnusedDataFor: 60,
```
fredrikekelund:
```suggestion
```
60 is already the default, so we could remove this.

fredrikekelund (approved):
LGTM 👍 This tests well, and the code looks good. I left a couple of small suggestions, but nothing blocking. My biggest concern was about the `tooltipText` prop for `ConnectButton`.

src/components/publish-site-button.tsx
epeicher: We need it for these changes as we have two buttons that behave differently (one redirects and the Sync one does not redirect), but that prop will be redundant once this PR https://github.com/Automattic/studio/pull/2171 lands, so we can remove that prop after the merge of the PR.

src/stores/sync/wpcom-sites.ts
epeicher: [That's right](https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#reducing-subscription-time-with-keepunuseddatafor), removing it 😄

src/components/publish-site-button.tsx
epeicher: I like this suggestion but I would delay adding changes to this until https://github.com/Automattic/studio/pull/2171 is merged, at that point we can also refactor the `publish-site-button.tsx` as [suggested here](https://github.com/Automattic/studio/pull/2132#discussion_r2584459482)

sejas (approved):
Works great. I confirm that it loads the connected sites correctly. I also tested it by logging in and logging out. And changing between accounts with and without sites.
I confirm the list of sites load correctly, they get refreshed and if the user doesn't have sites a CTA modal replaces the connect sites.
I didn't see any warning or error in the console.

Great job!!

e2e/localization.test.ts
```diff
@@ -4,7 +4,6 @@ import AddSiteModal from './page-objects/add-site-modal';
 import MainSidebar from './page-objects/main-sidebar';
 import Onboarding from './page-objects/onboarding';
 import SiteContent from './page-objects/site-content';
-import WhatsNewModal from './page-objects/whats-new-modal';
```
epeicher, line 7: This is to fix a lint error, it not used

e2e/overview-customize-links.test.ts
```diff
@@ -2,7 +2,6 @@ import { test, expect, type Page } from '@playwright/test';
 import { E2ESession } from './e2e-helpers';
 import Onboarding from './page-objects/onboarding';
 import SiteContent from './page-objects/site-content';
-import WhatsNewModal from './page-objects/whats-new-modal';
```
epeicher, line 5: This is to fix a lint error, it not used

src/components/publish-site-button.tsx
epeicher: After mergint #2171, I have done the refactor as part of ed59c461

src/hooks/sync-sites/use-listen-deep-link-connection.ts
epeicher: FYI, I've created STU-1086 for that

PR #2136, opened 2025-11-26
Studio: Add a warning when pushing a site with an outdated WordPress version

katinthehatsite:
## Related issues

Closes STU-875

## Proposed Changes

This PR adds a notice for the user when the PHP or WordPress versions on their Studio site are not the latest and are different from the versions used on WP.com in the Sync modal for the push process.

## Testing Instructions

* Change the WordPress version on one of your sites to be something like `6.4`
* Navigate to the Sync tab
* Connect at least one WP.com site
* Click on the `Push` option
* Confirm that you can see the warning like indicated below:

Screenshot 2025-11-26 at 2 08 56 PM

* Confirm that the modal works well and expands correctly when you have the file tree fully expanded
* You can also test the case when the site both uses an older PHP version or WordPress version and is over 2GB limit:

Screenshot 2025-11-26 at 2 08 39 PM

## Pre-merge Checklist


src/modules/sync/components/sync-dialog.tsx
```diff
@@ -334,6 +362,15 @@ export function SyncDialog( {
...
 						</Notice>
 					) }
+					{ shouldShowVersionMismatch && (
+						<Notice status="warning" isDismissible={ false } className="mb-4">
+							<p data-testid="push-version-mismatch-notice">
+								{ __(
+									'Your Studio site is using a different WordPress or PHP version than your WordPress.com site. The remote site will keep on using the newest supported versions.'
+								) }
+							</p>
+						</Notice>
```
katinthehatsite, line 396: We could use "info" version vs. "warning" and then it would look like this:

Image

To me, it seemed like the background was too subtle so I left it as warning but we can change this if needed.

src/modules/preview-site/components/create-preview-button.tsx
```diff
@@ -51,7 +51,9 @@ export function CreatePreviewButton( { onClick, selectedSite, user }: CreatePrev
 	const isCurrentSiteArchiving = !! activeOperationsForCurrentSite;
 	const isOtherSiteArchiving = isAnySiteArchiving && ! isCurrentSiteArchiving;
 
-	const latestWpVersion = wpVersions.find( ( version ) => version.value === 'latest' )?.value;
+	const latestWpVersion = wpVersions.find(
+		( version ) => version.value !== 'latest' && ! version.isBeta && ! version.isDevelopment
```
katinthehatsite: The tooltip on the `Create preview button` was broken for the semver comparison of versions so I am fixing it at the same time. Alternative fix, could be to add another prop to `interface WordPressVersion` with the actual version but it seemed like a bit of extra for this one edge case with `latest`.  (that's what I initially tried)

The reason why this is broken: We need value: 'latest' for the dropdown selector, but we need the actual version number like '6.7.1' for semver comparisons. Before this fix, the version mismatch detection was completely broken because it was trying to compare '6.2' with the string 'latest'.

katinthehatsite:
One last thing is that we can potentially combine these two messages in one notice when both have to be displayed but I kept them separate for now as they are relevant for two different things:

Screenshot 2025-11-26 at 2 08 39 PM

src/modules/sync/components/sync-dialog.tsx
ivan-ottinger: To me the "warning" variant stands out more as well. 👍🏼

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -149,6 +154,20 @@ export function SyncDialog( {
 	const { fetchChildren, rewindId, isLoadingRewindId, isErrorRewindId, isLoadingLocalFileTree } =
 		useDynamicTreeState( type, localSite.id, remoteSite.id, setTreeState );
 
+	const [ wpVersion ] = useGetWpVersion( localSite );
+	const minimumWordPressVersion = useRootSelector( selectMinimumWordPressVersion );
+	const { data: wpVersions = [] } = useGetWordPressVersions( {
+		minimumVersion: minimumWordPressVersion,
+	} );
+	const latestWpVersion = getLatestStableWpVersion( wpVersions );
```
ivan-ottinger: Maybe I am missing something, but what if we did something like this?

```suggestion
	const { data: wpVersions = [] } = useGetWordPressVersions( { minimumVersion: '' } );
	const latestWpVersion = getLatestStableWpVersion( wpVersions );
```

ivan-ottinger (commented):
Nice! The changes work as expected. 👌🏼 

---

One thing that came to my mind is how the scrolling would look like if we added a similar opacity as we have on the Sync tab:

Image

I believe it could help users noticed the content is scrollable in cases like this one:

Image

At the same time, this may be outside of scope for this PR.

---

> One last thing is that we can potentially combine these two messages in one notice when both have to be displayed but I kept them separate for now as they are relevant for two different things:

I agree and think it makes sense to keep them separate. 👍🏼 

---

I left one inline comment. Also noticed that unit tests are failing.

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -218,13 +237,20 @@ export function SyncDialog( {
...
+		noticeCount === 0 ? 'pb-[70px]' : noticeCount === 1 ? 'pb-[140px]' : 'pb-[210px]';
+
 	return (
 		<Modal
 			className="w-3/5 min-w-[550px] max-h-[84vh] [&>div]:!p-0"
 			onRequestClose={ onRequestClose }
 			title={ syncTexts.title }
 		>
-			<div className={ isPushSelectionOverLimit ? 'pb-[140px]' : 'pb-[70px]' }>
+			<div className={ footerPadding }>
```
sejas: Not a strong opinion, but we could declare the footer padding inside the class. Something like:
```
			<div
				className={ cx(
					noticeCount === 0 && 'pb-[70px]',
					noticeCount === 1 && 'pb-[140px]',
					noticeCount >= 2 && 'pb-[210px]'
				) }
			>
```

sejas (commented):
Great work Kat! I saw in the Linear issue that the notice was preferred over the tooltip 👍 .

I only noticed that the Database checkbox appears a bit cut. Could we show it all?

Image

src/modules/sync/components/sync-dialog.tsx
katinthehatsite: I agree, let's leave this as is for now. We can adjust it later if feedback related to this comes in 👍

katinthehatsite:
> I only noticed that the Database checkbox appears a bit cut. Could we show it all?

Nice, good catch! I was testing on bigger screen so it looked fully fitting there. I will adjust the padding a bit more 👍 

Updated in https://github.com/Automattic/studio/pull/2136/commits/e423be913e33d91093d37a36463ed3f188462ed9

src/modules/sync/components/sync-dialog.tsx
katinthehatsite: Good catch! Adjusted in https://github.com/Automattic/studio/pull/2136/commits/aa1cac95812c157666e324c45a489293cbd20950

katinthehatsite:
@ivan-ottinger I also added a background similar to the other footer for "Connect another site" . Feel free to give it another review 👍

ivan-ottinger (approved):
> @ivan-ottinger I also added a background similar to the other footer for "Connect another site" . Feel free to give it another review 👍

Thank you for the updates, Kat.

The changes look good and the notice is rendering correctly. 👍🏼 

Image

PR #2137, opened 2025-11-26
Move logic from `src/ipc-handlers` to separate modules

fredrikekelund:
## Related issues

- Fixes STU-1054

## Proposed Changes

This PR moves several functions from `src/ipc-handlers.ts` to `src/modules/*/lib/ipc-handlers.ts` in the relevant modules. `src/ipc-handlers.ts` is bloated, and we could easily split it into multiple files. 

The reason I still re-export the functions from the separate modules is that `src/ipc-handlers.ts` has special treatment in `src/ipc-types.d.ts`. Because IPC functions are exposed to the front-end as a single API, it makes sense to maintain this single file for exporting all methods. 

## Testing Instructions

Tests should pass and light smoke testing of the app shouldn't reveal any issues. 

## Pre-merge Checklist


gcsecsey (approved):
Thanks for carrying out the refactor @fredrikekelund! The changes overall look solid to me, and preserve the interface of ipc-handlers. 👌 

I also tested various features in Studio, and all of these continue to work:
- start/stop site
- site creation
- preview site creation
- pull
- push
- import/export
- updating settings

epeicher (approved):
Thanks @fredrikekelund for improving this and for the great explanation. I think it improves maintenance 👌.  I have smoked tested the affected areas, and I have not found any issues. LGTM! :shipit:

PR #2134, opened 2025-11-26
Added `studio site status` CLI command

fredrikekelund:
## Related issues

- Fixes STU-1018

## Proposed Changes

This PR adds a `studio site status` command that outputs an extensive list of details related to a Studio site:

- Site URL
- Auto Login URL (if the site is running)
- Site Path
- Status (online/offline)
- PHP Version
- WP Version
- Admin Username
- Admin Password

## Testing Instructions

1. Run `npm run cli:build`
4. Run `node dist/cli/main.js site status --path PATH_TO_SITE` (where `PATH_TO_SITE` is the path to a Studio site)
5. Ensure that the output is good
6. Run `node dist/cli/main.js site start --path PATH_TO_SITE`
7. Run `node dist/cli/main.js site status --path PATH_TO_SITE`
8. Ensure that the site status is reflected in the CLI output

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2140, opened 2025-11-26
Studio: Implement studio site stop CLI command

bcotrim:
## Related issues

- Fixes STU-951

## Proposed Changes

- Implemented `studio site stop` CLI command for gracefully stopping local WordPress sites
- Added graceful shutdown mechanism via IPC message to child process with 5s timeout before falling back to PM2 delete
- Clears `latestCliPid` from appdata

## Testing Instructions

- Create a test site: `node dist/cli/main.js site create --path path`
- Stop the site: `node dist/cli/main.js site stop --path path`
- Verify the site is stopped and no longer running
- Verify the `latestCliPid` is added on start and removed when site is stopped
- Test error scenarios: try stopping a non-existent site, already stopped site

## Pre-merge Checklist

### DIFF
--- a/cli/commands/site/stop.ts
+++ b/cli/commands/site/stop.ts
@@ -1,7 +1,6 @@
 import { __ } from '@wordpress/i18n';
-import { arePathsEqual } from 'common/lib/fs-utils';
 import { SiteCommandLoggerAction as LoggerAction } from 'common/logger-actions';
-import { clearSiteLatestCliPid, readAppdata } from 'cli/lib/appdata';
+import { clearSiteLatestCliPid, getSiteByFolder } from 'cli/lib/appdata';
 import { connect, disconnect } from 'cli/lib/pm2-manager';
 import { stopProxyIfNoSitesNeedIt } from 'cli/lib/site-utils';
 import { isServerRunning, stopWordPressServer } from 'cli/lib/wordpress-server-manager';
@@ -12,12 +11,7 @@ export async function runCommand( siteFolder: string ): Promise< void > {
 	const logger = new Logger< LoggerAction >();
 
 	try {
-		const appdata = await readAppdata();
-		const site = appdata.sites.find( ( s ) => arePathsEqual( s.path, siteFolder ) );
-
-		if ( ! site ) {
-			throw new LoggerError( __( 'Could not find Studio site.' ) );
-		}
+		const site = await getSiteByFolder( siteFolder );
 
 		await connect();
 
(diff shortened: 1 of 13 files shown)


cli/commands/site/stop.ts
```diff
@@ -0,0 +1,60 @@
...
+import { isServerRunning, stopWordPressServer } from 'cli/lib/wordpress-server-manager';
+import { Logger, LoggerError } from 'cli/logger';
+import { StudioArgv } from 'cli/types';
+
+export async function runCommand( siteFolder: string ): Promise< void > {
+	const logger = new Logger< LoggerAction >();
+
+	try {
+		const appdata = await readAppdata();
+		const site = appdata.sites.find( ( s ) => arePathsEqual( s.path, siteFolder ) );
```
fredrikekelund: Let's use `getSiteByFolder` here once https://github.com/Automattic/studio/pull/2134 has landed.

cli/commands/site/stop.ts
```diff
@@ -0,0 +1,60 @@
...
+	}
+}
+
+export const registerCommand = ( yargs: StudioArgv ) => {
+	return yargs.command( {
+		command: 'stop',
+		describe: __( 'Stop local site' ),
+		builder: ( yargs ) => {
+			return yargs;
+		},
```
fredrikekelund, line 51:
```suggestion
```

fredrikekelund (commented):
This is great overall 👍 My only issues with the implementation were that `sendStopMessage` was unnecessary, and that we shouldn't use the `as` keyword for any of the types from `cli/lib/types/wordpress-server-ipc.ts`. I went ahead and pushed a fix for both of those issues. 

There's one more thing I think we should look into before landing this, and that is stopping the proxy server if no remaining running sites depend on it.

bcotrim:
@fredrikekelund thanks for the review! I've updated the code to stop the proxy if no longer needed.
This should be ready for another review

fredrikekelund (approved):
Brilliant! Clean change, works beautifully 👍

cli/commands/site/stop.ts
```diff
@@ -0,0 +1,56 @@
...
+import { stopProxyIfNoSitesNeedIt } from 'cli/lib/site-utils';
+import { isServerRunning, stopWordPressServer } from 'cli/lib/wordpress-server-manager';
+import { Logger, LoggerError } from 'cli/logger';
+import { StudioArgv } from 'cli/types';
+
+export async function runCommand( siteFolder: string ): Promise< void > {
+	const logger = new Logger< LoggerAction >();
+
+	try {
+		const site = await getSiteByFolder( siteFolder );
```
fredrikekelund:
```suggestion
		const site = await getSiteByFolder( siteFolder, false );
```
This should fix the unit tests

PR #2139, opened 2025-11-26
Add wpcom to onboarding

nightnei:
## Related issues

- Fixes STU-972

## Proposed Changes
With this PR we are adding wpcom as option for onboarding step

## Testing Instructions
1. Open `appdata-v1.json`, set `"onboardingCompleted": false` and remove `authToken` if you have it
2. Open Studio
3. Assert that you see the next screen<br />Screenshot 2025-11-27 at 15 11 14
4. Click on "Skip"
5. Assert that onboarding is closed and you see initial Studio's empty page for creating new site
6. Repeat 1-st step to clean Studio
7. Click on "Create a free account"
8. Assert that you were landed to the appropriate wpcom page
9. Next, click on Log in and proceed
10. Assert that you landed to Studio and you see again Studio's empty page

PR #2147, opened 2025-11-27
Clean up siteForm from legacy

nightnei:
## Proposed Changes
While working on [Add wpcom to onboarding](https://github.com/Automattic/studio/pull/2139) I noticed:
1. We have a few [unused properties](https://github.com/Automattic/studio/pull/2147/files#diff-1642755f7f398091d0ea9bb2a219e293098cd01658801fee72c0a94517b0740eL64-L69), it seems they are from times when we had import process located inside the form for creating a site.
2. Also after the previously mentioned PR we have [2 more unused properties](https://github.com/Automattic/studio/pull/2147/files#diff-1642755f7f398091d0ea9bb2a219e293098cd01658801fee72c0a94517b0740eL56-L57) now

So I decided to clean it up.

I was also thinking about moving `create-site-form.tsx` inside `modules/add-site-form`, and also merging it with `modules/add-site/components/create-site`, but to keep this PR easy for review, I decided that it's better not to do it right now.

I was also thinking about merging create-site and create-site-form, but to make this PR easier for review, I decided not to do it. It's not something very critical, but more important to make the PR diff clearer and easier for review.

## Testing Instructions
1. Run Studio
2. Create a new site 
4. Assert that it was created and no errors in browser's console


sejas (approved):
Yay! great catch. Thanks for cleaning the code.
I tested it by importing a new site.

https://github.com/user-attachments/assets/50cccd16-59fd-4b9b-8452-d77cdc83800a

Let's also fix unit tests before merging.

src/modules/add-site/components/create-site-form.tsx
```diff
@@ -33,40 +25,23 @@ interface FormPathInputComponentProps {
...
-	placeholder?: string;
-}
-
 interface SiteFormErrorProps {
 	error?: string;
 	tipMessage?: string;
 	className?: string;
 }
 
 interface SiteFormProps {
```
sejas, line 37: Copilot made a great suggestions. We can rename the props name to match the renamed component.

```suggestion
interface CreateSiteFormProps {
```

fredrikekelund:
This is just a suggestion, and it does increase the scope of this refactor, but this component deserves a refactor to reduce the props drilling even further, IMO. 

The ideal scenario would be if `CreateSiteForm` received a `features` prop to determine which features to display in the form, and an `onSubmit` handler that receives an object with all the form values. It makes more sense for the state to live alongside the form controls than in the `useAddSite` hook, at the `AddSite` component level.

nightnei:
> This is just a suggestion, and it does increase the scope of this refactor, but this component deserves a refactor to reduce the props drilling even further, IMO.
The ideal scenario would be if CreateSiteForm received a features prop to determine which features to display in the form, and an onSubmit handler that receives an object with all the form values. It makes more sense for the state to live alongside the form controls than in the useAddSite hook, at the AddSite component level.

Great idea, but let's do it later as a separate PR

fredrikekelund:
> let's do it later as a separate PR

I created STU-1099 to track this

PR #1710, opened 2025-08-25
Preserve temporary site state during site list updates

ivan-ottinger:
## Related issues

- Related to STU-690
- related discussion: p1756125475562769-slack-C06DRMD6VPZ

## Proposed Changes

- ensure that `updatedSites` include also a site that is currently being created

## Testing Instructions

1. Check out the PR branch and build the app.
2. Ensure you have multiple sites.
3. Stop them.
4. Quickly start them one-by-one and initiate a new site creation right when they are starting.
5. The process should work correctly. It should not be possible to reproduce the issue mentioned in p1756125475562769-slack-C06DRMD6VPZ.

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2148, opened 2025-11-27
Studio: Refactor `studio preview create` to adjust new site logic

bcotrim:
## Related issues

- Fixes STU-991

## Proposed Changes

- `studio preview create` now requires the target folder to be an existing Studio site, failing with a helpful error if not
- Removed `newSites` concept from app and cli entirely

## Testing Instructions

- `npm start`
- Create a Studio site and remove it without deleting files
- Close Studio
- `node dist/cli/main.js preview --path <new site folder>`
- Verify it fails.
- Verify `studio preview create <existing-site-folder>` works correctly for existing Studio sites

## Pre-merge Checklist


cli/commands/preview/create.ts
```diff
@@ -21,6 +21,7 @@ export async function runCommand( siteFolder: string ): Promise< void > {
 	try {
 		logger.reportStart( LoggerAction.VALIDATE, __( 'Validating…' ) );
 		validateSiteFolder( siteFolder );
```
```suggestion
```
I think this can be removed. No need to validate the directory type when creating a site if we also use `getSiteByFolder` to ensure the current directory is added to Studio.

cli/commands/preview/tests/create.test.ts
fredrikekelund, line 1: Let's remove the `validateSiteFolder` mocking from here, too.

cli/commands/preview/update.ts
fredrikekelund, line 1: We should remove the `validateSiteFolder` call in `runCommand` here as well and replace it with a `getSiteByFolder` call.

cli/commands/preview/tests/update.test.ts
fredrikekelund, line 1: Same thing here about removing the `validateSiteFolder` mocking.

src/hooks/use-site-details.tsx
```diff
@@ -427,7 +426,7 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 
 	useEffect( () => {
 		const unsubscribe = window.ipcListener.subscribe( 'user-data-updated', async ( _, payload ) => {
```
fredrikekelund: Could we remove this event listener altogether now..? This might be a useful starting point for picking up `latestCliPid` changes, but all the logic around `addingSiteIds` seems to ultimately relate to `newSites`, which makes me think that we could clean up this file further.

fredrikekelund (commented):
Such a relief to clean this upp! I left a couple of comments. The most substantial one relates to the `user-data-updated` event listener in `src/hooks/use-site-details.tsx`.

src/hooks/use-site-details.tsx
bcotrim: I think it is related to the add site workflow from Studio, not the `newSites` workflow.
Since we will need to revist this code when we work on having Studio depending on Studio CLI for site management, I suggest adjusting this file at that time.
What do you think?

fredrikekelund: This was originally implemented by you in #1272 (d38ad878f27a6e7c20946863976364f12f920546 in response to https://github.com/Automattic/studio/pull/1272#pullrequestreview-2794057011). @ivan-ottinger then updated this listener in https://github.com/Automattic/studio/pull/1710 to fix a bug where sites would disappear from the sidebar while being created. 

The current logic has no purpose, really. If I attempt to summarize it, this listener will:
1. Trigger any time the appdata file changes
2. Load the list of sites from appdata, instantiating each of them as `SiteServer`
3. Proceed if `appdata.sites` is truthy (which it always is, because it's an array)
4. If there's no site currently being added by Studio, return early
5. If there IS a site currently being added to Studio, recreate the `data` state by adding the temporary site to the list of sites from appdata

We should remove this.

fredrikekelund (approved):
Thanks for the changes, @bcotrim, this looks good now 👍

I identified two remaining issues: that we were still using `validateSiteFolder` in the `preview list` command, and that the `user-data-updated` should be removed after all. Since I was planning to land this today (while you're AFK), I went ahead and pushed fixes for both of those.

src/hooks/use-site-details.tsx
```diff
@@ -425,33 +424,6 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
-
-					return updatedSites;
-				} );
-			}
-		} );
-
-		return () => {
-			unsubscribe();
-		};
-	}, [ addingSiteIds ] );
```
sejas, line 453: I think this listener is necessary. The site creation in Studio app is currently broken without it.

sejas (commented):
I noticed an issue when creating a new site. I created a Linear issue STU-1071, and I'm working on a quick fix.

https://github.com/user-attachments/assets/701e527a-6a09-4dd5-8329-4acd51f28dde

src/hooks/use-site-details.tsx
sejas: Quickfix PR https://github.com/Automattic/studio/pull/2161

fredrikekelund: Follow-up in https://github.com/Automattic/studio/pull/2164

PR #2149, opened 2025-11-28
Replace empty screen with add site flow and rename it to NoStudioSites

sejas:
## Related issues

- Fixes STU-1063

## Proposed Changes

- Remove the empty screen and replace it with add site options

## Testing Instructions

- Remove all your sites from Studio
- Observe the new screen
- Continue the flow to create any of the sites
- Confirm the site was correctly created

| Before | After |
|--------|--------|
|  Screenshot 2025-11-27 at 11 27 50| Screenshot 2025-11-28 at 00 21 26 |

## Pre-merge Checklist


gavande1 (changes requested):
Thanks @sejas for working on this. I have tested this found couple of issues. 

#### The whats new model appeared 
As soon as open Studio, the model appeared. I found bit odd to see that but might not be related this PR. Perhaps, we should add checks to prevent this model appearing on other screens. 

Image

#### The actions were not clickable
I could see the replaced "Add new site" screen on empty state but none of the links/buttons were clickable. I think it has to do something with making window movable by moving it. 

https://github.com/user-attachments/assets/0f6394f1-0982-4f0c-b534-cad524c98c20

sejas:
@gavande1 , thanks for the early review. Could you take another look.

> The whats new model appeared

I created a new issue STU-1066 so any decision won't block from merging this PR. 

> The actions were not clickable

This is fixed! Could you test it again?

epeicher (approved):
Thanks @sejas, this streamlines the flow to create a site by using the different options. I have tested it, and it works as expected, code changes also LGTM! :shipit: 

Image

P.S. I agree with the comment in the linked task about displaying the top bar, which can be helpful if the user has logged in in a previous step. But that is something that can be discussed for a follow-up

katinthehatsite (commented):
Should we disable then the `Add site` button in the topbar menu when this screen for adding a site is showing since it does nothing in that case?

Image

katinthehatsite (commented):
The changes look good 👍 , I just have one minor suggestion about disabling the `Add site` button in the top navigation

src/components/empty-studio/index.tsx
```diff
@@ -1,26 +1,14 @@
 import { useI18n } from '@wordpress/react-i18n';
-import emptyStudioIllustration from 'src/components/empty-studio/empty-studio.svg';
-import AddSite from 'src/modules/add-site';
+import { AddSiteModalContent } from 'src/modules/add-site';
 
 export function EmptyStudio() {
```
katinthehatsite: I am wondering if we might take the opportunity here to rename the `EmptyStudio` to something a bit more self-explanatory like `NoStudioSites` or similar. What do you think?

sejas:
> The changes look good 👍 , I just have one minor suggestion about disabling the `Add site` button in the top navigation

I created an issue to tackle that as a follow-up. STU-1068

src/components/empty-studio/index.tsx
sejas: Done! https://github.com/Automattic/studio/pull/2149/commits/e0b68e66b17b3a2648113d5f976977c23bdba5d0

katinthehatsite: looks good 👍

katinthehatsite (approved):

PR #2151, opened 2025-11-28
Added `studio site set-domain` and `studio site set-https` commands

fredrikekelund:
## Related issues

- Fixes STU-955
- Fixes STU-956

## Proposed Changes

- Added `studio site set-domain` command, which updates the appdata file and restarts the site if running. 
- Added `studio site set-https` command, which updates the appdata file and restarts the site if running. 

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js site set-https true --path PATH_TO_SITE` (where `PATH_TO_SITE` is the path to a Studio site WITH a custom domain)
5. Ensure that it succeeds
6. Run `node dist/cli/main.js site start --path PATH_TO_SITE`
7. Run `node dist/cli/main.js site set-https false --path PATH_TO_SITE`
8. Ensure that the site restarts successfully
9. Run `node dist/cli/main.js site set-domain example.com --path PATH_TO_SITE`
10. Ensure that the command returns an error about the domain name format
11. Run `node dist/cli/main.js site set-domain example.local --path PATH_TO_SITE`
12. Ensure that the site restarts successfully

## Pre-merge Checklist


cli/commands/site/set-domain.ts
```diff
@@ -0,0 +1,104 @@
...
+			const domainError = getDomainNameValidationError( true, domainName, existingDomains );
+
+			if ( domainError ) {
+				throw new LoggerError( domainError );
+			}
+
+			const site = appdata.sites.find( ( site ) => arePathsEqual( site.path, sitePath ) );
+			if ( ! site ) {
+				throw new LoggerError( __( 'The specified folder is not added to Studio.' ) );
+			}
```
bcotrim: I don't think we need both `getSiteByFolder` and this validation, I understand we want to search the site in appdata to make the changes, so we could remove `getSiteByFolder`?

cli/commands/site/tests/set-https.test.ts
```diff
@@ -0,0 +1,247 @@
+import { arePathsEqual } from 'common/lib/fs-utils';
+import { SiteCommandLoggerAction as LoggerAction } from 'common/logger-actions';
```
bcotrim: lint shows `warning  'LoggerAction' is defined but never used.`

cli/commands/site/set-https.ts
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		logger.reportStart( LoggerAction.START_DAEMON, __( 'Starting process daemon...' ) );
+		await connect();
+		logger.reportSuccess( __( 'Process daemon started' ) );
+
+		try {
+			await lockAppdata();
+			const appdata = await readAppdata();
+			const site = appdata.sites.find( ( site ) => arePathsEqual( site.path, sitePath ) );
```
bcotrim: Same as my comment on set-domain

bcotrim (changes requested):
Adding a custom domain for a running site isn't working for me, likely because `await setupCustomDomain( site, logger );` is using the site from `const site = await getSiteByFolder( sitePath, false );` that doesn't have a customDomain.

Cleaning up the logic to fetch the site (as I mentioned in the comments) should also fix this issue

cli/commands/site/set-domain.ts
fredrikekelund: Good call 👍

fredrikekelund:
> Adding a custom domain for a running site isn't working for me, likely because `await setupCustomDomain( site, logger );` is using the site from `const site = await getSiteByFolder( sitePath, false );` that doesn't have a customDomain.

Good catch 👍 I've updated the PR accordingly

bcotrim (approved):
Minor issue I noticed:
- When changing the domain of a site that already has a custom domain, we prompt for the admin password, twice.
We could avoid this by doing the change in one step "replace domain" instead of remove first and add later.
What do you think?

Code looks good and works as expected 👍

fredrikekelund:
> When changing the domain of a site that already has a custom domain, we prompt for the admin password, twice

I agree, this is annoying. I created STU-1077 to fix this in a follow-up PR

PR #2153, opened 2025-11-28
Fix regression related to Windows CLI installer

fredrikekelund:
## Related issues

- Related to #2041

## Proposed Changes

In #2041, we changed the CLI installer's behavior on Windows so that it no longer installs automatically when the app starts. Users can now install the CLI through the Studio settings modal. 

Unfortunately, my PR also introduced a regression, which is that the `%LOCALAPPDATA%\studio\bin\studio.bat` script would no longer be updated automatically after Studio updates. 

The `%LOCALAPPDATA%\studio` directory on Windows is structured so that the `resources` directory lives inside a versioned directory. For example, I have both `%LOCALAPPDATA%\studio\app-1.6.4` and `%LOCALAPPDATA%\studio\app-1.6.5` on my machine. Every time the app updates, a new directory is created, and we need to update the `%LOCALAPPDATA%\studio\bin\studio.bat` file to point to the newest directory (see #1152 for the original implementation).

## Testing Instructions

Code review should suffice

## Pre-merge Checklist

### DIFF
--- a/src/modules/cli/lib/windows-installation-manager.ts
+++ b/src/modules/cli/lib/windows-installation-manager.ts
@@ -17,9 +17,6 @@ const currentUserRegistry = new Registry( {
 	key: '\\Environment',
 } );
 
-/**
- * Windows implementation of CLI installation manager
- */
 export class WindowsCliInstallationManager implements StudioCliInstallationManager {
 	constructor() {
 		if ( process.platform !== 'win32' ) {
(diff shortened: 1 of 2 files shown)


src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -17,9 +17,6 @@ const currentUserRegistry = new Registry( {
 	key: '\\Environment',
 } );
 
-/**
- * Windows implementation of CLI installation manager
```
epeicher, line 21: 👍 The comment is almost the name of the class 😄

epeicher (approved):
LGTM!

bcotrim (approved):
LGTM 👍

src/modules/cli/lib/windows-installation-manager.ts
fredrikekelund: AI 😐

PR #2154, opened 2025-11-28
CLI: Implement `studio site set-php-version`

fredrikekelund:
## Related issues

- Fixes STU-953

## Proposed Changes

- Added a `studio site set-php-version` command, which updates the appdata file and restarts the site if running.

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js site set-php-version true --path PATH_TO_SITE` (where `PATH_TO_SITE` is the path to a Studio site)
5. Ensure that it succeeds
6. Run `node dist/cli/main.js site start --path PATH_TO_SITE`
7. Run `node dist/cli/main.js site set-php-version 8.2 --path PATH_TO_SITE`
8. Ensure that the site restarts successfully (assuming the site wasn't already running PHP 8.2, in which case it will return an error)

## Pre-merge Checklist


bcotrim (commented):
Code LGTM and works as described.

  While reviewing, I wondered if we should have a studio site update command instead of separate `set-*` commands.

  Advantages:
  1. Consistent with `studio create` - Same option names (`--php`, `--domain`, `--https`)
  2. Batch updates - One command, one restart instead of potentially 3
  3. Less code - Shared boilerplate (load site → lock appdata → update → restart if running)

  What do you think?

fredrikekelund:
>   While reviewing, I wondered if we should have a studio site update command instead of separate `set-*` commands.

I like that idea 👍 I have a whole train of PRs waiting to merge here, though (https://github.com/Automattic/studio/pull/2151, this, and https://github.com/Automattic/studio/pull/2155). I'd prefer to land all three and then circle back to refactor the `set-*` commands into a single one.

bcotrim (approved):

PR #2156, opened 2025-11-28
Remove backward compatibility for onboarding metrics tests

nightnei:
## Proposed Changes
We replaced onboarding in [this PR](https://github.com/Automattic/studio/pull/2139), but we run metrics on PR's branch and trunk branch to compare, so in that PR we added backward compatibility to metrics. And in this PR we are removing it.

## Testing Instructions
All tests on this PR should be green


sejas (approved):
Changes in e2e look good and confirm the pass correctly.

wojtekn (approved):

PR #2158, opened 2025-12-01
Improve CLI `site` subcommand tests

fredrikekelund:
## Related issues

- Fixes STU-1075

## Proposed Changes

The tests in `cli/commands/site/tests` are all integration tests, which means some level of mocking is required, and we focus more on behavior than output. The goal is not mock-free unit tests, but I still found room for improvement in two areas:

1. We consistently inspect logger output to verify function failures. This is not ideal, because logger output is a secondary effect of the failure.
2. Especially for the more complex commands (like `cli/commands/site/create.ts`), we mock A LOT of stuff. 

This PR addresses those concerns by:

1. Moving the `logger.reportError` output to the yargs handlers in the `registerCommand` functions. This allows us to actually expect exceptions to be thrown from `runCommand` calls in the tests. 
2. Remove mock calls where possible. Notably, the logger isn't mocked anywhere in `cli/commands/site/tests` now. `cli/commands/site/create.ts` also reads the blueprint file and parses it as JSON in the yargs handler, reducing the amount of mocking needed in the tests. 

## Testing Instructions

Green unit tests and code review

## Pre-merge Checklist


bcotrim (approved):
Amazing work, tests easier to read and write! 👍 

Image

PR #2160, opened 2025-12-01
Release 1.6.6-beta2

sejas:
## Related issues

Related to pfHvTO-Or-p2

## Proposed Changes

- Release 1.6.6-beta2

## Testing Instructions

- Confirm that the build succeeds

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2161, opened 2025-12-01
Introduce back the user-data-updated listener to quick fix the site creation

sejas:
## Related issues

- Fixes STU-1071

## Proposed Changes

- Quick fix site creation. Before this PR the site creation got stuck and never updated the UI.
- The issue was introduced in  #2148 
- That's the reason most of e2e tests were failing, and also the Performance CI, as it always failed on trunk.

## Testing Instructions

- Run `npm start`
- Click on Add site
- Observe that the site creation finishes correctly and you see the overview tab after a few seconds.

**Before**

https://github.com/user-attachments/assets/123e83c3-a09d-42ff-84b1-4cb92853c88f

**After**

https://github.com/user-attachments/assets/60281846-984f-426f-9795-26aa3dc389a7

 
## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -446,6 +446,17 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 		};
 	}, [ autoStartSites ] );
 
+	useEffect( () => {
+		const unsubscribe = window.ipcListener.subscribe( 'user-data-updated', async () => {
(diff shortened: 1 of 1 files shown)


sejas:
Merging this fix as I need it to unblock #2160. I can apply any feedback as a follow-up.

I also hope to fix `Performance Metrics Comparison / Run metrics tests` after merging this PR. The e2e passed on Mac, and failed on Windows due a timeout.

fredrikekelund:
I'm looking into the root cause of this

PR #2164, opened 2025-12-02
Remove `user-data-updated` listener in `useSiteDetails`

fredrikekelund:
## Related issues

- Fixes STU-1071
- Follow up to #2161 (which fixed a regression I caused in #2148)

## Proposed Changes

The data flow in and around `useSiteDetails` when creating a new site is pretty gnarly. Previously, we were unintentionally relying on the `user-data-updated` listener to refresh the state after writing to the appdata file. When I removed this listener in #2148, it caused a regression in which the `isAddingSite` property on the new site never cleared (which meant the loading screen would display indefinitely). @sejas merged a hotfix for this in #2161.

The `user-data-updated` listener is meant to let Studio pick up changes from the CLI. We should not rely on it for flows that run entirely in Studio, where we already have all the state easily accessible in memory.

This PR removes the `user-data-updated` listener again and fixes the underlying problem by:

1. Not mutating the `selectedSite` object in the `useMemo` callback (this is where we previously set the `isAddingSite` property, but we didn't have any explicit logic to unset it)
2. Updating the `createSite` function in `useSiteDetails` to accept a `phpVersion` argument. Previously, we would set this in a callback function in `useAddSite` that included a state update in`useSiteDetails` (which would unset `isAddingSite` flag).
3. Removing the `addingSiteIds` state and instead explicitly setting and unsetting the `isAddingSite` flag on the site object in the `createSite` function. 
4. Bonus: not using `useMemo` to store `selectedSite`. The way we did this had no effect.
5. Bonus: renaming the site list state from `data` to `sites` in `useSiteDetails`.

## Testing Instructions

1. Create a new site with all the default settings
2. Ensure that the flow behaves as before and that it finishes normally
3. Create another site with a PHP version different than the default
4. Ensure that the process finishes normally and that the PHP version was persisted correctly (check the Settings tab)
5. Create a site with a blueprint
6. Ensure that the process finishes normally
7. Create a site from an existing remote site (see pdtkmj-4cM-p2)
8. Ensure that the process finishes normally

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-add-site.ts
+++ b/src/hooks/use-add-site.ts
@@ -20,7 +20,7 @@ import type { SyncOption } from 'src/types';
 
 export function useAddSite() {
 	const { __ } = useI18n();
-	const { createSite, data: sites, loadingSites, startServer, updateSite } = useSiteDetails();
+	const { createSite, sites, loadingSites, startServer } = useSiteDetails();
 	const { importFile, clearImportState } = useImportExport();
 	const [ connectSite ] = useConnectSiteMutation();
 	const { pullSite } = useSyncSites();
@@ -126,20 +126,8 @@ export function useAddSite() {
 				usedCustomDomain,
 				useCustomDomain ? enableHttps : false,
 				selectedBlueprint,
+				phpVersion,
 				async ( newSite ) => {
-					let updatedSite = { ...newSite };
-
-					if ( newSite.phpVersion !== phpVersion ) {
-						updatedSite = {
-							...updatedSite,
-							phpVersion,
-						};
-					}
-
-					if ( updatedSite !== newSite ) {
-						await updateSite( updatedSite );
-					}
-
 					if ( fileForImport ) {
 						await importFile( fileForImport, newSite, {
 							showImportNotification: false,
@@ -183,7 +171,6 @@ export function useAddSite() {
 		siteName,
 		sitePath,
 		startServer,
-		updateSite,
 		wpVersion,
 		phpVersion,
 		customDomain,
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -22,7 +22,7 @@ import type { Blueprint } from 'src/stores/wpcom-api';
 interface SiteDetailsContext {
 	selectedSite: SiteDetails | null;
 	updateSite: ( site: SiteDetails ) => Promise< void >;
-	data: SiteDetails[];
+	sites: SiteDetails[];
 	setSelectedSiteId: ( selectedSiteId: string ) => void;
 	createSite: (
 		path: string,
@@ -31,6 +31,7 @@ interface SiteDetailsContext {
 		customDomain?: string,
 		enableHttps?: boolean,
 		blueprint?: Blueprint,
+		phpVersion?: string,
 		callback?: ( site: SiteDetails ) => Promise< void >
 	) => Promise< SiteDetails | void >;
 	startServer: ( id: string ) => Promise< void >;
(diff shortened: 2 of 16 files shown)


src/hooks/use-site-details.tsx
```diff
@@ -472,26 +467,20 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
-		setData( data.map( ( site ) => ( site.running ? { ...site, running: false } : site ) ) );
-	}, [ data ] );
+		setSites( sites.map( ( site ) => ( site.running ? { ...site, running: false } : site ) ) );
+	}, [ sites ] );
 
 	const [ isEditModalOpen, setIsEditModalOpen ] = useState( false );
-	const selectedSite = useMemo( () => {
-		const site = data.find( ( site ) => site.id === selectedSiteId ) || firstSite;
-		if ( addingSiteIds.includes( site?.id ) ) {
-			site.isAddingSite = true;
```
fredrikekelund, line 486: This is where we previously mutated the site object to add the `isAddingSite` property, but we didn't have any explicit logic to unset it.

src/hooks/use-add-site.ts
```diff
@@ -126,20 +126,8 @@ export function useAddSite() {
...
-					if ( newSite.phpVersion !== phpVersion ) {
-						updatedSite = {
-							...updatedSite,
-							phpVersion,
-						};
-					}
-
-					if ( updatedSite !== newSite ) {
-						await updateSite( updatedSite );
-					}
```
fredrikekelund, line 141: This is where we previously set the PHP version and updated the site list state in `useSiteDetails` by calling `updateSite`. 

I removed this logic and opted for passing the PHP version as an argument to `createSite` instead.

src/hooks/use-site-details.tsx
```diff
@@ -308,8 +304,13 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
 				} );
-				setData( ( prevData ) =>
-					prevData.map( ( site ) => ( site.id === tempSiteId ? newSite : site ) )
+				setSites( ( prevData ) =>
+					prevData.map( ( site ) => {
+						if ( site.id === newSite.id ) {
+							return { ...newSite, isAddingSite: true };
+						}
+						return site;
+					} )
```
fredrikekelund, line 313: This call—and the equivalent one just below this in the diff—is what replaces the `addingSiteIds` state.

src/hooks/use-site-details.tsx
```diff
@@ -472,26 +467,20 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
+	}, [ sites ] );
 
 	const [ isEditModalOpen, setIsEditModalOpen ] = useState( false );
-	const selectedSite = useMemo( () => {
-		const site = data.find( ( site ) => site.id === selectedSiteId ) || firstSite;
-		if ( addingSiteIds.includes( site?.id ) ) {
-			site.isAddingSite = true;
-		}
-		return site;
-	}, [ addingSiteIds, data, firstSite, selectedSiteId ] );
```
fredrikekelund, line 489: After removing the `site.isAddingSite = true` assignment, there was no longer any point in using `useMemo` here (see this [classic blog post](https://kentcdodds.com/blog/usememo-and-usecallback))

sejas (approved):
Thanks for creating the proper solution. I confirm the Add site creation works as expected, I also tested editing a the site name and PHP version. Also, thanks for the bonuses. The code is much cleaner.

bcotrim:
Nice cleanup! 👍

PR #2168, opened 2025-12-02
Clean up e2e tests

nightnei:
## Proposed Changes
See comments in the PR below

## Testing Instructions
PR's build should be green

### DIFF
--- a/e2e/page-objects/main-sidebar.ts
+++ b/e2e/page-objects/main-sidebar.ts
@@ -10,7 +10,7 @@ export default class MainSidebar {
 	}
 
 	get addSiteButton() {
-		return this.page.getByTestId( 'add-site-button' );
+		return this.locator.getByTestId( 'add-site-button' );
 	}
 
 	getSiteNavButton( siteName: string ) {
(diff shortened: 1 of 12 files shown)


e2e/page-objects/onboarding.ts
```diff
@@ -42,4 +43,11 @@ export default class Onboarding {
 			localPath,
 		};
 	}
+
+	async closeWhatsNew() {
```
nightnei, line 46: Moved closeWhatsNew to onboarding to avoid duplications and keep consistency

e2e/page-objects/whats-new-modal.ts
```diff
@@ -10,8 +10,4 @@ export default class WhatsNewModal {
 	get closeButton() {
 		return this.locator.getByRole( 'button', { name: 'Close' } );
 	}
-
-	async close() {
```
nightnei, line 14: We don't need it

e2e/page-objects/main-sidebar.ts
```diff
@@ -20,15 +20,13 @@ export default class MainSidebar {
...
 		const isButtonVisible = await this.addSiteButton.isVisible().catch( () => false );
 
-		if ( isButtonVisible ) {
-			// Sites exist - need to open the modal
-			await this.addSiteButton.click();
-			const dialog = new AddSiteModal( this.page );
-			await expect( dialog.locator ).toBeVisible();
-			return dialog;
+		if ( ! isButtonVisible ) {
+			throw new Error( 'Add Site button not found on the main sidebar' );
```
nightnei, line 24: I think it's better to keep this function more specific, since if it's not available in UI then we can mute an error 

Also, with [the current use case](https://github.com/Automattic/studio/pull/2168/files#diff-0c43fce9c70aa804feb4221cdf5a858f01bce48a1cc574a4d8167d74817b9616L23-L24) - there is no such thing as "sidebar" inside onboarding, so it's a bit misleading to create `MainSidebar` and running `openAddSiteModal`, instead of directly running `new AddSiteModal`.

e2e/page-objects/onboarding.ts
epeicher: 💯

epeicher (approved):
Thanks @nightnei for improving this! I can see the CI is green and the changes LGTM! :shipit:

PR #2172, opened 2025-12-02
Studio: Refactor Studio to depend on the CLI for starting sites

bcotrim:
## Related issues

- Fixes STU-1019

## Proposed Changes

- Add `--watch` to `studio site list`
- Add watcher to Studio to listen to site status changes
- Implemented `createCliServerProcess` so Studio can start and stop sites using Studio CLI

## Testing Instructions

- `npm install`
- `npm start`
- Start a new site using Studio
- Confirm site is running
- Stop site using Studio
- Confirm site is not running
- Confirm `studio site list` returns the correct status for the site
- Start site using Studio CLI
- Confirm Studio shows the site as running
- Stop the site using Studio CLI
- Confirm Studio shows the site as stopped

## Pre-merge Checklist


cli/commands/site/list.ts
```diff
@@ -96,16 +147,22 @@ export const registerCommand = ( yargs: StudioArgv ) => {
...
+				.option( 'watch', {
+					type: 'boolean',
+					default: false,
+					description: __( 'Watch for site status changes and update the list in real-time' ),
+				} );
 		},
 		handler: async ( argv ) => {
 			try {
-				await runCommand( argv.format as 'table' | 'json' );
+				await runCommand( argv.format as 'table' | 'json', argv.watch as boolean );
```
fredrikekelund:
```tsx
					choices: [ 'table', 'json' ] as const,
					default: 'table' as const,
					description: __( 'Output format' ),
				} )
				.option( 'watch', {
					type: 'boolean',
					default: false,
					description: __( 'Watch for site status changes and update the list in real-time' ),
				} );
		},
		handler: async ( argv ) => {
			try {
				await runCommand( argv.format, argv.watch );
```
Kind of a nitpick, but that's how we get yargs to provide the correct types for the options.

src/index.ts
```diff
@@ -391,8 +392,8 @@ async function appBoot() {
 
 	app.on( 'quit', () => {
 		void stopAllServersOnQuit();
```
fredrikekelund, line 397: This will kill all running processes, regardless of whether they were started from the CLI or from the app. We could just accept this behavior, but @nightnei also had a pretty good idea to display a simple modal before quitting the app that says "Do you want to stop the running sites? Yes / Leave them runing"

src/ipc-handlers.ts
```diff
@@ -725,6 +723,13 @@ export async function getThemeDetails(
 	if ( ! server.details.running || ! server.server ) {
 		return undefined;
 	}
+
+	// CLI-managed sites don't have PHP instance, return cached theme details for Now
+	// ToDo: Implement logic to fetch theme details using mu-plugin?
+	if ( ! server.server.php ) {
+		return server.details.themeDetails;
+	}
```
fredrikekelund, line 731: Once we have WP-CLI support, we might also consider using the `wp eval` command

src/modules/cli/lib/execute-command.ts
```diff
@@ -24,13 +26,36 @@ class CliCommandEventEmitter extends EventEmitter {
...
 
-export function executeCliCommand( args: string[] ): CliCommandEventEmitter {
+export interface ExecuteCliCommandOptions {
+	silent?: boolean;
+}
+
+export function executeCliCommand(
+	args: string[],
+	options: ExecuteCliCommandOptions = {}
+): CliCommandEventEmitter {
```
fredrikekelund:
```suggestion
): [CliCommandEventEmitter, ChildProcess] {
```
Optional, but we could return the child process instance from this function, too. It would save us from the dance of having to set the kill function on the EventEmitter

fredrikekelund (commented):
Great work, @bcotrim! 👍 Very satisfying to see this come together. It's working nicely in my testing so far. 

I left a couple of smaller comments.

I'd also be interested to hear your perspective of this `site list --watch` approach vs watching the config file for changes and using some mechanism (a version of the `studio site status` command, perhaps) to check the site status when `latestCliPid` changes. I don't really have a strong opinion there, but it would be good to have a documented reason for going one way or the other.

src/index.ts
bcotrim: I like that idea! I would add a "remember my decision" checkbox. I think that provides enough tools for our users to adjust their experience as it work best for them. We can always revisit in the future.
Should I handle that change in this PR or as part of STU-1078 ?

cli/logger.ts
```diff
@@ -24,69 +30,83 @@ export class LoggerError extends Error {
...
+		this._spinner = ora();
+	}
+
+	/**
+	 * Get the underlying ora spinner instance.
+	 * Useful for sharing with other modules that need to update progress.
+	 */
+	public get spinner(): Ora {
+		return this._spinner;
 	}
```
fredrikekelund: This seems like a leftover, right? We could probably change the `_spinner` prop back to `spinner`

cli/lib/pm2-manager.ts
```diff
@@ -212,3 +220,70 @@ export async function stopProcess( processName: string ): Promise< void > {
...
+ * @param handler - Callback invoked when a process message is received
+ * @returns Unsubscribe function to stop listening
+ */
+export async function subscribeProcessMessages(
+	handler: ( data: ProcessMessageData ) => void
+): Promise< () => void > {
+	const bus = await getPm2Bus();
+
+	const messageHandler = ( packet: unknown ) => {
+		const result = pm2ProcessMessageSchema.safeParse( packet );
```
```suggestion
		const result = childMessagePm2Schema.safeParse( packet );
```
Let's reuse `childMessagePm2Schema` here. We can update the schema definition to also include `data.process.name`

cli/lib/types/wordpress-server-ipc.ts
```diff
@@ -85,3 +93,26 @@ export const childMessagePm2Schema = z.object( {
...
+		name: z.string(),
+		pm_id: z.number(),
+	} ),
+	raw: z
+		.object( {
+			topic: z.string(),
+		} )
+		.passthrough(),
+} );
+export type Pm2ProcessMessage = z.infer< typeof pm2ProcessMessageSchema >;
```
fredrikekelund:
```suggestion
```
Let's remove this per my suggestion in `cli/lib/pm2-manager.ts`

cli/lib/types/wordpress-server-ipc.ts
```diff
@@ -85,3 +93,26 @@ export const childMessagePm2Schema = z.object( {
...
 } );
+
+// Zod schemas for PM2 process events (online, exit, stop, restart)
+export const pm2ProcessEventSchema = z.object( {
+	process: z.object( {
+		name: z.string(),
+	} ),
+	event: z.string(),
+} );
+export type Pm2ProcessEvent = z.infer< typeof pm2ProcessEventSchema >;
```
fredrikekelund:
```suggestion
```
We're only using the schema, so let's remove the inferred type.

src/index.ts
fredrikekelund: I created STU-1112 to tackle this separately

src/modules/cli/lib/execute-command.ts
```diff
@@ -10,7 +10,7 @@ type CliCommandEventMap = {
 	failure: void;
 };
 
-class CliCommandEventEmitter extends EventEmitter {
+export class CliCommandEventEmitter extends EventEmitter {
```
fredrikekelund:
```suggestion
class CliCommandEventEmitter extends EventEmitter {
```
No need to export this class, AFAICT

src/site-server.ts
```diff
@@ -58,6 +51,12 @@ type SiteServerMeta = {
 export class SiteServer {
 	server?: WordPressServerProcess;
 	wpCliExecutor?: WpCliProcess;
+	/**
+	 * Indicates whether a Studio-managed operation (start/stop) is in progress.
+	 * When true, file watchers should ignore site events to prevent interference
+	 * with the ongoing operation.
+	 */
+	hasOngoingOperation = false;
```
fredrikekelund, line 59: Nice 👍

cli/lib/wordpress-server-manager.ts
```diff
@@ -41,12 +55,15 @@ export async function isServerRunning( siteId: string ): Promise< ProcessDescrip
...
+	blueprint?: unknown;
+}
+
 export async function startWordPressServer(
 	site: SiteData,
-	options?: {
-		wpVersion?: string;
-		blueprint?: unknown;
-	}
+	logger: Logger< string >,
```
fredrikekelund, line 65: I don't think we need to address this in the current PR, but the pattern of passing in a logger instance seems to confuse responsibilities. I suggest following up later (maybe after the project launch) by refactoring this file to export a `WordPressServerManager` class that extends `EventEmitter`, and emitting the appropriate events (`progress`, `success`, `error`) as part of the startup flow.

fredrikekelund (approved):
It looks like the `e2e/blueprints.test.ts` tests are still failing, but I guess that's expected so long as we're in this intermittent state of using the Studio CLI to start sites, but still using Playground CLI to create sites. 

Other than that, this looks good! 👍 I left a couple of comments. No huge blockers. The biggest one is probably to remove `pm2ProcessMessageSchema` in favor of `childMessagePm2Schema`

nightnei (approved):
LGTM and works as expected 👍
I also tested the case when we do something in CLI (when Studio is closed) and then start the Studio, and all sites appears in Studio as started/stopped correctly 👍

bcotrim:
@fredrikekelund @nightnei thanks for the reviews!

> It looks like the e2e/blueprints.test.ts tests are still failing, but I guess that's expected so long as we're in this intermittent state of using the Studio CLI to start sites, but still using Playground CLI to create sites.

Yes, I think it will be easier to address once we have the full implementation

PR #2174, opened 2025-12-03
Remove `resolvePm2` and fix pm2 types

fredrikekelund:
## Related issues

- Fixes STU-1082

## Proposed Changes

This PR changes how we import pm2. pm2 exports two members: `default` (an instantiated `API`) and `custom` (the original `API` class). By importing `custom` and passing the `pm2_home` config prop, we can remove the `resolvePm2` logic. 

The only problem is that the `custom` member isn't accounted for in the pm2 type definitions. I've submitted a PR https://github.com/Unitech/pm2/pull/6064 to pm2 to fix this. Until that's merged, I've updated our pm2 patch to include my changes from that PR.

> [!TIP]
> View the upstream PR diff with whitespace changes hidden to see that I haven't actually changed all that much in `types/index.d.ts`

## Testing Instructions

1. Lint CI should pass
2. Run `npm run cli:build`
3. Run `node dist/cli/main.js site list`
4. Ensure that it finishes successfully

## Pre-merge Checklist


nightnei (approved):
Nice change and works as expected 👍

Only one NIT thing - I would avoid formatting changes in Studio's diff and in proposed PR to PM2 (or create a separate follow-up PR with formating), so that it's much easier to review and to see exact changes at a glance and maybe it will speed up review of the PR in PM2.

Screenshot 2025-12-03 at 13 50 33

cli/lib/pm2-manager.ts
```diff
@@ -101,7 +67,7 @@ const listProcesses = cacheFunctionTTL( () => {
 
 			const processDescriptions = ( processes || [] ).map( ( p ) => ( {
 				name: p.name || '',
-				pmId: p.pm_id || -1,
+				pmId: p.pm_id ?? -1,
```
fredrikekelund, line 70: This is unrelated to the core purpose of this PR, but it's an issue I ran into recently: if `p.pm_id` is 0, we would assign it a value of -1, which is incorrect.

katinthehatsite (approved):
This seems to work as expected:

Image

I have not looked in detail into this project so apologies if this question is obvious - should the status `Offline` be set to the sites that are not running and `Online` for those that are? Because I am seeing now `Offline` for everything.

fredrikekelund:
> should the status Offline be set to the sites that are not running and Online for those that are?

The online status isn't yet shared between the CLI and Studio, so `studio site list` will only show `Online` for sites that were started through the CLI. @bcotrim is working on this in STU-1019

PR #2176, opened 2025-12-03
Change texts in authentication onboarding

sejas:
## Related issues

- Fixes STU-1083

## Proposed Changes

- Change texts in onboarding Studio

## Testing Instructions

- Remove all your sites
- Remove the `onboardingCompleted` from your `~/Library/Application Support/Studio/appdata-v1.json`
- Run `npm start`
- Confirm the onboarding looks as expected

| Before | After |
|--------|--------|
|  Screenshot 2025-12-03 at 12 29 41 | Screenshot 2025-12-03 at 12 29 22 |

## Pre-merge Checklist

### DIFF
--- a/e2e/page-objects/onboarding.ts
+++ b/e2e/page-objects/onboarding.ts
@@ -10,9 +10,11 @@ export default class Onboarding {
 	}
 
 	get heading() {
-		return this.locator.getByRole( 'heading', {
-			name: /Connect to your WordPress.com account|Connect your WordPress.com account/,
-		} );
+		return this.locator.getByTestId( 'onboarding-welcome-title' ).or(
+			this.locator.getByRole( 'heading', {
+				name: /Connect to your WordPress\.com account|Connect your WordPress\.com account/,
+			} )
+		);
 	}
 
 	async completeOnboarding( options?: { customSiteName?: string; customFolderName?: string } ) {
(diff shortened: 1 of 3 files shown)


gcsecsey (approved):
The updated copies LGTM 👍 

Image

e2e/page-objects/onboarding.ts
```diff
@@ -10,9 +10,11 @@ export default class Onboarding {
 	}
 
 	get heading() {
-		return this.locator.getByRole( 'heading', {
-			name: /Connect to your WordPress.com account|Connect your WordPress.com account/,
-		} );
+		return this.locator.getByTestId( 'onboarding-welcome-title' ).or(
```
nightnei, line 13: Not something that should be done, I am just thinking for the future cases - are there some disadvantages of using `data-testid` vs roles 🤔 Interesting your thoughts why you decided to switch to testid?

nightnei (approved):
LGTM 👍

e2e/page-objects/onboarding.ts
sejas: I decided to use testId because this flow is used in every e2e and performance metric, so changing the text becomes more tricky as we need to support both text for a bit. Using only role could work, but I want to make sure the Heading displayed is this one and not the Add site flow.
I'll remove the text selector after a few days in favor of the testId.

PR #2178, opened 2025-12-03
Implement `studio wp` command

### LINKED ISSUE #47: Feature Request: add WP CLI support
### What

It would be nice if when opening the Terminal view, I had access to WP CLI out of the box, without the need for me to install anything else.

### Why

It would allow developers to quicky make changes to their local site from a different interface (CLI).

### How

_No response_

fredrikekelund:
## Related issues

- Fixes STU-959
- Fixes #47

## Proposed Changes

This PR implements the `studio wp` command, which exposes WP-CLI directly to users. 

If the site server is already started, the WP-CLI command is passed to the child process and executed on the existing Playground instance. If it's not started, we instantiate a temporary Playground and run the command there. It's not ideal to have this forked execution logic, but Playground startup is the greatest performance bottleneck, and this is a way to work around it. 

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js wp --path PATH_TO_STUDIO_SITE`
3. Ensure that the help contents are printed
4. Try a bunch of commands! There will be stuff that doesn't work. We might not be able to fix all of it, but let's do our best. 

## Pre-merge Checklist


bgrgicak:
> Find a way to prevent WP-CLI from passing help output through less or more. We might actually have to fork and rebundle WP-CLI to achieve this.

It might be worth opening an issue in the WP-CLI repo to discuss this, if Studio needs the feature, it could be useful to others.

cli/lib/sqlite-integration.ts
```diff
@@ -19,7 +21,7 @@ class CliSqliteProvider extends SqliteIntegrationProvider {
 		throw new Error( 'Unsupported platform' );
 	}
 
-	getSqliteFilename(): string {
+	getSqliteDirname(): string {
```
fredrikekelund, line 23: This change is unrelated to the core purpose of this PR. "Dirname" is the accurate term here, though, and I made this changed during my previous exploration.

patches/@wp-playground+cli+3.0.22.patch
fredrikekelund, line 1: I had to patch the `spawnWorkerThread` function in `packages/playground/cli/src/run-cli.ts` ([source](https://github.com/WordPress/wordpress-playground/blob/d567fd11bddb244ce1dd12f85d70777d08d67064/packages/playground/cli/src/run-cli.ts#L1183-L1202)), or at least I believe it's that function…

This is because the worker thread URL is constructed differently depending on whether `document` is defined, presumably to make Playground behave differently in Node than in the browser. The only issue is that we have to define `document` in the CLI context (see `cli/polyfills/browser-globals.js`), or else the `runCLI` function fails elsewhere because `document` is undefined. 

I'm happy to submit a bug for this unless @adamziel has any immediate ideas for how to fix this in upstream Playground.

fredrikekelund: I did some more thorough testing without `cli/polyfills/browser-globals.js` and couldn't reproduce the issues I thought I'd seen earlier. So, I removed the patch and `cli/polyfills/browser-globals.js` instead.

adamziel: @fredrikekelund are we good here, then? Or should we still revisit how `spawnWorkerThread` works?

fredrikekelund: We're good here. Thanks for double checking 👍

cli/lib/sqlite-integration.ts
```diff
@@ -3,6 +3,8 @@ import path from 'path';
 import { SqliteIntegrationProvider } from 'common/lib/sqlite-integration';
 
 const SQLITE_FILENAME = 'sqlite-database-integration';
+const SQLITE_COMMAND_DIRNAME = 'sqlite-command';
```
bcotrim: `warning  'SQLITE_COMMAND_DIRNAME' is assigned a value but never used.`

cli/commands/wp.ts
```diff
@@ -0,0 +1,156 @@
...
+	process.exit( await response.exitCode );
+}
+
+export async function commandHandler( argv: ArgumentsCamelCase< GlobalOptions > ) {
+	try {
+		const wpcliArgs = yargsParser( process.argv.slice( 3 ), {
+			config: {
+				'boolean-negation': false,
+				'camel-case-expansion': false,
+				'dot-notaton': false,
```
bcotrim: typo: should it be `dot-notation`?

bcotrim (approved):
Code looks good to me!
I found an issue while testing:

Bug: Post IDs (and other numeric args) fail because yargs-parser passes them as numbers, but the Zod schema expects strings.

Image

  UX suggestion: Even if Zod errors occur, we should parse them and display just the message field rather than the full JSON blob.

cli/commands/wp.ts
fredrikekelund: Good catch 👍 I also realized that the config object should be passed through the `configuration` key (which gives us proper type checking)

fredrikekelund:
Thanks for the review, @bcotrim! 

I chose to use `yargs-parser` to reliably be able to remove the `--path` option from the argv array before passing it to WP-CLI. Ultimately, I think I ended up undoing so much of what `yargs-parser` does that it's better to implement custom argv parsing to filter out the `--path` option. So, that's what I did.

nightnei:
@fredrikekelund I see this error, is it ok?
Screenshot 2025-12-10 at 12 44 08

fredrikekelund:
@nightnei, that's my mistake in writing the testing instructions. You should be running `node dist/cli/main.js wp`, not `studio wp`

nightnei:
@fredrikekelund oh, sorry, yeah, I already realized it after blindly copy-pasting 😅

nightnei (approved):
Works well and LGTM 👍

PR #2181, opened 2025-12-03
Publish site: Open OAuth page for logged-out users

epeicher:
## Related issues

- Closes STU-1062

## Proposed Changes

- When clicking the `Publish site` button, if the user is logged out, open the OAuth page to authenticate the user

|Studio | OAuth page opens|
|-------|------|
|CleanShot 2025-12-04 at 10 02 31@2x|CleanShot 2025-12-04 at 10 02 28@2x|

## Testing Instructions

- Log out if you are logged in
- Click on `Publish site` button on the header
- Check that the OAuth page opens and you can log in

## Pre-merge Checklist

### DIFF
--- a/src/components/publish-site-button.tsx
+++ b/src/components/publish-site-button.tsx
@@ -26,12 +26,11 @@ export const PublishSiteButton = () => {
 	const isAnySiteSyncing = isAnySitePulling || isAnySitePushing;
 
 	const handlePublishClick = useCallback( () => {
-		setSelectedTab( 'sync' );
 		if ( ! user ) {
 			authenticate();
-		} else {
-			dispatch( connectedSitesActions.openModal( 'push' ) );
 		}
+		dispatch( connectedSitesActions.openModal( 'push' ) );
+		setSelectedTab( 'sync' );
 	}, [ user, setSelectedTab, dispatch, authenticate ] );
 
 	if ( connectedSites.length !== 0 ) return null;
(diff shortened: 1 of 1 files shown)


katinthehatsite (approved):
This worked as expected for me 👍

sejas (approved):
I confirm the button now redirects to the sync tab and and opens the OAuth page when the user is not logged in 👍 , so the Publish site button now gives a hint what should be the next step.

src/components/publish-site-button.tsx
```diff
@@ -27,8 +27,12 @@ export const PublishSiteButton = () => {
 
 	const handlePublishClick = useCallback( () => {
 		setSelectedTab( 'sync' );
-		dispatch( connectedSitesActions.openModal( 'push' ) );
-	}, [ setSelectedTab, dispatch ] );
+		if ( ! user ) {
+			authenticate();
+		} else {
+			dispatch( connectedSitesActions.openModal( 'push' ) );
```
sejas: We can consider dispatching the action of open modal in any case. Outside of the else. That way the modal will open once the user is logged in.

Something like:

```
if ( ! user ) {
   authenticate();
}
dispatch( connectedSitesActions.openModal( 'push' ) );
```

sejas: Although we need to test it in different scenarios if the user clicks multiple times , in different sites etc.

epeicher: That works great! I am going to test different scenarios and push the changes if I don't find any issues

epeicher: I have tested clicking multiple times, and it opens multiple browsers similarly to clicking multiple times on `Log in to WordPress.com`, which I think is fine.

Also, when clicking multiple sites, _the last one wins_ which in my opinion is also correct.

I have pushed the change as part of 579661ee, please let me know if you find any issues

sejas: Thanks for the changes

PR #2183, opened 2025-12-03
Studio: Disable Add Site from the menu for no site empty screen

katinthehatsite:
## Related issues

Fixes STU-1068

## Proposed Changes

This PR disabled the `Add site` menu item in the top bar when the user is on the `Add site` modal. Currently, the menu item is enabled but it does nothing so it does not seem to be needed to have it enabled:

Screenshot 2025-12-04 at 9 48 26 AM

## Testing Instructions

* Click on the `Add site` button in the sidebar to open the `Add site` modal
* In the topbar menu, click on `File` > `Add site`
* Confirm that `Add site` is disabled

## Pre-merge Checklist

### DIFF
--- a/src/menu.ts
+++ b/src/menu.ts
@@ -29,7 +29,10 @@ import { promptWindowsSpeedUpSites } from 'src/lib/windows-helpers';
 import { getMainWindow } from 'src/main-window';
 import { isUpdateReadyToInstall, manualCheckForUpdates } from 'src/updates';
 
-export async function setupMenu( config: { needsOnboarding: boolean } ) {
+export async function setupMenu( config: {
+	needsOnboarding: boolean;
+	isAddSiteVisible?: boolean;
+} ) {
 	const mainWindow = await getMainWindow();
 	if ( ! mainWindow && process.platform !== 'darwin' ) {
 		Menu.setApplicationMenu( null );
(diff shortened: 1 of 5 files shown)


src/menu.ts
```diff
@@ -182,7 +188,7 @@ async function getAppMenu(
 					click: async () => {
 						void sendIpcEventToRenderer( 'add-site' );
 					},
-					enabled: ! needsOnboarding,
+					enabled: ! needsOnboarding && ! isAddSiteVisible,
```
katinthehatsite, line 191: For now, I disabled it as with the case with `needsOnboarding` but I am wondering if it makes sense to hide it completely.

epeicher: My initial thoughts were to hide it instead of disabling it, but after testing, I think it feels natural to leave the menu items the same and indicate through the disabling state that the `Add site` action cannot be performed on this screen. Therefore, I think this is a good approach. 👍

epeicher (approved):
Thanks @katinthehatsite! I have tested it, and it works as expected, I can see the menu disabled when the `Add a site` screen is open. LGTM! :shipit: 

|Add site screen | Others|
|-------|------|
| Image|  Image|

src/menu.ts
katinthehatsite: Nice, let's leave as is and we can adjust later if we receive feedback about this being confusing.

PR #2194, opened 2025-12-04
Update `studio auth status` response

nightnei:
## Proposed Changes

When I run `studio auth status` and receive `Successfully authenticated...` I had the impression that I ran the `login` command and was a bit confused. I [asked ChatGPT about his opinion](https://chatgpt.com/share/693199c7-26f4-8004-bb85-78f0de8a5a8b), w/o providing my confusion description, and it said exactly that improvement, so `Authenticated with WordPress.com as `username ' looks like reasonable.

## Testing Instructions
PR review should suffice


fredrikekelund (approved):
Fair 👍

PR #2204, opened 2025-12-05
CLI: Expand tilde in path option on Posix systems

fredrikekelund:
## Related issues

- Fixes STU-1101

## Proposed Changes

If I run `studio site create --path ~/Downloads/ddd2` (note the space after `--path`), the path is expanded as expected. However, if I run `studio site create --path=~/Downloads/ddd2`, it doesn't work. This is because `~` is a shell construct, and my shell doesn't expand `~` if it's not at the start of a string.

This PR fixes the problem by expanding `~` in the yargs `coerce` callback (so long as we're not running on Windows, where `~` isn't a thing).

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js site status --path=PATH_TO_SITE` (where `PATH_TO_SITE` is the path to a Studio site)
3. Ensure that it works as expected

## Pre-merge Checklist

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1,4 +1,5 @@
 import 'cli/polyfills/browser-globals.js';
+import os from 'node:os';
 import path from 'node:path';
 import { __ } from '@wordpress/i18n';
 import { suppressPunycodeWarning } from 'common/lib/suppress-punycode-warning';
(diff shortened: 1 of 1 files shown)


nightnei (approved):
LGTM and works perfectly 👍

PR #2212, opened 2025-12-08
Studio Sync: Implement resumable uploads for studio sync using Node process

gavande1:
## Related issues

- Fixes [STU-939](https://linear.app/a8c/issue/STU-939)

## Proposed Changes

- Implements resumable uploads for sync archives using Tus Protocol.
- Now it uploads the file first using new Tus protocal enabled endpoint, then it makes another request to initiate the import using attachment ID.

## Testing Instructions

> [!NOTE]
>  There are quite a number of steps to test this PR fully. I am happy to pair with you or help you out with the setup if you need.

- ~Checkout 196922-ghe-Automattic/wpcom your sandbox.~
- ~Sandbox public-api.wordpress.com and make it in FULL ACCESS mode.~
- Above PR has been deployed to production.
- Checkout this branch in Studio
- run `npm install & install start`
- Connect your wordpress.com account
- Add site for sync
- Run following tests

1. Do a full sync
2. Do a partial sync with different options
3. Start sync, go offline, wait few seconds, come online again. Make sure sync doesn't fail. This is to confirm resumable capability. 

- Checkout to master branch on Studio
- Run above tests again. Everything should work. The code handles backward compatibility. 

https://github.com/user-attachments/assets/dcac108e-f00b-4906-9c29-82d781078486

## Pre-merge Checklist

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -108,4 +108,4 @@ export const IPC_VOID_HANDLERS = < const >[
 ];
 
 // What's New
-export const FORCE_WHATS_NEW_WHEN_PATCH_CHANGED = true;
+export const FORCE_WHATS_NEW_WHEN_PATCH_CHANGED = false;
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -108,4 +108,10 @@ export const IPC_VOID_HANDLERS = < const >[
 ];
 
 // What's New
-export const FORCE_WHATS_NEW_WHEN_PATCH_CHANGED = false;
+export const FORCE_WHATS_NEW_WHEN_PATCH_CHANGED = true;
+
+// Studio File Upload constants
+export const STUDIO_FILE_UPLOADS_ENDPOINT_BASE =
+	'https://public-api.wordpress.com/rest/v1.1/studio-file-uploads';
+export const STUDIO_FILE_UPLOADS_CHUNK_SIZE = 500000;
+export const STUDIO_FILE_UPLOADS_RETRY_DELAYS = [ 0, 1000, 3000, 5000, 10000, 25000 ];
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -109,9 +109,3 @@ export const IPC_VOID_HANDLERS = < const >[
 
 // What's New
 export const FORCE_WHATS_NEW_WHEN_PATCH_CHANGED = true;
-
-// Studio File Upload constants
-export const STUDIO_FILE_UPLOADS_ENDPOINT_BASE =
-	'https://public-api.wordpress.com/rest/v1.1/studio-file-uploads';
-export const STUDIO_FILE_UPLOADS_CHUNK_SIZE = 500000;
-export const STUDIO_FILE_UPLOADS_RETRY_DELAYS = [ 0, 1000, 3000, 5000, 10000, 25000 ];
--- a/src/hooks/use-sync-states-progress-info.ts
+++ b/src/hooks/use-sync-states-progress-info.ts
@@ -166,7 +166,6 @@ export function useSyncStatesProgressInfo() {
 		const pushingStateKeys: PushStateProgressInfo[ 'key' ][] = [
 			'creatingBackup',
 			'uploading',
-			'uploadingPaused',
 			'uploadingResumed',
 			'creatingRemoteBackup',
 			'applyingChanges',
@@ -176,6 +175,10 @@ export function useSyncStatesProgressInfo() {
 			return false;
 		}
 		return pushingStateKeys.includes( key );
+	};
+
+	const isKeyUploadingPaused = ( key: PushStateProgressInfo[ 'key' ] | undefined ) => {
+		return key === 'uploadingPaused';
 	};
 
 	const isKeyImporting = ( key: PushStateProgressInfo[ 'key' ] | undefined ) => {
@@ -310,5 +313,6 @@ export function useSyncStatesProgressInfo() {
 		getBackupStatusWithProgress,
 		getPullStatusWithProgress,
 		getPushStatusWithProgress,
+		isKeyUploadingPaused,
 	};
 }
--- a/src/hooks/use-sync-states-progress-info.ts
+++ b/src/hooks/use-sync-states-progress-info.ts
@@ -16,7 +16,9 @@ export type PushStateProgressInfo = {
 		| 'finishing'
 		| 'finished'
 		| 'failed'
-		| 'cancelled';
+		| 'cancelled'
+		| 'uploadingPaused'
+		| 'uploadingResumed';
 	progress: number;
 	message: string;
 };
@@ -105,6 +107,16 @@ export function useSyncStatesProgressInfo() {
 				progress: 40,
 				message: __( 'Uploading Studio site…' ),
 			},
+			uploadingPaused: {
+				key: 'uploadingPaused',
+				progress: 45,
+				message: __( 'Uploading paused, retrying…' ),
+			},
+			uploadingResumed: {
+				key: 'uploadingResumed',
+				progress: 45,
+				message: __( 'Uploading resumed…' ),
+			},
 			creatingRemoteBackup: {
 				key: 'creatingRemoteBackup',
 				progress: 50,
@@ -154,6 +166,8 @@ export function useSyncStatesProgressInfo() {
 		const pushingStateKeys: PushStateProgressInfo[ 'key' ][] = [
 			'creatingBackup',
 			'uploading',
+			'uploadingPaused',
+			'uploadingResumed',
 			'creatingRemoteBackup',
 			'applyingChanges',
 			'finishing',
--- a/src/hooks/use-sync-states-progress-info.ts
+++ b/src/hooks/use-sync-states-progress-info.ts
@@ -17,8 +17,7 @@ export type PushStateProgressInfo = {
 		| 'finished'
 		| 'failed'
 		| 'cancelled'
-		| 'uploadingPaused'
-		| 'uploadingResumed';
+		| 'uploadingPaused';
 	progress: number;
 	message: string;
 };
@@ -110,12 +109,7 @@ export function useSyncStatesProgressInfo() {
 			uploadingPaused: {
 				key: 'uploadingPaused',
 				progress: 45,
-				message: __( 'Uploading paused, retrying…' ),
-			},
-			uploadingResumed: {
-				key: 'uploadingResumed',
-				progress: 45,
-				message: __( 'Uploading resumed…' ),
+				message: __( 'Uploading paused' ),
 			},
 			creatingRemoteBackup: {
 				key: 'creatingRemoteBackup',
@@ -166,7 +160,6 @@ export function useSyncStatesProgressInfo() {
 		const pushingStateKeys: PushStateProgressInfo[ 'key' ][] = [
 			'creatingBackup',
 			'uploading',
-			'uploadingResumed',
 			'creatingRemoteBackup',
 			'applyingChanges',
 			'finishing',
--- a/src/ipc-utils.ts
+++ b/src/ipc-utils.ts
@@ -38,8 +38,8 @@ export interface IpcEvents {
 		},
 	];
 	'site-context-menu-action': [ { action: string; siteId: string } ];
-	'studio-file-upload-paused': [ { error: string; selectedSiteId: string; remoteSiteId: number } ];
-	'studio-file-upload-resumed': [ { selectedSiteId: string; remoteSiteId: number } ];
+	'sync-upload-paused': [ { error: string; selectedSiteId: string; remoteSiteId: number } ];
+	'sync-upload-resumed': [ { selectedSiteId: string; remoteSiteId: number } ];
 	'snapshot-error': [ { operationId: crypto.UUID; data: SnapshotEventData } ];
 	'snapshot-fatal-error': [ { operationId: crypto.UUID; data: { message: string } } ];
 	'snapshot-output': [ { operationId: crypto.UUID; data: SnapshotEventData } ];
--- a/src/ipc-utils.ts
+++ b/src/ipc-utils.ts
@@ -38,6 +38,8 @@ export interface IpcEvents {
 		},
 	];
 	'site-context-menu-action': [ { action: string; siteId: string } ];
+	'studio-file-upload-paused': [ { error: string; selectedSiteId: string; remoteSiteId: number } ];
+	'studio-file-upload-resumed': [ { selectedSiteId: string; remoteSiteId: number } ];
 	'snapshot-error': [ { operationId: crypto.UUID; data: SnapshotEventData } ];
 	'snapshot-fatal-error': [ { operationId: crypto.UUID; data: { message: string } } ];
 	'snapshot-output': [ { operationId: crypto.UUID; data: SnapshotEventData } ];
--- a/src/modules/sync/lib/ipc-handlers.ts
+++ b/src/modules/sync/lib/ipc-handlers.ts
@@ -214,7 +214,7 @@ export async function pushArchive(
 					return;
 				}
 
-				const attachmentId = payload.lastResponse.getHeader( 'x-videopress-upload-media-id' );
+				const attachmentId = payload.lastResponse.getHeader( 'x-studio-file-upload-media-id' );
 				if ( attachmentId ) {
 					resolve( attachmentId );
 				} else {
--- a/src/modules/sync/lib/ipc-handlers.ts
+++ b/src/modules/sync/lib/ipc-handlers.ts
@@ -196,7 +196,7 @@ export async function pushArchive(
 			onProgress: ( bytesSent, bytesTotal ) => {
 				if ( isUploadingPaused ) {
 					isUploadingPaused = false;
-					void sendIpcEventToRenderer( 'studio-file-upload-resumed', {
+					void sendIpcEventToRenderer( 'sync-upload-resumed', {
 						selectedSiteId: selectedSiteId,
 						remoteSiteId: remoteSiteId,
 					} );
@@ -225,7 +225,7 @@ export async function pushArchive(
 			},
 			onShouldRetry: ( error ) => {
 				isUploadingPaused = true;
-				void sendIpcEventToRenderer( 'studio-file-upload-paused', {
+				void sendIpcEventToRenderer( 'sync-upload-paused', {
 					selectedSiteId: selectedSiteId,
 					remoteSiteId: remoteSiteId,
 					error: error.message,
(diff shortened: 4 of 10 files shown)


src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +166,94 @@ export async function pushArchive(
...
+					remoteSiteId: remoteSiteId,
+					error: error.message,
+				} );
+				console.error( 'Upload request error', error.message );
+				return true;
+			},
+		} );
+
+		upload.start();
+	} );
```
fredrikekelund, line 250:
```suggestion
	} ).finally( () => {
		file.destroy();
		file.close();
	} );

	const attachmentId = await attachmentPromise;
```
We recently encountered an issue with very similar logic in the CLI: the cleanup logic wouldn't run under specific circumstances. Your logic looks solid, but I still recommend this approach, as it's foolproof. Combine it with the following change on L174:

```ts
	const attachmentPromise = new Promise< string >( ( resolve, reject ) => {
```

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +166,94 @@ export async function pushArchive(
...
+	const cleanup = () => {
+		file.destroy();
+		file.close();
+	};
+
+	const attachmentId = await new Promise< string >( ( resolve, reject ) => {
+		const upload = new Upload( file, {
+			endpoint: `${ STUDIO_FILE_UPLOADS_ENDPOINT_BASE }/${ remoteSiteId }`,
+			chunkSize: STUDIO_FILE_UPLOADS_CHUNK_SIZE,
+			retryDelays: STUDIO_FILE_UPLOADS_RETRY_DELAYS,
```
fredrikekelund:
```suggestion
			endpoint: `https://public-api.wordpress.com/rest/v1.1/studio-file-uploads/${ remoteSiteId }`,
			chunkSize: 500000,
			retryDelays: [ 0, 1000, 3000, 5000, 10000, 25000 ],
```
Since these constants aren't being reused, I would argue it simplifies the code to inline them.

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +166,94 @@ export async function pushArchive(
...
+					} );
+					console.log( '[TUS] Upload resumed' );
+				}
+
+				if ( bytesTotal > 0 ) {
+					const progress = ( bytesSent / bytesTotal ) * 100;
+					console.log( '[TUS] Upload progress: %s%%', progress.toFixed( 2 ) );
+				}
+			},
+			onSuccess: ( payload: { lastResponse: HttpResponse } ) => {
```
fredrikekelund:
```suggestion
			onSuccess: ( payload ) => {
```
`tus-js-client` already provides the type.

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +166,94 @@ export async function pushArchive(
...
+
+				const attachmentId = payload.lastResponse.getHeader( 'x-videopress-upload-media-id' );
+				if ( attachmentId ) {
+					resolve( attachmentId );
+				} else {
+					console.error( 'Upload completed but required header not found' );
+					reject( new Error( 'Upload completed but required header not found' ) );
+				}
+			},
+			onShouldRetry: ( error: Error ) => {
```
fredrikekelund:
```suggestion
			onShouldRetry: ( error ) => {
```
`tus-js-client` already provides the type.

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +166,94 @@ export async function pushArchive(
...
+			storeFingerprintForResuming: true,
+			headers: {
+				Authorization: `Bearer ${ token.accessToken }`,
+			},
+			metadata: {
+				filename,
+				filetype: 'application/gzip',
+			},
+			uploadSize: fileSize,
+			onBeforeRequest: ( req: HttpRequest ) => {
```
fredrikekelund:
```suggestion
			onBeforeRequest: ( req ) => {
```
`tus-js-client` already provides the type.

src/ipc-utils.ts
```diff
@@ -38,6 +38,8 @@ export interface IpcEvents {
 		},
 	];
 	'site-context-menu-action': [ { action: string; siteId: string } ];
+	'studio-file-upload-paused': [ { error: string; selectedSiteId: string; remoteSiteId: number } ];
+	'studio-file-upload-resumed': [ { selectedSiteId: string; remoteSiteId: number } ];
```
fredrikekelund:
```suggestion
	'sync-upload-paused': [ { error: string; selectedSiteId: string; remoteSiteId: number } ];
	'sync-upload-resumed': [ { selectedSiteId: string; remoteSiteId: number } ];
```
Nit, but prefixing with `studio-` seems superfluous. Specifying that it's related to sync operations might be useful, though.

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -105,6 +107,16 @@ export function useSyncStatesProgressInfo() {
 				progress: 40,
 				message: __( 'Uploading Studio site…' ),
 			},
+			uploadingPaused: {
+				key: 'uploadingPaused',
+				progress: 45,
+				message: __( 'Uploading paused, retrying…' ),
```
fredrikekelund:
```suggestion
				message: __( 'Uploading paused' ),
```
Given that we have the tooltip, I think `Uploading paused` is enough.

src/hooks/use-sync-states-progress-info.ts
```diff
@@ -105,6 +107,16 @@ export function useSyncStatesProgressInfo() {
...
+			},
+			uploadingResumed: {
+				key: 'uploadingResumed',
+				progress: 45,
+				message: __( 'Uploading resumed…' ),
+			},
```
fredrikekelund: Do we actually need a separate status for this? Can't we just set the status back to `uploading`?

fredrikekelund (commented):
This looks very promising! I left a couple of comments, but overall, the architecture is clear and makes sense 👍

I wasn't able to test the changes, though, because the upload keeps failing, and I get errors like this in my Studio terminal session:

```
Upload request error tus: unexpected response while creating upload, originated from request (method: POST, url: https://public-api.wordpress.com/rest/v1.1/studio-file-uploads/241885143, response code: 500, response text: , request id: n/a)
```

src/modules/sync/lib/ipc-handlers.ts
fredrikekelund, line 1: The `wpcom.req.post` call in `pushArchive` should go to `/sites/${ remoteSiteId }/studio-app/sync/import/initiate`, I believe

gavande1: Thanks for pointing that. It have restored during merge conflict. I fixed in [f5d482c](https://github.com/Automattic/studio/pull/2212/commits/f5d482c1be6e0888b3a119f6cfcd48273b110edc).

src/hooks/use-sync-states-progress-info.ts
gavande1: Good call. It's not needed. Removed in [4b3c2bf](https://github.com/Automattic/studio/pull/2212/commits/4b3c2bf97fe1cbfa37fd2aee26d83ea1c5748323).

src/hooks/use-sync-states-progress-info.ts
gavande1: Makes sense. Fixed in [4b3c2bf](https://github.com/Automattic/studio/pull/2212/commits/4b3c2bf97fe1cbfa37fd2aee26d83ea1c5748323).

src/modules/sync/lib/ipc-handlers.ts
fredrikekelund: Also, combine it with removing the `cleanup` function

src/ipc-utils.ts
gavande1: It makes sense to keep it related to sync operations. Added in [b2a7760](https://github.com/Automattic/studio/pull/2212/commits/b2a7760069214b4f06be8b548b8aa72b3a0ca05e).

src/modules/sync/lib/ipc-handlers.ts
gavande1: Updated in [c1efbe1](https://github.com/Automattic/studio/pull/2212/commits/c1efbe13c0b111cb0f49e74ed76bc9f3eaffccf2).

src/modules/sync/lib/ipc-handlers.ts
gavande1: Updated in [c1efbe1](https://github.com/Automattic/studio/pull/2212/commits/c1efbe13c0b111cb0f49e74ed76bc9f3eaffccf2).

src/modules/sync/lib/ipc-handlers.ts
gavande1: Updated in [c1efbe1](https://github.com/Automattic/studio/pull/2212/commits/c1efbe13c0b111cb0f49e74ed76bc9f3eaffccf2).

src/modules/sync/lib/ipc-handlers.ts
gavande1: Makes sense to me. Updated in [9a3e9da](https://github.com/Automattic/studio/pull/2212/commits/9a3e9daf457c1a8e1a40236ce8a100a1e6eb5c7b). If needed in future, we can move them to constants.

src/modules/sync/lib/ipc-handlers.ts
gavande1: I have applied your suggestion in [7aaf10d](https://github.com/Automattic/studio/pull/2212/commits/7aaf10d0180f36708d273c4690597b925e1a75da). I am curious, under what condition it doesn't run? Did notice anything specific?

fredrikekelund: Your existing logic looked solid. I pointed it out because it was remarkably similar to a case that recently gave us a headache in the CLI project. Using a `finally` clause guarantees that cleanup logic runs at the right time, regardless of how the promise settles. Consider it future proofing 🙂

gavande1: Thanks for explaining it. It makes sense to use `finally` becuase no matter how promise settles, it will be always called.

src/constants.ts
```diff
@@ -108,4 +108,4 @@ export const IPC_VOID_HANDLERS = < const >[
 ];
 
 // What's New
-export const FORCE_WHATS_NEW_WHEN_PATCH_CHANGED = false;
+export const FORCE_WHATS_NEW_WHEN_PATCH_CHANGED = true;
```
fredrikekelund: Was this change intentional?

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +161,99 @@ export async function pushArchive(
...
+				}
+
+				if ( ! hasUploadStarted ) {
+					hasUploadStarted = true;
+				}
+
+				if ( bytesTotal > 0 ) {
+					const progress = ( bytesSent / bytesTotal ) * 100;
+					console.log( '[TUS] Upload progress: %s%%', progress.toFixed( 2 ) );
+				}
```
fredrikekelund:
```suggestion
```
This part makes a lot of noise. I say let's just remove it. Focusing on success and error conditions seems sufficient.

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +161,99 @@ export async function pushArchive(
...
+				if ( hasUploadStarted ) {
+					isUploadingPaused = true;
+					void sendIpcEventToRenderer( 'sync-upload-paused', {
+						selectedSiteId: selectedSiteId,
+						remoteSiteId: remoteSiteId,
+						error: error.message,
+					} );
+				}
+
+				console.error( 'Upload request error', error.message );
```
fredrikekelund: Should we log `Upload request error` here, or something that makes it clearer that the upload is now paused?

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +161,99 @@ export async function pushArchive(
...
+
+				if ( bytesTotal > 0 ) {
+					const progress = ( bytesSent / bytesTotal ) * 100;
+					console.log( '[TUS] Upload progress: %s%%', progress.toFixed( 2 ) );
+				}
+			},
+			onSuccess: ( payload ) => {
+				if ( ! payload.lastResponse ) {
+					console.error( 'Upload completed but no response received' );
+					reject( new Error( 'Upload completed but no response received' ) );
```
fredrikekelund: Do we really need to log and reject/resolve, like we do in multiple places here? The renderer process already sends errors to Sentry. Does it help to also log to the terminal? I think this is an open question.

fredrikekelund (approved):
Looks good, and works great 👍 I left a couple of final comments that would be good to address, but nothing blocking.

src/constants.ts
gavande1: No. I think it was recently changed. I have corrected it now.

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +161,99 @@ export async function pushArchive(
...
+				filetype: 'application/gzip',
+			},
+			uploadSize: fileSize,
+			onBeforeRequest: ( req ) => {
+				if ( req.getMethod() === 'HEAD' ) {
+					// @ts-expect-error We need to override the method to get the response headers.
+					req._method = 'GET';
+					req.setHeader( 'X-HTTP-Method-Override', 'HEAD' );
+				}
+			},
```
fredrikekelund, line 192: Why do we need this, btw? I'm still able to complete an upload without this logic.

src/modules/sync/lib/ipc-handlers.ts
gavande1: Removed in [1a966ec](https://github.com/Automattic/studio/pull/2212/commits/1a966ece6947a17a0de8978199fc3006fa7df57e).

src/modules/sync/lib/ipc-handlers.ts
gavande1: No. We don't really need. Removed in [1a966ec](https://github.com/Automattic/studio/pull/2212/commits/1a966ece6947a17a0de8978199fc3006fa7df57e).

src/modules/sync/lib/ipc-handlers.ts
gavande1: Good call. Its better to log upload is paused with error. I have updated it in [c29df5d](https://github.com/Automattic/studio/pull/2212/commits/c29df5d21f58f9b6c3af92546114d24f98b29f29). Thanks!

src/modules/sync/lib/ipc-handlers.ts
gavande1: Yes, we definitely need this. The `HEAD` method allows us to retrieve the upload status when resuming an upload. Without it, the file would restart uploading from the beginning.

epeicher (approved):
Thanks @gavande1 for implementing this functionality! I have tested it, and it works as expected, I can see the file upload resuming after disconnecting and re-connecting the wifi. This is a great functionality for the user 🙌 

I have tested also the `trunk` version and it works fine, while I can reproduce the previous error in `trunk` if there is a disconnection.

|Trunk with disconnection | Trunk successful upload|
|-------|------|
| Image| Image|


### Video with an upload and a disconnection while uploading

https://github.com/user-attachments/assets/2c8622b0-fed4-4560-913f-6ee5ad33af29


One question for a possible follow-up: do you think that once these changes are merged, we could work on resuming uploads even if the user closes Studio and reopens it? That would be another step toward improving the Sync functionality.

gcsecsey (approved):
Thanks @gavande1 for working on this feature, it'll be a great improvement for the upload experience!

I tested this together with the backend changes, and can confirm that the upload continues as described after disconnecting the wifi:

https://github.com/user-attachments/assets/e3aea4e7-118a-49d3-990a-644bae8d694a

One thing I noticed while testing this is that if the disconnect happens after the upload step, the renderer crashes. In a follow-up, we can ensure that we handle the offline states for the whole sync flow.

https://github.com/user-attachments/assets/a07e38ed-6dde-4549-a7cb-6b641a4cb5b4

```
hook.js:608 A non-serializable value was detected in the state, in the path: `wpcomSitesApi.queries.getWpComSites({"connectedSiteIds":[249663985],"userId":37463824}).error.data`. Value: Error: Request has been terminated
Possible causes: the network is offline, Origin is not allowed by Access-Control-Allow-Origin, the page is being unloaded, etc.
```

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -158,16 +161,91 @@ export async function pushArchive(
...
+			onError: ( error ) => {
+				console.error( '[TUS] Upload error', error );
+				reject( error );
+			},
+			onProgress: () => {
```
epeicher, line 197: Suggestion for a follow-up task, we can use the [parameters`bytesUploaded` and `bytestTotal`](https://github.com/tus/tus-js-client/blob/main/docs/usage.md#example-simple-file-upload) to add some progress to the UI

src/modules/sync/lib/ipc-handlers.ts
```diff
@@ -207,15 +207,9 @@ export async function pushArchive(
 				if ( ! hasUploadStarted ) {
 					hasUploadStarted = true;
 				}
-
-				if ( bytesTotal > 0 ) {
-					const progress = ( bytesSent / bytesTotal ) * 100;
-					console.log( '[TUS] Upload progress: %s%%', progress.toFixed( 2 ) );
```
gavande1: @epeicher, regarding your [comment](https://github.com/Automattic/studio/pull/2212/changes/1a966ece6947a17a0de8978199fc3006fa7df57e#r2610438087), I will create a follow‑up task to show upload progress in the UI. I originally had that in mind, but later removed it to keep this PR as simple as possible. We will definitely iterate on it.

src/modules/sync/lib/ipc-handlers.ts
gavande1: @epeicher Answered to the comment [here](https://github.com/Automattic/studio/pull/2212/changes/1a966ece6947a17a0de8978199fc3006fa7df57e#r2613603170)

gavande1:
@gcsecsey and @epeicher Thanks for the review and suggesting follow up tasks. 

> do you think that once these changes are merged, we could work on resuming uploads even if the user closes Studio and reopens it? 

Yes. that's one of the suggestion I received from Wojtek as well. It would greatly improve experience for the user. I will create follow up tasks for that once this is merged.

src/modules/sync/lib/ipc-handlers.ts
epeicher: Thanks @gavande1, I agree that it's better to do it iteratively, great work on this!

PR #2221, opened 2025-12-09
Fix broken CLI on Windows

fredrikekelund:
## Related issues

- Related to p1765274799851279/1765198096.499159-slack-C06DRMD6VPZ

## Proposed Changes

After upgrading yargs to v18, there's some logic in that module that breaks the Studio CLI on Windows. This happens because `cli/polyfills/browser-globals.js` sets a global `document` object, which leads yargs to think we're in a browser context. 

This PR removes that import (and a bunch of unused imports), which fixes the problem. I didn't notice any side effects, but we're also not sure exactly why @bcotrim had to add it in the first place.

Moreover, Studio didn't display any feedback to users when the CLI failed to start on Windows. I added a `failure` event listener to address this.

## Testing Instructions

1. Run `npm run cli:build` on Windows
2. Run `node dist\cli\main.js`
3. Ensure that a help page is printed

## Pre-merge Checklist

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -10,15 +10,7 @@ import { registerCommand as registerCreateCommand } from 'cli/commands/preview/c
 import { registerCommand as registerDeleteCommand } from 'cli/commands/preview/delete';
 import { registerCommand as registerListCommand } from 'cli/commands/preview/list';
 import { registerCommand as registerUpdateCommand } from 'cli/commands/preview/update';
-import { registerCommand as registerSiteCreateCommand } from 'cli/commands/site/create';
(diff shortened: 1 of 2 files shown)


gcsecsey (approved):
Thanks @fredrikekelund for the quick fix and for the detailed explanation! 🙌 This works well for me, and it also resolves the issue with the preview site flow.

| trunk | this branch |
|--------|--------|
| Image | Image |

bcotrim (approved):
LGTM 👍 

Test preview sites with standalone CLI and Studio for both dev and packaged Studio.
Could we also remove the file?

fredrikekelund:
> Could we also remove the file?

Probably, but I'll do that in https://github.com/Automattic/studio/pull/2178

PR #2224, opened 2025-12-09
E2E: fix on trunk

nightnei:
## Proposed Changes
In the history of commits you can see that I tried to set that old commit hash instead of `trunk` and we can see that error. Then I changed it to the latest commit hash, and we can see that now it's green.
I plan to merge this PR [E2E: remove backward compatibility for onboarding header](https://github.com/Automattic/studio/pull/2225) and add one more commit to thsi PR to point "Performance metrics" test to that one.

## Testing Instructions
Defined in "Proposed changes"

### DIFF
--- a/.github/workflows/metrics-comparison.yml
+++ b/.github/workflows/metrics-comparison.yml
@@ -41,7 +41,7 @@ jobs:
 
       - name: Compare performance with trunk
         if: github.event_name == 'pull_request'
-        run: cd scripts/compare-perf && npm ci && npm run compare -- perf $GITHUB_SHA 6917e54eea4bb37d8beaca746cd5861a38331264 --tests-branch $GITHUB_SHA --rounds 3
+        run: cd scripts/compare-perf && npm ci && npm run compare -- perf $GITHUB_SHA trunk --tests-branch $GITHUB_SHA --rounds 3
 
       - name: Post performance results to PR
         if: github.event_name == 'pull_request' && !cancelled()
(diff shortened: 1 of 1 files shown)


.github/workflows/metrics-comparison.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
       - name: Compare performance with trunk
         if: github.event_name == 'pull_request'
-        run: cd scripts/compare-perf && npm ci && npm run compare -- perf $GITHUB_SHA trunk --tests-branch $GITHUB_SHA --rounds 3
+        run: cd scripts/compare-perf && npm ci && npm run compare -- perf $GITHUB_SHA 6917e54eea4bb37d8beaca746cd5861a38331264 --tests-branch $GITHUB_SHA --rounds 3
```
sejas: I think we need to set this hash back to `trunk`:

```suggestion
        run: cd scripts/compare-perf && npm ci && npm run compare -- perf $GITHUB_SHA trunk --tests-branch $GITHUB_SHA --rounds 3
```

sejas (approved):
@nightnei , thanks for updating the hash were we compare the performance metrics.
I think we need to use `trunk` in the line 44. Aside from that, the change look good. Thank you!

.github/workflows/metrics-comparison.yml
nightnei: Sure
I did it intentionally, as I menthioned in the PR description - I replaced `trunk` temporary just for testing purposes, to make sure that CI is green and was going to provide the hash from this commit https://github.com/Automattic/studio/pull/2225 (Already pushed this change).

PR #2225, opened 2025-12-09
E2E: remove backward compatibility for onboarding header

nightnei:
## Proposed Changes
Previously, we added this `.or()` when we introduced `id` for the onboarding header. Now it looks like perfect time to remove it and at teh same time point the new commit hash for comparison Performance metrics in `trunk` tests (after mergign PRs to trunk)- https://github.com/Automattic/studio/pull/2224

# Testing instructions
Tests should be green

### DIFF
--- a/e2e/page-objects/onboarding.ts
+++ b/e2e/page-objects/onboarding.ts
@@ -11,11 +11,7 @@ export default class Onboarding {
 	}
 
 	get heading() {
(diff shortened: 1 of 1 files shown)


sejas (approved):
Thanks for removing the alternative header locator. I confirm the e2e tests pass correctly.

PR #2226, opened 2025-12-09
CLI: make error messages consistent

nightnei:
## Related issues

- Fixes STU-1090

## Proposed Changes

See comments in the code below.

This PR:
1. removes printing `--help` when an error occurs.
2. Introduces a function to render different errors in a consistent way
3. Proposes to move validators from 3 places to `yargs.option->coerce`

Also, I created a proposition PR with excluding validators to separate folders inside every command https://github.com/Automattic/studio/pull/2227

## Testing Instructions
1. Run:
  1.1 `node dist/cli/main.js site create --wp=12312312`
  1.2 `node dist/cli/main.js site create --php=12312312`
  1.3 Assert that you see the same error formatting
  1.4 [Assert that you don't see printed --help](https://github.com/Automattic/studio/pull/2226#discussion_r2603393620)
2. Try to create site with correct wp / php / blueprint and assert that there is no regression


cli/index.ts
```diff
@@ -76,7 +76,8 @@ async function main() {
 			previewYargs.demandCommand( 1, __( 'You must provide a valid command' ) );
 		} )
 		.demandCommand( 1, __( 'You must provide a valid command' ) )
-		.strict();
+		.strict()
+		.showHelpOnFail( false );
```
nightnei: This will turn off `--help` when error happens, for consistency with manually handled errors, for example - not valid `--wp` version.

<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Image
<td>Image<br />Image
</tr>
</table>

cli/lib/generate-yargs-error-message.ts
```diff
@@ -0,0 +1,3 @@
+export function generateYargsErrorMessage( argument: string, given: string, mustBe: string ) {
```
nightnei: When we handle errors manually, I propose to try to always prioritize using this template, for consistency with default yargs validation.

<table>
<tr>
<td>DEFAULT VALIDATION
<td>PROPOSED MANUAL VALIDATION
</tr>
<tr>
<td>Image
<td>Image
</tr>
</table>

cli/commands/site/create.ts
```diff
@@ -289,6 +277,31 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 					type: 'string',
 					describe: __( 'WordPress version (e.g., "latest", "6.4", "6.4.1")' ),
 					default: DEFAULT_VERSIONS.wp,
+					coerce: ( value: string ) => {
```
nightnei: Currently, our validators are spread across `registerCommand->choices`, `runCommand` and `handler`. So instead of using 3 places, I propose to use coerce option as kinda handler for command arguments. This way we will have straightforward structure as:

1. `coerce` is used for validation, type coercion and in general having dedicated place for prepering arguments before starting actual `runCommand`.
2. `runCommand` only for doing command job.

Of course, some validators may be inside `runCommand`, but I propose at least splitting responsibilities as much as possible.

cli/commands/site/create.ts
```diff
@@ -289,6 +277,31 @@ export const registerCommand = ( yargs: StudioArgv ) => {
...
+						if ( ! isValidWordPressVersion( value ) ) {
+							throw new LoggerError(
+								generateYargsErrorMessage(
+									'wp',
+									value,
+									__(
+										'"latest", "nightly", or a valid version number (e.g., "6.4", "6.4.1", "6.4-beta1")'
+									)
+								)
+							);
```
```suggestion
							throw new ValidationError(
								'wp',
								value,
								__(
									'Must be: "latest", "nightly", or a valid version number (e.g., "6.4", "6.4.1", "6.4-beta1")'
								)
							);
```
A couple of thoughts:
1. Let's use a new Error class in place of `generateYargsErrorMessage`.
2. Let's stick with complete, translatable sentences in `__()` calls

fredrikekelund (commented):
I'm onboard with moving the validation logic to the `coerce` methods 👍 I'm not really a fan of the `generateYargsErrorMessage` function, though. I suggested using a `ValidationError` class instead. The implementation might look something like this:

```ts
class ValidationError extends Error {
	optionName: string;
	givenValue: any;
	validationMessage: string;

	constructor( optionName: string, givenValue: any, validationMessage: string ) {
		super();
		this.name = 'ValidationError';
		this.optionName = optionName;
		this.givenValue = givenValue;
		this.validationMessage = validationMessage;
	}

	get message(): string {
		return `Invalid values: \n - Argument: ${ this.optionName }, Given: "${ this.givenValue }", ${ this.validationMessage }.`;
	}
}
```

cli/commands/site/create.ts
nightnei: Great idea with class and agree about complete messages 👍

cli/commands/site/create.ts
```diff
@@ -275,6 +263,42 @@ function readBlueprint( blueprintPath: string ) {
...
+	if ( value ) {
+		if ( /^https?:\/\//.test( value ) ) {
+			blueprintJson = await fetchBlueprint( value );
+		} else {
+			blueprintJson = readBlueprint( value );
+		}
+	}
+
+	return blueprintJson;
+}
```
```suggestion
export async function coerceBlueprint( value: string ) {
	if ( value ) {
		if ( /^https?:\/\//.test( value ) ) {
			return await fetchBlueprint( value );
		} else {
			return readBlueprint( value );
		}
	}
}
```
Nit. Now that this lives in a standalone function, we don't really need to assign to the `blueprintJson` variable.

cli/commands/site/tests/create.test.ts
```diff
@@ -689,4 +659,102 @@ describe( 'CLI: studio site create', () => {
...
+		} );
+
+		it( 'should error if WordPress version is below minimum', async () => {
+			const { coerceWpVersion } = await import( '../create' );
+
+			expect( () => coerceWpVersion( '6.0' ) ).toThrow( 'Invalid value' );
+		} );
+	} );
+
+	describe( 'coerceBlueprint', () => {
```
fredrikekelund: On the one hand, it makes sense to test anything labeled "validation logic". On the other hand, this seems like a typical example of adding so many mocks that we end up testing the implementation more than the output.

I'm really not convinced that these tests give us that much assurance…

cli/index.ts
```diff
@@ -98,7 +98,8 @@ async function main() {
 			handler: wpCliCommandHandler,
 		} )
 		.demandCommand( 1, __( 'You must provide a valid command' ) )
-		.strict();
+		.strict()
+		.showHelpOnFail( false );
```
fredrikekelund, line 102: 👍

fredrikekelund (approved):
The more validation we can move outside the `runCommand` functions, the better. Still, I don't think it's realistic that the `runCommand` functions wouldn't ever throw due to faulty parameters. So, this is a good direction, but I don't think we should contort ourselves completely in the name of accomplishing this. It's a balance. 

Anyway, this particular PR LGTM 👍 Except the fact that I would reduce or remove the `coerceBlueprint` tests.

cli/commands/site/create.ts
nightnei: 👍
Also, we don't need `if ( value ) {`, since the coerce function runs only if the argument is provided. One more small benefit of using this approach.

cli/commands/site/tests/create.test.ts
nightnei: Makes sense. Removed.

PR #2227, opened 2025-12-09
CLI: proposition with a files structure for validators

nightnei:
## Proposed Changes
Follow up to https://github.com/Automattic/studio/pull/2226
Just a proposition to make commands files smaller and also to make an agreement for code structure around validators.

## Testing Instructions
Same as [here](https://github.com/Automattic/studio/pull/2226)


cli/commands/preview/tests/create.test.ts
```diff
@@ -1,228 +1,270 @@
+import fs from 'fs';
 import os from 'os';
```
nightnei, line 2: I just merged tests with lib/tests/validation.test.ts

https://github.com/Automattic/studio/pull/2227/files#diff-90198c7b36ccac039a6c068994ee1e519aa075d0a0a56dd07ace8f323b233411

fredrikekelund:
I think we make the code more difficult to understand by moving validators out of the command files, so I'm instinctively against that. 

Seeing as we're on the home stretch of the CLI i2 project, I say let's keep the logic consolidated in command files for now.

nightnei:
@fredrikekelund sure, I am ok, it's just proposition, let's stick with existing approach 👍

PR #2228, opened 2025-12-09
Unify bumpAggregatedUniqueStat between CLI and App

epeicher:
## Related issues

- Solves STU-1052

## Proposed Changes

- Unify `bumpAggregatedUniqueStat` function between CLI and App
- Creates a new `AppdataProvider` type to abstract the differences between CLI and App access to the Appdata file
- Creates a new type to represent the `lastBumpStats` common property of UserData

## Testing Instructions

- Locate your `appdata-v1.json` file under `~/Library/Application Support/Studio/appdata-v1.json` and delete the values in the `lastBumpStats` entry
- Check in the terminal that you can see the following stats that would have been bumped:
```
Would have bumped stat: studio-app-launch-total=darwin
Would have bumped stat: studio-app-launch-uniqs-mon=darwin
Would have bumped stat: studio-app-launch-uniques=darwin
```
- Build the CLI by running `npm run cli:build`
- Run the CLI by opening a terminal and running a studio command, for example the following from the Studio repo root folder:
```sh
 NODE_ENV=development node dist/cli/main.js site list
```
- Check that you can see in the terminal your sties and `Would have bumped stat: studio-cli-usage-unique=success`

## Pre-merge Checklist


src/lib/tests/bump-stats.test.ts
```diff
@@ -1,15 +1,32 @@
...
 import { AggregateInterval, StatsGroup, StatsMetric } from 'common/types/stats';
-import { bumpAggregatedUniqueStat } from 'src/lib/bump-stats';
 
 jest.mock( 'atomically', () => ( {
 	readFile: jest.fn(),
 	writeFile: jest.fn(),
 } ) );
 
+// Mock appdata provider for tests
+const mockAppdataProvider: AppdataProvider< LastBumpStatsData > = {
```
katinthehatsite, line 18: I am wondering if we can simplify the mock to something like this: 

let mockData: LastBumpStatsData = {};

const mockAppdataProvider: AppdataProvider< LastBumpStatsData > = {
    load: async () => mockData,
    lock: async () => {},
    unlock: async () => {},
    save: async ( data ) => { mockData = data; },
};

Although, it is not a huge preference

katinthehatsite (approved):
The functionality worked correctly for me, I did not spot any regressions either for the app or for CLI:

Image

Image

Added a small suggestion and a question. Otherwise 👍

cli/index.ts
```diff
@@ -43,7 +55,8 @@ async function main() {
 				await bumpAggregatedUniqueStat(
 					StatsGroup.STUDIO_CLI_USAGE_UNIQUE,
 					StatsMetric.SUCCESS,
-					'weekly'
+					'weekly',
+					cliAppdataProvider
```
katinthehatsite: Should the error also be caught here? I see that this was an original code so this is not the question that is specifically related to the changes in this PR but more a general one :) 

Feel free to merge as is since this is replicating an original behaviour.

epeicher: That's a very good point, I didn't add error handling here to replicate original behaviour as you mention, but I think we can replicate the `.catch( ( err ) => Sentry.captureException( err ) );` of [the other usage](https://github.com/Automattic/studio/pull/2228/changes/BASE..784376ce9d9b24a2ff33ee944b66de9d2ef96d01#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R353)

src/lib/tests/bump-stats.test.ts
epeicher: Thanks for the suggestion, I will iterate on this, if it works, it seems simpler

katinthehatsite: Sounds good 👍 

It is not really necessary but could save a couple of lines of code.

cli/index.ts
katinthehatsite: Sounds good to me 👍

src/lib/tests/bump-stats.test.ts
epeicher: Ok, I have identified that the mock is asserted to be called [here for eample](https://github.com/Automattic/studio/pull/2228/changes#diff-cfbc6d0cda6ae1ff4b82a31f6414abf5b708f11393405bdfabbf64227e20706bR178), so it is helpful.

cli/index.ts
epeicher: After looking into this, I think we can wrap the call to `bumpAggregatedUniqueStat` in a `try/catch` and console.log the error message; at least the `yargs` flow will continue. What do you think @fredrikekelund?

epeicher: I have added the changes as part of [7f346c9](https://github.com/Automattic/studio/pull/2228/commits/7f346c9f90509a15852141c63b364fdf499fecb5), I will merge this after checks complete, and we can always iterate on other suggestions

PR #2233, opened 2025-12-10
CLI: Implement `studio site set-wp-version` command

fredrikekelund:
## Related issues

- Fixes STU-954

## Proposed Changes

This PR adds a `studio site set-wp-version` command that uses WP-CLI to update a site's WordPress version. If the site is already running, it's stopped and then restarted after WordPress is updated. 

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js site set-wp-version hello --path PATH_TO_SITE` (where `PATH_TO_SITE` is the path to a Studio site)
5. Ensure that it fails (`hello` is not a valid WP version)
6. Run `node dist/cli/main.js site set-wp-version 6.6 --path PATH_TO_SITE`
7. Ensure that it succeeds
8. Run `node dist/cli/main.js site start --path PATH_TO_SITE`
9. Run `node dist/cli/main.js site set-wp-version 6.7 --path PATH_TO_SITE`
10. Ensure that the site restarts successfully
11. Navigate to `/wp-admin/update-core.php` and ensure that 6.7 is, in fact, installed

## Pre-merge Checklist


nightnei (changes requested):
The incorrect version works well.

The correct version:
1. WP version is changed in Studio<br />Screenshot 2025-12-11 at 12 21 28
2. However, there are bugs - the command stuck at `WordPress site started` and can't be terminated with `control+C`<br >
Screenshot 2025-12-11 at 12 19 23

cli/commands/site/set-wp-version.ts
```diff
@@ -0,0 +1,128 @@
...
+							__(
+								'Invalid WordPress version. Must be "latest", "nightly", or a valid version number (e.g., "6.4", "6.4.1", "6.4-beta1").'
+							)
+						);
+					}
+					if ( ! isWordPressVersionAtLeast( value, MINIMUM_WORDPRESS_VERSION ) ) {
+						throw new LoggerError(
+							__( `WordPress version must be at least ${ MINIMUM_WORDPRESS_VERSION }.` )
+						);
+					}
```
nightnei, line 116: Let's use for consistency:
```suggestion
					if ( ! isValidWordPressVersion( value ) ) {
		throw new ValidationError(
			'wp',
			value,
			__(
				'Must be: "latest", "nightly", or a valid version number (e.g., "6.4", "6.4.1", "6.4-beta1")'
			)
		);
	}

	if ( ! isWordPressVersionAtLeast( value, MINIMUM_WORDPRESS_VERSION ) ) {
		throw new ValidationError(
			'wp',
			value,
			sprintf( __( 'Must be: at least %s' ), MINIMUM_WORDPRESS_VERSION )
		);
	}
```

cli/commands/site/set-wp-version.ts
```diff
@@ -0,0 +1,128 @@
...
+
+		await closeWpCliServer();
+	} finally {
+		disconnect();
+	}
+}
+
+export const registerCommand = ( yargs: StudioArgv ) => {
+	return yargs.command( {
+		command: 'set-wp-version <wp-version>',
```
nightnei, line 93: Should we make it short as `set-wp`? I think it shoudl be obvious, since we have `--wp` command, and probably users should not be confused with what to provide here

cli/commands/site/set-wp-version.ts
```diff
@@ -0,0 +1,128 @@
...
+		const processDescription = await isServerRunning( site.id );
+
+		if ( processDescription ) {
+			logger.reportStart( LoggerAction.STOP_SITE, __( 'Stopping WordPress site...' ) );
+			await stopWordPressServer( site.id );
+			logger.reportSuccess( __( 'WordPress site stopped' ) );
+		}
+
+		logger.reportStart( LoggerAction.SET_WP_VERSION, __( 'Changing WordPress version...' ) );
+		const phpVersion = validatePhpVersion( site.phpVersion );
```
nightnei, line 47: Do we need this validation? If it's already saved, I think it should be already valid
The same question is about this usecase https://github.com/Automattic/studio/pull/2233/files#diff-2a2c48d6e100d0f119ae75d1ee43474a44ca2e0c9d1de6ebeecaf28f4c6e47aaR32

fredrikekelund: We need this validation because the Zod schema for the appdata file parses `phpVersion` as a string rather than an enum. Playground uses a stricter type for the `php` config prop to the `runCLI`, and that's where this value ends up.

fredrikekelund:
> the command stuck at WordPress site started and can't be terminated with control+C

There used to be a `process.exit()` call. Unfortunately, I overlooked the fact that this was removed by my AI agent before pushing the PR 😐 I've brought it back.

cli/commands/site/set-wp-version.ts
fredrikekelund: I disagree that it would be obvious that `set-wp` means setting the WordPress version. I don't expect people to run this command very often, so I would argue it's preferable to keep it verbose (and immediately understandable) rather than short and quick to type.

fredrikekelund:
Thanks for the review, @nightnei. This is good for another look now 👍

cli/commands/site/set-wp-version.ts
```diff
@@ -0,0 +1,134 @@
...
+	logger.reportSuccess( __( 'Site loaded' ) );
+
+	try {
+		await connect();
+		const processDescription = await isServerRunning( site.id );
+
+		if ( processDescription ) {
+			logger.reportStart( LoggerAction.STOP_SITE, __( 'Stopping WordPress site...' ) );
+			await stopWordPressServer( site.id );
+			logger.reportSuccess( __( 'WordPress site stopped' ) );
```
bcotrim, line 43: Probably unrelated to this PR but I found this translation is broken in Portuguese:
("iniciado" means started and that should be the stopped site message)

Image

I checked Spanish and it seems to have the same behavior.

bcotrim (changes requested):
Code LGTM 👍 
One issue that I think we should address.
In Studio, 

| After install | After wp update |
| ------------ | ---------------- |
| Image | Image |

In CLI,
| After install | After wp update |
| ------------ | ---------------- |
| Image | Image |

The expected behavior would be to only receive updates when the site is configured to use `latest`. This should be controlled by the `isWpAutoUpdating` in SiteData/

cli/commands/site/set-wp-version.ts
nightnei: Good point 👍

nightnei (approved):
Thanks for addressing, LGTM and works as expected 👍

fredrikekelund:
> The expected behavior would be to only receive updates when the site is configured to use latest. This should be controlled by the isWpAutoUpdating in SiteData/

Good catch, @bcotrim! Thanks. I've addressed this. PR should be good for a final review now.

cli/commands/site/set-wp-version.ts
fredrikekelund: It looks like the `WordPress site stopped` string isn't translated yet. See https://translate.wordpress.com/projects/studio/pt/default?filters%5Bterm%5D=wordpress+site+stopped

PR #2235, opened 2025-12-10
Studio: Show all WordPress.com sites when importing, not just unconnected ones

epeicher:
## Related issues

- Fixes STU-1123

## Proposed Changes

- Made `connectedSiteIds` parameter optional in the `useGetWpComSitesQuery` API to avoid filtering them
- Updated the `transformSitesResponse` function to handle optional `connectedSiteIds` using optional chaining
- Do not pass `connectedSiteIds` parameter to the `useGetWpComSitesQuery` in  the PullRemoteSite component

## Testing Instructions

1. Apply this branch and run `npm start`
2. Ensure you have at least one local site already connected to a WordPress.com site
3. Click "Add Site" and select "Pull an existing site"
4. Verify that ALL your WordPress.com sites are displayed in the list, including those already connected to other local sites
5. Confirm you can select and import any site from the list, even if it's already connected elsewhere
6. Navigate to the Sync tab on a site with a connected site
7. Click on `Connect another site`
8. Verify that the connected site appears in the list greyed out with the `Already connected` label

## Pre-merge Checklist

### DIFF
--- a/src/modules/add-site/components/pull-remote-site.tsx
+++ b/src/modules/add-site/components/pull-remote-site.tsx
@@ -11,12 +11,10 @@ import offlineIcon from 'src/components/offline-icon';
 import { Tooltip } from 'src/components/tooltip';
 import { useAuth } from 'src/hooks/use-auth';
 import { useOffline } from 'src/hooks/use-offline';
-import { useSiteDetails } from 'src/hooks/use-site-details';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { NoWpcomSitesContent } from 'src/modules/sync/components/no-wpcom-sites-content';
 import { SitesListContent } from 'src/modules/sync/components/sync-sites-modal-selector';
 import { SyncTabImage } from 'src/modules/sync/components/sync-tab-image';
-import { useGetConnectedSitesForLocalSiteQuery } from 'src/stores/sync/connected-sites';
 import { useGetWpComSitesQuery } from 'src/stores/sync/wpcom-sites';
 import type { SyncSite } from 'src/modules/sync/types';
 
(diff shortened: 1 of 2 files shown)


sejas (approved):
Thanks for fixing the issue. I confirm that now the list of sites don't exclude the already connected sites when creating a new site, and it correctly appears when connecting a site.

I wonder if it would be possible to reuse the same request fingerprint to take advantage of the cached state. Currently, the list of sites in the Add Site modal takes longer to appear because it doesn't use the cached data from the other modal. I noticed that we accidentally removed the loading state in a different PR, which makes the issue more visible.

Image
Image

katinthehatsite:
This worked well for me:

Screenshot 2025-12-10 at 8 23 52 PM

Screenshot 2025-12-10 at 8 23 01 PM

> I wonder if it would be possible to reuse the same request fingerprint to take advantage of the cached state. Currently, the list of sites in the Add Site modal takes longer to appear because it doesn't use the cached data from the other modal. I noticed that we accidentally removed the loading state in a different PR, which makes the issue more visible.

I also noticed the same issue. I think we should definitely solve for this but it does not have to be in this PR. 

I also suspect that it might have been me who removed it 🙈

katinthehatsite (approved):

epeicher:
Thanks @sejas and @katinthehatsite for your reviews 🙇 ! 

> I wonder if it would be possible to reuse the same request fingerprint to take advantage of the cached state. Currently, the list of sites in the Add Site modal takes longer to appear because it doesn't use the cached data from the other modal. I noticed that we accidentally removed the loading state in a different PR, which makes the issue more visible.

I am going to create a follow-up to recover the loading state. I think the list of sites is cached already. I will investigate further in the follow-up

PR #2238, opened 2025-12-11
Auth: Display authorization denied message to the user access is denied

gavande1:
…ation process when access is denied.

## Related issues

- Fixes STU-1103

## Proposed Changes

- Display specific error message to the user when authorization is explicitly denied by them while login into Studio.

## Testing Instructions

- Switch to this branch  
- Run `npm start`  
- If you’re already logged into WordPress.com, log out  
- Click Log In  
- When the OAuth page opens up, click on Deny  
- You’ll be taken back to the Studio with a message saying you denied the request

| Before | After |
|--------|--------|
| CleanShot 2025-12-11 at 19 22 13@2x | CleanShot 2025-12-11 at 19 21 49@2x | 

## Pre-merge Checklist

### DIFF
--- a/src/components/auth-provider.tsx
+++ b/src/components/auth-provider.tsx
@@ -70,7 +70,7 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 			if ( payload.error instanceof Error && payload.error.message.includes( 'access_denied' ) ) {
 				title = __( 'Authorization denied' );
 				message = __(
-					'You have denied access to the authorization process. Please click the "Approve" option to authorize.'
+					'It looks like you denied the authorization request. To proceed, please click "Approve"'
 				);
 			}
 
--- a/src/components/auth-provider.tsx
+++ b/src/components/auth-provider.tsx
@@ -63,10 +63,18 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 
 	useIpcListener( 'auth-updated', ( _event, payload ) => {
 		if ( 'error' in payload ) {
-			getIpcApi().showErrorMessageBox( {
-				title: __( 'Authentication error' ),
-				message: __( 'Please try again.' ),
-			} );
+			let title: string = __( 'Authentication error' );
+			let message: string = __( 'Please try again.' );
+
+			// User has denied access to the authorization dialog.
+			if ( payload.error instanceof Error && payload.error.message.includes( 'access_denied' ) ) {
+				title = __( 'Authorization denied' );
+				message = __(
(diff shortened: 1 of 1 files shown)


src/components/auth-provider.tsx
```diff
@@ -63,10 +63,18 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
...
-				message: __( 'Please try again.' ),
-			} );
+			let title: string = __( 'Authentication error' );
+			let message: string = __( 'Please try again.' );
+
+			// User has denied access to the authorization dialog.
+			if ( payload.error instanceof Error && payload.error.message.includes( 'access_denied' ) ) {
+				title = __( 'Authorization denied' );
+				message = __(
+					'You have denied access to the authorization process. Please click the "Approve" option to authorize.'
```
gavande1: I’m not super happy with the copy, especially the last part. Please feel free to suggest a better version.

ivan-ottinger: What if we used something as follows?



```suggestion
					'It looks like you declined the authorization request. Please try to log in again and click "Approve".'
'
```

ivan-ottinger (approved):
Nice improvement! The code looks good. I have left a copy suggestion.

epeicher (approved):
Thanks for improving this @gavande1! I have tested it, and I can see now the new message, which sounds less scary. I agree with the suggestion from @ivan-ottinger, but overall this LGTM!

Image

src/components/auth-provider.tsx
katinthehatsite: Just wanted to add a note that the authorization page does not self-close when you click on "Deny" and after it brings you to electron so it still stays open. Technically, the user does not need to try to log in again if they did not close that page manually. The authorization page should still be available in that case.

katinthehatsite:
```suggestion
					'Access to the authorization process was denied. Please select “Approve” to continue.'
```

Another suggestion:


```suggestion
					'It looks like you denied the authorization request. To proceed, please click “Approve.”'
```

katinthehatsite (approved):
Looks good to me 👍 I left some suggestions for the copy.

src/components/auth-provider.tsx
gavande1: Thanks @ivan-ottinger for the suggestion. I agree with Kat that the screen does not close automatically when a user denies the request. This means the user does not need to log in again to approve it. For this reason, I think Kat’s suggestion is the best copy we can use.

ivan-ottinger: Sounds good to me. 👍🏼

PR #2239, opened 2025-12-11
Avoid showing a path conflict when pulling a remote site multiple times

sejas:
## Related issues

- Fixes STU-1118

## Proposed Changes

- When pulling an existing remote site multiple times, the user received an error conflicting in the path. In this PR I suggest appending a suffix number to generate a new clean path.

## Testing Instructions

- Run `npm start`
- Click on Add site and then select Pull an existing site
- Continue and create the site
- Observe that the site name does not contain any numeric suffix.
- Repeat the operation selecting the same site.
- Observe that the generated name is "Sitename 2" so it doesn't show any error to the user.

| Before | After |
|--------|--------|
| Screenshot 2025-12-11 at 17 19 41 |Screenshot 2025-12-11 at 17 21 30 |

https://github.com/user-attachments/assets/55e1f6c0-7b84-4c0f-ba59-c641a39d1acd

## Pre-merge Checklist

### DIFF
--- a/src/modules/add-site/index.tsx
+++ b/src/modules/add-site/index.tsx
@@ -123,12 +123,14 @@ function NavigationContent( props: NavigationContentProps ) {
 		}
 	}, [ createSiteProps, goTo ] );
 
+	const findAvailableSiteName = useFindAvailableSiteName();
 	const handlePullRemoteContinue = useCallback( async () => {
 		if ( selectedRemoteSite ) {
-			void createSiteProps.handleSiteNameChange( selectedRemoteSite.name );
+			const availableName = await findAvailableSiteName( selectedRemoteSite.name );
+			void createSiteProps.handleSiteNameChange( availableName );
 			goTo( '/pullRemote/create' );
 		}
-	}, [ createSiteProps, goTo, selectedRemoteSite ] );
+	}, [ createSiteProps, findAvailableSiteName, goTo, selectedRemoteSite ] );
 
 	const blueprints = useMemo(
 		() => blueprintsData?.blueprints.slice().reverse() || [],
--- a/src/modules/add-site/index.tsx
+++ b/src/modules/add-site/index.tsx
@@ -127,7 +127,7 @@ function NavigationContent( props: NavigationContentProps ) {
 	const handlePullRemoteContinue = useCallback( async () => {
 		if ( selectedRemoteSite ) {
 			const availableName = await findAvailableSiteName( selectedRemoteSite.name );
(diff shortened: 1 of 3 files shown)


src/modules/add-site/index.tsx
```diff
@@ -277,11 +281,8 @@ function NavigationContent( props: NavigationContentProps ) {
 			<Navigator.Screen className="flex-1 flex justify-center" path="/pullRemote">
 				<PullRemoteSite
 					selectedRemoteSite={ selectedRemoteSite }
-					setSelectedRemoteSite={ ( remoteSite?: SyncSite ) => {
+					setSelectedRemoteSite={ async ( remoteSite?: SyncSite ) => {
```
epeicher, line 284: I think we can remove the `async` keyword from here

src/modules/add-site/index.tsx
```diff
@@ -122,11 +123,14 @@ function NavigationContent( props: NavigationContentProps ) {
 		}
 	}, [ createSiteProps, goTo ] );
 
-	const handlePullRemoteContinue = useCallback( () => {
+	const findAvailableSiteName = useFindAvailableSiteName();
+	const handlePullRemoteContinue = useCallback( async () => {
 		if ( selectedRemoteSite ) {
+			const availableName = await findAvailableSiteName( selectedRemoteSite.name );
+			void createSiteProps.handleSiteNameChange( availableName );
```
epeicher: Now that we have moved this `createSiteProps.handleSiteNameChange( availableName )` to an async function, maybe we can get rid of the `void` operator and replace it for an `await`? I think we should minimize the use of `void` to very specific use cases. But this is my personal opinion, happy to leave it as is, as it was the previous behavior.

epeicher (approved):
Thanks @sejas for solving this! I have had that issue in the past, and I thought that receiving an error was not a great UX, so I like this! 

I have tested it, and it works as expected, I can see the sites with the number added and increasing. I have left a couple of minor comments, but this LGTM!

|Existing site | Selecting the same site| Suffix added | Suffix increased for subsequent sites|
|-------|------|------|------|
|         Image| Image|Image|Image|

src/modules/add-site/index.tsx
sejas: Changed it here https://github.com/Automattic/studio/pull/2239/commits/141c3f2913054df95652e9de6869842c10ffc37e

katinthehatsite (approved):
The changes look good to me, I am not seeing any regressions and the code reads well as well 👍

PR #2242, opened 2025-12-12
Add Xdebug support

wojtekn:
## Related issues

- Fixes STU-1156

## Proposed Changes

  - Add Xdebug support for WordPress sites with beta feature flag
  - Implement UI toggle in Edit Site Details modal with one-site-at-a-time constraint
  - Add full persistence across app restarts (site details, storage, and IPC layer)
  - Show tooltip when Xdebug is enabled on another site
  - Show Xdebug status in site's Settings

|**Xdebug enabled for a site**|**Restarting site on save**|**Xdebug not available for other site**|
|---|---|---|
|CleanShot 2025-12-17 at 16 09 38|CleanShot 2025-12-17 at 16 09 44|CleanShot 2025-12-17 at 16 10 02|

### Test feature in Studio

  1. Enable "Xdebug Support" beta feature in Settings → Beta Features
  2. Open Edit Site Details for any site
  3. Check "Enable Xdebug" checkbox and save
  4. Verify site restarts ~~and console shows "Enabling Xdebug support"~~
  5. Open Edit Site Details again - verify checkbox remains checked
  6. Restart the app - verify Xdebug setting persists
  7. Try enabling Xdebug on a second site - verify checkbox is disabled with tooltip explaining why
  8. Disable Xdebug on first site - verify second site's checkbox becomes enabled

### Test debugging in IDE

I used PhpStorm:

1. Create a site
2. Enable Xdebug and start the site
3. Open the site in PhpStorm
4. To stop breaking at the first line in PHPStorm, open Settings > PHP > Debug > Xdebug:
    - Disable > Force break at first line when no path mapping specified
    - Disable > Force break at first line when a script is outside the project
5. Open WP Admin using link in Studio
6. In PhpStorm, open the path mapping notification (or navigate to PHP|Servers) and configure mapping for your site's server:
   - Host: localhost:8884
   - Port: 80
   - File directory: /Users/USERNAME/Studio/SITEDIR
   - Absolute path on the server: `/wordpress`
7. Add a breakpoint somewhere, e.g. in wp-config.php file
8. Refresh WP Admin
9. Confirm PhpStorm stops execution on the breakpoint

## Pre-merge Checklist

### DIFF
--- a/src/lib/wordpress-provider/playground-cli/playground-server-process-child.ts
+++ b/src/lib/wordpress-provider/playground-cli/playground-server-process-child.ts
@@ -232,7 +232,6 @@ async function startServer(
 		if ( options.enableXdebug ) {
 			console.log( __( 'Enabling Xdebug support' ) );
 			args.xdebug = true;
-			//args.experimentalUnsafeIdeIntegration = 'phpstorm';
 		}
 
 		if ( options.phpVersion ) {
--- a/src/lib/wordpress-provider/playground-cli/playground-server-process-child.ts
+++ b/src/lib/wordpress-provider/playground-cli/playground-server-process-child.ts
@@ -228,6 +228,13 @@ async function startServer(
 			args.experimentalMultiWorker = workerCount;
 		}
 
+		// Enable Xdebug support if requested
+		if ( options.enableXdebug ) {
+			console.log( __( 'Enabling Xdebug support' ) );
+			args.xdebug = true;
+			//args.experimentalUnsafeIdeIntegration = 'phpstorm';
+		}
+
 		if ( options.phpVersion ) {
 			args.php = options.phpVersion as SupportedPHPVersion;
 		}
--- a/src/modules/site-settings/edit-site-details.tsx
+++ b/src/modules/site-settings/edit-site-details.tsx
@@ -342,7 +342,15 @@ const EditSiteDetails = ( { currentWpVersion, onSave }: EditSiteDetailsProps ) =
 							</div>
 
 							{ isXdebugFeatureEnabled && (
-								<div className="flex flex-col gap-2 mt-4 p-4 border border-a8c-gray-200 rounded">
+								<div
+								className={ cx(
+									'flex flex-col gap-2 mt-4 p-4 border border-a8c-gray-200 rounded',
+									isEditingSite ||
+										( xdebugEnabledSite && xdebugEnabledSite.id !== selectedSite?.id )
+										? 'opacity-50 cursor-not-allowed'
+										: ''
+								) }
+							>
 									<Tooltip
 										disabled={ ! xdebugEnabledSite || xdebugEnabledSite.id === selectedSite?.id }
 										text={ sprintf(
--- a/src/modules/site-settings/edit-site-details.tsx
+++ b/src/modules/site-settings/edit-site-details.tsx
@@ -343,14 +343,14 @@ const EditSiteDetails = ( { currentWpVersion, onSave }: EditSiteDetailsProps ) =
 
 							{ isXdebugFeatureEnabled && (
 								<div
-								className={ cx(
-									'flex flex-col gap-2 mt-4 p-4 border border-a8c-gray-200 rounded',
-									isEditingSite ||
-										( xdebugEnabledSite && xdebugEnabledSite.id !== selectedSite?.id )
-										? 'opacity-50 cursor-not-allowed'
-										: ''
-								) }
-							>
+									className={ cx(
+										'flex flex-col gap-2 mt-4',
+										isEditingSite ||
+											( xdebugEnabledSite && xdebugEnabledSite.id !== selectedSite?.id )
+											? 'opacity-50 cursor-not-allowed'
+											: ''
+									) }
+								>
 									<Tooltip
 										disabled={ ! xdebugEnabledSite || xdebugEnabledSite.id === selectedSite?.id }
 										text={ sprintf(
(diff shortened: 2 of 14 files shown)


src/modules/site-settings/edit-site-details.tsx
```diff
@@ -317,6 +340,43 @@ const EditSiteDetails = ( { currentWpVersion, onSave }: EditSiteDetailsProps ) =
...
 								) }
 							</div>
+
+							{ isXdebugFeatureEnabled && (
+								<div className="flex flex-col gap-2 mt-4 p-4 border border-a8c-gray-200 rounded">
+									<Tooltip
+										disabled={ ! xdebugEnabledSite || xdebugEnabledSite.id === selectedSite?.id }
+										text={ sprintf(
+											__(
+												'Xdebug is currently enabled for "%s" site. Disable it there first to enable it for this site.'
```
bcotrim, line 358: Have you considered allowing the user to enable here with a notice that it will disable the other site?

  The current flow requires ~6 steps: 

close modal → switch site → open modal → disable Xdebug → save → switch back → open modal → enable Xdebug. 

A single confirmation would be much smoother.

  What do you think?

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -317,6 +340,43 @@ const EditSiteDetails = ( { currentWpVersion, onSave }: EditSiteDetailsProps ) =
 									</div>
 								) }
 							</div>
+
+							{ isXdebugFeatureEnabled && (
+								<div className="flex flex-col gap-2 mt-4 p-4 border border-a8c-gray-200 rounded">
```
bcotrim: The current disabled state does not look very clear in my opinion. It uses the disabled format for the checkbox but the label could also have some visible changes.
The tooltip provides all the information, but we could make it clear to the user that the option is disabled just by looking at it:

Image

What do you think?

src/modules/site-settings/edit-site-details.tsx
```diff
@@ -317,6 +340,43 @@ const EditSiteDetails = ( { currentWpVersion, onSave }: EditSiteDetailsProps ) =
...
+												}
+											/>
+											<label htmlFor="enable-xdebug" className="font-semibold">
+												{ __( 'Enable Xdebug' ) }
+											</label>
+										</div>
+									</Tooltip>
+									<div className="text-a8c-gray-50 text-xs">
+										{ __(
+											'Enable PHP debugging with Xdebug. Only one site can have Xdebug enabled at a time.'
```
bcotrim: Could we have a link here to the Studio documentation with information on how to configure and use Xdebug?
I was thinking something similar to the HTTPS one:

Image

What do you think?

bcotrim (commented):
Works great! Amazing to see this coming to Studio

Have you considered adding the Xdebug flag in the site details section?

Image

src/lib/wordpress-provider/playground-cli/playground-server-process-child.ts
```diff
@@ -228,6 +228,12 @@ async function startServer(
 			args.experimentalMultiWorker = workerCount;
 		}
 
+		// Enable Xdebug support if requested
+		if ( options.enableXdebug ) {
+			console.log( __( 'Enabling Xdebug support' ) );
+			args.xdebug = true;
```
sejas, line 234: Should we also add `experimentalUnsafeIdeIntegration` argument or is not necessary? I tried it to add it, but it didn't help during my testing.

```
args.experimentalUnsafeIdeIntegration = [ 'vscode', 'phpstorm' ];`
```

sejas (commented):
Code looks good. I confirm it adds a feature flag, and enables the Xdebug parameter when the checkbox is active. I confirm the site property `enableXdebug` is saved in one site at a time in Studio settings file.

Image

I was not able to make it work with Cursor, VSCode and PhpStorm. @wojtekn, I would appreciate pairing to double check my testing steps. I followed your testing instructions and also those provided in https://make.wordpress.org/playground/2025/11/24/debugging-with-xdebug-is-now-available-in-wordpress-playground/

I agree with @bcotrim's [suggestion](https://github.com/Automattic/studio/pull/2242#discussion_r2626702024). It would be great if users could enable Xdebug on any site, and Studio could automatically remove the flag from the previous site. Maybe we could display a dismissible modal to confirm that action.

Image

A couple of ideas to improve the UI that could be follow-ups:

**1. Make the checkbox block consistent with Use custom domain and Enable HTTPS**

I suggest making the checkbox more consistent between them, by removing the border around the Xdebug option or adding the border to the rest of checkboxes.

Image

**2. Add the action to Enable/Disable Xdebug in the site actions menu.**

Image

Enable Xdebug would automatically disable the option from any other site, and start or restart the site.

**3. Replace the green dot with a green 🪲 when Xdebug is running in a given site.**

Image

PhpStorm and Cursor display the play button like this:

Image
Image

bcotrim:
> I was not able to make it work with Cursor, VSCode and PhpStorm.

I had some issue and was missing the `pathMappings` in VSCode. Sharing my launch.json in case it helps

```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Listen for Xdebug",
      "type": "php",
      "request": "launch",
      "port": 9003,
      "pathMappings": {
        "/wordpress": "${workspaceFolder}"
      }
    }
  ]
}

```

> 3. Replace the green dot with a green 🪲 when Xdebug is running in a given site.

❤️ I love this idea.

src/lib/wordpress-provider/playground-cli/playground-server-process-child.ts
wojtekn: No, because it's experimental and unsafe :D I tested it, and besides being broken, it was only updating IDE configs. I think it's better if Studio enables user to use Xdebug, instead of modifying their IDE configs.

src/modules/site-settings/edit-site-details.tsx
wojtekn: Yes, but I wanted to avoid overcomplicating this in this initial Beta Feature release.

This is a bit of a rabbit hole. If Studio disabled Xdebug for all sites, it should restart them. If it restarts sites, it should first ask users if they want to restart them.

wojtekn:
> Have you considered adding the Xdebug flag in the site details section?

Good idea. Added in 24204afd60028067acd0f49586ce4faf64ed546e

src/modules/site-settings/edit-site-details.tsx
wojtekn: I changed it in bc36147ab2b48bf5588d0a0eff4a2a51d6d76c64

src/modules/site-settings/edit-site-details.tsx
wojtekn: I added "Learn more" link placeholder: 1bd4aa1c4dd294ecffbf31bbd01f2f5c22561f4d

wojtekn:
> **1. Make the checkbox block consistent with Use custom domain and Enable HTTPS**
> 
> I suggest making the checkbox more consistent between them, by removing the border around the Xdebug option or adding the border to the rest of checkboxes.

Thanks, I updated it in 1bd4aa1c4dd294ecffbf31bbd01f2f5c22561f4d

> 2. Add the action to Enable/Disable Xdebug in the site actions menu.

The site context menu has "Edit site..." which contains Xdebug field, so I would hold with this change at this point.

> 3. Replace the green dot with a green 🪲 when Xdebug is running in a given site.

Sounds cool, but I would hold with this under this PR, and we could push follow-up after we check it more with the design team.

@sejas let's pair tomorrow to see how it works on your environment.

bcotrim:
@wojtekn  I added the CLI changes to support x-debug flag.
Please let me know if you find any regressions.

I also implemented the `studio site set-xdebug` to follow the other `set-*` commands we have.
cc @fredrikekelund if you want to take a quick look at the implementation

bcotrim (commented):
Thanks @wojtekn for the adjustments! 
The UI looks great.
Since I've made some changes to this PR I'll leave the approval for another reviewers but it's a LGTM on my side 👍

wojtekn:
Thanks @bcotrim .

sejas (approved):
It worked for me. I tested it using Cursor and the this [launch.json](https://github.com/Automattic/studio/pull/2242#issuecomment-3665228079) and it worked great. @bcotrim, thanks for sharing the `pathMappings` trick.

Note that the breakpoint set in wp-config.php was ignored, while breakpoints in other parts of the app, such as in a plugin, worked correctly.

Image

I just noticed that we are not consistent with the bold/regular font weight in the checkbox. I recommend making them consistent or creating a new issue to discuss that.

Image

wojtekn:
> Note that the breakpoint set in wp-config.php was ignored, while breakpoints in other parts of the app, such as in a plugin, worked correctly.

It works for me when I debug in PhpStorm - I can set breakpoints in `wp-config.php` file. 

> I just noticed that we are not consistent with the bold/regular font weight in the checkbox. I recommend making them consistent or creating a new issue to discuss that.

I fixed this, good catch.

I tested it again after Studio CLI changes, and it works fine, too. The only change is that console output doesn't show Playground CLI messaging so "Enabling Xdebug support" is not visible there.

PR #2244, opened 2025-12-12
Render app arch inside About dialog

nightnei:
## Proposed Changes
I am working on STU-1069, and I didn't know how to determine whether the arm version is installed or not. I had an assumption that maybe the bug is on the frontend and it's just UI bug. I decided that would be usefull to somewhere render the arch. So I added it as a title for the Studio version, to avoid being intrusive in the dialog. I think it will be useful for our team in the future.

## Testing Instructions
1. Build Studio and run it on macOS and Windows.
2. Assert that you see correct app version and arch inside About window<br />Screenshot 2025-12-12 at 15 58 09

### DIFF
--- a/src/about-menu/open-about-menu.ts
+++ b/src/about-menu/open-about-menu.ts
@@ -1,7 +1,7 @@
 import { BrowserWindow, app } from 'electron';
 import path from 'path';
 import * as Sentry from '@sentry/electron/renderer';
-import { sprintf, __ } from '@wordpress/i18n';
+import { __ } from '@wordpress/i18n';
 import { ABOUT_WINDOW_HEIGHT, ABOUT_WINDOW_WIDTH } from 'src/constants';
 import { shellOpenExternalWrapper } from 'src/lib/shell-open-external-wrapper';
 
@@ -16,10 +16,10 @@ function getPlatformLabel(): string {
 	const arch = process.arch;
 
 	if ( platform === 'darwin' ) {
-		return arch === 'arm64' ? 'macOS Apple Silicon' : 'macOS Intel';
+		return arch === 'arm64' ? __( 'Mac with Apple Silicon Chip' ) : __( 'Mac with Intel Chip' );
 	}
 	if ( platform === 'win32' ) {
-		return arch === 'arm64' ? 'Windows ARM64' : 'Windows x64';
+		return arch === 'arm64' ? __( 'Windows on ARM' ) : __( 'Windows on Intel/AMD' );
 	}
 	return `${ platform } ${ arch }`;
 }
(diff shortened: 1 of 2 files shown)


src/about-menu/open-about-menu.ts
```diff
@@ -51,11 +64,13 @@ export function openAboutWindow() {
 			const releasesText = escapeSingleQuotes( __( 'Release Notes' ) );
 			const demoSitesText = escapeSingleQuotes( __( 'Preview sites powered by' ) );
 			const localSitesText = escapeSingleQuotes( __( 'Local sites powered by' ) );
+			const systemInfoTooltip = escapeSingleQuotes( getPlatformLabel() );
 
 			const script = `
 				document.title = '${ aboutStudioText }';
 				document.getElementById('studio-by-wpcom').innerText = '${ studioByWpcomText }';
 				document.getElementById('version-text').innerText = '${ versionText }';
+				document.getElementById('version-text').title = '${ systemInfoTooltip }';
```
wojtekn: Actually, this information is essential for debugging, and it would be great if users could copy-paste it. What if we included it inline, in brackets, after the version?

src/about-menu/open-about-menu.ts
```diff
@@ -11,6 +11,19 @@ export function escapeSingleQuotes( str: string ) {
 	return str.replace( /\\/g, '\\\\' ).replace( /'/g, "\\'" );
 }
 
+function getPlatformLabel(): string {
+	const platform = process.platform;
+	const arch = process.arch;
+
+	if ( platform === 'darwin' ) {
+		return arch === 'arm64' ? 'macOS Apple Silicon' : 'macOS Intel';
```
wojtekn: What if we aligned those options with the exact wording we use on the landing page?

src/about-menu/open-about-menu.ts
nightnei: > What if we included it inline

I am totally on board with it. Actually, it was what I did initially, but then decided to hide 😄

bcotrim:
> The changes are quite readable in the code, and it will take a lot of time to generate all builds and test them. So I propose just to trust that I have tested it, to save your time :)

You should be able to trigger a build on CI and reviewers could use the generated artifacts to validate the changes without needing to build locally.

nightnei:
I printed arch and made the line selectable.
I wanted to add "Copy" button, but I think it's overkill for this case, because it's necessary to use IPC, since browser's clipboard doesn't work in Electron. And I tried to include preload from main window, but it feels too much and also there we have [validation](https://github.com/Automattic/studio/blob/trunk/src/index.ts#L164-L176) issues. I was thinking about implementing own preload for About window, but in teh result I decided that it's too much code and it doesn't worth it. So as a result, I think that the best is to go with just a selectable line of text.
Screenshot 2025-12-12 at 15 58 09

wojtekn (approved):
Looks good:

|**Apple Silicon Chip**|**Intel Chip**|
|---|---|
|Image|Image|

sejas (approved):
I tested it on Mac and it works as expected. I tested it in English and Spanish.

| English | Spanish |
|--------|--------|
|  Image | Image |

PR #2245, opened 2025-12-12
CLI: Support reading resources in blueprints

fredrikekelund:
## Related issues

- Fixes STU-1095

## Proposed Changes

Studio has previously lacked support for referencing bundled resources (i.e., files) in blueprints. Per the [Playground docs](https://wordpress.github.io/wordpress-playground/blueprints/bundles), this is supported if users supply a ZIP file as the blueprint. Playground will then look up the `blueprint.json` file in the ZIP archive and read any referenced files from the archive.

This PR implements a similar approach for Studio: `bundled` resources are read from disk, relative to the location of the blueprint file. I believe this is fairly intuitive behavior, but we should make sure to also document it.

From a technical perspective, here's what this PR changes:
- Blueprint JSON files are still read/fetched and parsed as part of the `site create` command. We could theoretically just pass the URI to Playground CLI, which supports reading/fetching blueprints, too. The reason we don't do this is that we amend the blueprint contents in a few places before passing it to Playground CLI. 
- Instead of passing the raw blueprint contents to Playground CLI, we now pass a so-called `BlueprintBundle`. This interface contains a `read` method that reads the blueprint file from a specific path (`/blueprint.json`) and any bundled resources from another location (e.g., the disk or the network).

## Testing Instructions

1. Save the following contents in `~/blueprint.json`:

```json
{
  "$schema": "https://playground.wordpress.net/blueprint-schema.json",
  "meta": {
    "title": "Non-profit Organization",
    "description": "A non-profit organization site created with the Koinonia theme.",
    "author": "WordPress.com",
    "categories": ["Website", "Organization"]
  },
  "login": true,
  "siteOptions": {
    "blogname": "Koinonia"
  },
  "steps": [
    {
      "step": "resetData"
    },
    {
      "step": "installTheme",
      "themeData": {
        "resource": "bundled",
        "path": "./koinonia.zip"
      },
      "options": {
        "activate": true
      }
    }
  ]
}
```

2. Download the Koinonia theme from [here](https://wordpress.org/themes/koinonia/) and move the ZIP to `~/koinonia.zip`
3. Run `npm run cli:build`
4. Run `node dist/cli/main.js site create --blueprint ~/blueprint.json --path PATH_TO_NEW_SITE` where `PATH_TO_NEW_SITE` is the path where you want to create the new site
5. Ensure that the setup process completes successfully
6. In wp-admin, ensure that the Koinonia theme is installed

## Pre-merge Checklist


nightnei (approved):
LGTM and works as expected 👍
Screenshot 2025-12-15 at 11 26 21
Screenshot 2025-12-15 at 11 26 34
Screenshot 2025-12-15 at 11 26 45

PR #2248, opened 2025-12-15
Update WordPress Components packages to latest versions

wojtekn:
## Related issues

- N/A

## Proposed Changes

I propose to update WordPress Components to the latest versions.

## Testing Instructions

- Confirm that the build works fine and that the UI doesn't have any new issues.

## Pre-merge Checklist


package.json
```diff
@@ -152,13 +158,7 @@
 		"wpcom-xhr-request": "^1.3.0",
 		"yargs": "^18.0.0",
 		"yauzl": "^3.2.0",
-		"zod": "^3.24.3",
```
wojtekn, line 155: @bcotrim, it seems NPM keeps them organized, sorted by name, so I think we may leave it as is instead of trying to position the Playground packages at the end.

ivan-ottinger (approved):
I have tested the build and there don't seem to be any regressions. I tried with RTL locale as well and clicked through the UI. Nothing stood out of place. 👍🏼

katinthehatsite (approved):
Works as expected 👍

gavande1 (approved):
I have tested this and nothing seems to broken in term of UI. LGTM 👍

PR #2249, opened 2025-12-15
refactor studio to use cli for wp cli

bcotrim:
## Related issues

- Fixes STU-1124

## Proposed Changes

This PR refactors Studio to depend on the CLI for running WP-CLI commands.

- **Studio uses CLI for WP-CLI**: Replaced direct Playground CLI usage with `studio wp` command
- **Theme details via WP-CLI**: Replaced `runPHP` approach with a custom `wp studio get-theme-details` command registered via mu-plugin
- **Removed deprecated code**: Deleted `wp-cli-process.ts`, `wp-cli-executor.ts`, `php-get-theme-details.ts`

## Testing Instructions

1. **Test theme details loading**:
   - Start a site in Studio
   - Verify theme details show correctly (name, block theme status, widgets/menus support)
   - Switch themes and verify the UI updates

2. **Test site export**:
   - Create a new site via Studio UI
   - Verify WordPress installs correctly and site starts
   - Export the site in Studio
   - Verify the export works and the generated file is correct

3. **Test WP-CLI execution**:
   - Start a site and run WP-CLI commands from the assistant
   - Verify commands execute correctly

## Pre-merge Checklist


cli/commands/wp.ts
```diff
@@ -1,81 +1,113 @@
...
+	const phpVersionSchema = z.enum( SupportedPHPVersions );
+	let phpVersion: SupportedPHPVersion;
+	const requestedPhpVersion = options.phpVersion ?? site.phpVersion;
+
 	try {
-		await connect();
+		phpVersion = phpVersionSchema.parse( requestedPhpVersion );
+	} catch ( error ) {
+		throw new LoggerError( sprintf( __( 'Unsupported PHP version: %s' ), requestedPhpVersion ) );
+	}
```
fredrikekelund:
```ts
	const phpVersion = validatePhpVersion( options.phpVersion ?? site.phpVersion );
```
I would move the `catch` clause to `validatePhpVersion` and keep using that helper function to encapsulate this boilerplate-y logic.

cli/commands/wp.ts
```diff
@@ -1,81 +1,113 @@
...
+	logger.reportKeyValuePair( 'stderr', stderr );
+	logger.reportKeyValuePair( 'exitCode', String( exitCode ) );
 
 	await closeWpCliServer();
-	process.exit( await response.exitCode );
+	process.exit( exitCode );
 }
 
-function removePathArgumentFromArgv( argv: string[] ) {
+export function removeArgumentFromArgv( argv: string[], argName: string ): string[] {
```
fredrikekelund:
```suggestion
function removeArgumentFromArgv( argv: string[], argName: string ): string[] {
```
No need to export, AFAICT

cli/commands/wp.ts
```diff
@@ -1,81 +1,113 @@
...
-		argv.splice( pathIndex, 2 );
+	while ( argv.indexOf( `--${ argName }` ) !== -1 ) {
+		const argIndex = argv.indexOf( `--${ argName }` );
+		argv.splice( argIndex, 2 );
 	}
 
-	while ( argv.find( ( arg ) => /^--path=/.test( arg ) ) ) {
-		const pathIndex = argv.findIndex( ( arg ) => /^--path=/.test( arg ) );
-		argv.splice( pathIndex, 1 );
+	while ( argv.find( ( arg ) => argPattern.test( arg ) ) ) {
```
fredrikekelund:
```suggestion
	while ( argv.find( ( arg ) => arg.startsWith( `--${ argName }=` ) ) ) {
```
Now that we're constructing the regexp dynamically, I think it'd be better to use `String.prototype.startsWith` to avoid any pitfalls from interpolating `argName` in the regex constructor.

cli/commands/wp.ts
```diff
@@ -1,81 +1,113 @@
...
+}
+
+function removeAvoidTelemetryArgumentFromArgv( argv: string[] ) {
+	return removeArgumentFromArgv( argv, 'avoid-telemetry' );
+}
+
 export async function commandHandler( argv: ArgumentsCamelCase< GlobalOptions > ) {
 	try {
-		const wpCliArgv = removePathArgumentFromArgv( process.argv.slice( 3 ) );
+		let wpCliArgv = removePathArgumentFromArgv( process.argv.slice( 3 ) );
```
fredrikekelund:
```suggestion
		let wpCliArgv = removeArgumentFromArgv( process.argv.slice( 3 ), 'path' );
```

cli/commands/wp.ts
```diff
@@ -86,7 +118,11 @@ export async function commandHandler( argv: ArgumentsCamelCase< GlobalOptions >
 			);
 		}
 
-		await runCommand( argv.path, wpCliArgv );
+		const phpVersion = parsedWpCliArgs[ 'php-version' ] as string | undefined;
+		wpCliArgv = removePhpVersionArgumentFromArgv( wpCliArgv );
```
fredrikekelund:
```suggestion
		wpCliArgv = removeArgumentFromArgv( wpCliArgv, 'php-version' );
```

cli/commands/wp.ts
```diff
@@ -86,7 +118,11 @@ export async function commandHandler( argv: ArgumentsCamelCase< GlobalOptions >
...
+		wpCliArgv = removeAvoidTelemetryArgumentFromArgv( wpCliArgv );
```
fredrikekelund:
```suggestion
		wpCliArgv = removeArgumentFromArgv( wpCliArgv, 'avoid-telemetry' );
```

common/lib/mu-plugins.ts
fredrikekelund, line 1: *chef's kiss*

src/ipc-handlers.ts
```diff
@@ -674,39 +701,31 @@ export async function getThemeDetails(
...
+		await updateSite( event, server.details );
+
+		void ( async () => {
+			try {
+				await server.updateCachedThumbnail();
+				await sendThumbnailChangedEvent( event, id );
+			} catch ( error ) {
+				console.error( `Failed to update thumbnail for server ${ id }:`, error );
+			}
+		} )();
```
```suggestion
		try {
			await server.updateCachedThumbnail();
			await sendThumbnailChangedEvent( event, id );
		} catch ( error ) {
			console.error( `Failed to update thumbnail for server ${ id }:`, error );
		}
```
This has the same effect. We just don't want to propagate any exceptions

src/site-server.ts
```diff
@@ -335,45 +339,123 @@ export class SiteServer {
...
+					exitCode: 1,
+				} );
+			}
+			wpCliArgs = parsedArgs as string[];
+		} else {
+			wpCliArgs = args;
 		}
 
-		const wpCliArgs = parse( args );
+		const cliArgs: string[] = [ 'wp', '--path', projectPath ];
```
fredrikekelund:
```suggestion
		const cliArgs: string[] = [ 'wp', '--path', this.details.path ];
```
Nit, but I don't think `projectPath` is used anywhere else, so we can probably remove it.

src/site-server.ts
```diff
@@ -335,45 +339,123 @@ export class SiteServer {
...
+			cliArgs.push( '--skip-plugins', '--skip-themes' );
 		}
 
-		try {
-			return await this.wpCliExecutor.execute( wpCliArgs as string[], { phpVersion } );
-		} catch ( error ) {
-			if ( ( error as MessageCanceled )?.canceled ) {
-				return {
+		const isImportExport =
+			wpCliArgs[ 0 ] === 'sqlite' && [ 'import', 'export' ].includes( wpCliArgs[ 1 ] as string );
```
fredrikekelund:
```suggestion
			wpCliArgs[ 0 ] === 'sqlite' && [ 'import', 'export' ].includes( wpCliArgs[ 1 ] );
```
I don't believe the `as` operator is needed here

src/site-server.ts
```diff
@@ -335,45 +339,123 @@ export class SiteServer {
...
+		const timeout = isImportExport
+			? WP_CLI_IMPORT_EXPORT_RESPONSE_TIMEOUT
+			: WP_CLI_DEFAULT_RESPONSE_TIMEOUT;
+
+		return new Promise( ( resolve ) => {
+			const result: WpCliResult = { stdout: '', stderr: '', exitCode: 1 };
+			const [ emitter ] = executeCliCommand( cliArgs, { silent: true } );
+
+			const timeoutId = setTimeout( () => {
+				resolve( {
```
fredrikekelund:
```suggestion
			const [ emitter, process ] = executeCliCommand( cliArgs, { silent: true } );

			const timeoutId = setTimeout( () => {
				process.kill( 'SIGKILL' );
				resolve( {
```
Seems like a good idea to also kill the process if the operation times out.

src/site-server.ts
```diff
@@ -335,45 +339,123 @@ export class SiteServer {
...
+			emitter.on( 'error', ( { error } ) => {
+				clearTimeout( timeoutId );
+				Sentry.captureException( error );
+				resolve( {
+					stdout: '',
+					stderr: `Error executing WP-CLI command: ${ error.message }`,
+					exitCode: 1,
+				} );
+			} );
+		} );
```
fredrikekelund:
```ts
		let timeoutId: NodeJS.Timeout;

		return new Promise< WpCliResult >( ( resolve ) => {
			const result: WpCliResult = { stdout: '', stderr: '', exitCode: 1 };
			const [ emitter, process ] = executeCliCommand( cliArgs, { silent: true } );

			timeoutId = setTimeout( () => {
				process.kill( 'SIGKILL' );
				resolve( {
					stdout: '',
					stderr: `WP-CLI command timed out after ${ timeout }ms`,
					exitCode: 1,
				} );
			}, timeout );

			emitter.on( 'data', ( { data } ) => {
				const parsed = data as { action?: string; key?: string; value?: string };
				if ( parsed.action === 'keyValuePair' ) {
					if ( parsed.key === 'stdout' ) {
						result.stdout = parsed.value ?? '';
					} else if ( parsed.key === 'stderr' ) {
						result.stderr = parsed.value ?? '';
					} else if ( parsed.key === 'exitCode' ) {
						result.exitCode = parseInt( parsed.value ?? '1', 10 );
					}
				}
			} );

			emitter.on( 'success', () => {
				resolve( result );
			} );

			emitter.on( 'failure', () => {
				resolve( result );
			} );

			emitter.on( 'error', ( { error } ) => {
				resolve( {
					stdout: '',
					stderr: `Error executing WP-CLI command: ${ error.message }`,
					exitCode: 1,
				} );
			} );
		} ).finally( () => {
			clearTimeout( timeoutId );
		} );
```
This is optional, but based on my experience from a previous PR in this project (can't remember exactly which one), putting promise cleanup logic in a `finally` callback is a good way to ensure it always runs, regardless of how the promise resolves.

fredrikekelund (approved):
This is great 👍 I left a couple of comments that would be good to address before landing, but this works as expected and the fundamentals of the implementation look solid to me.

src/ipc-handlers.ts
bcotrim: We also don't want to await the thumbnail update, but need those 2 operations to happen in sequence. I rewrote it using .then() since I believe it makes the intention easier to read.

fredrikekelund:
I struggled with crashing site processes and jumbled output between different WP-CLI child processes. I identified two underlying errors:

1. Studio sends WP-CLI commands relatively quickly after starting a site (presumably the `get-theme-details` command). This would sometimes cause `runWpCliCommand` in `cli/wordpress-server-child.ts` to throw, since `server` wasn’t yet defined. This caused the process to crash and made it unreliable to start the site from Studio.
2. Playground has a limit on the number of concurrent PHP instances (not the same thing as PHP workers, I don’t think). The Assistant executes three separate WP-CLI commands in parallel to gather data about a site (list of installed themes and plugins, as well as the `siteurl`). Executing these commands in such quick succession would also sometimes cause `cli/wordpress-server-child.ts` to crash when Playground threw a `MaxPhpInstancesError`.

I resolved the first error by making `runWpCliCommand` wait for a promise that settles once the server has started. Still, since I couldn't find an easy way to resolve the second issue, I decided to disable child-process WP-CLI execution altogether until further notice. This slows down WP-CLI if the server is already running, but ensures correctness.

fredrikekelund (approved):
LGTM 👍 Let's return to the issue with executing WP-CLI commands in child processes in a follow-up PR

PR #2253, opened 2025-12-15
Attempt to trigger dev builds on `dev/studio-cli-i2`

fredrikekelund:
## Related issues

–

## Proposed Changes

To publish the Studio CLI i2 CfT P2 post, we need a dev build that testers can download. Buildkite only creates dev builds for `trunk` commits, but we are using the `dev/studio-cli-i2` branch for the CLI i2 project. So, this PR attempts to also trigger dev builds on the `dev/studio-cli-i2` branch by tweaking `.buildkite/pipeline.yml`.

I already merged this to the `dev/studio-cli-i2` branch, but that didn't make any difference, so I wanted to try and merge it to `trunk`. It'd be great if someone from Apps infra could take a quick look 🙏

## Testing Instructions

N/A

## Pre-merge Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -164,7 +164,7 @@ steps:
       - step: dev-mac
       - step: dev-windows
     plugins: [$CI_TOOLKIT_PLUGIN, $NVM_PLUGIN]
-    if: build.branch == 'trunk' && build.tag !~ /^v[0-9]+/
+    if: ( build.branch == 'trunk' && build.tag !~ /^v[0-9]+/ ) || build.branch == 'dev/studio-cli-i2'
     notify:
       - github_commit_status:
           context: Distribute Dev Builds


bcotrim (commented):
Could we use something besides the branch name?
If we added a condition, tag for example, we could re-use this in future development branches if needed.

.buildkite/pipeline.yml
```diff
@@ -164,7 +164,7 @@ steps:
       - step: dev-mac
       - step: dev-windows
     plugins: [$CI_TOOLKIT_PLUGIN, $NVM_PLUGIN]
-    if: build.branch == 'trunk' && build.tag !~ /^v[0-9]+/
+    if: ( build.branch == 'trunk' && build.tag !~ /^v[0-9]+/ ) || build.branch == 'dev/studio-cli-i2'
```
iangmaia, line 167: >  already merged this to the dev/studio-cli-i2 branch, but that didn't make any difference, so I wanted to try and merge it to trunk.

iinm there's no need to merge it to `trunk` to have changes in the pipeline to take effect, so they should work on `dev/studio-cli-i2`.
I believe it didn't have the effect you expected (i.e. the job running) due to this:

```yaml
    depends_on:
      - step: dev-mac
      - step: dev-windows
```

That is: this step dependencies **also** have a `if` option (`if: build.tag !~ /^v[0-9]+/` for both of them), so if they don't run, the job that depends on them also won't run. You can try adding the same conditions in both steps and see if that works.

wojtekn:
@fredrikekelund, if you want to use the dev build, what prevents you from triggering the Build for Mac and Build for Windows actions in Buildkite?

Alternatively, if we consider the function stable and are ready for CfT, should we merge it and release 1.7.0-beta1?

fredrikekelund:
I'm dropping this approach. We'll merge to `trunk` instead and distribute a dev build from that commit. Thanks for weighing in, everyone 🙏

PR #2254, opened 2025-12-15
Studio: Ensure the site can't be deleted when syncing

katinthehatsite:
## Related issues

Fixes STU-1127

## Proposed Changes

This PR ensures that the local site can not be deleted when syncing to WP.com (either pulling or pushing).

## Testing Instructions

* Start Studio
* Navigate to the `Sync` tab
* Connect a site
* Start either push or pull
* Navigate to the `Settings` tab
* Click on the three ellipsis in the right corner
* Confirm that the `Delete` site button is disabled and the site can not be deleted
* Try to delete the Studio site from the context menu
* Confirm that the `Delete site` button is disabled

Screenshot 2025-12-15 at 4 22 38 PM

## Pre-merge Checklist


ivan-ottinger (commented):
Thank you for the fix, Kat! I have observed one inconsistency where the `Delete site...` button in the Settings tab is disabled for the whole process, while the similar option in the sidebar dropdown gets disabled only for a small part of the process:

https://github.com/user-attachments/assets/caa8588c-5d36-461c-bbe8-d8241cbc4efb

katinthehatsite:
> Thank you for the fix, Kat! I have observed one inconsistency where the Delete site... button in the Settings tab is disabled for the whole process, while the similar option in the sidebar dropdown gets disabled only for a small part of the process:

👀

katinthehatsite:
Great catch @ivan-ottinger - it should be fixed now. Would you be able to give it another try? Thank you!

src/components/site-menu.tsx
```diff
@@ -133,6 +135,8 @@ function SiteItem( { site }: { site: SiteDetails } ) {
 		tooltipText = __( 'Importing' );
 	} else if ( isPulling ) {
 		tooltipText = __( 'Syncing' );
+	} else if ( isPushing ) {
+		tooltipText = __( 'Syncing' );
```
ivan-ottinger, line 137: Good catch of noticing that label was missing! What if we used just one `else if` and checked for `isSyncing` directly? (instead of having two separate `Syncing` tooltip texts)?

ivan-ottinger (commented):
> Great catch @ivan-ottinger - it should be fixed now. Would you be able to give it another try? Thank you!

Thanks for the fix, Kat. The changes work as expected (bot delete buttons are now disabled in the same way). The code looks generally good to me. I have just shared one suggestion and noticed that tests are failing.

gcsecsey (approved):
Thanks @katinthehatsite for working on this! I tested both pushing and pulling and now I confirm not being able to delete the site for the whole duration, either from the sidebar or from the Settings tab.

https://github.com/user-attachments/assets/dca7cec0-d6eb-40e1-a3b0-607081977c3f

src/components/site-menu.tsx
katinthehatsite: Great suggestion, I applied it in https://github.com/Automattic/studio/pull/2254/changes/6ac53425daa288ca81f704eba60b30bc1c5519d2

ivan-ottinger: Thank you, Kat.

ivan-ottinger (approved):
Thanks for the updates, Kat! All is looking good to me now. ✅

PR #2258, opened 2025-12-16
Update Electron and Vite packages

wojtekn:
## Related issues

- N/A

## Proposed Changes

- I propose updating Electron and Vite packages.

## Testing Instructions

- Confirm builds go through.

## Pre-merge Checklist


.nvmrc
```diff
@@ -1 +1 @@
-22.18.0
+22.20.0
```
epeicher, line 1: Just curious, why this minor version and not the [latest minor](https://nodejs.org/en/download/archive/v22.21.1) for node 22?

epeicher (approved):
Build went through without any issues. I left a question out of curiosity, but changes LGTM! :shipit:

.nvmrc
wojtekn: Good question! I aligned it with the version listed in Electron docs: https://www.electronjs.org/docs/latest/tutorial/electron-timelines

fredrikekelund:
The `electron-vite` upgrade to v5.0.0 comes with breaking changes. We are now using a deprecated API in `electron.vite.config.ts`. See https://electron-vite.org/blog/#electron-vite-5-0-is-out

.nvmrc
epeicher: That makes total sense, thanks for the reply!

wojtekn:
Good catch @fredrikekelund ! I added follow-up: https://github.com/Automattic/studio/pull/2274

PR #2259, opened 2025-12-16
Merge changes from `dev/studio-cli-i2` branch

fredrikekelund:
## Related issues

N/A

## Proposed Changes

We've been using `dev/studio-cli-i2` as a base branch for the Studio CLI i2 project since ~Dec 8. This PR merges all changes from that branch to `trunk`.

## Testing Instructions

All PRs have been tested individually. We simply expect CI to pass for this PR.

## Pre-merge Checklist


fredrikekelund:
I'm working on fixing the E2E tests before we merge this PR

fredrikekelund:
The E2E tests should pass now, so this PR is good to review. cc @bcotrim @nightnei

fredrikekelund:
I've modified `.buildkite/commands/run-e2e-tests.sh` to run only a single test, to help us narrow down why the E2E tests are still failing.

fredrikekelund:
@nightnei and I have concluded that the tests get stuck on this screen:

image

This is the `NoStudioSites` component, but we expected the site to render `Onboarding`. This is the relevant source code:

https://github.com/Automattic/studio/blob/e2a0363871d8bc71f873859ab51c051b4a42ed67/src/components/app.tsx#L51

Now we just need to determine why `needsOnboarding` has an unexpected value when CI runs the E2E tests.

bcotrim:
> @nightnei and I have concluded that the tests get stuck on this screen:
> image
> 
> This is the `NoStudioSites` component, but we expected the site to render `Onboarding`. This is the relevant source code:
> 
> https://github.com/Automattic/studio/blob/e2a0363871d8bc71f873859ab51c051b4a42ed67/src/components/app.tsx#L51
> 
> Now we just need to determine why `needsOnboarding` has an unexpected value when CI runs the E2E tests.

My guess is that the initial site creation failed and as we cleanup the site from appdata it gets back to this step.

nightnei:
@fredrikekelund I was reading codebase and here are my observations - we run `expect(site-content)...` always only after `completeOnboarding`, what means that the flow definitely already went through:
  1.1. `Skip →` - so `needsOnboarding` can't be the culprit
  1.2. Also we had `modal.siteNameInput.fill( customSiteName )` and other steps, but as we see the test is stuck on the screen to choose button "Create a site". So I tested the flow with:
```
diff --git a/src/hooks/use-site-details.tsx b/src/hooks/use-site-details.tsx
index cb2ab8ad..73bd7417 100644
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -291,6 +291,7 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 
                        let newSite: SiteDetails;
                        try {
+                               qqq+www;
                                newSite = await getIpcApi().createSite( path, {
                                        siteName,
                                        wpVersion,
```
I as a result I got error and got the screen with chosing "Create a site" and other buttons.
So my assumption is - maybe we need to install some package or do other configurations for CI? Probably the issue is around `pm2` or some tool/configuration.

nightnei:
@bcotrim yeah, you are compleytelly right, I just wrote the summary of my investagation above :) 
I think that the best next step would be if you take a look at CLI command to create a site and think - what we missed to install/configure for CI? And also - why teh error is muted and we didn't get it in logs of tests. Maybe it was muted in this try/catch (src/hooks/use-site-details.tsx), since we don't throw teh error to teh upper level, however we do console inside the catch...

fredrikekelund:
After much hair-tearing, we finally narrowed down the problem to the fact that Studio and the CLI resolved the `server-files` directory differently in the E2E context. I've fixed the issue and confirmed that the E2E tests _should_ pass now.

nightnei (approved):
❤️

fredrikekelund:
The E2E tests have passed on macOS, so I'll go ahead and merge this

PR #2261, opened 2025-12-16
Re-enable WP-CLI execution in the child process

bcotrim:
## Related issues

- Fixes STU-1138

## Proposed Changes

- **Re-enable WP-CLI execution in child process** (`cli/commands/wp.ts`)
  - Set `IS_WP_CLI_CHILD_PROCESS_EXECUTION_ENABLED = true`
  - WP-CLI commands now use the running site's child process instead of spawning new Playground instances

- **Add WP-CLI command queue** (`cli/wordpress-server-child.ts`)
  - `WpCliCommandQueue` class serializes concurrent WP-CLI commands
  - `MAX_CONCURRENT_WP_CLI_COMMANDS = 3` - limits concurrent `playground.cli()` calls to prevent `MaxPhpInstancesError`
  - `MAX_WP_CLI_QUEUE_SIZE = 10` - prevents unbounded queue growth, rejects new commands when full

- **Fix messageId collision in IPC** (`cli/lib/wordpress-server-manager.ts`)
  - Replace sequential `nextMessageId++` with `generateMessageId()` using timestamp + random
  - Prevents concurrent CLI processes from receiving each other's responses

### Setup
```bash
npm run cli:build
node dist/cli/main.js site create --path /tmp/test-site --name "Test" --skip-browser
```

### Test 1: Concurrent commands return correct results
```bash
# Each command should return its own correct result (not mixed up)
node dist/cli/main.js wp --path /tmp/test-site option get blogname &
node dist/cli/main.js wp --path /tmp/test-site option get siteurl &
node dist/cli/main.js wp --path /tmp/test-site option get admin_email &
node dist/cli/main.js wp --path /tmp/test-site plugin list --format=count &
wait
```
**Expected:** Each returns different, correct values (e.g., "My WordPress Website", "http://localhost:XXXX", "admin@localhost.com", "3")

### Test 2: Site doesn't crash with concurrent commands
```bash
curl -s -o /dev/null -w "Before: %{http_code}" http://localhost:<PORT>
for i in {1..10}; do
  node dist/cli/main.js wp --path /tmp/test-site option get blogname &
done
wait
curl -s -o /dev/null -w "After: %{http_code}" http://localhost:<PORT>
```
**Expected:** Site returns HTTP 200 both before and after (no crash)

### Test 3: Failed commands don't crash the server
```bash
node dist/cli/main.js wp --path /tmp/test-site nonexistent-command
curl -s -o /dev/null -w "Site status: %{http_code}" http://localhost:<PORT>
```
**Expected:** Command shows error message, site still returns HTTP 200

### Test 4: Queue rejects commands when full
```bash
# Send 15 slow commands (3 active + 10 queue = 13 max, so 2 should fail)
for i in {1..15}; do
  node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1 &
done
wait
```
**Expected:** 13 commands succeed, 2 commands fail with "WP-CLI command queue is full (10 pending commands). Please try again later."

### Cleanup
```bash
node dist/cli/main.js site stop --path /tmp/test-site
rm -rf /tmp/test-site
```

## Pre-merge Checklist

### DIFF
--- a/cli/lib/wordpress-server-manager.ts
+++ b/cli/lib/wordpress-server-manager.ts
@@ -155,7 +155,13 @@ async function waitForReadyMessage( pmId: number ): Promise< void > {
  * - Checks periodically for inactivity
  * - Has both inactivity timeout and max total timeout
  */
-let nextMessageId = 0;
+/**
+ * Generate a unique message ID that won't collide across concurrent CLI processes.
+ */
+function generateMessageId(): number {
+	return Date.now() * 1000 + Math.floor( Math.random() * 1000 );
+}
+
 const messageActivityTrackers = new Map<
 	number,
 	{
@@ -175,7 +181,7 @@ async function sendMessage(
 ): Promise< unknown > {
 	const { maxTotalElapsedTime = PLAYGROUND_CLI_MAX_TIMEOUT, logger } = options;
 	const bus = await getPm2Bus();
-	const messageId = nextMessageId++;
+	const messageId = generateMessageId();
 	let responseHandler: ( packet: unknown ) => void;
 
 	return new Promise( ( resolve, reject ) => {
--- a/cli/lib/wordpress-server-manager.ts
+++ b/cli/lib/wordpress-server-manager.ts
@@ -229,7 +229,10 @@ async function sendMessage(
 			} else if ( validPacket.raw.topic === 'console-message' ) {
 				lastActivityTimestamp = Date.now();
 				logger?.reportProgress( validPacket.raw.message );
-			} else if ( validPacket.raw.topic === 'error' ) {
+			} else if (
+				validPacket.raw.topic === 'error' &&
+				validPacket.raw.originalMessageId === messageId
+			) {
 				const error = new Error( validPacket.raw.errorMessage ) as Error & {
 					cliArgs?: Record< string, unknown >;
 				};
--- a/cli/lib/wordpress-server-manager.ts
+++ b/cli/lib/wordpress-server-manager.ts
@@ -155,15 +155,8 @@ async function waitForReadyMessage( pmId: number ): Promise< void > {
  * - Checks periodically for inactivity
  * - Has both inactivity timeout and max total timeout
  */
-/**
- * Generate a unique message ID that won't collide across concurrent CLI processes.
- */
-function generateMessageId(): number {
-	return Date.now() * 1000 + Math.floor( Math.random() * 1000 );
-}
-
 const messageActivityTrackers = new Map<
-	number,
+	string,
 	{
 		activityCheckIntervalId: NodeJS.Timeout;
 	}
@@ -181,7 +174,7 @@ async function sendMessage(
 ): Promise< unknown > {
 	const { maxTotalElapsedTime = PLAYGROUND_CLI_MAX_TIMEOUT, logger } = options;
 	const bus = await getPm2Bus();
-	const messageId = generateMessageId();
+	const messageId = crypto.randomUUID();
 	let responseHandler: ( packet: unknown ) => void;
 
 	return new Promise( ( resolve, reject ) => {
--- a/cli/wordpress-server-child.ts
+++ b/cli/wordpress-server-child.ts
@@ -70,7 +70,7 @@ class WpCliCommandQueue {
 
 		return new Promise( ( resolve, reject ) => {
 			this.queue.push( { args, resolve, reject } );
-			this.processNext();
+			void this.processNext();
 		} );
 	}
 
@@ -89,7 +89,7 @@ class WpCliCommandQueue {
 			item.reject( error instanceof Error ? error : new Error( String( error ) ) );
 		} finally {
 			this.activeCount--;
-			this.processNext();
+			void this.processNext();
 		}
 	}
 }
--- a/cli/wordpress-server-child.ts
+++ b/cli/wordpress-server-child.ts
@@ -25,6 +25,7 @@ import { sanitizeRunCLIArgs } from 'common/lib/cli-args-sanitizer';
 import { isWordPressDirectory } from 'common/lib/fs-utils';
 import { getMuPlugins } from 'common/lib/mu-plugins';
 import { formatPlaygroundCliMessage } from 'common/lib/playground-cli-messages';
+import { sequential } from 'common/lib/sequential';
 import { isWordPressDevVersion } from 'common/lib/wordpress-version-utils';
 import { z } from 'zod';
 import { getSqliteCommandPath, getWpCliPharPath } from 'cli/lib/server-files';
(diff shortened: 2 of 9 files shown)


cli/wordpress-server-child.ts
```diff
@@ -308,6 +362,20 @@ async function runWpCliCommand(
...
+async function runWpCliCommand( args: string[] ): Promise< WpCliResult > {
+	await Promise.allSettled( [ startingPromise ] );
+
+	if ( ! server ) {
+		throw new Error( `Failed to run WP CLI command because server is not running` );
+	}
+
+	return wpCliQueue.enqueue( args );
+}
+
```
fredrikekelund: I would argue we should either inline this in `WpCliCommandQueue::processNext`, or make it a separate method on `WpCliCommandQueue`

cli/wordpress-server-child.ts
```diff
@@ -285,19 +343,15 @@ async function runBlueprint( config: ServerConfig ): Promise< void > {
 	}
 }
 
-async function runWpCliCommand(
-	args: string[]
-): Promise< { stdout: string; stderr: string; exitCode: number } > {
```
fredrikekelund, line 290:
```ts
const runWpCliCommand = sequential( 3, async ( args: string[] ): Promise< WpCliResult > => {
```

fredrikekelund (commented):
Seeing your solution in this PR, I realized that we already have a `sequential` helper function in `common/lib/sequential.ts` that can accomplish this a little more succinctly. If we apply the following diff, implement my inline suggestion and restore the function body of `runWpCliCommand`, I believe we should be good.

```diff
diff --git a/common/lib/sequential.ts b/common/lib/sequential.ts
index 593ee246..d7698147 100644
--- a/common/lib/sequential.ts
+++ b/common/lib/sequential.ts
@@ -2,6 +2,7 @@ const sequentialLocks = new Map< () => Promise< unknown >, Set< Promise< unknown
 
 // Ensures that calls to the provided function are executed sequentially
 export function sequential< Args extends unknown[], Return >(
+       concurrentCount: number,
        fn: ( ...args: Args ) => Promise< Return >
 ) {
        return async ( ...args: Args ) => {
@@ -10,7 +11,7 @@ export function sequential< Args extends unknown[], Return >(
                        sequentialLocks.set( fn, locks );
                }
 
-               while ( locks.size ) {
+               while ( locks.size >= concurrentCount ) {
                        await Promise.allSettled( [ ...locks ] );
                }
 
```

Let me know your thoughts, @bcotrim

cli/lib/wordpress-server-manager.ts
```diff
@@ -155,7 +155,13 @@ async function waitForReadyMessage( pmId: number ): Promise< void > {
...
  * - Has both inactivity timeout and max total timeout
  */
-let nextMessageId = 0;
+/**
+ * Generate a unique message ID that won't collide across concurrent CLI processes.
+ */
+function generateMessageId(): number {
+	return Date.now() * 1000 + Math.floor( Math.random() * 1000 );
+}
+
```
fredrikekelund: Oh man, good catch 👍 This is definitely part of the reason I was seeing jumbled output when running multiple WP-CLI commands in quick succession.

The current implementation is fine, but I guess we could also use `crypto.randomUUID()`. Any thoughts, @bcotrim?

cli/wordpress-server-child.ts
```diff
@@ -38,6 +39,11 @@ let server: RunCLIServer | null = null;
 let startingPromise: Promise< void > | null = null;
 let lastCliArgs: Record< string, unknown > | null = null;
 
+/**
+ * Result type for WP-CLI command execution
+ */
+type WpCliResult = { stdout: string; stderr: string; exitCode: number };
+
```
fredrikekelund: This is no big deal, but we could easily keep this inlined in the return type for `runWpCliCommand`, IMO. Since we're not reusing it, I mean.

common/lib/sequential.ts
```diff
@@ -1,23 +1,39 @@
...
-			await Promise.allSettled( [ ...locks ] );
+			queueCount++;
+			try {
+				while ( locks.size >= concurrentCount ) {
+					await Promise.allSettled( [ ...locks ] );
+				}
+			} finally {
+				queueCount--;
+			}
 		}
```
fredrikekelund:
```ts
		if ( maxQueueSize !== undefined && queueCount >= maxQueueSize ) {
			throw new Error(
				`Queue is full (${ maxQueueSize } pending commands). Please try again later.`
			);
		}

		while ( locks.size >= concurrentCount ) {
			queueCount++;
			await Promise.allSettled( [ ...locks ] );
			queueCount--;
		}
```
This is my suggestion ⬆️

A couple of comments:
1. The initial `if ( locks.size >= concurrentCount )` doesn't really make sense to me. We should check the queue count before checking the lock count. 
2. There's no need to wrap `Promise.allSettled` in a `try..catch`, since it never rejects (unlike `Promise.all`). 
3. By moving the `queueCount` increment and decrement operations inside the `while` loop, we ensure that the `queueCount` variable isn't touched if the lock count doesn't exceed `concurrentCount`. This doesn't _really_ matter, though, since we don't run any asynchronous operations if this is case. Consider this an optional nit that maybe just looks cleaner 🙂

common/lib/sequential.ts
```diff
@@ -1,23 +1,39 @@
...
 
 		const fnPromise = fn( ...args );
+		locks.add( fnPromise );
 
 		try {
-			locks.add( fnPromise );
```
fredrikekelund: This is probably fine, but I guess the `locks.add` call was previously part of the `try` clause to clarify its relationship to the `finally` clause. I don't believe this change makes any functional difference, though, so feel free to implement it whichever way you prefer, @bcotrim

common/lib/sequential.ts
```diff
@@ -1,23 +1,39 @@
...
 
-// Ensures that calls to the provided function are executed sequentially
 export function sequential< Args extends unknown[], Return >(
-	fn: ( ...args: Args ) => Promise< Return >
+	fn: ( ...args: Args ) => Promise< Return >,
+	options?: SequentialOptions
 ) {
+	const concurrentCount = options?.concurrent ?? 1;
+	const maxQueueSize = options?.max;
+	const locks = new Set< Promise< unknown > >();
```
fredrikekelund:
```suggestion
	const locks = new Set< Promise< Return > >();
```
Good idea moving `locks` inside the closure 👍 It definitely simplifies things. We might as well use the correct return type now, though.

fredrikekelund (approved):
The "Queue rejects commands when full" test case didn't work as expected for me (with or without my proposed changes to `common/lib/sequential.ts`). This definitely fixes the core issue, though. I've confirmed that I can run WP-CLI commands from the Assistant without any issues.

Great work identifying the `messageId` issue, too 👍

bcotrim:
> The "Queue rejects commands when full" test case didn't work as expected for me (with or without my proposed changes to `common/lib/sequential.ts`). This definitely fixes the core issue, though. I've confirmed that I can run WP-CLI commands from the Assistant without any issues.
> 
> Great work identifying the `messageId` issue, too 👍

```bash
❯ for i in {1..15}; do
  node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1 &
done
wait
[2] 41596
[3] 41597
[4] 41598
[5] 41599
[6] 41600
[7] 41601
[8] 41602
[9] 41603
[10] 41604
[11] 41605
[12] 41606
[13] 41607
[14] 41608
[15] 41609
[16] 41610
✖ Failed to run WP-CLI command: Queue is full (10 pending commands). Please try again later.
✖ Failed to run WP-CLI command: Queue is full (10 pending commands). Please try again later.
[14]    exit 1     node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[2]    exit 1     node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[5]    done       node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[9]    done       node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[12]    done       node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[4]    done       node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[16]  + done       node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[8]    done       node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[6]    done       node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[15]  + done       node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[7]    done       node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
✖ Failed to run WP-CLI command: Timeout waiting for response to message 2d2c6a3a-9e83-4114-8bcb-21adbdc8cef9: No activity for 120s
✖ Failed to run WP-CLI command: Timeout waiting for response to message ab1da4ca-bb4e-4971-8f6e-87b0a85c9564: No activity for 120s
✖ Failed to run WP-CLI command: Timeout waiting for response to message 6b933194-c256-4118-bc08-8f4ec6708500: No activity for 120s
✖ Failed to run WP-CLI command: Timeout waiting for response to message 0c5c871d-dff6-4150-9ee9-fda246e3d71c: No activity for 120s
[11]  - exit 1     node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[3]    exit 1     node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[10]  - exit 1     node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
[13]  + exit 1     node dist/cli/main.js wp --path /tmp/test-site eval "sleep(30);" 2>&1
```

Were you referencing the timeouts? I feel like that is expected, we could not timeout something while it is in queue, but it might be overthinking at this point.
What do you think?

fredrikekelund:
No, I meant that I didn't get any timeout errors. I didn't get any Playground errors about exceeding the maximum number of request handlers either, though. 

I say let's land this PR and revisit this later if needed 👍

fredrikekelund:
Let's change the base branch to `dev/studio-cli-i2` and merge this.

PR #2262, opened 2025-12-16
Fix WP auto-updates not being enforced when enabled in Studio

wojtekn:
## Related issues

- Fixes STU-1139

## Proposed Changes

Studio only disabled auto-updates when the setting was off, but didn't actively enable them when on. This meant WordPress sites with auto-updates already disabled in their database wouldn't respect Studio's enable setting.

This PR ensures Studio's auto-update setting is the source of truth by explicitly enabling auto-updates when `isWpAutoUpdating` is true.

## Testing Instructions

1. Create a site using WordPress 6.8
2. Open WP Admin
3. Confirm the site uses 6.8.3
4. Navigate to Updates and confirm that auto updates are disabled:

> This site will not receive automatic updates for new versions of WordPress

5. Open Studio site in Terminal
6. Disable auto-updates in WP using option:

```
wp option set auto_update_core_major unset
```

8. Stop Studio
9. Update site in appdata-v1.json file and set isWpAutoUpdating to true
10. Start Studio
11. Open WP Admin
12. Navigate to Updates and confirm that auto updates are enabled

## Pre-merge Checklist

### DIFF
--- a/common/lib/mu-plugins.ts
+++ b/common/lib/mu-plugins.ts
@@ -245,7 +245,7 @@ function getStandardMuPlugins( options: MuPluginOptions ): MuPlugin[] {
 		`,
 	} );
 
-	// Disable auto-updates if configured
+	// Configure auto-updates based on Studio settings
 	if ( ! options.isWpAutoUpdating ) {
 		muPlugins.push( {
 			filename: '0-disable-auto-updates.php',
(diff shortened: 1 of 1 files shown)


common/lib/mu-plugins.ts
```diff
@@ -256,6 +256,16 @@ function getStandardMuPlugins( options: MuPluginOptions ): MuPlugin[] {
 			add_filter( 'allow_major_auto_core_updates', '__return_false' );
 			`,
 		} );
+	} else {
+		muPlugins.push( {
+			filename: '0-enable-auto-updates.php',
+			content: `<?php
+			// Enable auto-updates
+			add_filter( 'allow_dev_auto_core_updates', '__return_true' );
```
sejas, line 264: I don't think this will make sites with a stable version automatically update to betas, but we should double-check.

sejas (approved):
I confirm it works as expected and after running `wp option set auto_update_core_major unset` I see the updates are still updated and the source of truth is Studio `isWpAutoUpdating` value.

Image

gavande1 (approved):
I tested this and works as described. LGTM 👍 

Image

common/lib/mu-plugins.ts
gavande1: I think this filter will only run when you are on dev version. So it won't be automatically update stable versions to dev versions. 
https://github.com/WordPress/wordpress-develop/blob/3a5da6c220aa50e1d928ec77257189ed353e4d0b/src/wp-admin/includes/class-core-upgrader.php#L346-L361

PR #2264, opened 2025-12-16
Studio: Make sure that Publish site button navigates to WP.com

katinthehatsite:
## Related issues

Fixes STU-1108

## Proposed Changes

This PR ensures that when you click on the `Publish site` button in Studio, it navigates to WP.com and takes you through the purchase flow, then bringing back to Studio. It should behave identical to `Create a new WordPress.com site` button inside the `Sync` modal.

## Testing Instructions

* Start Studio
* Navigate to the `Sync` tab
* Complete the purchase
* Confirm that the site you just created is connected in the `Sync` tab (might be a couple of seconds)

## Pre-merge Checklist

### DIFF
--- a/src/components/publish-site-button.tsx
+++ b/src/components/publish-site-button.tsx
@@ -1,8 +1,8 @@
 import { cloudUpload } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
 import { useCallback } from 'react';
+import { useSyncSites } from 'src/hooks/sync-sites';
 import { useAuth } from 'src/hooks/use-auth';
-import { useOffline } from 'src/hooks/use-offline';
 import { useSiteDetails } from 'src/hooks/use-site-details';
 import { generateCheckoutUrl } from 'src/lib/generate-checkout-url';
 import { getIpcApi } from 'src/lib/get-ipc-api';
@@ -11,13 +11,14 @@ import { useGetConnectedSitesForLocalSiteQuery } from 'src/stores/sync/connected
 
 export const PublishSiteButton = () => {
 	const { __ } = useI18n();
-	const isOffline = useOffline();
 	const { user } = useAuth();
 	const { selectedSite } = useSiteDetails();
 	const { data: connectedSites = [] } = useGetConnectedSitesForLocalSiteQuery( {
 		localSiteId: selectedSite?.id,
 		userId: user?.id,
 	} );
+	const { isAnySitePulling, isAnySitePushing } = useSyncSites();
+	const isAnySiteSyncing = isAnySitePulling || isAnySitePushing;
 
 	const handlePublishClick = useCallback( () => {
 		getIpcApi().openURL( generateCheckoutUrl( selectedSite ?? undefined ) );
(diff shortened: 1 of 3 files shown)


epeicher (approved):
Thanks @katinthehatsite for improving this! I have tested it, and it works as expected. I am redirected to the creation site flow, and the site is connected to Studio at the end of the process. Changes also LGTM!

|After clicking on Publish Site | Creating site|Site is created | Site is connected to Studio|
|-------|------|-------|------|
| Image|Image|Image|Image|

This PR is working as expected and could be merged. However, I would like to point out that the button says `Publish site`, and that is not actually happening until they click on `Push` in Studio. So I think we should create some follow-up task to add a clear Notice to the user that the site must be _pushed_ for the local site to be published.

katinthehatsite:
> This PR is working as expected and could be merged. However, I would like to point out that the button says Publish site, and that is not actually happening until they click on Push in Studio. So I think we should create some follow-up task to add a clear Notice to the user that the site must be pushed for the local site to be published.

Thanks for the review 👍 

@epeicher 100%, we will do the follow up PR as mentioned in Linear to do the following:

> In the follow up, we will make the Push to Productionmodal open once the user is redirected back to Studio. Alternatively, we might start the push automatically

src/modules/sync/components/create-button.tsx
```diff
@@ -1,11 +1,11 @@
 import { __ } from '@wordpress/i18n';
-import { DEFAULT_CUSTOM_DOMAIN_SUFFIX } from 'common/constants';
 import { ArrowIcon } from 'src/components/arrow-icon';
 import Button, { ButtonVariant } from 'src/components/button';
 import offlineIcon from 'src/components/offline-icon';
 import { Tooltip } from 'src/components/tooltip';
 import { useOffline } from 'src/hooks/use-offline';
 import { cx } from 'src/lib/cx';
+import { generateCheckoutUrl } from 'src/lib/generate-checkout-url';
```
sejas, line 8: Nice optimization!

src/components/publish-site-button.tsx
```diff
@@ -26,12 +21,8 @@ export const PublishSiteButton = () => {
 	const isAnySiteSyncing = isAnySitePulling || isAnySitePushing;
 
 	const handlePublishClick = useCallback( () => {
-		if ( ! user ) {
-			authenticate();
-		}
-		dispatch( connectedSitesActions.openModal( 'push' ) );
-		setSelectedTab( 'sync' );
-	}, [ user, setSelectedTab, dispatch, authenticate ] );
+		getIpcApi().openURL( generateCheckoutUrl( selectedSite ?? undefined ) );
```
sejas, line 24: Can we make `handlePublishClick` do nothing if selectedSite is nullish? Perhaps the button should be disabled or even don't appear at all in that edge case.

sejas (commented):
Thanks for pointing the Publish button to open the browser directly. It works great.
I left a small suggestion to make sure that we call `generateCheckoutUrl` always with a local site id.

src/lib/generate-checkout-url.ts
```diff
@@ -0,0 +1,20 @@
+import { DEFAULT_CUSTOM_DOMAIN_SUFFIX } from 'common/constants';
+
+export function generateCheckoutUrl( selectedSite?: SiteDetails ): string {
+	const url = new URL(
+		'https://wordpress.com/setup/new-hosted-site?ref=studio&section=studio-sync&showDomainStep'
```
sejas, line 5: @katinthehatsite, can we pass the `section` as a parameter? Currently it's hardcoded to `&section=studio-sync`, but I would like the Publish site button to have a different section in the URL so we can measure the amount of "clicks" on this button.

katinthehatsite: Sure, I will adress these comments in a follow up PR 👍

src/components/publish-site-button.tsx
katinthehatsite: I added a follow up here: *https://github.com/Automattic/studio/pull/2283

src/lib/generate-checkout-url.ts
katinthehatsite: Adressed in https://github.com/Automattic/studio/pull/2283

PR #2266, opened 2025-12-16
Add GH action to notify about stale PRs

wojtekn:
## Related issues

- Fixes STU-1134

## Proposed Changes

- I propose adding GH action to notify about stale PRs.

## Testing Instructions

- Code review should suffice, and we can test when merged to trunk. See how the test message looks on Slack: p1765903484005839-slack-C056D54B98U

## Pre-merge Checklist

### DIFF
--- a/.github/workflows/stale-pr-notifications.yml
+++ b/.github/workflows/stale-pr-notifications.yml
@@ -1,0 +1,90 @@
+name: Stale PR Notifications
+
+on:
+  schedule:
+    # Run daily at 9:00 AM UTC (adjust timezone as needed)
+    - cron: '0 9 * * *'
+  workflow_dispatch: # Allow manual triggering for testing
+    inputs:
+      channel_id:
+        description: 'Slack channel ID (leave empty to use default)'
+        required: false
+        type: string
+
+permissions:
+  pull-requests: read
+  contents: read
+
+jobs:
+  notify-stale-prs:
+    name: Notify team about stale PRs
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: '.nvmrc'
+
+      - name: Fetch stale PRs
+        id: fetch-prs
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Fetch open PRs not updated in the last 24 hours
+          # Exclude draft PRs and PRs that are already approved
+          STALE_PRS=$(gh pr list \
+            --repo ${{ github.repository }} \
+            --state open \
+            --limit 100 \
+            --json number,title,url,updatedAt,author,isDraft,reviewDecision,labels \
+            --jq '[.[] | select(
+              (.updatedAt | fromdateiso8601) < (now - 86400) and
+              .isDraft == false and
+              (.reviewDecision != "APPROVED")
+            )]')
+
+          # Count stale PRs
+          COUNT=$(echo "$STALE_PRS" | jq 'length')
+          echo "count=$COUNT" >> $GITHUB_OUTPUT
+
+          # Save to file for next step
+          echo "$STALE_PRS" > /tmp/stale-prs.json
+
+          echo "Found $COUNT stale PR(s) needing review"
+
+      - name: Format Slack message
+        id: format-message
+        if: steps.fetch-prs.outputs.count > 0
+        run: node scripts/format-stale-pr-message.js
+
+      - name: Send Slack notification
+        if: steps.fetch-prs.outputs.count > 0
+        env:
+          SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+          # Use manual input if provided, otherwise use secret, otherwise use default channel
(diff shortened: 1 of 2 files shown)


.github/workflows/stale-pr-notifications.yml
```diff
@@ -0,0 +1,90 @@
...
+        id: format-message
+        if: steps.fetch-prs.outputs.count > 0
+        run: node scripts/format-stale-pr-message.js
+
+      - name: Send Slack notification
+        if: steps.fetch-prs.outputs.count > 0
+        env:
+          SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+          # Use manual input if provided, otherwise use secret, otherwise use default channel
+          SLACK_CHANNEL_ID: ${{ github.event.inputs.channel_id || secrets.SLACK_CHANNEL_ID || 'C056D54B98U' }}
```
wojtekn, line 84: When merged, I will change the channel to Studio.

.github/workflows/stale-pr-notifications.yml
```diff
@@ -0,0 +1,90 @@
...
+        with:
+          node-version-file: '.nvmrc'
+
+      - name: Fetch stale PRs
+        id: fetch-prs
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Fetch open PRs not updated in the last 24 hours
+          # Exclude draft PRs and PRs that are already approved
```
katinthehatsite, line 53: Would we consider a PR that has some comments but is not approved also stale?

katinthehatsite (commented):
This looks good to me, I had one question about what we would consider to be stale PRs

.github/workflows/stale-pr-notifications.yml
gavande1: I believe those PRs would be captured in the "not updated in the last 24 hours" rule. If not, we should consider them stale, but only after 24 hours. Additionally, all open PRs (not approved) will become stale on Monday, so we need a method to manage that situation.

wojtekn: @gavande1 I think it's okay to have such a reminder on Monday, also, it could motivate us to finish PRs before the weekend. 😊

@katinthehatsite's comment makes the PR updated, so PRs with comments in the last 24 hours wouldn't be caught.

wojtekn: @gavande1 thinking more about your concern about weekends, I added 56afec40efcf7f93d2a866639368ab576a7a56f8. What do you think?

scripts/format-stale-pr-message.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Format PR status
+function getStatusEmoji( pr ) {
+	if ( pr.reviewDecision === 'REVIEW_REQUIRED' ) {
+		return '▶️';
+	} else if ( pr.reviewDecision === 'CHANGES_REQUESTED' ) {
+		return '⏩️️';
+	} else {
+		return '⏸️';
```
sejas, line 42: Have we considered using custom slack emojis?

```suggestion
		return ':pr-open:';
	} else if ( pr.reviewDecision === 'CHANGES_REQUESTED' ) {
		return ':github-changes-requested:';
	} else {
		return '⏸️';
```
For example: 

Image
Image

.github/workflows/stale-pr-notifications.yml
```diff
@@ -0,0 +1,104 @@
...
+              \"icon_emoji\": \":github:\",
+              \"text\": $(echo "$MESSAGE" | jq -r '.text'),
+              \"attachments\": $(echo "$MESSAGE" | jq -c '.attachments')
+            }"
+
+          echo "✅ Sent notification about ${{ steps.fetch-prs.outputs.count }} stale PR(s) to Slack channel $SLACK_CHANNEL_ID"
+
+      - name: No stale PRs
+        if: steps.fetch-prs.outputs.count == 0
+        run: echo "✨ No stale PRs found - all PRs are up to date!"
```
sejas, line 104: Nice! This should prevent any message from being sent if no PRs meet the criteria.

sejas (approved):
The action seems very helpful. Thanks for creating it.
I suggested changing the emojis to those that are more meaningful.

scripts/format-stale-pr-message.js
wojtekn: I tried using red/yellow/green icons, but the message seemed too aggressive.

.github/workflows/stale-pr-notifications.yml
gavande1: Thanks for addressing the concern. It looks good to me.

katinthehatsite: > @katinthehatsite's comment makes the PR updated, so PRs with comments in the last 24 hours wouldn't be caught.

Sounds good, I was just thinking that maybe we could skip PRs that have comments because it means that someone is already looking even though it might be longer than 24 hours but this approach is also good 👍

scripts/format-stale-pr-message.js
katinthehatsite: I think the current emojis are fine 👍 I like that they are not too forward which makes you feel like it is a normal part of your workflow instead of something urgent

katinthehatsite (approved):

.github/workflows/stale-pr-notifications.yml
wojtekn: Thanks for explaining that. If someone was looking, posted a comment, and then nothing happened for 24 hours, the PR is stale.

PR #2271, opened 2025-12-17
Studio: Add UI for when local tree for push has errors

katinthehatsite:
## Related issues

Fixes STU-966

## Proposed Changes

This PR adds a node to handle errors when there is an issue retrieving nodes from the local tree:

Screenshot 2025-12-17 at 10 25 18 AM

## Testing Instructions

* Apply the following diff: 

```diff --git a/src/stores/sync/sync-hooks.ts b/src/stores/sync/sync-hooks.ts
index 31db0fc9..b538c66b 100644
--- a/src/stores/sync/sync-hooks.ts
+++ b/src/stores/sync/sync-hooks.ts
@@ -83,6 +83,9 @@ export function useLocalFileTree() {
                        setIsLoading( true );
                        setError( null );
                        try {
+                               // TEMPORARY: Simulate error for testing
+                               throw new Error( 'Simulated error for testing error handling' );
+                               
                                const rawNodes = await getIpcApi().listLocalFileTree( siteId, path, 3 );
                                return convertRawToTreeNodes( rawNodes );
                        } catch ( err ) {
```

* Ensure that you have at least one WP.com site connected to Studio
* Navigate to the `Sync` tab
* Click on `Push`
* Select `Specific files and folders` option
* Confirm that when you open `wp-content`, you see the error shown on the screenshot above

## Pre-merge Checklist


src/modules/sync/components/sync-dialog.tsx
```diff
@@ -355,6 +374,19 @@ export function SyncDialog( {
...
 										return null;
 									} }
+									renderEmptyContent={ ( nodeId ) => {
+										if ( nodeId === 'wp-content' && type === 'push' && localFileTreeError ) {
+											return (
+												<div className="text-red-600 italic flex items-center gap-1.5">
+													<Icon icon={ cautionFilled } size={ 20 } className="fill-red-600" />
+													{ __(
+														'Error retrieving files and directories. Please close and reopen this dialog to try again.'
+													) }
```
katinthehatsite, line 377: I did not add an explicit retry because this seems to be rather an edge case but we can add it if we consider it necessary

gavande1 (approved):
I have tested this and works great. LGTM 👍 

Image

katinthehatsite:
Update: I updated the style to look like this based on the comment from Linear

Screenshot 2025-12-17 at 2 28 51 PM

gcsecsey (approved):
Thanks for adding the error message and for iterating on the styling @katinthehatsite!

Changes LGTM and work well:

Image

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -118,13 +117,21 @@ const useDynamicTreeState = (
 		localSiteId,
 	] );
 
+	// Handle local file tree errors by clearing children to show custom error message
+	useEffect( () => {
+		if ( type === 'push' && localFileTreeError ) {
+			setTreeState( ( treeState ) => updateNodeById( treeState, 'wp-content', { children: [] } ) );
+		}
+	}, [ type, localFileTreeError, setTreeState ] );
```
sejas, line 125: I'm not sure if this useEffect is necessary or it could be just an if when rendering the node.children.

src/components/tree-view.tsx
```diff
@@ -164,9 +167,13 @@ const TreeItem = ( {
...
 					{ node.children.length === 0 ? (
-						<div className="text-gray-500 italic" aria-label={ __( 'Empty folder' ) }>
-							{ __( 'Empty' ) }
-						</div>
+						renderEmptyContent && renderEmptyContent( node.id ) ? (
+							renderEmptyContent( node.id )
+						) : (
+							<div className="text-gray-500 italic" aria-label={ __( 'Empty folder' ) }>
+								{ __( 'Empty' ) }
+							</div>
```
sejas, line 175: I would suggest renaming `renderEmptyContent` to `renderErrorContent`. Or in case we want to keep it tied to empty children, then I think we should move the `Empty` logic into that function too. The condition `renderEmptyContent && renderEmptyContent( node.id ) ?` seems a bit tricky.

sejas (commented):
Thanks for adding the error case. Great work.
I left a couple of suggestions to consider. We could refactor `renderEmptyContent` and the `useEffect`.

src/modules/sync/components/sync-dialog.tsx
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2352

src/components/tree-view.tsx
katinthehatsite: I made adjustments in https://github.com/Automattic/studio/pull/2352 for this

PR #2277, opened 2025-12-17
Fix Featured Blueprints card grid styling

ivan-ottinger:
## Related issues

- Fixes STU-1150

## Proposed Changes

- fix classname (which was likely changed in a recent WordPress Components update)

| Before | After |
|--------|--------|
| CleanShot 2025-12-17 at 16 11 24@2x |CleanShot 2025-12-17 at 16 11 15@2x | 

## Testing Instructions

2. Navigate to _Add site → Start from a Blueprint_.
3. The featured blueprints should be formatted correctly.

## Pre-merge Checklist

### DIFF
--- a/src/modules/add-site/components/blueprints.css
+++ b/src/modules/add-site/components/blueprints.css
@@ -23,7 +23,7 @@
 	@apply absolute top-1.5 left-1.5 text-sm text-white rounded-sm px-1.5 py-0.5;
 }
 
-.blueprints-container .dataviews-view-grid {
+.blueprints-container .dataviews-view-grid__row {
 	@apply !grid !grid-cols-3 !gap-5 !items-start;
 }
 


gcsecsey (approved):
Thanks @ivan-ottinger for identifying and quickly fixing this! 🙌 

I could reproduce the issue on trunk, and I can confirm that the fix makes sense and works great! 🚢 

I also checked that we're not using the `dataviews-view` class anywhere else within the repo, so the change most likely affected only this page.

| trunk | branch |
|--------|--------|
| Image | Image |

epeicher (approved):
Thanks for fixing this nasty bug @ivan-ottinger 🙌 ! I have reproduced it in `trunk` and confirmed it is fixed as part of these changes. LGTM! :shipit: 

|trunk | This PR|
|-------|------|
|Image|Image|

PR #2278, opened 2025-12-17
Studio CLI: Fix miscellaneous issues raised in CfT

fredrikekelund:
## Related issues

- Fixes STU-1135
- Fixes STU-1146
- Fixes STU-1147
- Fixes STU-1145
- Fixes STU-1161

## Proposed Changes

- Refactor `SiteServer` to construct the `server` member (a `CliServerProcess` instance) in the constructor. It adds no overhead, so there's no point in setting and unsetting it at different times. 
- Parallelize WP-CLI operations during export to mitigate WP-CLI performance overhead.
- Refactor `SiteServer::delete` to call the `studio site delete` command under the hood
- Fail gracefully if user isn't logged in when running `studio site delete`, or if deleting a preview site fails
- Remove the `version` field from `cli/package.json`, and read the version number from the root `package.json`
- Only log `site.adminPassword` after starting or creating site if defined
- Re-enable the `showHelpOnFail` behavior

## Testing Instructions

1. Ensure that creating, starting, and stopping a site works as expected
2. Ensure that deleting a site from Studio works as expected (deleting files and NOT deleting files)
3. Create a preview site for one of your sites
4. Log out in Studio
5. Delete this site using the CLI
6. Ensure that it works as expected

## Pre-merge Checklist


src/lib/import-export/export/exporters/default-exporter.ts
fredrikekelund, line 1: Parallelize WP-CLI operations during export to mitigate WP-CLI performance overhead.

bcotrim (approved):
Tested and works as expected

LGTM 👍 

Wanted to note a different behavior:
Delete now completely removes files from the computer, while before they were being sent to the trash.
I don't have any strong preferences regarding this behavior, but since it's changed I wanted to confirm it was intentional or that we are aware of it.

fredrikekelund:
> Delete now completely removes files from the computer, while before they were being sent to the trash

Thanks for raising this. This was a conscious decision that I justified by thinking it would save us from introducing a new dependency. The current behavior is also not documented.

Coming back to it now, I feel it's probably wiser to accept the additional dependency and avoid the pain for users who are already familiar with the current behavior, or who end up in a situation where they want to restore a site they deleted.

fredrikekelund:
I've added the following final fixes:
- Move files to trash instead of removing them from disk when deleting a site (if the appropriate flag is set)
- Remove the "Finished running the Blueprint" string to avoid confusion where users might think that the entire setup process is finished (see pdtkmj-4fD-p2#comment-8352)
- Replace all instances of triple dots ("...") with proper ellipses ("…")

fredrikekelund:
@bcotrim, I'd appreciate a quick final look at the file trash logic 🙏

bcotrim (approved):
Thanks for addressing the changes!
LGTM 👍

PR #2281, opened 2025-12-17
Remove PHP-WASM dependencies from Studio

bcotrim:
## Related issues

- Fixes STU-960
- Fixes STU-693

## Proposed Changes

Remove PHP-WASM packages (~1.1 GB) from the desktop app's ASAR bundle. The desktop app now uses CLI for all WordPress operations, making these dependencies redundant.

- Create local type definitions (`common/types/php-versions.ts`) to replace `@php-wasm/universal` imports
- Remove WordPress provider abstraction layer (`src/lib/wordpress-provider/`)
- Remove child process entry points from desktop build
- Remove `getProviderConstants` IPC/Redux layer - components import directly from `common/constants`
- Move download utilities from wp-now to `src/lib/download-utils.ts`
- Remove `vendor/wp-now/` directory
- Move PHP-WASM packages from dependencies to devDependencies (for type checking only)
- Keep `forge.config.ts` ignore patterns to exclude PHP-WASM from bundle

## Testing Instructions

1. `npm install`
2. `npm start`
3. Create a new site - verify it works
4. Start/stop sites - verify server operations work
5. Import/export a site - verify backup/restore works
6. Run tests: `npm test`
7. Build installer: `npm run package` - verify app.asar is ~295 MB

## Pre-merge Checklist


fredrikekelund:
Let's remove the `unzipper` dependency from `package.json` as part of this PR. https://github.com/Automattic/studio/pull/2348 removed it from `scripts/download-wp-server-files.ts`, and the only remaining place it was used was in `vendor/wp-now` (which we'll remove with this PR).

epeicher (commented):
Thanks @bcotrim for this huge reduction in size 🙌

I have done some initial testing, and I have identified that creating a site does not start it for me. Also, when deleting it, an error is displayed, although the deletion completes. Please see a video below. 

https://github.com/user-attachments/assets/3a1fcb82-8244-4926-88b3-204fb3fc08ec


Please let me know if you need the logs.

bcotrim:
@epeicher
I couldn't replicate the issue, do you see anything in the Studio logs?
Can you share a bit more detail about your test? (`npm start` or bundled Studio from `npm run package`, etc)
Thanks testing and sharing the results!

epeicher:
> @epeicher I couldn't replicate the issue, do you see anything in the Studio logs? Can you share a bit more detail about your test? (`npm start` or bundled Studio from `npm run package`, etc) Thanks testing and sharing the results!

@bcotrim, I have been investigating more, and I have done the following that fix my issues:
1. Delete `node_modules` and `cli/node_modules`, I have also deleted the `dist` folder
2. Run `npm install` on the root folder to install npm packages
3. Run `cd cli && npm install` to install npm packages on the cli folder
4. On the root folder, run `npm start`

That fixes all the issues I was having, I am going to do some additional testing, but this looks good so far 🥳

epeicher (commented):
Thanks again @bcotrim for these massive changes. I have done a round of testing, and these are my findings. 

1. Create a new site - verify it works - ✅
This works fine, I have identified a minor issue that I am not able to reproduce in `trunk` but it is not a blocker, and it could be tackled as a follow-up if required. When creating a site, there is a moment where the list of sites displays the new one duplicated, please see screenshot: Image
3. Start/stop sites - verify server operations work - ✅
4. Export a site - verify backup works - ✅
5. Import a site - verify restore works - ❌
Importing a recently exported backup fails  with the error `Database import failed` - Please find error logs here 3a0e9-pb
5. Run tests: `npm test` -  ✅
6. Build installer: `npm run package` - verify app.asar is ~295 MB - ✅ - `294M Jan  9 13:15 app.asar` (in trunk it occupies 1.4G 🤯 )

I have measured the time it takes to create a site (approximately), and I have observed that on my Mac, it takes around 16 seconds on `trunk` (using `npm start`and having a few sites)  and it takes around 8 seconds on this branch, so this is a great improvement 🙌 

All in all, I would like to have another pair of eyes to review the changes, but this is looking great, amazing work.

bcotrim:
@fredrikekelund @epeicher with most (all?) of other on-going changes to the `dev/studio-cli-i2` I updated this PR and it should be ready for another review.

fredrikekelund:
https://github.com/Automattic/studio/pull/2404 is already huge, so I recommend we land that PR first and then merge this PR directly to `trunk`. 

I'm reviewing this PR now 👍

common/lib/extract-zip.ts
fredrikekelund, line 1: Could we use this function in `scripts/download-node-binary.mjs`, too? 

In the future, we might consider also targeting `src/lib/import-export/import/handlers/backup-handler-zip.ts`, but that's for another PR.

cli/commands/site/set.ts
fredrikekelund, line 1: 👍

common/types/blueprint.ts
fredrikekelund, line 1: Why can't we import these directly from `@wp-playground/blueprints`?

scripts/download-node-binary.ts
fredrikekelund, line 1: Can we use the `extractZip` function from `common` here?

src/lib/server-files-paths.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Get the base path for server files (WordPress versions, SQLite, etc.)
+ */
+function getBasePath(): string {
+	if ( process.env.NODE_ENV === 'test' ) {
+		return getTmpPath( 'server-files' );
+	}
+	return getServerFilesPath();
+}
```
fredrikekelund, line 28: The big reason for having this extra layer on top of `src/storage/paths` is that we add `getTmpPath`, right? How necessary is that?

src/lib/wordpress-server-types.ts
```diff
@@ -0,0 +1,20 @@
...
+		documentRoot: string;
+		request?: ( options: {
+			url: string;
+			method: string;
+			body: Record< string, string >;
+		} ) => Promise< { text: string } >;
+	};
+	start( siteId?: string ): Promise< void >;
+	stop(): Promise< void >;
+	runPhp( data: { code: string; [ key: string ]: unknown } ): Promise< string >;
```
fredrikekelund:
```suggestion
```
Can't we just remove the `runPhp` method now?

src/lib/wordpress-server-types.ts
fredrikekelund, line 1: Given that we now have just a single class implementing this interface, I'm wondering if we really need it..?

src/lib/wordpress-setup.ts
fredrikekelund, line 1: Looking forward to the day when we've consolidated this type of functionality in the CLI…

src/setup-wp-server-files.ts
```diff
@@ -131,10 +132,32 @@ export async function setupWPServerFiles() {
...
+				resolve( version ? `v${ version }` : '' );
+			} else {
+				resolve( '' );
+			}
+		} );
+
+		emitter.on( 'failure', () => resolve( '' ) );
+		emitter.on( 'error', () => resolve( '' ) );
+	} );
+}
```
fredrikekelund, line 152: This won't work. The `--path` argument is parsed by the Studio CLI and must point to a site directory. 

Off the top of my head, the best way to resolve this might be to add a hidden `--no-path` option to the `wp` command. Thoughts, @bcotrim?

fredrikekelund (commented):
Overall, this LGTM 👍 Nice work, @bcotrim!

The one thing we should look into before merging is the function for checking the current WP-CLI version. We'll need a different approach there.

wojtekn:
@bcotrim "Size reduction:" is a bit misleading as it compares cleanup with the cli-i2 branch.

1.6.8 is ~1.8 GB, so there is no reduction when compared to base brach.

fredrikekelund:
> there is no reduction when compared to base brach.

This was our original goal. The fact that we've added a new bundled node runtime since @bcotrim opened this PR will impact the installer size, and we expect it to be slightly larger than 1.6.8. 

But yeah, even if the PR description might lead one to believe that this change reduces the installer filesize, we are really just keeping it in check 👍

wojtekn:
@fredrikekelund, it's okay. I just wanted to clarify this to ensure the reader doesn't think this cleanup provides size reduction for the app.

fredrikekelund (approved):
LGTM 👍 There are still a couple of comments from my previous review that I'd love to discuss, but they're no blockers. 

Let's move ahead with landing this

fredrikekelund:
For the record, the reason I made the `--studio-no-path` change in 78015d0 was that TS complained locally about overriding the `path` option type (despite CI passing for some unclear reason 🤔)

bcotrim:
> @bcotrim "Size reduction:" is a bit misleading as it compares cleanup with the cli-i2 branch.
> 
> 1.6.8 is ~1.8 GB, so there is no reduction when compared to base brach.

Thanks for pointing it out! I Removed it, since it no longer makes sense

PR #2282, opened 2025-12-18
Ensure all CLI commands can be aborted

fredrikekelund:
## Related issues

- Fixes STU-1140

## Proposed Changes

- Add an `AbortController` in `cli/lib/wordpress-server-manager.ts` that interrupts `sendMessage` promises, ensuring that the CLI process exits when it receives a `SIGINT` or `SIGTERM` (<kbd>Ctrl+C</kbd>)
- Add an `abort` IPC message topic that aborts async operations in `cli/lib/types/wordpress-server-ipc.ts` (most oftenly killing the process). We'll refine this behavior once #2261 has landed.

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js site create --path PATH_TO_SITE` where `PATH_TO_SITE` is the path to a new folder where the site should be created
3. Once the process reaches the `Starting WordPress server` stage, press <kbd>Ctrl+C</kbd>
4. Ensure that the process is aborted

## Pre-merge Checklist

### DIFF
--- a/cli/lib/pm2-manager.ts
+++ b/cli/lib/pm2-manager.ts
@@ -60,9 +60,6 @@ export function disconnect(): void {
 	}
 }
 
-process.on( 'SIGINT', disconnect );
-process.on( 'SIGTERM', disconnect );
-
 // Cache the return value of `pm2.list` for a very short time to make multiple calls in quick
 // succession more efficient
 const listProcesses = cacheFunctionTTL( () => {
--- a/cli/lib/pm2-manager.ts
+++ b/cli/lib/pm2-manager.ts
@@ -60,7 +60,8 @@ export function disconnect(): void {
 	}
 }
 
-process.on( 'exit', disconnect );
+process.on( 'disconnect', disconnect );
+process.on( 'SIGHUP', disconnect );
 process.on( 'SIGINT', disconnect );
 process.on( 'SIGTERM', disconnect );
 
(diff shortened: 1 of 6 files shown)


cli/lib/pm2-manager.ts
```diff
@@ -60,10 +60,6 @@ export function disconnect(): void {
 	}
 }
 
-process.on( 'exit', disconnect );
-process.on( 'SIGINT', disconnect );
-process.on( 'SIGTERM', disconnect );
-
```
fredrikekelund, line 66: This isn't necessary. API consumers should be responsible for connecting and disconnecting from pm2. In fact, this previously resulted in a race condition where the `abort` message I implemented in this PR wouldn't be delivered (because the process disconnected from the pm2 daemon before sending it).

cli/lib/types/wordpress-server-ipc.ts
```diff
@@ -39,6 +44,7 @@ const managerMessageRunBlueprint = z.object( {
 
 const managerMessageStopServer = z.object( {
 	topic: z.literal( 'stop-server' ),
+	data: z.object( {} ),
```
fredrikekelund, line 47: pm2 requires all messages to have `id`, `topic`, and `data` fields. We didn't notice that the `stop-server` topic would generate errors because `stopWordPressServer` swallows exceptions.

bcotrim (approved):
LGTM as works as described 👍 

Image

PR #2288, opened 2025-12-18
Studio: Make the Publish button open "Push" dialog after coming from WP.com

katinthehatsite:
## Related issues

Fixes STU-1108

## Proposed Changes

This PR ensures that when you click on `Publish site` button and come back to WP.com, the WP.com site is not only connected but that we also open a `Push` modal.

## Testing Instructions

* This needs to be tested with https://github.com/Automattic/wp-calypso/pull/107779 which needs to be applied in Calypso first
* Start Studio
* Navigate to the `Sync` tab
* Complete the purchase
* Confirm that the site you just created is connected in the `Sync` tab (might be a couple of seconds)
* Confirm that the `Push` modal is open
* Test the same steps with `Create a new WordPress.com site` from the `Sync` modal and confirm that once you connect the site, the `Push` modal does not auto-open

## Pre-merge Checklist


epeicher (commented):
Thanks @katinthehatsite for implementing this! I reviewed [the related PR](https://github.com/Automattic/wp-calypso/pull/107779) in Calypso repo, so I will copy my comments and the screenshots here for completeness:

This nicely wraps up the process of publishing a local site to WordPress.com 🙌. I have tested it, and it works as expected. After purchasing the site, it is created as Atomic, and eventually, Studio connects the site and opens the Push modal. 

|Studio click on Publish site|Redirects to WP.com | Open in Studio|Push modal opens |Push completes| Site is published in WP.com
|---------|-------|------|------|------|------|
|Image| Image|Image|Image|Image|Image|

I have also tested the additional step of:
>Test the same steps with Create a new WordPress.com site from the Sync modal and confirm that once you connect the site, the Push modal does not auto-open

and it works as expected 👌, but I was wondering if that flow should be similar to the `Publish site` flow? 🤔 Anyways, that can be discussed as a possible follow-up. 

I see one Unit Test failing, but after addressing that, this can be progressed. LGTM! :shipit:

katinthehatsite:
> but I was wondering if that flow should be similar to the Publish site flow? 🤔 Anyways, that can be discussed as a possible follow-up.

I am not sure 😅 This was @sejas request to make the difference between these two buttons - `Publish site` should auto-open and the `Create a new site` should behave as is without changes. We can adjust it if needed, I don't have any strong preferences.

sejas:
> I am not sure 😅 This was @sejas request to make the difference between these two buttons - `Publish site` should auto-open and the `Create a new site` should behave as is without changes. We can adjust it if needed, I don't have any strong preferences.

Yep! I suggested keeping the same behaviour for the `Connect site` flow. The main reason is that connecting a site can be used for Push or Pull direction, so we can auto-open any modal. Correct me if I'm wrong. 🤔 
We are talking about this button-link:

Screenshot 2025-12-19 at 14 38 09

epeicher:
> Yep! I suggested keeping the same behaviour for the Connect site flow. The main reason is that connecting a site can be used for Push or Pull direction, so we can auto-open any modal. Correct me if I'm wrong. 🤔
We are talking about this button-link:

@sejas, yes, that's the one I meant. My reasoning here was that if users are creating a site, they probably want to push their local changes, but I agree they can:
- Create a site
- Customize the new created site on WordPress.com
- Pull their changes and continue working locally

So I agree opening the `Push` modal limits one possible scenario unnecessarily. I agree with leaving that as is. Thanks for the explanation!

sejas:
> > Yep! I suggested keeping the same behaviour for the Connect site flow. The main reason is that connecting a site can be used for Push or Pull direction, so we can auto-open any modal. Correct me if I'm wrong. 🤔
> > We are talking about this button-link:
> 
> @sejas, yes, that's the one I meant. My reasoning here was that if users are creating a site, they probably want to push their local changes, but I agree they can:
> 
> * Create a site
> * Customize the new created site on WordPress.com
> * Pull their changes and continue working locally
> 
> So I agree opening the `Push` modal limits one possible scenario unnecessarily. I agree with leaving that as is. Thanks for the explanation!

You make a valid point. Newly created sites will usually be empty, so the chances that the user wants to push rather than pull are quite high. I think it's worth discussing whether it's better to open the push modal further.

sejas (approved):
I tested it and it works as expected. It would be great to speed up the actions if possible, like connecting and opening the modal. It could be improved in a follow-up issue.

I'm approving the PR, although there are some conflicts and unit tests that need to be fixed before merging.

https://github.com/user-attachments/assets/cbbae869-d958-4dad-8280-3decaea9e791

katinthehatsite:
> It would be great to speed up the actions if possible, like connecting and opening the modal. It could be improved in a follow-up issue. 

Sounds good, I will open a follow up issue.

katinthehatsite:
I opened the follow up issue for speed improvement here: STU-1173

PR #2299, opened 2025-12-19
Studio CLI: Fix Windows E2E tests

fredrikekelund:
## Related issues

- Fixes STU-1155

## Proposed Changes

This PR fixes the Windows E2E tests by making the following changes:

- Disconnect from pm2 in the `site list --watch` command in response to `SIGINT` and `SIGTERM` signals. Without this, the `site list --watch` command couldn't be aborted with <kbd>Ctrl+C</kbd> and it would interfere with closing the Electron app.
- Revert the changes from #2313, because they make the E2E tests fail. See https://github.com/Automattic/studio/pull/2313#issuecomment-3715154268
- Modify the `site stop --all` command to kill all pm2 processes in one go, **including the daemon itself**.
- Refactor `E2ESession::cleanup` to not call `ElectronApplication::close()`, but always use `electronApp.evaluate( ( { app } ) => app.quit() )`.
- Wait for the `stopAllServersOnQuit` child process to finish before exiting the Electron app. This ensures **all** child processes have ended once the app quits, allowing auto-updates to be gracefully applied (if the bundled Node.js runtime were still used to run child processes, it would not be updatable on Windows).
- Kill child processes in response to `will-quit` Electron event.
- Clean up E2E session files after a slight delay, and use `rimraf` for smart retry behavior.
- Playground was logging warnings about `themeZipFile` and `pluginZipFile` being deprecated, so I also took the opportunity to replace those keys with `themeData` and `pluginData` in the test fixtures. 
- Start the site watcher (`site list --watch`) and wait for the `spawn` event from the child process before creating the main app window. After working on this for so long, I doubt this makes any difference, but it still seems like good practice. 

I also made changes related to general stability, not explicitly to the E2E tests:

- Added a `pm2-connection.lock` lockfile to ensure that multiple `site start` commands triggered in quick sequence are still able to properly connect to the pm2 daemon. Previously, if multiple sites had an `autoStart` config, only one site would start when Studio started. The other sites would stall indefinitely.
- Force all pm2 processes (including the daemon) to quit if there's a pending update. This allows the auto-updater to overwrite files that might otherwise be locked on Windows (since they're currently in use).
- Change the pm2 patch to clarify that the `pm2.disconnect` method takes a callback, and make all `disconnect` calls async. 

## Testing Instructions

CI should pass. 

## Pre-merge Checklist


cli/commands/site/list.ts
```diff
@@ -126,6 +126,8 @@ export async function runCommand( format: 'table' | 'json', watch: boolean ): Pr
 				},
 				{ debounceMs: 500 }
 			);
+
+			process.on( 'SIGINT', disconnect );
```
fredrikekelund, line 141: The `studio site list --watch` command would previously not interrupt in response to <kbd>Ctrl+C</kbd>

fredrikekelund:
Finally, finally, the E2E tests pass reliably here 😮‍💨 Except they don't… After I merged the latest changes from the base branch, they started failing again. However, this time, it's because something in the app is actually broken. 

I've narrowed it down to a regression coming from https://github.com/Automattic/studio/pull/2360 that makes it so the `Site Editor` link et al in the Overview tab open incorrect URLs after a site has been created (`localhost:0`). I'm currently diagnosing the problem, and I've also pinged @bcotrim for help.

Aside from this issue, there were multiple problems that led to test failures on the `dev/studio-cli-i2` branch. I've kept the PR description up-to-date to outline those.

fredrikekelund:
OK, after one more round of changes, the tests are now consistently green, and I've slimmed down the diff as much as possible.

cli/commands/site/start.ts
```diff
@@ -12,14 +12,14 @@ const logger = new Logger< LoggerAction >();
...
-		logger.reportSuccess( __( 'Site loaded' ) );
-
 		logger.reportStart( LoggerAction.START_DAEMON, __( 'Starting process daemon…' ) );
 		await connect();
 		logger.reportSuccess( __( 'Process daemon started' ) );
 
+		logger.reportStart( LoggerAction.LOAD_SITES, __( 'Loading site…' ) );
+		const site = await getSiteByFolder( sitePath );
+		logger.reportSuccess( __( 'Site loaded' ) );
+
```
fredrikekelund, line 22: I moved the `connect` call to the top of this function to fix a bug where triggering many `site start` commands in quick succession would cause only one site to start (the others would hang indefinitely).

bcotrim (approved):
LGTM 👍 
Amazing work not only on making E2E tests pass but ensuring the solution is robust and solid.

PR #2303, opened 2025-12-22
Studio CLI: Add bundled Node binary and fix Windows popup

bcotrim:
## Related issues

- Fixes STU-1148

## Proposed Changes

1. **Add bundled Node.js binary** - The CLI now uses a bundled Node.js binary instead of relying on the Electron run as Node.

2. **Fix Windows terminal popup** - Patched the `ps-man` package to add `windowsHide: true` to spawn calls. This prevents a brief console window popup (showing `C:\WINDOWS\system32\tasklist`) when running CLI commands on Windows.

### Windows
2. Run `node dist/cli/main.js site create --name=test-site`
3. Verify no popup windows appear during site creation
4. Verify the site is created successfully
5. Create a site using Studio and confirm operation times are similar

### macOS
2. Run `node dist/cli/main.js site create --name=test-site`
3. Verify the site is created successfully
4. Create a site using Studio and confirm operation times are similar

## Pre-merge Checklist


scripts/download-node-binary.sh
fredrikekelund, line 1: To make this more portable, let's do it in JS instead.

scripts/download-node-binary.js
```diff
@@ -0,0 +1,189 @@
...
+const { execSync } = require( 'child_process' );
+const os = require( 'os' );
+
+const LTS_FALLBACK = 'v22.12.0';
+
+function getNodeVersion() {
+	const nvmrcPath = path.join( __dirname, '..', '.nvmrc' );
+	if ( fs.existsSync( nvmrcPath ) ) {
+		const version = fs.readFileSync( nvmrcPath, 'utf-8' ).trim();
+		// Ensure version starts with 'v'
```
nightnei: The code looks self-explanatory
```suggestion
```

src/storage/paths.ts
```diff
@@ -66,6 +66,19 @@ export function getCliPath(): string {
 		: path.join( getResourcesPath(), 'cli', 'main.js' );
 }
 
+export function getBundledNodeBinaryPath(): string {
+	const nodeBinaryName = process.platform === 'win32' ? 'node.exe' : 'node';
+
+	if ( process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test' ) {
```
nightnei: I am thinking - should we keep it only for tests? To have it used in development, to work with the same as in prod, to avoid if in prod something will be unexpectedly broken?
```suggestion
	if ( process.env.NODE_ENV === 'test' ) {
```

nightnei (approved):
The code looks great and works as expected:
1. I don't see popup on windows anymore.
2. But, I encounter x2 difference in time between CLI and Studio, is it expected?

macOS:
CLI: 6 sec
Studio: 14 sec

Windows:
CLI: 1:40
Studio: 2:50

src/storage/paths.ts
fredrikekelund: The only time we can guarantee that the bundled node binary is available is in the production build. I flipped the logic in this function to return the bundled path only when `NODE_ENV === 'production'`

fredrikekelund (approved):
Great initiative! I'm very optimistic about the performance impact of this change.

I made a couple of changes to this PR:
- Refined `scripts/download-node-binary.mjs` in 17e2ead
- Updated `./bin` scripts to use the bundled node binary in 4e77d33 (meaning Electron isn't involved when users run `studio` from the terminal, either)
- Use the system-level node binary when forking CLI processes during development in f0b9873 (this addresses the performance problems @nightnei noted in https://github.com/Automattic/studio/pull/2303#pullrequestreview-3611541580)

With those changes in place, this PR should be good to land 👍

gavande1:
> But, I encounter x2 difference in time between CLI and Studio, is it expected?

@bcotrim @fredrikekelund I am curious, Did we addressed this issue?

fredrikekelund:
We did! The difference @nightnei noticed came from the fact that we previously used the Electron node runtime to fork CLI processes. In f0b9873, I made it so the system-level node binary is used instead, which means performance will be the same as when running the CLI directly, and we more closely match the production build behavior (which is to use the newly bundled node runtime).

scripts/download-node-binary.mjs
```diff
@@ -0,0 +1,166 @@
...
+
+const LTS_FALLBACK = 'v22.12.0';
+
+function getNodeVersion() {
+	const nvmrcPath = path.join( import.meta.dirname, '..', '.nvmrc' );
+	if ( fs.existsSync( nvmrcPath ) ) {
+		const version = fs.readFileSync( nvmrcPath, 'utf-8' ).trim();
+		return version.startsWith( 'v' ) ? version : `v${ version }`;
+	}
+	console.log( `.nvmrc not found, using fallback version ${ LTS_FALLBACK }` );
```
wojtekn, line 22: Would it be safer to remove fallback and fail the job/build if there is no version in `.nvmrc`?

bcotrim: We can even consider have a separate version for the bundled Node.
Keeping the Node version in sync with the recommended version from Playground CLI should be the best approach. cc @fredrikekelund

PR #2304, opened 2025-12-22
CLI: Update `studio-cli.bat` so we can read from stdin

fredrikekelund:
## Related issues

- Fixes STU-1162

## Proposed Changes

I've made the following changes to improve the Studio CLI on Windows:

- Remove the `call` statements preceding `ELECTRON_EXECUTABLE` and `node`. These were likely added as an analogue to `exec` in the `bin/studio-cli.sh`, but they are not necessary. In fact, they even break stdin parsing (see Linear issue for more details).
- Capture the exit code from the CLI script and pass it to the `exit` command. 
- I've added `enabledelayedexpansion`. This is a more defensive approach, as it makes it so variables are expanded at runtime rather than at parse time. 
- More quotes, to account for any spaces in paths. 

## Testing Instructions

1. Download the Windows build from pdtkmj-4fD-p2#testing
2. Install it on your Windows workstation
3. Open `C:\Users\fredrik\AppData\Local\studio\app-1.6.7-dev20\resources\bin\studio-cli.bat` in a text editor
4. Paste the contents of `bin/studio-cli.bat` from this PR into that file
5. Open a new Command Prompt window
6. Run `studio auth logout`
7. Run `studio auth login`
8. Ensure that you can complete the entire login flow successfully

## Pre-merge Checklist


gcsecsey (commented):
I took a dig at testing this, but I couldn't reproduce the original authentication issue on the `dev/studio-cli-i2` branch. When running the `auth login` command, I could paste the token as expected on both branches, in both PowerShell and Command Prompt.

| PowerShell | Terminal |
|--------|--------|
| Image | Image |

fredrikekelund:
@gcsecsey, this issue happens only when running the production CLI, which uses `bin/studio-cli.bat` as the actual executable "binary" that we make available on the `$PATH`. 

Testing this change isn't exactly simple, but I've updated the testing instructions so they're (hopefully) easy to follow. The PR is ready for review now.

epeicher (approved):
Thanks @fredrikekelund! I have tested it following the test instructions; I have reproduced the error with the build version and confirmed this PR fixes the issue. Changes also LGTM! :shipit: 

|Build version | Using the `bat` file from this PR|
|-------|------|
|Image|Image|

PR #2310, opened 2025-12-23
Migrate Performance Metrics to Buildkite pipeline

wojtekn:
## Related issues

- Fixes STU-978

## Proposed Changes

- I propose migrating Performance Metrics tests from GitHub action to Buildkite.

## Testing Instructions

- Confirm that the build goes through
- Confirm there is a performance metrics comment under the PR

## Pre-merge Checklist


.buildkite/commands/run-metrics-tests.sh
```diff
@@ -0,0 +1,44 @@
...
+
+  echo "--- :bar_chart: Logging metrics to CodeVitals"
+  COMMITTED_AT=$(git show -s $BUILDKITE_COMMIT --format="%cI")
+  npm run log-to-codevitals -- $CODEVITALS_AUTH_TOKEN trunk $BUILDKITE_COMMIT $BASELINE_COMMIT $COMMITTED_AT
+  cd -
+else
+  # Other branches - skip metrics
+  echo "--- :information_source: Skipping metrics for non-trunk branch"
+  exit 0
+fi
```
mokagio, line 44: Have you considered running this check at the start of the script? This way, if the "skip" branch of the conditional should run, we can avoid installing dependencies and the rest of the setup and save some time.

.buildkite/pipeline.yml
```diff
@@ -60,6 +60,20 @@ steps:
       - github_commit_status:
           context: E2E Tests
 
+  - label: ":chart_with_upwards_trend: Performance Metrics"
+    key: metrics
+    agents:
+      queue: mac
+    command: bash .buildkite/commands/run-metrics-tests.sh
```
mokagio, line 67: Nitpick: I think calling `bash` is redundant here because the script already has `!#/usr/bin/env bash`. Most of the `.sh` scripts we use as the `command` steps in other projects are invoked that way.


```suggestion
    command: .buildkite/commands/run-metrics-tests.sh
```

---

Update: After looking at the pipeline while working on #2327, I noticed that calling the script via `bash` is the standard in the pipeline, so it'd be out of place to change it in this PR.

.buildkite/commands/run-metrics-tests.sh
mokagio: Took a stab in https://github.com/Automattic/studio/pull/2327

mokagio (approved):

.buildkite/commands/run-metrics-tests.sh
wojtekn: Would it make sense to run this [condition as part of pipeline logic](https://buildkite.com/docs/pipelines/configure/conditionals), as we do in other steps? For example:

```
if: build.branch == 'trunk' || build.pull_request.base_branch == 'trunk'
```

.buildkite/pipeline.yml
wojtekn: Thanks for confirming that.

.buildkite/commands/run-metrics-tests.sh
mokagio: Sounds good @wojtekn . 

I see https://github.com/Automattic/studio/pull/2310/commits/716b93a67e4da643e5458874928c0778bb70b7e0 implements this and also adds the additional condition of the PR having to be ready for review. Neat.

_Ship it!_

PR #2311, opened 2025-12-23
Fix appearance Query Manager plugin in wp-admin

nightnei:
- Fixes STU-1005

## Proposed Changes
Previously, Query Manager plugin didn't work with Sites that are run under Studio. The reason is - `php_sapi_name()` return `cli`. Since it's essential to be "cli" in playground, that's why we needed to go with soem workaround. It would be great to go with some more specific fix as checkoing kinds "is playground-cli", but just for consistency I decided to go with the same fix as applyied for [playground](https://github.com/WordPress/wordpress-playground/pull/2415). It's a workaround, but actually it's robust.

## Testing Instructions
1. Create Studio site
2. Open wp-admin
3. Install "Query Manager" plugin and activate it
4. Assert that you see it in admin bar (before it wasn't rendered for sites in Studio)<br />
Screenshot 2025-12-23 at 15 14 23
5. Open "Installed plugins" page
6. Assert that you see all buttons
<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2025-12-23 at 15 15 58
<td>Screenshot 2025-12-23 at 15 16 08
</tr>
</table>

### DIFF
--- a/common/lib/mu-plugins.ts
+++ b/common/lib/mu-plugins.ts
@@ -78,6 +78,16 @@ async function createLoaderMuPlugin(): Promise< string > {
 function getStandardMuPlugins( options: MuPluginOptions ): MuPlugin[] {
 	const muPlugins: MuPlugin[] = [];
 
+	muPlugins.push( {
+		filename: '0-tmp-fix-qm-plugin-sapi.php',
+		content: `<?php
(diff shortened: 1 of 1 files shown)


wojtekn (approved):
The fix looks correct. It is good that we have the same workaround applied in Playground and Studio.

The plugin loads its UI now:

Image

.

PR #2334, opened 2025-12-30
Sync: Fix renderer crash when network disconnects after upload

gavande1:
## Related issues

- Fixes STU-1180

## Proposed Changes

- Added error handling in `getPushProgressInfo` to catch network errors gracefully
- Skip Sentry reporting for expected `crossDomain` network errors (offline scenarios)
- Prevent unhandled promise rejections when network goes offline

## Testing Instructions

1. Start a new sync operation
2. Allow the upload step to complete successfully
3. Disconnect the network (e.g., turn off Wi‑Fi) during subsequent sync steps
4. Verify the renderer no longer crashes and handles the offline state gracefully
5. Verify that when the network reconnects, the sync operation can continue

## Pre-merge Checklist

### DIFF
--- a/src/hooks/sync-sites/use-sync-push.ts
+++ b/src/hooks/sync-sites/use-sync-push.ts
@@ -151,11 +151,8 @@ export function useSyncPush( {
 					}
 				);
 			} catch ( error ) {
-				// Network errors (e.g., offline) should be handled gracefully
-				// The error will be handled when the user comes back online
-				console.error( 'Failed to get push progress:', error );
-
-				// Skip Sentry reporting for expected network errors (crossDomain errors)
+				// Skip Sentry reporting for expected network errors (crossDomain errors). The client throws this error
+				// when the user is offline.
 				if (
 					error instanceof Error &&
 					'crossDomain' in error &&
--- a/src/hooks/sync-sites/use-sync-push.ts
+++ b/src/hooks/sync-sites/use-sync-push.ts
@@ -142,12 +142,31 @@ export function useSyncPush( {
 				return;
 			}
 
-			const response = await client.req.get< ImportResponse >(
-				`/sites/${ remoteSiteId }/studio-app/sync/import`,
-				{
-					apiNamespace: 'wpcom/v2',
+			let response: ImportResponse;
+			try {
+				response = await client.req.get< ImportResponse >(
+					`/sites/${ remoteSiteId }/studio-app/sync/import`,
+					{
+						apiNamespace: 'wpcom/v2',
+					}
+				);
+			} catch ( error ) {
+				// Network errors (e.g., offline) should be handled gracefully
+				// The error will be handled when the user comes back online
+				console.error( 'Failed to get push progress:', error );
+
+				// Skip Sentry reporting for expected network errors (crossDomain errors)
+				if (
+					error instanceof Error &&
+					'crossDomain' in error &&
+					( error as Error & { crossDomain?: boolean } ).crossDomain
+				) {
+					return;
 				}
-			);
+
+				Sentry.captureException( error );
+				return;
+			}
 
 			let status: PushStateProgressInfo = pushStatesProgressInfo.creatingRemoteBackup;
 			if ( response.success && response.status === 'finished' ) {


src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -142,12 +142,31 @@ export function useSyncPush( {
...
+			let response: ImportResponse;
+			try {
+				response = await client.req.get< ImportResponse >(
+					`/sites/${ remoteSiteId }/studio-app/sync/import`,
+					{
+						apiNamespace: 'wpcom/v2',
+					}
+				);
+			} catch ( error ) {
+				// Network errors (e.g., offline) should be handled gracefully
```
katinthehatsite: > Network errors (e.g., offline) should be handled gracefully

I am not sure we need to include this comment though. It seems clear from the code itself. What do you think?

src/hooks/sync-sites/use-sync-push.ts
```diff
@@ -142,12 +142,31 @@ export function useSyncPush( {
...
+				// The error will be handled when the user comes back online
+				console.error( 'Failed to get push progress:', error );
```
katinthehatsite: What is the process of throwing the console.error in this case? What use would it have for the user?

katinthehatsite (approved):
It seems it works as described but when I was testing this, the sync progress did not indicate to me in any way that I was offline. Perhaps we could consider improving this area to be more indicative that the user is disconnected:

Image

What do you think?

katinthehatsite (commented):
It seems that the GH approved the changes but I was not intending to approve them just yet, sorry about that

src/hooks/sync-sites/use-sync-push.ts
gavande1: That is a good point. It makes sense to remove it, as the code itself is quite explanatory.

src/hooks/sync-sites/use-sync-push.ts
gavande1: AFAIU, the import status is pooled at intervals, but it continues to do so even when the network goes offline. On the trunk, the renderer crashes due to the network being offline. This is the expected behavior; however, the API itself might return an error from the server. I thought it would be useful to log this for debugging purposes, especially to handle any unknown cases.

gavande1:
> It seems it works as described but when I was testing this, the sync progress did not indicate to me in any way that I was offline. Perhaps we could consider improving this area to be more indicative that the user is disconnected:

Excellent observation. I added the offline indicator in #2212, but it only appears during the upload phase and when the network goes offline. I will create a follow-up issue to address this.

gavande1:
@katinthehatsite I have created STU-1195 linear issue to improve offline statuses during different phases of push.

katinthehatsite:
> @katinthehatsite I have created [STU-1195](https://linear.app/a8c/issue/STU-1195) linear issue to improve offline statuses during different phases of push.

Nice sounds good to me 👍

src/hooks/sync-sites/use-sync-push.ts
gavande1: Removed in [a1359a0](https://github.com/Automattic/studio/pull/2334/commits/a1359a0501e1c4cee5fe71ddc7f9d11b8fe4d9a0)

src/hooks/sync-sites/use-sync-push.ts
gavande1: Removed in [a1359a0](https://github.com/Automattic/studio/pull/2334/commits/a1359a0501e1c4cee5fe71ddc7f9d11b8fe4d9a0)

src/hooks/sync-sites/use-sync-push.ts
katinthehatsite: I see, thanks for explaining!

I am wondering if we should just skip logging it to the console to not pollute the console with something that we know will cause the error e.g. network interruption. What do you think?

Not a strong preference, just a suggestion on my end.

gavande1: Thanks for the suggestion Kate. I have removed logging after you suggestion.

gavande1:
> It seems that the GH approved the changes but I was not intending to approve them just yet, sorry about that

@katinthehatsite Do you mind giving another review?

katinthehatsite (approved):
Looks good on my end 👍 Thanks for implementing the adjustments!

PR #2336, opened 2025-12-30
Site Creation: Fix renderer crash for long site names when creating new site

gavande1:
## Related issues

- Fixes STU-1182

## Proposed Changes

- Added error handling in `handleSiteNameChange` to catch ENAMETOOLONG errors and display a user-friendly error message when site name is too long
- Added tests to verify ENAMETOOLONG error handling works correctly
- The renderer now gracefully handles long site names instead of crashing

## Testing Instructions

1. Open the "Add site" modal
2. Enter a very long site name (e.g., a string of 300+ characters)
3. Verify that instead of crashing, a user-friendly error message is displayed: "The site name is too long. Please choose a shorter site name."
4. Run the test suite: `npm test -- src/hooks/tests/use-add-site.test.tsx`
5. Verify all tests pass, including the new ENAMETOOLONG error handling tests

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-add-site.ts
+++ b/src/hooks/use-add-site.ts
@@ -237,7 +237,7 @@ export function useAddSite( options: UseAddSiteOptions = {} ) {
 		setSelectedTab,
 	] );
 
-		const handleSiteNameChange = useCallback(
+	const handleSiteNameChange = useCallback(
 		async ( name: string ) => {
 			setSiteName( name );
 			if ( sitePath ) {
@@ -271,8 +271,8 @@ export function useAddSite( options: UseAddSiteOptions = {} ) {
 				}
 				setDoesPathContainWordPress( ! isEmpty && isWordPress );
 			} catch ( err ) {
-				if ( err instanceof Error ) {
-					setError( err.message );
+				if ( err instanceof Error && err.message.includes( 'ENAMETOOLONG' ) ) {
+					setError( __( 'The site name is too long. Please choose a shorter site name.' ) );
 					return;
 				}
 				throw err;
--- a/src/hooks/use-add-site.ts
+++ b/src/hooks/use-add-site.ts
@@ -245,38 +245,36 @@ export function useAddSite( options: UseAddSiteOptions = {} ) {
 			}
 			setError( '' );
 
-			try {
-				const {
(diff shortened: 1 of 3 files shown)


nightnei (approved):
LGTM and works as expected 👍
Screenshot 2025-12-30 at 13 29 40

src/hooks/use-add-site.ts
```diff
@@ -244,30 +244,39 @@ export function useAddSite( options: UseAddSiteOptions = {} ) {
...
-				return;
-			}
-			if ( ! isEmpty && ! isWordPress ) {
-				setError(
-					__(
-						'This directory is not empty. Please select an empty directory or an existing WordPress folder.'
-					)
-				);
-				return;
+			try {
```
wojtekn: As the exception can be thrown by `generateProposedSitePath`, it would be clearer to wrap only this part with `try/catch`.

However, in `generateProposedSitePath`, we already catch the underlying exception and set returned props accordingly, so we could consider moving exception handling there.

gavande1: Initially, My first approach was to handle exceptions in `generateProposedSitePath`. However, the main issue was with the error message. If we handle the exception in IPC handlers, it prefixes the following error message. 

```
Error invoking remote method 'generateProposedSitePath'
```

if we want to remove this prefix, I would need to add an additional parameter, such as `isError`, to indicate that there was a problem. However, I believe this is unnecessary.

CleanShot 2025-12-31 at 14 07 53@2x

wojtekn: What if we handled the exception there, returned `isNameTooLong` prop, and set the error in the renderer process? It would be consistent with current error handling.

gavande1: Yes. To make it consistent we can do that. I will update the PR.

gavande1: Updated in [34e6d43](https://github.com/Automattic/studio/pull/2336/commits/34e6d431630d283f2fed6b057769b9c58c6650b7)!

katinthehatsite (approved):
This looks good to me 👍 

I got a visual indication when the site name was too long:

Image

PR #2338, opened 2025-12-30
Cleanup useSyncSites

nightnei:
## Related issues

STU-176

## Proposed Changes
I was looking at this issue (STU-176) and found out that we changed approach and got rid of syncSites array so while checking it I noticed that we have some leftovers, so this PR removes them.

## Testing Instructions
Tests in the PR are green

### DIFF
--- a/src/hooks/tests/use-add-site.test.tsx
+++ b/src/hooks/tests/use-add-site.test.tsx
@@ -9,6 +9,7 @@ import { useSiteDetails } from 'src/hooks/use-site-details';
 import { getWordPressProvider } from 'src/lib/wordpress-provider';
 import { store } from 'src/stores';
 import { setProviderConstants } from 'src/stores/provider-constants-slice';
+import type { SyncSitesContextType } from 'src/hooks/sync-sites/sync-sites-context';
 import type { SyncSite } from 'src/modules/sync/types';
 
 jest.mock( 'src/hooks/use-site-details' );
@@ -75,9 +76,6 @@ describe( 'useAddSite', () => {
 		mockPullSite.mockReset();
 		( useSyncSites as jest.Mock ).mockReturnValue( {
 			pullSite: mockPullSite,
-			syncSites: [],
-			refetchSites: jest.fn(),
-			isFetching: false,
 			isAnySitePulling: false,
 			isSiteIdPulling: jest.fn(),
 			clearPullState: jest.fn(),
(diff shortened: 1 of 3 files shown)


src/hooks/tests/use-add-site.test.tsx
```diff
@@ -89,7 +87,8 @@ describe( 'useAddSite', () => {
 			clearPushState: jest.fn(),
 			getPushState: jest.fn(),
 			getLastSyncTimeText: jest.fn(),
-		} );
+			cancelPush: jest.fn(),
+		} as SyncSitesContextType );
```
sejas, line 91:
```suggestion
		} satisfies SyncSitesContextType );
```

src/modules/add-site/tests/add-site.test.tsx
```diff
@@ -146,7 +144,8 @@ beforeEach( () => {
 		clearPushState: jest.fn(),
 		getPushState: jest.fn(),
 		getLastSyncTimeText: jest.fn(),
-	} );
+		cancelPush: jest.fn(),
+	} as SyncSitesContextType );
```
sejas, line 148:
```suggestion
	} satisfies SyncSitesContextType );
```

sejas (approved):
The changes look good. They only affect the tests, and they pass. I suggest using `satisfies` instead of `as` so it forces to keep the type correct.

src/modules/add-site/tests/add-site.test.tsx
nightnei: @sejas this way we have indeed more strcit check, but we wil lneed to make more specific with functions, I mean instead of just `jest.fn()` - we need to specify return types and the function type itself. For example `isSiteIdPulling: jest.fn< IsSiteIdPulling >().mockReturnValue( false ),`. And it will add extra exports, sicne for example `IsSiteIdPulling` is not exported so far outside of its module.

On the one hand it woudl be nice to be very specific with mocking, but on teh otehr side it makes everythign more wordy. Both options are good to me, so WDYT?

sejas: I'm ok keeping the `as` in this test file. I tried to find alternatives, but it seems that satisfies requires full match in this file. I don't fully understand why it works in `src/hooks/tests/use-add-site.test.tsx`.

nightnei: Hm, the only difference is how mock happens, but theoretically it should be the same, since we do `satisfies ` for the argument directly, interesting.
`( useSyncSites as jest.Mock ).mockReturnValue( {`
vs
`mockUseSyncSites.mockReturnValue( {`

PR #2345, opened 2026-01-02
Statically import `runCommand` functions in CLI tests

fredrikekelund:
## Related issues

- Fixes STU-1187

## Proposed Changes

@wojtekn asked (p1766602221612649-slack-C06DRMD6VPZ) why we dynamically import the `runCommand` functions in the CLI tests. When I dug into this, I quickly concluded that it's not necessary. Static imports yield better performance and better error propagation.

So, why did we use the dynamic imports previously? I believe it's because we used to call `process.exit()` in some of the CLI command implementations, and dynamically importing the functions helped work around that in the tests. 

## Testing Instructions

Unit tests should pass

## Pre-merge Checklist

### DIFF
--- a/cli/commands/auth/tests/login.test.ts
+++ b/cli/commands/auth/tests/login.test.ts
@@ -11,6 +11,7 @@ import {
 import { openBrowser } from 'cli/lib/browser';
 import { getAppLocale } from 'cli/lib/i18n';
 import { Logger, LoggerError } from 'cli/logger';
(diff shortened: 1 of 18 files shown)


nightnei (approved):
LGTM 👍
I restarted tests, it seems they are not relevant to the changes.

wojtekn:
Thanks for changing that - it will improve the developer and testing experience.

PR #2355, opened 2026-01-06
Consolidate studio site set-* commands into single command

bcotrim:
## Related issues

- Fixes STU-1079

## Proposed Changes

- Replace four separate `studio site set-*` commands (`set-domain`, `set-https`, `set-php-version`, `set-wp-version`) with a single `studio site set` command
- Support multiple options at once: `--name`, `--domain`, `--https`, `--php`, `--wp`
- Add new `--name` option to change site name (no restart needed, metadata only)
- Server restarts only once even when multiple options change
- Detect "no changes" scenario and show error instead of doing nothing
- Pass correct site URL to WP-CLI to avoid corrupting WordPress URL options when using custom domains with HTTPS

## Testing Instructions

1. Create a test site: `studio site create --path /tmp/test-site`
2. Start the site: `studio site start --path /tmp/test-site`

**Test single options:**
- `studio site set --name "New Name" --path /tmp/test-site` - should update name without restart
- `studio site set --php 8.2 --path /tmp/test-site` - should update PHP and restart

**Test multiple options:**
- `studio site set --name "Test" --php 8.3 --path /tmp/test-site` - should apply both changes with single restart

**Test edge cases:**
- `studio site set --path /tmp/test-site` - should show error "At least one option required"
- `studio site set --name "Test" --path /tmp/test-site` (when name is already "Test") - should show "No changes to apply"
- `studio site set --php invalid --path /tmp/test-site` - should show valid PHP version choices

**Test HTTPS with custom domain:**
- `studio site set --domain test.local --https true --path /tmp/test-site` - should work correctly with SSL

## Pre-merge Checklist


fredrikekelund:
Reviewing this now

cli/commands/site/set.ts
```diff
@@ -0,0 +1,289 @@
...
+		if ( needsRestart && wasRunning ) {
+			logger.reportStart( LoggerAction.STOP_SITE, __( 'Stopping WordPress site…' ) );
+			await stopWordPressServer( site.id );
+			logger.reportSuccess( __( 'WordPress site stopped' ) );
+		}
+
+		if ( wpChanged ) {
+			logger.reportStart( LoggerAction.SET_WP_VERSION, __( 'Changing WordPress version…' ) );
+			const phpVersion = validatePhpVersion( site.phpVersion );
+			const zipUrl = getWordPressVersionUrl( wp! );
```
fredrikekelund:
```suggestion
			const zipUrl = getWordPressVersionUrl( wp );
```
Nit, but we don't need to override TS here.

cli/commands/site/tests/set.test.ts
```diff
@@ -0,0 +1,360 @@
...
+		} );
+
+		it( 'should not restart running site when only name changes', async () => {
+			( isServerRunning as jest.Mock ).mockResolvedValue( testProcessDescription );
+
+			await runCommand( testSitePath, { name: 'New Name' } );
+
+			expect( stopWordPressServer ).not.toHaveBeenCalled();
+			expect( startWordPressServer ).not.toHaveBeenCalled();
+		} );
```
fredrikekelund, line 155: These two tests are almost identical. We could move the `isServerRunning` mock to the first test case and remove the second one.

cli/commands/site/tests/set.test.ts
```diff
@@ -0,0 +1,360 @@
...
+		id: 'site-1',
+		name: 'Test Site',
+		path: testSitePath,
+		port: 8080,
+		phpVersion: '8.0',
+		adminUsername: 'admin',
+		adminPassword: 'password123',
+	} );
+
+	const createTestSiteWithDomain = (): SiteData => ( {
```
fredrikekelund:
```suggestion
	const getTestSite = (): SiteData => ( {
		id: 'site-1',
		name: 'Test Site',
		path: testSitePath,
		port: 8080,
		phpVersion: '8.0',
		adminUsername: 'admin',
		adminPassword: 'password123',
	} );

	const getTestSiteWithDomain = (): SiteData => ( {
```
Truly nitpicking, but `create` seems to potentially imply that this function has side effects, whereas `get` implies that it doesn't. Up to you, though.

cli/commands/site/tests/set.test.ts
```diff
@@ -0,0 +1,360 @@
...
+			( readAppdata as jest.Mock ).mockResolvedValue( {
+				sites: [ siteWithDomain ],
+				snapshots: [],
+			} );
+			( isServerRunning as jest.Mock ).mockResolvedValue( testProcessDescription );
+
+			await runCommand( testSitePath, { https: true } );
+
+			expect( stopWordPressServer ).toHaveBeenCalledWith( 'site-1' );
+			expect( startWordPressServer ).toHaveBeenCalled();
```
fredrikekelund, line 213: Let's differentiate these tests by asserting that `stopWordPressServer` and `startWordPressServer` weren't called in the first test.

cli/commands/site/tests/set.test.ts
```diff
@@ -0,0 +1,360 @@
...
+			expect( savedAppdata.sites[ 0 ].phpVersion ).toBe( '8.2' );
+		} );
+
+		it( 'should restart running site when PHP version changes', async () => {
+			( isServerRunning as jest.Mock ).mockResolvedValue( testProcessDescription );
+
+			await runCommand( testSitePath, { php: '8.2' } );
+
+			expect( stopWordPressServer ).toHaveBeenCalledWith( 'site-1' );
+			expect( startWordPressServer ).toHaveBeenCalled();
```
fredrikekelund, line 233: Same thing here: let's differentiate the tests by asserting that `stopWordPressServer` and `startWordPressServer` weren't called in the first test.

cli/commands/site/tests/set.test.ts
```diff
@@ -0,0 +1,360 @@
...
+		const mockWpCliResponse = {
+			exitCode: Promise.resolve( 0 ),
+		};
+
+		beforeEach( () => {
+			( runWpCliCommand as jest.Mock ).mockResolvedValue( [
+				mockWpCliResponse,
+				jest.fn().mockResolvedValue( undefined ),
+			] );
+		} );
```
fredrikekelund:
```suggestion
		beforeEach( () => {
			( runWpCliCommand as jest.Mock ).mockResolvedValue( [
				{ exitCode: Promise.resolve( 0 ) },
				jest.fn().mockResolvedValue( undefined ),
			] );
		} );
```
Another true nit, but we might as well make this a bit more compact.

cli/commands/site/tests/set.test.ts
```diff
@@ -0,0 +1,360 @@
...
+				'Failed to update WordPress version to 6.7'
+			);
+		} );
+
+		it( 'should update isWpAutoUpdating when setting to latest', async () => {
+			await runCommand( testSitePath, { wp: 'latest' } );
+
+			// Second saveAppdata call updates isWpAutoUpdating
+			expect( saveAppdata ).toHaveBeenCalledTimes( 2 );
+		} );
```
fredrikekelund:
```suggestion
		it( 'should update isWpAutoUpdating to false when using specific version', async () => {
			await runCommand( testSitePath, { wp: '6.8' } );

			expect( saveAppdata ).toHaveBeenCalledWith(
				expect.objectContaining( {
					sites: expect.arrayContaining( [
						expect.objectContaining( { isWpAutoUpdating: false } ),
					] ),
				} )
			);
		} );
```
Since `isWpAutoUpdating` defaults to true, I think it's better to test that we set it to false. Also, let's actually test that and not just check the number of times `saveAppdata` is called.

cli/commands/site/tests/set.test.ts
```diff
@@ -0,0 +1,360 @@
...
+				domain: 'new.local',
+				php: '8.2',
+			} );
+
+			// Should stop once and start once
+			expect( stopWordPressServer ).toHaveBeenCalledTimes( 1 );
+			expect( startWordPressServer ).toHaveBeenCalledTimes( 1 );
+		} );
+
+		it( 'should handle name + domain change (only domain triggers restart)', async () => {
```
fredrikekelund:
```suggestion
		it( 'should restart if options contain changes that require restart and ones that don\'t', async () => {
```

cli/lib/run-wp-cli-command.ts
fredrikekelund, line 1: Could you elaborate on why we need this change, @bcotrim? I seem to remember testing the current setup and not having any trouble with it.

fredrikekelund (approved):
LGTM 👍 This turned out great. I left a couple of small comments on the test file. My only real concern is about `cli/lib/run-wp-cli-command.ts`, where I'm not sure why we need that change. It would be good to straighten that out before landing this. I'm still approving now to unblock you, though.

cli/lib/run-wp-cli-command.ts
bcotrim: Changing WP version on a site with a custom domain and https, would make the site no longer accessible.
My assumption was that the command to change wordpress version somehow uses the site url.
Using the correct site URL when running the wp upgrade command succeeded and the site was accessible.

fredrikekelund: OK, cool. Let's :shipit:

PR #2356, opened 2026-01-06
Fix sync tab items

nightnei:
## Proposed Changes
I had case when `connectedSite.environmentType` was `null` and UI was broken. It seems that we handled it previously with different grid, but it's not relevant to the current UI, that's why it breaks it.

<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2026-01-06 at 21 07 43
<td>Screenshot 2026-01-06 at 21 07 21
</tr>
</table>

## Testing Instructions
Visual code review should suffice


gavande1 (approved):
LGTM 👍

PR #2358, opened 2026-01-07
Add details about unsupported Sync item states

nightnei:
## Proposed Changes
On Sync tab we render “Upgrade plan” button, w/o any details about which one should be chosen.
It can be frustrating for users, if they click in “Upgrade plan” button, purchase for example “Personal” and still Studio asks to upgrade plan, w/o any details.
This PR adds tooltip to specify Business plan.
And also adds tooltip to explain why the site is unsupported. It even usefull for us, developers of Studio, since I had no idea why site can be unsupported and why that predicate can be true, so I had to read commits history.

## Testing Instructions
1. Open sync tab
2. Find any `Unsupported` item and `Upgrade plan` button
3. Assert that you see tooltips

### DIFF
--- a/src/modules/sync/components/sync-sites-modal-selector.tsx
+++ b/src/modules/sync/components/sync-sites-modal-selector.tsx
@@ -1,4 +1,4 @@
-import { Icon, SearchControl as SearchControlWp } from '@wordpress/components';
+import { Icon, SearchControl as SearchControlWp, Tooltip } from '@wordpress/components';
 import { __, sprintf } from '@wordpress/i18n';
 import { useI18n } from '@wordpress/react-i18n';
 import { useState, useEffect } from 'react';
@@ -339,13 +339,15 @@ function SiteItem( {
 			) }
 			{ needsUpgrade && (
 				<div className="a8c-body-small text-a8c-gray-30 shrink-0 text-right">
-					<Button
-						variant="link"
-						onClick={ () => getIpcApi().openURL( `https://wordpress.com/plans/${ site.id }` ) }
-					>
-						{ __( 'Upgrade plan' ) }
-						<ArrowIcon />
-					</Button>
+					<Tooltip text={ __( 'Sync support is available only with Business plan and above' ) }>
+						<Button
+							variant="link"
+							onClick={ () => getIpcApi().openURL( `https://wordpress.com/plans/${ site.id }` ) }
+						>
+							{ __( 'Upgrade plan' ) }
+							<ArrowIcon />
+						</Button>
+					</Tooltip>
 				</div>
 			) }
 			{ isNeedsTransfer && (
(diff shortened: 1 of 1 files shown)


gavande1:
I have tested this and noticed something unusual. It would be great if you could address it. There's a brief moment when the tooltip appears in the top left corner before moving back to its intended position. I captured a video and took a screenshot to demonstrate exactly what happens. You can see the tooltip appear at the 12-second mark, then move to its proper place.

CleanShot 2026-01-07 at 18 39 15@2x

https://github.com/user-attachments/assets/2f8e2078-122c-4aa2-a364-9e5659e874aa

gavande1 (changes requested):
I left a comment about the issue with flashing tooltip on top left corner.

nightnei:
@gavande1 Interesting, I can't reproduce it. But I added `placement="bottom"` to help lib decide where to render it, maybe it will fix the issue. If it doesn't fix, then I will take a look a bit more, but then it means that it's internal library issue and we should eitehr replace lib or live with it, since I don't think that debugging/fixing lib would be good approach, especially with such not critical thing.
Test it please one more time.

gavande1:
@nightnei Unfortunately that didn't fix the issue as it seems. 

https://github.com/user-attachments/assets/ec2cf039-1b76-4a77-b686-499795d34331

> but then it means that it's internal library issue and we should eitehr replace lib or live with it, since I don't think that debugging/fixing lib would be good approach

I totally agree. Would it make sense to identify the root cause and report the issue upstream repo in that case?

nightnei:
@gavande1 I created issue (STU-1201) for it and set low priority, since it's not a blocker and it doesn't cause any UX problems. Also, the lib is used across the whole Studio, so it should be reproducible for you everywhere, and the changes in this PR are not the root cause for it.
So I propose to merge this PR and handle that issue later separatelly.

katinthehatsite:
I see the same issue with the flashing tooltip although it does not happen with every tooltip, but for some sites it does.

gavande1 (approved):
Thanks for addressing this and creating STU-1201 for the tooltip issue!

After further review, I agree this PR should be merged:

1. ✅ The code changes add valuable UX improvements
2. ✅ The tooltip flashing is a pre-existing `@wordpress/components` library issue, not introduced by this PR
3. ✅ STU-1201 properly tracks the tooltip issue for separate resolution

Approving this PR. The tooltip library issue can be investigated and fixed separately without blocking these helpful improvements.

ivan-ottinger:
Thank you for introducing these tooltips, Vova!

src/modules/sync/components/sync-sites-modal-selector.tsx
```diff
@@ -1,4 +1,4 @@
-import { Icon, SearchControl as SearchControlWp } from '@wordpress/components';
+import { Icon, SearchControl as SearchControlWp, Tooltip } from '@wordpress/components';
```
gavande1, line 1: @nightnei This PR uses Tooltip from `@wordpress/components`, but Studio uses a custom Tooltip from `src/components/tooltip` everywhere else. Switching to the custom component should fix the flashing and maintain consistency.

PR #2359, opened 2026-01-07
Refactor sync-support.ts

nightnei:
## Proposed Changes
I worked around the sync-support module and didn't get how to reproduce cases, since the conditions are not clear.
See comments.

## Testing Instructions
Assert that everything still works well:
1. Create [jurassic.ninja](https://jurassic.ninja/) site, install jetpack and conntect to dotcom, open Studio sync tab and assert that it's marked as unsupported.
2. Create Business site, open [bulk-delete-test-sites](https://mc.a8c.com/tools/bulk-delete-test-sites/), revert to simple, open Studio sync tab and assert that the site is marked as "Enable hosting features"
3. Test wp.com sites and assert that they are marked as "syncable" only for Business+ plan.
4. Test Pressable sites and assert that they are always marked as "syncable"

### DIFF
--- a/src/modules/sync/lib/sync-support.ts
+++ b/src/modules/sync/lib/sync-support.ts
@@ -7,20 +7,23 @@ export function isPressableSite( site: SitesEndpointSite ): boolean {
 	return site.hosting_provider_guess === 'pressable';
 }
 
-export function isAtomicSite( site: SitesEndpointSite ): boolean {
+function isAtomicSite( site: SitesEndpointSite ): boolean {
 	return site.is_wpcom_atomic;
 }
 
-export function hasSupportedPlan( site: SitesEndpointSite ): boolean {
+function hasSupportedPlan( site: SitesEndpointSite ): boolean {
 	return site.plan?.features.active.includes( STUDIO_SYNC_FEATURE_NAME ) ?? false;
 }
 
-export function isJetpackSite( site: SitesEndpointSite ): boolean {
-	return !! site.jetpack && ! isAtomicSite( site ) && ! isPressableSite( site );
+// Sites hosted outside wp.com and Pressable (e.g. jurassic.ninja).
+function isUnsupported( site: SitesEndpointSite ): boolean {
+	return !! site.jetpack && ! isAtomicSite( site );
 }
 
-export function needsTransfer( site: SitesEndpointSite ): boolean {
-	return ! isJetpackSite( site ) && ! isPressableSite( site ) && ! isAtomicSite( site );
+// needsTransfer means that the site was reverted to "simple" via "bulk-delete-test-sites" tool.
+// Activating e.g. "VaultPress Backup" or "Hosting features" returns "is_wpcom_atomic === true && jetpack === true".
+function needsTransfer( site: SitesEndpointSite ): boolean {
+	return ! site.jetpack && ! isAtomicSite( site );
 }
 
 export function getSyncSupport( site: SitesEndpointSite, connectedSiteIds: number[] ): SyncSupport {
@@ -30,17 +33,18 @@ export function getSyncSupport( site: SitesEndpointSite, connectedSiteIds: numbe
 	if ( ! site.capabilities?.manage_options ) {
 		return 'missing-permissions';
 	}
-	if ( isJetpackSite( site ) ) {
+	if ( isUnsupported( site ) && ! isPressableSite( site ) ) {
 		return 'unsupported';
 	}
 	if ( ! hasSupportedPlan( site ) && ! isPressableSite( site ) ) {
 		return 'needs-upgrade';
 	}
-	if ( needsTransfer( site ) ) {
+	if ( needsTransfer( site ) && ! isPressableSite( site ) ) {
 		return 'needs-transfer';
 	}
 	if ( connectedSiteIds.some( ( id ) => id === site.ID ) ) {
 		return 'already-connected';
 	}
+
 	return 'syncable';
 }


src/modules/sync/lib/sync-support.ts
```diff
@@ -30,17 +33,18 @@ export function getSyncSupport( site: SitesEndpointSite, connectedSiteIds: numbe
 	if ( ! site.capabilities?.manage_options ) {
 		return 'missing-permissions';
 	}
-	if ( isJetpackSite( site ) ) {
```
nightnei, line 33: Why jetpack site is unsupported?
Jetpack is required. So I renamed to isUnsupported.

src/modules/sync/lib/sync-support.ts
```diff
@@ -7,20 +7,23 @@ export function isPressableSite( site: SitesEndpointSite ): boolean {
...
 	return site.is_wpcom_atomic;
 }
 
-export function hasSupportedPlan( site: SitesEndpointSite ): boolean {
+function hasSupportedPlan( site: SitesEndpointSite ): boolean {
 	return site.plan?.features.active.includes( STUDIO_SYNC_FEATURE_NAME ) ?? false;
 }
 
-export function isJetpackSite( site: SitesEndpointSite ): boolean {
-	return !! site.jetpack && ! isAtomicSite( site ) && ! isPressableSite( site );
```
nightnei, line 19: Why detection of jetpack includes ! isAtomic && ! isPressable? Theoretically `!! site.jetpack` should be enough.

I took a look at history and now the function directly says that **jetpack site, which is not atomic** is **unsupported**, and **the example is jurasic.ninja**

src/modules/sync/lib/sync-support.ts
```diff
@@ -7,20 +7,23 @@ export function isPressableSite( site: SitesEndpointSite ): boolean {
...
+// Sites hosted outside wp.com and Pressable (e.g. jurassic.ninja).
+function isUnsupported( site: SitesEndpointSite ): boolean {
+	return !! site.jetpack && ! isAtomicSite( site );
 }
 
-export function needsTransfer( site: SitesEndpointSite ): boolean {
-	return ! isJetpackSite( site ) && ! isPressableSite( site ) && ! isAtomicSite( site );
```
nightnei, line 23: If we read this condition, it says:
1. NOT `!! site.jetpack && ! isAtomicSite( site ) && ! isPressableSite( site )`
&&
2. `&& ! isPressableSite( site ) && ! isAtomicSite( site )`

It's dificult to undersathnd when it's actually true. I have read the history of comits and simplified to just `! site.jetpack && ! isAtomicSite( site )`. And added the comment to how to reproduce the case.

src/modules/sync/lib/sync-support.ts
```diff
@@ -30,17 +33,18 @@ export function getSyncSupport( site: SitesEndpointSite, connectedSiteIds: numbe
 	if ( ! site.capabilities?.manage_options ) {
 		return 'missing-permissions';
 	}
-	if ( isJetpackSite( site ) ) {
+	if ( isUnsupported( site ) && ! isPressableSite( site ) ) {
 		return 'unsupported';
 	}
 	if ( ! hasSupportedPlan( site ) && ! isPressableSite( site ) ) {
```
nightnei, line 39: In this condition we keep `isPressableSite` here, outside the `hasSupportedPlan` and imo it's good idea, to be clear at the upper level that the plan is about only dotcom, and it's not about Pressable.
So I added the same approach to two other cases (isUnsupported and needsTransfer).
Also, it's for consistency.

src/modules/sync/lib/sync-support.ts
```diff
@@ -7,20 +7,23 @@ export function isPressableSite( site: SitesEndpointSite ): boolean {
...
-	return !! site.jetpack && ! isAtomicSite( site ) && ! isPressableSite( site );
+// Sites hosted outside wp.com and Pressable (e.g. jurassic.ninja).
+function isUnsupported( site: SitesEndpointSite ): boolean {
+	return !! site.jetpack && ! isAtomicSite( site );
 }
 
-export function needsTransfer( site: SitesEndpointSite ): boolean {
-	return ! isJetpackSite( site ) && ! isPressableSite( site ) && ! isAtomicSite( site );
+// needsTransfer means that the site was reverted to "simple" via "bulk-delete-test-sites" tool.
+// Activating e.g. "VaultPress Backup" or "Hosting features" returns "is_wpcom_atomic === true && jetpack === true".
```
ivan-ottinger, line 24: I think we could simplify this comment to something like:

```
All WordPress.com Simple sites with sufficient plan fall into this category. For example, Simple site with a Business plan needs transfer.
```

src/modules/sync/lib/sync-support.ts
ivan-ottinger: > Why detection of jetpack includes ! isAtomic && ! isPressable?

As far as I remember, we used this in the past to filter out self-hosted sites that are not hosted by us (via WPCOM and Pressable).

In that logic, "Jetpack site" referred to "self-hosted WordPress site".

src/modules/sync/lib/sync-support.ts
ivan-ottinger: I think this was a workaround to check for a Simple site (a site that is not Jetpack, not Atomic and not Pressable).

It was as long time again since I looked at the old logic, but I believe there was some edge case that resulted in this more complex check.

src/modules/sync/lib/sync-support.ts
```diff
@@ -7,20 +7,23 @@ export function isPressableSite( site: SitesEndpointSite ): boolean {
...
 }
 
-export function hasSupportedPlan( site: SitesEndpointSite ): boolean {
+function hasSupportedPlan( site: SitesEndpointSite ): boolean {
 	return site.plan?.features.active.includes( STUDIO_SYNC_FEATURE_NAME ) ?? false;
 }
 
-export function isJetpackSite( site: SitesEndpointSite ): boolean {
-	return !! site.jetpack && ! isAtomicSite( site ) && ! isPressableSite( site );
+// Sites hosted outside wp.com and Pressable (e.g. jurassic.ninja).
```
ivan-ottinger, line 18: While reviewing the code I wanted to suggest moving `isPressableSite` check into the `isUnsupported` helper function directly, but then I read your other comment below:

> In this condition we keep isPressableSite here, outside the hasSupportedPlan and imo it's good idea, to be clear at the upper level that the plan is about only dotcom, and it's not about Pressable.
So I added the same approach to two other cases (isUnsupported and needsTransfer).
Also, it's for consistency.

Sounds good to me. 👍🏼 With that I think we may want to update the comment above:

```
// Sites hosted outside wp.com and Pressable (e.g. jurassic.ninja).
```

As now it looks as if `isUnsupported` also checks for Pressable sites.

src/modules/sync/lib/sync-support.ts
ivan-ottinger: Similar as in my previous comment:

> As far as I remember, we used this in the past to filter out self-hosted sites that are not hosted by us (via WPCOM and Pressable).

> In that logic, "Jetpack site" referred to "self-hosted WordPress site".

ivan-ottinger (commented):
Thanks for the improvements, Vova. This logic is indeed difficult to troubleshoot.

So far the changes look generally good to me and I haven't noticed any regressions so far. I have left a couple of comments.

Tomorrow I will give this a second look with fresh 🧠 . 🙂

src/modules/sync/lib/sync-support.ts
nightnei: The main point here - below we have a condition that Jetpack sites are unsupported:

```
if ( isJetpackSite( site ) ) {
	return 'unsupported';
}
```
But it's not true, since all supported sites *must* have Jetpack. I worked around this code, needed cases to reproduce it, and got stuck.

> In that logic, "Jetpack site" referred to "self-hosted WordPress site".

Then it should be either `isUnsupported` or `isSelfHosted`, but not `isJetpackSite`.

src/modules/sync/lib/sync-support.ts
nightnei: The main point here, that the condition is - `! ( !! site.jetpack && ! isAtomicSite( site ) && ! isPressableSite( site ) ) && ! isPressableSite( site ) && ! isAtomicSite( site )`, and it's not possible to understand it. And I haven't found any use case for something like this. Also reading commits history - I haven't found any "Testing instruction" or some PR description which says about some unusual use case.

src/modules/sync/lib/sync-support.ts
nightnei: Yeah, but it doesn't answer how it's possible to have simple site with Business plan, since all Business plan sites are atomic by default. I used this comment specifically since it says how it can be true (via `bulk-delete-test-sites`). And this way it will save time for developers in the future, who will be reading this code and thinking - how the site can be Simple on Business plan and how to test it.

ivan-ottinger: Makes sense. 👍🏼

ivan-ottinger (approved):
The changes looks good to me. 👍🏼 I have tested the Sync modal with different kinds of sites and did not observe any regressions:

Image

Image

katinthehatsite (commented):
I just wanted to approve this but it looks like it is already merged. The changes looked good to me, I did not see any regressions in the Connect your site modal.

PR #2360, opened 2026-01-07
Use CLI site set command to edit site details from Studio

bcotrim:
## Related issues

- Fixes STU-1149

## Proposed Changes

- Add `cli-site-editor.ts` module to invoke the CLI's `studio site set` command
- Update IPC handler `updateSite()` to delegate to CLI for site changes (domain, HTTPS, PHP version, WP version)
- Simplify `edit-site-details.tsx` - UI now just calls `updateSite()` and lets CLI handle restart logic
- Simplify `site-server.ts` `updateSiteDetails()` since CLI handles hosts file, SSL certs, and WP updates
- Create shared utility `siteNeedsRestart()` in `common/lib/` so CLI and Studio UI share the same restart logic

## Testing Instructions

1. Create a site in Studio
2. Start the site
3. Open Edit Site modal and change:
   - Site name (no restart needed)
   - PHP version (should restart)
   - WordPress version (should restart and update WP)
   - Add custom domain (should restart)
   - Enable HTTPS (should restart)
4. Verify each change works correctly and server restarts when needed
5. Test editing while site is stopped (should update without restart)

## Pre-merge Checklist

### DIFF
--- a/cli/commands/site/set.ts
+++ b/cli/commands/site/set.ts
@@ -3,6 +3,7 @@ import { __, sprintf } from '@wordpress/i18n';
 import { DEFAULT_WORDPRESS_VERSION, MINIMUM_WORDPRESS_VERSION } from 'common/constants';
 import { getDomainNameValidationError } from 'common/lib/domains';
 import { arePathsEqual } from 'common/lib/fs-utils';
+import { siteNeedsRestart } from 'common/lib/site-needs-restart';
 import {
 	getWordPressVersionUrl,
 	isValidWordPressVersion,
@@ -129,7 +130,13 @@ export async function runCommand(
 			);
 		}
 
-		const needsRestart = domainChanged || httpsChanged || phpChanged || wpChanged || xdebugChanged;
+		const needsRestart = siteNeedsRestart( {
+			domainChanged,
+			httpsChanged,
+			phpChanged,
+			wpChanged,
+			xdebugChanged,
+		} );
 		const oldDomain = site.customDomain;
 
 		try {
--- a/cli/commands/site/set.ts
+++ b/cli/commands/site/set.ts
@@ -69,7 +69,8 @@ export async function runCommand(
 		throw new LoggerError( __( 'Site name cannot be empty.' ) );
 	}
 
-	if ( xdebug !== undefined && ! ( await isXdebugBetaEnabled() ) ) {
+
+	if ( xdebug === true && ! ( await isXdebugBetaEnabled() ) ) {
 		throw new LoggerError(
 			__( 'Xdebug support is a beta feature. Enable it in Studio settings first.' )
 		);
@@ -82,7 +83,8 @@ export async function runCommand(
 
 		const initialAppdata = await readAppdata();
 
-		if ( domain !== undefined ) {
+		// Only validate when setting a domain (non-empty), not when clearing it (empty string)
+		if ( domain ) {
 			const existingDomainNames = initialAppdata.sites
 				.filter( ( s ) => s.id !== site.id )
 				.map( ( s ) => s.customDomain )
@@ -151,7 +153,8 @@ export async function runCommand(
 				foundSite.name = name!;
 			}
 			if ( domainChanged ) {
-				foundSite.customDomain = domain;
+				// Use undefined instead of empty string to remove the field from appdata
+				foundSite.customDomain = domain || undefined;
 			}
 			if ( httpsChanged ) {
 				foundSite.enableHttps = https;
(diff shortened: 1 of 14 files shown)


cli/commands/site/set.ts
```diff
@@ -319,6 +324,7 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 					const loggerError = new LoggerError( __( 'Failed to configure site' ), error );
 					logger.reportError( loggerError );
 				}
+				process.exit( 1 );
```
fredrikekelund, line 327: Nice catch 👍

cli/lib/wordpress-server-manager.ts
```diff
@@ -228,6 +232,17 @@ async function sendMessage(
...
+		processEventHandler = ( event: unknown ) => {
+			const result = pm2ProcessEventSchema.safeParse( event );
+			if ( ! result.success ) {
+				return;
+			}
+
+			if ( result.data.process.name === processName && result.data.event === 'exit' ) {
+				reject( new Error( 'WordPress server process exited unexpectedly' ) );
+			}
+		};
```
fredrikekelund, line 244: This seems sensible. Out of curiosity, did you add this because of something related to the core purpose of this PR, or was it just something that happened to come up here?

src/ipc-handlers.ts
```diff
@@ -347,23 +348,69 @@ export async function createSite(
 
 export async function updateSite(
```
fredrikekelund, line 349: Aside from being exposed as an IPC function, this function is also called in from multiple places in this file. It seems like the old purpose was to:
1. Write changes to the appdata file.
2. Update the `SiteServer` instance itself (which might have been entirely redundant, given that we pass `SiteServer.details` to this function already).

I'm curious if we really need to keep calling this function from `startServer`, `stopServer`, etc..?

cli/wordpress-server-child.ts
```diff
@@ -344,11 +344,48 @@ const runWpCliCommand = sequential(
...
+		if ( textContent ) {
+			return `WordPress error: ${ textContent }`;
+		}
+	}
+
+	// Check for PHP fatal error pattern
+	const fatalMatch = message.match( /PHP Fatal error:\s*(.+?)(?:\sin\s|$)/i );
+	if ( fatalMatch ) {
+		return `PHP Fatal error: ${ fatalMatch[ 1 ].trim() }`;
+	}
```
fredrikekelund, line 372: I'm curious what these two conditions are based on. XDebug errors?

src/modules/site-settings/edit-site-details.tsx
fredrikekelund, line 1: It's a small thing, but I still love the changes to this file. This is the kind of business logic consolidation that I think will do Studio a lot of good 👍

src/site-server.ts
```diff
@@ -207,40 +198,15 @@ export class SiteServer {
 			isWpAutoUpdating: site.isWpAutoUpdating,
 			customDomain: site.customDomain,
 			enableHttps: site.enableHttps,
+			tlsKey: site.tlsKey,
+			tlsCert: site.tlsCert,
```
fredrikekelund, line 202: Note for the future: we don't need to load these details in the Studio app anymore

src/modules/cli/lib/cli-site-editor.ts
```diff
@@ -0,0 +1,88 @@
...
+
+			if ( parsed.data.status === 'inprogress' ) {
+				console.log( `[CLI - ${ options.siteId }] ${ parsed.data.message }` );
+			} else if ( parsed.data.status === 'fail' ) {
+				lastErrorMessage = parsed.data.message;
+			}
+		} );
+
+		emitter.on( 'success', () => {
+			console.log( `[CLI Site Editor] Command completed successfully` );
```
fredrikekelund: Do we need this?

fredrikekelund (approved):
LGTM 👍 Sensible and manageable diff. Nice work!

cli/lib/wordpress-server-manager.ts
bcotrim: This came up during testing this PR. When changing WP version to an incompatible one (e.g., 6.4.7 with Twenty Twenty-Five theme which requires WP 6.5+), the CLI would hang forever instead of reporting the error.

src/ipc-handlers.ts
bcotrim: Good question, I think it deserves a closer look. I added a note on STU-1193 and will tackle it in that PR if you agree.

cli/wordpress-server-child.ts
bcotrim: It also related to the WordPress issue I came across during testing.
I just tried to clean the errors a bit so it's more UI friendly.
image

PR #2364, opened 2026-01-08
Performance: Use bundled `latest` WordPress version in online mode

ivan-ottinger:
## Related issues

- Related to STU-980
- Related to STU-873

## Proposed Changes

- use bundled `latest` WordPress version in the online mode

This PR proposal is to use the bundled WordPress for the online mode as well (not just offline mode). This way we will shorten the site creation time from approx. 37 seconds to 24 seconds (35% speed up).

One disadvantage in this approach I see is that WordPress can be released in the time between two Studio releases. This would mean users who create a new site in that period would not get the most latest WordPress version (as the bundled would be used instead).

Considering our frequent Studio releases and the performance improvement for everyone we can get, I think this is a good trade-off.

## Testing Instructions

1. Check out the PR branch and start the app with `npm install && npm start`.
2. Create a new site with the `latest` WordPress version selected.
3. Review the logs in terminal - the bundled WordPress should be used:

CleanShot 2026-01-08 at 12 52 39@2x

4. Create a new site with some older WordPress version (e.g. `6.7`).
5. Review the logs in terminal - there should be no bundled version copying, the older WordPress version should be downloaded:

CleanShot 2026-01-08 at 12 53 49@2x

Creating a site using the bundled WordPress version should be generally faster. 

## Pre-merge Checklist

### DIFF
--- a/cli/commands/site/tests/create.test.ts
+++ b/cli/commands/site/tests/create.test.ts
@@ -106,7 +106,7 @@ describe( 'CLI: studio site create', () => {
 			'wordpress-versions',
 			'latest'
 		);
-		return jest.fn().mockImplementation( ( path: string ) => {
+		const mock = jest.fn().mockImplementation( ( path: string ) => {
 			if ( path === bundledWPPath ) {
 				return Promise.resolve( true );
 			}
@@ -115,6 +115,7 @@ describe( 'CLI: studio site create', () => {
 			}
 			return Promise.resolve( false );
 		} );
+		( pathExists as jest.Mock ).mockImplementation( mock );
 	};
 
 	beforeEach( () => {
@@ -122,8 +123,7 @@ describe( 'CLI: studio site create', () => {
 
 		consoleLogSpy = jest.spyOn( console, 'log' ).mockImplementation();
 		fsMkdirSyncSpy = jest.spyOn( require( 'fs' ), 'mkdirSync' ).mockReturnValue( undefined );
-		const pathExistsMock = createPathExistsMock( false );
-		( pathExists as jest.Mock ).mockImplementation( pathExistsMock );
+		createPathExistsMock( false );
 		( isEmptyDir as jest.Mock ).mockResolvedValue( true );
 		( isWordPressDirectory as jest.Mock ).mockReturnValue( false );
 		( arePathsEqual as jest.Mock ).mockImplementation( ( a, b ) => a === b );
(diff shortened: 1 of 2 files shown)


cli/commands/site/tests/create.test.ts
```diff
@@ -94,15 +100,34 @@ describe( 'CLI: studio site create', () => {
...
+		return jest.fn().mockImplementation( ( path: string ) => {
+			if ( path === bundledWPPath ) {
+				return Promise.resolve( true );
+			}
+			if ( path === mockSitePath ) {
+				return Promise.resolve( sitePathExists );
+			}
+			return Promise.resolve( false );
+		} );
+	};
```
fredrikekelund, line 119:
```suggestion
	const createPathExistsMock = ( sitePathExists = false ) => {
		const bundledWPPath = require( 'path' ).join(
			'/test/server-files',
			'wordpress-versions',
			'latest'
		);
		const mock = jest.fn().mockImplementation( ( path: string ) => {
			if ( path === bundledWPPath ) {
				return Promise.resolve( true );
			}
			if ( path === mockSitePath ) {
				return Promise.resolve( sitePathExists );
			}
			return Promise.resolve( false );
		} );
		( pathExists as jest.Mock ).mockImplementation( mock );
	};
```
Optional nit, but I think we might as well set the `pathExists` mock inside this function.

fredrikekelund (approved):
LGTM and works as expected 👍

@ivan-ottinger, I've added STU-1202 as a future follow-up to run a regular background task that updates the bundled WordPress installation. This add robustness.

bcotrim (approved):
LGTM 👍

cli/commands/site/tests/create.test.ts
ivan-ottinger: I like that! Made the change in https://github.com/Automattic/studio/pull/2364/commits/15bc0497e5346fb41c861d880160e05160f8f7b4. Thanks!

ivan-ottinger:
Thank you for the reviews and testing, Fredrik and Bernardo!

> @ivan-ottinger, I've added [STU-1202](https://linear.app/a8c/issue/STU-1202) as a future follow-up to run a regular background task that updates the bundled WordPress installation. This add robustness.

Sounds good to me. 👍🏼  Thanks for creating the task!

fredrikekelund:
FYI, it's expected for the E2E tests to not pass. We're currently blocked by #2299 on that front

PR #2365, opened 2026-01-08
Use custom tooltip instead of wordpress components

nightnei:
Fixes STU-1201

## Proposed Changes
Follow up to https://github.com/Automattic/studio/pull/2358
To use our Tooltip instead of wordpress/components

## Testing Instructions
1. Open sync tab
2. Find any Unsupported or Upgrade plan button
3. Mouse over
4. Assert that it appears w/o flashing (previously it appeared sometimes for a bit in teh left top corner)


katinthehatsite (approved):
I don't see the weird flashing anymore 👍 

Image

We could try to make the tooltip stay on one line but I also think it is fine as is

gavande1 (approved):
Nice. I don't the issue with tooltip flashing anymore. Great! ✅

nightnei:
@gavande1 thanks for the thorough review of the original PR and noticing that the path was not to our Tooltip 👍

PR #2366, opened 2026-01-08
Studio: Handle error when reading backup directories from /ls endpoint

katinthehatsite:
## Related issues

Fixes [STU-702](https://linear.app/a8c/issue/STU-702)

## Proposed Changes

This PR adds the error handling for when there is an error fetching remote children in the file tree:

Screenshot 2026-01-08 at 4 02 21 PM

## Testing Instructions

* Apply the following diff:

```diff --git a/src/stores/sync/sync-api.ts b/src/stores/sync/sync-api.ts
index 7babf7fd..d98eaf13 100644
--- a/src/stores/sync/sync-api.ts
+++ b/src/stores/sync/sync-api.ts
@@ -112,6 +112,11 @@ export const fetchRemoteFileTree = createAsyncThunk(
 
                let rawResponse;
                try {
+                       // TEMPORARY: Simulate error for testing specific node
+                       if ( path === '/wp-content/themes/' ) {
+                               throw new Error( 'Simulated error for themes folder' );
+                       }
+
                        rawResponse = await client.req.post( {
                                path: `/sites/${ remoteSiteId }/rewind/backup/ls`,
                                apiNamespace: 'wpcom/v2',
```

* Ensure that you have at least one WP.com site connected to a local Studio site
* Navigate to the Sync tab
* Click on the Pull button
* Open the `Specific files and folders section`
* Open the `themes` folder
* Confirm that you can see the error indicated on the screenshot

## Pre-merge Checklist

### DIFF
--- a/src/modules/sync/components/sync-dialog.tsx
+++ b/src/modules/sync/components/sync-dialog.tsx
@@ -238,14 +238,14 @@ export function SyncDialog( {
 				try {
 					const children = await fetchChildren( remoteSite.id, rewindId, node.path, node.checked );
 					setTreeState( ( prev ) =>
-						updateNodeById( prev, node.id, { children, loading: false, error: undefined } )
+						updateNodeById( prev, node.id, { children, loading: false, hasError: false } )
 					);
 				} catch ( error ) {
 					setTreeState( ( prev ) =>
 						updateNodeById( prev, node.id, {
 							children: [],
 							loading: false,
-							error: 'Error retrieving files and directories',
+							hasError: true,
 						} )
 					);
 				}
@@ -402,7 +402,7 @@ export function SyncDialog( {
 												</div>
 											);
 										}
-										if ( node.error ) {
+										if ( node.hasError ) {
 											return (
 												<div className="text-gray-500 italic">
 													{ __(
--- a/src/modules/sync/components/sync-dialog.tsx
+++ b/src/modules/sync/components/sync-dialog.tsx
@@ -393,16 +393,10 @@ export function SyncDialog( {
 												</div>
 											);
 										}
-										if ( nodeId === 'wp-content' && type === 'pull' && remoteFileTreeError ) {
-											return (
-												<div className="text-gray-500 italic">
-													{ __(
-														'Error retrieving remote files and directories. Please close and reopen this dialog to try again.'
-													) }
-												</div>
-											);
-										}
-										if ( node.hasError ) {
+										if (
+											( nodeId === 'wp-content' && type === 'pull' && remoteFileTreeError ) ||
+											node.hasError
+										) {
 											return (
 												<div className="text-gray-500 italic">
 													{ __(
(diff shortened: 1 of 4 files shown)


katinthehatsite:
Just briefly wanted to mention that the design, style and functionality of this should be consistent with the case when we fail to fetch local tree.

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -371,6 +393,24 @@ export function SyncDialog( {
...
+										if ( nodeId === 'wp-content' && type === 'pull' && remoteFileTreeError ) {
+											return (
+												<div className="text-gray-500 italic">
+													{ __(
+														'Error retrieving remote files and directories. Please close and reopen this dialog to try again.'
+													) }
+												</div>
+											);
+										}
+										if ( node.error ) {
```
gavande1: I see there are two duplicate code segments. Can we move `node.error` to `||` (or) condition to avoid duplicating?

src/modules/sync/components/sync-dialog.tsx
```diff
@@ -225,8 +232,23 @@ export function SyncDialog( {
...
+					const children = await fetchChildren( remoteSite.id, rewindId, node.path, node.checked );
+					setTreeState( ( prev ) =>
+						updateNodeById( prev, node.id, { children, loading: false, error: undefined } )
+					);
+				} catch ( error ) {
+					setTreeState( ( prev ) =>
+						updateNodeById( prev, node.id, {
+							children: [],
+							loading: false,
+							error: 'Error retrieving files and directories',
```
gavande1: Should we use the `error` object to show the actual message? Or are we fine with a hard-coded message? It seems the `error` property is just to store if there was a problem. Could we just use a boolean value instead, like `hasError` or something similar?

gavande1 (approved):
I have tested this and it works as expected. LGTM ✅ 

I have left minor comments, it would be great if you could address them. 

Image

src/modules/sync/components/sync-dialog.tsx
katinthehatsite: > Should we use the error object to show the actual message? Or are we fine with a hard-coded message? 

My intention was to use the hard-coded message. I don't think displaying the exact error message would add much information value to the user. 

> It seems the error property is just to store if there was a problem. Could we just use a boolean value instead, like hasError or something similar?

Yes, exactly. We can use `hasError` boolean 👍

katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2366/changes/b398c386e381dc00aaa0e347eebb8418cdd85cbe

src/modules/sync/components/sync-dialog.tsx
katinthehatsite: Nice catch, adjusted in https://github.com/Automattic/studio/pull/2366/changes/762437bfb95ac22107f22f9ccc5f0bd54dea6735 👍

PR #2361, opened 2026-01-07
CLI: Replace `stop-all` command with `--all` option in `site stop`

nightnei:
## Related issues

- Fixes STU-1191

## Proposed Changes
Removes the `studio site stop-all` CLI command in favor of an `--all` option in the `studio site stop` command

Off topic: I am working on trying to simplify code to have one function to stop either one or all sites, but so far I don't like the result, so I decided to create this PR, and then I will work a bit more to try to make code more beautiful.

## Testing Instructions
1. `npm run cli:build`
2. Create 3 sites
3. `node dist/cli/main.js site list`
4. Assert that all of them are run
5. `node dist/cli/main.js site stop --all`
6. Assert that all of them are stopped
7. Run again `node dist/cli/main.js site stop-all`
8. Assert `No sites are currently running`
9. Start 1 site
10. `node dist/cli/main.js site stop --all`
11. Assert that you see the message that only one was stopped
12. `node dist/cli/main.js site stop --path=/path/to/site`
13. Assert that one site was stopped w/o regressions

PR #2367, opened 2026-01-08
CLI: Add `runOnSites` abstraction and utilize it to refactor `site stop` and `site stop --all` commands

nightnei:
## Proposed Changes
This is proposition PR and follow-up to https://github.com/Automattic/studio/pull/2361

This PR:
* Introduces `runOnSites` abstraction in `cli/lib/site-utils.ts` that provides a unified way to execute operations on either a single site or all sites
* Refactors `site stop` command to use this abstraction, merging `runCommand` and `runCommandAll` into a single function
* `runOnSites` can be used across the whole CLI, to avoid code duplications (e.g. `getSiteByFolder`, `readAppdata`, etc), having standardized way of running commands (especially when the command has `--all` flag).

#### Single site
```js
const result = await runOnSites( '/path/to/site', async ( site ) => {
  await stopServer( site.id );
} );
```
#### All sites
```js
const result = await runOnSites( ALL_SITES, async ( site ) => {
  await stopServer( site.id );
} );
```
#### All running sites
```js

const result = await runOnSites(
  ALL_SITES,
  async (site) => {
    await stopServer( site.id );
  },
  {
    filter: async (site) => await isServerRunning(site.id)
  },
);
```

## Testing Instructions
* The same as for the parent PR - https://github.com/Automattic/studio/pull/2361
* Also assert that tests

### DIFF
--- a/cli/commands/site/stop.ts
+++ b/cli/commands/site/stop.ts
@@ -1,138 +1,99 @@
 import { __, _n, sprintf } from '@wordpress/i18n';
 import { SiteCommandLoggerAction as LoggerAction } from 'common/logger-actions';
-import {
-	clearSiteLatestCliPid,
-	getSiteByFolder,
-	readAppdata,
-	updateSiteAutoStart,
-	type SiteData,
-} from 'cli/lib/appdata';
+import { clearSiteLatestCliPid, updateSiteAutoStart } from 'cli/lib/appdata';
 import { connect, disconnect } from 'cli/lib/pm2-manager';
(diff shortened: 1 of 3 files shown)


fredrikekelund (commented):
I find this a bit too clever, tbh. I prefer the simplicity of #2361, even if I would also prefer to have a single `runCommand` function. I'll submit a suggestion on #2361 for how we might simplify the logic there.

fredrikekelund:
Closing this in favor of #2374 as you're AFK, @nightnei

PR #2375, opened 2026-01-09
Experiment to speed up deeplink connection from WP.com

katinthehatsite:
## Related issues

Related to STU-1173

## Proposed Changes

This PR explores the speeding up of the deeplink connection from WP.com e.g. connecting a WP.com site and opening a modal. 

*  immediately create a minimal site object with data from the deeplink (this would need to be done on WP.com side, but we can add WP.com site's URL and site name which would be enough for our UI)
* display the loading skeleton while we are connecting
* fetch the remote site right away
*  and trigger the API refetch in parallel without blocking

## Testing Instructions

* Start Studio
* Navigate to the `Sync` tab
* Complete the purchase
* Confirm that you can see the loading skeleton on the `Sync` tab
* Confirm that the site you just created is connected in the `Sync` tab
* Confirm that the `Push` modal is open
* Confirm that there are no other regressions

## Pre-merge Checklist


katinthehatsite:
The code needs to be polished quite a bit so I am leaving this PR as a draft. What I am looking for is the feedback on the approach before adding any further polishing touches - does it make sense? is there a different way we could optimize the speed of connection that could be simpler?

One thing I dislike is creating a "fictional" site object that could potentially cause racing conditions and similar issues. Any opinions are welcome!

One option as an alternative that comes to mind could be adding a loading state to Sync tab if the user is connecting a site from deeplink and it is taking time. This way, they could have a visual indication of what is happening and this could be a simple and fast improvement in terms of the UX.

src/hooks/sync-sites/use-listen-deep-link-connection.ts
```diff
@@ -28,25 +29,65 @@ export function useListenDeepLinkConnection() {
...
+			const connectPromise = connectSite( { site: minimalSite, localSiteId: studioSiteId } );
+
+			// Only auto-open push dialog if explicitly requested (e.g., from "Publish site" button)
+			// Open modal immediately with minimal data
+			if ( autoOpenPush ) {
+				dispatch( connectedSitesActions.setSelectedRemoteSiteId( remoteSiteId ) );
+			}
+
+			// Fetch full site data in background (don't await - parallel operation)
+			const refetchPromise = refetchWpComSites();
```
sejas: Another optimization could be fetching only the information of the remoteSiteId. Currently fetching all the sites takes longer, specially for Automatticians.

src/hooks/sync-sites/use-listen-deep-link-connection.ts
```diff
@@ -28,25 +29,65 @@ export function useListenDeepLinkConnection() {
...
+				siteName?: string;
+				siteUrl?: string;
+			}
+		) => {
+			// Create minimal site object optimistically to connect immediately
+			// Use siteName and siteUrl from deeplink if available, otherwise use placeholders
+			const minimalSite: SyncSite = {
+				id: remoteSiteId,
+				localSiteId: studioSiteId,
+				name: siteName || 'Loading site...', // Use provided name or placeholder
```
sejas: We need to translate the placeholder.

sejas (commented):
Thanks for exploring how to speed up the deeplink connection.
In general the approach makes sense to me. Loading an optimistic connection until we fetch more information.
My concern is if we display wrong data, like isStaging, or isPressable. With the current flow where the user purchase a new site, those will always be false, but in the future we could add more deeplinks in Calypso.
What if we show a loading skeleton until we fetch the information?
As mentioned below, we can optimize the fetch call to get only the information for that specific site.

katinthehatsite:
Thank you for review @sejas - I will take a look at the next steps today 👀

src/hooks/sync-sites/use-listen-deep-link-connection.ts
katinthehatsite: I think it was not needed at the end after I added the loading placeholder so I removed it.

src/hooks/sync-sites/use-listen-deep-link-connection.ts
katinthehatsite: I found an endpoint that allowed to fetch just one site instead of all sites so I went with that solution.

katinthehatsite:
@sejas I made some adjustments on this branch based on your recommendations. Would you be able to give it another look?

apps/studio/src/hooks/sync-sites/use-listen-deep-link-connection.ts
```diff
@@ -28,25 +31,91 @@ export function useListenDeepLinkConnection() {
...
+						...singleSiteResult.data,
+						localSiteId: studioSiteId,
+					};
+					// Use updateSingleConnectedWpcomSite to overwrite the minimal site in storage.
+					// connectSite only adds new sites — it won't update an existing entry.
+					await getIpcApi().updateSingleConnectedWpcomSite( fullSiteData );
+					dispatch( connectedSitesApi.util.invalidateTags( [ 'ConnectedSites' ] ) );
 				}
+
+				fetchSingleSitePromise.unsubscribe();
```
sejas: What do you think about moving this unsubscribe call to the finally block, so it's always executed?

apps/studio/src/stores/sync/wpcom-sites.ts
```diff
@@ -122,6 +122,73 @@ export const wpcomSitesApi = createApi( {
...
 	endpoints: ( builder ) => ( {
+		getSingleWpComSite: builder.query< SyncSite, { siteId: number; userId?: number } >( {
+			queryFn: async ( { siteId } ) => {
+				const wpcomClient = getWpcomClient();
+				if ( ! wpcomClient ) {
+					return { error: { status: 401, data: 'Not authenticated' } };
+				}
+
+				try {
+					const fields = [
```
sejas: Would be possible to extract and re-use these fields? I think they are the same used in https://github.com/Automattic/studio/blob/e5630d537ef9d26a66dae0eeeb9a29586d641263/apps/studio/src/stores/sync/wpcom-sites.ts#L135-L139

apps/studio/src/modules/sync/components/sync-sites-modal-selector.tsx
```diff
@@ -285,48 +290,67 @@ function SiteItem( {
...
-						}
-					} }
-				>
-					<div className="truncate">{ site.url.replace( /^https?:\/\//, '' ) }</div>
-					<ArrowIcon />
-				</Button>
+				{ isSiteLoading ? (
+					<div className="flex items-center gap-1.5">
+						<div
+							className="w-3 h-3 rounded-full animate-pulse bg-gradient-to-r from-[#F6F7F7] via-[#DCDCDE] to-[#F6F7F7]"
```
sejas: Let's re-use the animate classes by using a shared string. Something like:
```
const skeletonBg = 'animate-pulse bg-gradient-to-r from-[#F6F7F7] via-[#DCDCDE] to-[#F6F7F7]';
```

We could simplify it further by re-using that string across the components, or creating a Tailwind utility class. We currently use a similar approach in https://github.com/Automattic/studio/blob/e5630d537ef9d26a66dae0eeeb9a29586d641263/apps/studio/src/components/content-tab-overview.tsx#L33 and https://github.com/Automattic/studio/blob/e5630d537ef9d26a66dae0eeeb9a29586d641263/apps/studio/src/modules/sync/components/tree-view-loading-skeleton.tsx#L4

apps/studio/src/ipc-utils.ts
```diff
@@ -43,7 +43,15 @@ export interface IpcEvents {
...
 	'snapshot-success': [ { operationId: crypto.UUID } ];
 	'show-whats-new': [ void ];
-	'sync-connect-site': [ { remoteSiteId: number; studioSiteId: string; autoOpenPush?: boolean } ];
+	'sync-connect-site': [
+		{
+			remoteSiteId: number;
+			studioSiteId: string;
+			autoOpenPush?: boolean;
+			siteName?: string;
+			siteUrl?: string;
```
sejas: Do we need siteName and siteUrl? Are we planning to create a PR to pass them? If not, then I suggest removing these optional parameters from everywhere. I think they are not necessary anymore.
The URL I received while testing was: `/home/252789215?studioSiteId=bd76fedc-75fa-4c85-867f-30a996394162&autoOpenPush=true`


Image

sejas (commented):
@katinthehatsite , great work. Thanks for addressing my suggestions.
I tested it and it works much better with the loading skeleton. Thanks for improving this flow.
I left a couple of comments. Let me know what you think. I also noticed that the loading skeleton is visible even after the single site request resolves, so it seems in the end it uses the whole sites request to decide if should be loading or not. Or at least that was my impression.
I also noticed we try to use the siteurl and site title, but that information is not passed from the deep link. I think it's ok to remove those because the skeleton gives enough context IMO.

Firs time connection:

https://github.com/user-attachments/assets/475f961e-3c5e-4f9b-9d93-65700682a82b

Second time connection:

I also tested it by disconnecting and repeating the deep link.

https://github.com/user-attachments/assets/be6d6d08-0007-48ce-b3f3-a280e5b56e44

apps/studio/src/ipc-utils.ts
katinthehatsite: I agree, we can probably skip it. I added the changes in: https://github.com/Automattic/studio/pull/2375/changes/c27b6e1604b59561b85dea3733324c60dc94658a

apps/studio/src/stores/sync/wpcom-sites.ts
katinthehatsite: Great idea, adjusted in https://github.com/Automattic/studio/pull/2375/changes/3b1f58db34e9c246bb6a3e15d9434c1ef095f778

apps/studio/src/hooks/sync-sites/use-listen-deep-link-connection.ts
katinthehatsite: Moved in https://github.com/Automattic/studio/pull/2375/changes/7bce366d3cb8629701b862790627c30307f9b3bc

katinthehatsite:
> I also noticed that the loading skeleton is visible even after the single site request resolves, so it seems in the end it uses the whole sites request to decide if should be loading or not. Or at least that was my impression.

I saw this as well before but thought it was resolved. Taking a look now 👀 

Update: I see it on the reconnection but not on the initial connection 🤔

apps/studio/src/modules/sync/components/sync-sites-modal-selector.tsx
katinthehatsite: Good idea, I used Tailwind utility class in https://github.com/Automattic/studio/pull/2375/changes/93a0ee7f1be5e797977f19db43c3ebdb9a3fea22

katinthehatsite:
I made some more adjustments - let me know what you think!

sejas (commented):
@katinthehatsite , thanks for applying the suggestions.
I found a bug where Studio opens the list of sites to connect if the user has selected another site.
Steps to reproduce:
1. Click publish site for site1, and continue the flow
2. Got to studio and switch to site 2
3. Go to the browser and click open in studio which should open the modal for site1
4. Observe that the site correctly connects in the sync tab and the modal that opens shows the list of sites
5. I would expect the modal that opens is the selective sync

It works as expected if I don't switch sites before clicking open in studio.

https://github.com/user-attachments/assets/441fb992-69ac-45c3-b88e-c1188a77e787

katinthehatsite:
> Click publish site for site1, and continue the flow
Got to studio and switch to site 2
Go to the browser and click open in studio which should open the modal for site1
Observe that the site correctly connects in the sync tab and the modal that opens shows the list of sites
I would expect the modal that opens is the selective sync
It works as expected if I don't switch sites before clicking open in studio.

Thank you for testing @sejas - I made some further adjustments as well as rebased trunk. Would you be able to give it another test?

sejas (approved):
@katinthehatsite , Thanks for bringing the PR to the finish line. I tested it and I couldn't find any regression.

I tested the deeplink using a the selected site, then a different selected site, and the regular pull push buttons.

https://github.com/user-attachments/assets/b01fb4c2-8aa9-4209-982f-99a892fcae03

PR #2380, opened 2026-01-12
Refactor site watchers into unified _watch CLI command

bcotrim:
## Related issues

- Fixes STU-1193

## Proposed Changes

This PR refactors site watching into a unified architecture using a hidden `_events` CLI command that streams site events to Studio:

- **New hidden CLI command (`studio _events`)**: Streams site lifecycle events (CREATED, UPDATED, DELETED) to Studio via IPC messages
- **New events relay process (`cli/events-relay.ts`)**: PM2-managed process that watches appdata file changes and emits events
- **New shared event types (`common/lib/site-events.ts`)**: Shared schema for site events between CLI and app
- **New CLI events subscriber (`cli-events-subscriber.ts`)**: Subscribes to `_events` command output and updates SiteServer instances

**Architecture:**
```
┌─────────────────────────────────────────────────────┐
│  _events command                                    │
│  • Subscribes to PM2 process events                 │
│  • Subscribes to events-relay for appdata changes   │
└───────────────┬─────────────────────────────────────┘
                │ IPC messages (site-event)
┌───────────────▼─────────────────────────────────────┐
│  App (cli-events-subscriber.ts)                     │
│  • Handles CREATED: register new sites              │
│  • Handles UPDATED: update existing sites only      │
│  • Handles DELETED: unregister sites                │
│  • Sends IPC event to renderer                      │
└───────────────┬─────────────────────────────────────┘
                │ IPC (site-event)
┌───────────────▼─────────────────────────────────────┐
│  Renderer (use-site-details.tsx)                    │
│  • CREATED: add site to state                       │
│  • UPDATED: update existing site in state           │
│  • DELETED: remove site from state                  │
└─────────────────────────────────────────────────────┘
```

**Key improvements:**
- Single event stream replaces dual watchers (user-data-watcher + site-watch-command)
- Debouncing prevents duplicate events when multiple sources fire
- Event-specific handling prevents race conditions during site creation
- Clean signal handling (SIGINT/SIGTERM) for proper E2E test cleanup

### Site Status Changes via CLI
1. Start Studio app
2. In a separate terminal, run `studio site start <site-name>` - verify UI updates
3. Run `studio site stop <site-name>` - verify UI updates

### Site Creation via CLI
1. With app open, run `studio site create --name "CLI Test Site" --path ~/Sites/cli-test`
2. Verify the new site appears in the app UI

### Site Editing via CLI
1. Run `studio site set --name "CLI Test Site" --php-version 8.2`
2. Verify the change reflects in the app UI

### Site Deletion via CLI
1. Run `studio site delete --name "CLI Test Site"`
2. Verify the site is removed from the app UI

### App Operations
1. Start/stop sites via the app UI - verify works as before
2. Create/edit/delete sites via the app UI - verify works as before

### Edge Cases
1. Rapid successive changes are handled correctly with debouncing
2. Creating a site in Studio and CLI at the same time does not cause duplicates
3. Quitting Studio exits cleanly without stuck processes

## Pre-merge Checklist


common/lib/site-events.ts
```diff
@@ -0,0 +1,42 @@
...
+export type SiteDetails = z.infer< typeof siteDetailsSchema >;
+
+export const SITE_EVENTS = {
+	CREATED: 'site-created',
+	UPDATED: 'site-updated',
+	DELETED: 'site-deleted',
+} as const;
+
+export const siteEventSchema = z.object( {
+	event: z.string(),
```
fredrikekelund:
```suggestion
export enum SITE_EVENTS {
	CREATED = 'site-created',
	UPDATED = 'site-updated',
	DELETED = 'site-deleted',
}

export const siteEventSchema = z.object( {
	event: z.nativeEnum( SITE_EVENTS ),
```
We'd benefit from stricter types here, and this is a pretty straightforward way to achieve it.

cli/lib/pm2-manager.ts
```diff
@@ -340,3 +342,76 @@ export async function subscribePm2KillEvent( handler: () => void ) {
...
+
+/**
+ * Emit a site event via the events relay process.
+ * The relay re-emits on PM2 bus for `_events` to receive.
+ *
+ * @param event - The event topic (e.g., 'site-created', 'site-updated', 'site-deleted')
+ * @param data - The event data (must include siteId)
+ */
+export async function emitSiteEvent(
+	event: string,
```
```suggestion
	event: SITE_EVENTS,
```
Let's make the type more strict.

cli/lib/pm2-manager.ts
```diff
@@ -340,3 +342,76 @@ export async function subscribePm2KillEvent( handler: () => void ) {
...
+): Promise< void > {
+	const relayProcess = await getProcessByName( EVENTS_RELAY_PROCESS_NAME );
+	if ( ! relayProcess?.pm_id ) {
+		// No relay running - that's fine, Studio might not be open
+		return;
+	}
+
+	return new Promise( ( resolve ) => {
+		pm2.sendDataToProcessId(
+			relayProcess.pm_id as number,
```
```suggestion
	const pmID = relayProcess?.pm_id;
	if ( ! pmID ) {
		// No relay running - that's fine, Studio might not be open
		return;
	}

	return new Promise( ( resolve ) => {
		pm2.sendDataToProcessId(
			pmID,
```
Somewhat nitpicky, but this satisfies TS without the `as` operator.

cli/commands/_events.ts
```diff
@@ -0,0 +1,146 @@
...
+	await startEventsRelayProcess( relayScriptPath );
+
+	await emitAllSitesStatus();
+
+	const relayProcessName = getEventsRelayProcessName();
+	await subscribeProcessMessages( ( { processName, topic, data } ) => {
+		if ( processName !== relayProcessName ) {
+			return;
+		}
+		if ( LIFECYCLE_EVENTS.includes( topic ) ) {
```
fredrikekelund:
```suggestion
		if (
			topic === SITE_EVENTS.UPDATED ||
			topic === SITE_EVENTS.DELETED ||
			topic === SITE_EVENTS.CREATED
		) {
```
An example of how we might handle stricter types after converting `SITE_EVENTS` to an enum

cli/lib/pm2-manager.ts
```diff
@@ -340,3 +342,76 @@ export async function subscribePm2KillEvent( handler: () => void ) {
...
+				type: 'process:msg',
+				topic: event,
+				data,
+			},
+			() => {
+				resolve();
+			}
+		);
+	} );
+}
```
fredrikekelund, line 379: I understand the reason for the current strategy of using a pm2-managed `events-relay` process to be that we need a pm2 process ID to be able to use the `pm2.sendDataToProcessId` API.

I spent some time trying a different strategy: interacting directly with the pm2 "bus". The bus is an instance of `sub-socket` that comes from the `pm2-axon` [module](https://www.npmjs.com/package/pm2-axon). Unfortunately, that module is old and seemingly unmaintained… Despite multiple attempts, I couldn't manage to get a simple pub/sub pattern (listening on a Unix socket) to work. 

If it had worked, I would argue it's a better pattern, but this relay-based solution is an OK option within our current constraints, IMO.

cli/lib/pm2-manager.ts
```suggestion
	const relayProcess = await isProcessRunning( EVENTS_RELAY_PROCESS_NAME );
	if ( ! relayProcess?.pmId ) {
		// No relay running - that's fine, Studio might not be open
		return;
	}

	return new Promise( ( resolve ) => {
		pm2.sendDataToProcessId(
			relayProcess.pmId,
```
`isProcessRunning` is really a misnomer… It already does the same thing as `getProcessByName`

cli/lib/pm2-manager.ts
```diff
@@ -340,3 +342,76 @@ export async function subscribePm2KillEvent( handler: () => void ) {
...
+			if ( error ) {
+				reject( error );
+				return;
+			}
+			const proc = ( processes || [] ).find( ( p ) => p.name === name );
+			resolve( proc ? { pm_id: proc.pm_id } : null );
+		} );
+	} );
+}
+
```
```suggestion
```
`isProcessRunning` already does the same thing

cli/lib/pm2-manager.ts
```diff
@@ -340,3 +342,76 @@ export async function subscribePm2KillEvent( handler: () => void ) {
...
+export function getEventsRelayProcessName(): string {
+	return EVENTS_RELAY_PROCESS_NAME;
+}
+
+export async function startEventsRelayProcess( relayScriptPath: string ): Promise< void > {
+	const existingProcess = await getProcessByName( EVENTS_RELAY_PROCESS_NAME );
+	if ( existingProcess?.pm_id !== undefined ) {
+		return;
+	}
```
fredrikekelund, line 389:
```suggestion
export async function startEventsRelayProcess( relayScriptPath: string ): Promise< void > {
	const existingProcess = await isProcessRunning( EVENTS_RELAY_PROCESS_NAME );
	if ( existingProcess?.pmId !== undefined ) {
		return;
	}
```

fredrikekelund (commented):
Submitting my review in pieces… First batch comes here.

I spent some time experimenting with a different pattern than the events-relay pm2 process. The current pattern works, but I agree with you that it's not ideal. Using a socket as an event emitter between processes would be much nicer, but I couldn't get that to work using pm2's built-in socket…

cli/commands/_events.ts
```diff
@@ -0,0 +1,166 @@
...
+				if ( error instanceof LoggerError ) {
+					logger.reportError( error );
+				} else {
+					const loggerError = new LoggerError( __( 'Events watcher failed' ), error );
+					logger.reportError( loggerError );
+				}
+			}
+		},
+	} );
+};
```
fredrikekelund:
```suggestion
export async function commandHandler() {
	try {
		await runCommand();
	} catch ( error ) {
		if ( error instanceof LoggerError ) {
			logger.reportError( error );
		} else {
			const loggerError = new LoggerError( __( 'Events watcher failed' ), error );
			logger.reportError( loggerError );
		}
	}
}
```
This is just a recommendation, but `cli/commands/wp.ts` has already established this pattern for running top-level commands. 

It mostly just comes down to `cli/index.ts` being easier to ready with this approach, since it allows us to keep chaining `.command()` calls, and all top-level command names are visible in that file.

On the other hand, the `registerCommand` pattern is also well established, so I could be swayed in that direction, too…

cli/index.ts
```diff
@@ -109,6 +110,8 @@ async function main() {
 		.demandCommand( 1, __( 'You must provide a valid command' ) )
 		.strict();
 
+	registerEventsCommand( studioArgv );
+
 	await studioArgv.argv;
```
fredrikekelund, line 118:
```suggestion
		.command( {
			command: '_events',
			describe: false, // Hidden command
			handler: _eventsCommandHandler,
		} )
		.demandCommand( 1, __( 'You must provide a valid command' ) )
		.strict();

	await studioArgv.argv;
```
Per my recommendation of exporting a `commandHandler` function from `cli/commands/_events` instead of a `registerCommand` callback.

cli/lib/wordpress-server-manager.ts
```diff
@@ -450,7 +451,8 @@ export async function subscribeSiteEvents(
 	let debounceTimeout: NodeJS.Timeout | null = null;
```
fredrikekelund: It looks like we aren't using the `debounceMs` option anywhere. Can we remove the debounce logic and simplify `invokeHandler`?

src/tests/index.test.ts
fredrikekelund, line 1: It looks like we can undo this change. The test still passes with the old code when I try it locally.

src/modules/cli/lib/cli-events-subscriber.ts
```diff
@@ -0,0 +1,123 @@
...
+		running,
+	};
+}
+
+async function handleSiteEvent( event: SiteEvent ): Promise< void > {
+	const { event: eventType, siteId, site, running } = event;
+	const previous = pendingUpdates.get( siteId ) ?? Promise.resolve();
+	const current = previous
+		.catch( () => {} )
+		.then( () => {
```
fredrikekelund:
```suggestion
const handleSiteEvent = sequential( async ( event: SiteEvent ): Promise< void > => {
	const { event: eventType, siteId, site, running } = event;
```
The `common/lib/sequential.ts` helper could be used here. I know the current logic sets up separate queues for each site, but since the event handler isn't actually executing any async logic, we still achieve the same effect with the non-partitioned `sequential` queue.

src/modules/cli/lib/cli-events-subscriber.ts
```diff
@@ -0,0 +1,123 @@
...
+		} );
+	} );
+}
+
+export function stopCliEventsSubscriber(): void {
+	if ( subscriber ) {
+		const [ , childProcess ] = subscriber;
+		if ( childProcess.connected ) {
+			childProcess.disconnect();
+		}
```
fredrikekelund, line 113:
```suggestion
```
We don't need this.

src/index.ts
```diff
@@ -513,7 +515,7 @@ async function appBoot() {
 	app.on( 'will-quit', ( event ) => {
 		globalShortcut.unregisterAll();
 		stopUserDataWatcher();
-		stopSiteWatcher();
+		void stopCliEventsSubscriber();
```
fredrikekelund:
```suggestion
		stopCliEventsSubscriber();
```
It's not an async function

fredrikekelund (commented):
I'm nearly done with the code review

cli/commands/_events.ts
```diff
@@ -0,0 +1,166 @@
...
+		}
+	}
+
+	const timeout = setTimeout( () => {
+		pendingEmits.delete( siteId );
+		void doEmitSiteEvent( effectiveEvent, siteId );
+	}, DEBOUNCE_MS );
+
+	pendingEmits.set( siteId, { event: effectiveEvent, timeout } );
+}
```
fredrikekelund: Do we really need to be this clever..? Can't we just use the `common/lib/sequential.ts` helper to ensure that events are sent in the order they're received?

fredrikekelund (approved):
Great work on this 👍 The `emitSiteEvent` API is easily understandable, and it'll be easy to extend in the future if we add more events.

I still want to land https://github.com/Automattic/studio/pull/2398 so that the communication doesn't break is users run `site stop --all` outside Studio, but we agreed to land this PR first, to unblock https://github.com/Automattic/studio/pull/2281

PR #2382, opened 2026-01-13
Studio: Testing pnpm with hoisted mode

katinthehatsite:
## Related issues

Related to STU-1011

## Proposed Changes

This PR is meant as an assessment / experiment to see if we can make `node-linker=hoisted` mode with pnpm to speed up the installation steps for Windows could be sped up. Using `node-linker=hoisted` would make `pnpm` because in a more flat way, similarly to `npm` when installing node modules which should prevent `pnpm` from changing the structure of `node_modules` folder and prevent from breaking current behavior of copying `cli/node_modules`.

## Testing Instructions

* We can observe if the installation steps for the Windows build are faster compared to PRs without these changes. 

## Pre-merge Checklist


katinthehatsite:
With the `node-linker=hoisted` mode, for MacOs, the installation steps took 21s:

Screenshot 2026-01-14 at 10 38 26 AM

With `npm`, it took 33s:

Screenshot 2026-01-14 at 10 39 37 AM

For Windows, with `node-linker=hoisted` mode for `pnpm`, the installation took 2m 3s + 53s for restoring cache:

Screenshot 2026-01-14 at 10 41 43 AM

The results from `npm` are 3m 11s + 11s for cache:

Screenshot 2026-01-14 at 10 42 30 AM

If we set cache aside, we can see some improvement on Windows, almost by a minute. However, there are a lot of other changes to make to make sure that we are able to produce a successful build. I managed to work it out for macOS but not for Windows yet. It feels though like a half-baked solution. I think our best approach would be to do a proper refactor for `pnpm` usage instead of going for the middle ground without seeing significant improvement.

katinthehatsite:
I will close this as this was meant as an experiment to see if the hoisted mode for node_linker will bring the performance improvement for Windows. We will not be implementing it in this mode as per Linear discussion.

PR #2389, opened 2026-01-13
Fix Blueprint deeplink `add-site-with-blueprint` IPC listener registration

ivan-ottinger:
## Related issues

- Fixes STU-1205

## Proposed Changes

- move `useBlueprintDeeplink` call from `useAddSite` to `AddSiteModal` (and `NoStudioSites`)
  -  this ensures the `add-site-with-blueprint` IPC listener is registered where it is needed - even when the _Add site_ modal is not open
- add new unit test covering `useBlueprintDeeplink`
- simplify logic in `useBlueprintDeeplink`
- extract and reuse `BlueprintPreferredVersions` type

## Testing Instructions

2. Remove all sites from Studio.
3. Try opening Blueprint deeplinks, for instance from here: https://jsbin.com/levaburate/2/edit?html,css,output or https://developer.wordpress.com/docs/developer-tools/studio/blueprints/open-in-wordpress-studio-button/.
4. The deeplinks should work as expected and let you create a site with related Blueprint.
5. Repeat the step 3, but now with at least one existing Studio site.
6. Again, the deellinks should work as expected.
7. There should be no unexpected regressions related to the _Add site_ modal. 
8. The `npm test -- src/modules/add-site/hooks/tests/use-blueprint-deeplink.test.tsx` test should pass.

## Pre-merge Checklist


src/modules/add-site/hooks/use-blueprint-deeplink.ts
```diff
@@ -3,118 +3,98 @@ import { useCallback } from 'react';
...
-			}
-		},
-		[ setBlueprintPreferredVersions, setPhpVersion, setWpVersion ]
-	);
+				try {
+					const blueprintJson = await getIpcApi().readBlueprintFile( blueprintPath );
+					const fileName = blueprintPath.split( /[/\\]/ ).pop() || 'blueprint.json';
+					const blueprintMeta = blueprintJson.meta as
+						| { title?: string; description?: string }
+						| undefined;
```
katinthehatsite: I am wondering if it makes sense to keep `type BlueprintMetadata` in the first place instead of having to specify types here. It is not a strong preference though

src/modules/add-site/hooks/use-blueprint-deeplink.ts
```diff
@@ -3,118 +3,98 @@ import { useCallback } from 'react';
...
-				const fileName = blueprintPath.split( /[/\\]/ ).pop() || 'blueprint.json';
-				const fileBlueprint = createBlueprintFromData(
-					blueprintJson,
-					`file:${ fileName }`,
-					fileName.replace( '.json', '' ),
-					__( 'Blueprint loaded from URL' )
-				);
+					if ( blueprintJson.preferredVersions ) {
+						const preferredVersions = blueprintJson.preferredVersions as {
+							php?: string;
```
katinthehatsite: I am wondering if it makes sense to move those to a type too similar to `type BlueprintMetadata` 🤔

katinthehatsite (approved):
The changes look good to me, thanks for fixing the regression from my PR 🙇  (I sort of half-expected there to be one because of how many things in touched 😅). I did not spot any issues in my testing 👍 

I left two minor suggestions but I don't think those are blockers in any way so approving.

ivan-ottinger:
> The changes look good to me, thanks for fixing the regression from my PR 🙇 (I sort of half-expected there to be one because of how many things in touched 😅).

No worries, at all! 🙂 That was not an obvious regression! Thanks for the review, Kat.

src/modules/add-site/hooks/use-blueprint-deeplink.ts
ivan-ottinger: Good point. I agree it is more readable if we keep the type separate. I have made the change in https://github.com/Automattic/studio/pull/2389/commits/d984a3ba93021ac29c1ad7e7c737b14267df7b19.

src/modules/add-site/hooks/use-blueprint-deeplink.ts
ivan-ottinger: Good catch! Turns out this type is used multiple times throughout the codebase, so I have extracted it out and reused in all places: https://github.com/Automattic/studio/pull/2389/commits/a303a03fe1795dc73d49b2188089c1de08c1d18a.

PR #2393, opened 2026-01-14
Studio: Fix error connecting to external MySQL database

gavande1:
## Related issues

- Fixes STU-1203

## Proposed Changes

- I propose to use `do-not-attempt-installing` WordPress installation mode when site is configured to use MySQL. 
- Otherwise use `install-from-existing-files-if-needed` as it was configured before.
- The MySQL config is checked using existence of sqlite plugin directory and `db.php`. 

## Why

The actual error is originating from [this line](https://github.com/WordPress/wordpress-playground/blob/fffde71af232231749175e6cb2ebf9d29ed85efb/packages/playground/wordpress/src/boot.ts#L360) in playground CLI. Its part of `assertValidDatabaseConnection` function where some pre-test are done to make sure sqlite is configured correctly. This validation only runs if either `install-from-existing-files-if-needed` or `download-and-install` installation mode is used. You can check condition [here](https://github.com/WordPress/wordpress-playground/blob/fffde71af232231749175e6cb2ebf9d29ed85efb/packages/playground/wordpress/src/boot.ts#L360). This PR updates WordPress installation mode based on sqlite plugin status to avoid these checks.

## Testing Instructions

1. Create a new site or open an existing one in WordPress Studio
2. Configure `wp-config.php` to use an external MySQL database (`127.0.0.1`). Follow this guide to [use Studio with MySQL server](https://developer.wordpress.com/docs/developer-tools/studio/frequently-asked-questions/#use-studio-with-mysql-server).
3. Start the server - it should attempt to connect to MySQL and server should start normally.
4. Verify the if site is using MySQL database. Goto `Tools -> Site heath -> Info -> Database`
CleanShot 2026-01-14 at 17 22 17@2x 
5. Check for regression. Existing sites should start normally. 

## Pre-merge Checklist

### DIFF
--- a/common/lib/sqlite-integration.ts
+++ b/common/lib/sqlite-integration.ts
@@ -64,4 +64,12 @@ export abstract class SqliteIntegrationProvider {
 			await this.installSqliteIntegration( sitePath );
 		}
 	}
+
+	async isSqliteInstalled( sitePath: string ): Promise< boolean > {
+		return (
+			fs.existsSync(
+				path.join( sitePath, 'wp-content', 'mu-plugins', this.getSqliteFilename() )
+			) && fs.existsSync( path.join( sitePath, 'wp-content', 'db.php' ) )
+		);
+	}
 }
--- a/common/lib/sqlite-integration.ts
+++ b/common/lib/sqlite-integration.ts
@@ -67,9 +67,8 @@ export abstract class SqliteIntegrationProvider {
 
 	async isSqliteInstalled( sitePath: string ): Promise< boolean > {
 		return (
-			fs.existsSync(
-				path.join( sitePath, 'wp-content', 'mu-plugins', this.getSqliteFilename() )
-			) && fs.existsSync( path.join( sitePath, 'wp-content', 'db.php' ) )
+			fs.existsSync( path.join( sitePath, 'wp-content', 'mu-plugins', this.getSqliteDirname() ) ) &&
+			fs.existsSync( path.join( sitePath, 'wp-content', 'db.php' ) )
 		);
 	}
 }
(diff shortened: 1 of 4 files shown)


epeicher (approved):
Thanks @gavande1 for solving this! I have tested it, and I can now connect to MySql database, while I have an error on trunk. Changes LGTM! :shipit: 

Image

common/lib/sqlite-integration.ts
```diff
@@ -66,4 +66,12 @@ export abstract class SqliteIntegrationProvider {
 			await this.installSqliteIntegration( sitePath );
 		}
 	}
+
+	async isSqliteInstalled( sitePath: string ): Promise< boolean > {
```
gavande1, line 68: I believe the `needsSqliteSetup` function does a similar think but in the opposite direction. It checks whether `wp-config.php` exists. In this case, we need that file to be present. So, I thought it would be best to create a separate function with the required functionality.

src/lib/wordpress-provider/playground-cli/playground-cli-provider.ts
```diff
@@ -79,14 +80,20 @@ export class PlaygroundCliProvider implements WordPressProvider {
 			console.log( '[PlaygroundCliProvider] Xdebug support is enabled for this site' );
 		}
 
+		let wordpressInstallMode: WordPressInstallMode = 'download-and-install';
```
wojtekn: What if we extracted the entire series of conditions to separate a function that uses early returns, making it more readable and easier to understand?

gavande1: Thanks for suggestion @wojtekn. It made sense to me. I have implemented your suggestion in [658a445](https://github.com/Automattic/studio/pull/2393/commits/658a445bca612c134d127256c69ccccb3691eeef).

wojtekn: Thanks, it's much cleaner now.

wojtekn (approved):

wojtekn:
@gavande1 I merged trunk to this branch to run tests with the cli i2 changes.

PR #2394, opened 2026-01-14
Studio: Update zod to v4

katinthehatsite:
## Related issues

Fixes STU-517

## Proposed Changes

This PR updates to Zod v4 and resolves incompatibility issues. 

## Testing Instructions

* Confirm that all the checks pass (I also triggered builds for this just in case and everything went through)
* Visual review should suffice
* Additionally, pull the changes from this branch, start the app with `npm start`, test the app holistically, especially around actions that make API requests, and confirm there are no errors in the console

## Pre-merge Checklist


katinthehatsite:
For reference on changes: https://zod.dev/v4/changelog (if you want to double-check things)

cli/lib/api.ts
```diff
@@ -18,7 +18,7 @@ const createSiteResponseSchema = z.object( {
 
 const statusResponseSchema = z.object( {
 	status: z.enum( [ SnapshotStatus.Pending, SnapshotStatus.Processing, SnapshotStatus.Active ], {
-		errorMap: () => ( { message: __( 'Invalid site status' ) } ),
+		message: __( 'Invalid site status' ),
```
gavande1: It seems `message` property is [deprecated](https://zod.dev/v4/changelog?id=deprecates-message-parameter). Should we update all instance of `message` with `error`? The message field is still usable but if we don't do it now then future versions will get complex to migrate IMO.

katinthehatsite: Let me take a look 👀 I was relying on my linter to indicate deprecation but it did not seem to catch this one

katinthehatsite: Updated in https://github.com/Automattic/studio/pull/2394/changes/aed8da1ae03eda0f8c659aadb3982903c4aee6ca

cli/lib/api.ts
```diff
@@ -18,7 +18,7 @@ const createSiteResponseSchema = z.object( {
 
 const statusResponseSchema = z.object( {
 	status: z.enum( [ SnapshotStatus.Pending, SnapshotStatus.Processing, SnapshotStatus.Active ], {
-		errorMap: () => ( { message: __( 'Invalid site status' ) } ),
+		invalid_type_error: __( 'Invalid site status' ),
```
epeicher: The `invalid_type_error` has been dropped [according to the docs](https://zod.dev/v4/changelog?id=drops-invalid_type_error-and-required_error). I think this needs to change to `error`. I also think that should fix the errors in the Unit Tests

epeicher (commented):
Thanks for upgrading this, Kat! I’ve done a smoke test on some parts of the application and haven’t found any issues, including nothing in the DevTools console.

I’ve left a comment about one change required for the new version, which should also fix the unit tests. Once that’s updated, this should be good to go.

cli/lib/api.ts
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2394/changes/8cfeadefee4841e53ca388061626529b320a7f64

Thank you for the review 🙇  I am not sure why my linter is not catching things, usually it indicates deprecated props and parameters. i will take a look into fixing that. In any case, it should now be adjusted.

ivan-ottinger:
> For reference on changes: https://zod.dev/v4/changelog (if you want to double-check things)

Thanks for linking the changelog link, Kat!

I have noticed that we may also want to update the `.merge` in `cli/lib/types/wordpress-server-ipc.ts`:

```diff
@@ -44,9 +44,9 @@ export type ManagerMessagePayload = z.infer< typeof _managerMessagePayloadSchem
 
 const managerMessageBase = z.object( { messageId: z.number() } );
 export const managerMessageSchema = z.discriminatedUnion( 'topic', [
-	managerMessageBase.merge( managerMessageStartServer ),
-	managerMessageBase.merge( managerMessageRunBlueprint ),
-	managerMessageBase.merge( managerMessageStopServer ),
+	managerMessageBase.extend( managerMessageStartServer.shape ),
+	managerMessageBase.extend( managerMessageRunBlueprint.shape ),
+	managerMessageBase.extend( managerMessageStopServer.shape ),
 ] );
 export type ManagerMessage = z.infer< typeof managerMessageSchema >;
```

also the following change in `src/stores/wpcom-api.ts`:

```diff
@@ -119,7 +119,7 @@ const wpcomPublicBaseQuery: BaseQueryFn<
 	}
 };
 
-function parseResponse< TSchema extends z.ZodTypeAny >(
+function parseResponse< TSchema extends z.ZodType >(
 	response: unknown,
 	schema: TSchema
 ): z.infer< TSchema > {
```

cli/lib/api.ts
epeicher: Thanks for updating it!

> I am not sure why my linter is not catching things, usually it indicates deprecated props and parameters

For me, it was being reported by the IDE, but not when running `npm run lint`, so I think there could be a configuration issue 🤔

katinthehatsite:
> I have noticed that we may also want to update the .merge in cli/lib/types/wordpress-server-ipc.ts:

I applied your suggestions in https://github.com/Automattic/studio/pull/2394/changes/6848bf1fc82d4d7210c8e8b4b01284e4b18a6f34 - please let me know what you think 🙇

ivan-ottinger (approved):
> I applied your suggestions in https://github.com/Automattic/studio/commit/6848bf1fc82d4d7210c8e8b4b01284e4b18a6f34 - please let me know what you think 🙇

Looks good! Thanks for the update, Kat. I did not observe any other changes that would be missing.

cli/lib/appdata.ts
```diff
@@ -44,14 +43,12 @@ const userDataSchema = z
 				email: z.string(),
 				displayName: z.string().default( '' ),
 			} )
-			.passthrough()
-			.optional(),
-		lastBumpStats: z
-			.record( z.string(), z.record( z.nativeEnum( StatsMetric ), z.number() ) )
+			.loose()
 			.optional(),
+		lastBumpStats: z.record( z.string(), z.record( z.string(), z.number() ) ).optional(),
```
fredrikekelund, line 48: @katinthehatsite, how come the `StatsMetric` enum isn't included in the new schema?

katinthehatsite: I think it was mostly something that I overlooked during the update. I added a follow-up PR here: https://github.com/Automattic/studio/pull/2417 . I also noticed that there were more updates needed after the CLI dev branch was merged so I addressed these changes there as well.

katinthehatsite: Actually I looked into it a bit more and ended up removing them even in the follow-up PR. Initially, I kept them but it required more code that seemed unnecessary. I think it is okay to remove it, and it feels like a more pragmatic choice.  TypeScript already enforces type safety at compile-time where the data is created. Zod will validate the data structure (nested records with numbers) in this case and does not necessarily need to duplicate TypeScript's business logic validation of enum values.

fredrikekelund: See https://github.com/Automattic/studio/pull/2417#issuecomment-3771802794

katinthehatsite: I tried that option 😅 but ultimately, it did not seem to be the best one. I added some details here: https://github.com/Automattic/studio/pull/2417#issuecomment-3771854264

PR #2397, opened 2026-01-15
Blueprints: Update Blueprint deeplink step design

ivan-ottinger:
## Related issues

- Resolves STU-1151

## Proposed Changes

- update Blueprint deeplink step design

| Before | After |
|--------|--------|
| CleanShot 2026-01-20 at 10 33 54@2x | 538092132-a9c2b72f-923f-4a67-a274-79d8ad583572 | 

ℹ️ Please note there are several differences compared to the design proposed in Figma (RToz6tIuQ7nlZrikBte4GU-fi-10353_7411), for example the check icon is blue in this PR to match existing design of checkmarks we already have in Studio.

## Testing Instructions

2. Try using different Blueprint deeplinks, for example the ones in: https://output.jsbin.com/kulaqijega or use https://developer.wordpress.com/docs/developer-tools/studio/blueprints/open-in-wordpress-studio-button/.
3. The design should look good in both LTR and RTL; there should be no regressions.

## Pre-merge Checklist

### DIFF
--- a/src/modules/add-site/components/blueprint-deeplink.tsx
+++ b/src/modules/add-site/components/blueprint-deeplink.tsx
@@ -35,7 +35,7 @@ export default function BlueprintDeeplink( {
 				fileName={ blueprintTitle }
 				className="w-full max-w-4xl mx-auto"
 			/>
-			<VStack className="max-w-[400px] min-w-[250px] mx-auto mt-16" spacing={ 0 }>
+			<VStack className="max-w-[400px] min-w-[350px] mx-auto mt-16 justify-start" spacing={ 0 }>
 				<VStack className="w-full max-h-[172px] p-6 border rounded-t-lg border-gray-200">
 					<HStack className="h-full" alignment="top" spacing={ 4 }>
 						<Icon className="fill-a8c-blue-50 shrink-0" icon={ check } size={ 29 } />
--- a/src/modules/add-site/components/blueprint-deeplink.tsx
+++ b/src/modules/add-site/components/blueprint-deeplink.tsx
@@ -40,13 +40,8 @@ export default function BlueprintDeeplink( {
 					<HStack className="h-full justify-start" alignment="top" spacing={ 4 }>
 						<Icon className="fill-a8c-blue-50 shrink-0" icon={ check } size={ 29 } />
 						<VStack alignment="left" spacing={ 1 }>
-							<Text className="text-lg font-medium text-gray-900">
-								{ __( 'Blueprint selected' ) }
-							</Text>
+							<Text className="text-lg font-medium text-gray-900">{ blueprintTitle }</Text>
 							<VStack alignment="left" spacing={ 4 }>
-								<Text className="text-[13px] text-[#50575E]" weight={ 400 }>
-									{ blueprintTitle }
-								</Text>
 								{ blueprintDescription && (
 									<Text
 										className="text-[13px] text-a8c-gray-700 line-clamp-3"
--- a/src/modules/add-site/components/blueprint-deeplink.tsx
+++ b/src/modules/add-site/components/blueprint-deeplink.tsx
@@ -62,7 +62,7 @@ export default function BlueprintDeeplink( {
 				</VStack>
 				<HStack
 					alignment="start"
-					className="w-full bg-[#FAFAFA] border border-t-0 rounded-b-lg p-[14px]"
+					className="w-full bg-[#FAFAFA] border border-t-0 rounded-b-lg py-[14px] pe-[14px] ps-[68px] justify-start"
 				>
 					<Icon icon={ link } size={ 18 } />
 					<Text className="text-[13px] text-[#2F2F2F]">{ __( 'Blueprint loaded from URL' ) }</Text>
--- a/src/modules/add-site/components/blueprint-deeplink.tsx
+++ b/src/modules/add-site/components/blueprint-deeplink.tsx
@@ -65,7 +65,9 @@ export default function BlueprintDeeplink( {
 					className="w-full bg-[#FAFAFA] border border-t-0 rounded-b-lg py-[14px] pe-[14px] ps-[68px] justify-start"
 				>
 					<Icon icon={ link } size={ 18 } />
-					<Text className="text-[13px] text-[#2F2F2F]">{ __( 'Blueprint loaded from URL' ) }</Text>
+					<Text className="text-[13px] text-a8c-gray-800">
+						{ __( 'Blueprint loaded from URL' ) }
+					</Text>
 				</HStack>
 			</VStack>
 		</VStack>
--- a/src/modules/add-site/components/blueprint-deeplink.tsx
+++ b/src/modules/add-site/components/blueprint-deeplink.tsx
@@ -22,7 +22,7 @@ export default function BlueprintDeeplink( {
 }: BlueprintDeeplinkProps ) {
 	const { __ } = useI18n();
 
-	const blueprintTitle = selectedBlueprint?.title || '';
+	const blueprintTitle = selectedBlueprint?.title || __( 'Blueprint loaded from URL' );
 	const blueprintDescription = selectedBlueprint?.excerpt || '';
 
 	return (
@@ -46,7 +46,7 @@ export default function BlueprintDeeplink( {
 						</Text>
 						<VStack alignment="left" spacing={ 4 }>
 							<Text className="text-[13px] text-[#50575E]" weight={ 400 }>
-								{ blueprintTitle || __( 'Blueprint loaded from URL' ) }
+								{ blueprintTitle }
 							</Text>
 							{ blueprintDescription && (
 								<Text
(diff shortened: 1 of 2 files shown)


wojtekn:
I tested Blueprints from the JS Bin, and it worked for the ideal case. For my Blueprint, it was a bit empty:

CleanShot 2026-01-15 at 16 21 56@2x

If Blueprint doesn't have a meta title and description, shouldn't we display some details about the Blueprint content, such as "Blueprint installs X plugins, Y themes, and runs Z blocks of PHP code"?

As regards layout, should we add "Blueprint loaded from URL", footer too? See p9Jlb4-gnb-p2

It doesn't seem to be part of this PR, but invalid Blueprints could benefit from a better UX. For now, I've got this:

|**1**|**2**|
|---|---|
|CleanShot 2026-01-15 at 16 19 45@2x|CleanShot 2026-01-15 at 16 19 59@2x|

ivan-ottinger:
Thanks for the review, Wojtek.

> If Blueprint doesn't have a meta title and description, shouldn't we display some details about the Blueprint content, such as "Blueprint installs X plugins, Y themes, and runs Z blocks of PHP code"?

Do you mean to parse the contents of the Blueprint JSON and use it to generate a message like the above? I think that could work nicely until we have a dedicated way to output and let the user manage the contents of the Blueprint we discussed earlier.

> As regards layout, should we add "Blueprint loaded from URL", footer too? See p9Jlb4-gnb-p2

This is something that feels redundant to me - as all Blueprints there are loaded from a deeplink URL (whether the JSON is included in the deeplink or referred from a different source). I have asked Marina earlier today about it in Figma: RToz6tIuQ7nlZrikBte4GU-fi-10352_72525#1587850255.

I am happy to add the footer there. If we implement the "Blueprint installs X plugins, Y themes, and runs Z blocks of PHP code" part for Blueprints that don't have any title/description, it may not feel that redundant.

> It doesn't seem to be part of this PR, but invalid Blueprints could benefit from a better UX. For now, I've got this:

Good idea. I have created a separate task to look into that: https://linear.app/a8c/issue/STU-1209/improve-the-display-of-blueprint-deeplink-errors

src/modules/add-site/components/blueprint-deeplink.tsx
```diff
@@ -33,23 +34,33 @@ export default function BlueprintDeeplink( {
...
-					<Text className="text-xl font-medium text-gray-900">{ __( 'Blueprint selected' ) }</Text>
-					<Text className="text-base font-medium text-gray-900 max-w-md px-4" weight={ 400 }>
-						{ blueprintTitle }
-					</Text>
-					{ blueprintDescription && (
-						<Text className="text-sm text-gray-600 max-w-md px-4 line-clamp-2">
-							{ blueprintDescription }
+			<VStack className="max-w-[400px] min-w-[250px] max-h-[172px] mx-auto p-6 border rounded-lg border-gray-200 mt-16">
+				<HStack className="h-full" alignment="top" spacing={ 4 }>
+					<Icon className="fill-a8c-blue-50 shrink-0" icon={ check } size={ 29 } />
```
katinthehatsite: I think it is a good idea to keep the consistency here with what we have in Studio even if it might be a bit different from design 👍

katinthehatsite (commented):
I did not mind that there was not a lot of information about the blueprint but to me, the spacing was slightly off when it was a shorter version:

Image

I am wondering if maybe moving the title closer to the icon would bring a better visual balance for this case. 

Alternatively, it wouldbe solved if it showed more content.

wojtekn:
> Do you mean to parse the contents of the Blueprint JSON and use it to generate a message like the above? I think that could work nicely until we have a dedicated way to output and let the user manage the contents of the Blueprint we discussed earlier.

Yes, but we can add that as a follow-up.

> This is something that feels redundant to me - as all Blueprints there are loaded from a deeplink URL (whether the JSON is included in the deeplink or referred from a different source). 

As this step will also allow the selection of a file from the user's computer (as in the import step), the footer will differentiate between the two cases: a file loaded from a URL and a file selected from the local system.

Also, I'm unsure if we need "Blueprint selected" text - shouldn't this text be the Blueprint title? Then we could have:

> Title (from Blueprint meta field, or just "Blueprint" if it's not set)
> Description (from Blueprint meta field, or generated one as mentioned above. Can be empty for now)
> Footer: "Blueprint loaded from URL" if it was 

Possible variations:

> Car service site
> Blueprint installs the car service theme and a few plugins that help with scheduling visits.
> [Blueprint loaded from URL]

> Car service site
> Blueprint installs X plugins, Y themes, and runs Z blocks of PHP code.
> [Blueprint loaded from URL]

> Blueprint
> Blueprint installs X plugins, Y themes, and runs Z blocks of PHP code.
> [Blueprint loaded from URL]

> Car service site
> Blueprint installs the car service theme and a few plugins that help with scheduling visits.

ivan-ottinger:
Thank you for your review and suggestions, Kat & Wojtek! I have made some of the changes and will continue next week. I have set the PR back to draft and will let you know when it will be ready for review agian.

ivan-ottinger:
I have updated the UI and also included the `Blueprint loaded from URL` message:

CleanShot 2026-01-20 at 12 10 28@2x

CleanShot 2026-01-20 at 10 42 40@2x

> Car service site
Blueprint installs X plugins, Y themes, and runs Z blocks of PHP code.
[Blueprint loaded from URL]

For this case where the description isn't available I have created a dedicated task for this change: STU-1224.

src/modules/add-site/components/blueprint-deeplink.tsx
```diff
@@ -29,27 +30,46 @@ export default function BlueprintDeeplink( {
...
-				className="w-full max-w-4xl mx-auto mb-6"
+				className="w-full max-w-4xl mx-auto"
 			/>
-
-			<div className="w-full max-w-[400px] h-[250px] mx-auto p-12 border-2 rounded-xl border-gray-300 bg-gray-50">
-				<VStack className="items-center justify-center h-full" spacing={ 2 }>
-					<Icon icon={ check } size={ 24 } className="text-green-600" />
-					<Text className="text-xl font-medium text-gray-900">{ __( 'Blueprint selected' ) }</Text>
-					<Text className="text-base font-medium text-gray-900 max-w-md px-4" weight={ 400 }>
-						{ blueprintTitle }
```
wojtekn, line 44: I'm unsure if this came up before, but should we filter/sanitize that data?

wojtekn (commented):
Looks good.

Do we need a "Blueprint selected" header for anything? Or could we replace this header content with `blueprintTitle` value?

Also, could we update jsbin with content that allows us to easily test all possible cases?

ivan-ottinger:
> Do we need a "Blueprint selected" header for anything? Or could we replace this header content with blueprintTitle value?

I like it there - in relation to the checkmark icon on the side. It is based on the Figma design. I am also open to replacing it with Blueprint title. @Marinatsu what do you think? Here's how it can look like:

CleanShot 2026-01-20 at 16 24 04@2x

---

> Also, could we update jsbin with content that allows us to easily test all possible cases?

Yes, we can update that to have more possibilities for testing in one place.

wojtekn:
@ivan-ottinger I think it's much better without that static title. It's cleaner, closer to the import UI, and thanks to "Start from Blueprint" screen header it's still clear where user lands:

CleanShot 2026-01-20 at 16 59 24@2x

ivan-ottinger:
> @ivan-ottinger I think it's much better without that static title. It's cleaner, closer to the import UI, and thanks to "Start from Blueprint" screen header it's still clear where user lands:

Sounds good. 👍🏼 I have removed the extra text in https://github.com/Automattic/studio/pull/2397/commits/e22f50893b4eadbf1243f6483f1a7c1a29fd5636.

src/modules/add-site/components/blueprint-deeplink.tsx
ivan-ottinger: I would say that in this case a clamp on the length should be enough. I have added it in https://github.com/Automattic/studio/pull/2397/commits/2b244d35393c3539f84855a860c9e9e7fe8ffa2c.

The Blueprint JSON also gets validated before we get title and description out of it.

ivan-ottinger:
> Yes, we can update that to have more possibilities for testing in one place.

I have added more testing deeplinks in https://output.jsbin.com/kulaqijega to test this PR more easily. We can extend that later for the PR where we add parsing for JSON contents and could also consider a different place than jsbin for those test deeplinks. 🙂

src/modules/add-site/components/blueprint-deeplink.tsx
wojtekn: Yep, React escapes any tags included in those, but we could consider some sanitization, too.

wojtekn:
All cases look great. Thanks for updating jsbin - let's use this one for subsequent PRs that target this feature.

wojtekn (approved):

PR #2398, opened 2026-01-15
CLI: Send site events over socket instead of through events-relay pm2 process

fredrikekelund:
## Related issues

- Related to #2380
- Related to STU-1193

## Proposed Changes

Use the `pm2-axon` library to send events over a socket/named pipe instead of using the events-relay child process. The benefit of this approach is that it doesn't break if users run `studio site stop --all` from their terminal while Studio is also running. 

> [!TIP]  
> Now that we use a socket, we should start thinking about whether there's a future scenario where Studio can be responsible for binding a server on the socket, and events could be sent directly from CLI processes to Studio. 

## Testing Instructions

1. `npm start`
2. Create a site
3. Ensure that it shows up in the sidebar
4. Leave Studio running in the background and open a new terminal window
5. Run `node dist/cli/main.js site stop --all`
6. Ensure that all sites are marked as stopped in Studio
7. Run `node dist/cli/main.js site start --path PATH_TO_SITE` where `PATH_TO_SITE` is the path to a Studio site
8. Ensure that the site is marked as running in Studio

## Pre-merge Checklist


fredrikekelund:
Status update: we've decided to follow this strategy, but we are currently trying to figure out why the E2E tests fail on macOS (on CI)

src/modules/cli/lib/cli-events-subscriber.ts
```diff
@@ -108,10 +110,9 @@ export async function startCliEventsSubscriber(): Promise< void > {
 export function stopCliEventsSubscriber(): void {
 	if ( subscriber ) {
 		const [ , childProcess ] = subscriber;
-		if ( childProcess.connected ) {
-			childProcess.disconnect();
-		}
-		childProcess.kill();
+		const pid = childProcess.pid;
+		const result = childProcess.kill( 'SIGKILL' );
```
fredrikekelund: I was battling failing E2E tests on macOS today and yesterday. The reason they were failing was that the first test suite (`e2e/app.test.ts`) runs very quickly. So quickly, in fact, that the events subscriber child process only ran for 25ms or so before `stopCliEventsSubscriber` was called. 

The `_events` command wouldn't even have time to register SIGINT/SIGTERM event listeners in that time, which meant that the process would not be killed. I initially tried to counteract this by hardening the cleanup logic in the `runCommand` function, not running any async IO before registering the SIGINT/SIGTERM listeners, etc.

Even after I did all that, the tests still failed the same way. This led me to the conclusion that the SIGTERM signal (which is what `ChildProcess::kill` sends by default on Posix platforms in node) was sent before `runCommand` was even triggered, while some other async work was happening. It could, for instance, be the translation loading in `cli/index.ts`, or something that yargs does internally. I guess it could potentially even be an issue with multiple cores executing logic in parallel…

In any case, SIGKILL signals don't allow the child process to perform any cleanup; it's essentially a force kill operation. pm2-axon has logic that gracefully binds to a socket if the process that originally bound to it is no longer running, which is why it's not critical to let the `_events` child process run the cleanup. 

The E2E tests are passing now, and I think this is a pragmatic solution to our problem 👍

bcotrim (commented):
Still reviewing the code but I noticed some deprecation warnings in the CLI now:

Image

Image

fredrikekelund:
The performance tests are working, at least 🙂

bcotrim (approved):
Code LGTM 👍 
Didn't find any regressions

fredrikekelund:
Thanks, @bcotrim. I just pushed a fix to mute the deprecation warning, too

PR #2399, opened 2026-01-15
Attempt to fix test flakiness

iangmaia:
Fixes AINFRA-1760

## Summary

Note: this was mostly a Claude Code session after I noticed the tests flakiness on https://github.com/Automattic/studio/pull/2383 to get them to a more stable situation. The build succeeded in the last few runs, so 🤞 

### E2E Test Fixes
- Add `retries: 1` to main `playwright.config.ts` for flaky test handling
- Add `actionTimeout: 30s` to prevent hanging on blocked elements
- Increase default `expect.timeout` to 30s
- **Fix missing `await`** in `metrics/tests/site-editor.test.ts` (was causing persistent metrics test failures)

## Test plan
- [x] Run unit tests in CI
- [x] Run E2E tests to verify changes don't cause false failures
- [x] Run metrics tests to verify the `await` fix resolves persistent failures

### Successful builds
The last [builds in this branch](https://buildkite.com/automattic/studio/builds?branch=iangmaia%2Ffix-test-flakiness) are successful:
- https://buildkite.com/automattic/studio/builds/10549
- https://buildkite.com/automattic/studio/builds/10553
- https://buildkite.com/automattic/studio/builds/10558

PR #2407, opened 2026-01-16
Fix metrics test timeout for slow site creation

iangmaia:
Fixes AINFRA-1760

## Summary

Fixes the metrics test failure introduced in #2399. The reduced global `expect.timeout` (30s) is too short for app startup and site creation operations.

Added explicit 120s timeouts matching the pattern used in E2E tests:
- `onboarding.heading` visibility check
- `siteContent.siteNameHeading` visibility check  
- `siteContent.runningButton` attached check

## Test plan
- [ ] Metrics test passes in CI


wojtekn (approved):
Thanks for the fix.

PR #2415, opened 2026-01-19
Studio: Add site does not duplicate sites

katinthehatsite:
## Related issues

Fixes STU-1210

## Proposed Changes

This PR ensures that when you add a site from `Add site` in the sidebar, only one site gets added and there is no duplication. 

## Testing Instructions

* In the sidebar, click on `Add site`
* Observe that once the site is added, it only gets added once and you don't see any duplicate sites

## Pre-merge Checklist

### DIFF
--- a/src/modules/cli/lib/cli-events-subscriber.ts
+++ b/src/modules/cli/lib/cli-events-subscriber.ts
@@ -26,16 +26,17 @@ function siteDetailsToServerDetails(
 	};
 }
 
-const handleSiteEvent = sequential( async ( event: SiteEvent ): Promise< boolean > => {
+const handleSiteEvent = sequential( async ( event: SiteEvent ): Promise< void > => {
 	const { event: eventType, siteId, site, running } = event;
 
 	if ( eventType === SITE_EVENTS.DELETED ) {
 		SiteServer.unregister( siteId );
-		return true;
+		void sendIpcEventToRenderer( 'site-event', event );
+		return;
 	}
 
 	if ( ! site ) {
-		return true;
+		return;
 	}
 
 	// Only register new sites on CREATED events to prevent duplicates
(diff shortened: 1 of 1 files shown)


bcotrim (commented):
Thanks @katinthehatsite for looking into this issue.
We already do a similar check in the main process before emitting the `site-event` event to the UI.
Have you confirmed what is the issue with that check? It would be preferable to keep the logic in the main process and leave the UI as simple as possible, in my opinion.

bcotrim:
> Thanks @katinthehatsite for looking into this issue. We already do a similar check in the main process before emitting the `site-event` event to the UI. Have you confirmed what is the issue with that check? It would be preferable to keep the logic in the main process and leave the UI as simple as possible, in my opinion.

I forgot to add the link to the existing check:
https://github.com/Automattic/studio/blob/trunk/src/modules/cli/lib/cli-events-subscriber.ts#L42

katinthehatsite:
> We already do a similar check in the main process before emitting the site-event event to the UI.
Have you confirmed what is the issue with that check? It would be preferable to keep the logic in the main process and leave the UI as simple as possible, in my opinion.

I see, thanks for adding the link @bcotrim . I looked into the check and it seems that it exited the main process correctly but did not notify the renderer process about the exit properly. I made some changes to `src/modules/cli/lib/cli-events-subscriber.ts` - let me know what you think!

bcotrim (approved):
Thanks for finding and fixing the issue @katinthehatsite 

Nit: If we move `sendIpcEventToRenderer` inside the `handleSiteEvent` function, we could structure it with early returns for the "don't send" cases and a single send at the end. In my opinion, it makes the function intention easier to read and understand.

katinthehatsite:
> Nit: If we move sendIpcEventToRenderer inside the handleSiteEvent function, we could structure it with early returns for the "don't send" cases and a single send at the end. In my opinion, it makes the function intention easier to read and understand.

 Thanks for the suggestion! I did the refactor in https://github.com/Automattic/studio/pull/2415/changes/bc0c681e7c989b48e1955d1243d6d50c88a78e12

PR #2420, opened 2026-01-19
Fix constant theme details refresh issue

ivan-ottinger:
## Related issues

- Fixes STU-1211

## Proposed Changes

CLI events subscriber overwrites `server.details`, losing the `themeDetails` property that it includes. This causes false theme change detection on every app focus.

To resolve the issue we are ensuring that existing server details are preserved.

## Testing Instructions

1. Try to reproduce the STU-1211 issue on the current `trunk`.
2. Check out the PR branch and build the app with `npm install && npm start`.
3. Try to reproduce the STU-1211 issue again. It should not be possible.
4. Try changing the site theme. The site thumbnail should automatically refresh itself.

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -767,8 +767,7 @@ export async function getThemeDetails(
 	}
 
 	const themeDetails = await server.getThemeDetails();
-	const themeChanged =
-		themeDetails?.path && themeDetails.path !== server.themeDetails?.path;
+	const themeChanged = themeDetails?.path && themeDetails.path !== server.themeDetails?.path;
 
 	if ( themeChanged ) {
 		const parentWindow = BrowserWindow.fromWebContents( event.sender );
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -767,8 +767,7 @@ export async function getThemeDetails(
 	}
 
 	const themeDetails = await server.getThemeDetails();
-	const themeChanged =
-		themeDetails?.path && themeDetails.path !== server.themeDetails?.path;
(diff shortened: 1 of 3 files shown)


bcotrim (commented):
Thanks for finding and fixing the issue @ivan-ottinger
I would take a simpler approach and never overwrite the `themeDetails` in the `cli-events-subscriber`.
`themeDetails` is a property managed and used only by Studio app, I believe the CLI is never updating or sending it, so we are safe to ignore it.

```typescript
 function siteDetailsToServerDetails(
        site: SiteDetails,
        running: boolean,
        existingDetails?: SiteServer[ 'details' ] 
  ): SiteServer[ 'details' ] {
        return {
                ...existingDetails,  // Preserve existing properties
                ...site,             // Override with CLI data
                running,
        };
  }

...

server.details = siteDetailsToServerDetails( site, running, server.details );
```

What do you think?

katinthehatsite (approved):
This worked well for me, I could no longer reproduce the issue from trunk 👍 

I see that the fix has also been refactored as per suggestion.

ivan-ottinger:
> Thanks for finding and fixing the issue @ivan-ottinger
I would take a simpler approach and never overwrite the themeDetails in the cli-events-subscriber.
themeDetails is a property managed and used only by Studio app, I believe the CLI is never updating or sending it, so we are safe to ignore it.

Thank you for your input, Bernardo. That makes sense. I have updated the logic in https://github.com/Automattic/studio/pull/2420/commits/1b8cead7d2be46f0bcff00b59659b211d32204dd.

---

> This worked well for me, I could no longer reproduce the issue from trunk 👍

> I see that the fix has also been refactored as per suggestion.

Thank you for your review and testing, Kat.

PR #2422, opened 2026-01-19
Fix backup import for Jetpack, Local, Wpress and Studio formats

bcotrim:
## Related issues

- Fixes STU-1217

## Proposed Changes

- Fix PHP version type conversion in CLI: `yargs-parser` converts `--php-version 8.3` from string to number, causing Zod validation to fail silently. Now explicitly converts back to string.
- Fix SQL file path for standalone PHP-WASM: The `sqlite import` command needs an absolute path `/wordpress/<file>` because standalone PHP-WASM mounts the site at `/wordpress` but doesn't set the working directory there.

## Testing Instructions

1. Create a new site in Studio
2. Export a backup from the site
3. Create another new site
4. Import the backup using "Import" option
5. Verify the import completes successfully without "Database import failed" error
6. Test with different backup formats: Jetpack, Local, Wpress, Studio backups

## Pre-merge Checklist

### DIFF
--- a/cli/commands/wp.ts
+++ b/cli/commands/wp.ts
@@ -132,7 +132,10 @@ export async function commandHandler( argv: ArgumentsCamelCase< WpCommandOptions
 			);
 		}
 
-		const phpVersion = parsedWpCliArgs[ 'php-version' ] as string | undefined;
+		const phpVersion =
+			parsedWpCliArgs[ 'php-version' ] !== undefined
+				? String( parsedWpCliArgs[ 'php-version' ] )
+				: undefined;
 		wpCliArgv = removeArgumentFromArgv( wpCliArgv, 'php-version' );
 		wpCliArgv = removeArgumentFromArgv( wpCliArgv, 'avoid-telemetry', false );
 
--- a/src/lib/import-export/import/importers/importer.ts
+++ b/src/lib/import-export/import/importers/importer.ts
@@ -74,7 +74,7 @@ abstract class BaseImporter extends EventEmitter implements Importer {
 				await this.prepareSqlFile( tmpPath );
 				console.log( `Importing ${ sqlFile }` );
 				const { stderr, exitCode, stdout } = await server.executeWpCliCommand(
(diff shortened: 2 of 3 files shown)


cli/commands/wp.ts
```diff
@@ -132,7 +132,10 @@ export async function commandHandler( argv: ArgumentsCamelCase< WpCommandOptions
 			);
 		}
 
-		const phpVersion = parsedWpCliArgs[ 'php-version' ] as string | undefined;
```
fredrikekelund, line 135:
```ts
		const parsedWpCliArgs = yargsParser( wpCliArgv, {
			string: [ 'php-version' ],
		} );

		if ( parsedWpCliArgs._[ 0 ] === 'shell' ) {
			throw new LoggerError(
				__(
					'Studio CLI does not support the WP-CLI `shell` command. Consider adding your code to a file and using the `eval` command.'
				)
			);
		}

		const phpVersion = parsedWpCliArgs[ 'php-version' ];
```
Unfortunately, `yargs-parser` doesn't properly type `parsedWpCliArgs[ 'php-version' ]` if we do this, so I still prefer your solution.

I wanted to mention the possibility of configuring `yargs-parser`, though, since it fixes the underlying issue of `--php-version` being parsed by yargs as a number.

fredrikekelund (approved):
Good catch! This works great 👍

katinthehatsite (approved):
The changes look good to me, I was able to import .wpress, Studio and Jetpack files. I got this warning with `.wpress` but I don't think it is related to this PR:


```
Warning during replacing URLs (https:\/\/mxhtestpremium3.wpcomstaging.com): Deprecated: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in /wordpress/wp-content/mu-plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-pdo-user-defined-functions.php on line 511
```

I also noticed something strange with the database export for older sites where it was failing so asked on Slack: p1768898334187009/1768816016.045449-slack-C06DRMD6VPZ but again it is not related to this PR 👍

wojtekn (approved):
Thanks for the fix. The code change looks reasonable.

PR #2424, opened 2026-01-20
Make `Stop all` button run `site stop --all` (which is much faster)

fredrikekelund:
## Related issues

- Fixes STU-1223

## Proposed Changes

- Make the `Stop all` button in the Studio sidebar run the `site stop --all` CLI command instead of iterating over the running sites and sequentially triggering `site stop`. The new solution is _much_ faster.
- Remove redundant `setSites` calls from the `startServer` and `stopServer` functions in `src/hooks/use-site-details.tsx`. The `site-event` event listener already does the same thing.
- Remove redundant `updateSite` calls from the `startServer` and `stopServer` functions in `src/ipc-handlers.ts`

## Testing Instructions

1. Run `npm start`
2. Start at least two sites
3. Ensure that starting them works as normal and that they are marked as running in the sidebar
4. Click the `Stop all` button
5. Ensure that all sites are immediately stopped
6. Start a site again
7. Stop that individual site (not by clicking the `Stop all` button`
8. Ensure that it works as expected

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2426, opened 2026-01-20
Change execution order in `site stop --all` command

fredrikekelund:
## Related issues

- Fixes STU-1225

## Proposed Changes

As noted by @epeicher in https://github.com/Automattic/studio/pull/2424#pullrequestreview-3681215893, the `site stop --all` command doesn't save the `autoStart` prop to the appdata file as expected. 

This apparently happens because pm2 sends a SIGQUIT signal from the daemon to the client process (the `site stop --all` process) as part of the shutdown sequence. SIGQUIT doesn't immediately interrupt the process – it's allowed to run a bit of cleanup logic – but the kernel forces it to quit eventually. That's why the logic that updates the `autoStart` properties in the appdata file never ran previously.

In this PR, we flip the order: the appdata file is updated first, and pm2 is killed second. I've also moved the `process.exit( 0 )` call into the `killDaemonAndChildrenAndExitProcess` function to clarify the logic flow.

## Testing Instructions

1. Run `npm start`
2. Start at least two sites
3. Click the `Stop all` button
4. Quit Studio
5. Run `npm start` again
6. Ensure that no sites are automatically started when the application launches

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2428, opened 2026-01-20
Always emit `site-event` events for site start and site stop operations

fredrikekelund:
## Related issues

- Fixes STU-1227

## Proposed Changes

In https://github.com/Automattic/studio/pull/2424, I changed the `startServer` and `stopServer` IPC functions to return `void`, so the front-end wouldn't immediately update the site's running state. Instead, the front-end would rely on the `site-event` listener to accomplish this.  

This collided with the changes from https://github.com/Automattic/studio/pull/2415, which made it so that no `site-event` events are submitted if `SiteServer.hasOngoingOperation` is true, which it is if the operation (start a site, stop a site) was initiated from Studio.

This is why performance tests started failing after I merged https://github.com/Automattic/studio/pull/2424 to trunk, even though they didn't fail on the PR itself. 

This PR solves this problem by changing the logic in `handleSiteEvent` so that `site-event` events are always emitted when a site starts or stops, even if the operation was initiated from Studio.  

## Testing Instructions

1. Run `npm start`
2. Start a site using the button in the header
3. Ensure that it starts correctly
4. Stop the site using the button in the header
5. Ensure that it stops correctly
6. Create a site
7. Ensure that it works as expected
8. Delete the newly created site
9. Ensure that it works as expected

## Pre-merge Checklist

### DIFF
--- a/src/modules/cli/lib/cli-events-subscriber.ts
+++ b/src/modules/cli/lib/cli-events-subscriber.ts
@@ -47,6 +47,10 @@ const handleSiteEvent = sequential( async ( event: SiteEvent ): Promise< void > 
 		if ( ! existingServer ) {
 			SiteServer.register( siteDetailsToServerDetails( site, running ) );
 		}
+		// Don't send to renderer if site is being created by UI (createSite IPC will handle it)
+		if ( existingServer?.hasOngoingOperation ) {
+			return;
+		}
 		void sendIpcEventToRenderer( 'site-event', event );
 		return;
 	}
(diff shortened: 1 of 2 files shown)


fredrikekelund:
It looks like the performance tests are failing when running the `site-startup` test suite on the `trunk` build. This is expected, since the code on `trunk` is currently broken. 

The `site-startup` test suite passes on aa642c1, which is what we want 👍

bcotrim (commented):
Overall I agree with the idea of relying on events, and ensuring all sites changes come from the same `events` workflow.
When testing a new site however I see duplicated sites in Studio:

Image

bcotrim (approved):
LGTM and works as described
👍

gcsecsey (approved):
LGTM and works well. 👍

epeicher (approved):
Thanks @fredrikekelund! I have tested it, and it works as advertised. I can start and stop sites as expected, and I can create sites without any issues. This is a nice complementary PR to https://github.com/Automattic/studio/pull/2426. LGTM!

PR #2429, opened 2026-01-20
Change `./cli` paths in `make-pot` script

fredrikekelund:
## Related issues

- Fixes STU-1226

## Proposed Changes

- Change the path globs passed to `wp-babel-makepot` in `scripts/make-pot.mjs`. Previously, top-level files in the `./cli` directory weren't targeted (like `cli/index.ts`).

> [!NOTE]  
> I've already generated a new pot file and imported it to the Studio project on translate.wordpress.com

## Testing Instructions

1. Run `node scripts/make-pot.mjs`
2. Open `out/pots/bundle-strings.pot`
3. Ensure that strings from `cli/index.ts` are included
4. Ensure that no strings from `cli/node_modules` are included

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2433, opened 2026-01-20
Fix SQLite setup for existing WordPress sites

sejas:
## Related issues

- Fixes STU-1214

## Issue

When adding an existing WordPress site directory to Studio, SQLite was incorrectly installed even when the site already had a `wp-config.php` file (likely configured for MySQL). This overwrote the user's existing database configuration.

## Proposed Changes

- Skip SQLite installation when `wp-config.php` exists and `db.php` doesn't exist
- Update tests to cover the new conditional behavior

## Testing Instructions

1. Create a site in Studio
2. Delete site from Studio (preserve the directory)
3. Remove SQLite files (mu-plugin/sqlite-database-integration, db.php, and database dir), leaving `wp-config.php`
4. Add the site again using the existing directory
5. **Expected:** SQLite is NOT installed, `wp-config.php` is preserved
6. Verify fresh site creation still installs SQLite normally

## Pre-merge Checklist


fredrikekelund:
The unit test failure looks like a fluke. I had the same error on https://github.com/Automattic/studio/pull/2429 (which didn't change anything that might affect the unit tests). Retrying the unit test job fixed it.

I took the liberty of merging the latest changes from trunk here too, to retrigger CI and see if that fixes the problem.

cli/commands/site/create.ts
```diff
@@ -171,16 +175,19 @@ export async function runCommand(
...
+				throw new LoggerError(
+					__(
+						'SQLite integration files not found. Please ensure Studio is installed and has been run at least once.'
+					)
+				);
+			}
+			logger.reportStart( LoggerAction.INSTALL_SQLITE, __( 'Setting up SQLite integration…' ) );
+			await installSqliteIntegration( sitePath );
+			logger.reportSuccess( __( 'SQLite integration configured' ) );
 		}
```
```suggestion
		logger.reportStart( LoggerAction.INSTALL_SQLITE, __( 'Setting up SQLite integration…' ) );
		await keepSqliteIntegrationUpdated( sitePath );
		logger.reportSuccess( __( 'SQLite integration configured' ) );
```
`keepSqliteIntegrationUpdated` actually implements all the same logic, so we might as well simplify this code and just use that function.

fredrikekelund (commented):
This change works, but `keepSqliteIntegrationUpdated` implements all the same logic, so I suggest using that function instead.

Side note, we could remove the `installSqliteIntegration` export from `src/lib/sqlite-versions.ts` in this PR. It's no longer used.

cli/commands/site/create.ts
sejas: Great suggestion. I implemented it here https://github.com/Automattic/studio/pull/2433/changes/8dc22735ffcc283e94dfb710801c97e9b1ba817d . I decided to change the success messages depending if SQLite was "updated" or not.

sejas:
@fredrikekelund , thanks for the review. I applied your suggestions, and it's ready for another review.

> Side note, we could remove the installSqliteIntegration export from src/lib/sqlite-versions.ts in this PR. It's no longer used.

Done https://github.com/Automattic/studio/pull/2433/commits/7f6f09a22b4f7187a769305199f9a3c12be439ff

wojtekn (approved):
It fixes the use case: Studio doesn't try to set up SQLite if there is no SQLite plugin installed and a `wp-config.php` file exists.

PR #2307, opened 2025-12-22
Replace electron-devtools-installer with native Electron API

wojtekn:
## Related issues

- N/A

## Proposed Changes

I propose removing `electron-devtools-installer` dependency which uses deprecated Electron APIs and implementing native extension loading using `session.extensions.loadExtension()`. It eliminates deprecation warnings for `session.loadExtension` and `session.getAllExtensions`.

## Testing Instructions

Confirm:
- Builds go through
- React DevTools and Redux DevTools load successfully in development mode
- There are no deprecation warnings for extension-related APIs

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1379, opened 2025-05-12
Add workaround for devtools installation

bcotrim:
## Related issues

- Fixes STU-268

## Proposed Changes

- Update `electron-devtools-installer` from to `v4.0.0`
- Integrate Redux DevTools in addition to React Developer Tools
- Added workaround for extensions support: https://github.com/electron/electron/issues/41613#issuecomment-2644018998
- Updated debugging documentation to reference browser extension tools
- Remove standalone React Dev Tools integration via REACT_DEV_TOOLS environment variable

## Testing Instructions

- Run `npm start` and verify that React Developer Tools and Redux DevTools are available in the dev tools panel
- Confirm that you can inspect React components and Redux state

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2437, opened 2026-01-21
Add electron devtools installer back

wojtekn:
## Related issues

- Follow-up to https://github.com/Automattic/studio/pull/2307

## Proposed Changes

PR #2307 removed `electron-devtools-installer` due to deprecated Electron APIs, replacing it with a custom implementation. However, this broke the automatic extension installation for new developers.

This PR takes a different approach by patching `electron-devtools-installer` to use the new Electron APIs (`session.extensions.*`) instead of the deprecated ones (`session.*`). This solution:
- Eliminates deprecation warnings
- Maintains automatic extension installation
- Uses the well-tested and maintained `electron-devtools-installer` package
- Applies the fix from upstream commit [f8ec609](https://github.com/MarshallOfSound/electron-devtools-installer/commit/f8ec609cddf6d28a42c0f90b775ac251158b776a) (not yet released)

## Testing Instructions

1. Start the app and confirm:
- Extensions should install automatically
- No deprecation warnings should appear in the console
2. Verify extensions work:
- Open DevTools (Cmd+Option+I / Ctrl+Shift+I)
- Check for "Redux" tab (Redux DevTools)

## Pre-merge Checklist


package.json
```diff
@@ -18,6 +18,7 @@
 	},
 	"scripts": {
 		"prestart": "npm run cli:build",
+		"setup:devtools": "ts-node ./scripts/setup-dev-extensions.ts",
```
bcotrim: I couldn't find this file in the `scripts` folder.

Image

bcotrim (commented):
I was able to start the app without seeing any errors.
I am not sure if this was happening previously, but starting the app with devtools already opened doesn't show React DevTools.
I remember that previously we had this workaround: https://github.com/Automattic/studio/pull/1379/changes#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R105
Maybe we still need that?

Image

Doing a refresh ( cmd + R) displays the React DevTools as expected

Image

package.json
wojtekn: Good catch! It seems that the change from another PR slipped in. I removed this line.

wojtekn:
> Maybe we still need that?

That code was still there. I adjusted it, but I'm unsure if it helps - it still required me to run CMD+R to refresh it at least one. On subsequent runs, it shows without that refresh.

bcotrim:
> That code was still there. I adjusted it, but I'm unsure if it helps - it still required me to run CMD+R to refresh it at least one. On subsequent runs, it shows without that refresh.

Thanks for following up on that!
I can replicate the same behavior, but it also happens in `trunk`. Since the original goal of this PR is working I think we can address that issue as a follow-up, if necessary.

bcotrim (approved):

PR #2438, opened 2026-01-21
Convert performance logging to use native fetch with redirect support

youknowriad:
The perf jobs is currently silently failing because codevitals hosting changed a bit and now the log api redirects first.

## Proposed Changes

- Replace `https.request()` with native `fetch()` to automatically follow HTTP redirects (301, 302, etc.)
- Add proper error handling with `process.exit(1)` on non-2xx responses or network errors
- Simplify code by using async/await instead of callback-based approach

## Testing Instructions

This script is run during `npm run test:metrics` in CI. The change is backwards compatible - if the endpoint doesn't redirect, behavior is identical to before. If it does redirect, the metrics will now properly reach the destination instead of failing silently.

## Pre-merge Checklist

- [x] TypeScript linting passed

### DIFF
--- a/scripts/compare-perf/log-to-codevitals.ts
+++ b/scripts/compare-perf/log-to-codevitals.ts
@@ -1,8 +1,8 @@
 #!/usr/bin/env node
 /* eslint-disable no-console */
 import fs from 'fs';
-import https from 'https';
 import path from 'path';
+
 const [ token, branch, hash, baseHash, timestamp ] = process.argv.slice( 2 );
 
 const resultsFiles = [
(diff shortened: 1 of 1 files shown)


wojtekn:
@youknowriad to clarify, it started failing today after deploying changes to CodeVitals, but was working correctly before now, correct?

wojtekn (approved):

youknowriad:
So I think this (the redirect) was probably not the issue. I had an issue in the hosting as well that I fixed. That said, this PR seems like a nice refactoring regardless :)

PR #2441, opened 2026-01-21
Avoid overwriting site name when adding existing WordPress installations

sejas:
## Related issues

- Fixes STU-1233

## Proposed Changes

- Ignore the `--name` argument if the target folder contains an existing WordPress site with a wp-config.php file, which means it was initialized in the past.
- Add new tests to cover this usage.

## Testing Instructions

**Existing WordPress site directory**

- Run `npm run cli:build`
- Run `studio site create /path/to/existing-wordpress --name "My New Site name that will be ignored"`
- Observe the started site uses the original site name

**New sites**

- Run `npm run cli:build`
- Run `studio site create /path/to/new-empty-folder --name "My New Site Name"`
- Observe the started site uses the provided name

## Pre-merge Checklist

### DIFF
--- a/cli/commands/site/create.ts
+++ b/cli/commands/site/create.ts
@@ -207,13 +207,21 @@ export async function runCommand(
 			}
 		}
 
+		logger.reportStart( LoggerAction.SET_SITE_NAME, __( 'Setting site name…' ) );
 		if ( options.name ) {
-			setupSteps.push( {
-				step: 'setSiteOptions',
-				options: {
-					blogname: options.name,
-				},
-			} );
+			if ( isWordPressDirResult ) {
+				logger.reportWarning(
+					__( 'Site name ignored because the directory contains a WordPress site.' )
+				);
+			} else {
+				logger.reportSuccess( sprintf( __( 'Site name set to: %s' ), options.name ) );
+				setupSteps.push( {
+					step: 'setSiteOptions',
+					options: {
+						blogname: options.name,
+					},
+				} );
+			}
 		}
 
 		if ( setupSteps.length > 0 ) {
--- a/cli/commands/site/create.ts
+++ b/cli/commands/site/create.ts
@@ -209,7 +209,8 @@ export async function runCommand(
 
 		logger.reportStart( LoggerAction.SET_SITE_NAME, __( 'Setting site name…' ) );
 		if ( options.name ) {
-			if ( isWordPressDirResult ) {
+			const hasWpConfig = await pathExists( path.join( sitePath, 'wp-config.php' ) );
+			if ( isWordPressDirResult && hasWpConfig ) {
 				logger.reportWarning(
 					__( 'Site name ignored because the directory contains a WordPress site.' )
 				);
--- a/cli/commands/site/create.ts
+++ b/cli/commands/site/create.ts
@@ -207,22 +207,15 @@ export async function runCommand(
 			}
 		}
 
-		logger.reportStart( LoggerAction.SET_SITE_NAME, __( 'Setting site name…' ) );
-		if ( options.name ) {
-			const hasWpConfig = await pathExists( path.join( sitePath, 'wp-config.php' ) );
-			if ( isWordPressDirResult && hasWpConfig ) {
-				logger.reportWarning(
-					__( 'Site name ignored because the directory contains a WordPress site.' )
-				);
-			} else {
-				logger.reportSuccess( sprintf( __( 'Site name set to: %s' ), options.name ) );
-				setupSteps.push( {
-					step: 'setSiteOptions',
-					options: {
-						blogname: options.name,
-					},
-				} );
-			}
+		const hasWpConfig = await pathExists( path.join( sitePath, 'wp-config.php' ) );
+		const isWordPressDirectoryInitialized = isWordPressDirResult && hasWpConfig;
+		if ( options.name && ! isWordPressDirectoryInitialized ) {
+			setupSteps.push( {
+				step: 'setSiteOptions',
+				options: {
+					blogname: options.name,
+				},
+			} );
 		}
 
 		if ( setupSteps.length > 0 ) {
--- a/cli/commands/site/tests/create.test.ts
+++ b/cli/commands/site/tests/create.test.ts
@@ -1,4 +1,4 @@
-import { Blueprint } from '@wp-playground/blueprints';
+import { Blueprint, StepDefinition } from '@wp-playground/blueprints';
 import {
 	filterUnsupportedBlueprintFeatures,
 	validateBlueprintData,
@@ -101,6 +101,7 @@ describe( 'CLI: studio site create', () => {
 	let consoleLogSpy: jest.SpyInstance;
 	let fsMkdirSyncSpy: jest.SpyInstance;
 	let loggerReportSuccessSpy: jest.SpyInstance;
+	let loggerReportWarningSpy: jest.SpyInstance;
 
 	const createPathExistsMock = ( sitePathExists = false ) => {
 		const bundledWPPath = require( 'path' ).join(
@@ -126,6 +127,7 @@ describe( 'CLI: studio site create', () => {
 		consoleLogSpy = jest.spyOn( console, 'log' ).mockImplementation();
 		fsMkdirSyncSpy = jest.spyOn( require( 'fs' ), 'mkdirSync' ).mockReturnValue( undefined );
 		loggerReportSuccessSpy = jest.spyOn( Logger.prototype, 'reportSuccess' );
+		loggerReportWarningSpy = jest.spyOn( Logger.prototype, 'reportWarning' );
 		createPathExistsMock( false );
 		( isEmptyDir as jest.Mock ).mockResolvedValue( true );
 		( isWordPressDirectory as jest.Mock ).mockReturnValue( false );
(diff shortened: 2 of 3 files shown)


cli/commands/site/create.ts
```diff
@@ -207,13 +207,21 @@ export async function runCommand(
...
 
+		logger.reportStart( LoggerAction.SET_SITE_NAME, __( 'Setting site name…' ) );
 		if ( options.name ) {
-			setupSteps.push( {
-				step: 'setSiteOptions',
-				options: {
-					blogname: options.name,
-				},
-			} );
+			if ( isWordPressDirResult ) {
```
wojtekn: Will it catch the case in which I'm adding a site from an existing WordPress directory that has database files set?

Shouldn't it be something like https://github.com/Automattic/studio/pull/2442

sejas: I don't think we should check if `hadExistingDatabase` because the site could be using MySQL.

I tested the case when I have a WordPress directory with database files set and it works as expected.
The `isWordPressDirectory` checks for wp-content, wp-includes and wp-load.php, and in that case it will skip setting the name.

https://github.com/Automattic/studio/blob/dd5c1d33f479d7e3668cce32fe59078521703e56/common/lib/fs-utils.ts#L45-L51

If we want to be more thorough, we could also check whether the wp-config.php file exists. So the Site name would be ignored if the site has been initialized in the past.

wojtekn: I see, so it would skip setting name even for cases in which user provided path that contains WordPress without database and Studio sets fresh one based on SQLite, correct?

sejas: Added a condition to check if wp-config.php exists and only in that case ignore the name.
https://github.com/Automattic/studio/pull/2441/changes/2707a774ef045c769a7d1242a83ec4dfc3134718

cli/commands/site/tests/create.test.ts
```diff
@@ -301,6 +303,97 @@ describe( 'CLI: studio site create', () => {
...
+					return Promise.resolve( true );
+				}
+				if ( path === wpConfigPath ) {
+					return Promise.resolve( true );
+				}
+				if ( path === mockSitePath ) {
+					return Promise.resolve( true );
+				}
+				return Promise.resolve( false );
+			} );
```
fredrikekelund:
```suggestion
			( pathExists as jest.Mock ).mockImplementation(
				async ( path: string ) =>
					path === bundledWPPath || path === wpConfigPath || path === mockSitePath
			);
```
Nit, but we might as well shorten this

cli/commands/site/tests/create.test.ts
```diff
@@ -301,6 +303,97 @@ describe( 'CLI: studio site create', () => {
...
+					return Promise.resolve( true );
+				}
+				if ( path === wpConfigPath ) {
+					return Promise.resolve( false );
+				}
+				if ( path === mockSitePath ) {
+					return Promise.resolve( true );
+				}
+				return Promise.resolve( false );
+			} );
```
fredrikekelund:
```suggestion
			const bundledWPPath = require( 'path' ).join(
				'/test/server-files',
				'wordpress-versions',
				'latest'
			);
			( pathExists as jest.Mock ).mockImplementation(
				async ( path: string ) => path === bundledWPPath || path === mockSitePath
			);
```
Again, nitpicking about the potential to shorten this.

common/logger-actions.ts
```diff
@@ -18,6 +18,7 @@ export enum PreviewCommandLoggerAction {
 }
 
 export enum SiteCommandLoggerAction {
+	SET_SITE_NAME = 'setName',
```
fredrikekelund: Please either move this a bit further down or sort the entire enum alphabetically. Right now, the values are roughly sorted in the order they run. I realize this will be difficult to maintain, though, so I'm personally inclined to sort this alphabetically instead

cli/commands/site/create.ts
```diff
@@ -207,13 +207,22 @@ export async function runCommand(
 			}
 		}
 
+		logger.reportStart( LoggerAction.SET_SITE_NAME, __( 'Setting site name…' ) );
```
```suggestion
		logger.reportStart( LoggerAction.SET_SITE_NAME, __( 'Setting WordPress site name…' ) );
```
"site name" seems a bit ambiguous to me

cli/commands/site/create.ts
sejas: > I see, so it would skip setting name even for cases in which user provided path that contains WordPress without database and Studio sets fresh one based on SQLite, correct?

Yes! that's correct. If the user removes the database and leaves the wp-config, then the site name will be skipped.

cli/commands/site/create.ts
```diff
@@ -207,13 +207,22 @@ export async function runCommand(
...
-			setupSteps.push( {
-				step: 'setSiteOptions',
-				options: {
-					blogname: options.name,
-				},
-			} );
+			const hasWpConfig = await pathExists( path.join( sitePath, 'wp-config.php' ) );
+			if ( isWordPressDirResult && hasWpConfig ) {
+				logger.reportWarning(
+					__( 'Site name ignored because the directory contains a WordPress site.' )
```
```suggestion
					__( 'Will skip configuring WordPress site name because the directory contains a WordPress site.' )
```
Not sure about the exact phrasing, but similarly to my previous comment, I think it's a little ambiguous what "site name ignored" means.

cli/commands/site/create.ts
```diff
@@ -207,13 +207,22 @@ export async function runCommand(
...
+				);
+			} else {
+				logger.reportSuccess( sprintf( __( 'Site name set to: %s' ), options.name ) );
```
```suggestion
				logger.reportSuccess( sprintf( __( 'WordPress site name set to: %s' ), options.name ) );
```

fredrikekelund (approved):
The core logic works great 👍 I left a couple of comments. None of them are necessarily blocking, but I'd recommend we look into changing the logger messages per my review comments.

bcotrim (commented):
Thanks for working on this! I have a concern about the current approach and a suggestion for a better UX.

When a user explicitly passes `--name` to the CLI (or enters a name in the Studio UI), we're now ignoring it with a warning if the directory contains an existing WordPress site. This feels counterintuitive. If the user explicitly provides a value, they likely want it applied.

Suggestion: Instead of ignoring the user's input, we could:

  1. Detect existing WordPress sites — When the user selects a directory that's a WordPress installation, use WP-CLI to fetch the current site name: `wp option get blogname`
  2. Pre-populate the form — In the Studio UI, suggest the fetched site name in the name field. The user can see what the site is currently called and change it if they want.
  3. Always respect explicit input. Any `--name` value provided by the user (CLI or UI) should be honored and set accordingly.

  This gives users visibility into the existing site name while still allowing them to override it when intended.

  What do you think?

sejas:
Yes, that's a good point. I agree that both options are valid. Respecting the existing site's name or respecting the name entered by the user in the UI or CLI even if that means that the user will override the site name for an existing site. See the discussion on [STU-1214](https://linear.app/a8c/issue/STU-1214/studio-sets-sqlite-for-site-that-already-has-wp-configphp#comment-afb19615). I’m curious what other folks think, too.<br><br>Pre-populating the site name input in the Studio app can be tricky because the site path is updated under the hood as the user types each character. So, if the user enters a name that matches an existing path, the entire site name could change instantly.

fredrikekelund:
> we're now ignoring it with a warning if the directory contains an existing WordPress site

Looking more closely at the logic in `cli/commands/site/create.ts`, I think we should either:

1. Keep the old logic and always set the WordPress site name if the `--name` option is passed, or…
2. Also remove the implicit `setSiteLanguage` blueprint [logic](https://github.com/Automattic/studio/blob/642fa1ca82cc3faffa1680595eecdc3a65548675/cli/commands/site/create.ts#L191-L208) if the directory contains an existing WordPress installation and a `wp-config.php` file.

wojtekn:
> When a user explicitly passes --name to the CLI (or enters a name in the Studio UI), we're now ignoring it with a warning if the directory contains an existing WordPress site. This feels counterintuitive. If the user explicitly provides a value, they likely want it applied.

I think this statement incorrectly assumes that the Studio site name must equal the WordPress site title. Let's think of that option as "How I want to name my site in Studio" and not "What title I want to use for WordPress site".

cli/commands/site/create.ts
wojtekn: I don't think we need this warning, as it implicitly assumes that it's expected that the WordPress site title must match the Studio site name, and that now something wrong happens, so the user is getting informed.

fredrikekelund:
I think that's fair, @wojtekn, but what about the `setSiteLanguage` language stuff? Any thoughts there?

wojtekn:
@fredrikekelund I think we should separately discuss what should be the source of truth if there is a collision between data coming from the user through UI/CLI, the existing database, and Blueprint.

cli/commands/site/tests/create.test.ts
sejas: Done! https://github.com/Automattic/studio/pull/2441/changes/c5990f964f2d1868d0e197d4ff66f7d5cde3be3f

cli/commands/site/tests/create.test.ts
sejas: Done https://github.com/Automattic/studio/pull/2441/changes/1f8215e1548cceb8ae4d501f91549c5807ef2eec

cli/commands/site/tests/create.test.ts
```diff
@@ -303,6 +305,70 @@ describe( 'CLI: studio site create', () => {
 			);
 		} );
 
+		it( 'should NOT override blogname when adding existing WordPress directory with wp-config.php and name', async () => {
```
wojtekn, line 308: Thanks for adding tests, it will better show the expected behavior.

cli/commands/site/create.ts
```diff
@@ -207,7 +207,9 @@ export async function runCommand(
 			}
 		}
 
-		if ( options.name ) {
+		const hasWpConfig = await pathExists( path.join( sitePath, 'wp-config.php' ) );
+		const isWordPressDirectoryInitialized = isWordPressDirResult && hasWpConfig;
```
wojtekn, line 211: It nicely documents the intent, too.

wojtekn (approved):

sejas:
@wojtekn , @fredrikekelund , @bcotrim , thanks for reviewing it and sharing your feedback.
After discussing with Wojtek, we decided to treat the name as Studio name and not site name, so I removed the logs displayed to the user. We might add follow-ups to communicate better this expectation.

bcotrim:
> @wojtekn , @fredrikekelund , @bcotrim , thanks for reviewing it and sharing your feedback. After discussing it with Wojtek, we decided to treat the name as Studio name and not site name, so I removed the logs displayed to the user. We might add follow-ups to communicate better this expectation.

Thanks for looking into the suggestions, that approach makes sense! 👍

bcotrim (approved):

PR #2442, opened 2026-01-21
Avoid changing site name if database already exists

wojtekn:
## Related issues

- Fixes #

## Proposed Changes

-

## Testing Instructions

-

## Pre-merge Checklist

### DIFF
--- a/cli/commands/site/create.ts
+++ b/cli/commands/site/create.ts
@@ -172,6 +172,8 @@ export async function runCommand(
 		}
 
 		logger.reportStart( LoggerAction.INSTALL_SQLITE, __( 'Setting up SQLite integration…' ) );
+		const sqliteDatabasePath = path.join( sitePath, 'wp-content', 'database', '.ht.sqlite' );
+		const hadExistingDatabase = await pathExists( sqliteDatabasePath );
 		const isSqliteUpdated = await keepSqliteIntegrationUpdated( sitePath );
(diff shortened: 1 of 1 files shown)


cli/commands/site/create.ts
```diff
@@ -207,7 +209,13 @@ export async function runCommand(
 			}
 		}
 
-		if ( options.name ) {
+		// Only set site name for fresh installations, not for existing WordPress sites
+		// that already have a database (either SQLite or MySQL). However, if we just created
+		// a fresh database (isSqliteUpdated is true AND no existing .ht.sqlite), we should set the name.
+		if (
+			options.name &&
+			( ! isWordPressDirResult || ( isSqliteUpdated && ! hadExistingDatabase ) )
```
sejas, line 217: The problem I see when using `hadExistingDatabase` and `isSqliteUpdated` is that it might not work for sites with MySQL. For those the .ht.sqlite database doesn't exist.

I implemented a similar check in https://github.com/Automattic/studio/pull/2441, but I only used `wp-config.php` to determine whether the site had been initialized previously.

fredrikekelund: I instinctively agree with Antonio's more simplistic approach. It's easier to explain to users, which is good.

wojtekn: Makes sense, thanks.

PR #2447, opened 2026-01-22
Fix regression where `themeDetails` are not persisted to appdata file

fredrikekelund:
## Related issues

- Fixes STU-1236

## Proposed Changes

We recently introduced a regression where `SiteDetails.themeDetails` objects stopped being persisted to the appdata file. This caused two problems:

1. After restarting Studio, the Overview tab was missing the appropriate navigation buttons for stopped sites with block themes (`Site Editor`, etc). This happened because `themeDetails` object contained all empty values. 
2. The Overview tab always showed a loading state when focusing the app window. This happened because the `SiteDetails.themeDetails` object wasn't updated properly. https://github.com/Automattic/studio/pull/2420 alleviated this problem, but it kept appearing due to the underlying regression. 

This PR fixes the problem by:

1. Consolidating the logic for updating `themeDetails` in a single function in `src/ipc-handlers.ts`. Previously, there were multiple execution paths to achieve this, which I believe explains why this regression surfaced in the first place. 
2. Clean up the last remaining `updateSite` calls in `src/ipc-handlers.ts` that were using this function for an ambiguous purpose (maybe update the settings, but also update the appdata file). 
3. Add a `SiteDetails.persistThemeDetails` method that does what you'd expect (update the `themeDetails` object of the site in the appdata file).
4. Clarifying when the different IPC events are emitted, renaming them for clarity, and adding a `thumbnail-load-error` to reset the loading state if there's an error when loading the thumbnail.

## Testing Instructions

1. Run `npm start`
2. Start a site
3. Open the appdata file (`~/Library/Application\ Support/Studio/appdata-v1.json`) and ensure that the site you just started has a `themeDetails` object that contains the values you'd expect
4. Focus the Studio window again
5. Ensure that there's no "loading state" flash in the Overview tab
6. Open wp-admin
7. Switch the theme
8. Focus the Studio window again
9. Ensure that the site thumbnail in the Overview tab reloads after a few seconds

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -117,56 +117,6 @@ export {
 	showUserSettings,
 } from 'src/modules/user-settings/lib/ipc-handlers';
 
-async function sendThumbnailChangedEvent( event: IpcMainInvokeEvent, id: string ) {
-	if ( event.sender.isDestroyed() ) {
-		return;
-	}
-	const thumbnailData = await getThumbnailData( event, id );
-	const parentWindow = BrowserWindow.fromWebContents( event.sender );
-	sendIpcEventToRendererWithWindow( parentWindow, 'thumbnail-changed', {
-		id,
-		imageData: thumbnailData,
-	} );
-}
-
-/**
- * Refreshes site metadata (theme details and thumbnail) and notifies the renderer.
- * This is typically called after a site is started or created.
- */
-async function refreshSiteMetadataAndNotify(
-	event: IpcMainInvokeEvent,
-	server: SiteServer,
-	parentWindow: BrowserWindow | null
-): Promise< void > {
-	const id = server.details.id;
-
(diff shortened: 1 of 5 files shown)


src/ipc-handlers.ts
```diff
@@ -331,6 +279,8 @@ export async function createSite(
 	}
 }
 
+// Update a site's details (name, custom domain, PHP version, etc). This function calls the
+// `site set` CLI command and updates the `SiteServer` instance after the CLI completes.
 export async function updateSite(
```
fredrikekelund, line 284: This function was previously used ambiguously (see https://github.com/Automattic/studio/pull/2360#discussion_r2676243841 and PR description). I've removed the last place where it was being used to update the appdata file (which it no longer did, anyway). I added this comment to prevent future confusion.

bcotrim (approved):
LGTM 👍 
The workflow is clearer and I didn't find any regressions

PR #2448, opened 2026-01-22
Revert winreg to 1.2.4 to avoid CLI install issue

ivan-ottinger:
## Summary
- Downgrade `winreg` from `^1.2.5` to `1.2.4` (pinned) to avoid a [known issue](https://github.com/fresc81/node-winreg/issues/65) that breaks the module
- add `winreg` to the Dependabot ignore list

Fixes STU-1240

## Test plan
1. Check out the PR branch and build the app on Windows.
2. Verify Windows CLI installation/uninstallation through the app settings → _Studio CLI_ toggle works correctly.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -44,6 +44,7 @@ updates:
           - "@php-wasm/*"
     ignore:
       - dependency-name: "eslint-plugin-studio"
+      - dependency-name: "winreg" # v1.2.5 has a known issue: https://github.com/fresc81/node-winreg/issues/65
 
   # Enable version updates for GitHub Actions
   - package-ecosystem: "github-actions"
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,7 +53,7 @@
 				"tar": "^7.5.2",
 				"ts-node": "^10.9.2",
 				"tus-js-client": "^4.3.1",
-				"winreg": "^1.2.5",
+				"winreg": "1.2.4",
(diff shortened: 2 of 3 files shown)


fredrikekelund:
@ivan-ottinger, I'm triggering a Windows dev build to be able to fully test the CLI installer. I've ensured no error is thrown on Windows when running the dev version, but the way the logic works means we can't fully test the installer without a production build.

ivan-ottinger:
> @ivan-ottinger, I'm triggering a Windows dev build to be able to fully test the CLI installer. I've ensured no error is thrown on Windows when running the dev version, but the way the logic works means we can't fully test the installer without a production build.

Yes, that's right. I have been testing this with the production build as well. Then I also noticed I have some old CLI-related files from previous Studio versions on the system.

fredrikekelund:
@ivan-ottinger, ah OK, so you've already verified that this works with a production build?

epeicher (approved):
Thanks @ivan-ottinger! I have created a production build using this version on my Windows machine and I have been able to install/uninstall the studio CLI now. LGTM! :shipit:

ivan-ottinger:
> @ivan-ottinger, ah OK, so you've already verified that this works with a production build?

Just noting that related conversation happened here: p1769085355925259-slack-C06DRMD6VPZ.

Thanks for reviews and testing, Roberto & Fredrik! 🙂

PR #2449, opened 2026-01-22
Fix PHP version not reflected in wp-admin Site Health

bcotrim:
## Related issues

- Fixes STU-1237

## Proposed Changes

- Include `preferredVersions` in the blueprint JSON passed to Playground CLI
- This ensures the PHP version is correctly applied when a `BlueprintBundle` is provided
- The fix establishes proper precedence: Studio config > Blueprint's preferredVersions > Default

**Root cause:** When Studio passes a `BlueprintBundle` to Playground CLI's `runCLI()`, the `args.php` value is ignored. The CLI only reads `preferredVersions.php` from inside the blueprint JSON itself.

## Testing Instructions

1. Create a new site with default settings
   - Verify wp-admin → Tools → Site Health → Info → Server shows PHP 8.3 (default)
2. Change PHP version to 8.1 in Studio
   - Verify wp-admin Site Health shows PHP 8.1.x
3. Change PHP version to 8.0
   - Verify wp-admin Site Health shows PHP 8.0.x
4. Change PHP version back to default
   - Verify wp-admin Site Health shows PHP 8.3.x

## Pre-merge Checklist

### DIFF
--- a/cli/wordpress-server-child.ts
+++ b/cli/wordpress-server-child.ts
@@ -22,6 +22,7 @@ import {
 	InMemoryFilesystem,
 } from '@wp-playground/storage';
 import { WordPressInstallMode } from '@wp-playground/wordpress';
+import { DEFAULT_PHP_VERSION } from 'common/constants';
 import { isWordPressDirectory } from 'common/lib/fs-utils';
 import { getMuPlugins } from 'common/lib/mu-plugins';
 import { decodePassword } from 'common/lib/passwords';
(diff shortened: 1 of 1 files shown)


gcsecsey (approved):
Thanks @bcotrim for the fix and for the explanation on why the previous approach didn't work correctly. This approach looks good to me, and works well both for PHP and WP version changes. 

| Studio | wp-admin |
|--------|--------|
| Image | Image | 
| Image | Image |
| Image | Image | 
| Image | Image |
| Image | Image |

sejas (approved):
Nice fix! I confirm now I'm able to change the PHP version and see it in the site health page.

Image

@bcotrim, If we are passing the version as blueprint config, should we remove the `args.php` parameter?

https://github.com/Automattic/studio/blob/83abd874d69c5f061a2bf5b1de4519337303c600/cli/wordpress-server-child.ts#L233-L235

fredrikekelund (approved):
LGTM 👍 Great catch, @bcotrim.

Maybe it's worth adding a brief comment above the `args.php` assignment to clarify that it's not effectively used now?

PR #2462, opened 2026-01-22
Fix Windows terminal not finding Studio CLI in PATH

epeicher:
## Related issues

- Closes STU-1246

## Root Cause:

The Studio CLI is installed from within Studio itself, which updates the Windows Registry PATH. However, the running Electron process doesn't automatically pick up registry changes, it keeps its original PATH from when it launched. So when Studio spawns Command Prompt, the terminal inherits the old PATH (without the CLI directory) from Studio.

Without this fix, users would need to close and reopen Studio after installing the CLI for the terminal shortcut to work.

## Proposed Changes

- Added `getWindowsCliBinPath()` function to `src/lib/windows-helpers.ts` that computes the Studio CLI bin directory path
- Updated `openTerminalAtPath()` in `src/ipc-handlers.ts` to prepend the CLI bin path to the PATH environment variable when spawning Command Prompt on Windows
- Solution preserves all system paths while ensuring the CLI bin directory is available

## Testing Instructions

- On Windows, build and install the app (`npm run make`)
- Open Studio and create a site (or use an existing one)
- Install the Studio CLI via Settings if not already installed
- Use Studio's "Open Terminal" shortcut to open Command Prompt
- Type `studio --version` in the terminal - it should work
- Verify that system commands like `where` still work (confirms system PATH is preserved)

## Pre-merge Checklist

### DIFF
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -833,17 +833,16 @@ export async function openTerminalAtPath( _event: IpcMainInvokeEvent, targetPath
 		// Ensure the Studio CLI bin directory is in the PATH for the spawned terminal.
 		// Child processes inherit the environment from the Electron process, which may have
 		// been started before the CLI was installed or PATH was updated in the registry.
-		const cliBinPath = unversionedBinDirPath;
 		const isCliInstalled = await isStudioCliInstalled();
 		let env: NodeJS.ProcessEnv | undefined;
-		if ( isCliInstalled && cliBinPath ) {
+		if ( isCliInstalled ) {
 			const currentPath = process.env.PATH || '';
 			const pathEntries = currentPath.split( ';' ).map( ( p ) => p.toLowerCase() );
-			if ( ! pathEntries.includes( cliBinPath.toLowerCase() ) ) {
+			if ( ! pathEntries.includes( unversionedBinDirPath.toLowerCase() ) ) {
 				env = { ...process.env };
 				delete env.PATH;
 				delete env.Path;
-				env.PATH = `${ cliBinPath };${ currentPath }`;
+				env.PATH = `${ unversionedBinDirPath };${ currentPath }`;
 			}
 		}
 
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -62,6 +62,8 @@ import { getLogsFilePath, writeLogToFile, type LogLevel } from 'src/logging';
 import { getMainWindow } from 'src/main-window';
 import { popupMenu, setupMenu } from 'src/menu';
 import { editSiteViaCli, EditSiteOptions } from 'src/modules/cli/lib/cli-site-editor';
+import { isStudioCliInstalled } from 'src/modules/cli/lib/ipc-handlers';
+import { unversionedBinDirPath } from 'src/modules/cli/lib/windows-installation-manager';
 import { shouldExcludeFromSync, shouldLimitDepth } from 'src/modules/sync/lib/tree-utils';
 import { supportedEditorConfig, SupportedEditor } from 'src/modules/user-settings/lib/editor';
 import { getUserTerminal } from 'src/modules/user-settings/lib/ipc-handlers';
@@ -831,9 +833,10 @@ export async function openTerminalAtPath( _event: IpcMainInvokeEvent, targetPath
 		// Ensure the Studio CLI bin directory is in the PATH for the spawned terminal.
 		// Child processes inherit the environment from the Electron process, which may have
 		// been started before the CLI was installed or PATH was updated in the registry.
-		const cliBinPath = windowsHelpers.getWindowsCliBinPath();
+		const cliBinPath = unversionedBinDirPath;
+		const isCliInstalled = await isStudioCliInstalled();
 		let env: NodeJS.ProcessEnv | undefined;
-		if ( cliBinPath ) {
+		if ( isCliInstalled && cliBinPath ) {
 			const currentPath = process.env.PATH || '';
 			const pathEntries = currentPath.split( ';' ).map( ( p ) => p.toLowerCase() );
 			if ( ! pathEntries.includes( cliBinPath.toLowerCase() ) ) {
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -834,10 +834,13 @@ export async function openTerminalAtPath( _event: IpcMainInvokeEvent, targetPath
 		const cliBinPath = windowsHelpers.getWindowsCliBinPath();
 		let env: NodeJS.ProcessEnv | undefined;
 		if ( cliBinPath ) {
-			const currentPath = process.env.Path || process.env.PATH || '';
+			const currentPath = process.env.PATH || '';
 			const pathEntries = currentPath.split( ';' ).map( ( p ) => p.toLowerCase() );
 			if ( ! pathEntries.includes( cliBinPath.toLowerCase() ) ) {
-				env = { ...process.env, Path: `${ cliBinPath };${ currentPath }` };
+				env = { ...process.env };
+				delete env.PATH;
+				delete env.Path;
+				env.PATH = `${ cliBinPath };${ currentPath }`;
 			}
 		}
 
--- a/src/ipc-handlers.ts
+++ b/src/ipc-handlers.ts
@@ -828,8 +828,22 @@ export async function openTerminalAtPath( _event: IpcMainInvokeEvent, targetPath
 			return promiseExec( `start "" "warp://action/new_tab?path=${ encodedPath }"` );
 		}
 
+		// Ensure the Studio CLI bin directory is in the PATH for the spawned terminal.
+		// Child processes inherit the environment from the Electron process, which may have
+		// been started before the CLI was installed or PATH was updated in the registry.
+		const cliBinPath = windowsHelpers.getWindowsCliBinPath();
+		let env: NodeJS.ProcessEnv | undefined;
+		if ( cliBinPath ) {
+			const currentPath = process.env.Path || process.env.PATH || '';
+			const pathEntries = currentPath.split( ';' ).map( ( p ) => p.toLowerCase() );
+			if ( ! pathEntries.includes( cliBinPath.toLowerCase() ) ) {
+				env = { ...process.env, Path: `${ cliBinPath };${ currentPath }` };
+			}
+		}
+
 		return promiseExec( `start "Command Prompt" ${ defaultShell }`, {
 			cwd: targetPath,
+			env,
 		} );
 	} else if ( platform === 'linux' ) {
 		return promiseExec( `gnome-terminal --working-directory=${ targetPath }` );
--- a/src/lib/windows-helpers.ts
+++ b/src/lib/windows-helpers.ts
@@ -3,8 +3,8 @@ import path from 'path';
 import sudo from '@vscode/sudo-prompt';
 import { __ } from '@wordpress/i18n';
 import { getMainWindow } from 'src/main-window';
+import { unversionedBinDirPath } from 'src/modules/cli/lib/windows-installation-manager';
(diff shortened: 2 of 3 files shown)


fredrikekelund (changes requested):
The root cause to this problem appears to be that `unversionedBinDirPath` gets wiped as part of the Studio update, which short circuits `WindowsCliInstallationManager::updateWindowsCliVersionedPathIfNeeded`.

I'm working on a separate PR to fix this.

fredrikekelund:
Hold on… It looks like we have two separate issue here. This one appears legit. Sorry for jumping to conclusions, @epeicher.

I'm reviewing this again now.

fredrikekelund:
I can't reproduce this 🤔

Here's what I did:
1. Uninstalled the Studio CLI from the settings modal
2. Uninstalled the Studio app
3. Installed 1.7.0-beta1 again
4. Opened the Command prompt using the button in the overview tab
5. Ran `where studio` to verify that the `studio` command wasn't in my PATH
6. Installed the Studio CLI from the settings modal
7. Opened the Command prompt using the button in the overview tab
8. Ran `where studio`
9. The `studio` command was now available in my PATH

epeicher:
> I can't reproduce this 🤔

That's weird @fredrikekelund, I have tested it again, and it has worked once, but it has failed the second time. Do you think there could be some racing conditions that make it work/fail depending on timing? 

Please see a video below with the issue being reproduced 

https://github.com/user-attachments/assets/a23864cb-8967-488a-b596-c2094e40d0ee

src/lib/windows-helpers.ts
```diff
@@ -5,6 +5,24 @@ import { __ } from '@wordpress/i18n';
...
+ */
+export function getWindowsCliBinPath(): string | undefined {
+	if ( process.platform !== 'win32' ) {
+		return undefined;
+	}
+
+	// This matches the `unversionedBinDirPath` calculation in windows-installation-manager.ts:
+	// path.resolve( path.dirname( app.getPath( 'exe' ) ), '../bin' )
+	return path.resolve( path.dirname( app.getPath( 'exe' ) ), '../bin' );
+}
```
fredrikekelund: Let's get this value from `src/modules/cli/lib/windows-installation-manager.ts` instead. `unversionedBinDirPath` is the one we want. Maybe add a static getter function to `WindowsCliInstallationManager`?

src/ipc-handlers.ts
```diff
@@ -828,8 +828,22 @@ export async function openTerminalAtPath( _event: IpcMainInvokeEvent, targetPath
 			return promiseExec( `start "" "warp://action/new_tab?path=${ encodedPath }"` );
 		}
 
+		// Ensure the Studio CLI bin directory is in the PATH for the spawned terminal.
+		// Child processes inherit the environment from the Electron process, which may have
+		// been started before the CLI was installed or PATH was updated in the registry.
+		const cliBinPath = windowsHelpers.getWindowsCliBinPath();
+		let env: NodeJS.ProcessEnv | undefined;
+		if ( cliBinPath ) {
+			const currentPath = process.env.Path || process.env.PATH || '';
```
```suggestion
			const currentPath = process.env.PATH || '';
```
Per the [Node.js process docs](https://nodejs.org/docs/latest-v22.x/api/process.html#processenv):

> On Windows operating systems, environment variables are case-insensitive.

src/ipc-handlers.ts
```diff
@@ -828,8 +828,22 @@ export async function openTerminalAtPath( _event: IpcMainInvokeEvent, targetPath
...
+			const pathEntries = currentPath.split( ';' ).map( ( p ) => p.toLowerCase() );
+			if ( ! pathEntries.includes( cliBinPath.toLowerCase() ) ) {
+				env = { ...process.env, Path: `${ cliBinPath };${ currentPath }` };
```
```suggestion
				delete env.PATH;
				delete env.Path;
				env.PATH = `${ cliBinPath };${ currentPath }` };
```
On the flip side, per the [Node.js child_process docs](https://nodejs.org/docs/latest-v22.x/api/child_process.html#child-process):

> On Windows, environment variables are case-insensitive. Node.js lexicographically sorts the env keys and uses the first one that case-insensitively matches. Only first (in lexicographic order) entry will be passed to the subprocess. This might lead to issues on Windows when passing objects to the env option that have multiple variants of the same key, such as PATH and Path.

src/ipc-handlers.ts
```diff
@@ -828,8 +828,22 @@ export async function openTerminalAtPath( _event: IpcMainInvokeEvent, targetPath
 			return promiseExec( `start "" "warp://action/new_tab?path=${ encodedPath }"` );
 		}
 
+		// Ensure the Studio CLI bin directory is in the PATH for the spawned terminal.
+		// Child processes inherit the environment from the Electron process, which may have
+		// been started before the CLI was installed or PATH was updated in the registry.
+		const cliBinPath = windowsHelpers.getWindowsCliBinPath();
+		let env: NodeJS.ProcessEnv | undefined;
+		if ( cliBinPath ) {
```
fredrikekelund: Should we run this logic even if the user hasn't installed the CLI? I'm probably onboard with that, but I still wanted to bring this up for discussion

src/ipc-handlers.ts
epeicher: TIL, thanks for the reference!

src/lib/windows-helpers.ts
epeicher: Are we OK with referencing `modules/cli/lib` from `lib`? I think [exporting the const](https://github.com/Automattic/studio/blob/df193de9876e4f561bab235813f55955cc8ce386/src/modules/cli/lib/windows-installation-manager.ts#L12) could be enough for this scenario.

src/ipc-handlers.ts
epeicher: I think it is better to have the guard there so we don't run that code if `cliBinPath` is empty

src/ipc-handlers.ts
epeicher: Applied as part of [8059bd7](https://github.com/Automattic/studio/pull/2462/commits/8059bd78444f63948da866f89e0b66c45cea56ef)

src/ipc-handlers.ts
epeicher: Updated as part of [8059bd7](https://github.com/Automattic/studio/pull/2462/commits/8059bd78444f63948da866f89e0b66c45cea56ef)

src/lib/windows-helpers.ts
epeicher: I have added the change as part of [8059bd7](https://github.com/Automattic/studio/pull/2462/commits/8059bd78444f63948da866f89e0b66c45cea56ef). I went for the `export const` approach, but if you think it is better to have a static getter, please let me know and I will update it

fredrikekelund: > Are we OK with referencing modules/cli/lib from lib

Yes. Let's have a single canonical reference to the CLI install path. 

> I think [exporting the const](https://github.com/Automattic/studio/blob/df193de9876e4f561bab235813f55955cc8ce386/src/modules/cli/lib/windows-installation-manager.ts#L12) could be enough for this scenario.

Makes sense 👍 https://github.com/Automattic/studio/pull/2463 renames this variable, so I was mostly thinking of merge conflicts, but they should be easy to resolve anyway.

src/ipc-handlers.ts
fredrikekelund: What I'm saying is that the logic currently runs regardless of whether the user has enabled the "Install CLI" setting. That's probably a good call, but I still wanted to make sure it's intentional

src/lib/windows-helpers.ts
```diff
@@ -3,7 +3,23 @@ import path from 'path';
 import sudo from '@vscode/sudo-prompt';
 import { __ } from '@wordpress/i18n';
 import { getMainWindow } from 'src/main-window';
+import { unversionedBinDirPath } from 'src/modules/cli/lib/windows-installation-manager';
```
fredrikekelund: Would it not make more sense to import this directly in `src/ipc-handlers.ts`?

epeicher: That makes sense, now this function does not provide that much value so we can get rid of changes in the component and import directly the `unversionedBinDirPath` constant in `ipc-handlers` directly. I will do that

src/ipc-handlers.ts
epeicher: Ah, sorry for the confusion. It was not intentional, but I think it would be better to add it for the same reason above, to avoid running that code unnecessarily

src/lib/windows-helpers.ts
epeicher: Removed changes in this file as they are not required anymore, the constant will be imported directly from `ipc-handlers`. Done as part of aae78866

src/ipc-handlers.ts
epeicher: Done as part of aae78866, let me know if you are OK with that

src/lib/windows-helpers.ts
epeicher: Added as part of aae78866

src/ipc-handlers.ts
```diff
@@ -828,8 +830,26 @@ export async function openTerminalAtPath( _event: IpcMainInvokeEvent, targetPath
...
+		if ( isCliInstalled && cliBinPath ) {
+			const currentPath = process.env.PATH || '';
+			const pathEntries = currentPath.split( ';' ).map( ( p ) => p.toLowerCase() );
+			if ( ! pathEntries.includes( cliBinPath.toLowerCase() ) ) {
+				env = { ...process.env };
+				delete env.PATH;
+				delete env.Path;
+				env.PATH = `${ cliBinPath };${ currentPath }`;
+			}
+		}
```
```suggestion
		const isCliInstalled = await isStudioCliInstalled();
		let env: NodeJS.ProcessEnv | undefined;
		if ( isCliInstalled ) {
			const currentPath = process.env.PATH || '';
			const pathEntries = currentPath.split( ';' ).map( ( p ) => p.toLowerCase() );
			if ( ! pathEntries.includes( unversionedBinDirPath.toLowerCase() ) ) {
				env = { ...process.env };
				delete env.PATH;
				delete env.Path;
				env.PATH = `${ unversionedBinDirPath };${ currentPath }`;
			}
		}
```
`unversionedBinDirPath` is static. No need to check if it's empty.

epeicher: True b9fdc357

src/ipc-handlers.ts
```diff
@@ -828,8 +830,25 @@ export async function openTerminalAtPath( _event: IpcMainInvokeEvent, targetPath
 			return promiseExec( `start "" "warp://action/new_tab?path=${ encodedPath }"` );
 		}
 
+		// Ensure the Studio CLI bin directory is in the PATH for the spawned terminal.
+		// Child processes inherit the environment from the Electron process, which may have
+		// been started before the CLI was installed or PATH was updated in the registry.
+		const isCliInstalled = await isStudioCliInstalled();
```
fredrikekelund, line 836:
```suggestion
		const manager = new WindowsCliInstallationManager();
		const isCliInstalled = await manager.updateWindowsCliVersionedPathIfNeeded();
```
Optional nit, but I figured that since we know we're only dealing with Windows here, we could use this method directly. It'll make the imports slightly cleaner.

fredrikekelund (approved):
This looks good and tests well 👍 I left one comment about an optional final adjustment we could make. Up to you, @epeicher

src/ipc-handlers.ts
epeicher: But this method `updateWindowsCliVersionedPathIfNeeded` [installs the bat file](https://github.com/Automattic/studio/blob/b9fdc3579e7abb552f42451aca180c13dcfdffab/src/modules/cli/lib/windows-installation-manager.ts#L56) if the Cli is installed, right? I personally think it is clearer to just check using `isStudioCliInstalled()`

PR #2463, opened 2026-01-23
Make Windows CLI installation tolerate app reinstalls

fredrikekelund:
## Related issues

- Fixes STU-1247

## Proposed Changes

> [!NOTE]  
> From the user's perspective, I don't believe it's completely necessary to change this behavior. However, it will make the CLI installation more resilient in our internal smoke testing, which seems like a noble cause in and of itself.

When installing the CLI on Windows, `WindowsCliInstallationManager` does two things:
1. It adds `C:\Users\<USERNAME>\AppData\Local\studio\bin` to the `PATH` variable in the registry
2. It creates the `C:\Users\<USERNAME>\AppData\Local\studio\bin` directory and writes a `studio.bat` file to it

The reason we do it this way is that the app resources live in a versioned directory (i.e. `C:\Users\<USERNAME>\AppData\Local\studio\app-1.7.0-beta3`). Instead of changing the `PATH` variable in the registry every time the app starts, we use a known, stable path and change the contents of `C:\Users\<USERNAME>\AppData\Local\studio\bin\studio.bat` when the app starts.

This logic plays nicely with the auto-updater. However, it breaks when the app is **manually reinstalled**. When that happens, the `PATH` variable in the registry isn't touched, but the entire `C:\Users\<USERNAME>\AppData\Local\studio` directory is wiped, effectively uninstalling the CLI.

The `WindowsCliInstallationManager::updateWindowsCliVersionedPathIfNeeded` logic (which runs when the app starts and updates the contents of `C:\Users\<USERNAME>\AppData\Local\studio\bin\studio.bat`) relies on the presence of the `C:\Users\<USERNAME>\AppData\Local\studio` directory. This PR changes that method to only look for `C:\Users\<USERNAME>\AppData\Local\studio\bin` in the registry PATH and proceed with writing the `studio.bat` script if the registry key is there. 

## Testing Instructions

1. Have Studio installed on Windows
2. Have the Studio CLI installed
3. Download the Windows dev build
4. Install it
5. Ensure that the settings modal still shows the CLI to be installed
6. Open a command prompt using the button in the Overview tab
7. Ensure that the `studio` command is available in the command prompt

## Pre-merge Checklist

### DIFF
--- a/src/modules/cli/lib/windows-installation-manager.ts
+++ b/src/modules/cli/lib/windows-installation-manager.ts
@@ -8,7 +8,7 @@ import Registry from 'winreg'; // don't update winreg to 1.2.5 - https://github.
 import { getMainWindow } from 'src/main-window';
 import { StudioCliInstallationManager } from 'src/modules/cli/lib/ipc-handlers';
 
-// `stableBinDirPath` resolves to C:\Users\<USERNAME>\AppData\Local\studio\bin
+// `STABLE_BIN_DIR_PATH` resolves to C:\Users\<USERNAME>\AppData\Local\studio\bin
 export const STABLE_BIN_DIR_PATH = path.resolve( path.dirname( app.getPath( 'exe' ) ), '../bin' );
 const PATH_KEY = 'Path';
 
--- a/src/modules/cli/lib/windows-installation-manager.ts
+++ b/src/modules/cli/lib/windows-installation-manager.ts
@@ -29,25 +29,8 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
 	 */
 	async isCliInstalled(): Promise< boolean > {
 		try {
-			const currentPath = await this.getPathFromRegistry();
-
-			// Return true if we are running the development version of the app and the production CLI is installed
-			if ( process.env.NODE_ENV !== 'production' && process.env.LOCALAPPDATA ) {
-				const prodStudioCliDir = path.join( process.env.LOCALAPPDATA, 'studio', 'bin' );
-				if ( this.isStudioCliInPath( currentPath, prodStudioCliDir ) ) {
-					return true;
-				}
-			}
-
-			if ( ! this.isStudioCliInPath( currentPath ) ) {
-				return false;
-			}
-
-			if ( ! existsSync( stableBinDirPath ) ) {
-				return false;
-			}
-
-			return true;
+			const isStudioCliDirInPath = await this.isStudioCliDirInPath();
+			return isStudioCliDirInPath && existsSync( stableBinDirPath );
 		} catch ( error ) {
 			console.error( 'Failed to check installation status of CLI', error );
 			return false;
@@ -55,8 +38,7 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
 	}
 
 	async updateWindowsCliVersionedPathIfNeeded(): Promise< void > {
-		const currentPath = await this.getPathFromRegistry();
-		if ( this.isStudioCliInPath( currentPath ) ) {
+		if ( await this.isStudioCliDirInPath() ) {
 			await this.installProxyBatFile();
 		}
 	}
@@ -145,8 +127,16 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
 		} );
 	}
 
-	private isStudioCliInPath( pathValue: string, studioCliDir: string = stableBinDirPath ): boolean {
-		return pathValue
+	private async isStudioCliDirInPath(): Promise< boolean > {
+		let studioCliDir = stableBinDirPath;
+
+		// Return true if we are running the development version of the app and the production CLI is installed
+		if ( process.env.NODE_ENV !== 'production' && process.env.LOCALAPPDATA ) {
+			studioCliDir = path.join( process.env.LOCALAPPDATA, 'studio', 'bin' );
+		}
+
+		const currentPath = await this.getPathFromRegistry();
+		return currentPath
 			.split( ';' )
 			.map( ( item ) => item.trim().toLowerCase() )
 			.includes( studioCliDir.toLowerCase() );
@@ -155,11 +145,6 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
 	private async installPath(): Promise< void > {
 		try {
 			const currentPath = await this.getPathFromRegistry();
-
-			if ( this.isStudioCliInPath( currentPath ) ) {
-				return;
-			}
-
 			const updatedPath = currentPath
 				.split( ';' )
 				.map( ( p ) => p.trim() )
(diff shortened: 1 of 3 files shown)


src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -8,8 +8,8 @@ import Registry from 'winreg'; // don't update winreg to 1.2.5 - https://github.
 import { getMainWindow } from 'src/main-window';
 import { StudioCliInstallationManager } from 'src/modules/cli/lib/ipc-handlers';
 
-// `unversionedBinDirPath` resolves to C:\Users\<USERNAME>\AppData\Local\studio\bin
-const unversionedBinDirPath = path.resolve( path.dirname( app.getPath( 'exe' ) ), '../bin' );
+// `stableBinDirPath` resolves to C:\Users\<USERNAME>\AppData\Local\studio\bin
+const stableBinDirPath = path.resolve( path.dirname( app.getPath( 'exe' ) ), '../bin' );
```
fredrikekelund: I found this name easier to understand.

src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -154,16 +145,11 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
 	private async installPath(): Promise< void > {
 		try {
 			const currentPath = await this.getPathFromRegistry();
-
-			if ( this.isStudioCliInPath( currentPath ) ) {
-				return;
-			}
```
fredrikekelund, line 160: This check is redundant. Better to do this before calling `WindowsCliInstallationManager::installPath`

src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -24,35 +24,21 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
...
+			return isStudioCliDirInPath && existsSync( stableBinDirPath );
 		} catch ( error ) {
 			console.error( 'Failed to check installation status of CLI', error );
 			return false;
 		}
 	}
 
 	async updateWindowsCliVersionedPathIfNeeded(): Promise< void > {
-		if ( await this.isCliInstalled() ) {
+		if ( await this.isStudioCliDirInPath() ) {
```
fredrikekelund, line 41: This is really the core change I made: instead of checking both whether the CLI path is added to the registry PATH and whether the CLI path exists, we only check the registry PATH.

src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -24,35 +24,21 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
...
 		try {
-			const currentPath = await this.getPathFromRegistry();
-
-			// Return true if we are running the development version of the app and the production CLI is installed
-			if ( process.env.NODE_ENV !== 'production' && process.env.LOCALAPPDATA ) {
-				const prodStudioCliDir = path.join( process.env.LOCALAPPDATA, 'studio', 'bin' );
-				if ( this.isStudioCliInPath( currentPath, prodStudioCliDir ) ) {
-					return true;
-				}
-			}
```
fredrikekelund, line 37: I moved this logic to the `isStudioCliDirInPath` method.

ivan-ottinger (approved):
Thank you for including detailed description and comments in the code. That helped.

The changes look good to me and testing steps work as expected. CLI stayed operational after app update. I did not observe any regressions.

ivan-ottinger:
~Wait, I just realized I tested with incorrect dev build.~

Update: Sorry, false alarm, I just got confused by the version in the About window. All is good. 👍🏼

gcsecsey (commented):
The changes make sense to me, but I ran into an issue when testing this on Parallels.

After following these steps:
- Installed Studio 1.6.8 release
- Installed the CLI through the modal
- Checked that the `C:\Users\<USERNAME>\AppData\Local\studio\bin` directory is appended to `PATH` and that the `studio` command works
- Installed the dev build
- Checked the CLI status on the modal

I can see the modal showing CLI as not installed. However, clicking the "Command Prompt" button then invoking `studio` command works as expected, the `C:\Users\<USERNAME>\AppData\Local\studio\bin` still exists and it's in `PATH`.

Image

fredrikekelund:
Hmm, let me test that same flow to check, @gcsecsey

gcsecsey:
> The changes make sense to me, but I ran into an issue when testing this on Parallels.
> 
> After following these steps:
> 
> * Installed Studio 1.6.8 release
> * Installed the CLI through the modal
> * Checked that the `C:\Users\<USERNAME>\AppData\Local\studio\bin` directory is appended to `PATH` and that the `studio` command works
> * Installed the dev build
> * Checked the CLI status on the modal
> 
> I can see the modal showing CLI as not installed. However, clicking the "Command Prompt" button then invoking `studio` command works as expected, the `C:\Users\<USERNAME>\AppData\Local\studio\bin` still exists and it's in `PATH`.
> 
> Image

I installed the dev build using the .appx image, maybe that's the issue? I can see in my PATH that it has populated some new directories:

```
PS C:\Users\gcsecsey> $env:Path
C:\Users\gcsecsey\AppData\Local\fnm_multishells\10092_1769169635781;C:\Program Files\Parallels\Parallels Tools\Applications;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\WINDOWS\System32\OpenSSH\;C:\ProgramData\chocolatey\bin;C:\Program Files\Git\cmd;C:\Users\gcsecsey\AppData\Local\Microsoft\WindowsApps;C:\Users\gcsecsey\AppData\Local\Programs\Microsoft VS Code\bin;C:\Users\gcsecsey\AppData\Local\Programs\cursor\resources\app\bin;C:\Program Files\WindowsApps\22490Automattic.StudiobyWordPress.com_1.6.3.0_arm64__9qb1eemr42qj0\bin;C:\Users\gcsecsey\AppData\Local\Programs\Warp\bin;C:\Users\gcsecsey\AppData\Local\studio\bin;C:\Program Files\WindowsApps\22490Automattic.StudiobyWordPress.com_1.6.8.0_arm64__9qb1eemr42qj0\bin;C:\Program Files\WindowsApps\22490Automattic.StudiobyWordPress.com_1.6.8.0_arm64__9qb1eemr42qj0\bin;C:\Program Files\WindowsApps\22490Automattic.StudiobyWordPress.com_1.6.8.0_arm64__9qb1eemr42qj0\bin
PS C:\Users\gcsecsey> ls C:\Users\gcsecsey\AppData\Local\studio\bin


    Directory: C:\Users\gcsecsey\AppData\Local\studio\bin


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----        23/01/2026     11:52             62 studio.bat
```

src/modules/cli/lib/windows-installation-manager.ts
fredrikekelund: I take it back. Without this check, we might install the same path multiple times

fredrikekelund:
We're investigating the issue @gcsecsey uncovered here ☝️

It doesn't seem related to the changes in this PR, though, so we're landing this first and potentially following up with another PR for the AppX issue.

PR #2465, opened 2026-01-23
Always use LOCALAPPDATA to generate Windows CLI install path

fredrikekelund:
## Related issues

- Fixes STU-1249

## Proposed Changes

As noted by @gcsecsey in https://github.com/Automattic/studio/pull/2463#issuecomment-3789927177, CLI installation doesn't work when the app was installed using an AppX installer on Windows.

When Studio installs the CLI on Windows, it writes a "proxy `studio.bat` file" to a directory relative to its installation directory (as determined by the value Electron returns from `app.getPath( 'exe' )`. However, Microsoft Store installs Studio in a directory like `C:\Program Files\WindowsApps\22490Automattic.StudiobyWordPress.com_1.6.8.0_arm64__9h07f78gwnchp` that has heavy read and write limitations. Therefore, we believe the CLI installation fails because Studio cannot write the file.

This PR solves the problem by always writing the proxy batch file to a known stable location in `%LOCALAPPDATA%`.

## Testing Instructions

I haven't figured out how to install an AppX dev build on Windows, so code review might have to suffice

## Pre-merge Checklist


fredrikekelund:
[This Sentry issue](https://a8c.sentry.io/issues/7064963192/?environment=production&query=&referrer=issue-stream) confirms our theory. It was triggered by me and the error message reads:

```
Error: EPERM: operation not permitted, mkdir 'C:\Program Files\WindowsApps\22490Automattic.StudiobyWordPress.com_1.6.8.0_arm64__9h07f78gwnchp\bin'
```

apps/studio/src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -24,13 +21,21 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
 		}
 	}
 
+	getStableBinDirPath() {
+		if ( ! process.env.LOCALAPPDATA ) {
+			throw new Error( 'LOCALAPPDATA environment variable is not set' );
+		}
```
fredrikekelund, line 27: An obvious question is, how do we know that `process.env.LOCALAPPDATA` is defined? Well, there's no official guarantee, and I can't find any Microsoft documentation on this, but it appears the system sets this variable for each process. It's not editable by users either, AFAICT. That's probably as good a guarantee as we'll get.

Still, it makes sense to ensure that we handle errors thrown from here gracefully, so I'll take a look at that now…

fredrikekelund: Done in b942405

src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -172,20 +182,18 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
...
+			await mkdir( this.getStableBinDirPath(), { recursive: true } );
 
 			const versionedCliPath = path.join(
 				path.dirname( app.getPath( 'exe' ) ),
 				'resources/bin/studio-cli.bat'
 			);
-			const relativeVersionedCliPath = path.relative( STABLE_BIN_DIR_PATH, versionedCliPath );
-
-			const content = `@echo off\n"%~dp0\\${ relativeVersionedCliPath }" %*`;
+			const content = `@echo off\n"%~dp0\\${ versionedCliPath }" %*`;
```
gcsecsey: In the earlier version we used `relativeVersionedCliPath` but now we're using `versionedCliPath` which seems to be an absolute path.

I think appending the absolute path to the `%~dp0` variable will produce an invalid path.

fredrikekelund: Thanks for being alert, @gcsecsey 🙏

fredrikekelund: I've updated this to use just the absolute path. It seemed unnecessary to me to use a relative path

gcsecsey (commented):
This LGTM and I can consistently see the toggle working correctly now, both on Parallels and on x64. 👍 

When trying to open a Command Prompt using the button on the site details page, I see an `Access denied` message when I try to use the `studio` command.

Image

Opening a new tab and using the command in a different directory also prints the same message:

Image

The contents of the bat file seems correct to me, I think the issue might be that we don't have rights to run the batch file under `WindowsApps`:

```batch
@echo off
"C:\Program Files\WindowsApps\22490Automattic.StudiobyWordPress.com_1.6.8.0_arm64__9qb1eemr42qj0\app\resources\bin\studio-cli.bat" %*
```

fredrikekelund:
> I think the issue might be that we don't have rights to run the batch file under WindowsApps

That seems like a good theory. We probably lack permission to execute that file (but we can read it).

I'm not immediately sure what the best way to solve that is…

gcsecsey:
@fredrikekelund I dug a bit deeper into this, and added a commit that fixes the CLI for the AppX version as far as I could test.

The Windows-recommended solution is https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-uap5-appexecutionalias. When declared in the manifest, this creates a lightweight stub in `%LOCALAPPDATA%\Microsoft\WindowsApps\`, which activates the package through the container APIs. The big downside is that this approach only supports .exe files, so we need to create/maintain a proxy binary for AppX versions only.

The commit adds:
  - `bin/studio-cli-launcher.js` — a small script that does what studio-cli.bat does (finds bundled node.exe and cli/main.js, spawns the CLI). This is compiled to a standalone .exe via @yao-pkg/pkg during the prePackage build step.
  - `scripts/package-appx.mjs` — adds the uap5:AppExecutionAlias extension to the AppxManifest, pointing at studio-cli.exe with alias studio.exe.

I could test it on Windows by removing all previous installations and also uninstalling the CLI from the dev folder using the Settings dialog. Then, downloading the appx artifact from the build and installing it, invoking the `studio` command from the terminal works as expected.

The `where` command now prints the path of the main executable, but the CLI features still seem to work normally, even when Studio is not launched:
CleanShot 2026-02-04 at 15 32 03@2x

To package the exe, I also needed to add a new dependency. The alternative would be to use Node's [Single Executable Application](https://nodejs.org/api/single-executable-applications.html) feature to package the .exe. However, it would produce a larger binary (~70-80MB vs ~40-50MB from pkg) and requires a more complex multi-step build (generate blob, copy node.exe, inject blob, remove signature, re-sign). Still, it could be worth revisiting if we want to drop the @yao-pkg/pkg dependency.

gcsecsey:
I tested this approach further, and realized that the AppXManifest approach makes the CLI instantly available when the app is installed, so the %LOCALAPPDATA% fix is not tightly related to it. I added conditional rendering to the toggle and raised #2547 with these changes separately.

I think the `%LOCALAPPDATA%` fix would still make sense for the installer version, if the app can be installed in different locations, as it uses absolute paths instead of relative ones. I'll restore this PR's changes to only contain this fix.

fredrikekelund:
Since @gcsecsey already solved the AppX installer issue differently (and properly) in https://github.com/Automattic/studio/pull/2547, I don't really see a reason to proceed with this change. `WindowsCliInstallationManager` will still depend on `app.getPath` from Electron, and there's no apparent reason to push hard to decouple it from it. 

So, I'll go ahead and close this PR 👍

PR #2459, opened 2026-01-22
Update site selection behavior after site deletion

gcsecsey:
- Update site selection logic to preserve the currently selected site if it still exists after deletion.
- Add tests to verify that the first remaining site is selected when the currently selected site is deleted, and that selection remains unchanged when a different site is deleted.

## Related issues

- Fixes STU-1241

## Proposed Changes

- Updated `onDeleteSite` to check if the currently selected site still exists after deletion
- If the selected site still exists (user switched to another site), the selection is preserved
- If the selected site was deleted, it switches to the first remaining site
- Added unit tests to verify both scenarios

## Testing Instructions

1. Start Studio with `npm start`
2. Create three sites (Site 1, Site 2, Site 3)
3. Delete Site 3
4. While the deletion is going on, quickly select Site 2
5. Check that Site 2 remains selected, and the selection doesn't change to Site 1

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2468, opened 2026-01-23
Update tab selection logic after site deletion

gcsecsey:
## Related issues

- Fixes STU-1241
- Follow-up to #2459

## Proposed Changes

When the deleted site was the selected site, the tab should reset to "Overview". However, calling `setSelectedTab('overview')` from the site details hook didn't work reliably. The `ContentTabsProvider` wraps `SiteDetailsProvider`, so when both states update, the site change renders the component before the tab update takes effect, causing the TabPanel to remount with the old tab value.

This PR moves the state update call from the `onDeleteSite` callback to `SiteContentTabs`, and adds logic to compute an `effectiveTab` value at render time. The component tracks the previous site ID in state and, during render, checks if the site changed AND the previous site was deleted. If both conditions are true, it uses `'overview'` as the effective tab for `initialTabName`. A useEffect then syncs this back to the tab context.

## Testing Instructions

1. Start Studio with `npm start`
2. Create three sites (Site 1, Site 2, Site 3)
3. Delete Site 3
4. While the deletion is going on, quickly select Site 2 and switch to the "Sync" tab or any other tab
6. Check that after the deletion, Site 2 and the previously selected tab remain selected
7. Delete Site 2, and stay on the Settings tab until the deletion is completed
8. Check that Site 1 on the Overview tab is selected after the deletion

https://github.com/user-attachments/assets/35ffc9dc-cfc0-4019-a0bf-ad82d8e486ef

## Pre-merge Checklist

### DIFF
--- a/src/components/site-content-tabs.tsx
+++ b/src/components/site-content-tabs.tsx
@@ -9,45 +9,26 @@ import { ContentTabSettings } from 'src/components/content-tab-settings';
 import Header from 'src/components/header';
 import { SiteIsBeingCreated } from 'src/components/site-is-being-created';
 import { MIN_WIDTH_CLASS_TO_MEASURE } from 'src/constants';
-import { TabName, useContentTabs } from 'src/hooks/use-content-tabs';
+import { TabName } from 'src/hooks/use-content-tabs';
+import { useEffectiveTab } from 'src/hooks/use-effective-tab';
 import { useImportExport } from 'src/hooks/use-import-export';
 import { useSiteDetails } from 'src/hooks/use-site-details';
 import { cx } from 'src/lib/cx';
 import { ContentTabSync } from 'src/modules/sync';
 
 export function SiteContentTabs() {
-	const { selectedSite, sites, siteCreationMessages } = useSiteDetails();
+	const { selectedSite, siteCreationMessages } = useSiteDetails();
 	const { importState } = useImportExport();
-	const { tabs, selectedTab, setSelectedTab } = useContentTabs();
+	const { effectiveTab, selectedTab, setSelectedTab, tabs } = useEffectiveTab();
 	const { __ } = useI18n();
 
 	// Remount: Avoid focus loss on user tab changes (no remount),
 	// but remount on programmatic changes and site switches so initial tab/content state resets.
 	const [ keyCounter, setKeyCounter ] = useState( 0 );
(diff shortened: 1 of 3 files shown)


sejas (commented):
I tested it and it mostly worked, the only issue is that my selected site changed from one existing site to another automatically.
Steps to reproduce:
- Delete site 3
- Quickly switch to site 2
- Observe that Studio switches automatically to the first site instead of keeping you in the existing site 2.

Here is a recording:

https://github.com/user-attachments/assets/efe754dc-94e7-4e4e-a8fc-4f97616e7dc9

sejas (commented):
@gcsecsey , thanks for making the changes. I tested it and it works as expected, althoug I think we added more changes in site-content-tabs whcih I think are not necesssary.

I created this alternate PR using yours as base and it seems to work. https://github.com/Automattic/studio/pull/2480

src/components/site-content-tabs.tsx
```diff
@@ -29,6 +29,26 @@ export function SiteContentTabs() {
...
+	// Update prevSiteId selectedTab after render
+	useEffect( () => {
+		if ( siteChanged ) {
+			setPrevSiteId( selectedSite?.id );
+		}
+
+		if ( effectiveTab !== selectedTab ) {
+			setSelectedTab( effectiveTab as TabName );
+		}
+	}, [ siteChanged, selectedSite?.id, effectiveTab, selectedTab, setSelectedTab ] );
```
sejas: Maybe we could isolate these new hooks into separate custom hook.

sejas (approved):
Works as expected! Great work!

src/components/site-content-tabs.tsx
gcsecsey: Thanks @sejas that's a good suggestion. I moved it to a new hook in [27d6e4d](https://github.com/Automattic/studio/pull/2468/commits/27d6e4def44a1a631fb39de86c6a016beb59bbe5).

epeicher (approved):
Thanks @gcsecsey for solving this! I have tested it, and it works as expected. The site selected and the tab selected remains after the site deletion completes. I have also confirmed that if no other sites are selected, the first site in the Overview tab is opened. Changes also LGTM! :shipit:

sejas (approved):

PR #2470, opened 2026-01-23
Allow enableMultisite blueprint step

bcotrim:
## Related issues

- Fixes STU-831

### Allow `enableMultisite` blueprint step
- Remove `enableMultisite` from the unsupported blueprint features list, allowing blueprints to configure WordPress multisite

### Bump `@wp-playground` packages to 3.0.54
- Includes upstream fix for `$_SERVER['HTTP_HOST']` handling in multisite, which fixes site restart failures

### Add multisite validation (CLI + Desktop)

**CLI:** Error early when a blueprint contains `enableMultisite` but no `--domain` flag is provided. Multisite doesn't support custom ports, so a custom domain is required.

**Desktop app:** When a blueprint with `enableMultisite` is detected:
- Auto-enable the custom domain field and show advanced settings
- Lock the custom domain toggle so users can't disable it
- Show an info notice: _"WordPress multisite requires a custom domain."_

### Shared utility
- Add `blueprintHasMultisite()` helper in `common/lib/blueprint-settings.ts`, used by both CLI and desktop

### CLI
1. Create a blueprint JSON with an `enableMultisite` step
2. Run `npm run cli:build && node dist/cli/main.js site create ~/test-multisite --blueprint /tmp/multisite.json`
3. Verify it errors with a message about requiring `--domain`
4. Run again with `--domain test.local` — should proceed normally

### Desktop
1. Open Studio, go to Add Site
2. Use a blueprint URL containing `enableMultisite` (e.g., via deeplink)
3. Verify advanced settings auto-open, custom domain is enabled and locked
4. Verify the info notice about multisite requiring a custom domain is shown

### Tests
```
npm test -- common/lib/tests/blueprint-settings.test.ts cli/commands/site/tests/create.test.ts
```

## Pre-merge Checklist


sejas (commented):
I tested this PR and I confirm I was able to create a new site using a blueprint with `enableMultisite` step. Feature wise it behaves as described in the PR.
I have two concerns that could be raised by users once we make multisites easier to create.
1. AFAIK, Users won't be able to Push these sites to WordPress.com and only [some Pressable sites](https://pressable.com/knowledgebase/will-my-multisite-work-on-pressable/) support it. I wonder if we should at least keep a warning for this step.
2. There are some warnings visible in wp-admin:
```
Warning: file_get_contents(/wordpress/wp-content/plugins/{SQLITE_PLUGIN}): Failed to open stream: No such file or directory in /wordpress/wp-includes/functions.php on line 6910

Warning: file_get_contents(/wordpress/wp-content/plugins/sqlite-database-integration/load.php): Failed to open stream: No such file or directory in /wordpress/wp-includes/functions.php on line 6910
```
3. The assumption of a custom domain being required is false. I was able to setup a Multisite using localhost:8881 port.

Blueprint I used for testing: [enableMultisite-blueprint.json](https://github.com/user-attachments/files/25450427/enableMultisite-blueprint.json)

**Step is allowed and the custom domain is forced:**

Image
Image

**wp-admin warnings:** 

Image

**Multisite using localhost**

I noticed that the site created with the blueprint added this first line in the wp-config.php : `$_SERVER['HTTP_HOST'] = json_decode(base64_decode('InRlc3QtbXUtc2l0ZS53cC5sb2NhbCI='), true);`, so I used a similar setting in my localhost Multisite and worked great. I was able to restart the server without errors in Studio:

```
/* Multisite */
$_SERVER["HTTP_HOST"] = "localhost:8886";
define("WP_ALLOW_MULTISITE", true);
define("MULTISITE", true);
define("SUBDOMAIN_INSTALL", false);
define("DOMAIN_CURRENT_SITE", "localhost:8886");
define("PATH_CURRENT_SITE", "/");
define("SITE_ID_CURRENT_SITE", 1);
define("BLOG_ID_CURRENT_SITE", 1);
```

@bcotrim, should we allow users using localhost:port?

Image

wojtekn:
> AFAIK, Users won't be able to Push these sites to WordPress.com and only [some Pressable sites](https://pressable.com/knowledgebase/will-my-multisite-work-on-pressable/) support it. I wonder if we should at least keep a warning for this step.

@sejas let's keep the UI clean, the purpose of this PR is to avoid blocking features that can work in Studio and to reduce the number of unnecessary warnings.

bcotrim:
> The assumption of a custom domain being required is false. I was able to setup a Multisite using localhost:8881 port.

Were you able to start a site in Playground CLI without a domain? I didn't introduce that restriction, on my tests, creating a multisite using Playground CLI blueprint would result in an error if a domain wasn't provided.

apps/studio/src/modules/add-site/components/create-site-form.tsx
```diff
@@ -227,6 +230,14 @@ export const CreateSiteForm = ( {
 		setAdvancedSettingsVisible( true );
 	}, [ blueprintSuggestedDomain, blueprintSuggestedHttps ] );
 
+	useEffect( () => {
+		if ( ! blueprintRequiresCustomDomain ) {
+			return;
+		}
+		setUseCustomDomain( true );
```
katinthehatsite, line 269: We could consider auto-generating the suggested custom domain for users e.g. something like `my-site.local` as an option and then users who want to modify it, could change them further

apps/studio/src/modules/add-site/index.tsx
```diff
@@ -113,6 +115,8 @@ function NavigationContent( props: NavigationContentProps ) {
 		setBlueprintSuggestedHttps,
 		blueprintSuggestedSiteName,
 		setBlueprintSuggestedSiteName,
+		blueprintRequiresCustomDomain,
+		setBlueprintRequiresCustomDomain,
```
katinthehatsite, line 121: Some of the logic here is similar to the logic in `use-blueprint-deeplink.ts `. This is not specifically related to this PR but we could refactor it in some kind of shared helper. What do you think?

apps/studio/src/modules/add-site/components/create-site-form.tsx
katinthehatsite: Oh I see that it already worked this way for me so you can ignore the comment 🙈

katinthehatsite (commented):
In terms of functionality, I did not find any issues: I was able to create another site from this site in `My sites` in wp-admin, and it worked great. Nothing crashed on my end 🤞 

The only thing the I saw were the warning which I think it would be nice if they were resolved:

Image

katinthehatsite (approved):
thanks for fixing the warnings @bcotrim  - I no longer see them:

Image

It would be nice if this notice had a background or so to stand out more but I think it is a minor design thing: 

Image

bcotrim:
thanks for the review @katinthehatsite
I changed the notice from "info" to "warning" does this look better?
image

@sejas do you agree to land this with the domain restriction, or should we try to look more into it, and check with Playground if there's a way we can allow sites without a custom domain?

sejas (approved):
Sure! Let’s land the PR with the domain restriction. We can always remove it in the future if we find an easier way to handle it.

Currently, this approach hardcodes a constant in wp-config.php, which may cause issues in production, etc. But it’s a small thing, considering that multisites are not supported by many wp.cloud providers. I would suggest creating an issue to disable the publish and connect buttons for multisites.

bcotrim:
> Currently, this approach hardcodes a constant in wp-config.php, which may cause issues in production, etc. But it’s a small thing, considering that multisites are not supported by many wp.cloud providers. I would suggest creating an issue to disable the publish and connect buttons for multisites.

Done STU-1377, thank you!

tools/common/lib/sqlite-integration.ts
```diff
@@ -49,10 +49,15 @@ export abstract class SqliteIntegrationProvider {
...
-			`realpath( __DIR__ . '/mu-plugins/${ sqliteDirname }' )`
-		);
+		const updatedContent = dbCopyContent
+			.replace(
+				"'{SQLITE_IMPLEMENTATION_FOLDER_PATH}'",
+				`realpath( __DIR__ . '/mu-plugins/${ sqliteDirname }' )`
+			)
+			.replace(
+				"define( 'SQLITE_DB_DROPIN_VERSION'",
+				`define( 'SQLITE_MAIN_FILE', __FILE__ );\ndefine( 'SQLITE_DB_DROPIN_VERSION'`
```
wojtekn, line 59: @bcotrim what was the reason to define SQLITE_MAIN_FILE here? CC @JanJakes

bcotrim: it was to fix the errors reported by @katinthehatsite here https://github.com/Automattic/studio/pull/2470#pullrequestreview-3853309658

wojtekn: Thanks. I will remove this change in https://github.com/Automattic/studio/pull/2834 as it redefines this constant in the incorrect path, leading to issues in other places.

It may be that multisite loads plugins differently, and we will need to address those warnings differently. Probably the code that handfles `performance-lab` in `db.php` should be rewritten, as now it assumes SQLite plugin lives in plugins dir.

PR #2472, opened 2026-01-26
Fix excessive WP-CLI command invocations in Assistant tab

fredrikekelund:
## Related issues

- Fixes STU-1254

## Proposed Changes

The 1.7.0 release introduced a bug that caused an excessive number of WP-CLI invocations in the Assistant tab (triggered by the `updateFromSite` thunk). This led to the commands failing and showing up as JSON parsing errors in the browser console. This seems to be caused by mutations of the `selectedSite` object returned by the `useSiteDetails` hook. 

This PR fixes the issue by changing the dependency array of the `useEffect` hook that triggers the `updateFromSite` thunk, and by adding a `condition` config to the thunk itself that prevents parallel invocations. 

## Testing Instructions

1. Run `npm start`
2. Open the Assistant tab
3. Send a message
4. Ensure that there are no `Unexpected end of JSON input` messages in the devtools console

## Pre-merge Checklist

### DIFF
--- a/src/components/content-tab-assistant.tsx
+++ b/src/components/content-tab-assistant.tsx
@@ -386,7 +386,8 @@ export function ContentTabAssistant( { selectedSite }: ContentTabAssistantProps 
 
 	useEffect( () => {
 		void dispatch( chatThunks.updateFromSite( { site: selectedSite } ) );
-	}, [ dispatch, selectedSite ] );
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [ dispatch, selectedSite.id ] );
 
 	useEffect( () => {
 		if ( themeDetails ) {
--- a/src/stores/chat-slice.ts
+++ b/src/stores/chat-slice.ts
@@ -1,12 +1,10 @@
-import { createSlice, createAsyncThunk, PayloadAction } from '@reduxjs/toolkit';
+import { createSlice, createAsyncThunk, PayloadAction, isAnyOf } from '@reduxjs/toolkit';
 import * as Sentry from '@sentry/electron/renderer';
 import { WPCOM } from 'wpcom/types';
 import { z } from 'zod';
+import { DEFAULT_PHP_VERSION } from 'common/constants';
 import { LOCAL_STORAGE_CHAT_API_IDS_KEY, LOCAL_STORAGE_CHAT_MESSAGES_KEY } from 'src/constants';
 import { getIpcApi } from 'src/lib/get-ipc-api';
-// Provider constants are retrieved via IPC when needed
-// Default PHP version for initial state
-const DEFAULT_PHP_VERSION = '8.3';
 import { AppDispatch, RootState } from 'src/stores';
 import { assistantQuotaSchema, wpcomApi } from 'src/stores/wpcom-api';
 
@@ -90,6 +88,13 @@ const updateFromSite = createTypedAsyncThunk(
 			themes,
 			siteUrl,
 		};
+	},
+	{
+		condition: ( { site }, { getState } ) => {
+			const { chat } = getState();
+			const thunkIsCurrentlyLoading = chat.isLoadingUpdateFromSiteDict[ site.id ];
+			return ! thunkIsCurrentlyLoading;
+		},
 	}
 );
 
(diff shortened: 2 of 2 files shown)


src/components/content-tab-assistant.tsx
```diff
@@ -386,7 +386,8 @@ export function ContentTabAssistant( { selectedSite }: ContentTabAssistantProps
 
 	useEffect( () => {
 		void dispatch( chatThunks.updateFromSite( { site: selectedSite } ) );
-	}, [ dispatch, selectedSite ] );
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [ dispatch, selectedSite.id ] );
```
fredrikekelund, line 390: This is the key change. The `selectedSite` object is being mutated frequently somewhere, so we change the dependency to `selectedSite.id`

src/stores/chat-slice.ts
```diff
@@ -111,7 +116,7 @@ const assistantResponseSchema = z.object( {
 const assistantHeadersSchema = z.object( {
 	'x-quota-max': z.coerce.number(),
 	'x-quota-remaining': z.coerce.number(),
-	'x-quota-reset': z.string().datetime( { offset: true } ),
+	'x-quota-reset': z.iso.datetime( { offset: true } ),
```
fredrikekelund, line 119: https://zod.dev/v4/changelog#zstring-updates

src/stores/chat-slice.ts
```diff
@@ -90,6 +88,13 @@ const updateFromSite = createTypedAsyncThunk(
...
 			siteUrl,
 		};
+	},
+	{
+		condition: ( { site }, { getState } ) => {
+			const { chat } = getState();
+			const thunkIsCurrentlyLoading = chat.isLoadingUpdateFromSiteDict[ site.id ];
+			return ! thunkIsCurrentlyLoading;
+		},
 	}
```
fredrikekelund, line 98: Adds additional safety by preventing multiple parallel `updateFromSite` invocations

src/components/content-tab-assistant.tsx
epeicher: Good catch! I think we could create a follow-up to identify the root cause, but for the time being, this is a good fix that prevents the re-runs when properties other than `id` change.

epeicher (approved):
Thanks for the fix @fredrikekelund! I have tested it, and it works as expected. I don't see any errors in the console after using the AI assistant. 

|Before | After|
|-------|------|
| Image| Image|

src/stores/chat-slice.ts
```diff
@@ -1,12 +1,10 @@
-import { createSlice, createAsyncThunk, PayloadAction } from '@reduxjs/toolkit';
+import { createSlice, createAsyncThunk, PayloadAction, isAnyOf } from '@reduxjs/toolkit';
 import * as Sentry from '@sentry/electron/renderer';
 import { WPCOM } from 'wpcom/types';
 import { z } from 'zod';
+import { DEFAULT_PHP_VERSION } from 'common/constants';
```
ivan-ottinger, line 5: Nice catch!

gcsecsey (commented):
The approach and the changes make sense to me. I couldn't reproduce the issue on trunk on Mac, so I can't confirm yet if the changes eliminated the errors on my machine.

I tested sending messages to ask about the installed plugins/themes on the site, or to print some WP CLI commands for me that I ran, but didn't get any errors in the console:

| trunk | After |
|--------|--------|
| Image | Image |  

I see in Sentry that the issue is platform agnostic, so I think I should be able to reproduce it on Mac:
Image

@fredrikekelund do you have any tips or steps on how to produce the excessive WPCLI calls on trunk?

ivan-ottinger (approved):
Thanks for the fix, Fredrik. The change looks good and I did not observe any regressions. The Assistant still works as expected.

Regarding the root issue - have you been able to reproduce it? I haven't been able so far.

fredrikekelund:
> have you been able to reproduce it? 

I have. It looks like @epeicher could, too (see https://github.com/Automattic/studio/pull/2472#pullrequestreview-3705293042).

The issue comes from excessive `updateFromSite` thunk invocations. `chat-slice` starts triggering those as soon as the Assistant tab is opened, basically. 

@gcsecsey @ivan-ottinger, maybe you cannot trigger the errors on trunk because the server wasn't running…

src/components/content-tab-assistant.tsx
fredrikekelund: > I think we could create a follow-up to identify the root cause

It could be that the `themeDetails` object is overwritten more frequently now. That's not necessarily an error, IMO. Depending on the entire `selectedSite` object here might just have been a flawed strategy…

Still, it would be good to understand why this happened in the first place.

PR #2473, opened 2026-01-26
Fix bundled node permissions

bcotrim:
## Related issues

- Fixes STU-1253

## Proposed Changes

- Add custom entitlements for the bundled Node.js binary to fix SIGTRAP crashes on Intel Macs

### Root Cause

When electron-forge signs the app, it re-signs the bundled Node binary with Electron's default entitlements. These entitlements are missing critical permissions that Node.js/V8 requires for JIT compilation:

| Entitlement | Official Node | Bundled (before) | Bundled (after) |
|-------------|:-------------:|:----------------:|:---------------:|
| `cs.allow-jit` | ✅ | ✅ | ✅ |
| `cs.allow-unsigned-executable-memory` | ✅ | ❌ | ✅ |
| `cs.disable-executable-page-protection` | ✅ | ❌ | ✅ |
| `cs.disable-library-validation` | ✅ | ❌ | ✅ |

Without these entitlements, V8's baseline compiler crashes when trying to allocate executable memory:
```
Fatal error in , line 0
Check failed: 12 == (*__error()).  // ENOMEM
```

### The Fix

- Created `entitlements/node.plist` with the required V8/JIT entitlements
- Updated `forge.config.ts` to use `optionsForFile` to apply these entitlements specifically to `bin/node` during code signing

## Testing Instructions

1. Download the **Intel (x64)** build from CI
2. Install and run Studio
3. Create or start a site - it should work without crashing

**To verify the fix is applied:**
```bash
codesign -d --entitlements - /Applications/Studio.app/Contents/Resources/bin/node
```
Should show:
- `com.apple.security.cs.allow-jit`
- `com.apple.security.cs.allow-unsigned-executable-memory`
- `com.apple.security.cs.disable-executable-page-protection`
- `com.apple.security.cs.disable-library-validation`

## Pre-merge Checklist

### DIFF
--- a/entitlements/node.plist
+++ b/entitlements/node.plist
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <!-- Required for V8 JIT compilation -->
+    <key>com.apple.security.cs.allow-jit</key>
+    <true/>
+    <!-- Required for V8 to generate unsigned executable code -->
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <true/>
+    <!-- Required for V8 memory page management -->
+    <key>com.apple.security.cs.disable-executable-page-protection</key>
+    <true/>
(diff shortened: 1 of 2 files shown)


fredrikekelund (approved):
I confirm that this fixes the issue: the bundled node runtime now spawns as expected in the macOS Intel build.

PR #2474, opened 2026-01-26
Fix thumbnail not loading for newly created sites (approach 1)

ivan-ottinger:
## Summary
- Fixes an issue where newly created sites would show "Preview unavailable" instead of loading the thumbnail
- The thumbnail would only appear after navigating away and back to the Overview tab
- Root cause: the error state was set when the img tried to load with empty src, and could never be reset because the img was replaced by a div

## Changes
- Added `thumbnailLoadErrors` tracking at the context level in `use-theme-details.tsx`
- Added a `useEffect` to reset local error state when thumbnail loading starts
- Updated loading condition to show skeleton for running sites without thumbnail data (unless a load error occurred)

## Test plan
- [ ] Create a new site
- [ ] Verify the thumbnail loads without needing to navigate away
- [ ] Verify existing sites still show thumbnails correctly
- [ ] Verify "Preview unavailable" still shows when thumbnail generation actually fails

Fixes [STU-1212](https://linear.app/a8c/issue/STU-1212/newly-added-site-doesnt-have-thumbnail)

### DIFF
--- a/src/components/content-tab-overview.tsx
+++ b/src/components/content-tab-overview.tsx
@@ -202,10 +202,10 @@ export function ContentTabOverview( { selectedSite }: ContentTabOverviewProps ) 
 	const isServerLoading = loadingServer[ selectedSite.id ];
 
 	useEffect( () => {
-		if ( loadingThumbnails ) {
+		if ( thumbnailData ) {
 			setIsThumbnailError( false );
 		}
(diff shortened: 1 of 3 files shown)


ivan-ottinger:
Closing in favor of https://github.com/Automattic/studio/pull/2484.

PR #2475, opened 2026-01-26
Timeout `site stop --all` after 6s when quitting

fredrikekelund:
## Related issues

- Fixes STU-1256

## Proposed Changes

When Studio quits, the `stopAllServers` function runs the `site stop --all` CLI command to stop running sites. This function resolves if the command fails or if the Node.js `ChildProcess` instance emits an `error` event. 

Unfortunately, when testing the Studio 1.7.0 Intel build on my Apple Silicon laptop (using Rosetta emulation) none of these things appear to happen. I don't yet know what happens for regular macOS Intel users, but there are some indications that `stopAllServers` does resolve. In any case, it's likely not working for me because the bundled node runtime initializes but can't proceed, getting stuck in an intermittent state. 

To mitigate against similar issues in the future, this PR adds an optional timeout to the `stopAllServers` function that kills the child process and resolves the promise. 

## Testing Instructions

1. Run `npm start`
2. Wait a second or two
3. Quit the app
4. Ensure that it exits cleanly

## Pre-merge Checklist


gcsecsey (approved):
Changes LGTM, and the process still exists cleanly.

Image

src/index.ts
```diff
@@ -517,7 +517,7 @@ async function appBoot() {
 
 		if ( shouldStopSitesOnQuit ) {
 			event.preventDefault();
-			stopAllServers( true )
+			stopAllServers( true, 6_000 )
```
epeicher, line 499: Nit: I think adding a constant with a meaningful name would help to identify what this number is

epeicher (approved):
Thanks @fredrikekelund! I have tested it, and I can see that the application exits cleanly after different amounts of time. I left a minor comment, but this can progress anyway. LGTM! 

Image

fredrikekelund:
Thanks for the reviews, @gcsecsey and @epeicher 🙏 This fell by the wayside as I was sick and then AFK. Coming back and landing this now.

src/index.ts
fredrikekelund: I think the function argument name already provides similar clarity. Extracting the value into a constant would also introduce a jump for the reader between where the value is defined and this, the only place where it's used. So, in this particular case, I'll go ahead and land it as is 👍

PR #2480, opened 2026-01-26
Improve delete site focus tab and selected site id

sejas:
## Related issues

- Related to https://github.com/Automattic/studio/pull/2468

## Proposed Changes

- Fix delete site behaviour but keep the logic simple

## Testing Instructions

- See https://github.com/Automattic/studio/pull/2468

## Pre-merge Checklist

### DIFF
--- a/src/components/site-content-tabs.tsx
+++ b/src/components/site-content-tabs.tsx
@@ -16,7 +16,7 @@ import { cx } from 'src/lib/cx';
 import { ContentTabSync } from 'src/modules/sync';
 
 export function SiteContentTabs() {
-	const { selectedSite, siteCreationMessages } = useSiteDetails();
+	const { selectedSite, sites, siteCreationMessages } = useSiteDetails();
 	const { importState } = useImportExport();
 	const { tabs, selectedTab, setSelectedTab } = useContentTabs();
 	const { __ } = useI18n();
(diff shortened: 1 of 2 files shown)


gcsecsey (commented):
Thanks for suggesting this approach @sejas! I experimented with this approach earlier as well, and I found that it doesn't always work as expected when following these steps:

1. Delete a site
2. Don't switch site, but optionally switch tabs
3. Observe that the selected site is updated to the first site, but the selected tab isn't updated to "Overview"

https://github.com/user-attachments/assets/4528cf3d-f287-480e-afa9-6b78db4a1d61

I think this might be caused by the selectedSiteId being set from the `SITE_EVENTS.DELETED` [event listener](https://github.com/Automattic/studio/blob/86a88f2b6f7ec307e85dbfff6ce26c8b045fe318/src/hooks/use-site-details.tsx#L201), and the tab state update being prevented from running.

sejas:
@gcsecsey, thanks for reviewing the PR. I tested it again and I'm not able to reproduce your issue. 

https://github.com/user-attachments/assets/5d2d5f6f-b194-497a-8adf-9380d2a6ea48

gcsecsey:
Thanks for huddling about these PRs @sejas! I tested these changes on Windows x64 too because I was suspicious that something's off with my environment on Mac. I see the same issue when deleting a site, and staying on the same site:

https://github.com/user-attachments/assets/0ee5cd75-6a8a-443e-846e-1f5ddf090eab

As we discussed on the huddle, it might be worth looking into unifying the deletion logic across [`SITE_EVENTS.DELETED` listener](https://github.com/Automattic/studio/blob/86a88f2b6f7ec307e85dbfff6ce26c8b045fe318/src/hooks/use-site-details.tsx#L204) and the [onDeleteSite](https://github.com/Automattic/studio/blob/86a88f2b6f7ec307e85dbfff6ce26c8b045fe318/src/hooks/use-site-details.tsx#L254) handler. As both of these call `setSelectedSiteId`, there might be a race condition on the state updates.

sejas:
Thank you @gcsecsey , let's go with your approach!

PR #2481, opened 2026-01-27
Fix site export in offline mode

bcotrim:
## Related issues

- Fixes STU-1258

## Proposed Changes

- Fix `sqlite export` command to use absolute path `/wordpress/<filename>` instead of relative path. The WASM filesystem mounts the site at `/wordpress` but PHP's working directory isn't set there, causing relative paths to fail. This is the same fix applied to `sqlite import` in PR #2422.
- Fix `getSitePlugins` and `getSiteThemes` to tolerate stderr warnings when stdout contains valid JSON. When offline, WordPress produces warnings about being unable to connect to WordPress.org for update checks, but the plugin/theme list commands still return valid data.

## Testing Instructions

1. Create a site in Studio and start it
2. Disconnect from the internet
3. Export the site using "Export site" option
4. Verify the export completes successfully without errors
5. Reconnect to the internet
6. Export the site again to verify it still works when online
7. (Optional) Import the exported backup to verify it's valid

## Pre-merge Checklist

### DIFF
--- a/cli/commands/site/tests/create.test.ts
+++ b/cli/commands/site/tests/create.test.ts
@@ -101,7 +101,6 @@ describe( 'CLI: studio site create', () => {
 	let consoleLogSpy: jest.SpyInstance;
 	let fsMkdirSyncSpy: jest.SpyInstance;
 	let loggerReportSuccessSpy: jest.SpyInstance;
-	let loggerReportWarningSpy: jest.SpyInstance;
 
 	const createPathExistsMock = ( sitePathExists = false ) => {
 		const bundledWPPath = require( 'path' ).join(
@@ -127,7 +126,7 @@ describe( 'CLI: studio site create', () => {
 		consoleLogSpy = jest.spyOn( console, 'log' ).mockImplementation();
 		fsMkdirSyncSpy = jest.spyOn( require( 'fs' ), 'mkdirSync' ).mockReturnValue( undefined );
 		loggerReportSuccessSpy = jest.spyOn( Logger.prototype, 'reportSuccess' );
-		loggerReportWarningSpy = jest.spyOn( Logger.prototype, 'reportWarning' );
+		jest.spyOn( Logger.prototype, 'reportWarning' );
 		createPathExistsMock( false );
 		( isEmptyDir as jest.Mock ).mockResolvedValue( true );
 		( isWordPressDirectory as jest.Mock ).mockReturnValue( false );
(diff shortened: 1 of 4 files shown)


cli/commands/site/tests/create.test.ts
```diff
@@ -99,7 +101,6 @@ describe( 'CLI: studio site create', () => {
 	let consoleLogSpy: jest.SpyInstance;
 	let fsMkdirSyncSpy: jest.SpyInstance;
 	let loggerReportSuccessSpy: jest.SpyInstance;
-	let loggerReportWarningSpy: jest.SpyInstance;
 
 	const createPathExistsMock = ( sitePathExists = false ) => {
```
bcotrim, line 104: Unrelated change to fix `npm run lint` warning.

ivan-ottinger (approved):
Thank you for the fix, Bernardo! The changes look good and I can confirm they resolve the issue. Export now works well while being offline.

Image

I did not observe any regressions - online export still runs as before. Import as well. 👍🏼

src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -306,16 +306,17 @@ export class DefaultExporter extends EventEmitter implements Exporter {
...
 		);
 
-		if ( stderr ) {
-			console.error( `Could not get information about plugins: ${ stderr }` );
-			throw new Error(
-				'Could not get information about installed plugins to create meta.json file.'
-			);
-		}
-
+		// Try to parse the output first - WordPress may produce warnings on stderr
```
epeicher: Parsing the output first makes sense 👍 

One question about the code is that we are parsing `stdout`, but the comment mentions `WordPress may produce warnings on stderr`, should that be `stdout` in the comment 🤔 ?

epeicher (approved):
Thanks @bcotrim for fixing this! I have tested it, and it works as expected, I am able to export site in offline mode while I can reproduce the error in `trunk`. I have left a question out of curiosity, but changes LGTM!

Image

src/lib/import-export/export/exporters/default-exporter.ts
bcotrim: Good catch, I updated the comment in https://github.com/Automattic/studio/pull/2481/changes/f798c84af438db74daecfa0f0d0c5e67833f00a4
Let me know if it's clear now 🙏

gcsecsey (approved):
Thanks for the fix @bcotrim! These changes LGTM, I tested them on Windows to make sure the path updates work there too. I confirm being able to export and reimport sites without issues now. 👍 

https://github.com/user-attachments/assets/956efbf5-b435-4536-9a73-d1308b8755dc

src/lib/import-export/export/exporters/default-exporter.ts
epeicher: That's much clearer now, and it makes total sense, thanks for updating!

PR #2483, opened 2026-01-27
Record more error details on CLI failure

fredrikekelund:
## Related issues

- Related to STU-1260

## Proposed Changes

After the 1.7.0 launch, we've seen an increased number of `Failed to start server` and `Failed to create server` errors. We know this stems from the CLI launch, and the errors are captured in Sentry, but we don't record enough details to know exactly what's going on. 

This PR fixes that by:

1. Changing the `CliCommandEventEmitter`'s `failure` event to contain a new custom error class in the payload.
2. This custom error class (`CliCommandError`) captures the last error message sent over IPC (probably the most vital piece of information), stderr, stdout, the process's exit code, or the type of signal that interrupted the process. 
3. All site CLI commands now pass the `output: 'capture'` config to `executeCliCommand`. If something unexpected is printed to stderr or stdout in the child process, we will now capture it, too. 

## Testing Instructions

1. Add a line somewhere in `cli/commands/site/start.ts` that just throws an error, like `throw new Error( 'TEST ERROR' );`
2. Run `npm start`
3. Start a server
4. Ensure that you see the following in your terminal:

```
Sentry Logger [log]: Captured error event `[Last error message] Failed to start WordPress server: TEST ERROR
```

## Pre-merge Checklist

### DIFF
--- a/src/modules/cli/lib/cli-server-process.ts
+++ b/src/modules/cli/lib/cli-server-process.ts
@@ -43,13 +43,9 @@ export class CliServerProcess implements WordPressServerProcess {
 				resolve();
 			} );
 
-			emitter.on( 'failure', ( { result, lastErrorMessage } ) => {
-				const errorDetail =
-					lastErrorMessage ||
-					result.stderr.trim() ||
-					result.stdout.trim() ||
-					`Failed to start site ${ this.siteId }`;
-				reject( new Error( errorDetail ) );
+			emitter.on( 'failure', ( { error } ) => {
+				error.baseMessage = 'Failed to start site';
+				reject( error );
 			} );
 
 			emitter.on( 'error', ( { error } ) => {
@@ -69,13 +65,9 @@ export class CliServerProcess implements WordPressServerProcess {
 				resolve();
 			} );
 
-			emitter.on( 'failure', ( { result, lastErrorMessage } ) => {
-				const errorDetail =
-					lastErrorMessage ||
-					result.stderr.trim() ||
-					result.stdout.trim() ||
-					`Failed to stop site ${ this.siteId }`;
-				reject( new Error( errorDetail ) );
+			emitter.on( 'failure', ( { error } ) => {
+				error.baseMessage = 'Failed to stop site';
+				reject( error );
 			} );
 
 			emitter.on( 'error', ( { error } ) => {
@@ -99,13 +91,9 @@ export class CliServerProcess implements WordPressServerProcess {
 				resolve();
 			} );
 
-			emitter.on( 'failure', ( { result, lastErrorMessage } ) => {
-				const errorDetail =
-					lastErrorMessage ||
-					result.stderr.trim() ||
-					result.stdout.trim() ||
-					`Failed to delete site ${ this.siteId }`;
-				reject( new Error( errorDetail ) );
+			emitter.on( 'failure', ( { error } ) => {
+				error.baseMessage = 'Failed to delete site';
+				reject( error );
 			} );
 
 			emitter.on( 'error', ( { error } ) => {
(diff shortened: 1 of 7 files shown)


src/modules/cli/lib/execute-command.ts
```diff
@@ -95,20 +171,20 @@ export function executeCliCommand(
...
-	} );
-
-	let capturedExitCode: number | null = null;
+		const errorParsed = cliErrorMessageSchema.safeParse( message );
+		if ( errorParsed.success ) {
+			lastErrorMessage = errorParsed.data.message;
+		}
 
-	child.on( 'exit', ( code ) => {
-		capturedExitCode = code;
```
fredrikekelund, line 104: I removed the `exit` event handler, because it's redundant. The `close` event already provides the same params as the `exit` event.

src/modules/cli/lib/execute-command.ts
```diff
@@ -95,20 +171,20 @@ export function executeCliCommand(
...
+		eventEmitter.emit( 'data', { data: message } );
 	} );
 
 	function appQuitHandler() {
 		const pid = child.pid;
+		child.removeAllListeners();
```
fredrikekelund, line 184: This is unlikely to have any discernible effect, but it still makes sense in principle: this function runs before we quit the app. It kills the child process with a `SIGTERM` signal. If we initiated the process shutdown and the app is quitting anyway, then we don't need to listen for any more events on `child`

src/modules/cli/lib/cli-server-process.ts
```diff
@@ -71,20 +76,24 @@ export class CliServerProcess implements WordPressServerProcess {
 		} );
 	}
 
-	async delete( deleteFiles: boolean ): Promise< void > {
+	async delete( trashFiles: boolean ): Promise< void > {
```
sejas, line 79: Renaming the trashFiles parameter seems not related to this PR but it's ok to keep it.

sejas (approved):
I edited `stop.ts` to throw an error, tested it, and confirmed that I see the error message in the Node console.

Image

I also tested importing a SQL file. And the error appeared in the console and in the dialog.

Image
Image

src/modules/cli/lib/cli-server-process.ts
fredrikekelund: Yeah, I should've commented on that, but that was intentional on my part. Just clarifies what the param does 👍

gcsecsey (approved):
I added the test error, and it's shown both in the alert box and in the console. 👍 

Image

Image

src/modules/cli/lib/execute-command.ts
```diff
@@ -117,18 +193,30 @@ export function executeCliCommand(
 		}
 	}
 
-	child.on( 'close', ( code ) => {
+	child.on( 'close', ( exitCode, signal ) => {
```
fredrikekelund, line 196: As a general FYI, here's from the [Node.js docs](https://nodejs.org/api/child_process.html#event-close):

> If the process exited, `code` is the final exit code of the process, otherwise `null`. If the process terminated due to receipt of a `signal`, signal is the string name of the signal, otherwise `null`. One of the two will always be non-`null`.

src/modules/cli/lib/execute-command.ts
```diff
@@ -2,56 +2,133 @@ import { app } from 'electron';
...
 import EventEmitter from 'node:events';
 import * as Sentry from '@sentry/electron/main';
+import { z } from 'zod';
 import { getBundledNodeBinaryPath, getCliPath } from 'src/storage/paths';
 
-export interface CliCommandResult {
+export type CliCommandResult = {
 	stdout: string;
 	stderr: string;
-	exitCode: number;
```
fredrikekelund, line 10: The value of `exitCode` is now implicit based on whether we emit a `success` or `failure` event.

PR #2486, opened 2026-01-27
Update `.nvmrc` to use Node.js v24

fredrikekelund:
## Related issues

- Fixes STU-1234

## Proposed Changes

- Updates the bundled node version to v24 by updating `.nvmrc`

## Testing Instructions

* Confirm that everything passes in CI and it installs the node 24.11 version (you can check here in the installing node.js step: https://buildkite.com/automattic/studio/builds/11169#019c293b-6360-4a77-bf3b-d7d5bb06c7af )

Screenshot 2026-02-04 at 4 27 37 PM

* Alternatively, pull the changes from this branch
* Run `npm install`
* Run `nvm use`
* Confirm that you get node version 24.11 installed 

## Pre-merge Checklist


katinthehatsite:
Just a small note that I did not update the fallback version and left it as is for safety

gcsecsey (commented):
Using the new `npm` version works as expected:

Image

I can also start the app and everything works nicely.

However, when running `npm i`, I see this warning:
```
npm warn Unknown project config "10.9.4". This will stop working in the next major version of npm.
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'studio@1.7.3',
npm warn EBADENGINE   required: { npm: '>=10.0.0 <11.0.0', node: '>=22.0.0 <23.0.0' },
npm warn EBADENGINE   current: { node: 'v24.11.0', npm: '11.6.1' }
npm warn EBADENGINE }
```

I think we'll also need to update the [engines property](https://github.com/Automattic/studio/blob/2811e12ec3fad628fc5d74b35f81974dd3d8ca8e/package.json#L10-L12) in package.json.

katinthehatsite:
> I think we'll also need to update the [engines property](https://github.com/Automattic/studio/blob/2811e12ec3fad628fc5d74b35f81974dd3d8ca8e/package.json#L10-L12) in package.json.

👍 Made the update as suggested.

gcsecsey (approved):
Thanks for the update, I see no more warnings! 👍

ivan-ottinger:
Thank you for taking Fredrik's PR to the finish line, @katinthehatsite! The change looks good, however, I noticed that we may need to update npm as well (in engines):

```bash
npm warn Unknown project config "10.9.4". This will stop working in the next major version of npm.
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'studio@1.7.3',
npm warn EBADENGINE   required: { npm: '>=10.0.0 <11.0.0', node: '>=24.0.0 <25.0.0' },
npm warn EBADENGINE   current: { node: 'v24.11.0', npm: '11.6.1' }
npm warn EBADENGINE }
npm warn skipping integrity check for git dependency ssh://git@github.com/electron/node-gyp.git
```

katinthehatsite:
> Thank you for taking Fredrik's PR to the finish line, @katinthehatsite! The change looks good, however, I noticed that we may need to update npm as well (in engines):

I will take a look in a follow-up PR 👀

wojtekn:
It doesn't look like a correct change.

Studio currently uses Electron 39.x, which requires Node v22.20. After modifying the default `.nvmrc` file, when we run `nvm use` in the project root, it installs Node 24 for the whole Studio app, not just the CLI.

katinthehatsite:
> Studio currently uses Electron 39.x, which requires Node v22.20. After modifying the default .nvmrc file, when we run nvm use in the project root, it installs Node 24 for the whole Studio app, 

I am checking this now in a different PR

wojtekn:
> I am checking this now in a different PR

Thanks. However, even if we update Electron in another PR, it still won't fix the underlying issue. The idea of bundling Node with the CLI was to untangle the Node version required by the Electron app and the CLI process, but it seems we tangled it again.

katinthehatsite:
> Thanks. However, even if we update Electron in another PR, it still won't fix the underlying issue. The idea of bundling Node with the CLI was to untangle the Node version required by the Electron app and the CLI process, but it seems we tangled it again.

From what I see so far, I don't think we will be able to upgrade Electron just yet anyway 😅 There is a relatively big blocker with `fs-ext` which is not compatible with the new version of Electron. At the end, I went with the approach of CLI and Electron using two different versions. I have a PR incoming.

katinthehatsite:
Added PR here: https://github.com/Automattic/studio/pull/2537 cc @wojtekn @ivan-ottinger

fredrikekelund:
@wojtekn, I don't entirely understand your argument here. `.nvmrc` dictates which node version nvm installs in the repo, not which node version Electron bundles. 

It's possible that Electron relies on a system-level node installation during development (though I doubt it), but it's entirely fine to have Electron 39 bundle node 22 for the "Electron back-end logic" and separately bundle node 24 as the runtime for spawning Studio CLI child processes. 

Am I missing something?

wojtekn:
> It's possible that Electron relies on a system-level node installation during development (though I doubt it), but it's entirely fine to have Electron 39 bundle node 22 for the "Electron back-end logic" and separately bundle node 24 as the runtime for spawning Studio CLI child processes.

I don't have an exact error noted, but it didn't work for me when I used `nvm use`, `npm install` and `npm start`.

PR #2492, opened 2026-01-28
Studio: Fix site deletion lock contention for multiple deletions

epeicher:
## Related issues

- Fixes STU-1266

## Root Cause
When deleting a site, both the main process and CLI tried to lock and update the same `appdata-v1.json.lock` file:

Main process acquires lock → spawns CLI
CLI tries to acquire the same lock → blocked
CLI waits 5 seconds for lock to become "stale", then takes over
Single site: Hidden 5-second delay on every deletion

Multiple sites: Race condition when multiple processes try to take over the stale lock simultaneously → EEXIST errors → deletion failures

## Proposed Changes

- Fixed site deletion lock contention that caused failures when deleting multiple sites in succession
- Removed redundant appdata locking from the main process `deleteSite` handler
- The CLI already handles appdata updates with proper locking, so the main process lock was causing a deadlock
- Also removes the hidden 5-second delay that occurred during single site deletion while waiting for the lock to become stale

## Testing Instructions

- Create multiple test sites in Studio. I use the `studio` cli for that by moving to `~/Studio` folder and running:
```sh
for n in 1 2 3 4 5 6 7 8 9 0; do studio site create --path="site-number-$n" --skip-browser --skip-log-details; done
```
- Delete them one after another in quick succession
- Verify all sites are deleted successfully without errors
- No "appdata-v1.json" locking error should appear
- Single site deletion should complete without delays

## Pre-merge Checklist


bcotrim (approved):
Nice catch and cleanup! @epeicher 
Works as described
LGTM 👍

PR #2498, opened 2026-01-28
Add drag-and-drop site reordering with sortOrder persistence

### LINKED ISSUE #588: Feature Request: Reorder sites on the left sidebar
### What

Currently, when creating a new site in WP Studio, it adds it to the top of the list. This list cannot have its order changed, and so all sites are always in descending order of creation time. I request the ability to drag the names of sites to reorder them. At the very least, to set a "priority" number to sites, which effectively reorders them.

### Why

Finding the site names in a list of 20+ sites within WP Studio can be tedious. Having the ability to reorder them at will would make this seeking process less tedious, as I could expect where to find a site in the list if I have ordered it myself.

### How

_No response_

mikaoelitiana:
## Summary

This PR implements drag-and-drop site reordering functionality. Users can now reorder sites in the sidebar by dragging and dropping them, with the new sort order being saved to disk. Existing sites without a sort order will be sorted alphabetically after those with an explicit order.

The main changes involve making site items draggable and updating `sortOrder` property upon item drop. This new field is added to the data stored locally.

https://github.com/user-attachments/assets/c28b20db-47e8-4bdc-8c85-3ef028eb69f9

Resolves #588


profucius:
Hi there, thanks for taking this feature! I watched your video above and it looks great so far.

wojtekn:
Thanks for working on that @mikaoelitiana .

I think we could drop the drag indicator to keep the interface cleaner.

I opened https://github.com/Automattic/studio/pull/2531 to trigger CI.

mikaoelitiana:
> Thanks for working on that @mikaoelitiana .
> 
> I think we could drop the drag indicator to keep the interface cleaner.
> 
> I opened #2531 to trigger CI.

OK I will remove it

mikaoelitiana:
@wojtekn I have remove the drag icon and updated the video on the PR description

tools/common/lib/sort-sites.ts
```diff
@@ -1,3 +1,18 @@
...
+		if ( a.sortOrder !== undefined ) {
+			return -1;
+		}
+		// If only b has sortOrder, b comes first
+		if ( b.sortOrder !== undefined ) {
+			return 1;
+		}
+		// Neither has sortOrder, sort by name
+		return a.name.localeCompare( b.name, undefined, { numeric: true } );
+	} );
```
bcotrim, line 8: This would make the function easier to read and fallback to `localCompare` if the `sortOrder` is the same

```suggestion
export function sortSites< T extends { name: string; sortOrder?: number } >( sites: T[] ): T[] {
	const MAX = Number.MAX_SAFE_INTEGER;
	return sites.sort( ( a, b ) => {
		return (
			( a.sortOrder ?? MAX ) - ( b.sortOrder ?? MAX ) ||
			a.name.localeCompare( b.name, undefined, { numeric: true } )
		);
	} );
}
```

src/hooks/use-site-details.tsx
```diff
@@ -390,6 +394,12 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 		setSites( updatedSites );
 	}, [] );
 
+	const reorderSites = useCallback( async ( updates: { siteId: string; sortOrder: number }[] ) => {
```
bcotrim: The reorder should update the UI optimistically — call `setSites` with the new order immediately, then persist to the ipc in the background (ideally debounced to avoid excessive writes during rapid drag-and-drop). Right now the list waits for the full IPC roundtrip before reflecting the new order, which causes a visible delay.

bcotrim (changes requested):
Thanks for the contribution @mikaoelitiana and welcome to the project!

I added some comments, please let me know what you think.

tools/common/lib/sort-sites.ts
wojtekn: It's a little bit cryptic that way. Could we make this more explicit?

```
// Sites with sortOrder come before those without
const orderA = a.sortOrder ?? Infinity;
const orderB = b.sortOrder ?? Infinity;

if ( orderA !== orderB ) {
    return orderA - orderB;
}

// Fall back
```

src/hooks/use-site-details.tsx
mikaoelitiana: @bcotrim I made some refactoring to address your comments. For the debounce part, I checked if we already do that somewhere in the project but I don't seem to find anything. Do you think we should install a dependency for that or create it in-house?

bcotrim: No need for an external dependency, a simple setTimeout/clearTimeout with a useRef does the trick. The key is to keep setSites immediate (optimistic) and only debounce the IPC persist:

```typescript
const saveTimeoutRef = useRef< ReturnType< typeof setTimeout > >();
const DEBOUNCE_SAVE_MS = 300;

const updateSitesSortOrder = useCallback( ( sites: SiteDetails[] ) => {
    setSites( sites );

    const updates = sites.map( ( site, index ) => ( {
        siteId: site.id,
        sortOrder: ( index + 1 ) * 1000,
    } ) );

    clearTimeout( saveTimeoutRef.current );
    saveTimeoutRef.current = setTimeout( () => {
        getIpcApi().updateSitesSortOrder( updates );
    }, DEBOUNCE_SAVE_MS );
}, [] );
```

wojtekn:
@mikaoelitiana, we recently changed the repo structure to a monorepo, and I merged the trunk to your branch and solved conflicts.

Please let me know if we should proceed with applying @bcotrim's comment.

mikaoelitiana:
> @mikaoelitiana, we recently changed the repo structure to a monorepo, and I merged the trunk to your branch and solved conflicts.
> 
> Please let me know if we should proceed with applying @bcotrim's comment.

Hello, thanks for updating my branch. I will work on the last comment from @bcotrim today

src/hooks/use-site-details.tsx
mikaoelitiana: Looks great, thanks! I will update that

mikaoelitiana:
@bcotrim @wojtekn I have addressed the comments and updated my PR

bcotrim (approved):
Thanks for making the changes @mikaoelitiana 
LGTM 👍

wojtekn (approved):

wojtekn:
Thanks for your contribution @mikaoelitiana !

PR #1671, opened 2025-08-13
Add a patch that removes setting DB_NAME constant to wp-config.php file

wojtekn:
## Related issues

- Fixes STU-691

## Proposed Changes

I propose to add a patch that removes setting DB_NAME constant to `wp-config.php` file.

It's a temporary change until we find a solution on the Playground CLI side to support it better.

## Testing Instructions

1. Run `npm install` and start Studio
2. Enable Blueprints feature flag
3. Create Studio site
4. Stop the site
5. Edit site's `wp-config.php` and remove DB_NAME constan (or all DB_* constants)
6. Start site
7. Confirm that site opens fine in browser
8. Confirm site exports fine
9. Confirm `wp-config.php` doesn't have additional DB_NAME constant definition prepended

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2499, opened 2026-01-28
Update WordPress Playground dependencies to 3.0.46

epeicher:
## Related issues

- Related to STU-1114

## Proposed Changes

- Update `@wp-playground/blueprints` to 3.0.46 in package.json
- Update `@php-wasm/universal`, `@wp-playground/blueprints`, `@wp-playground/cli`, `@wp-playground/common`, `@wp-playground/storage` to 3.0.46 in cli/package.json
- Remove PHP 7.2, 7.3, and 8.5 support (7.2/7.3 deprecated upstream; 8.5 causes WP-CLI deprecation warnings that break JSON parsing)
- Update blueprint validation to use new `validateBlueprint` export
- Fix `workerThreadCount` access via `internalsKeyForTesting` symbol
- Add external config for `@php-wasm` packages in electron.vite.config.ts
- Remove old patches for 3.0.22 (no longer needed with 3.0.46)

### Why the external config for @php-wasm?

The `@wp-playground/blueprints` package (used for blueprint validation in the main process) imports `@php-wasm/web-*` packages which contain WASM files. Vite cannot bundle these WASM files and fails with "ESM integration proposal for Wasm is not supported". Marking `@php-wasm/*` as external tells Vite to skip bundling them, resolving the build failure.

## Testing Instructions

1. Run `npm install` in root and `cli/` directories
2. Run `npm start` and verify the app launches without WASM errors
3. Run `npm run make` and verify the production build succeeds
4. Confirm that creating, starting, and stopping the site works fine
5. Confirm that you can create a site using the normal flow, using a Blueprint and using an Import
7. Check the PHP version selector shows 7.4-8.4 range

## Pre-merge Checklist


electron.vite.config.ts
```diff
@@ -37,6 +37,9 @@ export default defineConfig( {
 				output: {
 					entryFileNames: '[name].js',
 				},
+				external: [
+					/^@php-wasm\/.*/,
+				],
```
epeicher, line 43: This has been added to avoid an issue with the new imports of [`import { validateBlueprint } from '@wp-playground/blueprints';`](https://github.com/Automattic/studio/pull/2499/changes#diff-b0b46050adfe4716fac15d4e380215725ef340ffb6f6637d2c17365f7abb5987R2). That import transitively imports `@php-wasm/*` packages that contain `.wasm` files that fail bundling.

Those `@php-wasm` are not needed by the UI so it is OK not to bundle them.

patches/@wp-playground+wordpress+3.0.22.patch
epeicher, line 1: I tested the following PR https://github.com/Automattic/studio/pull/1671#issue-3318802432 and it works as expected so I don't think this patch is required anymore.

cli/lib/run-wp-cli-command.ts
epeicher, line 1: This commit has a great explanation of why this change is required https://github.com/Automattic/studio/pull/2499/changes/62b1dc28a874dc7e0d686906a52004e61d754e42

electron.vite.config.ts
epeicher: Also this commit bf31ab5e contains the explanation for adding `/^@wp-playground\/.*/,`

katinthehatsite (approved):
In my testing, I did not spot any regressions so approving.

I am curious - do you know why we are setting PHP 8.3 instead of PHP 8.4 as default when creating sites?

epeicher:
Thanks @katinthehatsite for your review!

> I am curious - do you know why we are setting PHP 8.3 instead of PHP 8.4 as default when creating sites?

Good question, I don't really know, but it probably comes from playgrounds [recommendation here](https://github.com/WordPress/wordpress-playground/blob/9e9c6c95dd6a0de924b16d39ef14bd70a2a61fb0/packages/playground/common/src/index.ts#L17). I think we copy that in [this file](https://github.com/Automattic/studio/pull/2499/changes#diff-bae5563c0623500754c7d6c7eaa4bd69bb79514691d7502b53379ed491ad1497) to avoid adding a dependency to `wp-playground` there

wojtekn:
> I am curious - do you know why we are setting PHP 8.3 instead of PHP 8.4 as default when creating sites?

We do this to match WordPress.com's default.

common/types/php-versions.ts
```diff
@@ -5,16 +5,7 @@
...
-	'8.4',
-	'8.3',
-	'8.2',
-	'8.1',
-	'8.0',
-	'7.4',
-	'7.3',
-	'7.2',
-] as const;
+export const SupportedPHPVersions = [ '8.4', '8.3', '8.2', '8.1', '8.0', '7.4' ] as const;
```
wojtekn, line 8: Should we add 8.5?

patches/@wp-playground+wordpress+3.0.22.patch
```diff
@@ -1,25 +0,0 @@
...
- 			if (!file_exists($wp_load)) {
-diff --git a/node_modules/@wp-playground/wordpress/index.js b/node_modules/@wp-playground/wordpress/index.js
-index 219ccf3..311981e 100644
---- a/node_modules/@wp-playground/wordpress/index.js
-+++ b/node_modules/@wp-playground/wordpress/index.js
-@@ -366,7 +366,6 @@ async function x(e, n, t, i = "rewrite") {
- }
- async function L(e, n) {
-   const t = s(n, "wp-config.php"), i = {
--    DB_NAME: "wordpress"
```
wojtekn, line 22: What makes this patch unnecessary? Was this removed from Playground code?

cli/lib/run-wp-cli-command.ts
```diff
@@ -87,7 +90,10 @@ export async function runGlobalWpCliCommand(
 		await setPhpIniEntries( php, {
 			'openssl.cafile': '/tmp/ca-bundle.crt',
 			allow_url_fopen: 1,
-			disable_functions: '',
+			// Disable process spawning functions as they're not supported in WASM
+			// without a spawn handler. WP-CLI's search-replace tries to use these.
+			disable_functions:
+				'exec,passthru,shell_exec,system,proc_open,popen,proc_close,proc_get_status,proc_nice,proc_terminate,pcntl_exec',
```
wojtekn: Should we extract the list to avoid repeating?

bcotrim (commented):
Good job on these changes!

What happens to user sites currently on deprecated versions?
I tried to start a site using Studio CLI and it seems Playground automatically uses the `7.4` version, but it is invisible to the users. Should we display a warning?
Opening Studio app, seems to change the version to `8.0` without any warning to the users. I think we should follow a similar approach and show a warning to users letting them know that PHP version is deprecated and their site will boot with the next one.

What do you think?

common/types/php-versions.ts
epeicher: PHP 8.5 returns a deprecation error when running the `wp` cli that is not correctly handled. That could be addressed as a follow-up. The deprecation errors returned are [these ones](https://www.php.net/releases/8.5/en.php#:~:text=Terminating%20case%20statements,empty%20string%20instead.)

patches/@wp-playground+wordpress+3.0.22.patch
epeicher: I commented my reasoning [here](https://github.com/Automattic/studio/pull/2499#discussion_r2741716869), but I would like to double-check this removal. The playground package [still add that constant](https://github.com/WordPress/wordpress-playground/blob/0f591c4ee9ae2d8baac4c33e9715f056a3555356/packages/playground/wordpress/src/rewrite-wp-config.ts#L56), but I don't reproduce the issue in the [PR that added the patch ](https://github.com/Automattic/studio/pull/1671), I am able to Export using these changes. 

Shall I add the patch back with the correct version?

cli/lib/run-wp-cli-command.ts
epeicher: Great idea, updated as part of da8e5002

epeicher:
Thanks for your review @bcotrim!

> What happens to user sites currently on deprecated versions? I tried to start a site using Studio CLI and it seems Playground automatically uses the `7.4` version, but it is invisible to the users. Should we display a warning? Opening Studio app, seems to change the version to `8.0` without any warning to the users. I think we should follow a similar approach and show a warning to users letting them know that PHP version is deprecated and their site will boot with the next one.
> 
> What do you think?

In Studio, unknown versions are moved automatically to `8.0` - [source code here](https://github.com/Automattic/studio/blob/ff3a60159e67723dc3f0abb8f14c458967e5ce7d/src/storage/user-data.ts#L46)

I agree this can be confusing, so your suggestion of displaying a warning is a good option 👍

wojtekn:
@bcotrim @epeicher I think this isn't crucial enough to warrant a warning. Let's just use the lowest available in such a case. In the other case, we will decrease UX by a growing number of dialogs and warnings.

common/types/php-versions.ts
wojtekn: Thanks, so let's work on adding PHP 8.5 under a separate issue.

patches/@wp-playground+wordpress+3.0.22.patch
wojtekn: In my test, it re-added the DB_NAME to the `wp-config.php` file.

epeicher: Ok, I will add the patch updated to the current version

common/types/php-versions.ts
epeicher: Follow-up issue created [STU-1282: Add PHP 8.5 to Studio php-versions file](https://linear.app/a8c/issue/STU-1282/add-php-85-to-studio-php-versions-file)

epeicher:
> @bcotrim @epeicher I think this isn't crucial enough to warrant a warning. Let's just use the lowest available in such a case. In the other case, we will decrease UX by a growing number of dialogs and warnings.

Thanks @wojtekn, I will update the site [logic here](https://github.com/Automattic/studio/blob/ff3a60159e67723dc3f0abb8f14c458967e5ce7d/src/storage/user-data.ts#L46) to be more precise.

patches/@wp-playground+wordpress+3.0.22.patch
epeicher: Added the patch back as part of [1eabbce](https://github.com/Automattic/studio/pull/2499/commits/1eabbcee9a6187ffaf232f26f82b9269a1ab707e)

patches/@wp-playground+wordpress+3.0.22.patch
epeicher: Added patch back for the new version as part of 1eabbce

epeicher:
@bcotrim, @wojtekn I have updated the `populatePhjpVersion` logic as part of 25e0b79b. That change and others are ready for another review 🙏

bcotrim (approved):
LGTM 👍

wojtekn:
@epeicher I still reproduce the issue. I see it applies the patch, but when I start a site that doesn't have DB_NAME defined, it prepends the file with that.

Looking into how CLI is packaged now, should we copy this patch to `cli/`, too

cli/lib/run-wp-cli-command.ts
```diff
@@ -14,6 +14,11 @@ import { getSqliteCommandPath, getWpCliPharPath } from 'cli/lib/server-files';
 
 const PLAYGROUND_INTERNAL_SHARED_FOLDER = '/internal/shared';
 
+// Disable process spawning functions as they're not supported in WASM
+// without a spawn handler. WP-CLI's search-replace tries to use these.
+const DISABLED_FUNCTIONS =
```
adamziel: Could we have a spawn handler? Wouldn't PHP now fail with a fatal error if one of these functions is called? If you don't want to implement a full spawn handler, it could be just one that always returns exit code 1.

common/types/php-versions.ts
adamziel: I've ran into that as well! And dealt with it by silencing deprecation errors, which is not great but not terrible. Is WP-CLI ready for PHP 8.5 yet?

electron.vite.config.ts
adamziel: good call! In the Playground repo, we use `external` all the time, especially with all the node std lib packages.

common/types/php-versions.ts
wojtekn: @adamziel, let's tackle this in a separate PR to avoid scope creep here.

epeicher: > Is WP-CLI ready for PHP 8.5 yet?

@adamziel I don't think it is yet, the latest version still returns deprecations 
CleanShot 2026-02-06 at 13 56 19@2x

cli/lib/run-wp-cli-command.ts
epeicher: This failed originally because Studio was doing `wp search-replace` during the pull process. After adding this fix to disable functions, it worked fine. But I have implemented your suggestion in commit 2ef4a20b, and it still works fine. Please have a look at it and let me know if there are any issues.

adamziel: Looks great! :heart:

wojtekn (approved):
The code changes looks correct. It works as expected now.

PR #2516, opened 2026-02-02
Add support for Blueprints defineSiteUrl

bcotrim:
## Related issues

- Fixes https://linear.app/a8c/issue/STU-830

## Proposed Changes

- Extract `defineSiteUrl` from blueprint steps to pre-populate the custom domain field
- Create bidirectional mapping utilities between blueprint settings and form values (`extractFormValuesFromBlueprint` / `updateBlueprintWithFormValues`)
- Add `BlueprintSiteSettings` type for shared settings between blueprints and forms
- Pre-populate custom domain from `defineSiteUrl` step and expand advanced settings
- Pre-populate HTTPS toggle based on the protocol in `defineSiteUrl` (https vs http)
- Update blueprint with final user selections before site creation (only updates existing properties)

### Test 1: HTTPS URL in blueprint
1. Create a blueprint file with a `defineSiteUrl` step using **https**:
```json
{
  "steps": [
    {
      "step": "defineSiteUrl",
      "siteUrl": "https://mysite.local"
    }
  ]
}
```

2. Open Studio and go to Add Site → Start from Blueprint → Choose Blueprint file
3. Upload the blueprint file
4. Verify:
   - Custom domain field shows `mysite.local`
   - Advanced settings are expanded
   - **HTTPS toggle is ON**
5. Create the site and verify it works with HTTPS

### Test 2: HTTP URL in blueprint
1. Create a blueprint file with a `defineSiteUrl` step using **http**:
```json
{
  "steps": [
    {
      "step": "defineSiteUrl",
      "siteUrl": "http://mysite.local"
    }
  ]
}
```

2. Upload the blueprint file
3. Verify:
   - Custom domain field shows `mysite.local`
   - **HTTPS toggle is OFF**

### Test 3: URL with port (port is stripped)
1. Create a blueprint file with a port in the URL:
```json
{
  "steps": [
    {
      "step": "defineSiteUrl",
      "siteUrl": "https://mysite.local:8443"
    }
  ]
}
```

2. Upload the blueprint file
3. Verify custom domain field shows `mysite.local` (port is stripped - domain validation doesn't support ports)

## Pre-merge Checklist

### DIFF
--- a/common/lib/blueprint-settings.ts
+++ b/common/lib/blueprint-settings.ts
@@ -1,5 +1,10 @@
 import type { Blueprint } from '@wp-playground/blueprints';
-import type { BlueprintSiteSettings } from 'common/types/site-settings';
+
+type BlueprintSiteSettings = Partial<
+	Pick< StoppedSiteDetails, 'phpVersion' | 'customDomain' | 'enableHttps' >
+> & {
+	wpVersion?: string;
+};
 
 /**
  * Extracts form-relevant values from a blueprint.
--- a/common/lib/blueprint-settings.ts
+++ b/common/lib/blueprint-settings.ts
@@ -1,0 +1,75 @@
+import type { Blueprint } from '@wp-playground/blueprints';
+import type { BlueprintSiteSettings } from 'common/types/site-settings';
+
+/**
+ * Extracts form-relevant values from a blueprint.
+ */
+export function extractFormValuesFromBlueprint( blueprintJson: Blueprint ): BlueprintSiteSettings {
+	const values: BlueprintSiteSettings = {};
+
+	if ( blueprintJson.preferredVersions ) {
+		if ( blueprintJson.preferredVersions.php && blueprintJson.preferredVersions.php !== 'latest' ) {
+			values.phpVersion = blueprintJson.preferredVersions.php;
+		}
+		if ( blueprintJson.preferredVersions.wp && blueprintJson.preferredVersions.wp !== 'latest' ) {
+			values.wpVersion = blueprintJson.preferredVersions.wp;
+		}
+	}
+
+	if ( blueprintJson.steps && Array.isArray( blueprintJson.steps ) ) {
+		const defineSiteUrlStep = blueprintJson.steps.find(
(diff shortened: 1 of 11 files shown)


common/lib/blueprint-settings.ts
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	if ( blueprintJson.steps && Array.isArray( blueprintJson.steps ) ) {
+		const defineSiteUrlStep = blueprintJson.steps.find(
+			( step: { step?: string } ) => step.step === 'defineSiteUrl'
+		);
+		if ( defineSiteUrlStep?.siteUrl ) {
+			try {
+				const url = new URL( defineSiteUrlStep.siteUrl );
+				values.customDomain = url.hostname;
```
sejas, line 31: > Verify custom domain field shows mysite.local:8443 (port preserved)

If we want to preserve the port we might need to use `host` instead of `hostname`. Although I'm not sure if we should just show a warning or error to avoid falling into a rabbit hole.

```suggestion
				values.customDomain = url.host;
```

sejas (commented):
I tested the PR with and without https and I confirm Studio pre-fills the custom domain extracting it from `defineSiteUrl` step.

The only case that fails is `Test 3: URL with port - Verify custom domain field shows mysite.local:8443 (port preserved)`. I don't see the port preserved, and I'm not sure if we should support it at all.

bcotrim:
Good catch! This is intentional - we're using `url.hostname` instead of `url.host` to strip ports.

The reason is that domain validation (`getDomainNameValidationError` in `common/lib/domains.ts`) doesn't support ports:
1. The regex pattern doesn't allow `:` characters
2. Validation requires the domain to end with `.local`

If we preserved the port (e.g., `mysite.local:8443`), users would see a validation error in the form. Stripping the port avoids this issue.

I've updated the PR description to reflect this behavior. 👍

sejas (approved):
It works as expected. The advance settings are expanded when suing the site url step definition, and the custom domain is pre-populated as expected. I also modified the pre-populated domain and I confirm it respected my custom domain.

Image

PR #2518, opened 2026-02-03
Run E2E on Windows x64 and both Mac architectures

mokagio:
## Related issues

https://linear.app/a8c/issue/AINFRA-1888/e2e-test-coverage-for-all-studio-platforms

## Proposed Changes

- Adds Buildkite build matrix for the E2E tests to run additionally run on Windows and Intel Macs (simulated via Rosetta)
- Unfortunately, there are virtualizations limitations for Windows on ARM, so that platform is not part of this PR

## Testing Instructions

- Refer to green E2E steps in CI

image

## Pre-merge Checklist

### DIFF
--- a/.buildkite/commands/run-e2e-tests.sh
+++ b/.buildkite/commands/run-e2e-tests.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-MATRIX=${1:-}
+PLATFORM=${1:-}
+ARCH=${2:-}
 
 if .buildkite/commands/should-skip-job.sh --job-type validation; then
   exit 0
@@ -12,8 +13,19 @@ bash .buildkite/commands/install-node-dependencies.sh
 
 export IS_DEV_BUILD=true
 
-echo '--- :package: Package app for testing'
-npm run package
+echo "--- :package: Package app for testing ($PLATFORM-$ARCH)"
+case "$PLATFORM" in
+  mac)
+    npm run make:macos-"$ARCH"
+    ;;
+  windows)
+    npm run make:windows-"$ARCH"
+    ;;
+  *)
+    echo "Unknown platform: $PLATFORM"
+    exit 1
+    ;;
+esac
 
 echo '--- :playwright: Run End To End Tests'
 echo 'Installing Playwright browsers...'
--- a/.buildkite/commands/run-e2e-tests.sh
+++ b/.buildkite/commands/run-e2e-tests.sh
@@ -32,7 +32,12 @@ esac
 # `make` creates signed distributables (installers), which requires code signing setup.
 # `package` creates an unsigned app bundle, sufficient for E2E testing.
 echo "--- :package: Package app for testing ($PLATFORM-$ARCH)"
-npx electron-vite build --outDir=dist && npx electron-forge package --arch="$ARCH" --platform="$FORGE_PLATFORM"
+npx electron-vite build --outDir=dist
+npx electron-forge package --arch="$ARCH" --platform="$FORGE_PLATFORM"
+
+# This empty echo prevents the next one from being swallowed.
+# See https://buildkite.com/automattic/studio/builds/11098/steps/canvas?jid=019c2287-365e-49c6-b0ba-f7c1d06bc5f5#019c2287-365e-49c6-b0ba-f7c1d06bc5f5/L392
+echo ''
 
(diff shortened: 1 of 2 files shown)


.buildkite/commands/run-e2e-tests.sh
```diff
@@ -1,22 +1,56 @@
...
+# `package` creates an unsigned app bundle, sufficient for E2E testing.
+echo "--- :package: Package app for testing ($PLATFORM-$ARCH)"
+npx electron-vite build --outDir=dist
+npx electron-forge package --arch="$ARCH" --platform="$FORGE_PLATFORM"
+
+# This empty echo prevents the next one from being swallowed.
+# See https://buildkite.com/automattic/studio/builds/11098/steps/canvas?jid=019c2287-365e-49c6-b0ba-f7c1d06bc5f5#019c2287-365e-49c6-b0ba-f7c1d06bc5f5/L392
+echo ''
 
 echo '--- :playwright: Run End To End Tests'
```
mokagio, line 42: Before 

Image

After

Image

mokagio:
@AliSoftware @iangmaia I'd appreciate feedback on this before I remove the TODO and switch back to the original E2E running configuration. Haven't done it yet in case there's something to change, so I could run the E2E easily again. Thanks.

AliSoftware:
I was surprised that testing those apps on Mac Intel via Rosetta was possible, because I vaguely remembered that there was a limitation with some framework the app used that was not compatible with Rosetta. I asked LibreChat to find the P2 to refresh my memory, and it turns out that this problem was for Beeper Mac (internal ref: pdnsEh-1Yv-p2), not for WP Studio.

But still, the issue we identified back then (ref: pdnsEh-1Yv-p2#comment-4803) was that [Rosetta 2 was not supported by `Virtualization.framework`](https://developer.apple.com/forums/thread/759898) so it wasn't possible to run Rosetta 2 inside a VM on our CI. Which I'd expect to be a blocker for Studio too?

.buildkite/commands/run-e2e-tests.sh
```diff
@@ -1,22 +1,56 @@
...
-echo 'Installing Playwright browsers...'
-npx playwright install
-echo 'Running Playwright tests...'
-npx playwright test
+
+# For mac-x64 on Apple Silicon: install Rosetta and run under x86_64 arch
+arch_prefix=()
+if [ "$PLATFORM" = "mac" ] && [ "$ARCH" = "x64" ]; then
+  echo 'Installing Rosetta for x64 emulation...'
+  softwareupdate --install-rosetta --agree-to-license || true
```
AliSoftware, line 48: Have things changes since the last time we tried this and [found out](https://developer.apple.com/forums/thread/759898) that Installing Rosetta 2 wasn't supported inside VMs and by Virtualization.framework? 🤔 

[EDIT] [Seems like it passed though](https://buildkite.com/automattic/studio/builds/11103/steps/canvas?jid=019c2323-5f6a-40a5-a39c-f86ff1c73cc7#019c2323-5f6a-40a5-a39c-f86ff1c73cc7/L543)? Super interesting if that works then, that suggests the bug and feedback has been solved by Apple since?

Btw, do we know if there's a way we can be sure that the app version that is being tested (via playwright, iiuc) is really the Intel binary running through Rosetta, as opposed to it being a red herring and actually opening and testing the Apple Silicon version of Studio instead? (even if maybe `playwright` itself is run via Rosetta due to the `arch -x86_64`, does that apply to the binary it launches? 🤔 )

mokagio: I haven't checked if anything changed in Rosetta and/or Virtualization.framework when working on this.

---

Interesting question about ensuring the Playwright is actually running the E2E tests on the Intel version. Three ideas:

- Dive into the docs to verify the behavior based on the options (i.e. `arch -x86_64)
- Deleting any app binary that is not the x64 before running the tests
- Adding a bit of UI (if not there already) that tells the type of build we are running. On top of the UI work, this might also require updating the E2E script to either introspect or receive the OS and architecture it expects to run on, so that it can assert accordingly. (cc @bcotrim)

mokagio:
Thanks for the review @AliSoftware . You raise great questions as always. I'll extract the Windows Intel addition from this PR so we can lock that in, the continue the investigation for Mac on Intel. _Stay tuned_.

mokagio:
Closing in favor of https://github.com/Automattic/studio/pull/2530

PR #2526, opened 2026-02-03
Fix translation and adjust copy

nightnei:
## Proposed Changes
Adjust copy and fix translation (see comments in code)

## Testing Instructions
Visual review should suffice

### DIFF
--- a/cli/commands/site/create.ts
+++ b/cli/commands/site/create.ts
@@ -454,7 +454,7 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 				} )
 				.option( 'skip-log-details', {
 					type: 'boolean',
-					describe: __( 'Skip logging default wp-admin user details after starting' ),
+					describe: __( 'Skip logging site URL and admin credentials after creating' ),
 					default: false,
 				} );
 		},
--- a/cli/commands/site/create.ts
+++ b/cli/commands/site/create.ts
@@ -454,7 +454,7 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 				} )
 				.option( 'skip-log-details', {
 					type: 'boolean',
-					describe: __( 'Skip logging site URL and admin credentials after creating' ),
+					describe: __( 'Skip printing site URL and admin credentials after creating' ),
 					default: false,
 				} );
 		},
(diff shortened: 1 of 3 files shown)


cli/commands/site/create.ts
```diff
@@ -454,7 +454,7 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 				} )
 				.option( 'skip-log-details', {
 					type: 'boolean',
-					describe: __( 'Skip logging default wp-admin user details after starting' ),
+					describe: __( 'Skip logging site URL and admin credentials after creating' ),
```
nightnei: In this command it's more accurate to say "creating", since we are creating site. I suppose that it was jsut missed while copy-pasted from "start" command.
Also, just IMO, "default wp-admin user details" is a bit misleading, since we are printing also "Site URL" and saying "credentials" is a bit more clear to understand at a glance.

common/lib/blueprint-validation.ts
```diff
@@ -154,7 +154,7 @@ export async function validateBlueprintData(
 		const errorPath = firstError.instancePath || '/';
 		const additionalProp = firstError.params?.additionalProperty;
 		const errorMessage = additionalProp
-			? __( `"${ additionalProp }" is not a valid Blueprint property` )
+			? sprintf( __( '"%s" is not a valid Blueprint property' ), additionalProp )
```
nightnei, line 157: While adding translation to GlotPress I noticed it as `"" is not a valid Blueprint property`.

katinthehatsite (approved):
This looks good to me, thanks for fixing the issue with Blueprints translations 👍

common/lib/blueprint-validation.ts
sejas: Good catch!

sejas (approved):
The new descriptions make sense! 👌

cli/commands/site/create.ts
wojtekn: Good improvement. Should we also change `logging` to `printing`?

nightnei: Yeah, makes sense 👍 Done

Issue #87, opened 2024-05-03
Feature Request: Default Admin name email


ghost:
### What

I would love the option to change the default username, email and password for spining up new sites directly within the Studio application (a "preferences" setting maybe?)... directly on the local computer.

### Why

I don't know anyone who ever uses "admin" for an admin account. We all have our own information for this as best practices. This would save a step of having to go in and add my normal info in a new admin account, and then deleting the default one. It would just save everyone a few steps.

### How

Sorry I can't help with "How" - Only thing I can think of is that it could be something within the default settings or preferences (whatever term you are using for that option in the menu).

p.s. you might take a look at how Local does the site setup in order to do something similar. It allows you to set up a default admin setting OR change it right there when you enter the site name.

paaljoachim:
I agree it would be very helpful to be able to add a username, email and password when setting up a site.
Here is how LocalWP does this.

<img width="753" height="469" alt="Image" src="https://github.com/user-attachments/assets/820735d7-96a1-41f5-b106-37968fc8e601" />

wojtekn:
We've added that feature in Studio. The change is merged to the trunk and will be available in the next version.

PR #2527, opened 2026-02-03
Add support for Blueprints login

bcotrim:
## Related issues

- Fixes STU-829 STU-13 #87 

## Proposed Changes

- Remove `login` step from unsupported Blueprint features
- Allow creating a site with custom admin username and password
- Allow editing a site with custom admin username and password

| Create site | Edit site |
|--------|--------|
| image | image | 

## Testing Instructions

- [x] Create site with Blueprint containing `login: { username: "custom", password: "test123" }` — verify credentials prefill in Advanced settings and work in wp-admin
- [x] Create site with Blueprint containing `steps: [{ step: "login", username: "studio-admin", password: "Studio-Admin-Password" }]` — verify same behavior
- [x] Create site without Blueprint — verify default `admin`/random password behavior preserved
- [x] Modify pre-filled admin credentials in Advanced settings — verify form values override Blueprint
- [x] Create site via CLI with `--admin-username` and `--admin-password` — verify credentials work
- [x] Edit existing site credentials via Settings modal — verify restart prompt and credentials applied
- [x] Edit existing site credentials via CLI `set --admin-username newadmin --admin-password newpass` — verify restart and credentials applied
- [x] Change username on a site that already has a custom username — verify new user created, old user left intact, auto-login works
- [x] Try invalid username (special chars, >60 chars, empty) — verify validation errors in both UI and CLI
- [x] Edit a legacy site (no stored credentials) without changing credentials — verify no unnecessary restart
- [x] Verify existing sites, import/export, and sync functionality unaffected

## Pre-merge Checklist


nightnei:
> Change username on a site that already has a custom username — verify new user created, old user left intact, auto-login works

I see this <br />Screenshot 2026-02-12 at 14 30 03
Is it expected?

nightnei:
I see that updating username creates a new user, is it expected?
I had expectation that the old one will be changed.
Screenshot 2026-02-12 at 14 31 18

Maybe we should add some comment here. WDYT?
Screenshot 2026-02-12 at 14 31 47

nightnei:
> Edit a legacy site (no stored credentials) without changing credentials — verify no unnecessary restart

Could you please provide more testing instruction or/and context for this test?

nightnei:
I also tested email changes and everything works well 👍
Shallowly reviewed code changes - all LGTM 👍
Just left a few comments

bcotrim:
Thanks for the review @nightnei 

> > Change username on a site that already has a custom username — verify new user created, old user left intact, auto-login works
> 
> I see this Screenshot 2026-02-12 at 14 30 03 Is it expected?

Yup, this is expected! AFAIK WordPress doesn’t support renaming user_login. So when someone changes the username, we create a new admin user and the old one stays. Added a code comment to make this clearer 👍

> Could you please provide more testing instruction or/and context for this test?

A site created before this PR, we would always use `admin` `admin@localhost.com` and a generated password. Unrelated changes to the site settings (like site name for example) should not prompt for a server restart and have no side effects on user login.

wojtekn:
Can we group and name fields in 'Add site' and 'Edit site' forms consistently?

bcotrim:
> Can we group and name fields in 'Add site' and 'Edit site' forms consistently?

Done

bcotrim:
@wojtekn @nightnei could you give this PR another review please?

tools/common/lib/mu-plugins.ts
```diff
@@ -416,18 +416,49 @@ function getStandardMuPlugins( options: MuPluginOptions ): MuPlugin[] {
...
+						}
 					} else {
+						// WordPress doesn't support renaming user_login, so we create a new admin user.
+						// The old user is left intact — this is intentional.
+						// Generate a unique email to avoid conflicts with existing users
+						$email = $provided_email ? $provided_email : 'admin@localhost.com';
+						if ( ! $provided_email ) {
+							$counter = 1;
+							while ( email_exists( $email ) && $counter < 100 ) {
+								$email = 'admin' . $counter . '@localhost.com';
```
katinthehatsite, line 444: I am wondering if we should use something like: `$email = 'admin-' . uniqid() . '@localhost.com';` for guaranteed uniqueness. It is just a minor thing though so feel free to skip the suggestion

tools/common/lib/mu-plugins.ts
```diff
@@ -416,18 +416,49 @@ function getStandardMuPlugins( options: MuPluginOptions ): MuPlugin[] {
...
+					$user = get_user_by( 'login', $username );
+					$provided_email = ! empty( $_POST['email'] ) ? sanitize_email( $_POST['email'] ) : '';
+
 					if ( $user ) {
 						wp_set_password( $_POST['password'], $user->ID );
+						if ( $provided_email ) {
+							wp_update_user( array( 'ID' => $user->ID, 'user_email' => $provided_email ) );
+						}
 					} else {
+						// WordPress doesn't support renaming user_login, so we create a new admin user.
```
katinthehatsite, line 437: I am wondering if this needs to be communicated to users in some way in the UI 🤔 Could be a notice like: `Changing the username will create a new admin user. The previous admin user will remain in the database.` but then this might be overengineering 🤔

tools/common/lib/passwords.ts
```diff
@@ -1,20 +1,68 @@
...
+
+/**
+ * Validates an admin username and returns an error message, or empty string if valid.
+ */
+export function validateAdminUsername( username: string ): string {
+	if ( ! username.trim() ) {
+		return __( 'Admin username cannot be empty.' );
+	}
+	if ( ! /^[a-zA-Z0-9_.@-]+$/.test( username ) ) {
+		return __( 'Username can only contain letters, numbers, and _.@- characters' );
```
katinthehatsite: I am wondering if we should stretch this along one line even if it goes under Password field. I think it looks a bit cut off now and I am not seeing any harm if it streches out. This will make it consistent with other 

Image

katinthehatsite: As a sidenote, should this end with a period? It looks like all the other strings contain period

tools/common/lib/mu-plugins.ts
katinthehatsite: The reason why I was asking this is because the user could still see it on their local site which could be potentially confusing:

Image

apps/studio/src/modules/add-site/components/create-site-form.tsx
```diff
@@ -490,6 +546,66 @@ export const CreateSiteForm = ( {
...
+										{ __( 'Email' ) }
+									</label>
+									<TextControlComponent
+										id="admin-email"
+										value={ adminEmail }
+										onChange={ ( value: string ) => {
+											hasUserEditedCredentials.current = true;
+											setAdminEmail( value );
+										} }
+										placeholder="admin@localhost.com"
```
katinthehatsite, line 604: I am wondering if instead of a placeholder, we should add below a description such as `The email defaults to `admin@localhost.com` if not added or something along these lines. This could make it clearer to the user what the default is. Not a strong prefernce on my end though

katinthehatsite (commented):
Overall, the functionality looked good to me and I did not find anything major 👍 I left some minor comments for consideration but nothing blocking. I tested with the following file (in case someone wants to use it for testing):

[blueprints.json](https://github.com/user-attachments/files/25416896/blueprints.json)

nightnei (approved):
Thanks @bcotrim, everything LGTM 👍
Also, I like the comment from Kat about adding a description to the UI, that the new user will be created. I would add it.

bcotrim:
@katinthehatsite thanks for the review!
I've addressed most of the feedback. Regarding the username validation error I tried it but it looks strange if both username and password have validation errors:

image

I have decided to keep the error in the field column.

Would love another review when you get a chance! 🙏

katinthehatsite:
> I've addressed most of the feedback. Regarding the username validation error I tried it but it looks strange if both username and password have validation errors:

Oh I see, thanks for exploring @bcotrim - I think what I mostly meant is taking the current error of the username and making it stretch under both fields without introducting the second error for password but I also think we can leave it as it 👍

katinthehatsite (approved):
The changes look good to me, thanks for making the adjustments 👍

wojtekn:
CleanShot 2026-02-27 at 14 16 37@2x

Could we just prefill the field with default value and remove the comment? Wouldn't it be simpler?

bcotrim:
> Could we just prefill the field with default value and remove the comment? Wouldn't it be simpler?

I've added it as a follow-up PR: [Prefill admin email and make it mandatory](https://github.com/Automattic/studio/pull/2681)

PR #2529, opened 2026-02-03
Remove default DB constants from existing sites on startup and site start

sejas:
## Related issues

- Fixes STU-1257

## Proposed Changes

- Remove DB_ default constants from wp-config.php when exporting a full site
- Remove DB_ default constants when a new site is created
- Sites with custom (non-default) database configurations are not modified

## Testing Instructions

- Remove node_modules in the root project and in cli folder
- Run `npm install` in the root folder

**Remove DB_ constants on Studio exports**

- Open the wp-config in any of your Studio sites
- Confirm it contains the DB_NAME and other constants with the default values. If not, you can use wp-config-sample.php to replace your wp-config.php
- Run `npm start`
- Go to the export tab
- Click export entire site
- Observe that the constants were removed from your wp-config.php

https://github.com/user-attachments/assets/45f317d8-6fa1-4621-9c1a-0554190f74e7

**Remove DB_ constants when a site is created**
- Run `npm start`
- Create a new site
- Observe that the site starts correctly
- Open the wp-config.php
- Confirm that it doesn't contain any DB_ constant.

https://github.com/user-attachments/assets/ac622f87-7cf5-424a-a2ea-8b5b368ad718

## Pre-merge Checklist


wojtekn:
@sejas, wouldn't it be enough to remove constants at the time of export/push?

sejas:
I can move the logic to remove the constants on existing sites from Studio startup and run it during export/push 👍 .
From my point of view, it’s also great to remove them during site creation for new sites. So we are consistent and from now on all Studio sites won't have or use DB_ constants in wp-config.php

epeicher (commented):
Thanks @sejas for working on this! I have tested it, and I can confirm that for existing sites, the `wp-config` DB_ constants are removed when starting Studio, but when I create a site, I can still see the constants there and if I push it, I can see the PHP Warnings in the remote site. If I close Studio and open it again, the newly created site has the constants removed

|Before starting Studio | After starting Studio|Creating a site| Warnings when pushing
|-------|------|------|-------|
|  Image  | Image | Image|Image

sejas:
@wojtekn , @epeicher , thanks for reviewing it.
I changed the logic to remove the DB_ constants when a site is created, or a site is exported.
I see I have an issue applying the patch in CI. I'll fix it.

sejas:
@wojtekn, @epeicher, could you review this PR again?

epeicher (approved):
Thanks Antonio! I have tested it and it works as advertised. I was able to:
- Export a site
- Push a site to a remote
- Create a site

And verified that the `wp-config.php` constants are removed as expected

|Before | After|
|-------|------|
| Image| Image|

After pushing a site with the `DB_` constants in `wp-config.php`, I can confirm that I don't see the error logs in my conected wpcom site. 

One question to consider is that if the user modifies `wp-config.php` to add a comment, this solution will not replace the constants. Should we consider this use case? I wonder if an alternative approach, where we parse the file to identify and remove the constants, could be more robust. However, I believe we can address that as a follow-up if we decide it's necessary.

sejas:
> One question to consider is that if the user modifies wp-config.php to add a comment, this solution will not replace the constants. Should we consider this use case? I wonder if an alternative approach, where we parse the file to identify and remove the constants, could be more robust. However, I believe we can address that as a follow-up if we decide it's necessary.

It's a great question. I think the goal of this PR is to remove the constants for new sites, which I don’t expect to require modifications to wp-config.php. In the beginning, I implemented a set of regexes, which was more robust, but I felt more confident using strict string matching.

wojtekn:
@sejas, I'm unsure if it's still a thing, but I have at least one local site where all comments in the sample file are translated to Polish. Also, the original comment could change over time, too.

sejas:
I created a few sites in Polish, and they had the wp-config constnats correctly replaced. Anyway I agree that we can make the logic more flexible to accept any comments. My first solution was the simplest one to write clear code that is safe for all the users. I created [STU-1308](https://linear.app/a8c/issue/STU-1308/studio-make-the-removal-logic-for-db-name-constants-more-flexible) to iterate on this logic.

wojtekn:
Thanks for checking that. My site was created a long time ago, and it may be that at some point, translated WP packages included translated docblocks, too.

PR #2530, opened 2026-02-04
CI: Run E2E tests on `windows` queue, too

mokagio:
## Related issues

- https://linear.app/a8c/issue/AINFRA-1903/run-e2e-on-windows-queue-alongside-mac

## Proposed Changes

- Sets up Buildkite matrix to run E2E tests on the `windows` queue, alongside the already established `mac` run
- Leaves the implementation open for further additions (see discussion in https://github.com/Automattic/studio/pull/2518#discussion_r2760549088)

## Testing Instructions

Refer to the green CI run from https://buildkite.com/automattic/studio/builds/11139/steps/canvas, which run before the commit where I restored the job skip checks that prevent E2E from running when nothing relevant changed.

## Pre-merge Checklist

### DIFF
--- a/.buildkite/commands/run-e2e-tests.sh
+++ b/.buildkite/commands/run-e2e-tests.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-MATRIX=${1:-}
+PLATFORM=${1:-}
+ARCH=${2:-}
 
 if .buildkite/commands/should-skip-job.sh --job-type validation; then
   exit 0
@@ -12,8 +13,19 @@ bash .buildkite/commands/install-node-dependencies.sh
 
 export IS_DEV_BUILD=true
 
-echo '--- :package: Package app for testing'
-npm run package
+echo "--- :package: Package app for testing ($PLATFORM-$ARCH)"
+case "$PLATFORM" in
+  mac)
+    npm run make:macos-"$ARCH"
+    ;;
+  windows)
+    npm run make:windows-"$ARCH"
+    ;;
+  *)
+    echo "Unknown platform: $PLATFORM"
+    exit 1
+    ;;
+esac
 
 echo '--- :playwright: Run End To End Tests'
 echo 'Installing Playwright browsers...'
--- a/.buildkite/commands/run-e2e-tests.sh
+++ b/.buildkite/commands/run-e2e-tests.sh
@@ -32,7 +32,12 @@ esac
 # `make` creates signed distributables (installers), which requires code signing setup.
 # `package` creates an unsigned app bundle, sufficient for E2E testing.
 echo "--- :package: Package app for testing ($PLATFORM-$ARCH)"
-npx electron-vite build --outDir=dist && npx electron-forge package --arch="$ARCH" --platform="$FORGE_PLATFORM"
+npx electron-vite build --outDir=dist
+npx electron-forge package --arch="$ARCH" --platform="$FORGE_PLATFORM"
+
+# This empty echo prevents the next one from being swallowed.
+# See https://buildkite.com/automattic/studio/builds/11098/steps/canvas?jid=019c2287-365e-49c6-b0ba-f7c1d06bc5f5#019c2287-365e-49c6-b0ba-f7c1d06bc5f5/L392
+echo ''
 
 echo '--- :playwright: Run End To End Tests'
 
@@ -44,8 +49,8 @@ if [ "$PLATFORM" = "mac" ] && [ "$ARCH" = "x64" ]; then
   arch_prefix=(arch -x86_64)
 fi
 
-echo 'Installing Playwright browsers...'
+echo '~~~ Installing Playwright browsers...'
 ${arch_prefix[@]+"${arch_prefix[@]}"} npx playwright install
 
-echo 'Running Playwright tests...'
+echo '~~~ Running Playwright tests...'
 ${arch_prefix[@]+"${arch_prefix[@]}"} npx playwright test
--- a/.buildkite/commands/run-e2e-tests.sh
+++ b/.buildkite/commands/run-e2e-tests.sh
@@ -14,19 +14,25 @@ bash .buildkite/commands/install-node-dependencies.sh
 
 export IS_DEV_BUILD=true
 
-echo "--- :package: Package app for testing ($PLATFORM-$ARCH)"
+# Map platform names to electron-forge platform values
 case "$PLATFORM" in
   mac)
-    npm run make:macos-"$ARCH"
+    FORGE_PLATFORM="darwin"
     ;;
   windows)
-    npm run make:windows-"$ARCH"
+    FORGE_PLATFORM="win32"
     ;;
   *)
     echo "Unknown platform: $PLATFORM"
     exit 1
     ;;
 esac
+
+# Use `electron-forge package` instead of `npm run make:*` for E2E tests.
+# `make` creates signed distributables (installers), which requires code signing setup.
+# `package` creates an unsigned app bundle, sufficient for E2E testing.
+echo "--- :package: Package app for testing ($PLATFORM-$ARCH)"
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -30,29 +30,46 @@ steps:
...
+      adjustments:
+        - with:
+            platform: mac
+            arch: x64
+          skip: true
+        - with:
+            platform: windows
+            arch: arm64
+          skip: true
+    # TODO: Re-enable after validating E2E tests on all platforms (AINFRA-1888)
```
mokagio:
```suggestion
```

.buildkite/commands/run-e2e-tests.sh
```diff
@@ -12,11 +13,31 @@ bash .buildkite/commands/install-node-dependencies.sh
...
+    ;;
+  *)
+    echo "Unknown platform: $PLATFORM"
+    exit 1
+    ;;
+esac
+
+# Use `electron-forge package` instead of `npm run make:*` for E2E tests.
+# `make` creates signed distributables (installers), which requires code signing setup.
+# `package` creates an unsigned app bundle, sufficient for E2E testing.
```
AliSoftware, line 32: Thanks for documenting this 🙇 ❤️ 🙏

.buildkite/commands/run-e2e-tests.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-MATRIX=${1:-}
+PLATFORM=${1:-}
+ARCH=${2:-}
```
AliSoftware: The `${1:-}` syntax means that if `$1` is not defined this will default to the empty string instead of failing and exiting early (thanks to `set -u`).

But I think we want the `$1` and `$2` parameters to be mandatory in our case, so we might as well fail early?

> [!TIP]
> If you want to ensure you get a more explicit message when those parameters are missing (as opposed to the generic one provided by `set -u` triggering), you can use the `${1:?Missing platform as first argument to the script}` syntax

AliSoftware:
```suggestion
PLATFORM=${1:?Expected platform to be provided as first parameter}
ARCH=${2:?Expected architecture to be provided as second parameter}
```

.buildkite/pipeline.yml
```diff
@@ -30,26 +30,42 @@ steps:
...
+          - arm64
+      adjustments:
+        - with:
+            platform: mac
+            arch: x64
+          skip: true
+        - with:
+            platform: windows
+            arch: arm64
+          skip: true
```
AliSoftware: Side note: I really wish Buildkite offered a nicer "API" / YAML syntax for those use cases, especially one that would allow us to list the allowed combinations instead of having to use `adjustments` to list the disallowed ones…

It seems I'm not the only one thinking that as [I just found a Feature Request for such a thing in Buildkite forums](https://forum.buildkite.community/t/matrix-object-typed-dimension-items/2292). I'm out of votes on the Buildkite Forums for Feature Requests but if you still have some remaining I'd invite you to +1 that one… who knows, maybe one day we'll get it 🤞

AliSoftware: For the record, I tried tricking the system by using this opt-in syntax instead (i.e. providing empty lists for `matrix.setup` then using `adjustments.with` to [add combinations to the build matrix](https://buildkite.com/docs/pipelines/configure/workflows/build-matrix#adding-combinations-to-the-build-matrix):
```yaml
    matrix:
      setup: { platform: [], arch: [] }
      adjustments:
        - with: { platform: mac, arch: x86_64 }
        - with: { platform: windows, arch: x64 }
```
Sadly, [this resulted in Buildkite not generating any combination at all (0/0)](https://buildkite.com/automattic/studio/builds/11154/steps/canvas)

I then tried to adding at least one entry in each key of `matrix.setup` to check if this was only a side-effect of the initial setup being empty… and it seems that it's the bug I feared indeed:
```yaml
    matrix:
      setup: { platform: [mac], arch: [arm64] }
      adjustments:
        - with: { platform: mac, arch: x86_64 }
        - with: { platform: windows, arch: x64 }
```
This results in [a build generating 3 matrix entries](https://buildkite.com/automattic/studio/builds/11155/steps/canvas) this time, as I'd expect. So this works and I like it better to list what combinations we want rather than having to list the ones we dont (`skip:true`)… but apparently there's a bug that doesn't allow to use that syntax with empty arrays for `matrix.setup` keys and don't create any job at all if we have empty values even if we ask to add more entries via `adjustments`.

This might be worth reporting as a bug to Buildkite Support, as this might have more chances to being fixed if that's just a quick fix (e.g. an overly-eager early-exit path in their code), while [the Feature Request to ask for a different YAML syntax](https://forum.buildkite.community/t/matrix-object-typed-dimension-items/2292) would likely not be prioritized anytime soon…

AliSoftware (approved):
Only left a suggestion about `${1:-}` vs `${1:?error message}`, otherwise LGTM 👍

.buildkite/pipeline.yml
AliSoftware: Taking a quick look at the source code in Buildkite's GitHub I stumbled upon [this comment](https://github.com/buildkite/go-pipeline/blob/a40d93b5ef090f83410d8a96e274d7f422f7cc2d/step_command_matrix.go#L151-L152) in particular, which suggests that my first attempt with `setup: { platform: [], arch: [] }` is supposed to work, and thus it generating an empty list of items is indeed a bug?

[EDIT] I've now sent an email to Buildkite Support with the details of the bug.

AliSoftware: Buildkite Support replied confirming this is a bug and they're going to follow-up on it.

In the meantime they suggested a workaround of [either adding one of the variants in the initial `matrix.setup` and all the other in `adjustments`](https://forum.buildkite.community/t/matrix-object-typed-dimension-items/2292/9), or [add a dummy one in the initial `matrix.setup`, then `skip:true` that dummy one in `adjustments` to then add all the combinations we actually want in `adjustements`](https://github.com/Automattic/studio/blob/5e824a547ea2258ee9ba2caac4639e59f62ad751/.buildkite/pipeline.yml)

So the alternatives we could consider for this particular use case here could be:

```suggestion
      setup: { platform: [mac], arch: [arm64] }
      adjustments:
        - with: { platform: windows, arch: x64 }
```
or this (tested [here](https://buildkite.com/automattic/studio/builds/11176/steps/canvas))
```suggestion
      setup: { platform: [_], arch: [_] } # Dummy base value (because empty array doesn't work, known Buildkite bug)
      adjustments:
        - with: { platform: _, arch: _ }
           skip: true # Skip dummy base value
        - with: { platform: mac, arch: arm64 }
        - with: { platform: windows, arch: x64 }
```

mokagio: Thanks @AliSoftware 

I went for the last option you and support suggested because I like how it keeps the combinations together. The alternative made it look to me like the one in the `setup` line was more important.

mokagio:
I'm going to cash in on @AliSoftware's approval after applying his two suggestions (fail early with error message and streamline `matrix`).

This branch has been restored to not run E2E unless there are changes, but notice I gave one last run on the final setup _just in case_: https://buildkite.com/automattic/studio/builds/11185/steps/canvas?sid=019c2b04-08e9-4bff-9d06-22df680fa296

wojtekn:
Thanks for setting up E2E tests to run on Windows - it will let us catch issues earlier.

.buildkite/pipeline.yml
AliSoftware: Buildkite just told us they fixed the issue on their end 🎉  (and this was apparently a fix on the control node / server side on the infra, i.e. this fix won't even require us to update to a new version of `buildkite-agent` to benefit from it 🎉 )

I've thus opened a follow-up PR to remove the now-unnecessary workaround: https://github.com/Automattic/studio/pull/2574

PR #2532, opened 2026-02-04
Update Xcode to 26.2

wojtekn:
## Related issues

- Related to paaHJt-9AT-p2

## Proposed Changes

- I propose updating Xcode to 26.2.

## Testing Instructions

- Confirm builds go through.

## Pre-merge Checklist


epeicher (approved):
Thanks @wojtekn, the build went through, and I don't think the performance test failures are related so this LGTM!

wojtekn:
@epeicher I think it may be related to the new Xcode image, checking with Apps & Infra.

.buildkite/commands/run-metrics-tests.sh
```diff
@@ -5,6 +5,12 @@ if .buildkite/commands/should-skip-job.sh --job-type metrics; then
   exit 0
 fi
 
+echo "--- :echo: Clean up disk space"
+# This job requires a lot of disk space, because it builds 2 versions of the app to compare their performance.
```
wojtekn: It's a nitpick, but I wouldn't call it "a lot".  😅

.buildkite/commands/run-metrics-tests.sh
```diff
@@ -5,6 +5,12 @@ if .buildkite/commands/should-skip-job.sh --job-type metrics; then
...
+# Our current Xcode VM 26.2 have 23GB of free space but that's not enough, so until we have larger VMs in the future
+# we need to delete some large unused things in the VM to leave enough space for this build. MetalToolchain is a good candidate for that
+xcodebuild -deleteComponent MetalToolchain
```
wojtekn: Thanks for the fix, looks like a good workaround.

AliSoftware: Turns out the MetalToolchain apparently only takes 717MB, so contrary to what I initially thought this is apparently not the thing that makes the 26.2 VM have so much less space than the 26.1 one after all 🤔 

image

iangmaia: ~I also mentioned in Linear, but I guess the additional space comes mostly from the `iOS 18.6` / `watchOS 11.5` runtimes (well, we could delete those as well as a workaround).~
Nvm, I caught up with the thread and saw this was already figured out and fixed in the image 👍

AliSoftware:
@wojtekn After re-publishing the `xcode-26.2` VM with increased disk size, the Performance Metrics job is now passing without any "no space left on device" anymore 🎉

AliSoftware (approved):

wojtekn:
Thanks for helping with this @AliSoftware 💪

PR #2534, opened 2026-02-04
Fix site start/stop status by removing stale socket file

bcotrim:
## Related issues

- Fixes STU-1278 STU-1255

## Proposed Changes

- Remove stale Unix socket file (`~/.studio/pm2/events.sock`) before binding in the `_events` command
- Only applies to Unix-based systems (macOS/Linux) since Windows named pipes are automatically cleaned up by the OS

**Root cause:** When Studio crashes or is force-killed, the socket file is left behind. On restart, the `_events` command cannot bind to this socket (EADDRINUSE), preventing site-event updates from flowing to the UI.

## Testing Instructions

1. Start Studio in dev mode (`npm start`)
2. Verify normal start/stop works - click Start/Stop on a site, status should update
3. Force-kill Studio without proper cleanup:
   ```bash
   pkill -9 -f "Electron.app"
   ```
4. Verify the stale socket file exists:
   ```bash
   ls -la ~/.studio/pm2/events.sock
   ```
5. Start Studio again (`npm start`)
6. Click Start/Stop on a site - status should update correctly (the fix removes the stale socket automatically)

## Pre-merge Checklist

### DIFF
--- a/cli/commands/_events.ts
+++ b/cli/commands/_events.ts
@@ -6,6 +6,7 @@
  * stdout key-value pairs that Studio parses.
  *
  */
+import fs from 'fs';
 import { __ } from '@wordpress/i18n';
 import { sequential } from 'common/lib/sequential';
 import { SITE_EVENTS, siteDetailsSchema, SiteEvent } from 'common/lib/site-events';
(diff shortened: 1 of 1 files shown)


sejas (approved):
Nice! it solved my issue! 👌
Now I can see the sites starting and stopping as expected.

fredrikekelund:
pm2-axon is _supposed_ to automatically handle `EADDRINUSE` errors in `Socket::bind`. See [source](https://github.com/Unitech/pm2-axon/blob/e012e208f82b697fd8c0fa9795602cb3a7c82f54/lib/sockets/sock.js#L344-L415).

The gist of the logic is that:
- If `net.createServer` emits an `EADDRINUSE` error.
  - Create a `net.Socket` and try to connect to the socket.
  - If this fails, it means that no server is listening on the socket, so pm2-axon removes the socket file and connects again. 

If this fixes the underlying problem, that's great 👍 Still, I'm curious how the pm2-axon logic fails and why…

PR #2537, opened 2026-02-05
Upgrade node to v24 only for CLI

katinthehatsite:
## Related issues

Related to: https://github.com/Automattic/studio/pull/2486#issuecomment-3851973868

## Proposed Changes

This PR ensures that we upgrade node version only for CLI and not for the whole Studio project. Electron is currently using v39.x and is not compatible with Node v24. I explored upgrading Electron so that CLI and the rest of the Studio project could use the same configuration but ran into a blocker with `fs-ext`. Additionally, according to this comment: https://github.com/Automattic/studio/pull/2486#issuecomment-3852363811 , it seems that our goal is to untangle with node versions used by CLI and Studio anyway. 

## Testing Instructions

* Confirm that in CI, we get the node 22.x installed:

Screenshot 2026-02-05 at 11 17 04 AM

To test CLI node version: 

* Run `nvm use && npm install`
* Run `npm run cli:build`
* Run `./bin/node --version`
* Confirm you see `node v24.11.0`
* For good measure, confirm that `./bin/node dist/cli/main.js --help` works as expected

## Pre-merge Checklist


scripts/download-node-binary.ts
```diff
@@ -100,7 +102,11 @@ async function download( downloadUrl: string, dest: string ): Promise< void > {
 	console.log( 'Download complete.' );
 }
 
-async function extractTarGz( archivePath: string, destDir: string, binaryName: string ): Promise< void > {
+async function extractTarGz(
+	archivePath: string,
+	destDir: string,
+	binaryName: string
+): Promise< void > {
```
katinthehatsite, line 97: These are unrelated linter fixes

scripts/download-node-binary.ts
```diff
@@ -163,4 +173,4 @@ async function main(): Promise< void > {
 	}
 }
 
-main();
+void main();
```
katinthehatsite, line 164: Unrelated linter fix

.nvmrc
```diff
@@ -1 +1 @@
-24.11.0
+22.20.0
```
katinthehatsite, line 1: These are the changes reverted back from https://github.com/Automattic/studio/pull/2486

.bundled-node-version
```diff
@@ -0,0 +1 @@
+24.11.0
```
katinthehatsite: This file could be called something else if this is not clear. I also thought about hard-coding it directly into `download-node-binary` but it could be slightly harder to find that way

wojtekn: If it's used only in one place and is not an industry-standard file like `.nvmrc`, we could hardcode it there. There is no value in creating a new file format for that.

We follow the same approach with the SQLite plugin version.

katinthehatsite: Sounds good to me, made changes in https://github.com/Automattic/studio/pull/2537/changes/d9021ad87b4f7708494bb6e737b558c603ecdc96

I also removed the fallback as it should not be needed as the version is always defined.

scripts/download-node-binary.ts
```diff
@@ -6,21 +6,17 @@
...
 import fs from 'fs';
-import path from 'path';
 import os from 'os';
+import path from 'path';
 import { extract } from 'tar';
 import { extractZip } from '../common/lib/extract-zip';
 
-const LTS_FALLBACK = 'v22.12.0';
+// Node.js version to bundle with the CLI
+// This is separate from .nvmrc which controls the development environment
```
wojtekn:
```suggestion
// This is separate from .nvmrc which controls the development environment and Electron app
```

wojtekn (approved):

scripts/download-node-binary.ts
```diff
@@ -6,21 +6,17 @@
...
+import path from 'path';
 import { extract } from 'tar';
 import { extractZip } from '../common/lib/extract-zip';
 
-const LTS_FALLBACK = 'v22.12.0';
+// Node.js version to bundle with the CLI
+// This is separate from .nvmrc which controls the development environment and Electron app
+const BUNDLED_NODE_VERSION = 'v24.11.0';
 
 function getNodeVersion(): string {
```
ivan-ottinger: I think we could now get rid of the wrapper `getNodeVersion` now - since it returns one constant only.

ivan-ottinger (commented):
Thanks for the changes, Kat.

scripts/download-node-binary.ts
katinthehatsite: Thanks, should be updated 👍

ivan-ottinger: Thanks for the update, Kat! Looks good! 👍🏼

ivan-ottinger (approved):

fredrikekelund:
Connecting the dots between https://github.com/Automattic/studio/pull/2486#issuecomment-3876177073 and this change…

`.nvmrc` dictates which node version we use for development, primarily for Studio CLI and potentially (though unlikely, IMO) also for the Electron "back-end logic". It doesn't dictate which node version Electron uses for the "back-end logic" after the app is packaged.

We should not use a different node version for Studio CLI development compared to what we ship in the installer, since we won't be able to catch potential issues related to the node version. 

@wojtekn, I'll experiment with the approach we took in https://github.com/Automattic/studio/pull/2486 and try to reproduce the error you described in https://github.com/Automattic/studio/pull/2486#issuecomment-3876327484

PR #2538, opened 2026-02-05
Fix WP-CLI command queue saturation during push operations

bcotrim:
## Related issues

- Fixes STU-1283

## Proposed Changes

- Add opt-in `deduplicateKey` option to `sequential.ts` to prevent identical commands from consuming queue slots. When enabled, concurrent calls with the same key share a single in-flight promise.
- Enable dedup for `runWpCliCommand` using command args as key, so repeated `get-theme-details` calls don't pile up in the queue.
- Increase WP-CLI queue max from 10 to 100 as a safety net.
- Surface WP-CLI error messages in failure handler by falling back to `error.lastErrorMessage` when stderr is empty, so "Queue is full" errors appear in logs.

## Testing Instructions

- `npm test -- common/lib/tests/sequential.test.ts` — Verify dedup behavior: concurrent calls with same key share a promise, different keys execute separately, dedup covers queued calls.
- `npm test` — Verify no regressions (1177 tests pass).
- Manual: Fire 15 concurrent `studio wp option get siteurl` commands — should no longer hit queue limit.
- Manual: Trigger export on site with many tables — should complete without "Queue is full" error.
- Verify logs show error messages when WP-CLI commands fail (not empty stderr).

## Pre-merge Checklist


common/lib/sequential.ts
```diff
@@ -1,37 +1,62 @@
...
 
-		try {
-			locks.add( fnPromise );
-			return await fnPromise;
-		} finally {
-			locks.delete( fnPromise );
+		const resultPromise = execute();
+		if ( key ) {
+			inFlightByKey.set( key, resultPromise );
 		}
```
sejas, line 57: If the queue is full I think we are still saving the failed promise in the `inFlightByKey` and blocking that space "for ever". I think the next the same command with same parameters is executed it will return the queue is full.

sejas (commented):
Code looks good. I just left a suggestion to investigate about what happens when reaching the 100 limit.

common/lib/sequential.ts
bcotrim: Nice catch! I've fixed the issue

sejas (approved):
It works as expected. I was able to export my site and run wp-cli commands on it.
Thanks for addressing the feedback and adding a new test case for that.

PR #2545, opened 2026-02-06
Fix npm version enforcement and remove invalid .npmrc

wojtekn:
## Related issues

- Related to npm/cli#7961

## Proposed Changes

This fixes the potential Buildkite build failure where appdmg was not being installed.

The invalid .npmrc can cause npm to misbehave and skip installing optional dependencies like appdmg (required for macOS DMG builds). This issue emerged after the Node 22 update, as npm 10.9.x has a known bug with platform-specific optional dependencies being pruned unexpectedly.

The packageManager field is the correct, official way to enforce npm versions across local dev and CI environments using Node's built-in Corepack.

## Testing Instructions

- Confirm build goes through.

## Pre-merge Checklist


katinthehatsite:
The changes look good to me. I actually had a very similar commit to my PR yesterday - https://github.com/Automattic/studio/pull/2537/changes/5e2adc04ce7612099f470e7954b1084349a07952 - when I was researching upgrading the node version. It wasn't the fix for the issue I was working on, but it came up as something that needed to be potentially fixed.

epeicher (approved):
Thanks for the great explanation @wojtekn! The changes make sense to me and the build went through. LGTM!

package.json
```diff
@@ -7,6 +7,7 @@
 	"description": "Local WordPress development environment using Playgrounds",
 	"license": "GPL-2.0-or-later",
 	"main": "dist/main/index.js",
+	"packageManager": "npm@10.9.4",
```
epeicher, line 10: [According to the documentation](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#devengines), this should be under `devEngines` entry. I am not familiar with this, but I wanted to double-check. 

EDIT: It seems that this field is used by Corepack [reference here](https://github.com/nodejs/corepack?tab=readme-ov-file#when-authoring-packages) so this is probably fine 👍

PR #2547, opened 2026-02-06
Make Studio CLI available in AppX via AppExecutionAlias

gcsecsey:
## Related issues

- Fixes STU-1249
- Follow-up to #2465

## Proposed Changes

On Windows AppX (Microsoft Store) installations, the `studio` CLI command was not available in the terminal because AppX packages require special handling. This PR:

- Adds AppExecutionAlias support by compiling a Node.js launcher script to a standalone .exe via `@yao-pkg/pkg`, which is configured as a Windows AppExecutionAlias in the AppxManifest. When users type `studio` in a terminal, Windows activates the package context and runs the CLI.
- Hides the CLI toggle for Store users. Since the CLI is automatically installed with the AppX package, the Settings toggle to install/uninstall the CLI is hidden for Store installations using Electron's `process.windowsStore` flag for detection.

The PR uses a new dependency `@yao-pkg/pkg`, but node.js also has a [single executable applications](https://nodejs.org/api/single-executable-applications.html) feature which could be used to compile it. The downsides of using SEA would be:
- no cross-compilation at this time (Node 26+ will ship with --build-sea command that supports this)
- slightly bigger binary
- requires a slightly more complex build script

If not adding a new dependency is more important than the binary size, we could also adapt this approach, because our CI builds are run on Windows already.

## Testing Instructions

Store version:

1. Download and install the signed .appx package from the CI run
2. Open a terminal and run `where studio`, and check that it outputs a path similar to `C:\Users\gcsecsey\AppData\Local\Microsoft\WindowsApps\studio.exe`
3. Run other CLI commands, eg. `studio site list`, these should work as expected
4. Start Studio and check that Settings > Preferences > Studio CLI toggle is NOT visible

Installer version:

0. Remove the Store version if it's installed
1. Download and install the studio-setup.exe from the CI run
2. Open a terminal and run `where studio`, and check that it outputs a path similar to `C:\Users\gcsecsey\AppData\Local\studio\bin\studio.bat`
3. Run other CLI commands, eg. `studio site list`, these should work as expected
4. Start Studio and check that Settings > Preferences > Studio CLI toggle IS visible

Also check for any regressions on Mac.

| Windows Store version | Installer version | Mac (regression) |
|--------|--------|------|
| CleanShot 2026-02-06 at 12 37 16@2x | CleanShot 2026-02-06 at 12 47 39@2x | CleanShot 2026-02-09 at 12 32 49@2x |
| image | CleanShot 2026-02-06 at 12 48 07@2x | CleanShot 2026-02-09 at 12 32 36@2x |

## Pre-merge Checklist


ivan-ottinger (commented):
Thank you for detailed description and testing instructions, Gergely. That helps!

I can confirm the changes work correctly. I did not observe any regressions.

**Installer version:**

Image

Image

**Store version:**

Image

Image

**macOS version**

Image

Image

I haven't reviewed the code yet though, but will do so tomorrow with fresh brain.

ivan-ottinger (approved):
The changes look good as well! I did not observe any issues. 👍🏼 

We will just need to resolve conflicts In `package-lock.json`.

PR #2548, opened 2026-02-06
Add ability to start all stopped sites

wojtekn:
## Related issues

- Fixes STU-84

## Proposed Changes

I propose to add an ability to start all stopped sites. I kept "Start all" only for case in which no sites are running, and "Stop all" for all other cases to keep UI simple.

CleanShot 2026-02-06 at 14 01 29

CleanShot 2026-02-06 at 14 01 47

CleanShot 2026-02-06 at 14 02 24

## Testing Instructions

1. Create 2+ sites in the app
2. Stop all sites — verify "Start all" button appears in sidebar
3. Click "Start all" — verify all sites start
4. Stop some sites — verify "Stop all" button is shown and "Start" is disabled

### DIFF
--- a/src/components/running-sites.tsx
+++ b/src/components/running-sites.tsx
@@ -4,34 +4,45 @@ import Button from 'src/components/button';
 import { useSiteDetails } from 'src/hooks/use-site-details';
 import { cx } from 'src/lib/cx';
 
+const linkButtonClassName = cx(
+	'[&.is-link]:text-white [&.is-link:disabled]:hover:text-white [&.is-link:not(:disabled)]:hover:text-a8c-gray-10 [&.is-link]:text-right text-xxs leading-4 !mb-0 items-start'
+);
+
 export function RunningSites() {
 	const { __, _n } = useI18n();
-	const { sites, stopAllRunningSites } = useSiteDetails();
+	const { sites, stopAllRunningSites, startAllStoppedSites, loadingServer } = useSiteDetails();
 
 	const runningSites = sites.filter( ( site ) => site.running );
+	const realSites = sites.filter( ( site ) => ! site.isAddingSite );
+	const hasLoadingSites = Object.values( loadingServer ).some( Boolean );
+	const anyRunning = runningSites.length > 0;
+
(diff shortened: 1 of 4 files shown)


src/components/running-sites.tsx
```diff
@@ -4,34 +4,45 @@ import Button from 'src/components/button';
...
-						onClick={ stopAllRunningSites }
-						variant="link"
-					>
-						{ runningSites.length === 1 ? __( 'Stop' ) : __( 'Stop all' ) }
-					</Button>
-				</div>
+		<div className="flex flex-row px-5 pb-1 justify-between align-center self-stretch opacity-70">
+			<p className="text-xxs leading-4">
+				{ sprintf(
+					_n( '%d site running', '%d sites running', runningSites.length ),
```
gcsecsey: This label was earlier only shown when any site was started, but now `0 sites running` are shown by default. Maybe we could handle this case separately and show a `No sites running` label?

Image

gcsecsey (approved):
Thanks for working on this @wojtekn, it's a very useful small enhancement. 🙌 

The changes LGTM and when testing starting/stopping sites, it works well.

| All sites stopped | Some sites stopped | All sites started | 
|--------|--------|----|
| Image | Image | Image |

sejas (approved):
I confirm I'm able to start all the sites. The button is visible only if all the sites are stopped.
Code looks good. It iterates for each site and call startServer for each.
https://github.com/user-attachments/assets/64ac3441-0453-493c-9c91-4f5a3074081e

src/components/running-sites.tsx
wojtekn: Makes sense, I changed it in e38caf70b254012a6b51aeb1d11c90837700c409 so we will keep the current behavior. I will keep the opposite case as is, as it is a nice indicator of the number of all sites.

bcotrim (approved):
LGTM 👍 
I wonder if we could add a `studio site start --all` similar to the existing `studio site stop --all` command, to keep CLI parity.
This discussion could be a follow up PR, if we decide to implement it, since the current change works perfectly already

PR #2549, opened 2026-02-06
Fix site thumbnail not refreshing after content changes

sejas:
## Related issues

- Fixes STU-1284

## Proposed Changes

- Regenerate the selected site thumbnail each time the user changes the theme, or focus on Studio

## Testing Instructions

- Run `npm start`
- Start your site
- See the thumbnail works as expected
- Start a wp-admin
- Publish a new post, or edit the site
- Focus back to Studio App
- Confirm the thumbnail has been updated

https://github.com/user-attachments/assets/a0425252-32ca-4da5-9043-1f4773c655c9

## Pre-merge Checklist


sejas:
I moved the thumbnail update to run every time `loadThemeDetails` is executed. This is the most effective way to keep the thumbnail updated and I didn't observe any performance issue. Other alternatives could include checking whether the database has been modified or loading an iframe of the site if it is running.

src/tests/ipc-handlers.test.ts
```diff
@@ -349,3 +358,65 @@ describe( 'getXdebugEnabledSite', () => {
 		} );
 	} );
 } );
+
+describe( 'loadThemeDetails', () => {
+	it( 'should update thumbnail even when theme has not changed', async () => {
```
ivan-ottinger: I wonder if we could merge this first new test with the last one.

ivan-ottinger (approved):
Thank you for the fix, Antonio! I can confirm the issue is no longer there and the thumbnail gets refreshed correctly 🙂👍🏼 :

https://github.com/user-attachments/assets/b080cbd8-f1a5-4cc1-bb82-2bbe17c67d22

The refresh happens correctly on theme change as well. 👍🏼 

The only thing I noticed is that we are not using the thumbnail-loading event when the thumbnail is refreshing, but I think that's even better in this case - as I found the loading on placeholder on app focus distracting.

---

I also left one inline comment.

src/tests/ipc-handlers.test.ts
sejas: Great suggestion. Done! https://github.com/Automattic/studio/pull/2549/commits/7f421b8d6f37b0a62d6037a373c04a217d8d690f

ivan-ottinger: Thanks, Antonio! 🙂

Issue #23, opened 2024-04-24
Feature Request: one click to enable debug.log


mdrovdahl:
### What

I would love a UI button that updates wp-config.php and enables debug.log per https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/#example-wp-config-php-for-debugging

### Why

This is a very common step for local development, easily automated, and adding it as a UI option would save me time over-and-over.

### How

Probably belongs under "Settings"

CHEWX:
Ideally options for `WP_DEBUG_LOG` and `WP_DEBUG_DISPLAY` independently.

andrewjmead:
This would be amazing. It would be nice to see styled logs right there in the app too.

ndiego:
+1 for this

faisalahammad:
+1

ahegyes:
+1

getdave:
+1 This needs to happen 🙏

bgrgicak:
+1

PR #2553, opened 2026-02-08
Add WP_DEBUG_LOG and WP_DEBUG_DISPLAY toggles

bcotrim:
## Related issues

Fixes STU-4 #23

## Proposed Changes

- Add WP_DEBUG_LOG and WP_DEBUG_DISPLAY toggles in a new "Developer" tab within the Edit Site modal
- Xdebug control moved to Developer tab alongside new debug toggles
- WP_DEBUG automatically enabled when either WP_DEBUG_LOG or WP_DEBUG_DISPLAY is toggled on
- Developer settings grouped in read-only Settings view
- Full CLI support: `studio site set --debug-log --debug-display` with `--no-` variants
- Server restart triggered when debug settings change

image

## Testing Instructions

1. Open WordPress Studio and select a site
2. Click "Edit site" → Developer tab
3. Toggle "Enable debug log" and "Show errors in browser"
4. Save and verify the site restarts
5. Check that wp-content/debug.log is created
6. Via CLI: `studio site set --path /path/to/site --debug-log --debug-display`

## Pre-merge Checklist


ivan-ottinger (commented):
This will be a nice improvement! Great work!

Despite disabling the logging settings, the `debug.log` file is still being created on a fresh site.

From what I have seen so far, this is likely related to Playground's default PHP error logging that is overriding our settings.

---

Another, but minor design thing I noticed is that the tabs line does not stretch to the sides in a way we have on other existing modal in the app:

Settings:
Image

Edit site:
Image

---

Apart from the above, the UI looks clean to me and CLI commands run well too. 🙂 

Image

Image

Image

wojtekn:
> From what I have seen so far, this is likely related to Playground's default PHP error logging that is overriding our settings.

There was a discussion with the Playground team on reverting those to the default state. I removed those under https://github.com/WordPress/wordpress-playground/pull/3131, which was closed, but @adamziel mentioned it was solved in another way.

katinthehatsite:
> Despite disabling the logging settings, the debug.log file is still being created on a fresh site.

I am not sure why but I did not see this issue when testing: I tried with the new site and enabled and disabled it multiple times and the setting was reflected correctly 🤔

apps/studio/src/components/content-tab-settings.tsx
```diff
@@ -138,9 +138,20 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
 							<span className="line-clamp-1 break-all">{ selectedSite.phpVersion }</span>
 						</div>
 					</SettingsRow>
+					<tr>
+						<th colSpan={ 2 } className="pb-4 ltr:text-left rtl:text-right">
+							<h3 className="text-black text-sm font-semibold mt-4">{ __( 'Developer' ) }</h3>
```
katinthehatsite: I think we might want to remove the padding of 1rem for the `General` tab to make sure that the tab looks more aligned.

bcotrim:
@ivan-ottinger nice catch!

The debug.log being created with settings off was caused by Playground’s 0-playground.php which unconditionally sets error_log to `wp-content/debug.log` and has `log_errors=1` in its default php.ini. Any PHP warning would create the file.

Fixed by adding a Studio mu-plugin (`0-configure-error-logging.php`) that sets ini_set('log_errors', '0') when WP_DEBUG_LOG is disabled. This runs after Playground’s setup but before any warnings get triggered.

The site behavior should now match the selected options. Let me know what you think.

wojtekn:
@bcotrim, can we check this with the Playground team? It seems like something that should be fixed on the Playground side.

apps/studio/src/modules/site-settings/edit-site-details.tsx
```diff
@@ -205,170 +220,247 @@ const EditSiteDetails = ( { currentWpVersion, onSave }: EditSiteDetailsProps ) =
...
-									disabled={ isEditingSite }
-									onChange={ setSiteName }
-									value={ siteName }
-								></TextControlComponent>
-							</label>
+						<TabPanel
+							className="w-full [&>[role=tabpanel]]:h-64 [&>[role=tabpanel]]:overflow-auto"
+							tabs={ [
+								{ name: 'general', title: __( 'General' ) },
+								{ name: 'developer', title: __( 'Debugging' ) },
```
ivan-ottinger: I think we could change the name here (and in related references) as well (to `debugging`).

ivan-ottinger (approved):
> The debug.log being created with settings off was caused by Playground’s 0-playground.php which unconditionally sets error_log to `wp-content/debug.log` and has `log_errors=1` in its default php.ini. Any PHP warning would create the file.
> 
> Fixed by adding a Studio mu-plugin (`0-configure-error-logging.php`) that sets ini_set('log_errors', '0') when WP_DEBUG_LOG is disabled. This runs after Playground’s setup but before any warnings get triggered.
> 
> The site behavior should now match the selected options. Let me know what you think.

Thank you for the updates, Bernardo! The changes look good to me and work well. I did not observe any further major issues, just a tiny naming thing (please see the inline comment below) and a different styling for disabled elements when when saving:

https://github.com/user-attachments/assets/de1be7dc-23f7-4c53-a861-b7dc449af7f7

(xdebug checkbox is disabled differently)

---

I am aware of Wojtek's comment on fixing the override on the Playground side. I am approving this PR in case we decide to move forward with the current approach.

bcotrim:
> @bcotrim, can we check this with the Playground team? It seems like something that should be fixed on the Playground side.

Shared: p1771600191281759-slack-C09GMMV3T6J

Meanwhile should we merge this PR or hold on to Playground and design feedback?

apps/studio/src/modules/site-settings/edit-site-details.tsx
katinthehatsite: > xdebug checkbox is disabled differently

I noted this as well while reviewing this PR, I think it looks better now with recent changes

katinthehatsite (approved):
I think this looks good, I tested this on my end once again and checked the code and did not see any regressions. 👍

wojtekn:
> Meanwhile should we merge this PR or hold on to Playground and design feedback?

UI looks solid, so I think we should be good to merge.

We could consider changing the MU plugin to an NPM patch for the Playground package, so it will be easier to notice when the fix arrives in the Playground core and to clean up our hotfix.

wojtekn:
@bcotrim one more thing - looking at labels, it may be unclear for some users what happens behind those new controls. Would it make sense to update descriptions and say "[...] by setting WP_DEBUG constant"?

bcotrim:
@wojtekn done in https://github.com/Automattic/studio/pull/2553/commits/0ddde2a76d6b4712d97bfda58bb94dbe8fee2214

wojtekn:
We've added settings that allow enabling the logs:

CleanShot 2026-02-24 at 09 31 24@2x

The feature will be included in the incoming 1.7.5 release.

PR #2558, opened 2026-02-09
Bundle WordPress translation files for faster site creation

ivan-ottinger:
## Related issues

Closes STU-1250
Related to STU-980

## Proposed Changes

The PR proposes to bundle WordPress core, plugin and theme translation files. This saves about 3 - 4 seconds in localized site creation time while adding only about 29 MB to the app bundle and additional time when running `npm install`.

Feel free to check the comments with the background in the related task (STU-1250).

- Download WordPress core, plugin, and theme translation files at build time via `scripts/download-wp-server-files.ts` for all 19 supported non-English locales.
- At app startup, copy bundled language packs from app resources to the runtime `server-files/language-packs/` directory.
- During site creation for the latest WP version, copy locale-specific `.l10n.php` and `.json` files directly into the site's `wp-content/languages/` directory and set `WPLANG` via blueprint steps (`defineWpConfigConsts` + `setSiteOptions`) — no network round-trip needed.
- Fall back to the original `setSiteLanguage` blueprint step for non-latest WP versions or when bundled packs are unavailable.
- Auto-detect bundled plugins and themes from the extracted WordPress installation for translation downloads.
- Only bundle `.l10n.php` and `.json` files (WordPress 6.5+ format), skipping `.po` and `.mo` to reduce size (~29MB vs ~82MB).
- Make `setupWPServerFiles` resilient so individual step failures don't prevent subsequent steps from running.

## Testing Instructions
1. Run `npm install` and verify language packs are downloaded without errors. You can find them in the project root: `wp-files/latest/languages`, `wp-files/latest/languages/plugins` and `wp-files/latest/languages/themes`.
CleanShot 2026-02-10 at 13 49 40@2x

2. Run `npm start` and set your system locale to a non-English language (e.g. Swedish).
3. Create a new site and verify:
   - Site creation completes without downloading translations from WordPress.org → It should take similar time as when creating an English site (6 - 8 seconds vs 11 - 12 seconds in my tests).
   - The WordPress admin is displayed in the selected language.
   - Plugin and theme translations are present (please note that some locales are missing translations for certain plugins and themes).
4. Create a site with a non-latest WP version (e.g. 6.5) and verify it falls back to downloading translations → The site creation should take about 3 - 4 seconds longer.
5. Create a site with English locale and verify everything is going well there too. There should be no translations in `wp-content`.
6. Run `npm test -- cli/commands/site/tests/create.test.ts` and verify all tests pass.

## Pre-merge Checklist

### DIFF
--- a/scripts/download-wp-server-files.ts
+++ b/scripts/download-wp-server-files.ts
@@ -166,7 +166,8 @@ async function downloadTranslationsFromApi(
 			continue;
 		}
 
-		const zipPath = path.join( os.tmpdir(), `wp-language-${ label }-${ language }.zip` );
+		const safeLabel = label.replace( /\//g, '-' );
+		const zipPath = path.join( os.tmpdir(), `wp-language-${ safeLabel }-${ language }.zip` );
 		const buffer = Buffer.from( await zipResponse.arrayBuffer() );
 		await fs.writeFile( zipPath, buffer );
 		await extractZip( zipPath, destPath );
--- a/scripts/download-wp-server-files.ts
+++ b/scripts/download-wp-server-files.ts
@@ -131,14 +131,18 @@ interface TranslationsApiResponse {
 	translations: TranslationEntry[];
 }
 
-async function downloadLanguagePacks(): Promise< void > {
-	const languagesPath = path.join( WP_SERVER_FILES_PATH, 'latest', 'languages' );
-	await fs.ensureDir( languagesPath );
+// Plugins and themes bundled with the default WordPress installation.
+const BUNDLED_PLUGINS = [ 'akismet' ];
+const BUNDLED_THEMES = [ 'twentytwentyfive', 'twentytwentyfour', 'twentytwentythree' ];
 
-	console.log( '[language-packs] Fetching available translations ...' );
-	const response = await fetch( 'https://api.wordpress.org/translations/core/1.0/' );
+async function downloadTranslationsFromApi(
+	apiUrl: string,
+	destPath: string,
+	label: string
+): Promise< void > {
+	const response = await fetch( apiUrl );
 	if ( ! response.ok ) {
-		throw new Error( `Failed to fetch translations API (status ${ response.status })` );
+		throw new Error( `Failed to fetch ${ label } translations API (status ${ response.status })` );
 	}
 	const data: TranslationsApiResponse = await response.json();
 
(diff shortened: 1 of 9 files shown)


scripts/download-wp-server-files.ts
```diff
@@ -99,6 +100,125 @@ async function downloadFile( file: FileToDownload ): Promise< void > {
...
+		const safeLabel = label.replace( /\//g, '-' );
+		const zipPath = path.join( os.tmpdir(), `wp-language-${ safeLabel }-${ language }.zip` );
+		const buffer = Buffer.from( await zipResponse.arrayBuffer() );
+		await fs.writeFile( zipPath, buffer );
+		await extractZip( zipPath, destPath );
+		await fs.remove( zipPath );
+	}
+}
+
+async function removePoAndMoFiles( dirPath: string ): Promise< void > {
```
ivan-ottinger, line 173: I decided to drop the `.po` and `.mo` translation files here - in favor of performant `.l10n.php` (more context: https://make.wordpress.org/core/2024/02/27/i18n-improvements-6-5-performant-translations/).

This approach takes the additional bundle size down to about 29 MB - compared to around 80 MB if we included `.po` and `.mo` files.

One drawback I see in this approach is that `hello-dolly` plugin has the new `.l10n.php` translation files for French and Korean only so far. This means that this plugin won't be translated in other locales when a new site is created.

Curious to hear your point of view.

scripts/download-wp-server-files.ts
```diff
@@ -99,6 +100,125 @@ async function downloadFile( file: FileToDownload ): Promise< void > {
...
+
+	// Core translations
+	await downloadTranslationsFromApi(
+		'https://api.wordpress.org/translations/core/1.0/',
+		languagesPath,
+		'core translations'
+	);
+
+	// Plugin translations
+	const pluginSlugs = getBundledSlugs( path.join( wpContentPath, 'plugins' ) );
```
ivan-ottinger, line 197: Initially I was just listing specific plugins and themes to bundle their translations, but then decided to automatically get the list based on what's included in the current WordPress.

This approach will ensure we won't need to update the list of plugins and themes when they change in the WordPress (e.g. when a new core default theme is added).

scripts/download-wp-server-files.ts
gcsecsey: Overall I think this tradeoff would be worth it to save on the bundle size.

I also saw that Akismet also only has 19 translations in the `.l10n.php` format out of the [75 shown on WP.org](https://wordpress.org/plugins/akismet/).

Image

Do you know if there's a plan to migrate these to the more performant version in the future maybe?

scripts/download-wp-server-files.ts
gcsecsey: I think this is a great approach to avoid the maintenance burden. 👍

gcsecsey (approved):
Thanks for working on this @ivan-ottinger! 🙌 Functionally, this is working well!

I confirm the language packs are now downloaded when starting the app:

Image

The `wp-files/latest/languages` directory and its subdirectories are populated:

Image

When creating a site using the newest WordPress version, translations are present in the site's `wp-content/languages` directory, and omitted for English sites:

| German site | English site |
|--------|--------|
| Image | Image |

scripts/download-wp-server-files.ts
ivan-ottinger: > I also saw that Akismet also only has 19 translations in the .l10n.php format out of the [75 shown on WP.org](https://wordpress.org/plugins/akismet/).

We only keep the files of locales we need for Studio (19 locales) and drop the rest. 🙂 

> Do you know if there's a plan to migrate these to the more performant version in the future maybe?

I haven't found any info on that, but for what we need, only hello-dolly seems to be affected. Moving forward into the future, I expect all new plugins and themes will rely on the new translation format.

gcsecsey: > We only keep the files of locales we need for Studio (19 locales) and drop the rest. 🙂

Aaah, that makes sense, thanks! 😄 

> I haven't found any info on that, but for what we need, only hello-dolly seems to be affected. Moving forward into the future, I expect all new plugins and themes will rely on the new translation format.

Yeah, I think so too.

sejas (approved):
This is a super cool feature! and the best of all it helps when creating a new site and the computer is offline.

Creating a WordPress site in Spanish in offline mode:
| Before | After |
|--------|--------|
|  Image | Image |

It increases the time of installing dependencies by one minute, but I think we can improve it as a follow-up by downloading the wp-files in parallel. Happy to own that task after you merge this PR.

**BEFORE:**

```
> time npm install
npm install  11.26s user 3.56s system 15% cpu 1:36.16 total
```
**AFTER:**

```
> time npm install
npm install  6.05s user 1.85s system 50% cpu 15.705 total
```

Also, I would recommend skipping the languages for older themes like 2024 and 2023 themes.

Image

ivan-ottinger:
Thank you for your reviews and testing, Gergely and Antonio!

> It increases the time of installing dependencies by one minute, but I think we can improve it as a follow-up by downloading the wp-files in parallel. Happy to own that task after you merge this PR.

Great idea! Looks like you already have the code locally so feel free to create the PR for it. 🙂 Happy to review it.

> Also, I would recommend skipping the languages for older themes like 2024 and 2023 themes.

At the moment it downloads translations for all bundled themes - but true is that only one is active. I agree it is a good idea to save a bit more storage and time. We could indeed skip 2023 and 2024. That should be a quick change. If you like, feel free to include it in your wp-files parallel copy PR Antonio. Otherwise I will add a new PR for it. 🙂

PR #2460, opened 2026-01-22
Release 1.7.0

wojtekn:
## Related issues

- Related to pfHvTO-UG-p2

## Proposed Changes

- Release 1.7.0.

## Testing Instructions

- N/A

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2561, opened 2026-02-10
Change performance profiler to use 50ms threshold

fredrikekelund:
## Related issues

- Fixes STU-1245

## Proposed Changes

Performance benchmark CI runs yield much more stable results these days 🎉 So stable in fact that most PRs report improvements or regressions in <100ms.

Take https://github.com/Automattic/studio/pull/2460#issuecomment-3784944097, for example. The underlying information is that performance didn't change at all, yet the comment still states it degraded across all areas. This PR updates the performance profiler so that any change below a 50ms threshold registers as "No change".

We still display the timing change, but we use the "⚪" legend to indicate that there was no reliable, significant change. 

## Testing Instructions

Code review should suffice

## Pre-merge Checklist

### DIFF
--- a/scripts/compare-perf/post-to-github.ts
+++ b/scripts/compare-perf/post-to-github.ts
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-/* eslint-disable no-console */
+
 import fs from 'fs';
 import https from 'https';
 import path from 'path';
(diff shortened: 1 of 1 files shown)


gcsecsey (approved):
LGTM 👍

wojtekn:
@fredrikekelund  looks good, it will reduce noise. Would it make sense to add a note about this in the bot message?

> No change

> No change (difference under 50ms)

wojtekn (approved):

fredrikekelund:
> No change (difference under 50ms)

Good call 👍

PR #2562, opened 2026-02-10
Fix build size increase by using schema validator subpath import

bcotrim:
## Related issues

- N/A

## Proposed Changes

- Move `@wp-playground/blueprints` back to devDependencies (it's only needed at build time, not runtime)
- Import the lightweight schema validator directly (`blueprint-schema-validator.js`) instead of the full `validateBlueprint` export
- Add Vite resolve aliases to bypass the package's exports restriction and allow direct file imports
- Remove `@wp-playground/*` from external patterns in `electron.vite.config.ts` (only `@php-wasm/*` needs to stay external for the CLI)

This allows the schema validator to be bundled directly into the main process (pure JS, no dependencies) while keeping the WASM-heavy packages out of the packaged app.

## Testing Instructions

- Run `npm run cli:build` to verify CLI builds
- Run `npm test -- common/lib/tests/blueprint-validation.test.ts` to verify schema validation still works (11 tests)
- Run `SKIP_DMG=1 npm run make` to verify packaged app builds
- Check packaged app size: should be ~1.9GB (down from ~3GB)
- Verify no @php-wasm WASM files in asar: `npx asar list out/Studio-darwin-arm64/Studio.app/Contents/Resources/app.asar | grep -c @php-wasm` should return 0 files

## Pre-merge Checklist

### DIFF
--- a/common/lib/blueprint-validation.ts
+++ b/common/lib/blueprint-validation.ts
@@ -1,5 +1,6 @@
 import { __, sprintf } from '@wordpress/i18n';
-import { Blueprint, validateBlueprint } from '@wp-playground/blueprints';
+import validateBlueprintSchema from '@wp-playground/blueprints/blueprint-schema-validator';
+import type { Blueprint } from '@wp-playground/blueprints';
 
 interface UnsupportedFeature {
 	type: 'step' | 'property';
@@ -131,21 +132,19 @@ export type BlueprintValidationResult = BlueprintValidationError | BlueprintVali
 export async function validateBlueprintData(
 	blueprintJson: unknown
 ): Promise< BlueprintValidationResult > {
-	const result = validateBlueprint( blueprintJson as object );
+	const isValid = validateBlueprintSchema( blueprintJson );
 
-	if ( ! result.valid ) {
-		const firstError = result.errors?.[ 0 ] as
-			| {
-					instancePath?: string;
-					message?: string;
-					params?: { additionalProperty?: string };
-			  }
-			| undefined;
-		const errorPath = firstError?.instancePath || '/';
-		const additionalProp = firstError?.params?.additionalProperty;
+	if ( ! isValid && validateBlueprintSchema.errors ) {
+		const firstError = validateBlueprintSchema.errors[ 0 ];
+		// The schema validator uses ajv v8 internally (instancePath, additionalProperty)
+		// but ships with ajv v6 types (dataPath, ErrorParameters).
+		type RuntimeError = { instancePath?: string; params?: { additionalProperty?: string } };
+		const error = firstError as unknown as RuntimeError;
+		const errorPath = error.instancePath || '/';
+		const additionalProp = error.params?.additionalProperty;
 		const errorMessage = additionalProp
 			? sprintf( __( '"%s" is not a valid Blueprint property' ), additionalProp )
-			: firstError?.message || __( 'Invalid blueprint' );
+			: firstError.message || __( 'Invalid blueprint' );
 
 		return {
 			valid: false,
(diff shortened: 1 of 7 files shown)


common/lib/blueprint-validation.ts
```diff
@@ -131,21 +132,19 @@ export type BlueprintValidationResult = BlueprintValidationError | BlueprintVali
...
-					instancePath?: string;
-					message?: string;
-					params?: { additionalProperty?: string };
-			  }
-			| undefined;
-		const errorPath = firstError?.instancePath || '/';
-		const additionalProp = firstError?.params?.additionalProperty;
+	const isValid = validateBlueprintSchema( blueprintJson );
+
+	if ( ! isValid && validateBlueprintSchema.errors ) {
```
fredrikekelund, line 137: Funky return signature. Returning a boolean and attaching an `errors` property to the function itself 🤔

common/lib/blueprint-validation.ts
```diff
@@ -131,21 +132,19 @@ export type BlueprintValidationResult = BlueprintValidationError | BlueprintVali
...
+		const firstError = validateBlueprintSchema.errors[ 0 ] as {
+			instancePath?: string;
+			message?: string;
+			params?: { additionalProperty?: string };
+		};
```
fredrikekelund:
```suggestion
		const firstError = validateBlueprintSchema.errors[ 0 ];
```
No need to use the `as` operator here

common/lib/blueprint-validation.ts
```diff
@@ -131,21 +132,19 @@ export type BlueprintValidationResult = BlueprintValidationError | BlueprintVali
...
+		const errorPath = firstError.instancePath || '/';
+		const additionalProp = firstError.params?.additionalProperty;
```
fredrikekelund:
```suggestion
		const additionalProp = firstError.params.additionalProperty;
```
Nit

vite.cli.config.ts
```diff
@@ -81,6 +81,10 @@ export default defineConfig( {
 			src: resolve( __dirname, 'src' ),
 			vendor: resolve( __dirname, 'vendor' ),
 			common: resolve( __dirname, 'common' ),
+			'@wp-playground/blueprints/blueprint-schema-validator': resolve(
+				__dirname,
+				'node_modules/@wp-playground/blueprints/blueprint-schema-validator.js'
+			),
```
fredrikekelund, line 87: Do we need this because `@wp-playground/blueprints` doesn't define this file as part of the `exports` object in `package.json`? The `@wp-playground/blueprints/blueprint-schema-validator` locator matches where I'd expect node's require algorithm to look for that file on disk, but maybe this is a CommonJS/ESM difference that I haven't internalized yet…

fredrikekelund (approved):
I haven't run the full build yet, but based on a code review, this LGTM 👍 

I left a couple of nit comments that'd be nice to address, but they're not blockers.

fredrikekelund:
I triggered the macOS dev builds on CI to verify that the change decreases the installer file size

vite.cli.config.ts
bcotrim: That is also my understanding, since it's not in the exports object we need to provide an alias so Vite can find it.

common/lib/blueprint-validation.ts
bcotrim: The previous `validateBlueprint` wrapper from the main entry point had a cleaner { valid, errors } return, but importing it pulls in the entire package

fredrikekelund:
The macOS arm64 dev build comes out at 579MB 👍

PR #2565, opened 2026-02-11
Adopt monorepo structure

fredrikekelund:
## Related issues

- Fixes STU-1288

## Proposed Changes

> [!NOTE]  
> The grunt work in this PR was done using OpenAI Codex. 🚨 I know the size of the diff looks daunting, but aside from `package-lock.json` changes, this PR is mostly just moving files around, changing import paths and tweaking config files 🚨

> [!NOTE]  
> As noted in https://github.com/Automattic/studio/pull/2565#issuecomment-3884605148, it's expected that the performance tests are failing. 

> [!WARNING]  
> A consequence of this change is that `package-lock.json` will no longer govern the version of dependencies we ship in production releases. That's because we now have a single lockfile for all npm workspaces, but we use only the package-specific `package.json` files to install the `node_modules` directories shipped in production (without a lockfile). This is probably fine, but it means we should be a bit more conservative about the semver ranges we use in `package.json`. For example, we cannot use `^3.0.46` for `@wp-playground` dependencies if the intention is to ship `3.0.46` to production. For those cases, we should save the exact version in `package.json`.

The CLI used to be more of an add-on to Studio, but as of v1.7.0, it's become integral to how Studio works. To reflect this, and to make the codebase easier to work with (for humans and AI agents), this PR adopts a proper monorepo structure. 

So, what does "proper monorepo structure" mean here?

- The CLI code is moved to `apps/cli` and the Electron app code is moved to `apps/studio`. They now live side by side.
- The `apps/cli` code and the `apps/studio` code are now isolated – there are no cross imports. 
- `./common` had been made into a proper package, using `@studio/common` as the package name. It now lives in `tools/common`.
  - We still use a Vite import alias for `@studio/common`, so it's not consumed as a true npm module. This is because we'd need to add package exports for everything we want to consume outside that package (i.e., for basically every function/type/class). This seemed tedious to me, but if it's good practice for whatever reason, we can easily revisit it later. 
- Other auxiliary tools have also been moved into `tools/`, like `tools/eslint-plugin-studio`.
- We use [npm workspaces](https://docs.npmjs.com/cli/v10/using-npm/workspaces) to make it easy and convenient to install dependencies for all packages. Run `npm install` once in the repo root, and you're good to go. 
  - This means that all packages have individual `package.json` files and npm reconciles and dedupes everything into a single `node_modules` directory and a single `package-lock.json` file. 
- The CLI and the app have special requirements for packaging (see pfHvTO-XV-p2#comment-820), namely that they each need separate `node_modules` directories that get copied into the packaged output. We handle this with `install:bundle` npm scripts in `apps/cli/package.json` and `apps/studio/package.json`. 
  - A notable caveat of this approach is that it mutates the `apps/cli/node_modules` and `apps/studio/node_modules` directories, which potentially messes up the npm workspace-powered dependency tree. So, if you run `npm run package` locally, you would typically need to rerun `npm ci` after the script finishes to reconcile the dependency tree. It's easy for other developers to be stumped by this requirement, so I addressed it by adding a `scripts/package-in-isolation.ts` script. The gist is that it copies the source files from the repo to a temporary directory, installs dependencies, runs `install:bundle` and the package/make script, and then copies the output back to the repo. 

## Testing Instructions

CI should pass, and the dev build should install fine

## Pre-merge Checklist


fredrikekelund:
The performance metrics CI job fails because it uses the same setup command on `trunk` as on this branch. There's no easy way to reconcile that, so I'd prefer to just let it fail in this PR and have it work for the next PR that runs after this one.

.gitignore
gcsecsey:
```suggestion
apps/studio/e2e/imports/*.tar.gz
apps/studio/e2e/imports/*.zip
apps/studio/e2e/imports/*.wpress
```

gcsecsey, line 106:
```suggestion
apps/cli/vendor/
```

gcsecsey (commented):
Thanks for taking on the task to reorganize to a monorepo @fredrikekelund, I think this is going to make working with the different packages easier on the long run.

Also thanks for the detailed description of the changes, these made reviewing much easier! 🙌

I could `npm i` from the project root as expected, and using `npm ls -ws` I can see the newly added packages in the workspace:

Image

When starting the app, I'm currently getting failures to start the site servers, it seems as if the cli package dependencies are not yet installed, or it's not built:

```
Node.js v22.12.0
  [Exit code] 1
    at ChildProcess.<anonymous> (/Users/gcsecsey/a8c-projects/studio/.conductor/melbourne/dist/main/index.js:62877:22)
    at ChildProcess.emit (node:events:519:28)
    at ChildProcess.emit (node:domain:489:12)
    at maybeClose (node:internal/child_process:1101:16)
    at Socket.<anonymous> (node:internal/child_process:456:11)
    at Socket.emit (node:events:519:28)
    at Socket.emit (node:domain:489:12)
    at Pipe.<anonymous> (node:net:346:12)
[2026-02-12T11:31:54.373Z][info][main] Sentry Logger [log]: Captured error event `[Base message] Failed to start site
  [stderr] node:internal/modules/cjs/loader:1252
  throw err;
  ^

Error: Cannot find module '/Users/gcsecsey/a8c-projects/studio/.conductor/melbourne/dist/cli/main.js'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1249:15)
    at Function._load (node:internal/modules/cjs/loader:1075:27)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:219:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:170:5)
    at node:internal/main/run_main_module:36:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

apps/cli/vite.config.ts
gcsecsey, line 46: I could resolve the start site issue above by changing the outDir to the root.

```suggestion
		outDir: '../../dist/cli',
```

apps/cli/vite.config.ts
```diff
@@ -1,10 +1,13 @@
...
 import { viteStaticCopy } from 'vite-plugin-static-copy';
-import { existsSync } from 'fs';
 
-const yargsLocalesPath = resolve( __dirname, 'node_modules/yargs/locales' );
-const cliNodeModulesPath = resolve( __dirname, 'cli/node_modules' );
+const yargsLocalesPath = resolve( __dirname, '../../node_modules/yargs/locales' );
+const cliNodeModulesPath = resolve( __dirname, 'node_modules' );
+const packageVersion = JSON.parse(
+	readFileSync( resolve( __dirname, 'package.json' ), 'utf-8' )
+).version;
```
gcsecsey, line 11: We didn't add a separate `version` field to the CLI package.json yet. We could either add one, or point this to the root package.json if we want to have a canonical version.


```suggestion
const packageVersion = JSON.parse(
	readFileSync( resolve( __dirname, '../../package.json' ), 'utf-8' )
).version;
```

fredrikekelund:
> When starting the app, I'm currently getting failures to start the site servers, it seems as if the cli package dependencies are not yet installed, or it's not built

Hmm, would you mind pulling the latest changes and running `npm ci` from the repo root? I'm not seeing this locally, and since the builds are also passing on CI, I'm inclined to believe that some leftover is causing the issue for you locally.

apps/cli/vite.config.ts
fredrikekelund: Good point. We used to have a separate version field, but decided to drop it to keep the `studio --version` command output in sync with the Studio UI. 

It makes sense to consolidate this and maintain a single canonical version field. My instinctive preference would be to use the root package.json for this, but if Electron Forge complains, maybe it's better to use apps/studio/package.json. I'll take a look

.gitignore
fredrikekelund: This was apparently added in #2033. @bcotrim, do you know if this is still needed? The directory doesn't exist locally for me, and I don't know when or how it'd be created.

apps/cli/vite.config.ts
fredrikekelund: I believe this shouldn't be needed. See https://github.com/Automattic/studio/pull/2565#issuecomment-3890709921

gcsecsey:
> > When starting the app, I'm currently getting failures to start the site servers, it seems as if the cli package dependencies are not yet installed, or it's not built
> 
> Hmm, would you mind pulling the latest changes and running `npm ci` from the repo root? I'm not seeing this locally, and since the builds are also passing on CI, I'm inclined to believe that some leftover is causing the issue for you locally.

Thanks for checking @fredrikekelund. After pulling the latest changes, both `npm ci` and `npm i` are working as expected, and I can start sites normally. 👍

.gitignore
bcotrim: I'm pretty sure that was generated by Playground CLI at some point. Maybe it's not the case anymore

fredrikekelund:
The macOS dev build is 601 MB. 1.7.4-beta2 is 579 MB, so we have a 22 MB increase. 

I don't know exactly what that comes down to, but it's something we can revisit after landing this PR, IMO.

.gitignore
wojtekn: I cleaned up old directories on local, installed and built the app, and I can't see it anymore. It seems as unused.

AGENTS.md
wojtekn: Unrelated to this PR, but we could update the Electron version here.

wojtekn (commented):
The changes look good, and `npm` scripts used in development work fine. 

Some notes below.

We parse the version from `package.json` in a few scripts: `confirm-tag-matches-version.mjs`, `package-appx.mjs`, and `prepare-dev-build-version.mjs`. Should we adjust it? Or if they already use the correct file, would it make sense to use an absolute path, as in `package-in-isolation.ts`, to make it clearer?

We also refer to `package.json` in some `docs/` and other scripts - it would make sense to clarify which one we refer to.

Also, I built the app successfully, but it crashed when I started it:

```

Termination Reason:  Namespace DYLD, Code 1, Library missing
Library not loaded: @rpath/Electron Framework.framework/Electron Framework
Referenced from: <4C4C4492-5555-3144-A1A8-5F0D2C2A6582> /Users/USER/Sites/*/Studio.app/Contents/MacOS/Studio
Reason: tried: '/Users/cyphelf/Sites/studio/apps/studio/out/Studio-darwin-arm64/Studio.app/Contents/Frameworks/Electron Framework.framework/Electron Framework' (no such file), '/Users/cyphelf/Sites/studio/apps/studio/out/Studio-darwin-arm64/Studio.app/Contents/Frameworks/Electron Framework.framework/Electron Framework' (no such file)
(terminated at launch; ignore backtrace)
```

fredrikekelund:
> We parse the version from package.json in a few scripts

Good catch. I've updated the relevant scripts 👍 I also took the opportunity to just use a plain `import` statement where appropriate. And again, for clarity, `apps/studio/package.json` now dictates the version, not the root `package.json`.

> I built the app successfully, but it crashed when I started it

I'll test this, but as noted in https://github.com/Automattic/studio/pull/2565#issuecomment-3890709921, the E2E tests are passing on CI, which suggests this might be an issue with your local setup. Please run `npm ci` from the project root to ensure a clean install, then package again and see if that helps. 

I should note that I've also successfully generated, installed, and run macOS dev builds from previous commits.

fredrikekelund:
@wojtekn, the dev build installs and runs fine on macOS for me.

AGENTS.md
fredrikekelund: Even if we updated it, there's a good chance it would drift again, so I went ahead and removed the version number

epeicher (approved):
Thanks @fredrikekelund for implementing these massive changes! I have tested different areas, and I have not found any issues. With regards to the scripts, I have tested:

|Platform | Test|Result|
|-------|------|------|
|Mac| `npm install`|✅|
|Mac| `npm start`|✅|
|Mac| `npm run make`|✅|
|Windows| `npm install`|✅|
|Windows| `npm start`|✅|

I approve this as it works for me, but I'm open to feedback from others.

apps/cli/lib/tests/api.test.ts
```diff
@@ -1,6 +1,6 @@
 import fs from 'fs';
+import wpcomFactory from '@studio/common/lib/wpcom-factory';
 import { createMock } from 'src/lib/test-utils';
```
epeicher: Nit: The `src` alias maps to `apps/studio/src`, so this CLI test file imports from the Studio app. This breaks the isolation goal. `createMock` should be moved to `tools/common` or the test should use its own utility.

epeicher:
> A consequence of this change is that package-lock.json will no longer govern the version of dependencies we ship in production releases. That's because we now have a single lockfile for all npm workspaces, but we use only the package-specific package.json files to install the node_modules directories shipped in production (without a lockfile). This is probably fine, but it means we should be a bit more conservative about the semver ranges we use in package.json. For example, we cannot use ^3.0.46 for @wp-playground dependencies if the intention is to ship 3.0.46 to production. For those cases, we should save the exact version in package.json.

For a follow-up, we could consider improving this, an alternative to investigate is `pnpm deploy`, which creates a standalone install of a workspace package with locked versions from the root lockfile. It's purpose-built for this exact use case. But migrating from npm to pnpm is a separate undertaking.

epeicher:
Also, I would like to note that when running `npm run make` on Windows, I'm getting the following error:
```
✗ Build failed in 31.12s
error during build:
[vite:build-html] The "fileName" or "name" properties of emitted chunks and assets must be strings that are neither absolute nor relative paths, received "../../../../../../../../username/AppData/Local/Temp/studio-package-NbjiPI/repo/apps/studio/index.html".
    at getRollupError (file:///C:/Users/username/AppData/Local/Temp/studio-package-NbjiPI/repo/node_modules/rollup/dist/es/shared/parseAst.js:401:41)
```

wojtekn:
@fredrikekelund I tried again, and it still crashes.

I did the following:

1. Pull the changes from the branch
2. Remove old directories and dependencies (`dist`, `node_modules`, `assets`, `bin`, `common`, `cli`, `e2e`, `src`, `out`)
3. Run `npm install`
4. Run `npm run make`

When I launch Studio, it immediately crashes:

CleanShot 2026-02-18 at 09 28 52@2x

It doesn't seem as flakiness as it clearly misses library:

```
Library not loaded: @rpath/Electron Framework.framework/Electron Framework
Referenced from: <4C4C4492-5555-3144-A1A8-5F0D2C2A6582> /Users/USER/Sites/*/Studio.app/Contents/MacOS/Studio
Reason: tried: '/Users/cyphelf/Sites/studio/apps/studio/out/Studio-darwin-arm64/Studio.app/Contents/Frameworks/Electron Framework.framework/Electron Framework' (no such file), '/Users/cyphelf/Sites/studio/apps/studio/out/Studio-darwin-arm64/Studio.app/Contents/Frameworks/Electron Framework.framework/Electron Framework' (no such file)
(terminated at launch; ignore backtrace)
```

I checked it more, and it is a symlink pointing to a build-time directory, which I guess, was removed after the build:

```
lrwxr-xr-x   1 cyphelf  _www  209 Feb 18 08:57 Electron Framework -> /private/var/folders/6x/h7nh9c9d5fg_jxtx7bc053240000gr/T/studio-package-IfODkM/repo/apps/studio/out/Studio-darwin-arm64/Studio.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework
```

Studio 1.7.4 installed from DMG has a symlink, too, but it seems to be relative, pointing to another dir from the package:

```
lrwxr-xr-x@  1 cyphelf  staff   35 Feb 17 12:24 Electron Framework -> Versions/Current/Electron Framework
```

fredrikekelund:
> pnpm deploy

That's super interesting! Thanks for highlighting, @epeicher. Worth considering.

apps/cli/lib/tests/api.test.ts
fredrikekelund: Great catch 👍 This made me realize we should split the Vitest config the same way we do with the Vite config. This ensures the resolve aliases are identical between Vite and Vitest, and tests cannot import files that aren't accessible to source code files. I did this in 7ddeec8.

fredrikekelund:
> I tried again, and it still crashes

I'm taking another look and trying to reproduce this. The package/make routine varies between local environments and CI now (we package in isolation locally only). It's possible that this is causing the error you encountered, @wojtekn 

> I would like to note that when running npm run make on Windows, I'm getting the following error

Looking into this, too. Again, CI passed, but I'm running `npm run make` locally to confirm.

fredrikekelund:
@epeicher, I couldn't reproduce the Windows error you described. `npm run make` works fine for me, and I can start `apps\studio\out\Studio-win32-arm64\Studio.exe` just fine after building. 

@wojtekn, I could reproduce the error you described, and it was indeed related to the `package-in-isolation.ts` logic (meaning it only happens locally, and not on CI). The fix was simple. See 1895468.

PR #2574, opened 2026-02-11
[CI] Remove workaround for Buildkite's `matrix.setup`

AliSoftware:
## Related Issues

This is a follow-up on https://github.com/Automattic/studio/pull/2530 and [AINFRA-1903](https://linear.app/a8c/issue/AINFRA-1903/run-e2e-on-windows-queue-alongside-mac)

## Proposed Changes

Buildkite have told us they fixed their bug with empty `matrix.setup` so we should not need the workaround anymore

> Hey Oliver,
>
> I'm an engineer on the product team 👋
> 
> An update from us to let you know we've indeed confirmed this was a unintentional behaviour on our end (Great sleuthing through the bk/go-pipeline source code. So we're happy to inform that we've shipped a fix to enable matrix job creation through adjustments only.
>
> Using your above provided example pipeline, this will now result in the creation of two jobs with the following params:
> 
> Job #1 { arch: x86_64, platform: mac }
> Job #2 { arch: x64, platform: windows }
> Thanks again for the report and patience as we patched things up.
> 
> Let us know how you go 🙏
> 
> Jonathan

## Testing Instructions

Verify that the build on Buildkite creates 2 jobs for `E2E Tests on {platform}-{arch}`
image
image

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -33,9 +33,6 @@ steps:
   # E2E tests run on supported platform/architecture combinations.
   # - mac-arm64: Native on Apple Silicon agents
   # - windows-x64: Native on x64 agents
-  # Skipped via matrix adjustments:
-  # - mac-x64: https://linear.app/a8c/issue/AINFRA-1904
-  # - windows-arm64: x64 Windows cannot run arm64 binaries (no emulation layer)
   - label: E2E Tests on {{matrix.platform}}-{{matrix.arch}}
     key: e2e_tests
     command: bash .buildkite/commands/run-e2e-tests.sh "{{matrix.platform}}" "{{matrix.arch}}"
@@ -50,17 +47,10 @@ steps:
       # See https://playwright.dev/docs/ci#debugging-browser-launches
       DEBUG: "pw:browser"
     matrix:
-      # More details on this setup at https://github.com/Automattic/studio/pull/2530#discussion_r2763853223
-      #
-      # TL;DR: we're bending the syntax to limit the combinations.
-      setup: { platform: [_], arch: [_] }
+      setup: { platform: [], arch: [] }
       adjustments:
-        # Combinations to run
         - with: { platform: mac, arch: arm64 }
         - with: { platform: windows, arch: x64 }
-        # Skip the dummy value used to initiate the syntax.
-        - with: { platform: _, arch: _ }
-          skip: true
     # Skip E2E tests on draft PRs to save CI resources
     # Tests will run automatically when PR is marked ready for review
     if: build.branch == 'trunk' || build.tag =~ /^v[0-9]+/ || !build.pull_request.draft


.buildkite/pipeline.yml
```diff
@@ -33,9 +33,6 @@ steps:
   # E2E tests run on supported platform/architecture combinations.
   # - mac-arm64: Native on Apple Silicon agents
   # - windows-x64: Native on x64 agents
-  # Skipped via matrix adjustments:
-  # - mac-x64: https://linear.app/a8c/issue/AINFRA-1904
-  # - windows-arm64: x64 Windows cannot run arm64 binaries (no emulation layer)
```
mokagio, line 38: What do you think of keeping this comment to explain why we have the adjustments?

mokagio (approved):

.buildkite/pipeline.yml
AliSoftware: I'm not sure I'm following your idea 🤔

The 3 lines removed from that part of the diff are the part of the comment that mentions **skipping** jobs we wanted to exclude from the matrix (back from **even before** we switched to the alternative approach that used the `[_]` and workaround—at which point we should already have removed those 3 lines that weren't true anymore after we switched to the `[_]` trick)

AliSoftware: I'm ok if you want to _add_ a **different** comment line to explain that we use `[]` in `setup` entries to specify individual combinations explicitly in `adjustments` (otherwise it would generate the cross product of all combinations), if you think such a comment would be necessary/useful.

_(Even though personally I think the YAML is already self-explanatory and not super necessary given that way of using the `matrix` syntax matches the Buildkite's doc for `matrix` now that they fixed the bug…)_ 🤷‍♂️

mokagio: What I think might be worth tracking, or rather duplicating in the YAML with a link to Linear, it the reason we are not using the full matrix [mac, win] x [arm, x86_64] but only a subset, due to the open question about rosetta and the lack of Arm Windows VMs.

Looking at the pipeline without the context of having set up the E2E, I'd want to know why the other platform/architecture pairing are not supported.

But now that I write about the Linear link, maybe once we get to that one and add the adjustment for it, we can also reintroduce the line for Windows ARM somewhere.

AliSoftware: > Looking at the pipeline without the context of having set up the E2E, I'd want to know why the other platform/architecture pairing are not supported.

Oh I see so your point was not about documenting/explaining the YAML syntax, but explaining _why_ we decided not to include the `{ platform: windows, arch: arm64 }` configuration in the set, I see! Makes more sense now 👍

> But now that I write about the Linear link, maybe once we get to that one and add the adjustment for it, we can also reintroduce the line for Windows ARM somewhere.

Yeah we can just add the comment about win-arm64 not being technically possible closer to the `adjustments` attribute and its lists of combinations to explain why we didn't include this one?

```yaml
adjustments:
 - with { platform: mac, arch: arm64 }
 # with { platform: mac, arch: x64 } # TODO: AINFRA-1904
 - with { platform: windows, arch: x64 }
 # NOTE: { platform: windows, arch: arm64 } is not possible because our windows CI agent is x64 and cannot run arm64 binaries (no emulation)
```

---

_Btw even if we ended up adding all the 2x2 combinations of the full matrix dot product (eg in a world where running win-arm64 was technically possible), I think personally I'd still prefer us keeping the explicit list of the 4 combinations (with 4 dedicated `adjustments`) rather than switch back to just `setup: { platform: [mac,windows], arch: [x86,arm64] }` just because I like the explicit listing better for that use case anyway._

mokagio: > I think personally I'd still prefer us keeping the explicit list of the 4 combinations (with 4 dedicated adjustments) rather than switch back to just setup: { platform: [mac,windows], arch: [x86,arm64] } just because I like the explicit listing better for that use case anyway.

You know what, now that you mention it I think I'd like how that would read better, too.

PR #2572, opened 2026-02-11
Revert "Bundle WordPress translation files for faster site creation"

ivan-ottinger:
## Summary

- Reverts #2558 due to intermittent `ConnectTimeoutError` failures on `downloads.wordpress.org` during CI builds
- The sequential download of ~114 translation ZIP files causes transient network timeouts that fail the entire build
- Related discussion: p1770812916266539-slack-C06DRMD6VPZ

## Testing

Running `npm install && npm start` should work correctly. There should be no regressions.

PR #2575, opened 2026-02-11
Bundle WordPress translation files with download retry and failure handling

ivan-ottinger:
## Related issues

Closes STU-1250
Related to STU-980

## Context

#2558 bundled WordPress translation files for faster site creation but was reverted in #2572 due to CI build failures — `ConnectTimeoutError` when downloading ~114 translation ZIP files sequentially from `downloads.wordpress.org` during `npm install`. The root cause was that language pack downloads ran on every `npm install` via `postinstall`.

Related discussion: p1770812916266539-slack-C06DRMD6VPZ.

## What changed from the original PR

This PR re-applies #2558 with the following fixes:

- **Language pack downloads moved out of `postinstall`** into a standalone script (`scripts/download-language-packs.ts`) that runs during `npm run make` via the forge `prePackage` hook.
- **`npm install` no longer downloads language packs**
- **Retry with exponential backoff** — failed HTTP requests are retried up to 3 times with exponential backoff (1s, 2s, 4s) to handle transient network issues like `ConnectTimeoutError`.
- **Download failure is fatal during `make`** — if a download still fails after retries, the build fails, ensuring all translations are included in the release.
- **`npm run download-language-packs`** available as a standalone command for manual use.
- **Skips translations for older bundled themes** (twentytwentythree, twentytwentyfour) to reduce download count.

## Proposed Changes (from original PR)

- Download WordPress core, plugin, and theme translation files at build time for all 19 supported non-English locales.
- At app startup, copy bundled language packs from app resources to the runtime `server-files/language-packs/` directory.
- During site creation for the latest WP version, copy locale-specific `.l10n.php` and `.json` files directly into the site's `wp-content/languages/` directory and set `WPLANG` via blueprint steps (`defineWpConfigConsts` + `setSiteOptions`) — no network round-trip needed.
- Fall back to the original `setSiteLanguage` blueprint step for non-latest WP versions or when bundled packs are unavailable.
- Auto-detect bundled plugins and themes from the extracted WordPress installation for translation downloads.
- Only bundle `.l10n.php` and `.json` files (WordPress 6.5+ format), skipping `.po` and `.mo` to reduce size (~29MB vs ~82MB).
- Make `setupWPServerFiles` resilient so individual step failures don't prevent subsequent steps from running.

### Testing the built app

1. Build the app with `npm run make`.
2. Open the content of the built app:

CleanShot 2026-02-12 at 12 01 41@2x

3. Ensure the translation files are bundled: `Studio.app/Contents/Resources/wp-files/latest/languages`:

CleanShot 2026-02-12 at 12 10 47@2x

4. Set the locale in the app settings to a non-English language (e.g. Swedish).
5. Create a new site and verify:
   - Site creation completes without downloading translations from WordPress.org. → To test this, you can go offline. Also, the site creation process should take about 3 - 4 seconds less than when testing with the current `trunk`.
   - The WP Admin is displayed in the selected language.
   - Plugin and theme translations are present.
   - You should also find translations in the site's `wp-content`.
6. Ensure you are online. Create a site with a non-latest WP version (e.g. 6.5) and verify it falls back to downloading translations. → The process should take 3 - 4 seconds longer compared to creating site with the latest WP version.
7. Again as with the previous site, everything in WP Admin should be translated, including themes and plugins.
8. Switch the locale to English.
9. Create a site and verify no translations appear in the site's `wp-content`.

### Testing the dev environment

10. Run `npm install`.
11. Head over to your project root and confirm there is no `languages` directory present in `wp-files/latest`.
12. Run `npm run download-language-packs` and verify language packs are downloaded. You can find them in the project root: `wp-files/latest/languages`.
13. Run `npm start` and set the locale in the app settings to a non-English language (e.g. Swedish).
14. You can follow similar steps as with testing the built app (steps 5 - 9).
15. Run `npm test` and verify all tests pass.

## Pre-merge Checklist

### DIFF
--- a/cli/commands/site/create.ts
+++ b/cli/commands/site/create.ts
@@ -44,6 +44,7 @@ import {
 	updateSiteAutoStart,
 	updateSiteLatestCliPid,
 } from 'cli/lib/appdata';
+import { copyLanguagePackToSite } from 'cli/lib/language-packs';
 import { connect, disconnect, emitSiteEvent } from 'cli/lib/pm2-manager';
 import { getServerFilesPath } from 'cli/lib/server-files';
 import { getPreferredSiteLanguage } from 'cli/lib/site-language';
@@ -190,10 +191,33 @@ export async function runCommand(
 
 		const setupSteps: StepDefinition[] = [];
 
-		if ( isOnlineStatus ) {
-			const siteLanguage = await getPreferredSiteLanguage( options.wpVersion );
+		const siteLanguage = await getPreferredSiteLanguage( options.wpVersion );
 
-			if ( siteLanguage && siteLanguage !== DEFAULT_LOCALE ) {
+		if ( siteLanguage && siteLanguage !== DEFAULT_LOCALE ) {
+			// For the 'latest' WP version, try using bundled language packs first to avoid
+			// a network round-trip. Fall back to the Playground setSiteLanguage step for
+			// non-latest versions or when bundled packs aren't available.
+			let usedBundledPacks = false;
+			if ( options.wpVersion === DEFAULT_WORDPRESS_VERSION ) {
+				usedBundledPacks = await copyLanguagePackToSite( sitePath, siteLanguage );
+			}
+
+			if ( usedBundledPacks ) {
+				setupSteps.push(
+					{
+						step: 'defineWpConfigConsts',
+						consts: {
+							WPLANG: siteLanguage,
+						},
+					},
+					{
+						step: 'setSiteOptions',
+						options: {
+							WPLANG: siteLanguage,
+						},
+					}
+				);
+			} else if ( isOnlineStatus ) {
 				setupSteps.push(
 					{
 						step: 'setSiteLanguage',
--- a/cli/commands/site/create.ts
+++ b/cli/commands/site/create.ts
@@ -197,12 +197,12 @@ export async function runCommand(
 			// For the 'latest' WP version, try using bundled language packs first to avoid
 			// a network round-trip. Fall back to the Playground setSiteLanguage step for
 			// non-latest versions or when bundled packs aren't available.
-			let usedBundledPacks = false;
+			let isUsingBundledLanguagePacks = false;
 			if ( options.wpVersion === DEFAULT_WORDPRESS_VERSION ) {
-				usedBundledPacks = await copyLanguagePackToSite( sitePath, siteLanguage );
+				isUsingBundledLanguagePacks = await copyLanguagePackToSite( sitePath, siteLanguage );
 			}
 
-			if ( usedBundledPacks ) {
+			if ( isUsingBundledLanguagePacks ) {
 				setupSteps.push(
 					{
 						step: 'defineWpConfigConsts',
--- a/common/lib/wp-locales.ts
+++ b/common/lib/wp-locales.ts
@@ -8,6 +8,7 @@ export const WP_LOCALES = [
 	'es_ES',
 	'fr_FR',
 	'he_IL',
+	'hu_HU',
 	'id_ID',
 	'it_IT',
 	'ja',
--- a/common/lib/wp-locales.ts
+++ b/common/lib/wp-locales.ts
@@ -1,0 +1,25 @@
+/**
+ * WordPress locale codes for all non-English locales supported by Studio.
+ * Used for downloading and installing WordPress core language packs.
+ */
+export const WP_LOCALES = [
+	'ar',
+	'de_DE',
+	'es_ES',
+	'fr_FR',
+	'he_IL',
+	'id_ID',
+	'it_IT',
+	'ja',
+	'ko_KR',
+	'nl_NL',
+	'pl_PL',
+	'pt_BR',
+	'ru_RU',
+	'sv_SE',
+	'tr_TR',
+	'vi',
+	'uk',
+	'zh_CN',
+	'zh_TW',
+];
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
 		"e2e": "npx playwright install && npx playwright test",
 		"test:metrics": "npx playwright test --config=./metrics/playwright.metrics.config.ts",
 		"make-pot": "node ./scripts/make-pot.mjs",
-		"download:language-packs": "ts-node ./scripts/download-language-packs.ts"
+		"download-language-packs": "ts-node ./scripts/download-language-packs.ts"
 	},
 	"devDependencies": {
 		"@automattic/color-studio": "^4.1.0",
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
 		"test:watch": "vitest",
 		"e2e": "npx playwright install && npx playwright test",
 		"test:metrics": "npx playwright test --config=./metrics/playwright.metrics.config.ts",
-		"make-pot": "node ./scripts/make-pot.mjs"
+		"make-pot": "node ./scripts/make-pot.mjs",
+		"download:language-packs": "ts-node ./scripts/download-language-packs.ts"
 	},
 	"devDependencies": {
 		"@automattic/color-studio": "^4.1.0",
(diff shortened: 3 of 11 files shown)


wojtekn:
@ivan-ottinger I think we need to address the request limits part, too, so we don't have flaky builds.

ivan-ottinger:
> @ivan-ottinger I think we need to address the request limits part, too, so we don't have flaky builds.

I have applied logic with three retries with delay of 1, 2 and 4 seconds. We could also add a small delay between each request on top of that. What do you think?

Or would you prefer adding that caching logic as well? I wonder if it is not too much though - in case it turns out there are no issues with the current approach anymore.

common/lib/wp-locales.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * WordPress locale codes for all non-English locales supported by Studio.
+ * Used for downloading and installing WordPress core language packs.
+ */
+export const WP_LOCALES = [
```
nightnei, line 5: Let' also add Hungarian localization - https://github.com/Automattic/studio/pull/2578

scripts/download-language-packs.ts
```diff
@@ -0,0 +1,166 @@
...
+					`[language-packs] Request failed (${ error instanceof Error ? error.message : error }), retrying in ${ delay }ms (attempt ${ attempt }/${ MAX_RETRIES })...`
+				);
+				await new Promise( ( resolve ) => setTimeout( resolve, delay ) );
+				continue;
+			}
+			throw error;
+		}
+	}
+	// This should never be reached, but TypeScript needs it.
+	throw new Error( 'Unexpected: exceeded max retries' );
```
nightnei: NIT, but maybe it would not be necessary if we use recursively call of `fetchWithRetry` instead for `for` loop. Also it's just metter of preference, I like recurive call instead of for loop :) Anyway, it's totally NIT

nightnei (approved):
@ivan-ottinger I tested and everything works well, only one step which I didn't get how to test - `The WordPress admin is displayed in the selected language.`. What is WP admin?
Also left some small comments.

But I would also wait for Wojtek's feedback on his comment, so I propose postponing this PR to the next release. WDYT, is it ok?

ivan-ottinger:
Thank you for the review, Vova!

> @ivan-ottinger I tested and everything works well, only one step which I didn't get how to test - The WordPress admin is displayed in the selected language.. What is WP admin?

Yes, WP Admin. I have updated it in the description to make it clear. 🙂 

> Also left some small comments.

Thanks! I will take a look at them.

> But I would also wait for Wojtek's feedback on his comment, so I propose postponing this PR to the next release. WDYT, is it ok?

I agree, let's leave this PR for the next release.

common/lib/wp-locales.ts
ivan-ottinger: Good catch! I have added it in https://github.com/Automattic/studio/pull/2575/changes/d1dc96494d67a8155531fe5887bf6807296c51b6.

scripts/download-language-packs.ts
ivan-ottinger: Sounds good! Change made in https://github.com/Automattic/studio/pull/2575/changes/e1bf9273e2592e4873bbb87a80c30e4d7d9ab45f. We saved two lines. 🙂

package.json
```diff
@@ -39,7 +39,8 @@
 		"test:watch": "vitest",
 		"e2e": "npx playwright install && npx playwright test",
 		"test:metrics": "npx playwright test --config=./metrics/playwright.metrics.config.ts",
-		"make-pot": "node ./scripts/make-pot.mjs"
+		"make-pot": "node ./scripts/make-pot.mjs",
+		"download:language-packs": "ts-node ./scripts/download-language-packs.ts"
```
```suggestion
		"download-language-packs": "ts-node ./scripts/download-language-packs.ts"
```
Nitpicky, but since `download` isn't really a category of scripts, I'd name the script like so

cli/commands/site/create.ts
```diff
@@ -190,10 +191,33 @@ export async function runCommand(
...
-		if ( isOnlineStatus ) {
-			const siteLanguage = await getPreferredSiteLanguage( options.wpVersion );
+		const siteLanguage = await getPreferredSiteLanguage( options.wpVersion );
 
-			if ( siteLanguage && siteLanguage !== DEFAULT_LOCALE ) {
+		if ( siteLanguage && siteLanguage !== DEFAULT_LOCALE ) {
+			// For the 'latest' WP version, try using bundled language packs first to avoid
+			// a network round-trip. Fall back to the Playground setSiteLanguage step for
+			// non-latest versions or when bundled packs aren't available.
+			let usedBundledPacks = false;
```
```suggestion
			let isUsingBundledLanguagePacks = false;
```
The exact name isn't super important, but I'd prefix the variable name with `is` for clarity.

src/setup-wp-server-files.ts
```diff
@@ -108,12 +113,38 @@ async function copyBundledTranslations() {
...
+	];
+
+	for ( const [ name, step ] of steps ) {
+		try {
+			await step();
+		} catch ( error ) {
+			console.error( `Failed to set up bundled ${ name }:`, error );
+		}
+	}
 }
```
fredrikekelund, line 148: This isn't the fault of this PR, but I really want to revisit why we need to do this. We're basically copying the entire `wp-files` directory to a single `server-files` location. Why do we need to do this? Seems like potential legacy logic that we could simplify…

fredrikekelund (approved):
This works as expected and LGTM 👍 I left a couple of nit comments, but nothing blocking. We could potentially land this PR and follow up with a separate PR for addressing any small tweaks we want to do.

package.json
ivan-ottinger: Good catch! That was a leftover from one of the previous implementations I tried before. I have made the change in https://github.com/Automattic/studio/pull/2575/commits/e260e39f3f1a53f3d4dcfc64245ea8dc0f880d9a.

cli/commands/site/create.ts
ivan-ottinger: I agree. It is more clear with your proposed name. → Made the change in https://github.com/Automattic/studio/pull/2575/commits/942a060c9830ce23eec9f571e61029656bd6ed6f.

src/setup-wp-server-files.ts
ivan-ottinger: Good observation. I let Claude look at this and here are its observations (I haven't verified everything):

```
  1. The app bundle is read-only — Electron's Resources/ directory
  (ASAR-packed) can't be modified at runtime. The app needs a writable
  location because it downloads newer versions of WordPress, SQLite, and
  WP-CLI from the internet.
  2. Auto-updates — The app checks for and downloads newer WordPress
  versions, storing them in server-files/. The bundled version is just the
  initial seed.
  3. Version management — server-files/wordpress-versions/ holds multiple WP
   versions side by side (e.g., latest/, 6.4.3/). Not possible in the static
   app bundle.
  4. Offline resilience — The bundled files guarantee the app works on first
   launch without internet.
```

wojtekn:
> I have applied logic with three retries with delay of 1, 2 and 4 seconds. We could also add a small delay between each request on top of that. What do you think?

Okay, let's try with that. It seems that the chance of getting rate-limited is lower after moving the script out from `npm install`, so maybe that would be enough.

ivan-ottinger:
> Okay, let's try with that. It seems that the chance of getting rate-limited is lower after moving the script out from npm install, so maybe that would be enough.

Thanks, Wojtek! Yes, I think it should be far less likely to hit the limit now. In case we still run into issues, I am happy to revert again and we can apply additional measures.

fredrikekelund:
I realize we still package the app at least four times for every PR push (E2E tests for macOS and Windows, and performance job builds for the PR branch and trunk). 

So, there's still a fair amount of downloading going on, even with this optimization. We previously discussed improving CI efficiency by reusing a single packaging artifact across multiple jobs. We should look into that.

ivan-ottinger:
> So, there's still a fair amount of downloading going on, even with this optimization. We previously discussed improving CI efficiency by reusing a single packaging artifact across multiple jobs. We should look into that.

That would be great. I see we have a related Linear issue here already: STU-1045.

PR #2577, opened 2026-02-11
Add site name to start site dialog error

sejas:
## Related issues

- Fixes STU-1287

## Proposed Changes

- Adds the site name to error dialog title to all 5 errors when a site fails to start, helping users identify which site failed (especially useful during "Start all" or autostart sites on launch)
- Optimizes auto start sites to run it only on mounting the provider.
- Adds comprehensive unit tests for all error branches, with a shared `setupStartServerError` helper to reduce boilerplate
- Changes startSite parameter to receive the whole site to make easier to get the site name.

### Automated
-  `npm test -- src/hooks/tests/use-site-details.test.tsx`

### Manual
- Create a site
- Add some code to make it fail at start, for example add these constants to wp-config.php
```
define( 'WP_ALLOW_MULTISITE', true );
define( 'MULTISITE', true );
```
- Start that specific site or start all the sites
- Observe that the alert appears, and confirm the site name is included in the title.

Screenshot 2026-02-11 at 20 02 28

## Pre-merge Checklist

### DIFF
--- a/src/hooks/use-site-details.tsx
+++ b/src/hooks/use-site-details.tsx
@@ -407,12 +407,10 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 			try {
 				await getIpcApi().startServer( id );
 			} catch ( error ) {
-				const siteName = sitesRef.current.find( ( site ) => site.id === id )?.name;
+				const siteName = sitesRef.current.find( ( site ) => site.id === id )?.name || __( 'site' );
 				if ( error instanceof Error && error.message.includes( 'PROXY_ERROR_PORT_IN_USE' ) ) {
 					getIpcApi().showErrorMessageBox( {
-						title: siteName
-							? sprintf( __( "Failed to initialize custom domains for '%s'" ), siteName )
-							: __( 'Studio failed to initialize custom domains' ),
+						title: sprintf( __( "Failed to initialize custom domains for '%s'" ), siteName ),
 						message: __(
 							'Studio needs to use port 80 and 443 to enable custom domains and SSL, but one of these ports are already in use by another app. Close any local development apps and restart Studio.'
 						),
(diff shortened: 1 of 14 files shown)


src/hooks/use-site-details.tsx
```diff
@@ -168,6 +169,10 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 	const { Provider } = siteDetailsContext;
 
 	const [ sites, setSites ] = useState< SiteDetails[] >( [] );
+	const sitesRef = useRef( sites );
+	useEffect( () => {
+		sitesRef.current = sites;
+	}, [ sites ] );
```
nightnei, line 175: Why do we need this trick with useRef?
Theotretically it should work well if we include `sites` into `useCallback` dependencies of `startServer` function, unless I am missing something.

nightnei (approved):
Works well and LGTM, only one tiny thing - take a look please at the comment. Anyway approved, as it's not very critical 👍

Screenshot 2026-02-12 at 11 54 26

nightnei:
@sejas I see the changes, so you can ignore my previous comment, reviewing again 🙂

sejas:
@nightnei , great feedback! Consider last changes addressing your suggestion :D.
The useRef was used to avoid an infinite loop producedby the auto start logic. In my last changes I've optimized the auto start useEffect to run only on mount, and I also optimized the startSite logic to grab the site name from the parameter. 👍

src/hooks/use-site-details.tsx
sejas: Fixed on https://github.com/Automattic/studio/pull/2577/changes/ad7d1979f9358e2b654ccbf3b9d1aff357296247
And then optimized array dependencies on https://github.com/Automattic/studio/pull/2577/commits/3f34c0e2889f985955d40ffc12502a32ec7d1c8a

nightnei (approved):
@sejas It's unreally awesome commit - having full site will definitely be usefull and simplify the codebase in future changes, the same as in this case 🔥
Looks great and works as expected 👍

PR #2557, opened 2026-02-09
Update translations for 1.7.4

nightnei:
pfHvTO-YU-p2

## Proposed Changes

Update translations for 1.7.4

## Testing Instructions
1. Run Studio
2. Change language to any non-english
3. Assert that it looks good

PR #2583, opened 2026-02-12
[Tooling] Add Buildkite release pipelines and fastlane lanes for ReleasesV2

iangmaia:
Fixes AINFRA-1806, AINFRA-1943

ReleasesV2 PR: https://github.a8c.com/Automattic/wpcom/pull/203624

## Proposed changes

Adds CI automation for Studio releases, to be triggered from [ReleasesV2](https://mc.a8c.com/releases-v2/). This is the companion PR to the wpcom ReleasesV2 configuration PR.

### Fastlane lanes added

- `code_freeze` — Creates release branch from trunk, extracts translatable strings, delegates to `new_beta_release`
- `new_beta_release` — Bumps beta version, commits, pushes, creates GitHub prerelease, triggers release build
- `finalize_release` — Bumps to final version, creates draft GitHub release, triggers release build
- `publish_release` — Publishes draft GitHub release, creates backmerge PR to trunk
- `new_hotfix_release` — Creates hotfix branch from latest release tag, bumps version
- `distribute_release_build` — Distributes build artifacts (called from release-build-and-distribute pipeline)

### Buildkite pipelines added

- `release-pipelines/code-freeze.yml`
- `release-pipelines/new-beta-release.yml`
- `release-pipelines/finalize-release.yml`
- `release-pipelines/publish-release.yml`
- `release-pipelines/download-translations.yml`
- `release-pipelines/new-hotfix-release.yml`
- `release-build-and-distribute.yml` — Build jobs extracted from the main `pipeline.yml` and triggered by Fastlane lanes to build Mac/Windows artifacts and distribute.

### How it works

ReleasesV2 triggers a Buildkite build with a `PIPELINE` env var pointing to the release pipeline YAML added in this PR. The release pipeline runs the corresponding Fastlane lane.

For lanes that produce builds (`new_beta_release`, `finalize_release`), we use `buildkite_add_trigger_step` to trigger a separate `release-build-and-distribute` build that handles Mac/Windows builds, artifact distribution and release tagging / GitHub release creation.

## Testing instructions

- [ ] Test lanes locally (e.g. `code_freeze`) locally (without `skip_confirm`) to verify prompts and flow
- [ ] Trigger a test build from ReleasesV2 on a sandbox to verify end-to-end integration; if needed, cancel Buildkite jobs once you validate them. Note that this is pretty hard to test E2E without having everything merged; I've been triggering the pipelines manually via Buildktie, simulating ReleasesV2.


fastlane/Fastfile
```diff
@@ -104,21 +112,350 @@ lane :distribute_release_build do |_options|
...
+    message: 'Update translations',
+    allow_nothing_to_commit: true
+  )
+
+  if result.nil?
+    Fastlane::Helper::GitHelper.checkout_and_pull(current_branch)
+    Fastlane::Helper::GitHelper.delete_local_branch_if_exists!(translations_branch)
+    UI.important('No translation changes detected.')
+
+    if is_ci?
```
mokagio, line 409: Should this be `is_ci`? I didn't know there was a `is_ci?` method:

Image

mokagio (commented):
@iangmaia great effort! I run out of time to look in details at this, and given the surface area and established workflow, I wonder whether it's best to just merge and run a few beta/internal-only releases to work out the kinks.

I asked Claude Code to review with three agents for correctness, maintainability, and consistency with existing approach. I'll get it to post its findings next.

mokagio:
## PR Review — ReleasesV2 for Studio

### Critical

**1. Beta builds distributed with wrong version string**
`trigger_release_build` strips the beta suffix from `RELEASE_VERSION` (e.g. `1.7.4-beta1` → `1.7.4`). The build pipeline then passes `version:"1.7.4"` to `distribute_release_build`, so CDN uploads and Slack notifications would say `v1.7.4` instead of `v1.7.4-beta1`.
**Fix**: Either pass the full version as a separate env var, or drop the explicit `version:` param from the pipeline so the lane reads `package.json` directly (its default behavior).

**2. Missing `configure_apply` in pipeline YAMLs**
Every WPiOS/WCiOS release pipeline runs `bundle exec fastlane run configure_apply` to fetch secrets before lane execution. None of Studio's 7 new pipelines include this.
**Action**: Verify whether Studio uses a different secrets mechanism. If it relies on the release-toolkit's `configure`, this will fail on CI.

**3. `is_ci?` may not exist**
The `download_translations` lane and `create_backmerge_pr` helper use `is_ci?` (with `?`). Fastlane's built-in is `is_ci` (no `?`). If there's no custom definition elsewhere in the Fastfile, this will fail at runtime.

### Warnings

**4. Missing git safety guards**
WPiOS/WCiOS lanes start with `ensure_git_status_clean` and `ensure_git_branch_is_release_branch!`. Studio's lanes have neither, making local runs fragile.

**5. `DRY_RUN` doesn't cover the new lanes**
`DRY_RUN=true` skips CDN uploads and Slack, but the new lanes still push branches, create tags, create GitHub releases, trigger Buildkite builds, and create backmerge PRs. The docs suggest `DRY_RUN=true bundle exec fastlane code_freeze ...` is safe for local testing — it isn't.

**6. `new_beta_release` relies on `nil.to_i == 0`**
When `code_freeze` calls `new_beta_release`, `package.json` has no beta suffix. The regex returns `nil`, and `nil.to_i` silently becomes `0` → `beta1`. Correct by accident, but a re-run on a finalized version would also silently produce `beta1`. Deserves at minimum a comment, ideally a guard.

**7. `checkout-release-branch.sh`: `git pull` is redundant and risky**
WPiOS/WCiOS scripts do `git fetch` + `git checkout` only — no `git pull`. The pull is redundant after a fresh checkout and could merge unexpectedly if the branch diverges. Remove it or use `--ff-only`.

**8. `extract_release_notes` regex may over-match**
The lookahead `(?=\n\d+\.\d+|\z)` matches any line starting with a version-like number, not just version headers. Anchoring on the full header pattern (`\n\d+\.\d+\.\d+\n=+`) would be more robust.

**9. `update_release_notes_draft` silently no-ops**
If `RELEASE-NOTES.txt` lacks an "Unreleased" header, `content.sub!` returns `nil` and the notes are silently dropped. Check the return value and `UI.user_error!` on failure.

**10. No `complete_code_freeze` two-step pattern**
WPiOS/WCiOS split code freeze into two steps with a human gate before the first beta. Studio combines everything, meaning the first beta ships automatically with no review pause. Confirm this is intentional.

**11. Version passed as parameter vs. read from file**
WPiOS/WCiOS lanes read the version from the project file (xcconfig). Studio passes `version:` from the env var through the pipeline. This creates a risk of mismatches between what ReleasesV2 sends and what the repo contains.

### Nits

**12. Constant naming divergence**
Studio uses `MAIN_BRANCH` and `BUILDKITE_ORG` where WPiOS uses `DEFAULT_BRANCH` and `BUILDKITE_ORGANIZATION`.

**13. Repeated version-bump-commit-push pattern**
Three lanes (`new_beta_release`, `finalize_release`, `new_hotfix_release`) repeat the same `set_package_json_version` → `git_add` → `git_commit` → `push_to_git_remote` block. A small helper would reduce future drift risk.

**14. Pipeline YAML duplication**
All 6 release pipelines share ~80% boilerplate (env, plugins, use-bot-for-git, install_gems, agents, retry policy). Acceptable given Buildkite's YAML limitations, but a comment noting the shared structure would help future maintainers.

**15. Fastfile approaching size threshold**
Growing from ~340 to ~820 lines. Well-sectioned today, but the release management helpers (~480 lines) are logically independent and could be extracted to a separate file if the Fastfile grows further.

**16. `release-process.md` local testing examples incomplete**
`new_beta_release` example omits the required `version:` param. `new_hotfix_release` and `download_translations` aren't shown at all. The "Pre-Release" section says "manually update `RELEASE-NOTES.txt`" but code freeze already auto-generates a draft — should say "review and edit."

**17. Hardcoded `update/latest-translations` branch name**
Static branch name could collide if two releases overlap with concurrent translation downloads.

**18. `BUILDKITE_PIPELINE` constant shadows the Buildkite env var of the same name**
Different namespaces (Ruby constant vs. env var), but could confuse readers.

---

*Posted by Claude Code (Opus 4.6) on behalf of @mokagio with approval.*

fastlane/Fastfile
iangmaia: I knew this was a pattern we [used](https://github.com/Automattic/pocket-casts-android/blob/d21dcd20f696baf93d8dbec2d00ff6539ace01b6/fastlane/Fastfile#L508) in other apps, but the `is_ci?` vs. `is_ci` detail never really occurred to me. Fastlane is doing [this trick](https://github.com/fastlane/fastlane/blob/737aae7ce100404630bef42bd9302228bf49c478/fastlane/lib/fastlane/runner.rb#L97) to remove the `?` from the action call, to support the more Ruby-idiomatic boolean method naming.

iangmaia:
@mokagio 

About the critical items:
> **1. Beta builds distributed with wrong version string**

This was an actual issue due to a change I did last minute last Friday; removing the `version` parameter from the pipeline should fix the issue. Updated on d4d3a3952b4d8c32fc3eb99c6e279bc6e6aa0c6c

> **2. Missing `configure_apply` in pipeline YAMLs** 

This is done in `prepare-environment.sh`

> **3. `is_ci?` may not exist** 

It does, see [this comment](https://github.com/Automattic/studio/pull/2583#discussion_r2840782230).

---

I've implemented other changes related to some of the comments:
- Consolidate version bump: 0b834711d48f3efea4a537a109a62fc5450d3696
- Regex validation improvement: bcec68ac4876ed2009d315532fa6d890aa611cfc
- RELEASE-NOTES top section validation: b008eb725aceb491c73427f1b33cebc7d99c90c3
- Git checkout change: f8732e7d4b494a3c1ab277f35837ca5559922759
- Documentation update: 13a71824254e03768bac55bb05c080291b321cdf, 1881972cbf7b1371b1b24fb2411b65e24b36e9ac

.buildkite/commands/build-for-windows.ps1
```diff
@@ -46,8 +46,6 @@ if ($BuildType -eq $BUILD_TYPE_DEV) {
     $env:IS_DEV_BUILD="true"
 } else {
     Write-Host "Preparing release build..."
-    node ./scripts/confirm-tag-matches-version.mjs
-	If ($LastExitCode -ne 0) { Exit $LastExitCode }
```
AliSoftware, line 50: I'm unclear for the rationale about removing those 2 lines, can you elaborate?

fastlane/Fastfile
```diff
@@ -104,21 +112,326 @@ lane :distribute_release_build do |_options|
...
+    '--ignore', "#{PROJECT_ROOT_FOLDER}/apps/cli/node_modules/**/*,**/*.d.ts",
+    '--base', PROJECT_ROOT_FOLDER,
+    '--dir', pot_dir,
+    '--output', pot_output
+  )
+
+  # Commit and push the .pot file so the wpcom cron can import it to GlotPress
+  git_add(path: ['./i18n/bundle-strings.pot'])
+  git_commit(
+    path: ['./i18n/bundle-strings.pot'],
```
AliSoftware: Shouldn't we use `pot_output` variable instead to avoid repeating the path name?
Or is this because using absolute paths (with `PROJECT_ROOT_DIR`) in this context would not work as expected?

fastlane/Fastfile
```diff
@@ -104,21 +112,326 @@ lane :distribute_release_build do |_options|
...
+# - Determines the next beta number automatically from package.json
+# - Bumps the version, commits, tags, and triggers a release build
+#
+# @param version [String] The base release version (e.g., '1.7.4')
+# @param skip_confirm [Boolean] Skip interactive confirmation prompts (default: false)
+#
+lane :new_beta_release do |version:, skip_confirm: false|
+  current_version = read_package_json_version
+  current_beta = current_version[/beta(\d+)$/, 1].to_i
+  new_version = "#{version}-beta#{current_beta + 1}"
```
AliSoftware, line 229: ⚠️ There is a risk of discrepancy here iinm, where `version` could contain one version number (say `12.3`) while `read_package_json_version` would return a different base version (e.g. `12.2-beta3`), leading to `12.3-beta4` in that case.

This shouldn't happen in theory, at least if the whole release process is run as expected—because `new_beta_release` should only be triggered after `code_freeze` has run for the current version, and `new_beta_release` lane should in theory always run from the `release/*` branch corresponding of the `version` value (based on `.buildkite/release-pipelines/new-beta-release.yml`). But would still be worth adding guardrails for this is case some edge case make us end up in an unexpected situation? (for example if someone accidentally merges `trunk` back into `release/12.3` in an attempt to solve a merge conflict or something, and thus accidentally reverts the version in `Package.json` to an unexpected value that doesn't match `12.3-*`)

fastlane/Fastfile
```diff
@@ -104,21 +112,326 @@ lane :distribute_release_build do |_options|
...
+      - Notify #dotcom-studio on Slack
+  PROMPT
+  next unless skip_confirm || UI.confirm('Continue?')
+
+  bump_version_commit_and_push(version: new_version)
+
+  # Tag the beta release
+  tag_name = "v#{new_version}"
+  add_git_tag(tag: tag_name)
+  push_git_tags(tag: tag_name)
```
AliSoftware: Should we really create the tag _before_ the CI job that builds the beta release has run?

This means that if the beta build fails for some reason (e.g. compilation error or whatnot), we'd have created the git tag even if the beta didn't really get built and might require additional commits to fix the compilation error.

I think it'd be better for the CI job that is triggered by `trigger_release_build` to do the actual tag (i.e. detect if it's a beta or final build—or being passed that info as a CI env var—and decide to just do a git tag for betas vs a GitHub Release for final builds)

fastlane/Fastfile
```diff
@@ -104,21 +112,326 @@ lane :distribute_release_build do |_options|
...
+  tag_name = "v#{version}"
+  create_github_release(
+    repository: GITHUB_REPO,
+    version: tag_name,
+    target: "release/#{version}",
+    release_notes_file_path: release_notes_path,
+    release_assets: [],
+    prerelease: false,
+    is_draft: true
+  )
```
AliSoftware: Same thing, it would probably be better to only create the GitHub Release draft at the end of the build being finished. So that the Draft isn't created if the final build fails for some reason (and potentially need to be fixed then re-triggered)

Besides, it would allow us to attach the binaries generated by those builds as assets to the GitHub Release if we wanted to (though probably not, as we publish them to AppsCDN and add the links to AppsCDN as part of `append_download_links_to_github_release`), and simplify the logic that adds the AppsCDN links in `append_download_links_to_github_release` if we were to create the GitHub Release draft as a dependent job _after_ all builds have been uploaded to AppsCDN, thus creating the GitHub Release + providing its description with AppsCDN links in one go, instead of creating the GitHub Release in one step, then having to find it using the API during `append_download_links_to_github_release` to modify its description after the fact

fastlane/Fastfile
```diff
@@ -329,8 +639,171 @@ def upload_file_to_apps_cdn(site_id:, product:, file_path:, platform:, arch:, bu
...
+  escaped = Regexp.escape(version)
+  match = content.match(/^#{escaped}\n=+\n(.*?)(?=^\d+\.\d+\S*\n=+|\z)/m)
+
+  if match
+    match[1].strip
+  else
+    UI.important("No release notes found for #{version} in RELEASE-NOTES.txt")
+    "Release #{version}"
+  end
+end
```
AliSoftware, line 711: Can't we use https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/extract_release_notes_for_version_action.rb for this? 🤔

fastlane/Fastfile
```diff
@@ -329,8 +639,171 @@ def upload_file_to_apps_cdn(site_id:, product:, file_path:, platform:, arch:, bu
...
+
+# Generate a draft release notes section from merged PRs and prepend it to RELEASE-NOTES.txt.
+# Uses GitHub's generate-notes API to list PRs since the last published release.
+def update_release_notes_draft(version:)
+  # Find the latest stable release tag reachable from HEAD to use as the starting point
+  previous_tag = find_previous_tag(pattern: 'v*')
```
AliSoftware: ❌   This will also include the previous betas in the search. I don't think that's what we want?

Looking at the existing `RELEASE-NOTES.txt` file it seems that new sections are only added per production release, not for each beta. And this `update_release_notes_draft` helper is also only called during `code_freeze`, so that confirms that we want the list of PRs since last release, not since the last intermediate beta.
If we want to keep this same behavior, we thus need to compare the PRs with the last final release tag, **excluding beta tags**.

I don't think our `find_previous_tag` action currently support custom `--exclude` patterns in order to avoid batching beta tags like `v1.7.4-beta3`. So either we need to make a PR in `release-toolkit` add this feature to `find_previous_tag` (which would probably be useful to make it more flexible in general anyway), or we need to start adopting a different tag naming convention (e.g. like we do with Tumblr and prefixing them with "folder names" e.g.`beta/1.7.4-beta3` vs `final/1.7.4`), so that we can write a `find_previous_tag(pattern:…)` that only matches final tags.

fastlane/Fastfile
```diff
@@ -329,8 +639,171 @@ def upload_file_to_apps_cdn(site_id:, product:, file_path:, platform:, arch:, bu
...
+
+  changelog = get_prs_between_tags(
+    repository: GITHUB_REPO,
+    tag_name: "v#{version}",
+    previous_tag: previous_tag,
+    target_commitish: 'HEAD'
+  )
+
+  # Convert "* Title by @user in https://github.com/Org/Repo/pull/123" to "* Title #123"
+  lines = changelog.lines.reject { |l| l.start_with?('##') || l.strip.empty? }
```
AliSoftware, line 727: ⚠️  Note that if the Studio dev team wants to adopt `.github/releases.yml` config file in the future, so that `get_prs_between_tags` groups PRs in different sections based on GitHub Labels (for other products use this to differentiate e.g. "Bug Fixes" vs "New Feature" PRs and separate them in the release notes), this code would have the side effect of removing those subsections.

This is not a problem right now as there's no `.github/releases.yml` file configured yet and their `RELEASE-NOTES.txt` file is listing a flat list of PRs per release. But to keep in mind if they want to consider their release notes to list changes classified by different kinds in the future.

fastlane/Fastfile
```diff
@@ -329,8 +639,171 @@ def upload_file_to_apps_cdn(site_id:, product:, file_path:, platform:, arch:, bu
...
+  notes_path = File.join(PROJECT_ROOT_FOLDER, 'RELEASE-NOTES.txt')
+  content = File.read(notes_path)
+
+  new_section = "#{version}\n=====\n#{formatted}\n"
+
+  # Insert the new version section after the "Unreleased" header
+  unless content.sub!(/^(Unreleased\n=+\n)\n*/, "\\1\n#{new_section}\n")
+    UI.user_error!("Could not find 'Unreleased' header in RELEASE-NOTES.txt. Please ensure the file has an 'Unreleased' section.")
+  end
+  File.write(notes_path, content)
```
AliSoftware, line 746: I thought we had a dedicated common action to handle those insertions in the right place in `RELEASE-NOTES.txt` markdown-formatted files, but maybe not (I didn't find it in `common/*` actions of `release-toolkit` at least). Would be nice to work on better actions in `release-toolkit` to manipulate Markdown-formatted files like those `release-notes.txt` / `CHANGELOG.md` files that most of our repos have to consolidate the implementation and logic across all of them without having to re-invent the wheel every time.

docs/release-process.md
```diff
@@ -1,19 +1,83 @@
...
+- Removes beta suffix (sets version to `<version>`)
+- Creates a **draft** GitHub release with notes from `RELEASE-NOTES.txt`
+- Triggers the final release build, which uploads to the Apps CDN, appends download links to the draft GitHub release, and notifies Slack
+
+### 5. Publish Release
+
+**Fastlane lane**: `publish_release`
+
+- Publishes the draft GitHub release
+- Creates a backmerge PR from `release/<version>` into `trunk`
```
AliSoftware, line 47: Just to validate with the team: is this split between "finalize" and "publish" intentional and useful to you?

In all our other release processes for mobile apps this makes sense because the apps need to be submitted for review and approved by Apple / Google before we can publish them. So we finalize the release to submit it to Apple/Google for review in one step, let them review and hopefully approve the build, and once we're ready we publish to end users.

But with Studio being distributed outside of the AppStores and not dependent on Apple/Google review/validation of each build, I wonder if this split is necessary.

Though to be honest I still think it'd be useful, if only for the release manager and your team to at least smoke-test the release build before publishing the GitHub Release officially.

But given the "Finalize Release" stage already uploads the final build to AppsCDN—which would make it publicly available and add it to the Sparkle RSS feed thus being proposed as a new version update to users via Sparkle even before you go through the Publish Release stage, that raises the question if this is useful, and if it is, if we wouldn't need an additional mechanism that makes builds uploaded to  AppsCDN via `Finalize Release` would upload it as private at first, and "Publish Release" would change the status from private to public once you had the opportunity to smoke-test the final build.

fastlane/Fastfile
```diff
@@ -329,8 +639,171 @@ def upload_file_to_apps_cdn(site_id:, product:, file_path:, platform:, arch:, bu
...
+  sh('npm', 'version', version, '--no-git-tag-version', '-w', 'studio-app', '--prefix', PROJECT_ROOT_FOLDER)
+end
+
+# Extract release notes for a specific version from RELEASE-NOTES.txt
+def extract_release_notes(version:)
+  notes_path = File.join(PROJECT_ROOT_FOLDER, 'RELEASE-NOTES.txt')
+  content = File.read(notes_path)
+
+  escaped = Regexp.escape(version)
+  match = content.match(/^#{escaped}\n=+\n(.*?)(?=^\d+\.\d+\S*\n=+|\z)/m)
```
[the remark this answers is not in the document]
AliSoftware: This is an assumption we're ok with, because `RELEASE-NOTES.txt` file has a fixed format.

.buildkite/commands/build-for-windows.ps1
iangmaia: Previously, the release process was based on tags. The first line was a call to a script to check whether the current tag matches the version from the `package.json`... but since we moved to a release branch based process, this doesn't make a lot of sense. The second line is just checking the exit code from the script.

fastlane/Fastfile
iangmaia: Indeed shouldn't happen but a guard is cheap. Updated on cdbc5a80fa57f6c518fe270f2d857e5107617687.

fastlane/Fastfile
iangmaia: Nice catch!

Updated on dcb9cf8304010046e4cdfefa14677b15f0bbb826 + 93bfe7f7bcaee9d30d3dc1b2350d9ff888a44980, I think this should work? -- I wanted to avoid changing the tag convention (though it's probably not a big deal). I've created AINFRA-2056 to add this improvement to release-toolkit.

fastlane/Fastfile
iangmaia: Yeah, I wasn't sure if the absolute path would work but apparently it is (Fastlane just sends the command to Git); updated on f7708f34a9068d5f04cd2187908638c34bf39d6b.

fastlane/Fastfile
AliSoftware: Yep I think it should work 👍

But thinking about it I think right now only Tumblr uses this action which is why with its tag naming convention using folder names it doesn't have this limitation… but all our other products would, just like Studio here? Eg most of our other apps use a `x.y-rc-N` version for their betas and would have the same problem (but it just happens that they don't use generated change logs based on PRs list and `get_previous_tag` yet hence why we never realized about that limitation until now.

So AINFRA-2056 would definitely be useful not just for Studio in fact!

fastlane/Fastfile
iangmaia: Good point. It crossed my mind but I preferred to start simple, but it's probably a good idea to handle potential error cases from the start (we could have tag conflicts in case of failures, for example). Updated on b694d47c1180bca4b4658cb35f42143ce76de658.
I also like the fact that the code to add tag / create release is next to each other; another thing I considered was creating GitHub pre-releases for betas, though the team mentioned we could skip it for now.

fastlane/Fastfile
iangmaia: Also updated on b694d47c1180bca4b4658cb35f42143ce76de658.

docs/release-process.md
iangmaia: cc @sejas

docs/release-process.md
```diff
@@ -1,19 +1,82 @@
...
+**ReleasesV2 milestone**: Release | **Fastlane lane**: `finalize_release`
+
+- Removes beta suffix (sets version to `<version>`)
+- Triggers the final release build, which uploads to the Apps CDN, creates a **draft** GitHub release with notes and download links, and notifies Slack
+
+### 5. Publish Release
+
+**Fastlane lane**: `publish_release`
+
+- Publishes the draft GitHub release
```
AliSoftware:
```suggestion
- Publishes the draft GitHub release (which creates the corresponding GitHub tag too)
```

fastlane/Fastfile
iangmaia: True! I've [added](https://github.com/wordpress-mobile/release-toolkit/pull/696) the `exclude` parameter to `find_previous_tag` and released a new `release-toolkit` version, updating the code here on 83505a2786e3893fca447097c93399d49269cb5d.

AliSoftware (approved):

fastlane/Fastfile
iangmaia: The formats are slightly different 😓 

- Studio's RELEASE-NOTES.txt format uses `=====` underline separators and `*` bullets:
```
1.7.4
=====
* Added etc etc ...
* Removed ...
```

The release toolkit action expects bare version numbers with `-` bullets and no separator:
```
1.7.4
- Added etc etc ...
- Removed ...
```

Related, I think what you mentioned [here](https://github.com/Automattic/studio/pull/2583#discussion_r2842247006) is spot on: we should probably consolidate these different formats into a single action.

iangmaia: Created AINFRA-2077

AliSoftware: > The release toolkit action expects bare version numbers with - bullets and no separator:

~~The release-toolkit action expects `##` in front of the version line iinm, but yeah.~~ Actually looking at the code maybe not, you're right. Strange, I could have sworn our other repos using that action used `## 1.2.3` for section titles. But turns out [they expect `----` after the line containing the version to consider it a section title](https://github.com/wordpress-mobile/release-toolkit/blob/e871c7d8f6556b91cab91d6fe44023c783ce4323/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/extract_release_notes_for_version_action.rb#L34) (while here we're using `====`). They don't care about the type of bullet point though (`-` vs `*`), as any line between the section title and next section is captured verbatim.

I guess we could consider changing the format of the `RELEASE-NOTES.txt` file while at it as I'm guessing the team is not too strict about it using one or the other? 🤷 And maybe use the occasion to rename the file to the more standard `CHANGELOG.md` name?

Not a blocker for this PR though

fastlane/Fastfile
iangmaia: Indeed. I've created AINFRA-2077 to handle it at some point.

fastlane/Fastfile
iangmaia: 🎗️ We need to also remember there's some [release notes parsing](https://github.a8c.com/Automattic/missioncontrol/blob/00a6c955f89e6fe9c55639dc1ab62b2234c92bca/mobile/includes/class-native-app.php#L559) going on in ReleasesV2 😮‍💨

AliSoftware: Oh TIL I completely forgot about that (I guess it's to generate the Release Summary P2 post, right?).
Good catch and good point 👍 

Also notice how the format expected by the parsing logic in ReleasesV2 is the same format than what `extract_release_notes_for_version`, namely expecting `----` instead of `====` for section separators.

package.json
```diff
@@ -43,7 +43,6 @@
 		"test:watch": "vitest",
 		"e2e": "npx playwright install && npx playwright test",
 		"test:metrics": "npx playwright test --config=./tools/metrics/playwright.metrics.config.ts",
-		"make-pot": "node ./scripts/make-pot.mjs",
```
sejas, line 46: Could we keep the `make-pot` script and file in case we want to upload them manually? Or will be possible to trigger it from Buildkite?

fastlane/Fastfile
```diff
@@ -106,19 +126,306 @@ lane :distribute_release_build do |_options|
...
+  current_branch = Fastlane::Helper::GitHelper.current_git_branch
+
+  UI.important <<~PROMPT
+    Downloading translations. This will:
+    - Download latest translations from GlotPress
+    - Create a PR to merge them into `#{current_branch}`
+  PROMPT
+  next unless skip_confirm || UI.confirm('Continue?')
+
+  sh('node', File.join(PROJECT_ROOT_FOLDER, 'scripts', 'download-available-site-translations.mjs'))
```
sejas: I executed the script `node scripts/download-available-site-translations.mjs` and it downloaded the WordPress core translations in the path `wp-files/latest/available-site-translations.json` which is ignored by git.

I think we need to create a script to download the WordPress translations from [WPcom Studio project](https://translate.wordpress.com/projects/studio).

sejas (commented):
@iangmaia, thanks for creating this PR and moving forward with the Releases V2 migration. The code looks good, and I’m looking forward to testing it together.

I’ve only spotted a misleading step: `download-available-site-translations.mjs` is not downloading the translations for WordPress Studio. Currently, this is a manual step, and here is an example PR https://github.com/Automattic/studio/pull/2557

docs/release-process.md
sejas: I like that idea! Uploading it as private in the Finalize Release step and then making it public in the Publish Release step is really smart. It’s also consistent with the name of each stage.

AliSoftware: [❓] @hannahtinkler: Is there already an API to change the `visibility` of a binary already uploaded to AppsCDN? I looked at the [FG Technical Documentation](<PCYsg-15tP-p2>) and didn't see such an API mentioned but maybe it exists and is just not yet documented, or maybe it's just already implemented as a regular/generic WP API call if it's just changing attributes of a post anyway like it's possible for regular posts?

By the way I'm not sure if the best approach here would be to play with:

1. The `visibility` attribute—starting with `Internal`, then change it to `External` once ready—to make the binary published and easily testable by a12s internally before making it public?
2. Or the `post_status` attribute—starting with `draft` status while a12s are smoke-testing the build, then change it to `publish`?

I'm not 100% sure of the consequence of one vs the other.

* I'm guessing that if we're using `post_status` we can probably already use the official WP API to change the status later like we'd do for any post or page, so at least the API is already there… but how will this reflect on a binary uploaded to AppsCDN, will it be downloadable/testable internally easily still, or will it risk not reflect well enough on how the download experience will look on the final product once published? (Depending on if the RM who wants to smoke-test it wants to test it via the Internal track of Sparkle vs manual install…)
* While `visibility` is probably a better one for testability (because a build that is published but with `visibility: Internal` is, by design, easier to download and test for a12s, since that's what the `Internal` track is for after all)… but since that's a custom attribute of the taxonomy, does this mean it will require a specific API to be changed (and for AppsCDN to account for that change, regenerate the Sparkle RSS XML if needed, and whatnot)?

fastlane/Fastfile
iangmaia: Ahh, I think this is probably a legacy script then and I probably got confused? I see that https://github.com/Automattic/studio/blob/trunk/docs/localization.md mentions a manual export. I'll fix it quickly!

package.json
iangmaia: What `make-pot` was doing is in this PR in `code_freeze`, but I created a follow-up PR to extract it so it can be executed separately: https://github.com/Automattic/studio/pull/2658, so you should be able to run `bundle exec fastlane generate_pot_file` -- the POT file now living in the repo.

docs/release-process.md
hannahtinkler: There's no dedicated endpoint to change a build's visibility, but it's just post meta, so you _should_ be able to use the [REST API](https://developer.wordpress.org/rest-api/reference/posts/#update-a-post)? 🤔

> does this mean it will require a specific API to be changed (and for AppsCDN to account for that change, regenerate the Sparkle RSS XML if needed, and whatnot)?

This is all dynamically generated from the post data, so should be fine 🙂

fastlane/Fastfile
iangmaia: Done, thanks for catching it! Updated on 7da895556e0e5a7646ce67ce86d7b342595a035b. Just tested it here and it worked well -- we can double check once we run a real test later today.

docs/release-process.md
iangmaia: Thanks Hannah! 👍 
I've created AINFRA-2102 to handle the visibility update as a follow-up.

package.json
sejas: Thank you!

PR #2584, opened 2026-02-12
Hide hidden files in DMG volume

sejas:
## Related issues

- Fixes [STU-1294](https://linear.app/a8c/issue/STU-1294/studio-hidden-icons-are-too-visible-when-dmg-volume-mounts)

## Proposed Changes

- Position `.background`, and `.VolumeIcon.icns` outside the visible DMG window area so they don't appear when macOS is configured to show hidden files
- Uses appdmg's `type: "position"` to move hidden files to coordinates (900, 900), well outside the 710x502 window
- If the user scrolls the window the icons might be visible.

## Testing Instructions

- Build a DMG with `nvm use && npm run make:dmg-arm64` (or x64)
- Mount the DMG with hidden files enabled (Toggle it by pressing `cmd+shift+.`)
- Verify `.background` and `.VolumeIcon.icns` are no longer visible in the DMG window
- Verify the app icon and Applications link still display correctly
- Restore Finder settings (Toggle it by pressing `cmd+shift+.`)

| Before | After |
|--------|--------|
|  Screenshot 2026-02-12 at 22 32 24 | Screenshot 2026-02-12 at 22 55 55 |

If the user has hidden files disabled, the window won’t scroll. If the user has hidden files enabled, the hidden files will appear in a corner outside the default view, and you’ll need to scroll to see them. You can toggle between these two modes by pressing `cmd+shift+.`.

https://github.com/user-attachments/assets/b2d9abc6-893d-41f6-96a6-519d50e4f29b

## Pre-merge Checklist

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -110,6 +110,10 @@ const config: ForgeConfig = {
 									}/Studio.app`,
 								},
 								{ x: 533, y: 354, type: 'link', path: '/Applications' },
+								{ x: 900, y: 900, type: 'position', path: '.background' },
+								{ x: 900, y: 900, type: 'position', path: '.DS_Store' },
+								{ x: 900, y: 900, type: 'position', path: '.Trashes' },
+								{ x: 900, y: 900, type: 'position', path: '.VolumeIcon.icns' },
 							],
 							additionalDMGOptions: {
 								window: {
--- a/scripts/make-dmg.mjs
+++ b/scripts/make-dmg.mjs
@@ -31,6 +31,10 @@ const dmgSpecs = {
 	contents: [
 		{ type: 'file', path: appPath, x: 533, y: 122 },
 		{ type: 'link', path: '/Applications', x: 533, y: 354 },
+		{ type: 'position', path: '.background', x: 900, y: 900 },
+		{ type: 'position', path: '.DS_Store', x: 900, y: 900 },
+		{ type: 'position', path: '.Trashes', x: 900, y: 900 },
+		{ type: 'position', path: '.VolumeIcon.icns', x: 900, y: 900 },
 	],
 };
 


wojtekn:
I think it may be confusing for Mac users if hidden files are not visible after they enable the display of hidden files. Let's avoid doing such hacks.

sejas:
> I think it may be confusing for Mac users if hidden files are not visible after they enable the display of hidden files. Let's avoid doing such hacks.

Other successful apps like [Figma](https://www.figma.com/downloads/) follow the same pattern. The goal is that the hidden files don't overlap with logos or text, and those icons don't distract users from dragging the correct file into Applications.

https://github.com/user-attachments/assets/51a521d2-cdcc-45e6-9c64-6334e5131295

wojtekn:
Fair point about Figma - if popular apps are using this pattern successfully, I'm happy to go with it. I still have a slight preference for not hiding files, but it's not something I feel strongly about. Let's move forward with your approach.

epeicher:
Hi @sejas, I am testing this but I still see the hidden files 
CleanShot 2026-02-13 at 10 48 38@2x

I've confirmed:
- I am using this branch
- I have deleted the `out` folder 
- I have run `npm make` 
- The `out` folder is correctly generated with my `arm64` architecture
- I checked that the `dmg` generated has the expected date

What could I be missing? 🤔

sejas:
@epeicher , thanks for testing it. Make sure to eject the previous Studio mounted device:
Screenshot 2026-02-13 at 09 54 20

Happy to huddle sync, in case is still not working.

epeicher:
After a huddle with @sejas we identified that `npm run make` generates the `dmg` with the hidden files visible while `npm run make:dmg-arm64` generates the `dmg` with the hidden files outside the visible window.

I've started the Mac build, and we will check the downloadable file that will be generated for the release.

sejas:
@epeicher , great catch, and thanks for pointing me out in the right direction. I added the same position configuration in both places that produce dmg files. So now this PR works with `npm run make:dmg-arm64` and also with `npm run make`.

Screenshot 2026-02-13 at 10 53 52
Screenshot 2026-02-13 at 10 53 37

epeicher (approved):
Thanks @sejas for solving this! I have seen that @wojtekn has approved to move forward, so I have tested it, and it works as expected for both `npm run make` and `npm run make:dmg-arm64` scripts. Changes also LGTM!

PR #2589, opened 2026-02-13
Tweak `getPrettyPath` behavior

fredrikekelund:
## Related issues

- Fixes STU-1270

## Proposed Changes

When the current working directory is the system root, `getPrettyPath` replaces the initial `/` with a dot. `/Users/fredrik/Studio/hello` becomes `.Users/fredrik/Studio/hello`. This PR fixes that by replacing only the cwd portion of the path when cwd isn't the path root. 

I also took the opportunity to change the behavior on Windows so that cwd is replaced with `.` there, too. `~` is the only Posix-exclusive construct here. 

Lastly, I added some tests for good measure. 

## Testing Instructions

Code review should honestly suffice

## Pre-merge Checklist


nightnei (approved):
Looks great 👍 Thanks for the fix

PR #2594, opened 2026-02-16
Upgrade `.nvmrc` to Node 24 and fix deprecation warnings

fredrikekelund:
## Related issues

- Fixes STU-1296
- Fixes STU-1297

## Proposed Changes

- Update `.nvmrc` to use Node.js 24
- <del>Add relevant custom domain-related dependencies to `cli/package.json` and</del> bump version compared to `package.json` (the `Node.util.isObject` error came from a previous version of `@vscode/sudo-prompt`) <ins>I realized that the CLI build bundles dependencies from the root `node_modules` dir, and `cli/node_modules` is only for what we copy into the production build verbatim.</ins>
- Patch `pm2-axon` to avoid `url.parse` deprecation warning (I also added STU-1301 for replacing pm2-axon in favor of a homegrown solution in the future)
- <del>Remove unused dependencies from `package.json`</del>
- Remove unused functions from `src/lib/certificate-manager.ts`
- Revert `scripts/download-node-binary.ts` to pre-https://github.com/Automattic/studio/pull/2537 version

## Background

In https://github.com/Automattic/studio/pull/2486, we upgraded the bundled Node.js runtime to v24. However, after some discussion prompted by https://github.com/Automattic/studio/pull/2486#issuecomment-3852290036, we took a different approach in https://github.com/Automattic/studio/pull/2537 and updated the Node.js version only in the `scripts/download-node-binary.ts` script (meaning we still run Node 22 for development, but Node 24 in production releases).

When smoke testing v1.7.4, we encountered two Node.js deprecation warnings from the CLI. This is a good example of why we should keep the local development version of Node.js (as specified by `.nvmrc`) in sync with the version bundled in the installer, because it would have allowed us to notice these deprecation warnings during development. 

> Isn't it a problem if the Node version bundled with Electron differs from what we have in `.nvmrc`?

Very rarely. The only way we can affect the Node version that Electron uses is by upgrading Electron itself. We use v39, which gives us a Node v22 environment on the "back-end" (powered by a custom V8 version). If `bin/node` is Node v24, that's totally fine, so long as there are no breaking changes to IPC communication in the [child_process module](https://nodejs.org/docs/latest-v24.x/api/child_process.html) (which there aren't, and which I believe are rare). 

## Testing Instructions

1. Run `nvm use`. Follow nvm's instructions if you don't already have the relevant Node version installed.
2. Run `npm install`
3. Run `cd cli` and `npm install` again
4. Run `cd ..`
5. Run `npm run cli:build`
6. Run `node dist/cli/main.js site create --domain node24.wp.local --path PATH_TO_SITE`
7. Ensure that the command finishes successfully and without deprecation warnings
8. Run `npm start`
9. Create a new site with a custom domain
10. Ensure that it works as expected

## Pre-merge Checklist


epeicher (commented):
Thanks @fredrikekelund for the detailed explanations! I have tested it and I am finding this warning when running `npm install`, should we update the `engines` versions in the `package.json` files? 

Image

fredrikekelund:
> I am finding this warning when running npm install, should we update the engines versions in the package.json files?

Ah, good catch 👍 I'll update

gcsecsey (commented):
Thanks for the in-depth explanation and for raising the fix @fredrikekelund! The changes LGTM and work well, the URL rewrite is working as described:  

Image

I encountered the same issue as @epeicher above when installing the dependencies.

Image

fredrikekelund:
I've updated the `engines` field and decided to drop the upper version limit.

gcsecsey (approved):
Nice, I get no warnings now. 👍

package.json
epeicher, line 65: I think this can be removed too if we're not using it

nightnei (approved):
LGTM and works as expected
And thanks for detailed info in the description 👍

Screenshot 2026-02-16 at 11 26 27
Screenshot 2026-02-16 at 11 29 07

epeicher (approved):
Thanks @fredrikekelund for fixing this! I have tested creating a site with a custom domain and creating a site using the CLI and they both work as expected and don't emit any warnings. I have also tested `npm run make` for any regression, and it works fine. Changes LGTM! :shipit:

fredrikekelund:
It looks like the performance CI job is failing when building the `trunk` version of the app. That's expected, as the lockfiles contain packages targeting Node 22. The performance job should resume in the next PR after this lands.

package.json
fredrikekelund: Good catch 👍 This actually prompted me to rethink some of my changes where I had moved dependencies from `package.json` to `cli/package.json`. Until we've landed https://github.com/Automattic/studio/pull/2565, this isn't the right approach, since `cli/node_modules` is only used to copy to the production build. Vite/Rollup takes care of bundling and looks up dependencies from the root `node_modules` dir just fine. 

In short, the solution was to move `@vscode/sudo-prompt` and `node-forge` back to the root `package.json` file.

PR #2596, opened 2026-02-16
Disable Copy and Delete in context menu when any site is being added

sejas:
## Related issues

- Fixes STU-1298

## Proposed Changes

The Settings tab disables "Copy site" and "Delete site" for **all** sites when any site is being created (`sites.some(site => site.isAddingSite)`), but the right-click context menu only checked `site.isAddingSite` for the individual site. This created an inconsistency where right-clicking Site B while Site A was being created showed Copy/Delete as enabled.

- Added `isAnySiteAdding` (computed from `sites.some(s => s.isAddingSite)`) in the renderer and pass it to the main process via `showSiteContextMenu()`
- Use `isAnySiteAdding` instead of `isAddingSite` for the "Copy site" and "Delete site" enabled conditions
- Per-site `isAddingSite` continues to be used for all other menu items (Start, Stop, Open, etc.)

## Testing Instructions

1. Start the app with `npm start`
2. Create a new site
3. While the site is being created, right-click another existing site in the left sidebar
4. Verify "Copy site…" and "Delete site…" are greyed out and disabled
5. Verify other menu items (Stop, Open site, WP admin, etc.) remain enabled
6. Once site creation completes, verify all menu items return to normal

https://github.com/user-attachments/assets/4b0fd42a-d7bc-43b6-bce4-3d2dfe16daae

## Pre-merge Checklist

### DIFF
--- a/src/components/site-menu.tsx
+++ b/src/components/site-menu.tsx
@@ -129,7 +129,8 @@ function ButtonToRun( site: SiteDetails ) {
 	);
 }
 function SiteItem( { site }: { site: SiteDetails } ) {
-	const { selectedSite, setSelectedSiteId, loadingServer, isSiteDeleting } = useSiteDetails();
+	const { sites, selectedSite, setSelectedSiteId, loadingServer, isSiteDeleting } =
+		useSiteDetails();
 	const isSelected = site === selectedSite;
 	const { isSiteImporting, isSiteExporting } = useImportExport();
 	const { isSiteIdPulling, isSiteIdPushing } = useSyncSites();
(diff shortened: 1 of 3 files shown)


ivan-ottinger (approved):
The change looks good and works as expected. 👍🏼 

https://github.com/user-attachments/assets/23636141-6d56-4dee-9141-52b2e3a50891

I have asked myself the same question as you did in the related task:

> Although I wonder if we should enable the delete and copy site menu items in site settings. I don't see any issue of adding or removing other sites.

When testing, I did not observe any issues with deleting or copying a site while another site is being created.

epeicher (approved):
Thanks @sejas for solving this! I have tested it, and it works as expected. `Copy` and `Delete` site menu items are disabled when a site is being created. Changes LGTM! :shipit: 

Image

PR #2597, opened 2026-02-17
[WIP] Reuse package artifacts between CI jobs

fredrikekelund:
## Related issues

- Fixes STU-1045

## Proposed Changes

> [!NOTE]  
> This PR uses #2565 as the base branch, to save myself some headaches in having to deal with merge conflicts later. 

Several CI jobs build and package the app. Previously, each job handled this individually rather than reusing the same packaged artifact. This PR fixes that by implementing the following changes:

- Added `.buildkite/commands/package-app-for-ci.sh` script that packages the app and saves it to the Buildkite cache.
- Added a Buildkite step to run the script, and to make E2E tests, performance comparison, and dev builds depend on it. 
- Updated `scripts/package-in-isolation.ts` to only build and package the app (i.e., not make it). 
- Consolidated `make:*` scripts in `apps/studio/package.json` into a single `make` script. The root `package.json` file now passes `--arch` and `--platform` options as needed. Reduces repetition and clarifies that we orchestrate from the root, and just package/make in `apps/studio`.
- Cache package artifacts on `trunk` dev builds using a fixed cache key. 
- In the compare-perf job, restore the `trunk` package artifacts. This should be a big win and help us reduce overall compute time. For comparison, the current compare-perf job builds and packages the `trunk` version of the app with every commit to every PR branch. 

## Testing Instructions

CI should pass. The artifacts from the `package_app_artifacts` step should be reused by the E2E tests and the compare-perf job. 

## Pre-merge Checklist


fredrikekelund:
This PR isn't in a stable state, but the core conclusion is that I didn't see any performance improvements with this approach over what we have on `trunk`.

Building a dev build on macos-x64 takes around 24 minutes here, but just 20 minutes on trunk. E2E tests are also slower. They take just under 10 minutes to run on macOS-arm64 on `trunk`, but 13.5 minutes here. 

Rather than spend a bunch more time trying to optimize this, I'll simply declare that we gave this approach a shot, but found that it didn't meet expectations.

PR #2598, opened 2026-02-17
Fix Preview site and Push size validation to exclude archived-out directories

epeicher:
## Related issues

- Related to STU-1303

## Proposed Changes

- Extract archive exclusion logic (`.git`, `node_modules`) into a shared `isExcludedFromArchive` function in `common/lib/fs-utils.ts`
- Apply exclusions in `calculateDirectorySizeForArchive` (renamed from `calculateDirectorySize`) so size validation matches what actually gets archived
- Use the shared function in `cli/lib/archive.ts` instead of inline string checks
- Add post-change verification instructions to `AGENTS.md`

## Testing Instructions

- Create a local WordPress site with a plugin that contains a large `node_modules` directory inside `wp-content/plugins/`. For example, you can clone the [Gutenberg repo](https://github.com/WordPress/gutenberg) into` wp-content/plugins` and run `nvm use && npm install` there
- Ensure the total `wp-content` size exceeds 2GB but the size without `node_modules`/`.git` is under 2GB
- Attempt to create a preview site — it should no longer be rejected by the size validation
- Connect the site to a WordPress.com site
- Push the local site, and check the size reported is the expected one (without `node_modules`)
- Check that if the site contains large files, the validation still applies as expected

## Pre-merge Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,6 +7,7 @@ WordPress Studio - Electron desktop app for managing local WordPress sites. Buil
 **Dev/Build**: `npm start` | `npm run cli:build` | `node dist/cli/main.js`
 **Test**: `npm test [-- path/to/test.test.ts]` | `npm run e2e`
 **Quality**: `npm run lint` | `npx prettier --write <files>` (format ONLY modified files)
+**IMPORTANT - Post-Change Verification**: After applying code changes, always run the linter (`npm run lint`), format modified files (`npx prettier --write <files>`), and run relevant tests (`npm test [-- path/to/test]`) before considering the work complete.
 **Package**: `npm run make` (builds installers for current platform)
 
 **IMPORTANT - Hot Reload**: Renderer auto-reloads, Main process needs restart (or `rs` in terminal). Changes to Main process IPC handlers require full restart.
--- a/common/lib/fs-utils.ts
+++ b/common/lib/fs-utils.ts
@@ -2,13 +2,20 @@ import fs, { promises as fsPromises } from 'fs';
 import path from 'path';
 import { isErrnoException } from './is-errno-exception';
 
+const ARCHIVE_EXCLUDED_DIRECTORIES = [ '.git', 'node_modules' ];
+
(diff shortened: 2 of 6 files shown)


AGENTS.md
```diff
@@ -7,6 +7,7 @@ WordPress Studio - Electron desktop app for managing local WordPress sites. Buil
 **Dev/Build**: `npm start` | `npm run cli:build` | `node dist/cli/main.js`
 **Test**: `npm test [-- path/to/test.test.ts]` | `npm run e2e`
 **Quality**: `npm run lint` | `npx prettier --write <files>` (format ONLY modified files)
+**IMPORTANT - Post-Change Verification**: After applying code changes, always run the linter (`npm run lint`), format modified files (`npx prettier --write <files>`), and run relevant tests (`npm test [-- path/to/test]`) before considering the work complete.
```
epeicher, line 10: I added this as the coding agent was including changes with linting errors

common/lib/fs-utils.ts
```diff
@@ -2,13 +2,20 @@ import fs, { promises as fsPromises } from 'fs';
...
+
 /**
  * Calculates the total size of a directory by recursively traversing its contents.
+ * Excludes directories that are not included in archives (e.g., .git, node_modules).
  *
  * @param directoryPath - The path to the directory to calculate the size of
  * @returns A promise that resolves to the total size in bytes
  */
-export function calculateDirectorySize( directoryPath: string ): Promise< number > {
+export function calculateDirectorySizeForArchive( directoryPath: string ): Promise< number > {
```
epeicher, line 18: Renaming this was not strictly required, but I think it better resembles that the directory size has some excluded folders

ivan-ottinger (approved):
Nice fix! The proposed changes look good to me and work as expected:

**Before:**

Preview site:
Image

Push:
Image

**After:**

Preview site:
Image

Push:
Image

Thank you for clear testing instructions.

epeicher:
Thanks for your review and thorough testing, including before/after screenshots @ivan-ottinger!

PR #2599, opened 2026-02-17
Testing AppX execution alias without yao-pkg

fredrikekelund:
## Related issues

- Related to https://github.com/Automattic/studio/pull/2547

## Proposed Changes

I'm not familiar with AppX, but looking at https://github.com/Automattic/studio/pull/2547, I wondered whether we've tried an approach like this. In this PR, the `Executable` parameter points to the node runtime we already bundle, and a `uap10:Parameters` contains command line parameters passed to Node (in this case, simply the path to `cli/main.js`).

## Testing Instructions

We need to build an AppX installer for the app to test this. 

## Pre-merge Checklist


gcsecsey (commented):
The approach looked good to me on the surface, but when I tried to test it, I couldn't access the `studio` command from any shell (Command Prompt or PowerShell).

After digging a bit into how Windows execution aliases work, I think this approach won't work because of a limitation of the alias mechanism. Based on [this article](https://www.tiraniddo.dev/2019/09/overview-of-windows-execution-aliases.html), the AppExecutionAlias doesn't seem to support receiving parameters. I think the `uap10:Parameters` attribute is not being applied to the execution alias.

It'd be great to be able to pass a parameter to avoid packaging a standalone exe, but I don't have any better ideas at the moment for how to do that.

fredrikekelund:
Thanks for digging, @gcsecsey 🙏 I did some more testing with this last week, too, but also couldn't get it to work. I'm actually not even sure this approach had a very big effect on filesize… I'll just close this.

PR #2608, opened 2026-02-18
Fix CLI uninstall on Windows not removing bin directory

bcotrim:
## Related issues

- Fixes STU-1299

## Proposed Changes

- Add `rm(STABLE_BIN_DIR_PATH, { recursive: true, force: true })` to `uninstallCli()` to remove the `studio.bat` proxy file and bin directory on uninstall

**Root cause:** The Windows `uninstallCli()` method only removed the bin directory from the registry PATH but never deleted the `studio.bat` file or the directory itself. This matches the macOS behavior where the symlink is removed on uninstall, ensuring the CLI is no longer callable from already-open terminal sessions.

## Testing Instructions

1. Install Studio on Windows
2. Open Settings → enable CLI → Save
3. Open a **new** terminal → run `studio` → should work
4. Open Settings → disable CLI → Save
5. Open a **new** terminal → run `studio` → should NOT be found
6. Open Settings → enable CLI again → Save
7. Open a **new** terminal → run `studio` → should work again

> Note: Registry PATH changes only take effect in new terminal sessions.

## Pre-merge Checklist


fredrikekelund:
> The Windows `uninstallCli()` method only removed the bin directory from the registry PATH but never deleted the `studio.bat` file or the directory itself. This meant the CLI remained callable even after disabling it in Studio settings.

I don't understand this logic. If the directory isn't in `%PATH%`, why would the CLI remain callable..?

bcotrim:
> I don't understand this logic. If the directory isn't in `%PATH%`, why would the CLI remain callable..?

Sorry this could have been better detailed. If you uninstall Studio CLI from the app while having an open terminal you would still be able to call `studio`
This should match the behavior on macOS where the bat file symlink is removed.

fredrikekelund:
@nightnei, the issue you reported in STU-1299 was that you were still able to run the `studio` CLI command in a _new_ terminal window after uninstalling the CLI, right? This could easily have been a local quirk, where `studio` was also available elsewhere on your `%PATH%`, but I just wanted to confirm…

> This should match the behavior on macOS where the bat file symlink is removed

Seems fair enough. I kinda think we're fine without this, too, but I see how it makes sense. The install step does two things, and the uninstall step undoes just one of them. Not entirely logical on the face of it…

bcotrim:
@fredrikekelund @nightnei I am unsure if there are any pending changes to this PR, or if we should close it.
Can you confirm, please?

apps/studio/src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -195,13 +196,15 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
...
+		const binDir = this.getStudioCliBinDir();
 		const currentPath = await this.getPathFromRegistry();
 		const newPath = currentPath
 			.split( ';' )
-			.filter( ( item ) => item.trim().toLowerCase() !== STABLE_BIN_DIR_PATH.toLowerCase() )
+			.filter( ( item ) => item.trim().toLowerCase() !== binDir.toLowerCase() )
 			.join( ';' );
 
 		await this.setPathInRegistry( newPath );
+		await rm( binDir, { recursive: true, force: true } );
```
fredrikekelund:
```suggestion

		if ( process.env.NODE_ENV === 'production' ) {
			await rm( STABLE_BIN_DIR_PATH, { recursive: true, force: true } );
		}
```

fredrikekelund (commented):
I recommend updating this PR so that it only adds the call to remove `STABLE_BIN_DIR_PATH` when uninstalling, and only doing that in production environments. 

The other change here is to generate the bin dir path using the `LOCALAPPDATA` environment variable, making it more deterministic. We've also explored that in 
https://github.com/Automattic/studio/pull/2465. 

I think it'll be easier to land this is we reduce scope and pick up the discussion about the `LOCALAPPDATA` approach in the other PR.

apps/studio/src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -202,6 +202,7 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
 			.join( ';' );
 
 		await this.setPathInRegistry( newPath );
+		await rm( STABLE_BIN_DIR_PATH, { recursive: true, force: true } );
```
fredrikekelund:
```suggestion
		if ( process.env.NODE_ENV === 'production' ) {
			await rm( STABLE_BIN_DIR_PATH, { recursive: true, force: true } );
		}
```
`STABLE_BIN_DIR_PATH` points to source-controlled files in development. That's why I think it's better to do this only in production.

fredrikekelund (approved):
👍

PR #2600, opened 2026-02-17
Add Site Editor performance benchmark across environments

gcsecsey:
## Related issues

Fixes STU-1290

## Proposed Changes

We want to measure how site editor performance varies across Studio, Playground CLI, and Playground Web. Also we'd like to measure the impact on performance when plugins are installed. Currently there’s no automated way to compare these environments side by side.

This PR iterates on the test script created for STU-1199, and adds a self-contained benchmark script at `scripts/benchmark-site-editor/` that:

- Orchestrates a 10-environment matrix (benchmark.ts) across Studio, Playground CLI, and Playground Web. Each environment has bare, multi-worker, plugins, and multi-worker+plugins variants. The script handles site creation (Studio CLI with isolated appdata + symlinked server-files, or Playground CLI server), runs N measurement rounds, calculates medians, and tears everything down.

- Measures 5 site editor metrics (measure-site-editor.ts) by launching a fresh Chromium browser per round using playwright directly, navigating through the site editor, and returning raw timings for siteEditorLoad, templatesViewLoad, templateOpen, blockAdd, and templateSave.

- Installs 10 plugins via a blueprint (plugins-blueprint.json): WooCommerce, Jetpack, Elementor, Contact Form 7, and others.

- Results are printed as a comparison table and saved as JSON to metrics/artifacts/.

## Testing Instructions

- Check out this branch and install dependencies:
  ```
  cd scripts/benchmark-site-editor && npm install
  ```
- Run the test for a single environment:
  ```
  npm run benchmark -- --only=studio --rounds=1`
  ```
- Check that the Studio site is created, the the benchmark runs against it, results are printed in a table, and the site is cleaned up
- Optionally test with plugins:
  ```
  npm run benchmark -- --only=studio,studio-plugins --rounds=1
  ```
- Check that results are also saved to `metrics/artifacts/benchmark-comparison-<timestamp>.json`

<details>
  <summary>
    As an example, this is the output of running the benchmark across Studio and Studio with plugins environments
  </summary>
<pre>
npm run benchmark -- --only=studio,studio-plugins --rounds=3

> benchmark-site-editor@0.0.1 benchmark
> tsx benchmark.ts --only=studio,studio-plugins --rounds=3

=== Site Editor Performance Benchmark ===
Platform: darwin arm64
Node: v22.12.0
CPUs: 16
Rounds: 3
Date: 2026-02-18T14:39:43.044Z
Environments: studio, studio-plugins

Setup:
  Playwright chromium ready
[...]

PR #2614, opened 2026-02-19
Site Editor Benchmark: add custom WordPress environment support

gcsecsey:
## Related issues

Related to STU-1305
Follow up to #2600

## Proposed Changes

Adds the ability to benchmark any running WordPress site, not just Studio and Playground environments. This PR:

- Adds a repeatable `--custom` flag. A running WordPress site can be added to the benchmark run via `--custom=<name>,<url>[,<user>,<password>]`. The flag is repeatable, so multiple custom sites can be compared in a single run (eg. bare vs with plugins). Username defaults to admin, password to password.
- Adds an `--install-plugins` flag. This installs the plugins in the `plugins-blueprint.json` file using the standard WordPress REST API (cookie auth + nonce). `--install-plugins` installs plugins for all custom environments, or `--install-plugins=<name1>,<name2>` targets specific ones.
- Updates the navigation to handle setup wizard redirects. Custom environments log in via the WordPress login form, then navigate to `/wp-admin/` explicitly.

## Examples

```bash
# Benchmark a single custom site
npm run benchmark -- --custom=my-site,http://localhost:10003

# Compare bare vs. with plugins
npm run benchmark -- \
  --custom=custom-bare,http://localhost:10003 \
  --custom=custom-plugins,http://localhost:10004 \
  --install-plugins=custom-plugins

# Compare Studio vs. a custom site
npm run benchmark -- --only=studio,my-site \
  --custom=my-site,http://localhost:10003 --rounds=3
```

## Testing Instructions

- Set up a WordPress site (Local, Docker, MAMP, etc.) with an admin account
- Run a bare custom benchmark:
  ```bash
  cd tools/benchmark-site-editor
  npm install
  npm run benchmark -- --skip-studio --skip-playground-cli --skip-playground-web \
    --custom=my-site,http://localhost:10003
  ```
- Check that the benchmark completes and prints results
- Run with `--install-plugins` and verify plugins are installed
- Run again to verify already-installed plugins are skipped
- Run with multiple `--custom` flags to verify multiple sites work
- Verify existing environments still work: `npm run benchmark -- --only=studio`

## Pre-merge Checklist

- [x] TypeScript compilation clean
- [x] Plugin installation works via REST API with proper activation hooks
- [x] Teardown guaranteed even on plugin install failure
- [x] Local environments excluded by default


tools/benchmark-site-editor/package.json
```diff
@@ -6,16 +6,14 @@
...
 	"type": "module",
 	"scripts": {
-		"benchmark": "ts-node ./benchmark.ts"
-	},
-	"dependencies": {
-		"@wp-playground/cli": "3.1.4",
-		"chalk": "^5.3.0",
-		"playwright": "^1.58.1",
-		"ts-node": "^10.9.2"
+		"benchmark": "tsx benchmark.ts"
```
fredrikekelund: It does seem like ts-node is unmaintained (https://github.com/TypeStrong/ts-node/issues/2140), and the recommended alternative (https://github.com/TypeStrong/ts-node/issues/2140) is indeed tsx. 

It irks me to have them both in our dependency tree, though… 

My ideal solution would be to revert this PR to using ts-node and to create a follow-up issue to use `node --experimental-strip-types` instead of either ts-node or tsx.

tools/benchmark-site-editor/tsconfig.json
fredrikekelund, line 1: This file was adapted to ts-node. If we follow my suggestion to revert this PR to use ts-node instead of tsx, we should revert this file as well.

tools/benchmark-site-editor/benchmark.ts
```diff
@@ -620,6 +691,21 @@ function saveResultsSummary( results: BenchmarkResult[] ): void {
...
+			execSync( 'npm install', {
+				cwd: path.resolve( STUDIO_ROOT, 'apps/cli' ),
+				stdio: 'inherit',
+			} );
+		} catch {
+			console.error( chalk.red( '  Failed to install CLI dependencies' ) );
+			return false;
+		}
+	}
+
```
fredrikekelund: Now that everything is connected with npm workspaces, I'd argue that this isn't really necessary. Users are required to run `npm install` to install the `tools/benchmark-site-editor` dependencies anyway.

tools/benchmark-site-editor/benchmark.ts
```diff
@@ -123,6 +159,15 @@ function parseArgs(): Options {
 			opts.skipPlaygroundCli = true;
 		} else if ( arg === '--skip-playground-web' ) {
 			opts.skipPlaygroundWeb = true;
+		} else if ( arg.startsWith( '--custom=' ) ) {
+			opts.customs.push( parseCustomFlag( arg.split( '=' ).slice( 1 ).join( '=' ) ) );
```
fredrikekelund: This is just an auxiliary tool, so I'm not sure we should really make this change, but [yargs](https://www.npmjs.com/package/yargs) is already in our dependency tree and would be a great fit here…

fredrikekelund (approved):
LGTM 👍 Thanks for adding this. I left a couple of comments – they should hopefully be pretty straightforward to address.

tools/benchmark-site-editor/tsconfig.json
gcsecsey: Also done in [0a542dd](https://github.com/Automattic/studio/pull/2614/commits/0a542ddae02803fab8429f5decd65d56f64a098c)

tools/benchmark-site-editor/package.json
gcsecsey: That's a good idea, thanks! I've added STU-1376 as a follow-up and reverted this PR to use ts-node in 0a542dd

tools/benchmark-site-editor/benchmark.ts
gcsecsey: Good catch, thanks! I raised this before the monorepo PR was merged, but it indeed makes sense to remove this now. Done in [0dad301](https://github.com/Automattic/studio/pull/2614/commits/0dad30113c9f6d5e1661ab29b044ee601a3121d6)

tools/benchmark-site-editor/benchmark.ts
gcsecsey: I agree, and it also makes maintenance easier, it helps to keep the args well structured. Done in a00b0c3

PR #2615, opened 2026-02-19
Patch fs-ext-extra-prebuilt to remove other-platform binaries on install

wojtekn:
## Related issues

- N/A - trying to fix builds broken after recent Playground update.

## Proposed Changes

- Applies [upstream fix (adamziel/fs-ext-prebuilt#6)](https://github.com/adamziel/fs-ext-prebuilt/pull/6) via custom script
- The script deletes `.node` binaries for other platforms during postinstall and bundle steps
- Fixes Performance Metrics broken after monorepo merge
- Fixes builds, mainly around `out/` paths

## Testing Instructions

- Run `npm install` and verify no errors
- On Windows: verify code signing doesn't fail on Linux/macOS `.node` files from `fs-ext-extra-prebuilt`

## Pre-merge Checklist


package.json
```diff
@@ -22,7 +22,7 @@
 	"scripts": {
 		"start": "npm -w studio-app run start",
 		"start-wayland": "npm -w studio-app run start-wayland",
-		"postinstall": "patch-package --patch-dir apps/cli/patches && patch-package --patch-dir apps/studio/patches && ts-node ./scripts/download-wp-server-files.ts && node ./scripts/download-available-site-translations.mjs",
+		"postinstall": "patch-package && patch-package --patch-dir apps/cli/patches && patch-package --patch-dir apps/studio/patches && ts-node ./scripts/download-wp-server-files.ts && node ./scripts/download-available-site-translations.mjs",
```
wojtekn: @fredrikekelund this packages is dependency for `@studio/common`, `benchmark-site-editor` and `studio-cli`. With the monorepo setup, is applying that in the prokject root's `node_modules/` a way to go?

fredrikekelund: Good question. This is probably one of the biggest gotchas with our monorepo setup now.

As part of the packaging flow, we now run the `install:bundle` npm scripts in `apps/cli` and `apps/studio`. They do this:

```
npm install [--omit=dev] --no-package-lock --no-progress --install-links --no-workspaces 
```

The goal is to produce self-contained `apps/cli/node_modules` and `apps/studio/node_modules` directories, because those are copied into the packaged output. 

This goal conflicts with how dependency management typically works with our npm workspaces setup, which is that most (not all) dependencies are hoisted to the root `node_modules` directory. 

So, how does this affect patches? Well, `patch-package` fails if it finds a patch without a matching dependency. That's why we have separate `apps/cli/patches` and `apps/studio/patches` directories, because otherwise the `install:bundle` npm scripts would fail. 

To answer your question directly: just put the patch in `apps/cli/patches`. `fs-ext-extra-prebuilt` is a dev dependency of `@studio/common`, so it won't be included in the self-contained packaged `node_modules` directories.

fredrikekelund (approved):
LGTM 👍 I've triggered a Windows dev build on Buildkite. Let's see how that fares before we land this

package.json
wojtekn: Thanks for the clarification. I moved it, and it works fine.

package.json
```diff
@@ -22,7 +22,7 @@
 	"scripts": {
 		"start": "npm -w studio-app run start",
 		"start-wayland": "npm -w studio-app run start-wayland",
-		"postinstall": "patch-package --patch-dir apps/cli/patches && patch-package --patch-dir apps/studio/patches && ts-node ./scripts/download-wp-server-files.ts && node ./scripts/download-available-site-translations.mjs",
+		"postinstall": "patch-package --patch-dir apps/cli/patches && patch-package --patch-dir apps/studio/patches && node ./scripts/remove-fs-ext-other-platform-binaries.mjs && ts-node ./scripts/download-wp-server-files.ts && node ./scripts/download-available-site-translations.mjs",
```
wojtekn, line 25: It didn't work with the patch, as the patch was applied after the package was installed. I moved the logic to custom scripts as a hotfix.

We can remove those when the library is fixed.

.buildkite/commands/build-for-windows.ps1
```diff
@@ -58,7 +58,7 @@ npm -w studio-app run "make:windows-$Architecture"
 If ($LastExitCode -ne 0) { Exit $LastExitCode }
 
 # Rename NuGet package files with generic name
-$artifactsPath = Get-Item ".\out" | Select-Object -ExpandProperty FullName
+$artifactsPath = Get-Item ".\apps\studio\out" | Select-Object -ExpandProperty FullName
```
wojtekn, line 61: We missed that during the monorepo change.

.buildkite/pipeline.yml
```diff
@@ -142,10 +142,10 @@ steps:
...
         command: powershell -File .buildkite/commands/build-for-windows.ps1 -BuildType dev -Architecture {{matrix}}
         artifact_paths:
-          - out\**\studio-setup.exe
-          - out\**\studio-update.nupkg
-          - out\**\RELEASES
-          - out\**\*.appx
+          - apps\studio\out\**\studio-setup.exe
+          - apps\studio\out\**\studio-update.nupkg
+          - apps\studio\out\**\RELEASES
+          - apps\studio\out\**\*.appx
```
wojtekn, line 148: Those are the remaining changes that follow the monorepo change.

PR #2618, opened 2026-02-19
Fix missing aria-level on heading role in site settings

sejas:
## Related issues

- Related to STU-1307

## Proposed Changes

- Improve accessibility issue found by React Doctor, where the explicit `role="heading"` was missing the required `aria-level` prop, which is an accessibility violation (`jsx-a11y/role-has-required-aria-props`)
This PR fixes the error by using the `Heading` component from `@wordpress/components`. It could also be fixed by simply removing `role="heading"`, but we preferred to go one step further and use a WordPress component instead a plain h3.

## Testing Instructions

Visual regressions:

- Navigate to a site's Settings tab
- Verify the "Site details" heading renders correctly with no visual regression

Tests:

- Run `npm test -- src/components/tests/content-tab-settings.test.tsx` — all 7 tests pass, including the one that checks `getByRole('heading', { name: 'Site details' })`

React Doctor:

- Run `npx -y react-doctor`
- Observe the error doesn't appear anymore:
```
  ✗ `heading` role is missing required aria props `aria-level`.
    Add missing aria props `aria-level` to the element with `heading` role.
```

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/components/content-tab-settings.tsx
+++ b/apps/studio/src/components/content-tab-settings.tsx
@@ -1,5 +1,10 @@
 import { decodePassword } from '@studio/common/lib/passwords';
-import { DropdownMenu, MenuGroup, Button } from '@wordpress/components';
+import {
+	DropdownMenu,
+	MenuGroup,
+	Button,
+	__experimentalHeading as Heading,
+} from '@wordpress/components';
 import { moreVertical } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
 import { PropsWithChildren } from 'react';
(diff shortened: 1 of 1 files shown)


fredrikekelund (changes requested):
> It could also be fixed by simply removing `role="heading"`, which was unnecessary.

This. I say let's just remove `role="heading"` and use the semantic `h3` tag. It already conveys the same meaning as `role="heading" aria-level="3"`.

katinthehatsite (approved):
For me, it works as expected 👍 I don't have preference of one approach over the other so I will approve on my end.

sejas:
> > It could also be fixed by simply removing `role="heading"`, which was unnecessary.
> 
> This. I say let's just remove `role="heading"` and use the semantic `h3` tag. It already conveys the same meaning as `role="heading" aria-level="3"`.

@fredrikekelund, Yes, that’s correct. The problem was that having only `role="heading"` required an `aria-level`. The simplest fix would be removing the `role="heading"` and leaving the `h3` as it was, but I went one step further and replaced the `h3` with the WordPress component `Heading`. Both solutions are equivalent, but I thought it was a good opportunity to rely more on the WordPress components. If you’d prefer I use an `h3`, I’m happy to modify the PR.

fredrikekelund (approved):
Looking at how `__experimentalHeading` behaves internally, I see that it renders a plain `<h3>` tag with this config (i.e., without the a11y props). I also confirmed this in the app. 

That behavior makes sense to me 👍

PR #2627, opened 2026-02-20
Fix performance tests to support pre-monorepo commits

wojtekn:
## Related issues

- Fixes performance test failures when comparing against pre-monorepo commits

## Proposed Changes

- Adds condition that detects monorepo vs old structure
- Old structure: runs `npm ci && npm run package`
- Monorepo structure: runs `npm ci && npm -w studio-cli run install:bundle && npm -w studio-app run install:bundle && npm run package`

## Testing Instructions

- Run performance comparison between a monorepo commit (e.g., `44012f24`) and a pre-monorepo commit (e.g., `58c52bfe`)
- Verify both builds succeed

## Pre-merge Checklist


fredrikekelund (commented):
I preferred the cdc1365acec0fc311abd95560eb5ed19af30371c approach, except I would have moved the script into `scripts/compare-perf`. It's more difficult to read and follow when it's all on a single line like this

tools/compare-perf/config.ts
```diff
@@ -8,7 +8,7 @@ const config = {
 	setupTestRunner: 'npm ci && npx playwright install chromium',
 	testCommand: 'npm run test:metrics',
 	setupCommand:
-		'npm ci && npm -w studio-cli run install:bundle && npm -w studio-app run install:bundle && npm run package',
+		'if [ -d "apps/studio" ]; then npm ci && npm -w studio-cli run install:bundle && npm -w studio-app run install:bundle && npm run package; else npm ci && npm run package; fi',
```
fredrikekelund:
```suggestion
		'npm ci && npm run package',
```
I realized that `npm -w studio-cli run install:bundle` and `npm -w studio-app run install:bundle` aren't needed here, because that's already handled by the packaging script. If we remove those, then we no longer need to check if this is a monorepo or not, either 😅

fredrikekelund (approved):
LGTM! Thanks for fixing this, @wojtekn 👍

tools/compare-perf/config.ts
wojtekn: Good catch! I pushed this change, and it works. I'm giving it another test locally to see if it works vs older baseline commit and will merge soon.

PR #2629, opened 2026-02-20
Fix performance test to support old structure for both build and test runner paths

wojtekn:
## Related issues

- Completes the performance test backward compatibility fixes

## Proposed Changes

- Makes both `outDir` and `testRunnerOutDir` conditional to support monorepo vs old structure
- `outDir` checks if `apps/studio/out` exists, falls back to `out`
- `testRunnerOutDir` checks if `apps/studio` directory exists to determine structure
- This ensures performance tests work when comparing commits across the monorepo migration boundary

## Testing Instructions

- Run performance comparison between a monorepo commit and a pre-monorepo commit:
  ```bash
  npm run compare:perf perf 44012f244faaac682d4473ee6f176ddaa8bcf34c 58c52bfee7e585614ced202f43f217a01f94f029 --tests-branch trunk --rounds 1
  ```
- Verify both builds succeed and tests run

## Pre-merge Checklist

### DIFF
--- a/tools/compare-perf/performance.ts
+++ b/tools/compare-perf/performance.ts
@@ -86,11 +86,21 @@ async function runTestSuite(
 	runKey: string,
 	branchDir: string
 ) {
-	const outDir = path.join( branchDir, 'apps', 'studio', 'out' );
+	// Support both monorepo (apps/studio/out) and old structure (out)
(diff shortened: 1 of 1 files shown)


fredrikekelund (approved):
I haven't tested this, but from a logical standpoint, this seems good to me 👍

PR #2641, opened 2026-02-23
Exclude @php-wasm packages from app.asar bundle

wojtekn:
## Related issues

- N/A

## Proposed Changes

- Remove unnecessary @wp-playground/wordpress dependency.

`@php-wasm` packages (~502MB of WASM files, 7+ PHP version variants) are only needed by the CLI — never by the Electron renderer or main process. The CLI bundles its own copy in the `cli/` extra resources directory.

This single change brings `app.asar` back to its pre-regression size (~305MB instead of ~801MB).

## Testing Instructions

- Build the app: `npm run make` (or `npm run package`)
- Check `app.asar` size — should be ~305MB (back to pre-regression level, down from ~801MB)
- Confirm the app still works: CLI site creation, PHP execution still function correctly since the CLI uses its own bundled `@php-wasm` in the `cli/` resource directory

## Pre-merge Checklist

### DIFF
--- a/apps/cli/vite.config.ts
+++ b/apps/cli/vite.config.ts
@@ -33,9 +33,10 @@ export default defineConfig( {
 						],
 					} ),
 					{
-						// Remove asyncify PHP-WASM builds from dist. JSPI is a newer and faster technology, and
-						// there's no need for us to bundle both build formats. Removing asyncify saves ~250MB.
-						name: 'prune-php-wasm-asyncify',
+						// Remove unnecessary PHP-WASM binaries from dist: asyncify binaries for node and all web
+						// binaries. JSPI is a newer and faster than asyncify, and there's no need for us to bundle
+						// both build formats. Removing asyncify saves ~250MB.
+						name: 'prune-php-wasm',
 						apply: 'build' as const,
 						closeBundle() {
 							const asyncifyPaths = globSync( '@php-wasm/node-*/asyncify/', {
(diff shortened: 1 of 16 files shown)


fredrikekelund:
This may resolve the issue, but I'm currently also investigating and considering our options. Please hold off on merging this for a little while…

fredrikekelund:
I investigated this by listing the contents of `Studio.app/Contents/Resources/app.asar` using the [asar module](https://www.npmjs.com/package/@electron/asar). Turns out the issue was that `tools/common` incorrectly depended on `@wp-playground/wordpress`. That module isn't even used in `@studio/common`, so its inclusion was just a mistake. 

When I removed it, all PHP-WASM dependencies were excluded from the Electron package output, and the installer filesize was reduced to 1.75 GB.

apps/cli/vite.config.ts
```diff
@@ -33,18 +33,24 @@ export default defineConfig( {
 						],
 					} ),
 					{
-						// Remove asyncify PHP-WASM builds from dist. JSPI is a newer and faster technology, and
-						// there's no need for us to bundle both build formats. Removing asyncify saves ~250MB.
-						name: 'prune-php-wasm-asyncify',
+						// Remove unnecessary PHP-WASM binaries from dist: asyncify binaries for node and all web
+						// binaries. JSPI is a newer and faster than asyncify, and there's no need for us to bundle
+						// both build formats. Removing asyncify saves ~250MB.
```
wojtekn: We can update the size part to account for the new removal.

fredrikekelund: See 3a104b0 in #2644

wojtekn:
Nice catch @fredrikekelund , thanks for improving that.

fredrikekelund (approved):
Approving my own change, I guess, but since the E2E tests are passing, we should be safe to merge 👍

PR #2649, opened 2026-02-23
Add open debug log file button in Settings

bcotrim:
## Related issues

- Fixes STU-1311

## Proposed Changes

- Add "Open log file" button next to the debug log toggle in Settings, visible only when debug logging is enabled
- Uses `showItemInFolder` to reveal the `wp-content/debug.log` file in the system file manager

image

## Testing Instructions

- Enable debug logging for a site (Settings > Debug log > Enabled)
- Verify an "Open log file" link appears next to "Enabled"
- Click the link — it should open Finder/Explorer with `debug.log` selected
- Disable debug logging and verify the link disappears

## Pre-merge Checklist


ivan-ottinger (commented):
This is a nice shortcut! It works well when the `debug.log` file exists:

https://github.com/user-attachments/assets/b312d851-d25c-4981-93cd-4041a533e82c

But if it is missing, the link is not clickable:

https://github.com/user-attachments/assets/7c0e1715-ea64-4dd0-9ce1-3e736308095d

Maybe we could consider disabling it in that case.

wojtekn:
@bcotrim, it works well for me on Mac.

I was wondering what would be the best UX - e.g., if we should add "Debug log path" field below,  with `wp-content/debug.log` value, and add a link in the form of an icon. However, the current approach is consistent with the "Trust Certificate" CTA, so it looks good.

ivan-ottinger (approved):
Thank you for the updates, Bernardo.

I can confirm the link is not displayed anymore when the `debug.log` does not exist. 👍🏼 

There's still a slight edge case: When the user is already on the Settings tab, the link won't appear/disappear if the `debug.log` is added/removed from the file system, but I think that is acceptable.

bcotrim:
> There's still a slight edge case: When the user is already on the Settings tab, the link won't appear/disappear if the `debug.log` is added/removed from the file system, but I think that is acceptable.

I considered it, but it would require a file watcher. To me it seems the added complexity doesn't pay considering it should be an edge case.

bcotrim:
@wojtekn I agree that looks better! Please let me know if this is good or if you prefer the previous version.

image

wojtekn:
> @wojtekn I agree that looks better! Please let me know if this is good or if you prefer the previous version.

Thanks for trying that. On second thought, the file name will be the same in most cases, and your first approach was consistent with certificate nudge. What do you think?

bcotrim:
> Thanks for trying that. On second thought, the file name will be the same in most cases, and your first approach was consistent with certificate nudge. What do you think?

I think the file location being always the same is a good point, it doesn't make much sense to have it here.
Going to revert it back to the initial version and deploy, thanks for the feedback!

PR #2651, opened 2026-02-24
feat: add site icons to sidebar with fallback colors

shaunandrews:
## Summary
- Fetches and caches custom site icons (favicons) from running WordPress sites using `get_site_icon_url()`
- Sites without a custom icon get a distinguishable muted color from an 8-color palette with a centered WP logo
- Color is assigned round-robin at creation time and persisted to `appdata-v1.json`
- Existing sites without `iconColorIndex` get a deterministic color derived from hashing their site ID (no migration needed)
- Widened sidebar from 210px to 230px

image

## Test plan
- [ ] Create a new site — verify it gets a colored fallback icon immediately (no color flash)
- [ ] Create multiple sites — verify colors cycle through the palette
- [ ] Delete a site and create another — color may repeat but stays stable
- [ ] Copy a site — verify the copy gets its own distinct color
- [ ] Set a custom site icon in WP admin — verify the real icon appears after refresh
- [ ] Restart the app — verify colors persist across restarts
- [ ] Verify existing sites (created before this change) show stable hash-derived colors

### DIFF
--- a/apps/studio/src/components/app.tsx
+++ b/apps/studio/src/components/app.tsx
@@ -74,7 +74,7 @@ export default function App() {
 						<MainSidebar
 							className={ cx(
 								'h-full transition-all duration-500',
-								isSidebarVisible ? 'basis-52 flex-shrink-0' : 'basis-0 !min-w-[10px]'
+								isSidebarVisible ? 'basis-[248px] flex-shrink-0' : 'basis-0 !min-w-[10px]'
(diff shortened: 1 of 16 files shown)


wojtekn:
@shaunandrews I'm closing this, as those icons are included in the https://github.com/Automattic/studio/pull/3717 (and they look even better there).

PR #2656, opened 2026-02-24
Add "Open Application Logs" to Help menu

ivan-ottinger:
## Related issues

- N/A (no existing Linear issue)

## Proposed Changes

I think it could be useful to make it easy for our users to access app logs in case they need to troubleshoot something - or if we need to ask them for the logs.

At the moment, it is possible to easily get to logs from error modals only (after some error is triggered already). What we could to is to add a menu item as follows. What do you think?

CleanShot 2026-02-24 at 17 08 24@2x

- Add an "Open Application Logs" menu item to the Help menu, placed before "Report an Issue"
- Clicking it opens the Studio log file (`current.log`) in the user's default text editor via `shell.openPath()`
- Uses the existing `getLogsFilePath()` utility from `src/logging`, following the same pattern as the error dialog's "Open Studio Logs" button

## Testing Instructions

1. Run `npm start`
2. Open the **Help** menu
3. Verify "Open Application Logs" appears between the Windows speed-up item (or the separator on macOS) and "Report an Issue"
4. Click "Open Application Logs"
5. Verify the Studio log file opens in your default text editor

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/menu.ts
+++ b/apps/studio/src/menu.ts
@@ -5,6 +5,7 @@ import {
 	BrowserWindow,
 	autoUpdater,
 	MenuItem,
+	shell,
 } from 'electron';
 import { __ } from '@wordpress/i18n';
 import { openAboutWindow } from 'src/about-menu/open-about-menu';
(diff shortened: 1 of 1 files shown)


sejas (approved):
It works as expected. I tested it on a Mac, and it opens the log file in the Console app. Another alternative could be opening Finder with `showItemInFolder`, but I think the current behavior of opening the log is more consistent with the current dialogs.

https://github.com/user-attachments/assets/ac930d7e-356e-46bb-8a32-8ccaef9a27ce

katinthehatsite (approved):
I think this looks good, and also works as expected 👍

PR #2657, opened 2026-02-24
Make site create CLI command interactive

bcotrim:
## Related issues

- Fixes STU-1229

## Proposed Changes

- Added interactive prompts to the `site create` CLI command for site name, path, WordPress version, PHP version, custom domain, and HTTPS
- CLI detects interactive mode via `process.stdin.isTTY` (automatically disabled when forked by the desktop app)
- Moved site name generation and WordPress version fetching to `tools/common` so both CLI and desktop app can reuse them
- Shared code uses callback pattern for `isNameAvailable`, allowing each consumer to provide their own availability checks
- WordPress versions are fetched from the WordPress.org API with fallback to "Latest (recommended)" if offline
- User can press Ctrl+C to cancel prompts gracefully

## Testing Instructions

2. Test interactive mode (should prompt): `node dist/cli/main.js site create`
3. Test non-interactive mode (no prompts): `echo | node dist/cli/main.js site create`
4. Test with flags (skip corresponding prompts): `node dist/cli/main.js site create --name "Test" --path ~/Studio/test`
5. Run tests: `npm test -- apps/cli/commands/site/tests/create-interactive.test.ts apps/cli/lib/tests/`
6. Verify lint: `npm run lint`

## Pre-merge Checklist


apps/cli/commands/site/create.ts
```diff
@@ -502,13 +506,98 @@ export const registerCommand = ( yargs: StudioArgv ) => {
...
+								{
+									name: sprintf( __( '%s (recommended)' ), __( 'Latest' ) ),
+									value: 'latest',
+								},
+							];
+						}
+						wpVersion = await select( {
+							message: __( 'WordPress version:' ),
+							choices: wpChoices,
+							default: DEFAULT_WORDPRESS_VERSION,
```
fredrikekelund, line 555:
```suggestion
							default: DEFAULT_WORDPRESS_VERSION,
							loop: false,
```
I found the looping functionality to not be a good fit here.

It's unlikely that I'd want WP 6.2. With the looping functionality, if I move a few steps down and then back up, WP 6.2 is the closest choice to "latest".

apps/cli/commands/site/create.ts
```diff
@@ -502,13 +506,98 @@ export const registerCommand = ( yargs: StudioArgv ) => {
...
+							} ) ),
+							default: DEFAULT_PHP_VERSION,
+						} );
+					}
+
+					if ( ! customDomain ) {
+						customDomain =
+							( await input( {
+								message: __( 'Custom domain (leave empty to skip):' ),
+							} ) ) || undefined;
```
fredrikekelund:
```suggestion
						const appdata = await readAppdata();
						const existingDomains = appdata.sites
							.map( ( site ) => site.customDomain )
							.filter( ( domain ): domain is string => Boolean( domain ) );

						customDomain = await input( {
							message: __( 'Custom domain (leave empty to skip):' ),
							validate: ( value ) =>
								getDomainNameValidationError( !! value, value, existingDomains ) || true,
						} );
```
I don't understand the need for `|| undefined` here. Also, let's validate 👍

apps/cli/commands/site/create.ts
```diff
@@ -502,13 +506,98 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 				} );
 		},
 		handler: async ( argv ) => {
+			let siteName = argv.name as string | undefined;
+			let sitePath = argv.path;
+			let wpVersion = argv.wp as string | undefined;
+			let phpVersion = argv.php as string | undefined;
+			let customDomain = argv.domain as string | undefined;
+			let enableHttps = !! argv.https;
```
fredrikekelund:
```suggestion
			let siteName = argv.name;
			let sitePath = argv.path;
			let wpVersion = argv.wp;
			let phpVersion = argv.php;
			let customDomain = argv.domain;
			let enableHttps = !! argv.https;
```
The types are already well-defined. `phpVersion`, especially, has higher fidelity.

apps/cli/commands/site/tests/create-interactive.test.ts
fredrikekelund, line 1: I'm not convinced we get the kind of assurances we're looking for in tests here… There's just too much mocking going on. TS already provides strong assurances about which values `@inquirer/prompts` can return, and the runtime logic in `apps/cli/commands/site/create.ts` is already tested elsewhere.

I suggest removing this file.

apps/cli/lib/is-interactive.ts
fredrikekelund, line 1: I don't believe this needs to be a standalone function. Let's just inline the `process.stdin.isTTY` check in `apps/cli/commands/site/create.ts`. We also don't need to wrap it in `Boolean()`

apps/cli/lib/tests/is-interactive.test.ts
fredrikekelund, line 1: Let's remove this.

tools/common/lib/wordpress-versions.ts
```diff
@@ -0,0 +1,130 @@
...
+type ProcessedOffer = { version: string; shortName: string; isDevelopment: boolean };
+
+function processOffers(
+	offers: unknown[],
+	isDevelopment: boolean,
+	shortNameCounts: Map< string, number >
+): ProcessedOffer[] {
+	return offers
+		.map( ( offer ) => offerSchema.safeParse( offer ) )
+		.filter( ( r ): r is { success: true; data: z.infer< typeof offerSchema > } => r.success )
```
fredrikekelund:
```suggestion
		.filter( ( r ) => r.success )
```
I know this code is from prior to this PR, but we don't need this type guard.

fredrikekelund (commented):
This is working great overall 👍 Thanks for tackling it, @bcotrim. 

I left a couple of comments that we should look into before merging, though. 

I'm also curious about the 20% `siteStartup` performance improvement 🤔 Is that a fluke, or did this actually speed up site startup? Any idea how?

tools/common/lib/tests/generate-site-name.test.ts
fredrikekelund, line 1: Could we move this to `tools/common/lib/tests`? It would reduce the amount of mocking required. Good opportunity to apply some clean unit testing logic.

fredrikekelund (approved):
LGTM 👍 I left one final comment. Not blocking, but would be nice to address. 

The E2E tests are failing. Not sure the cause, but we need to fix before landing regardless.

bcotrim:
Thanks for the reviews @fredrikekelund 

> I'm also curious about the 20% siteStartup performance improvement 🤔 Is that a fluke, or did this actually speed up site startup? Any idea how?

To be honest, I don't look at those stats very often as I find them unreliable. The 20% improvement on `siteStartup` is now a `14% decrease` in `siteCreation`. I would call it unrelated at this point, I don't see a reason why this PR would impact `siteStartup` or `siteCreation`

wojtekn:
> To be honest, I don't look at those stats very often as I find them unreliable. 

You can look at them more, as they are much more reliable after migrating to Buildkite. There are still cases in which they vary by 10-20% in some cases, but when recently we had a PR that improved performance heavily, it was consistently visible in metrics comments.

Feel free to rerun if you think the result was off.

PR #2659, opened 2026-02-24
Add React Doctor CI Buildkite job

sejas:
## Related issues

- Fixes STU-1307

## Proposed Changes

- Add a new `React Doctor` Buildkite CI step that runs in parallel with lint, unit tests, e2e, and metrics
- New script `.buildkite/commands/run-react-doctor.sh` that:
  - Runs `npx -y react-doctor --no-ami --yes --fail-on error` to detect React code health issues
  - Parses the score from the output and fails CI if it drops below 95/100. I chose **87 as an arbitrary threshold**. We can set it to any number we consider appropriate. Currently, the score is 90.
  - Posts a Buildkite annotation (error/success styled) with full diagnostics in a collapsible block
  - Skips on non-code-only changes (docs, config, etc.) via `should-skip-job.sh`
- Reports as a GitHub commit status check ("React Doctor")
- Add `apps/studio/react-doctor.config.json` to exclude Electron entry points from being omitted because they are not dead code, and omits the `jsx-a11y/no-autofocus` rule.
- Remove unused files and export that was not caught before to bump the score to 90.
- The job takes around 1 minute 34 seconds.
Screenshot 2026-02-25 at 10 36 41

## Testing Instructions

- Confirm the CI is green.
- Check the Buildkite annotation shows the score and diagnostics
- Verify the GitHub action shows the error when score < 95 or error-level issues are found. See https://buildkite.com/automattic/studio/builds/11920#019c9163-6deb-42e1-8dcd-537314da3875

Screenshot 2026-02-24 at 20 58 16

Screenshot 2026-02-24 at 20 46 35

## Pre-merge Checklist


.buildkite/commands/run-react-doctor.sh
```diff
@@ -0,0 +1,74 @@
...
+
+echo '--- :react: React Doctor'
+
+SCORE_THRESHOLD=90
+DOCTOR_EXIT=0
+
+# Run react-doctor and capture output
+# Don't use --offline so score gets calculated
+# Use --fail-on error to catch error-level issues
+OUTPUT=$(npx -y react-doctor --no-ami --yes --fail-on error 2>&1) || DOCTOR_EXIT=$?
```
katinthehatsite, line 20: I think this always runs the latest version, right? I am wondering if we should use specific version instead? The main reason why I am thinking that is if the latest version has some breaking changes, the CI build could potentially break. But it is not a strong opinion, mostly something to consider in my opinion.

katinthehatsite (commented):
the changes look good to me. I am not sure if we need React doctor in CI so I will let others share their opinions. I left one small comment for consideration

.buildkite/commands/run-react-doctor.sh
sejas: Yes! it uses the latest version. I had a similar concern where the CI fails because new rules are added or the score algorithm is updated, but in that case we could change the threshold if necessary. Running the latest version allows us to benefit from new best practices and bugfixes.

gcsecsey (approved):
This looks good to me, and I think this is a nice-to-have addition to our CI suite. The checks it runs are valuable on top of eg. eslint rules that are already applied locally. Also, as it's run in parallel with other jobs and usually completes in ~1.5m, I don't think it adds much overhead or potentially becomes a bottleneck on any CI run.

epeicher (approved):
Thanks @sejas for adding this! I think adding quality gates to the CI is good as long as they don't add noise, and preventing dead code and similar warnings are a good example. We can always iterate on it if we need to refine it. The current approach of targeting around 90 (87 is a good number 😄 ) looks good to me!

PR #2661, opened 2026-02-25
Lint and format with a single command in `AGENTS.md`

fredrikekelund:
## Related issues

N/A

## Proposed Changes

The current `AGENTS.md` instructions make agents run a full lint operation after each change. It also uses `prettier` to format the code instead of using `eslint --fix` (which incorporates Prettier, but also auto-fixes other issues).

This PR updates the instructions to ask agents to only run `npx eslint --fix <files>` after each change. This speeds up the process significantly. 

## Testing Instructions

Code review suffices

## Pre-merge Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,10 +4,10 @@ WordPress Studio - Electron desktop app for managing local WordPress sites. Buil
 
 ## Essential Commands
 
-**Dev/Build**: `npm start` | `npm run cli:build` | `node dist/cli/main.js`
+**Dev/Build**: `npm start` | `npm run cli:build` | `node apps/cli/dist/cli/main.js`
 **Test**: `npm test [-- path/to/test.test.ts]` | `npm run e2e`
-**Quality**: `npm run lint` | `npx prettier --write <files>` (format ONLY modified files)
-**IMPORTANT - Post-Change Verification**: After applying code changes, always run the linter (`npm run lint`), format modified files (`npx prettier --write <files>`), and run relevant tests (`npm test [-- path/to/test]`) before considering the work complete.
+**Quality**: `npx eslint --fix <files>` (lint and format ONLY modified files)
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -4,10 +4,10 @@ WordPress Studio - Electron desktop app for managing local WordPress sites. Buil
 
 ## Essential Commands
 
-**Dev/Build**: `npm start` | `npm run cli:build` | `node dist/cli/main.js`
+**Dev/Build**: `npm start` | `npm run cli:build` | `node apps/cli/dist/cli/main.js`
 **Test**: `npm test [-- path/to/test.test.ts]` | `npm run e2e`
-**Quality**: `npm run lint` | `npx prettier --write <files>` (format ONLY modified files)
-**IMPORTANT - Post-Change Verification**: After applying code changes, always run the linter (`npm run lint`), format modified files (`npx prettier --write <files>`), and run relevant tests (`npm test [-- path/to/test]`) before considering the work complete.
+**Quality**: `npx eslint --fix <files>` (lint and format ONLY modified files)
```
wojtekn, line 9: Treating the commands defined in `package.json` as recommended entry points, would it be cleaner to propose `npm run lint -- --fix` here? Or add another `npm` command that automatically fixes issues?

fredrikekelund: I considered that, but `npm run lint` already passes a glob to ESLint, which isn't overridden by also passing `-- <files>`. 

> add another npm command that automatically fixes issues

There's `npm run format`, but that has the same problem with existing globs.

wojtekn: I see. We could redefine `format` as `npx eslint --fix` and then run `npm run format -- src/foo.ts`, but as it's just for agent use, using npx eslint seems fine, too.

fredrikekelund: I think it's very convenient that engineers can run `npm run lint` or `npm run format` without specifying a path or glob. I get the argument that it would be ideal to have a single npm script for linting and formatting, respectively, to prevent drift between `package.json` and `AGENTS.md`, but I don't think it's a strong enough case that we should jump through very many hoops for is. `npx eslint --fix` is very straightforward and it respects the ESLint config just the same as `npm run format`.

wojtekn (approved):

PR #2665, opened 2026-02-25
Refactor deep link connection handling and connectSite mutation

epeicher:
## Related issues

- Solves STU-1086
- Follow up of this comment - https://github.com/Automattic/studio/pull/2132#discussion_r2580931078

## Proposed Changes

- Simplified the `useListenDeepLinkConnection` hook by removing unnecessary queries and directly using the `connectSite` function with updated parameters.
- Updated the `handleConnect` function in `ContentTabSync` to accept `remoteSiteId` instead of a full site object, streamlining the connection process.
- Enhanced the `connectSite` mutation in the Redux slice to handle remote site connections more effectively, including error handling for site not found scenarios.

#### Connect a site via the Sync tab

- Select a local site, go to the Sync tab, connect a WordPress.com site, and verify it connects successfully.

#### Deep link connection
- Trigger a deep link connection (e.g., via WordPress.com "Connect to Studio" flow). Verify the correct local site is selected, the Sync tab opens, and the site connects. One example could be: `wp-studio://sync-connect-site?studioSiteId=<local_studio_id>&remoteSiteId=<remote_blog_id>`

#### Deep link with auto-open push
- Create a Studio site or open one that is not connected to WP.com, select `Publish site`, and follow the process to create the site
- Click on Open in Studio
- Verify that the Sync modal open as expected.

 #### Add a new connected site
- Click Add Site
- Choose Connect to a WordPress.com site and select a remote site
- Fill in the site name and create the site
- Verify the new local site is created and it starts pulling from the remote site (you should see the Sync tab activate with a pull in progress)

## Pre-merge Checklist


katinthehatsite:
One this is that we might want to double-check this with https://github.com/Automattic/studio/pull/2375/commits as they are likely going to overlap 😅

katinthehatsite:
Also, perhaps this PR already solves the issue from the other PR - I will give it a test now 👀

katinthehatsite (commented):
I did not finish testing everything yet but I noticed that there is a small delay for me when I click on `Connect site` or `Connect another site`:

https://github.com/user-attachments/assets/450e1939-7a7d-42f6-939c-720d87aa5c7a

Initially, I thought that it was not connecting so I clicked multiple times before seeing it connect. In the second try, I click and waited and then it connected. I think if we have the small delay, we should probably set the `Connect` button to loading state to indicate to the user that the connection is happening.

apps/studio/src/modules/sync/index.tsx
```diff
@@ -187,7 +191,7 @@ export function ContentTabSync( { selectedSite }: { selectedSite: SiteDetails }
 			dispatch( connectedSitesActions.openModal( reduxModalMode ) );
 			setSelectedRemoteSite( selectedSiteFromList );
 		} else {
-			await handleConnect( selectedSiteFromList );
+			await handleConnect( siteId );
```
katinthehatsite: There are two more calls for `handleConnect` in this file in `onPull` and `onPush` - do we also need to adjust passing an `id` there now that it expects a number?

epeicher: Well spotted! That code might be called when using deeplinks and it needed to be updated accordingly! Done as part of f78b4fd8

epeicher:
> One this is that we might want to double-check this with https://github.com/Automattic/studio/pull/2375/commits as they are likely going to overlap 😅

Ok, this can wait for that PR to be merged, this PR is not urgent as it is a refactor

epeicher:
> I think if we have the small delay, we should probably set the Connect button to loading state to indicate to the user that the connection is happening.

Absolutely, that will improve the UX. I will compare also with trunk and check if there are differences, but anyway, adding a loading state will prevent confusion

katinthehatsite:
> Absolutely, that will improve the UX. I will compare also with trunk and check if there are differences, but anyway, adding a loading state will prevent confusion

I am not sure if this was ready for review again but wanted to to mention that for me the deeplink connection no longer worked along with sites that are available to be connected after trunk rebase:

Screenshot 2026-03-06 at 4 08 07 PM

epeicher:
Thanks for your review @katinthehatsite! I have merged with trunk and applied some changes to avoid the extra fetch when opening the Connect modal. Now the `useListenDeepLinkConnection` is simplified as requested in [the original PR](https://github.com/Automattic/studio/pull/2132#discussion_r2580931078) and the existing connect functionality keeps working as it was.

I guess this can progress and then you can rebase and apply the changes on https://github.com/Automattic/studio/pull/2375 on top of them, but I am also happy to wait for that #2375 to merge and apply these changes on top of them.

epeicher:
Closing this PR — trunk has since taken a different architectural direction for the deep link connection flow (optimistic connect with parallel site data fetch in #2375), which makes the refactor proposed in STU-1086 no longer applicable.

PR #2668, opened 2026-02-25
Fix code_freeze lane committing beta bump to wrong branch

iangmaia:
Relates to AINFRA-2108

## Summary

- Move `create_backmerge_pr` after `new_beta_release` in the `code_freeze` lane to fix the beta version bump being committed to the backmerge branch instead of the release branch

## Root cause

The `create_release_backmerge_pull_request` action (from release-toolkit) creates a `merge/release-X.Y.Z-into-trunk` intermediate branch and leaves git HEAD checked out on it. When `new_beta_release` subsequently runs `bump_version_commit_and_push`, it commits the version bump to the **backmerge branch** instead of the **release branch**.

The triggered build correctly targets `release/X.Y.Z`, but that branch never receives the beta bump commit, so it builds the wrong version.

## Fix

Move `create_backmerge_pr` to after `new_beta_release` in `code_freeze`. The backmerge action's branch switch no longer matters since it happens after the beta bump and build trigger.

## Pre-merge Checklist

- [x] Changes are tested locally

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -211,11 +211,12 @@ lane :code_freeze do |version:, skip_confirm: false, github_username: nil|
 
   push_to_git_remote(set_upstream: true)
 
-  # Create backmerge PR from the release branch to MAIN_BRANCH so that the strings bundle is uploaded to GlotPress
-  create_backmerge_pr(source_branch: branch_name, github_username: github_username)
-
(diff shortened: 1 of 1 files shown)


sejas (approved):
The code looks good to me. The build version should be beta now, instead of the previous version.

PR #2644, opened 2026-02-23
Remove PHP-WASM web builds from output

fredrikekelund:
## Related issues

- N/A

## Proposed Changes

This PR follows up https://github.com/Automattic/studio/pull/2641 by removing PHP-WASM web binaries from the packaged CLI output. Combined with that PR, the macOS application comes out at just 1,31 GB 👍

## Testing Instructions

CI should pass. 

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2670, opened 2026-02-26
Only remove `@php-wasm/\d-\d` packages

fredrikekelund:
## Related issues

- Following up on https://github.com/Automattic/studio/pull/2644

## Proposed Changes

https://github.com/Automattic/studio/pull/2644 removed all `@php-wasm/web-*` packages from the packaged `cli/node_modules` directory. This turned out to be incorrect, because `@wp-playground/blueprints` depends on `@php-wasm/web-service-worker`, which shouldn't be removed the same as `@php-wasm/web-8-2`, for example.

This PR changes the glob that looks up which modules to remove so it only targets `@php-wasm/\d-\d`.

## Testing Instructions

First and foremost, CI should pass. Second, we should generate a dev build and ensure it can start sites as intended. 

## Pre-merge Checklist

### DIFF
--- a/apps/cli/vite.config.ts
+++ b/apps/cli/vite.config.ts
@@ -44,7 +44,7 @@ export default defineConfig( {
 								absolute: true,
 							} );
 
-							const webPaths = globSync( '@php-wasm/web-*/', {
+							const webPaths = globSync( '@php-wasm/web-[0-9]-[0-9]/', {
 								cwd: distCliNodeModulesPath,
 								absolute: true,
 							} );


sejas (approved):
The new restriction looks good; we could even consider removing the `web-service-worker`.

Image

epeicher (approved):
Thanks @fredrikekelund for the prompt fix! I have packaged the app running `npm run make` and I can confirm that the sites start as expected when running `Studio.app`. Changes also LGTM! Let's wait for the CI to pass and :shipit: 

Image

fredrikekelund:
> we could even consider removing the web-service-worker

That's what caused the breakage to begin with

fredrikekelund:
I've confirmed that the dev build works as expected. Thanks for also confirming this, @epeicher 👍

The performance metric failures are due to a socket hangup, i.e., a network fluke. I'm landing this PR now

PR #2675, opened 2026-02-26
Redesign titlebar: window title, button sizing, offline indicator

shaunandrews:
## Before

image
image

## After

image
image

## Summary

- **Remove "Howdy, {user}" text** — Gravatar-only button with tooltip showing the user's name on hover. Cleaner, less cluttered.
- **Add centered window title** — "Studio • {site name}" centered in the titlebar (macOS). "Studio" is dimmed/light, bullet is subtle, site name is full white. Falls back to just "Studio" when no site is selected.
- **Improve titlebar button sizing** — All icon buttons (sidebar toggle, gravatar, settings, help) get consistent padding (`p-1.5`) and `rounded-lg` hover states. Sidebar toggle gets extra spacing from the traffic lights.
- **Reduce top/right padding** — Removed redundant `pt-1.5` from TopBar (chrome padding already handles it), reduced right padding so buttons sit closer to the window edge.
- **Refactor offline indicator** — Changed from `Button` to `span[role=status]` since it's not interactive. Removes misleading hover state while keeping tooltip and accessibility attributes. Added `fill-white` for SVG visibility on dark chrome.
- **Add settings tooltip** — Settings cog was missing a tooltip (pre-existing omission). Now shows "Settings" on hover.
- **Right-align right-side tooltips** — Gravatar, settings, and help tooltips use `bottom-end` placement so they don't overshoot the window edge.

## Test plan

- [ ] macOS: Verify window title is centered between traffic lights and right controls
- [ ] macOS: Verify title updates when selecting different sites
- [ ] macOS: Verify title shows just "Studio" with no site selected
- [ ] Verify gravatar tooltip shows user display name on hover
- [ ] Verify all titlebar buttons have visible rounded hover states
- [ ] Verify offline indicator shows tooltip but no hover background (test by toggling network off)
- [ ] Verify sidebar toggle has breathing room from traffic lights
- [ ] Verify settings cog shows "Settings" tooltip on hover
- [ ] Verify right-side tooltips align to the right (bottom-end)
- [ ] Windows: Verify no regressions in Windows titlebar layout
- [ ] Unit tests pass (`npm test`)


wojtekn:
@shaunandrews nice improvements! I merged the trunk into the branch and solved the conflict. Some notes:

- Left padding for sidebar toggle looks too large in full-screen mode (when three MacOS buttons are gone)
- Clicking on the user thumbnail could open Settings -> Account instead of General. However, I see it works that way in trunk, too, so maybe we broke it there at some point.
- We need to check how those changes look on Windows - it would be great to add a Windows screenshot, too.

katinthehatsite (commented):
Nice work, thanks for making these improvements!

I noticed two small things. The padding looks slightly too large when it is full size screen (I see it was also mentioned by another reviewer):

Image

I also noticed that when a site has a very long name, it looks a bit weird but not sure if it is a big issue. We can also leave as is or crop it shorter as an alternative:

Image

shaunandrews:
Thanks for the feedback, y'all. I finally got a Windows build and... it is indeed busted on Windows. I'm working on fixes now. 

image

shaunandrews:
Been toying with this for a bit. After some more feedback, I removed the title (Studio • Site Title) as its duplicative in most scenarios, with the site title showing in the sidebar and the frame header. And on windows the Studio title is already shown, and the large window controls and menu element cause overlapping/truncation issues.

image

image

image

image

wojtekn:
I tested it again, and it looks good. However, this label is confusing:

CleanShot 2026-03-27 at 10 17 40

I understand that the intention is to explain why it's worth logging in, but with such short sentence, it is unclear. If we want to keep it there, it would make sense to make it more descriptive e.g. "Log in to benefit from Sync, Previews, and  Assistant features." or similar.

wojtekn:
> Clicking on the user thumbnail could open Settings -> Account instead of General. However, I see it works that way in trunk, too, so maybe we broke it there at some point.

I fixed this in https://github.com/Automattic/studio/pull/2930, it was a regression introduced at some point.

shaunandrews:
> I understand that the intention is to explain why it's worth logging in, but with such short sentence, it is unclear. If we want to keep it there, it would make sense to make it more descriptive e.g. "Log in to benefit from Sync, Previews, and Assistant features." or similar.

Tooltips don't have to be sentences, and should rarely be more than a few words. I think the button's label ("WordPress.com Log In") and the tooltip ("Enable Sync and Previews") together provide enough context to explain the connection between logging in and these gated features.

FWIW I tried a few other options:
- Log in to enable sync and previews
- Login to enable more features
- Enable more features
- Sync and preview require login

Its also possible that we don't need a tooltip here; the login button itself is already fairly unambiguous, but I thought it might be helpful to try and explain _why_ you might click to login.

wojtekn:
@shaunandrews I think this tooltip starts to make sense as soon as we prepend it with "Log in to" to read "Log in to enable Sync and Previews," as you propose.

Without "Log in to", the tooltip describes a side effect (sync and previews) rather than the primary action (logging in).

shaunandrews:
> Without "Log in to", the tooltip describes a side effect

With "Log in to" the tooltip is too long.

If you feel strongly I'll remove it—I want to clean up some of these misaligned elements more than I want to add the tooltip—but with the surrounding context I think "Enable Sync and Previews" is good.

shaunandrews:
I guess we may have discrepancies with different plans now making the tooltip more complex. I'll just remove it for now.

wojtekn:
@shaunandrews Maybe in the end, we don't need a tooltip for that link? It's not an icon, and is already descriptive.

shaunandrews:
I removed the tooltip for the login button, and updated the branch.

wojtekn (approved):

PR #2676, opened 2026-02-26
[Tooling] Make beta releases self-contained with translations and backmerge

iangmaia:
Relates to AINFRA-2108

## Related issues

- Related to #2583
- wpcom PR: https://github.a8c.com/Automattic/wpcom/pull/205520

## Proposed Changes

- **`code_freeze` lane**: No longer triggers the first beta. Now only creates the release branch, extracts translatable strings, generates draft release notes, and creates a backmerge PR. The first beta is triggered manually via ReleasesV2.
- **`new_beta_release` lane**: Now self-contained — downloads latest translations from GlotPress, bumps the version, triggers the build, and creates a backmerge PR. Added `github_username` parameter for backmerge PR reviewer.
- **Removed `download_translations` lane and pipeline**: Translations are now downloaded as part of each beta release. The standalone `fetch_glotpress_translations` lane remains available for manual use.
- **Updated docs**: `release-process.md` and `localization.md` reflect the new flow.
- **Updated pipeline YAMLs**: `new-beta-release.yml` now passes `github_username`, `download-translations.yml` deleted.

Companion wpcom PR needed for scenario file updates (wpstudio.php).

## Testing Instructions

- Review the Fastfile changes for correctness
- Verify the code freeze and beta release flows match the updated documentation

## Pre-merge Checklist


docs/release-process.md
```diff
@@ -14,21 +14,21 @@ Builds are signed, notarized (macOS), and uploaded to the Apps CDN automatically
...
 - Triggers a build (the build tags the new version and uploads to CDN)
+- Creates a backmerge PR from the release branch into `trunk`
 - Repeat as needed for additional betas
 
 ### 3. Pre-Release
 
 **ReleasesV2 milestone**: Pre-Release
 
-- **Download translations**: Button triggers `download_translations` lane, which fetches translations from GlotPress and commits them to the release branch
 - **Release notes**: Review and refine the draft release notes in `RELEASE-NOTES.txt` on the `release/<version>` branch (a draft is auto-generated during code freeze)
```
AliSoftware, line 32: ⚠️ ❓ Does the removal of this step means that if a release have it's first beta on the same day as the code freeze… but then no additional beta for the rest of the release (because no bugs was found during beta-testing), we won't download the latest translations during release finalization and might thus miss the opportunity to get more recent translations that had been translated by translators post-code-freeze while the app was in beta?!

I'd suggest to always download the latest translations at the time of release finalization scenario phase milestone (or the day before, like we do for Tumblr) to ensure that they get downloaded late in the release cycle and thus give the most time for translators to translate as much as possible.

fastlane/Fastfile
```diff
@@ -211,30 +212,29 @@ lane :code_freeze do |version:, skip_confirm: false, github_username: nil|
...
+  UI.success("Code freeze complete! Created #{branch_name}")
 end
 
 # Create a new beta release on the current release branch.
 #
+# - Downloads latest translations from GlotPress and commits them
 # - Determines the next beta number automatically from package.json
-# - Bumps the version, commits, tags, and triggers a release build
+# - Bumps the version, commits, pushes, and triggers a release build
+# - Creates a backmerge PR from the release branch to trunk
```
AliSoftware, line 226: Note that this means that code-freeze will create one backmerge PR (to commit the version bump + extracted translatable strings etc), then on the step close after that in the scenario, we will separately trigger the first new_beta_release… which will also create yet another backmerge PR (whose diff this time will mostly only contain the beta version bump and not much else), right?

I'm OK with that if that's acknowledged, but just wanted to highlight that side effect of the separation, since that now means 2 backmerge PRs close to one another instead of a single one (and means waiting for CI to go green on each of those… but for that we can improve the RMs life via `pr_changed_files`-based filtering to reduce the number of CI jobs to run and having to wait for especially on the 2nd PR I guess)

docs/release-process.md
iangmaia: That's a good point -- the previous implementation, at the "Pre-Release" stage, wasn't fully in line with the process either. Running a final fetch during finalize release makes total sense 👍

fastlane/Fastfile
iangmaia: Yeah, good point, I discussed this a bit with @sejas. I like the fact that we'll have backmerge PRs for each beta + download translations and I think that's the main point I wanted to address after the discussions.

But one point I'm not totally convinced about is the need (or not) of the "automatic" first beta build during code freeze (which could create the first backmerge instead of having it both on code freeze + first beta). The implementation removed it, but on a second thought I think having it saves a couple of clicks and binds the release process together a bit more. 🤷

AliSoftware: Another option would be to remove the backmerge PR at the time of `code_freeze`, relying on the fact that a `new_beta_release` will be triggered afterwards as a separate step in the release scenario.

Thinking about it, this is kinda what we do in some other product repos that need a pause between the start of the code freeze and the first beta (e.g. DOAndroid)… except in those repos we've used the convention to have a lane named `complete_code_freeze` for the similar case.
I'm ok to not have a lane called `complete_code_freeze` here if all it does in the case of Studio is doing the exact same as just calling `new_beta_release` directly and nothing more btw. Just drawing parallels.

This also re-enforces my suggestion on the ReleasesV2 PR (205520-ghe-Automattic/wpcom#discussion_r201295) that this step to create the first beta post code-freeze, even if it's now on demand and not done as part of the `code_freeze` lane, should still be a `Task` that already exists in the scenario from scratch when the scenario is created, instead of relying on the Release Manager having to remember to click the "New Beta" button manually.

[EDIT] At the time I wrote this comment I hadn't seen your intermediate reply (as I hadn't refreshed the GitHub page) [/EDIT]

iangmaia: One important point about the Code Freeze backmerge is the .pot file update, so that's the main reason I kept the backmerge.

AliSoftware: Yeah but if we're guaranteed to have a beta being done as a follow up step after the backmerge (be it because the `code_freeze` lane calls the `new_beta_release` lane itself or because we have a 'Task::buildkite` in the scenario to trigger by `new_beta_release` when ready at the end of the `Milestone::code_freeze`), then that backmerge PR that is going to be created as part of that first beta will also contain the change to the `.pot`. So that would just lead to one (not so) big PR containing everything, instead of 2 smaller ones back to back.

iangmaia: Yep, perhaps a `Task::buildkite` is a good middle ground... but at that point, why not just trigger the lane together with `code_freeze`? The main advantage would be to choose when (and given it's part of the same milestone, it's implicit it should be soon, but not enforced).
I think our question comes more from the fact that this gap between code freeze and first beta is a bit of an unknown currently. I checked with @sejas on [Slack](https://a8c.slack.com/archives/C06DRMD6VPZ/p1772211622941149?thread_ts=1771256551.851819&cid=C06DRMD6VPZ) just to be sure.

---

By the way, tangentially related, one small bug we found in the first release was that, if we do an automated (or separate task, for that matter) first beta, the second beta created using "Add Beta" button will still say `Beta Release 1` 😄  not sure if this came up in other apps, but in Studio this became more visible as the increment is part of the version used everywhere.

AliSoftware: > By the way, tangentially related, one small bug we found in the first release was that, if we do an automated (or separate task, for that matter) first beta, the second beta created using "Add Beta" button will still say Beta Release 1 😄 not sure if this came up in other apps, but in Studio this became more visible as the increment is part of the version used everywhere.

I think this has always been the case even in other products? And probably part of why we've been calling the betas added via the button as "**Intermediate** beta" (as a cheat to suggest that there's the first beta made during code freeze, then a first intermediate beta after that, etc) 😅

So even in other products where we use different naming conventions of where we do the first beta as part of the code freeze lane this is also the case (eg first beta done during code freeze is named `-rc-1` and Intermediate Beta 1 makes this bumped to `-rc-2`)

AliSoftware: Random example with PockatCasts Android:
 - [Code Freeze](https://buildkite.com/automattic/pocket-casts-android/builds/15481/steps/canvas) does the version bump then builds a beta, [which results in `app-8.6-rc-1.apk`](https://buildkite.com/automattic/pocket-casts-android/builds/15481/steps/canvas?sid=019c6529-c3b0-4451-ad59-d32b5e95b59f&tab=artifacts)
 - [Intermediate Beta 1](https://mc.a8c.com/releases-v2/release.php?product=PCAndroid&version=8.6#task-df18ac36-b633-4930-bcca-05bfbc225af3-fragment) triggers [a build](https://buildkite.com/automattic/pocket-casts-android/builds/15520/steps/canvas) which bumps the version to `8.6-rc-2`

So that off-by-one in the title has always been there.

Maybe if we renamed the task in Releases V2 from "Intermediate Beta 1" to "Intermediate Bugfix build 1" to make the distinction even clearer between "the first build done during code freeze" vs "subsequent beta builds that are only done if we found a bug and merged a fix and want to do a new beta to include that bugfix". But at that point I think that'd be wayyy too nitpicky 😄 

Besides, there is also the seldom but possible case that a failed beta build would need to be retried and end up bumping the version number multiple times (see the case we had this morning with DOAndroid — p1772201201312009-slack-C06CKSPHYA1) which would lead to the version being `-rc-2` at the end of the code freeze, and the "Intermediate Beta 1" would then produce an `-rc-3` 😛 

So in the end I think we don't need to be too much attached with this off-by-one offset between "Intermediate Beta" index and `-betaN` version name?

docs/release-process.md
iangmaia: Updated on f4ccc0108b8248dc336a453192bc6e9f5e535bf6.

fastlane/Fastfile
sejas: Would it be possible to separate the pots generation and the submission of strings to GlotPress from the code freeze milestone?
- Day 1:  Submit strings (Back merge)
- Day 3: Produce Beta1 (Codefreeze + beta + backmerge)
- Day 3+: Produce more betas manually + backmerge
- Day 5 : Build release
- Day 8: Publish release

sejas: We decided p1772217237435699/1771256551.851819-slack-C06DRMD6VPZ to produce the code freeze and then a different milestone to produce a new beta1 a couple of days later. We always can iterate in these decisions after trying the real process a couple of cycles.

AliSoftware: So if I understand the discussion from Slack correctly, that will be:

 - Day 1: Freeze the code and the associated strings (i.e. Code Freeze to create the release branch and generate the pot then backmerge to trunk so that the strings are uploaded to GlotPress)
 - Day 3: Produce Beta1 (new beta build + backmerge)
 - Day 3+: Produce more betas manually + backmerge
 - Day 5 : Build release
 - Day 8: Publish release

If so, then that sounds good to me 👍

iangmaia: I've updated the [wpcom PR](https://github.a8c.com/Automattic/wpcom/pull/205520) to follow this pattern 👍

AliSoftware (approved):

PR #2681, opened 2026-02-27
Prefill admin email and make it mandatory

bcotrim:
## Related issues

Related to #2527 (Add support for Blueprints login)

## Proposed Changes

- Prefill the admin email field with `admin@localhost.com` instead of leaving it empty
- Make email validation mandatory (always enabled, not conditional on user input)
- Remove the informational comment explaining the default value
- For pre-existing sites without a stored email, automatically fetch it from WordPress via `wp user get <username> --field=user_email` when opening site settings (uses the stored admin username or defaults to `admin`)
- Simplify form submission logic to always send the email value (since it's now always populated)

## Testing Instructions

**For new sites:**
1. Create a new site
2. In the advanced settings, verify the email field is prefilled with `admin@localhost.com`
3. Try to clear the field and save — should show validation error "Admin email cannot be empty."
4. Enter a valid email and confirm the site is created

**For pre-existing sites:**
1. If you have a site created before PR #2527 without stored email:
   - Open its settings
   - Watch the email field populate with the real email from WordPress (fetched via WP-CLI)
2. Edit the email and save — should persist to appdata

## Pre-merge Checklist


gcsecsey (approved):
Thanks @bcotrim for improving this! 🙌 This is looking good to me, and tests well. Unfortunately I didn't have a site from earlier, but I confirm the validation works great for a new site:
 
https://github.com/user-attachments/assets/9e4565aa-3130-4199-86e6-3930f43e13f4

epeicher (approved):
Thanks @bcotrim! I have tested it, and I see the email validation when creating the site and when deleting the email from an existing (or new) site. LGTM!

One small detail that can be tacked as a follow-up is that the validation message displayed when editing a site and deleting the email is hidden from the viewport and the user needs to scroll down. Initially, I saw the red outline but didn't see the message, then I realised I needed to scroll, please see the video below 

https://github.com/user-attachments/assets/cb26c98c-577d-411c-a0c4-96293f1cfb4b

This is such a minor detail, but I believe addressing it will enhance the validation UX.

bcotrim:
> One small detail that can be tacked as a follow-up is that the validation message displayed when editing a site and deleting the email is hidden from the viewport and the user needs to scroll down. Initially, I saw the red outline but didn't see the message, then I realised I needed to scroll, please see the video below

Nice catch, thanks! I've addressed it in https://github.com/Automattic/studio/pull/2681/commits/f133abf1fa45b4ab04d1c41c75b9c5443f7858e3

PR #2684, opened 2026-03-02
Studio Agents Suite: Add starter agents.md file and ship it to site root

katinthehatsite:
## Related issues

Fixes https://linear.app/a8c/issue/STU-1332/create-starter-agentsmd-template
Related to https://linear.app/a8c/issue/STU-1334/implement-auto-placement-of-agentsmd-in-site-root

## Proposed Changes

This PR implements the `AGENTS.md` file to the root of Studio sites. The file includes guidance on Studio CLI commands and usage, WordPress development practices as well as SQLite database documentation. 

## Testing Instructions

* Create a Studio site
* Navigate to the `Overview` tab and open the site in `Finder`
* Confirm that `AGENTS.md` file is present in the root of the site
* Next, open a Terminal and `cd` into the site
* Start, for example, Claude
* Ask Claude about the database integration 
* Confirm that you get the answers related to SQLite database integration
* Ask Claude about Studio CLI commands
* Confirm that it gets answers based on the `AGENTS.md` file

## Pre-merge Checklist


apps/cli/lib/tests/agents-md.test.ts
```diff
@@ -0,0 +1,70 @@
...
+		await writeAgentsMd( sitePath );
+
+		expect( fs.promises.writeFile ).toHaveBeenCalledWith(
+			expectedFilePath,
+			expect.stringContaining( '# AI Instructions' ),
+			'utf-8'
+		);
+	} );
+
+	it( 'does not overwrite an existing AGENTS.md', async () => {
```
katinthehatsite, line 36: For now, we do not overwrite the file. We will consider some more flexibility here to make sure the users get the most recent versions of `AGENTS.md` shall we have to modify it. It will be solved under https://linear.app/a8c/issue/STU-1332/create-starter-agentsmd-template

apps/cli/lib/agents-md.ts
```diff
@@ -0,0 +1,119 @@
+import fs from 'fs';
+import path from 'path';
+
+const AGENTS_MD_FILENAME = 'AGENTS.md';
+
+const AGENTS_MD_TEMPLATE = `# AI Instructions
```
epeicher, line 6: Have we considered extracting this template to an external `.md` file for easier maintenance?

apps/cli/commands/site/create.ts
```diff
@@ -205,6 +206,8 @@ export async function runCommand(
 			isSqliteUpdated ? __( 'SQLite integration configured' ) : __( 'SQLite integration skipped' )
 		);
 
+		await writeAgentsMd( sitePath );
```
epeicher: The `writeAgentsMd` call here could benefit from a try/catch with a non-fatal error, similar to how `deletePreviewSites` handles failures in `delete.ts`. Writing `AGENTS.md` is a nice-to-have; if it fails (e.g., permissions), it shouldn't block site creation.

Something like:

```typescript
try {
    await writeAgentsMd( sitePath );
} catch ( error ) {
    logger.reportError(
        new LoggerError(
            __( 'Failed to write AGENTS.md. Proceeding anyway…' ),
            error
        ),
        false
    );
}
```

The `false` flag on `reportError` marks it as non-fatal.

epeicher (commented):
Thanks @katinthehatsite for adding that! I have reviewed the Agents.md and it looks great, we can iterate over it as we use it (if required) but it's a great starting point in my opinion. 

 have added a couple of comments for consideration, the main one about externalising the file, but interested in other points of view.

apps/cli/lib/agents-md.ts
katinthehatsite: Yeah I initially did that :) Then, I went back on it but I think we will have to do it regardless based on my exploration in https://github.com/Automattic/studio/pull/2698 

I will move it in that PR if that's okay so that I don't have to do another rebase 😅

epeicher: > I will move it in that PR if that's okay so that I don't have to do another rebase

That's fine 👍

epeicher (approved):
I have been testing using Claude and Codex, and both agents work as expected, they see the Agents.md and allow to ask and do actions using natural language, like `Please create a preview site`

Image

I am approving this as my comments are non-blocking

apps/cli/commands/site/create.ts
katinthehatsite: I added the try catch block in this commit: https://github.com/Automattic/studio/pull/2684/changes/00f3022764d52d70a341559583afe1cc8298ce10

sejas (commented):
Thanks, Kat! Agents.md is a great start. I think we could put it behind a feature flag to avoid adding it to all sites yet. Although, if we install it from the UI, that could be sufficient.

katinthehatsite:
> Thanks, Kat! Agents.md is a great start. I think we could put it behind a feature flag to avoid adding it to all sites yet. Although, if we install it from the UI, that could be sufficient.

Sounds good to me, I will adjust https://github.com/Automattic/studio/pull/2698 based on this suggestion to make sure that we don't install it automatically until the user does the installation from the UI. 

At the moment, it gets installed to all sites automatically but then the user can remove it or reinstall it back from the UI. We can also adjust the initial installation to happen only for the UI.

PR #2609, opened 2026-02-18
Fix Studio CLI yargs locale strings displaying in English

bcotrim:
## Related issues

- Fixes STU-1232

## Proposed Changes

- Import yargs locale JSON files directly into the bundle to bypass y18n's broken path resolution
- Use `yargs.updateLocale()` to apply inline locale data instead of relying on filesystem-based locale loading
- Add missing `uk` -> `uk_UA` locale mapping in `mapToYargsLocale()`
- Return locale data from `loadTranslations()` for yargs to use

## Testing Instructions

Set a non-English locale in Studio UI and run `studio --help`:

**Portuguese (Brazil):**
```
studio --help
```
Verify "Comandos:", "Opções:", "Exibe ajuda" appear (not "Commands:", "Options:", "Show help")

**Ukrainian:**
```
studio --help
```
Verify "Команди:", "Опції:", "Показати довідку" appear

**Chinese (Simplified):**
```
studio --help
```
Verify Chinese characters appear for yargs strings

**English:**
Verify fallback to English still works correctly.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #2685, opened 2026-03-02
Fix Typescript errors and clean up TS configs

fredrikekelund:
## Related issues

Fixes STU-1373

## Proposed Changes

We currently have a few TS errors in our code. This PR addresses that with the following changes:

- Add a `typecheck` npm script.
- Update `AGENTS.md` to tell agents to do type checking.
- Run that script as part of the lint CI job. 
- Removed `"*": [ "node_modules/*" ]` resolvers in `tsconfig.json` files. This means it's no longer possible to import any code from `node_modules` files. Instead, we use regular module resolution to import available modules. 
  - This may sound like a big change, but it's really not. I only had to apply a minor edit to a single file (`use-update-button-tooltip.ts`) to make this work. 
- Install `@types/glob` and `wpcom` to prevent TS from complaining. 
- Externalize `yargs` dependency during build. This is a different way of tackling the issue raised in https://github.com/Automattic/studio/pull/2609. I had to address this because yarg's `updateStrings` function didn't play nice with the type definitions from `apps/cli/lib/i18n.ts`. By not bundling the yargs code, we sidestep the problem altogether. 
- Fix all other type errors.

## Testing Instructions

CI should pass.

## Pre-merge Checklist


apps/cli/patches/yargs+18.0.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/yargs/lib/platform-shims/esm.mjs b/node_modules/yargs/lib/platform-shims/esm.mjs
+index 1b8efe8..0e34ec4 100644
+--- a/node_modules/yargs/lib/platform-shims/esm.mjs
++++ b/node_modules/yargs/lib/platform-shims/esm.mjs
+@@ -61,7 +61,7 @@ export default {
+   },
+   stringWidth,
+   y18n: y18n({
+-    directory: resolve(__dirname, '../../../locales'),
++    directory: resolve(dirname(require.resolve('yargs')), 'locales'),
```
fredrikekelund: The `__dirname` relative path lookup fails because we bundle yargs into `apps/cli/dist/cli/main.js`. Weirdly, translations still worked for me even without this change (after I reverted https://github.com/Automattic/studio/pull/2609)… In any case, this change produces the correct path.

sejas (approved):
Changes look good. I tested it the app by starting a site and logging in. I also tried executing the new CLI command `npm run typecheck`.

fredrikekelund:
Thanks for the review, @sejas 🙏

PR #2694, opened 2026-03-03
Studio: Resolve type error for blueprint steps

katinthehatsite:
## Related issues

N/A

I noticed this issue while working on something else.

## Proposed Changes

This PR resolves the type error in `studio/apps/cli/commands/site/create.ts`: 

```
Property 'steps' does not exist on type 'Blueprint'.
  Property 'steps' does not exist on type 'ReadableFilesystemBackend'.
```

## Testing Instructions

* Try creating a Studio site and confirm you can create it with no further errors
* Open `studio/apps/cli/commands/site/create.ts` and confirm there are no type errors

## Pre-merge Checklist


epeicher (approved):
Thanks @katinthehatsite! I can confirm that I no longer see the type errors. LGTM! :shipit: 

|Before | After|
|-------|------|
|Image| Image|

I think the `package-lock.json` changes are not required, and if you merge with latest `trunk` the E2E error will be fixed

apps/cli/commands/site/create.ts
```diff
@@ -296,8 +296,9 @@ export async function runCommand(
 				const blueprintDir = fs.mkdtempSync( path.join( os.tmpdir(), 'studio-empty-blueprint-' ) );
 				blueprintUri = path.join( blueprintDir, 'blueprint.json' );
 			}
-			const existingSteps = blueprint.steps || [];
-			blueprint.steps = [ ...setupSteps, ...existingSteps ];
+			const blueprintDecl = blueprint as BlueprintV1Declaration;
```
nightnei, line 299: I don't see error in my Cursor, maybe something is broken in my local configuration, but in general the solution seems like a workaround for wp-playground and we need to report it.
What concerns me is:
1. Above in the code we define `let blueprint: Blueprint | undefined;`
2. Under the hood we have `export type Blueprint = BlueprintV1 | BlueprintV2;`
3. Under the hood of the previous statement, we have `export type BlueprintV1 = BlueprintV1Declaration | BlueprintBundle;`
4. So, ultimately it should be `BlueprintV1Declaration`. Since you see that error and casting `as BlueprintV1Declaration` fixes the issue, then we indeed should report this ts bug instead of adding workaround.

It's ts, so it shouldn't affect anything, and I would go with keeping the error as is and reporting the bug. Unless it breaks something right now, but it seems it isn't.
But as a temporary solution, I think we can go with the approach in this PR.

WDYT? Am I missing something?

UPDATE: restarting the Cursor - showed the error to me.

nightnei (changes requested):

gcsecsey (approved):
Thanks @katinthehatsite for fixing this. This LGTM and I see no more issues on the create file. 👍 

Also, when creating a new site, I see no errors:

Image

nightnei (approved):
The changes are fixing the issue 👍 So feel free to move forward and merge it. I just [shared my thoughts above](https://github.com/Automattic/studio/pull/2694#discussion_r2878212104).

apps/cli/commands/site/create.ts
katinthehatsite: I think we can merge this for now and I will report this to Playground team. It is likely will be marked a low priority so it might take a while for the fix to come in.

katinthehatsite: I added follow up here: https://github.com/WordPress/wordpress-playground/issues/3336

fredrikekelund: This isn't a safe type assertion. Typescript is flagging a legitimate type error here.

I recommend changing the type of the `let blueprint` declaration to `BlueprintV1Declaration`. This will require us to also change the return type of `filterUnsupportedBlueprintFeatures`, and potentially some other related functions. All of this is what we want, though, because `BlueprintV1Declaration` is the data format we support.

katinthehatsite: I will take a look 👀

fredrikekelund: I went down a rabbit hole with this and ended up publishing https://github.com/Automattic/studio/pull/2719, which addresses the problem and some related issues.

PR #2689, opened 2026-03-02
[Tooling] Upload CDN builds as internal first, publish as external later

iangmaia:
## Related issues

- Fixes AINFRA-2102

## Proposed Changes

- Upload all CDN build artifacts with `visibility: internal` (Automatticians-only) during `finalize_release` / `distribute_release_build`, instead of `external`
- Embed CDN post IDs as an HTML comment in the draft GitHub release body so they can be retrieved later
- In `publish_release`, read the draft release body, extract post IDs, and call `update_apps_cdn_build_metadata` for each build to flip visibility to `external` before publishing the GitHub release
- This gives the team a window for internal testing between finalize and publish

**Depends on:** wordpress-mobile/release-toolkit PR for the new `update_apps_cdn_build_metadata` action.

## Testing Instructions

- The `update_apps_cdn_build_metadata` action needs to be manually verified against the WPCOM API first (see release-toolkit PR)
- Once confirmed, test by running a release through the full flow:
  1. `finalize_release` → verify builds appear on CDN with Internal visibility
  2. `publish_release` → verify builds are updated to External visibility and GitHub release is published

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2695, opened 2026-03-03
[Tooling] Upload CDN builds as drafts, publish in publish_release

iangmaia:
## Related issues

- This PR depends on the PRs:
  - wordpress-mobile/release-toolkit#701 (adds `update_apps_cdn_build_metadata` action)
  - wordpress-mobile/release-toolkit#702 (adds `list_apps_cdn_builds` action)
- Fixes AINFRA-2102

## How AI was used in this PR

Claude Code was used to implement the Fastfile changes based on a detailed plan I designed. I reviewed all generated code.

## Proposed Changes

- Upload all non-beta CDN builds as **drafts** with external visibility (previously published immediately), so they're not publicly accessible during the testing period
- Beta builds continue to be published immediately (no `post_status: 'draft'`)
- Add `make_cdn_builds_public` helper that queries the CDN API via `list_apps_cdn_builds` to find draft builds for the release version, then publishes each via `update_apps_cdn_build_metadata`
- Call `make_cdn_builds_public` in `publish_release` before publishing the GitHub release
- Update `finalize_release` prompt to mention the new draft behavior

### Why drafts instead of internal visibility?

Integration testing against the real CDN API revealed that posts with `visibility: internal` are completely hidden from all listing queries (by CDN plugin design), making it impossible to find and flip them later. The draft/publish approach works reliably and is idempotent — re-running when no drafts exist is a safe no-op.

## Testing Instructions

- Full end-to-end testing will happen during the next release cycle
- The draft→publish flow was manually verified against the real CDN API

## Pre-merge Checklist

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -274,7 +274,7 @@ lane :finalize_release do |version:, skip_confirm: false|
     Finalizing release #{version}. This will:
     - Bump version to #{version} (remove beta suffix)
     - Trigger a release build for all platforms (macOS, Windows), which will then:
-      - Upload build artifacts to the Apps CDN (as Internal, visible only to Automatticians)
+      - Upload build artifacts to the Apps CDN (as drafts, published later by publish_release)
       - Create a draft GitHub release with release notes and download links
       - Notify #dotcom-studio on Slack
   PROMPT
@@ -300,7 +300,7 @@ lane :publish_release do |version:, skip_confirm: false, github_username: nil|
 
   UI.important <<~PROMPT
     Publishing release #{version}. This will:
-    - Make CDN builds public (change visibility from Internal to External)
+    - Publish CDN builds (change post status from draft to publish)
     - Publish the draft GitHub release for v#{version}
     - Create a backmerge PR from `#{release_branch}` into `#{MAIN_BRANCH}`
     - Delete the `#{release_branch}` branch after creating the backmerge PR
@@ -585,7 +585,8 @@ def distribute_builds(
       arch: build[:arch],
       build_type: build_type,
       install_type: build[:install_type],
-      visibility: :internal,
+      visibility: :external,
+      post_status: release_tag&.match?(/beta/i) ? nil : 'draft',
       version: version,
       build_number: build_number,
       release_notes: release_notes,
@@ -823,10 +824,11 @@ def create_draft_github_release(version:, release_tag:, builds:)
   UI.success("Created draft GitHub release #{release_tag} with download links")
 end
 
-# Make CDN builds public by updating their visibility from Internal to External.
+# Publish CDN builds by changing their post status from draft to publish.
 #
-# Queries the Apps CDN API to find internal builds matching the version,
-# then calls {update_apps_cdn_build_metadata} for each to set visibility to External.
+# Queries the Apps CDN API to find draft builds matching the version,
+# then calls {update_apps_cdn_build_metadata} for each to set status to publish.
+# This is idempotent: if no drafts are found (already published), it logs a warning and returns.
 #
 # @param version [String] The version to publish (e.g., '1.7.5')
 #
@@ -835,26 +837,26 @@ def make_cdn_builds_public(version:)
   builds = list_apps_cdn_builds(
     site_id: WPCOM_STUDIO_SITE_ID,
     version: release_version,
-    visibility: :internal
+    post_status: 'draft'
   )
 
   if builds.empty?
-    UI.important("No internal CDN builds found for #{release_version}. They may have already been made public.")
+    UI.important("No draft CDN builds found for #{release_version}. They may have already been published.")
     return
   end
 
-  UI.message("Updating CDN build visibility to External for #{builds.size} builds...")
+  UI.message("Publishing #{builds.size} CDN build(s) for #{release_version}...")
 
   builds.each do |build|
-    UI.message("  Updating #{build[:title]} (post #{build[:post_id]})...")
+    UI.message("  Publishing #{build[:title]} (post #{build[:post_id]})...")
     update_apps_cdn_build_metadata(
       site_id: WPCOM_STUDIO_SITE_ID,
       post_id: build[:post_id],
-      visibility: :external
+      post_status: 'publish'
     )
   end
 
-  UI.success('All CDN builds are now public (External visibility)')
+  UI.success('All CDN builds are now published')
 end
 
 # Trigger a release build in Buildkite for the given version.
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -274,7 +274,7 @@ lane :finalize_release do |version:, skip_confirm: false|
     Finalizing release #{version}. This will:
     - Bump version to #{version} (remove beta suffix)
     - Trigger a release build for all platforms (macOS, Windows), which will then:
-      - Upload build artifacts to the Apps CDN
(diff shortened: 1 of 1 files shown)


iangmaia:
Converted to draft until the release-toolkit PRs aren't merged and we can use a new version with both actions.

AliSoftware:
> Integration testing against the real CDN API revealed that posts with `visibility: internal` are completely hidden from all listing queries (by CDN plugin design), making it impossible to find and flip them later. The draft/publish approach works reliably and is idempotent — re-running when no drafts exist is a safe no-op.

What is the consequence of uploading a build to AppsCDN as "Draft", in terms of testability?
I'm guessing that just like a WP post, an app uploaded as "Draft" would not be visible when you load the WP site that lists the builds (just like draft posts aren't visible on a blog), so how would a Release Manager or the team be able to download and test the build in practice?

Maybe it'd be better to adjust the CDN plugin design to allow `list_apps_cdn_builds` to return internal builds somehow… though I guess it'd only allow listing internal builds if the call is done from a proxied IP, while CI machines (where those lanes run) are not proxied, so I guess that's the core of the problem… In that case I could think of 2 possible solutions:
 - Either use Draft status instead of visibility, like you did, but then ensuring it'd be possible (and easy) for the Release Manager to download the build for testing
 - Or store the `post_id` of the build that is uploaded as draft — but the question is where… maybe as some `<!-- post_id:… -->` hidden metadata in the GitHub Release draft description? Or as an annotation to the Buildkite build? — so you wouldn't have to use `list_apps_cdn_builds` to retrieve it later.

fastlane/Fastfile
```diff
@@ -581,7 +585,8 @@ def distribute_builds(
       arch: build[:arch],
       build_type: build_type,
       install_type: build[:install_type],
-      visibility: 'external',
+      visibility: :external,
+      post_status: release_tag&.match?(/beta/i) ? nil : 'draft',
```
AliSoftware, line 589: I'm not sure the condition is right here?

This means that the first beta post code freeze as well as the intermediate betas that occur in the middle of the release will be uploaded as draft (and then never published). While those should probably be internal instead.

That also means that the final build we do at the end of the release process, the one that drops the `beta` suffix because it's supposed to be the final version, will not match the `/beta/i` RegEx, and thus will be uploaded as non-draft. While it's exactly not what we want.

Also, based on the code at the start of this lane, there also exists "Nightly Builds"—for which `release_tag` is `nil`, which means means this will publish those as non-draft as well.

---

I think instead we should:
 - For `build_type` of both `Nightly` and `Beta`, use `visibility: :internal` but `post_status: 'publish'`, so that they are published but only internally
 - For `build_type == 'Production` builds, especially the final build done at the end of the release scenario, use `visibility: :external` but `post_status: 'draft'`… until we get to the next stage in the Release Scenario to publish it once it has been smoke-tested.

fastlane/Fastfile
```diff
@@ -300,12 +300,16 @@ lane :publish_release do |version:, skip_confirm: false, github_username: nil|
...
     Publishing release #{version}. This will:
+    - Publish CDN builds (change post status from draft to publish)
     - Publish the draft GitHub release for v#{version}
     - Create a backmerge PR from `#{release_branch}` into `#{MAIN_BRANCH}`
     - Delete the `#{release_branch}` branch after creating the backmerge PR
   PROMPT
   next unless skip_confirm || UI.confirm('Continue?')
 
+  # Update CDN build visibility from Internal to External
+  make_cdn_builds_public(version: version)
```
AliSoftware, line 311: This is not consistent with what it currently does, which is not about making it public and changing it from internal to external, but about changing it from draft to published…

Of course this might change back depending of if using the Draft state was indeed the right move (i.e. [if it still allows this build to be easily downloadable and testable by the team and RM for smoke-testing](https://github.com/Automattic/studio/pull/2695#issuecomment-3999727134)) or if we're gonna switch back to using internal vs external.

fastlane/Fastfile
```diff
@@ -819,6 +824,41 @@ def create_draft_github_release(version:, release_tag:, builds:)
...
+#
+# @param version [String] The version to publish (e.g., '1.7.5')
+#
+def make_cdn_builds_public(version:)
+  release_version = "v#{version}"
+  builds = list_apps_cdn_builds(
+    site_id: WPCOM_STUDIO_SITE_ID,
+    version: release_version,
+    post_status: 'draft'
+  )
```
AliSoftware, line 841: Is there a risk—especially in the rare case that issues was found in a final release build and the team decides to make a new commit on the `release/` branch then re-upload a new final build—that there would be more than one build with version `v#{version}` in the CDN for each platform, and that we'd then end up publishing all of them (the bad one as well as the subsequent one with the fix) instead of only the last one?

Tbh I'm not sure this is a case the current setup fully support yet anyway.
 - e.g. what happens to the version bump logic if you `finalize_release` once, which removes the `-betaN` suffix from the version, then find an issue and redo another `finalize_release`, would the logic that removes the `-betaN` be a no-op on the version already being `X.Y.Z`? Will it crash?
 - What about the AppsCDN side of things, is the CDN plugin ok with uploading a build with the same metadata (version, platform…) twice? Does it replace the old one with the new one in that case, or keep both entries (which is where your `list_apps_cdn_builds` call would return duplicates)

_Maybe it's only a concern for a future iteration though. But I figured I'd still raise the question, including in case this informs our thoughts about using `visibility: :internal` and not using `list_apps_cdn_builds` but storing the `post_id` values at initial upload time somewhere instead._

iangmaia:
@AliSoftware 
> * Either use Draft status instead of visibility, like you did, but then ensuring it'd be possible (and easy) for the Release Manager to download the build for testing

Yeah, the testing part with Drafts might indeed be tricky 😕 I moved from using visibility to drafts after testing the API but indeed was yet to check the user flow.

> * Or store the `post_id` of the build that is uploaded as draft — but the question is where… maybe as some `<!-- post_id:… -->` hidden metadata in the GitHub Release draft description? Or as an annotation to the Buildkite build? — so you wouldn't have to use `list_apps_cdn_builds` to retrieve it later.

This was actually my original approach (see https://github.com/Automattic/studio/pull/2689, which I closed in favor of this PR using drafts). I considered to be somewhat fragile. But thinking about it again... perhaps it isn't such a bad idea as it has less dependencies, the posts are tied up nicely with ids (without relying on listing / searching) and changing GitHub releases (which could be a concern) requires GitHub repo permissions anyway.

Wdyt? I think using visibility will probably be less hassle on the testing side, plus it isn't much more complicated than using status. Using `<!-- post_id ... ->` connected to the release itself doesn't sound that bad. 🤔

AliSoftware:
One other thing I like about storing the ist of `post_id` values during `finalize_release` and re-read them during `publish_release` is that it feels more deterministic, as in it will only publish builds for the last version that was finalized.

So even if you `finalize_release` more than once (because you found a very-last-minute bug during smoke-testing after finalization and added commits to fix it), then during `publish_release` you will still only fetch the _last_ GitHub Release draft for that version, then read the list of `post_id` values (one for each platform/arch tuple) for just that release, even if a previous GitHub Release draft was created for the previous attempts.

iangmaia:
Closing in favor of #2724, which goes back to the internal visibility + post IDs approach from #2689 (as discussed in https://github.com/Automattic/studio/pull/2695#issuecomment-4004921627). The draft-based approach had concerns around testability for the Release Manager and was less deterministic than storing post IDs in the GitHub release body.

PR #2696, opened 2026-03-03
Patch Playground to reduce worker count on low-memory machines

sejas:
## Related issues

- Fixes STU-1369

## How AI was used in this PR

Planned with Opus 4.6 and created the code, patch, and PR fully automated. Tweaked a bit the testing instructions. Reviewed and tested the PR.

## Proposed Changes

- Patch `@wp-playground/cli` to dynamically set worker count based on available memory
- When total RAM (totalmem) is less than 8GB, use 3 workers instead of the hardcoded 6
- Reduces memory pressure on Windows and other memory-constrained environments
- Patches both the ESM and CJS bundles

## Testing Instructions

- Run `nvm use && npm i`
- Observe the patches are applied correctly, with no errors.
- Create a new site and verify it boots correctly
- Check `node_modules/@wp-playground/cli/run-cli-Cv4FHhj_.js` line 1284 contains `q.freemem() < 8 * 1024 ** 3 ? 3 : 6`
- Check `node_modules/@wp-playground/cli/run-cli-NCnYfPz7.cjs` line 47 contains `G.freemem()<8*1024**3?3:6`
- Run `npm install` and verify the patch applies without errors

## Pre-merge Checklist


fredrikekelund:
> When free RAM is less than 8GB, use 3 workers instead of the hardcoded 6

This rule would make more sense if the total available system memory was 8GB. Given that a site process uses ~1.6 GB, this seems like a too harsh limit right now. 

There's also the fact that this will make some sites slower than others with no clear explanation as to why…

I'm not immediately sure about the best solution here, but looking at `os.totalmem()` instead of `os.freemem()` seems like one place we could start

sejas:
What if we limit the number of workers based on OS? Windows -> 3 , else -> 6.<br>My Mac used the SWAP for the sites that were beyond the physical RAM, and as a user I didn't notice the difference.

sejas:
That would make consistent the performance of each site for both platforms.

sejas:
With 3 workers, Parallels (8GB RAM) can run 6 sites in parallel (from 3 without the patch).

sejas:
@fredrikekelund, I think your suggestion is the best approach. Let's use `totalmem`: [https://github.com/Automattic/studio/pull/2696/changes/31ea19fe6e0efd78c31935506f8499982731564b](<https://github.com/Automattic/studio/pull/2696/changes/31ea19fe6e0efd78c31935506f8499982731564b>)

epeicher (approved):
Thanks @sejas for the fix! I can confirm that this improves the performance on Windows, and it allows to create more than 3 sites, I have tested with 5. Also, I agree with using a fix number better than a dynamic so 👍 to https://github.com/Automattic/studio/commit/31ea19fe6e0efd78c31935506f8499982731564b.

sejas:
Closing in favor of https://github.com/Automattic/studio/pull/2704 and https://github.com/Automattic/studio/pull/2702.

With the new Playground version, Windows does a better job managing swap and doesn’t fail immediately when allocating memory for a new site.

PR #2697, opened 2026-03-03
Configure CLI build pipeline for npm publishing

ivan-ottinger:
## Related issues

- Fixes STU-1362

## How AI was used in this PR

This PR has been through plenty of iterations. First by @ivan-ottinger and then by @fredrikekelund. For the current iteration, Fredrik has used AI to fix targeted issues and smaller tasks. 

## Proposed Changes

- Update `apps/cli/package.json` with the basic properties needed to publish to npm (name, version, bin, etc).
- Added a separate Vite config (`vite.config.npm.ts`) specifically for publishing to npm. It defines some global variables that enable Vite/Rollup to perform better tree-shaking, and it adds a "banner" to the main bin JS file. 
  - All external `package.json` dependencies are now defined as external in all builds. This prevents us from distributing the same code twice (in the bundle and in `node_modules`).
  - Marking dependencies as external typically increases the memory footprint of the CLI (I haven't tested the current iteration, but I did so a couple of weeks ago). To counteract this, I've changed the command imports in `apps/cli/index.ts` to be dynamic rather than static, so the CLI mostly imports only the code it actually uses for the command at hand. 
- Adjusted the `postinstall` script for the CLI to work both during development and distribution. The end goal is to remove this script altogether, as it's correctly perceived as a security risk. 
- Added README and LICENSE files that will be distributed along with the npm package. 

## Testing Instructions

1. Run `npm run cli:build` to generate a dev build for the CLI
2. Ensure that `node apps/cli/dist/cli/main.js site list` works as expected
3. Run `npm run cli:build:npm` to generate an npm build for the CLI
4. Ensure that `node apps/cli/dist/cli/main.js site list` works as expected again
5. Open Studio, go to the settings modal and uninstall the CLI
6. Run `npm i -g @automattic/wp-studio@next`
7. Run `studio site list`
8. Ensure that it works as expected

## Pre-merge Checklist


apps/cli/package.json
```diff
@@ -1,40 +1,74 @@
 {
-	"name": "studio-cli",
+	"name": "@automattic/studio-cli",
 	"author": "Automattic Inc.",
-	"private": true,
+	"version": "0.1.0",
 	"productName": "Studio CLI",
 	"description": "WordPress Studio CLI",
 	"license": "GPL-2.0-or-later",
 	"main": "index.js",
```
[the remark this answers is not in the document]
ivan-ottinger: Good catch — removed the `main` field in 531f4cbd. This is a CLI-only package, so it's not needed.

apps/cli/package.json
```diff
@@ -1,40 +1,74 @@
...
+	"bin": {
+		"studio": "./dist/cli/main.js"
+	},
+	"files": [
+		"dist/cli/*.js",
+		"dist/locales/",
+		"patches/",
+		"scripts/"
+	],
```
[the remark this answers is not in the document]
ivan-ottinger: Fixed in 531f4cbd — added `dist/cli/*.map` to the `files` whitelist.

apps/cli/scripts/postinstall-npm.mjs
```diff
@@ -0,0 +1,76 @@
...
+	process.exit( 0 );
+}
+
+// Apply patches via patch-package
+try {
+	execSync( 'npx patch-package --patch-dir ./patches', {
+		cwd: packageDir,
+		stdio: 'inherit',
+	} );
+} catch {
```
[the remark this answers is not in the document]
ivan-ottinger: Fixed in 531f4cbd — now using `npx --no-install patch-package`.

apps/cli/scripts/postinstall-npm.mjs
```diff
@@ -0,0 +1,76 @@
...
+					rmSync( asyncifyPath, { recursive: true, force: true } );
+					console.log( `Pruned ${ asyncifyPath }` );
+				}
+			}
+			// Remove web-* packages entirely
+			if ( /^web-\d+-\d+/.test( entry ) ) {
+				const webPath = join( phpWasmDir, entry );
+				rmSync( webPath, { recursive: true, force: true } );
+				console.log( `Pruned ${ webPath }` );
+			}
```
fredrikekelund: https://github.com/WordPress/wordpress-playground/pull/3315 made this redundant. I just haven't had a change to update `apps/cli/vite.config.ts` yet.

apps/cli/scripts/postinstall-npm.mjs
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+} catch {
+	// Fail silently — dependencies may not be present
+}
```
fredrikekelund: The asyncify builds take up around 250 MB. That's significant, but I'm still dubious of removing `node_modules` files in this way for the standalone CLI. I would prefer to let npm just do its thing here, honestly.

apps/cli/scripts/postinstall-npm.mjs
```diff
@@ -0,0 +1,76 @@
...
+				unlinkSync( join( binDir, file ) );
+				console.log( `Removed ${ file }` );
+			} catch ( e ) {
+				console.log( `Could not remove ${ file }: ${ e.message }` );
+			}
+		}
+	}
+} catch {
+	// Fail silently — fs-ext-extra-prebuilt may not be present
+}
```
fredrikekelund: I don't believe we need to touch these files in this context. We had trouble with the `fs-ext-extra-prebuilt` binaries during code signing. Windows was complaining about signing macOS binaries. That won't be an issue here.

apps/cli/scripts/postinstall-npm.mjs
```diff
@@ -0,0 +1,76 @@
...
+
+// Apply patches via patch-package
+try {
+	execSync( 'npx --no-install patch-package --patch-dir ./patches', {
+		cwd: packageDir,
+		stdio: 'inherit',
+	} );
+} catch {
+	console.log( 'patch-package failed — patches may not apply cleanly' );
+}
```
fredrikekelund: Removing `patch-package` isn't trivial… The `pm2-axon` and `pm2` patches will soon be redundant. We might be able to manage without the `ps-man` patch. The `@wp-playground/wordpress` patch could potentially be merged upstream in some way (gated by some config flag).

apps/cli/package.json
```diff
@@ -1,40 +1,74 @@
...
 	},
 	"scripts": {
 		"build": "vite build --config ./vite.config.ts",
+		"build:npm": "vite build --config ./vite.config.npm.ts",
 		"install:bundle": "npm install --omit=dev --no-package-lock --no-progress --install-links --no-workspaces && patch-package && node ../../scripts/remove-fs-ext-other-platform-binaries.mjs",
 		"package": "npm run install:bundle && npm run build",
-		"watch": "vite build --config ./vite.config.ts --watch"
+		"watch": "vite build --config ./vite.config.ts --watch",
+		"prepublishOnly": "npm run build:npm",
+		"postinstall": "node ./scripts/postinstall-npm.mjs"
```
fredrikekelund: It would be great if we could remove this. I've left a couple of comments in `apps/cli/scripts/postinstall-npm.mjs` about what it would take to do that.

The reason I would prefer to avoid `postinstall` scripts is that they're seen as a security risk, and pnpm, for example, [disables](https://pnpm.io/supply-chain-security) them by default.

apps/cli/package.json
```diff
@@ -1,40 +1,74 @@
...
 		"pm2": "^6.0.14",
 		"pm2-axon": "^4.0.1",
 		"trash": "^10.0.1",
+		"wpcom": "^7.1.1",
+		"wpcom-xhr-request": "^1.3.0",
 		"yargs": "^18.0.0",
-		"yargs-parser": "^22.0.0"
+		"yargs-parser": "^22.0.0",
+		"yauzl": "^3.2.0",
+		"zod": "^4.0.0"
```
fredrikekelund: Did we add all the `@studio/common` dependencies here? Why is that needed? They're not marked as external in the Vite config, so bundling should already include them, no?

apps/cli/vite.config.npm.ts
fredrikekelund, line 1: Could we make this config extend `apps/cli/vitest.config.ts` directly, or, alternatively, use a shared base config that both files extend? This will make it easier to maintain them.

fredrikekelund (commented):
Thanks for tackling this! The main questions I see here are:

1. We need to think about how to handle `@studio/common` when publishing to npm. I don't think the current approach of just including those dependencies in `apps/cli/package.json` is reliable enough. 
2. We should reduce the postinstall script as much as possible. In a follow-up PR, we should consider even removing `patch-package` altogether.

apps/cli/scripts/postinstall-npm.mjs
wojtekn: Is there anything against keeping the `patch-package`?

Even if we clear current ones, there is a chance we will need hotfix something in the future, so it seems reasonable to have that way available for both UI and CLI.

fredrikekelund: I expanded a little bit on the "why" in https://github.com/Automattic/studio/pull/2697#discussion_r2889125516

Basically, the ideal end state to me is that we don't have a postinstall script when people install this package. pnpm disables the postinstall script by default and touts this as a security feature (which it is). Yarn mentions this in their [docs](https://yarnpkg.com/advanced/lifecycle-scripts#postinstall):

> Postinstall scripts should be avoided at all cost, as they make installs slower and riskier. Many users will refuse to install dependencies that have postinstall scripts. Additionally, since the output isn't shown out of the box, using them to print a message to the user will not work as you expect.

I don't know that the npm CLI has taken a stance on this, but given the multiple high-profile security incidents in the npm ecosystem in recent years, it's not unreasonable to expect them to address this. 

For us, `patch-package` is the most difficult thing to remove. It's not feasible to tackle that in this PR, but the end goal of removing the postinstall script from the CLI package is definitely something we should consider.

apps/cli/scripts/postinstall-npm.mjs
ivan-ottinger: Makes sense! I have addressed this in https://github.com/Automattic/studio/pull/2697/changes/96c832c4f3c3d5316e03e9befe4f2312f8c366ca.

apps/cli/scripts/postinstall-npm.mjs
ivan-ottinger: I have removed pruning in https://github.com/Automattic/studio/pull/2697/changes/3564d2c2512b4a95f6f0532867febef769a8d85a.

apps/cli/scripts/postinstall-npm.mjs
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2697/changes/96c832c4f3c3d5316e03e9befe4f2312f8c366ca.

apps/cli/package.json
ivan-ottinger: Good point. I have made the change in https://github.com/Automattic/studio/pull/2697/commits/55016e8fcb5f62c04f0644246806661e757a7ac8.

apps/cli/vite.config.npm.ts
ivan-ottinger: Sounds good. We save some lines of code as well. Made the change in https://github.com/Automattic/studio/pull/2697/commits/0b59b52030af0c1bca963d894450e969579de373

ivan-ottinger:
> We should reduce the postinstall script as much as possible. In a follow-up PR, we should consider even removing patch-package altogether.

I have created related issue: STU-1375.

apps/cli/package.json
ivan-ottinger: I left the link to the new Linear task below already, but for completeness adding it here too: STU-1375.

apps/cli/vite.config.npm.ts
```diff
@@ -0,0 +1,73 @@
...
+import { baseConfig, yargsLocalesCopyPlugin } from './vite.config.base';
+
+const packageJson = JSON.parse( readFileSync( resolve( __dirname, 'package.json' ), 'utf-8' ) );
+const packageVersion = packageJson.version;
+
+// Externalize all runtime dependencies except @studio/common (which is bundled via alias)
+const externalDeps = Object.keys( packageJson.dependencies || {} ).filter(
+	( dep ) => dep !== '@studio/common'
+);
+
```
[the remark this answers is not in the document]
ivan-ottinger: Good catch — simplified in c57d86c0. Removed the filter and updated the comment since `@studio/common` is in `devDependencies`, not `dependencies`.

apps/cli/scripts/postinstall-npm.mjs
```diff
@@ -0,0 +1,32 @@
...
+
+// Apply patches via patch-package
+try {
+	execSync( 'npx --no-install patch-package --patch-dir ./patches', {
+		cwd: packageDir,
+		stdio: 'inherit',
+	} );
+} catch {
+	console.log( 'patch-package failed — patches may not apply cleanly' );
```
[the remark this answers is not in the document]
ivan-ottinger: Fixed in 4ee75993 — the postinstall now logs the error to stderr and exits with code 1 so installs fail loudly when patches can't be applied.

apps/cli/vite.config.npm.ts
```diff
@@ -0,0 +1,73 @@
...
+function shebangPlugin(): Plugin {
+	return {
+		name: 'shebang',
+		apply: 'build',
+		generateBundle( _options, bundle ) {
+			const mainChunk = bundle[ 'main.js' ];
+			if ( mainChunk && mainChunk.type === 'chunk' ) {
+				mainChunk.code = '#!/usr/bin/env node\n' + mainChunk.code;
+			}
+		},
```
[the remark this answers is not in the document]
ivan-ottinger: Fixed in 22d33ad4 — replaced the `generateBundle` plugin with Rollup's `output.banner` option, which is sourcemap-aware. The banner function conditionally adds the shebang only to `main.js`.

apps/cli/vite.config.npm.ts
```diff
@@ -0,0 +1,58 @@
...
+					if (
+						/^(path|fs|os|child_process|crypto|http|https|http2|url|querystring|stream|util|events|buffer|assert|net|tty|readline|zlib|constants|tls|domain|dns)$/.test(
+							id
+						)
+					) {
+						return true;
+					}
+					if ( id === 'fs/promises' || id === 'dns/promises' ) {
+						return true;
+					}
```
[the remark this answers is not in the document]
ivan-ottinger: Fixed in 59e0919d — the npm config now imports and reuses `nodeBuiltinExternals` from `vite.config.base.ts` instead of duplicating the patterns.

apps/cli/package.json
```diff
@@ -25,16 +45,19 @@
 	},
 	"scripts": {
 		"build": "vite build --config ./vite.config.ts",
+		"build:npm": "vite build --config ./vite.config.npm.ts",
 		"install:bundle": "npm install --omit=dev --no-package-lock --no-progress --install-links --no-workspaces && patch-package && node ../../scripts/remove-fs-ext-other-platform-binaries.mjs",
 		"package": "npm run install:bundle && npm run build",
-		"watch": "vite build --config ./vite.config.ts --watch"
+		"watch": "vite build --config ./vite.config.ts --watch",
```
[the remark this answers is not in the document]
ivan-ottinger: No change needed — `prepublishOnly` does run for `npm pack` as of npm 7+. From the [npm docs](https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-scripts): "*prepublishOnly*: Runs BEFORE the package is prepared and packed, ONLY on `npm publish`." However, in practice since npm 7, both `npm publish` and `npm pack` trigger `prepare` and `prepublishOnly`. Verified locally — running `npm pack` from `apps/cli/` does trigger `prepublishOnly` and builds the npm dist.

apps/cli/scripts/postinstall-npm.mjs
```diff
@@ -0,0 +1,34 @@
...
+import { fileURLToPath } from 'url';
+
+const scriptDir = dirname( fileURLToPath( import.meta.url ) );
+const packageDir = resolve( scriptDir, '..' );
+const nodeModulesPath = join( packageDir, 'node_modules' );
+
+// Skip in workspace context where deps are hoisted to the root
+if ( ! existsSync( nodeModulesPath ) ) {
+	process.exit( 0 );
+}
```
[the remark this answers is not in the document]
ivan-ottinger: The detection is correct for our use case. The check looks for `node_modules` relative to *the package itself* (`packageDir`), not the consumer's root.

- **In our monorepo**: `apps/cli/node_modules` doesn't exist (deps are hoisted to root) → script correctly exits early, avoiding running `patch-package` against hoisted deps that may not need patching.
- **For npm consumers**: The package lives at `<project>/node_modules/@automattic/wp-studio/`. Its own dependencies (like `pm2`, `pm2-axon`) are installed there by npm — either nested under `node_modules/@automattic/wp-studio/node_modules/` or hoisted. Since `patch-package` is a declared dependency, npm will install it, and `npx --no-install patch-package --patch-dir ./patches` runs with `cwd` set to the package directory, where it can resolve the patched packages from the nearest `node_modules`.

The key insight: even when npm hoists deps to the consumer's top-level `node_modules`, the *package directory itself* still exists at `node_modules/@automattic/wp-studio/`, and `patch-package` walks up from `cwd` to find the target packages.

apps/cli/package.json
```diff
@@ -1,14 +1,33 @@
 {
-	"name": "studio-cli",
+	"name": "@automattic/wp-studio",
 	"author": "Automattic Inc.",
-	"private": true,
+	"version": "1.7.5",
```
ivan-ottinger: Based on the discussion in STU-1386 I set the version to `1.7.5` - to match the current Studio app version.

Before merging this PR I could test the publishing it to npm as `@automattic/wp-studio@1.7.5-beta1`. This is because if we published as `1.7.5` already and then made further changes to this PR, we would need to bump the version number in order to publish the update to npm.

Cleaner approach could just to wait for this PR to get merged and then publish to npm as `1.7.5` right away. Then, all the next changes we are working on could get published as `1.7.6-beta1` etc... With the final release as `1.7.6` - together with the next Studio app release.

What do you think?

fredrikekelund: > Before merging this PR I could test the publishing it to npm as @automattic/wp-studio@1.7.5-beta1

Good idea 👍 Makes sense to not publish `1.7.6-beta1` until the actual next beta is ready

ivan-ottinger: One more approach I am considering - to test everything:

1. Merge this PR (version `1.7.5`).
2. Manually publish `1.7.5` to npm to verify credentials and package work.
3. Create a separate PR for the the CI workflow (I haven't started working on it yet).
4. Bump version to `1.7.6-beta1` in that PR.
5. Test the workflow with manual dispatch. Which will publish `1.7.6-beta1`.
6. Merge the CI workflow PR.
7. When the next Studio release (`1.7.6`) is ready, publish as latest.

apps/cli/package.json
```diff
@@ -1,14 +1,33 @@
...
 	"productName": "Studio CLI",
 	"description": "WordPress Studio CLI",
 	"license": "GPL-2.0-or-later",
-	"main": "index.js",
+	"bin": {
+		"studio": "./dist/cli/main.js"
+	},
+	"files": [
+		"dist/cli/*.js",
+		"dist/cli/*.map",
```
fredrikekelund:
```suggestion
```
Let's not include the source map. The code is open source, so if users are interested in the source code, they can just look up the repo.

We also don't have any Sentry error tracking in the CLI that'd warrant source maps for generating stack traces.

apps/cli/package.json
```diff
@@ -17,24 +36,28 @@
...
 	},
 	"scripts": {
 		"build": "vite build --config ./vite.config.ts",
+		"build:npm": "vite build --config ./vite.config.npm.ts",
 		"install:bundle": "npm install --omit=dev --no-package-lock --no-progress --install-links --no-workspaces && patch-package && node ../../scripts/remove-fs-ext-other-platform-binaries.mjs",
 		"package": "npm run install:bundle && npm run build",
-		"watch": "vite build --config ./vite.config.ts --watch"
+		"watch": "vite build --config ./vite.config.ts --watch",
+		"prepublishOnly": "npm run build:npm",
+		"postinstall": "node ./scripts/postinstall-npm.mjs"
```
fredrikekelund:
```suggestion
		"postinstall": "patch-package"
```
No need to keep a separate script file now that the only thing we do on `postinstall` is to run `patch-package`.

apps/cli/scripts/postinstall-npm.mjs
fredrikekelund, line 1: Let's remove this file. It only runs `patch-package` now, which is easier to do directly from the `postinstall` script in package.json.

apps/cli/package.json
```diff
@@ -1,14 +1,33 @@
...
-	"main": "index.js",
+	"bin": {
+		"studio": "./dist/cli/main.js"
+	},
+	"files": [
+		"dist/cli/*.js",
+		"dist/cli/*.map",
+		"dist/locales/",
+		"patches/",
+		"scripts/"
```
fredrikekelund, line 14:
```suggestion
```

apps/cli/vite.config.base.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const yargsLocalesCopyPlugin = viteStaticCopy( {
+	targets: [
+		{
+			src: normalizePath( join( yargsLocalesPath, '*' ) ),
+			dest: '../locales',
+		},
+	],
+} );
+
```
fredrikekelund:
```suggestion
import { resolve } from 'path';
import { defineConfig } from 'vite';

export const nodeBuiltinExternals = [
	/^node:/,
	/^(path|fs|os|child_process|crypto|http|https|http2|url|querystring|stream|util|events|buffer|assert|net|tty|readline|zlib|constants|tls|domain|dns)$/,
	'fs/promises',
	'dns/promises',
];

```
This is something I should have fixed in https://github.com/Automattic/studio/pull/2685, but I overlooked it. 

In that PR, I marked `yargs` as an external package to fix the locale lookup issue, meaning we no longer need to also copy that directory to `dist`

apps/cli/package.json
```diff
@@ -1,14 +1,33 @@
...
 	"description": "WordPress Studio CLI",
 	"license": "GPL-2.0-or-later",
-	"main": "index.js",
+	"bin": {
+		"studio": "./dist/cli/main.js"
+	},
+	"files": [
+		"dist/cli/*.js",
+		"dist/cli/*.map",
+		"dist/locales/",
```
fredrikekelund:
```suggestion
```
See my comment in `apps/cli/vite.config.base.ts` about not needing to copy this directory anymore.

apps/cli/vite.config.base.ts
```diff
@@ -0,0 +1,61 @@
...
+export const baseConfig: UserConfig = {
```
fredrikekelund:
```suggestion
export const baseConfig = defineConfig( {
```
Kind of a nitpick, but `defineConfig` works fine with `mergeConfig`, and this is the API we already use elsewhere to define Vite configs.

apps/cli/vite.config.npm.ts
```diff
@@ -0,0 +1,52 @@
...
+					// Bundle the blueprint-schema-validator subpath (not exported by the package)
+					if ( id.includes( 'blueprint-schema-validator' ) ) {
+						return false;
+					}
+
+					// Externalize @php-wasm/* and @wp-playground/* (transitive deps resolved by npm)
+					if ( /^@php-wasm\//.test( id ) || /^@wp-playground\//.test( id ) ) {
+						return true;
+					}
+
```
fredrikekelund:
```suggestion
```
I may be wrong, but this looks redundant to me.

apps/cli/vite.config.npm.ts
```diff
@@ -0,0 +1,52 @@
...
+					// Externalize all declared runtime dependencies
+					return externalDeps.some( ( dep ) => id === dep || id.startsWith( dep + '/' ) );
+				},
+			},
+		},
+		define: {
+			__STUDIO_CLI_VERSION__: JSON.stringify( packageVersion ),
+		},
```
fredrikekelund, line 19: This could be part of `apps/cli/vite.config.base.ts`, right?

apps/cli/package.json
```diff
@@ -1,14 +1,33 @@
...
+	},
+	"publishConfig": {
+		"access": "restricted"
+	},
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/Automattic/studio.git",
+		"directory": "apps/cli"
+	},
 	"dependencies": {
```
fredrikekelund, line 24: A few notes on this section:

1. We aren't using `@wp-playground/common` directly anywhere. Let's go ahead and remove it.
2. We haven't included `zod` here. Let's do that.
3. We haven't included `@wordpress/i18n` here. Let's do that.

apps/cli/package.json
fredrikekelund: That also works 👍 The one thing I'd say is that we can use a different version than `1.7.6-beta1` unless we're truly publishing `1.7.6-beta1`. `1.7.6-alpha1`, for example.

apps/cli/vite.config.npm.ts
ivan-ottinger: I checked this and it catches transitive dependencies like `@php-wasm/node-8-5` or `@php-wasm/node-8-3`.

Without the check, Vite tries to bundle them and fails. I updated the comment to clarify why it's needed.

apps/cli/package.json
ivan-ottinger: Good point. Removed in 07ae336f, also disabled sourcemap generation in the npm build config.

apps/cli/package.json
ivan-ottinger: That sounds good, we can use `1.7.6-alpha1` - so it won't clash with proper `beta1`, etc. I am also curious if/how we would be able to utilize the new release process for the npm package as well. But we can look into that separately.

apps/cli/package.json
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2697/commits/bccdb50ffd868490359de3f2988ccf6f0c917c17.

apps/cli/scripts/postinstall-npm.mjs
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2697/commits/bccdb50ffd868490359de3f2988ccf6f0c917c17.

apps/cli/package.json
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2697/commits/bccdb50ffd868490359de3f2988ccf6f0c917c17.

apps/cli/package.json
ivan-ottinger: Sounds good. Addresed in https://github.com/Automattic/studio/pull/2697/commits/8455b4dc460f25d0e6989a7ec5c1005855bb700e.

apps/cli/vite.config.base.ts
ivan-ottinger: 👍🏼 https://github.com/Automattic/studio/pull/2697/commits/e55cc43545da31187962764b43cff8bf8b522916

apps/cli/vite.config.npm.ts
ivan-ottinger: Yes, that's a good idea. Moved it in https://github.com/Automattic/studio/commit/2da2890d.

apps/cli/vite.config.base.ts
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2697/changes/bdce54e4d39d9bc0a28f55fb8046ef61fa03cdec.

apps/cli/package.json
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2697/changes/bdce54e4d39d9bc0a28f55fb8046ef61fa03cdec.

apps/cli/vite.config.npm.ts
fredrikekelund: Those are second-level dependencies (i.e., we don't import them directly). Do we really need to mark second-level dependencies as external?

apps/cli/package.json
fredrikekelund: I'm positive we could (and indeed should), but it's fine to publish the first version/-s manually. We need to tweak all the knobs, anyway

fredrikekelund:
Friendly reminder that it's fine to do regular merge commits instead of rebasing the branch, @ivan-ottinger 🙂

ivan-ottinger:
> Friendly reminder that it's fine to do regular merge commits instead of rebasing the branch, @ivan-ottinger 🙂

Thanks. Yes, I already remember it for https://github.com/Automattic/studio/pull/2697/commits/e060e4069d9c4e817359107f0e093fb22baa0a59. 🙂 

Just the rebase habit is so automatic, also considering other codebases we work with.

apps/cli/scripts/postinstall-npm.mjs
ivan-ottinger: At the end I have restored the postinstall script (in https://github.com/Automattic/studio/pull/2697/commits/45c5bb1b09465654f2cc768058f3e61866842532) which detects the monorepo workspace context and skips patching there, so it only runs for standalone npm consumers. This keeps patches working for npm installs without breaking CI.

I tried to find a better solution, but cannot think of anything at the moment. Do you have any ideas?

ivan-ottinger: One additional change: https://github.com/Automattic/studio/pull/2697/commits/b6771029048295c559a5d580c20ee483a1371432.

apps/cli/vite.config.npm.ts
ivan-ottinger: Without the regex, the build fails with:

```
     ../../node_modules/@php-wasm/node-8-5/jspi/php_8_5.js (4:9): "createRequire" is
     not exported by "__vite-browser-external", imported by
     "../../node_modules/@php-wasm/node-8-5/jspi/php_8_5.js".
     2: // We need to explicitly create a require function to avoid errors when
     running
         at getRollupError (file:///Users/ivanottinger/a8c/studio/node_modules/rollup
     /dist/es/shared/parseAst.js:402:41)
         at error (file:///Users/ivanottinger/a8c/studio/node_modules/rollup/dist/es/
     shared/parseAst.js:398:42)
         at Module.error (file:///Users/ivanottinger/a8c/studio/node_modules/rollup/d
     ist/es/shared/node-entry.js:17040:16)
```

It still once to include them.

sejas (approved):
I followed the testing steps and I was able to run `npm run cli:build` and `npm run cli:build:npm`, and `node apps/cli/dist/cli/main.js site list` worked as expected.

I got a permissions error when uninstalling the CLI from the settings modal, so I had to run `sudo  rm /usr/local/bin/studio` to uninstall it manually.

After that I was able to instal the CLI globally:

Image

and run:

Image

I also tried running it using `npx @automattic/wp-studio@latest site list`:

Image


Here is the error I encountered when I tried uninstalling the CLI, but it’s not related to this PR, and it may be due to my environment.:

Image

PR #2511, opened 2026-02-01
Add agent suite (chat UI, ACP integration, skills, sync CLI)

Jameswlepage:
Summary
- Introduce the agent chat UI and ACP-backed workflows inside Studio.
- Add agent skills discovery/install and instruction management.
- Expand CLI sync workflows (connect/pull/push/status) and related UX updates.

Key features
- Agent chat experience with model selection, tool results, and UI polish.
- ACP session management, agent detection, and permissions flows.
- Agent skills discovery/install and instruction templating.
- Sync CLI workflows for connecting sites and pushing/pulling content.

Security & safety
- Guard ACP file access with path traversal checks.
- Prompt on agent permission requests.
- Restrict agent server CORS to trusted origins.

Testing
- npm run lint
- npm run test

Notes
- Tests emit React act(...) warnings in assistant UI tests.
- jsdom CSS parse warnings from @wordpress/ui styles.
- npm warns about engine/config mismatch (node/npm versions).

PR #2698, opened 2026-03-03
Studio: Add AGENTS.md management

katinthehatsite:
## Related issues

- Fixes STU-1333

## How AI was used in this PR

I used it to adapt the base from https://github.com/Automattic/studio/pull/2511/changes#diff-c69fa338d4e3c20084125a9fa7186cf40116a7460210922cfd4b4021935f4c18

## Proposed Changes

This PR adds an option to manage `AI Settings` from the AI Assistant tab:

Screenshot 2026-03-05 at 4 23 44 PM

Screenshot 2026-03-05 at 4 23 37 PM

## Testing Instructions

* Activate `Enable Agent Suite` feature flag from the topbar from `Electron > Feature Flags`
* Create a Studo site
* Navigate to the `Assistant` tab
* Click on the three dots at the end of the input field
* Observe that you can see the `AI Settings` option
* Open the modal and confirm that you see the `AGENTS.md` and the `Install` button
* Click on the `Install` button
* Click on `Open` and confirm it opens the file
* Try deleting the file
* Observe that now you see the option to `Install` the file again
* Observe that you can also reinstall the default `AGENTS.md`

## Pre-merge Checklist


epeicher (approved):
Thanks for the changes @katinthehatsite, they look great! I have tested the new AI settings, and I have not found any big issues. When I click on `Open` the `Agents.md` file, it asks for permissions twice, please see video below:

https://github.com/user-attachments/assets/f75e891c-5a64-4282-80f0-4a2e761e0814

But that could be investigated as a follow-up as it's a minor detail. Apart from that, everything works as expected. I can see the file, I can edit it using my IDE, and then reinstall the original version. When deleting it from the folder, I see the Reinstall option working as expected.  `Install All` works as expected. Changes also LGTM! 

Image

apps/studio/src/components/site-content-tabs.tsx
```diff
@@ -1,7 +1,6 @@
 import { TabPanel } from '@wordpress/components';
 import { useI18n } from '@wordpress/react-i18n';
 import { useEffect, useRef, useState } from 'react';
-import { ContentTabAgents } from 'src/components/content-tab-agents';
```
epeicher, line 4: If it's not used, should that file be deleted as well?

katinthehatsite:
Thanks for testing @epeicher ❤️ 

> But that could be investigated as a follow-up as it's a minor detail.

Oh, I saw this as well but was not sure if it is my setup or it happened for everyone. I will open an issue to check this further 👀

katinthehatsite:
Issue added here: https://linear.app/a8c/issue/STU-1385/studio-agents-suite-ensure-the-permissions-to-open-agentsmd-does-not

apps/studio/src/modules/agent-instructions/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+	displayName: string;
+	description: string;
+}
+
+export const INSTRUCTION_FILES: Record< InstructionFileType, InstructionFileConfig > = {
+	agents: {
+		id: 'agents',
+		fileName: AGENTS_MD_FILE_NAME,
+		displayName: AGENTS_MD_FILE_NAME,
+		description: 'Instructions for Codex, Goose, and other AI agents',
```
epeicher, line 17: Should this `description` string be i18n'd?

apps/studio/src/components/ai-settings-modal.tsx
```diff
@@ -0,0 +1,201 @@
...
+									{ status.isOutdated && (
+										<span className="block mt-1 text-orange-600">
+											{ __(
+												'A newer version is available. Reinstall to get the latest commands.'
+											) }
+										</span>
+									) }
+								</div>
+							</div>
+							<div className={ cx( 'flex items-center gap-2 flex-shrink-0' ) }>
```
epeicher: This `cx` seems to be unnecessary

epeicher (commented):
I have added some minor comments to my review, but I don't see any of them as blockers

apps/studio/src/components/site-content-tabs.tsx
katinthehatsite: I think we can remove it; I initially left it because there was an explicit request to add it in the task where I added the feature flag but I think it makes much more sense to work with the existing tab instead of rebuilding the whole UI considering that the chat interface looks very similar in the new design so that's when I changed the approach 😅 

I will remove it before merging 👍

sejas (commented):
@katinthehatsite , thanks for creating the UI and moving the Agents installation logic to the UI.
This is a great improvement 🥳 and the AI settings will be really useful for other features like Skills settings.

Let's merge this PR, and create a follow-ups to improve these:
- Open doesn't use the preferred  IDE, this is an existing bug  in `openFileInIDE` function and I'll create the issue to fix it, so no need to take any action in your side.
- What do you think should happen if users remove `<!-- Studio Instructions Version: 20250131.1 -->` or write Agents.md from scratch? Should we warn them that they’ll lose their changes? I’ll create an issue to improve this behavior.
- We mention re-install, but the button says update. Let's use the same term in both places. TIL about Goose, we can improve the text later.
- Do we need an update all button for the Agents file?

**Installed**

Image

**Modified**

If the user modified the Agents, it's not detected by Studio, and if the user removes `<!-- Studio Instructions Version: 20250131.1 -->` it's detected like outdated.

Image

katinthehatsite:
> Do we need an update all button for the Agents file?

I left it thinking there would be more Agents.md files but we can remove for now. There was also `Reinstall all` button previously but I removed that one as well.

katinthehatsite:
> What do you think should happen if users remove  or write Agents.md from scratch? Should we warn them that they’ll lose their changes? I’ll create an issue to improve this behavior.

In my opinion and from what I researched, we should leave the file as is because it is the user's preference and setting which we should not overwrite. I think the rule of thumb should be - if the user modified it, don't touch it. What do you think? @sejas

wojtekn:
Adding to what @katinthehatsite said above, I think we can have helpers in Studio that let users add those files to a project - e.g., "Install" button which adds them, but Studio shouldn't automagically try to update, or re-add them if the user modifies or removes them later.

apps/studio/src/components/site-content-tabs.tsx
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2698/changes/bb98b0241ad1e7c70c3bb3f9dca2494c8ab111d4

apps/studio/src/modules/agent-instructions/constants.ts
katinthehatsite: added in https://github.com/Automattic/studio/pull/2698/changes/b10a2bfc2516c07bc5fd7ada6633c1d66dcda9ce

apps/studio/src/components/ai-settings-modal.tsx
katinthehatsite: Added in https://github.com/Automattic/studio/pull/2698/changes/15d8c30891c6950c3ce0fcc3ba6ec4d176597faf

katinthehatsite:
> Do we need an update all button for the Agents file?

I removed it in https://github.com/Automattic/studio/pull/2698/changes/2e1e771c775247359aeef5e334b0dad002451f49

apps/cli/commands/site/create.ts
fredrikekelund, line 1: I haven't followed the previous discussion about this, but if the goal is to make it optional to create an `AGENTS.md` when creating a site, would it not make sense to add an `--agentsmd` flag (or whatever name) to the CLI? 

This definitely seems like something that'll be relevant for CLI-only users, too.

katinthehatsite: > I haven't followed the previous discussion about this, but if the goal is to make it optional to create an AGENTS.md

I am not sure we have decided on this yet. Initially, we planned to always install it but then we decided to not make it to be the default, for now at least. And only add from the UI for now. I am not sure what is the final direction we will be taking here so I will let @sejas comment on it.

nightnei (approved):
LGTM and works as expected 👍

Screenshot 2026-03-10 at 12 33 37
Screenshot 2026-03-10 at 12 35 00
Screenshot 2026-03-10 at 12 35 37

apps/cli/commands/site/create.ts
fredrikekelund: It makes sense for the UI to have more options for managing this file, but my point is mainly that `AGENTS.md` will be just as useful for sites created directly with the CLI, and I noticed that this PR removed the `writeAgentsMd` from the `site create` command.

sejas: I created a separate issue to bring the AGENTS.md to CLI STU-1393 . In any case we don't want to ship AGENTS.md yet. At least it should be behind a feature flag. Let's continue with the solution implemented in this PR and we can iterate later.

apps/studio/src/ipc-handlers.ts
```diff
@@ -126,6 +135,34 @@ export {
...
 } from 'src/modules/user-settings/lib/ipc-handlers';
 
+async function getAgentInstructionsSitePath( siteId: string ): Promise< string > {
+	const userData = await loadUserData();
+	const site = userData.sites.find( ( s ) => s.id === siteId );
+	if ( ! site ) {
+		throw new Error( `Site not found: ${ siteId }` );
+	}
+	return site.path;
+}
```
sejas: Do we need this function? It seems to return the site path. I wonder if we have an equivalent function already, or if we can pass the site path directly to the IPC function.
If we really need it, we can consider renaming it to a more generic name. We could consider using `SiteServer.get` instead of loading loadUserData.

katinthehatsite: I am still adjusting a couple of things. I will take a look 👀

katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2698/changes/bed8cc28a605531c5e6ccee20ba87555e107d907

katinthehatsite:
@sejas  I made a small adjustment regarding how the UI looks when the `AGENTS.MD` was modified by the user. I will now be merging this PR and we can make further adjustments after:

Screenshot 2026-03-10 at 4 36 39 PM

sejas (approved):
Thanks for making the adjustments!

PR #2700, opened 2026-03-03
Add Studio CLI skill for agent-assisted site management

epeicher:
## Related issues

- Fixes STU-1336

## How AI was used in this PR

AI assisted in drafting the skill file content and structuring the implementation. I reviewed and tested manually.

## Proposed Changes

- Add `apps/cli/lib/skill-md.ts` — a module that writes a `SKILL.md` file to `.agents/skills/studio-cli/` inside each Studio site directory, teaching AI agents all major Studio CLI commands (site management, authentication, preview sites, WP-CLI)
- Create a `.claude/skills/studio-cli` symlink pointing to the `.agents/skills/studio-cli/` directory for Claude Code skill discovery
- Install the skill automatically during site creation (alongside `AGENTS.md`)
- Skip writing if files already exist to preserve user customisations
- Add comprehensive tests for the skill-md module

> **Note:** This feature is gated behind the `ENABLE_AGENT_SUITE` feature flag (disabled by default). The flag must be set to `true` for SKILL.md to be installed during site creation.

## Testing Instructions

- Enable the feature flag: set `ENABLE_AGENT_SUITE=true` in your environment
- Create a new site via CLI and verify `.agents/skills/studio-cli/SKILL.md` and `.claude/skills/studio-cli` symlink are created in the site directory
- Create a new site by using Studio UI and verify that the skills are created too
- Open a Terminal, open `Claude` or `Codex` and prompt something like `Create a preview site` or `Can you create a home page with a hero image of a nice paradisiacal beach?`
- Verify that the `studio` CLI is used as expected
- Create the same site again and verify existing files are not overwritten
- Without the feature flag enabled, verify SKILL.md is **not** installed during site creation

## Pre-merge Checklist


apps/cli/lib/skill-md.ts
```diff
@@ -0,0 +1,174 @@
...
+const SKILL_MD_FILENAME = 'SKILL.md';
+const CLAUDE_SKILLS_DIR = path.join( '.claude', 'skills' );
+const SYMLINK_NAME = 'studio-cli';
+
+const SKILL_MD_TEMPLATE = `---
+name: studio-cli
+description: Use the Studio CLI to manage local WordPress sites, authentication, and preview sites.
+---
+
+# Studio CLI
```
katinthehatsite, line 14: I am wondering if we should reference `AGENTS.MD` in this file as well to let the skills know that it exists. Could be something like one line at the bottom. what do you think?

katinthehatsite (commented):
I just wanted to mention briefly that I am still testing this but for some reason I am not seeing the skills file created after adding the site with CLI. I am checking if this is perhaps an issue with my setup.

katinthehatsite (approved):
Everything works now for me, there was a problem in my setup before 😅 

The changes look good to me, I added one minor comment but it is non-blocking.

apps/cli/lib/skill-md.ts
epeicher: Because `Agents.md` is read when starting any conversation with the agents (checked for [Claude](https://code.claude.com/docs/en/best-practices#:~:text=CLAUDE.md%20is%20a%20special%20file%20that%20Claude%20reads%20at%20the%20start%20of%20every%20conversation) and [Codex](https://developers.openai.com/codex/guides/agents-md/#:~:text=Codex%20reads%20AGENTS.md%20files%20before%20doing%20any%20work.)) I think it is redundant referencing it from the skill, although this involves trial and error, we can test some use cases and check how it behaves

apps/cli/lib/skill-md.ts
```diff
@@ -0,0 +1,174 @@
...
+`;
+
+/**
+ * Writes the default SKILL.md file to `.agents/skills/studio-cli/SKILL.md` within the
+ * site directory and creates a `.claude/skills/studio-cli` symlink pointing to it.
+ *
+ * Returns `false` if both the SKILL.md and the symlink already exist (nothing was written).
+ * Returns `true` if any files were created.
+ */
+export async function writeSkillMd( sitePath: string ): Promise< boolean > {
```
nightnei: @epeicher  I would propose either fo with Promise<void> and just use `return` inside or adjust the returned value for AGENTS.md [here](https://github.com/Automattic/studio/pull/2684/changes#diff-fb5d9c48b172625307d29babf818e4c8f43ef0b984e23b64d7bc9f7f3a562ffeR116), for consistency.

cc @katinthehatsite you worked on AGENTS.md integration, so you have better context than me and maybe you could also share your opinion about which way is better - either boolean or void.

nightnei (approved):
I decided to test via Studio UI - everything works great and I see both AGENTS and SKILL. Also tested with borh codex and claude. LGTM 👍
Just left one comment to keep consistency with AGENTS.md implementation

apps/cli/lib/skill-md.ts
epeicher: That's a good point, thanks @nightnei! I will change it to a `Promise<void>` for consistency. We can use the boolean if we are going to handle the return value for a specific requirement 👍

apps/cli/lib/skill-md.ts
nightnei: Agree with Roberto. 
Regarding cross-referencing - we could consider some references from AGENTS.md to the skills, AFAIK it's the right way, but not opposite.

.claude/skills
```diff
@@ -0,0 +1 @@
+../.agent/skills
```
sejas: I think this is not related to this PR.

```suggestion
```

apps/cli/commands/site/create.ts
```diff
@@ -226,6 +227,15 @@ export async function runCommand(
 			);
 		}
 
+		try {
+			await writeSkillMd( sitePath );
```
sejas: Let’s add it behind the feature flag. We might move this installation step behind a UI action in the near future.

sejas (approved):
Thanks for creating this PR. I think it's a great start.
Let's add it behind a feature flag to avoid modifying new sites.
I'm still deciding the best path forward, but probably iterating and testing real use cases is the best approach.
We might need a UI to enable it in existing sites.

apps/studio/src/modules/cli/lib/execute-command.ts
```diff
@@ -139,6 +139,7 @@ export function executeCliCommand(
 		stdio,
 		execPath: getBundledNodeBinaryPath(),
 		execArgv: [ '--experimental-wasm-jspi' ],
+		env: { ...process.env },
```
epeicher, line 142: Added to access the environment variables of the feature flags

PR #2701, opened 2026-03-03
Read site data from CLI instead of appdata

bcotrim:
## Related issues

- Related to STU-1350

## How AI was used in this PR

AI assisted in implementation and iteration. Code was reviewed and tested manually.

## Proposed Changes

- **Site data comes from the CLI, not appdata**: IPC handlers now read site details from the in-memory server registry (populated by the CLI) instead of reading directly from the appdata JSON file. Studio-only fields like sort order and theme details are still hydrated from appdata.
- **Startup fetches sites from CLI**: On launch, Studio asks the CLI for all known sites to populate its registry, replacing the old pattern of reading from appdata and lazily registering servers.
- **Richer CLI JSON output**: `site list --format json` now returns full site details (validated through the shared schema) instead of just table-oriented fields.
- **Simpler site lifecycle**: Creating, starting, stopping, and updating sites no longer re-read appdata to reconstruct state. Instead, the CLI events subscriber handles state updates consistently.
- **Site copy uses the standard creation flow** instead of manually constructing details, allocating ports, and writing to appdata.
- **Site deletion cleans up Studio-only appdata** (sort order, theme details) on delete.
- **Boot order fix**: Sites are fetched and the events subscriber starts before the data watcher, preventing a race where the UI could render before sites are loaded.

## Testing Instructions

1. **Fresh startup** — Verify all existing sites appear in the sidebar
2. **Create a new site** — Site should appear and be functional
3. **Start/stop a site** — Should transition states correctly
4. **Update site settings** — Name, PHP version, domain changes should persist
5. **Delete a site** — Should be removed from UI and appdata
6. **Copy a site** — Should create a working copy with theme details preserved

## Pre-merge Checklist


apps/cli/commands/site/list.ts
```diff
@@ -8,35 +9,51 @@ import { getColumnWidths, getPrettyPath } from 'cli/lib/utils';
...
 			status,
 			name: site.name,
 			path: getPrettyPath( site.path ),
 			url,
 		} );
+
+		jsonEntries.push( {
+			...siteDetailsSchema.parse( { ...site, url } ),
+			running,
+		} );
```
fredrikekelund, line 48:
```suggestion
		jsonEntries.push( {
			...site,
			url,
			running,
		} );
```
No need to use the schema here. TS is already statically analyzing the type.

apps/studio/src/site-server.ts
```diff
@@ -108,10 +102,56 @@ export class SiteServer {
...
+				}
+			);
+
+			for ( const site of sites ) {
+				if ( ! SiteServer.get( site.id ) ) {
+					SiteServer.register( { ...site } );
+				}
+			}
+		} catch ( error ) {
+			console.error( 'Failed to fetch sites from CLI:', error );
```
fredrikekelund, line 151: I don't have a great answer for how, but I really think we should try to improve the error-handling story here. 

If this operation fails, the app effectively becomes broken.

apps/studio/src/site-server.ts
```diff
@@ -108,10 +102,56 @@ export class SiteServer {
...
+
 	static isDeleted( id: string ) {
 		return deletedServers.includes( id );
 	}
 
+	private static siteListJsonSchema = z.array(
+		siteDetailsSchema.extend( {
+			running: z.boolean(),
+		} )
+	);
```
fredrikekelund: Let's move this to a new schema/type pair in `tools/common/lib/site-events.ts`. So far, we've only added one prop to `siteDetailsSchema`, but we should really share the entire definition between the CLI and the app for maximum safety.

apps/studio/src/site-server.ts
```diff
@@ -108,10 +102,56 @@ export class SiteServer {
...
+					} );
+
+					emitter.on( 'failure', ( { error } ) => reject( error ) );
+					emitter.on( 'error', ( { error } ) => reject( error ) );
+				}
+			);
+
+			for ( const site of sites ) {
+				if ( ! SiteServer.get( site.id ) ) {
+					SiteServer.register( { ...site } );
```
fredrikekelund:
```suggestion
					SiteServer.register( site );
```
Nitpick, but I don't see a need to recreate the `site` object here.

apps/studio/src/site-server.ts
```diff
@@ -137,46 +177,14 @@ export class SiteServer {
...
-			}
-
-			return { server, details: siteDetails };
-		} finally {
-			server.hasOngoingOperation = false;
+		if ( result.running ) {
+			server.details.running = true;
 		}
+
+		return { server, details: server.details };
```
fredrikekelund, line 187: Nice 👍

apps/studio/src/site-server.ts
```diff
@@ -198,21 +206,12 @@ export class SiteServer {
...
-			this.details.port = freshSiteData.port;
-		}
-
 		const url = getAbsoluteUrl( this.details );
-
 		this.details = {
 			...this.details,
 			url,
 			running: true,
 			autoStart: true,
```
fredrikekelund: Open question: do we still need to assign these props, despite the increased reliance on site events from the CLI?

apps/studio/src/modules/cli/lib/cli-events-subscriber.ts
```diff
@@ -35,23 +35,22 @@ const handleSiteEvent = sequential( async ( event: SiteEvent ): Promise< void >
...
 	}
 
-	// For UPDATED events, only update if the site already exists
+	// For UPDATED events, update existing server or register new one (e.g., during startup)
 	const server = SiteServer.get( siteId ) ?? SiteServer.getByPath( site.path );
 	if ( ! server ) {
+		SiteServer.register( siteDetailsToServerDetails( site, running ) );
+		void sendIpcEventToRenderer( 'site-event', event );
 		return;
 	}
```
fredrikekelund, line 54: Is this correct? It seems to me like startup server registration is handled by `apps/studio/src/site-server.ts`

apps/studio/src/ipc-handlers.ts
```diff
@@ -635,33 +606,26 @@ export async function copySite(
...
-		const userData = await loadUserData();
-		userData.sites.push( newSiteDetails );
-		await saveUserData( userData );
-	} finally {
-		await unlockAppdata();
+	// Persist themeDetails to appdata (Studio-only data)
+	if ( sourceSite.themeDetails ) {
+		server.details.themeDetails = sourceSite.themeDetails;
+		await server.persistThemeDetails();
 	}
```
fredrikekelund, line 613:
```suggestion
```
This is redundant, right?

apps/studio/src/ipc-handlers.ts
```diff
@@ -635,33 +606,26 @@ export async function copySite(
...
 		adminUsername: sourceSite.adminUsername,
-		adminPassword: sourceSite.adminPassword,
+		adminPassword: sourceSite.adminPassword
+			? decodePassword( sourceSite.adminPassword )
+			: undefined,
 		adminEmail: sourceSite.adminEmail,
-		themeDetails: sourceSite.themeDetails,
-	};
+		noStart: true,
+	} );
```
fredrikekelund, line 607: Nice 👍

apps/studio/src/ipc-handlers.ts
```diff
@@ -602,6 +561,19 @@ export async function deleteSite( event: IpcMainInvokeEvent, id: string, deleteF
...
+		await lockAppdata();
+		const userData = await loadUserData();
+		const siteIndex = userData.sites.findIndex( ( s ) => s.id === id );
+		if ( siteIndex !== -1 ) {
+			userData.sites.splice( siteIndex, 1 );
+			await saveUserData( userData );
+		}
+	} finally {
+		await unlockAppdata();
+	}
```
fredrikekelund: Maybe this is something we'll need in https://github.com/Automattic/studio/pull/2731, but is this truly needed here..? Seems to be reimplementing parts of `apps/cli/commands/site/delete.ts`

fredrikekelund (commented):
I left some questions and some feedback. Overall, I agree with your vision that this makes for a cleaner architecture and mental model, @bcotrim 👍

The biggest question mark to me is how we effectively deal with errors in `SiteServer::fetchAll`. Doesn't have to block this PR, though.

I'll do some more testing to verify this change, but figured I'd leave my initial feedback first.

apps/studio/src/site-server.ts
bcotrim: I agree it's something we should improve, but it's a preexisting condition, reading the appdata file could also fail and break the app. There's the argument that running the CLI could be more prone to errors, but both share the same issue.
I would propose revisiting it in a follow-up PR, if we re-use this approach for other data points we should add a similar fallback mechanism

fredrikekelund: Agreed this is something we can revisit 👍

apps/studio/src/site-server.ts
```diff
@@ -108,10 +102,56 @@ export class SiteServer {
...
+					} );
+
+					emitter.on( 'success', ( { result } ) => {
+						try {
+							const parsed = JSON.parse( result.stdout );
+							resolve( SiteServer.siteListJsonSchema.parse( parsed ) );
+						} catch ( error ) {
+							reject( error );
+						}
+					} );
```
fredrikekelund: Relying on stdout isn't wrong or bad, but it's riskier than something like `logger.reportKeyValuePair`, because anything can output to stdout. A deprecation warning, for example, could break the entire site-loading logic. 

There's also the fact that passing `output: 'capture'` leads to the output being logged to the terminal. Logging the entire site list is a little noisy, IMO. 

`Logger` was designed around the requirements of the preview site commands, and maybe we need to rethink parts of the API at some point. 

My main point here, though, is that `logger.reportKeyValuePair` seems like a safer fit and will reduce the noise in the terminal.

fredrikekelund (approved):
This PR tests well for me 👍 There are things from my previous review that we should address before landing this, and there's the `logger.reportKeyValuePair` question. Still, I'm approving now to unblock you, @bcotrim

apps/studio/src/ipc-handlers.ts
bcotrim: Since themeDetails is Studio only data and not managed by the CLI, I think we still need it.
Did you notice something I missed?

apps/studio/src/site-server.ts
bcotrim: I think it's a good approach, let's experiment with it

bcotrim:
> This PR tests well for me 👍 There are things from my previous review that we should address before landing this, and there's the logger.reportKeyValuePair question. Still, I'm approving now to unblock you, @bcotrim

Thanks for the review @fredrikekelund 
I think I've addressed all the feedback, so I'm merging this one. Please let me know if there's anything I've missed and we can follow-up in another PR.

PR #2704, opened 2026-03-04
Handle Windows WASM memory allocation errors with user-friendly message

sejas:
## Related issues

- Fixes STU-1369

## Proposed Changes

Studio displayed an error, `WordPress server process exited unexpectedly`, on Windows when a user started or created a site with low available RAM. That behaviour has been improved in Playground CLI 3.1.4, but I think it is worth adding this heuristic to guide the user to stop some sites. This also adds more error strings that I encountered when starting a site on Windows; it was not common, but I saw them about once every 10 tries.

This PR:
- Adds `isWasmMemoryError()` helper function in `ipc-handlers.ts` that detects WASM memory allocation failures through:
- A Windows-specific heuristic: when the server process exits unexpectedly, other sites are already running, and free system memory is below 600MB — this covers cases where Windows doesn't propagate a clear WASM error message
- Simplify both `createSite` and `startServer` catch blocks to use the new helper
- When detected, throw `WASM_ERROR_NOT_ENOUGH_MEMORY` which the renderer already handles by showing a user-friendly "Not enough memory" dialog

**Context**: Windows doesn't overcommit memory like macOS. When multiple WordPress sites are running via PHP WASM, Windows fails to allocate memory for new WASM instances and the server process crashes. Previously, users saw a generic "verify your site's local path directory" error instead of a helpful memory message.

## Testing Instructions

- Start the app with `npm start` on Windows
- Start enough WordPress sites to consume your RAM memory. 6-7 for 8GB machine.
- Attempt to start another site
- Verify that the "Not enough memory" dialog appears instead of the generic error
- On macOS, verify existing behavior is unchanged (the `'Cannot allocate Wasm memory for new instance'` check still works)
- The error might now is hard to reproduce due to the new improvements on Playground CLI 3.1.4

## Pre-merge Checklist


sejas:
Closing this PR as it seems not necessary anymore. I was able to start all my sites in Windows without any error similarly to Mac behaviour.

Screenshot 2026-03-04 at 13 52 45

PR #2705, opened 2026-03-04
Fix removing sites when site folder is already removed

nightnei:
## Related issues
- Fixes STU-1319

## How AI was used in this PR
AI wrote code; I thoroughly reviewed it and did a few more iterations to polish the solution. I considered another solution (p1772624822355439/1772022345.939889-slack-C06DRMD6VPZ), but ultimately decided to go with this mutual  proposition from AI and Fredrik (p1772626864886569/1772022345.939889-slack-C06DRMD6VPZ).

## Proposed Changes
The issue - if user removed site folder manually, then `studio site delete` fails and the site remains in the `studio site list`.
This PR fixes delete command in thsi case.

## Testing Instructions
1. `npm run cli:build`
2. Create a site - `node apps/cli/dist/cli/main.js site create --path=~/Studio/qqq`
3. Remove site folder - `rm -rf ~/Studio/qqq`
4. Remove site via cli - `node apps/cli/dist/cli/main.js site delete --path=~/Studio/qqq --files`
5. Previously we got error. Now - `Site files already removed` is printed and no errors and `node apps/cli/dist/cli/main.js site list` shows that the site was removed from Studio

## Pre-merge Checklist

### DIFF
--- a/tools/common/lib/fs-utils.ts
+++ b/tools/common/lib/fs-utils.ts
@@ -64,8 +64,7 @@ export function isWordPressDirectory( projectPath: string ): boolean {
 // `fs.Stats.dev` signifies the device ID, and `fs.Stats.ino` signifies the inode number
 // that uniquely identifies the file or directory. This approach respects the current file
 // system's case sensitivity.
-// Falls back to string comparison when either path doesn't exist (e.g., deleted directories),
-// using case-insensitive comparison on macOS/Windows.
+// Falls back to string comparison when either path doesn't exist (e.g., deleted directories).
 export function arePathsEqual( path1: string, path2: string ) {
 	try {
 		const stats1 = fs.statSync( path.resolve( path1 ) );
--- a/tools/common/lib/fs-utils.ts
+++ b/tools/common/lib/fs-utils.ts
@@ -73,7 +73,7 @@ export function arePathsEqual( path1: string, path2: string ) {
 
 		return stats1.ino === stats2.ino && stats1.dev === stats2.dev;
 	} catch ( error ) {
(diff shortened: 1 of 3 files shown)


epeicher (approved):
Thanks for fixing this @nightnei! I have tested it, and I can see that the site is removed from the list when the folder have been removed. Changes also LGTM! :shipit: 

|Before | After|
|-------|------|
| Image|Image

tools/common/lib/fs-utils.ts
```diff
@@ -60,18 +60,20 @@ export function isWordPressDirectory( projectPath: string ): boolean {
...
-// Compare paths in a case-insensitive manner. `fs.Stats.dev` signifies the device ID, and
-// `fs.Stats.ino` signifies the inode number that uniquely identifies the file or directory.
-// The benefit of this approach over converting the entire path to lowercase is that it respects
-// the current file system's case sensitivity.
+// Compare paths, preferring inode comparison when both paths exist on disk.
+// `fs.Stats.dev` signifies the device ID, and `fs.Stats.ino` signifies the inode number
+// that uniquely identifies the file or directory. This approach respects the current file
+// system's case sensitivity.
+// Falls back to string comparison when either path doesn't exist (e.g., deleted directories),
+// using case-insensitive comparison on macOS/Windows.
```
```suggestion
// Falls back to string comparison when either path doesn't exist (e.g., deleted directories).
```
The second line looks like a leftover

fredrikekelund (approved):
LGTM! Thanks for tackling this, @nightnei 👍

tools/common/lib/fs-utils.ts
nightnei: Forgot to update 🤦‍♂️ Thanks 👍

PR #2707, opened 2026-03-04
Clear sync state on user logout

gcsecsey:
## Related issues

Fixes STU-166

## How AI was used in this PR

I used AI to research the existing logout flow, then implemented and iterated on the changes using it.

## Proposed Changes

When a user logged out, sync-related Redux state was not cleared, and the stale data remained in memory. If a different user logged in, they could see leftover sync UI state from the previous session.

This PR adds a userLoggedOut Redux action that is dispatched either when logging out explicitly or when an invalid token is detected. Multiple slices and a store listener react to this:

- `syncOperations` slice resets `pullStates` and `pushStates` to empty
- `sync` slice clears the remote file tree cache
- `connectedSites` slice resets modal and selection state
- A store listener cancels active main-process sync operations, stops renderer-side push/pull pollers, and resets authenticated RTK Query caches

## Testing Instructions

- Log in to a WordPress.com account and connect a remote site
- Start a push or pull operation
- Log out
- Log in again with the same user or a different one
- Check that the Sync tab shows the logged out state
- Log in again (same or different account) and check that initial pull/push states are shown

https://github.com/user-attachments/assets/1969f4f1-cb08-481b-ab60-dc2f20e08285

## Pre-merge Checklist


katinthehatsite (commented):
Thanks for making the changes @gcsecsey - nice work!

I found some issues on my end when testing this PR. This is what I saw on my end: I started the pull operation and waited until the downloading backup step. Then, I logged out and noticed that some of the buttons were stuck in the "Importing" state: 

Image

(see the top right corner)

Then, when I log in as a different user, I see the following issue on the import screen:

Image

gcsecsey:
> Thanks for making the changes @gcsecsey - nice work!
> 
> I found some issues on my end when testing this PR. This is what I saw on my end: I started the pull operation and waited until the downloading backup step. Then, I logged out and noticed that some of the buttons were stuck in the "Importing" state:
> 
> Image
> (see the top right corner)
> 
> Then, when I log in as a different user, I see the following issue on the import screen:
> 
> Image

Thanks for the review @katinthehatsite, great catch! This was caused by the import/export also being tracked in the ImportExportProvider context, not just Redux. I set this to also be reset on logout.

I also encountered a stuck state when trying to pull a site that caused a db import issue:

image

I updated the component to treat this state as a pull error as well, to unblock this state. However, I'm not sure if this is the best approach to handle this. I think ideally we'd show the pull operation as completed, and separately also show the db import error in a dialog, similarly to how it's done when importing, right?

apps/studio/src/stores/index.ts
```diff
@@ -184,6 +185,42 @@ startAppListening( {
...
+		PUSH_POLLERS.clear();
+		for ( const controller of PULL_POLLERS.values() ) {
+			controller.abort();
+		}
+		PULL_POLLERS.clear();
+
+		// Reset authenticated RTK Query caches
+		listenerApi.dispatch( connectedSitesApi.util.resetApiState() );
+		listenerApi.dispatch( wpcomSitesApi.util.resetApiState() );
+		listenerApi.dispatch( wpcomApi.util.resetApiState() );
```
fredrikekelund: Note for the future (potentially for https://github.com/Automattic/studio/pull/2778): the queries in these slices should declare the user ID as a dependency, so that RTK-Query automatically invalidates them when the user state changes.

apps/studio/src/stores/index.ts
```diff
@@ -184,6 +185,42 @@ startAppListening( {
...
+
+		// Stop renderer-side pollers
+		for ( const controller of PUSH_POLLERS.values() ) {
+			controller.abort();
+		}
+		PUSH_POLLERS.clear();
+		for ( const controller of PULL_POLLERS.values() ) {
+			controller.abort();
+		}
+		PULL_POLLERS.clear();
```
fredrikekelund:
```suggestion
		for ( const pullState of Object.values( state.syncOperations.pullStates ) ) {
			void store.dispatch(
				syncOperationsThunks.cancelPull( {
					selectedSiteId: pullState.selectedSite.id,
					remoteSiteId: pullState.remoteSiteId,
					displayNotification: false,
				} )
			);
		}

		for ( const pushState of Object.values( state.syncOperations.pushStates ) ) {
			void store.dispatch(
				syncOperationsThunks.cancelPush( {
					selectedSiteId: pushState.selectedSite.id,
					remoteSiteId: pushState.remoteSiteId,
					displayNotification: false,
				} )
			);
		}
```
I recommend using the existing thunks to cancel the sync operations. The fewer ways we trigger state updates, the easier they are to reason about.

We need to consider notifications because they shouldn't be displayed when sync operations are automatically cancelled in this way. That's easy enough to fix, though. It's why I included the `displayNotification` prop here

gcsecsey: Thank you @fredrikekelund, I agree it's simpler to invoke the thunks here. In f816995 I implemented this change and also added the `displayNotification` prop as suggested.

apps/studio/src/stores/index.ts
gcsecsey: Good call! That would be a cleaner approach than manually resetting caches on logout. I’ll leave this as-is for now, and we can address it either as part of #2778 or separately.

katinthehatsite (commented):
Thanks for making the changes @gcsecsey ! I was testing the flow a bit more but now I am getting strange `session expired` messages:

Image

If I do manage to log in, I am getting the errors related to not being logged in while doing the push or pull:

Image

I did not debug the code for this but wanted to point this out early in the review.

gcsecsey:
> Thanks for making the changes @gcsecsey ! I was testing the flow a bit more but now I am getting strange `session expired` messages:
> 
> Image
> If I do manage to log in, I am getting the errors related to not being logged in while doing the push or pull:
> 
> Image
> I did not debug the code for this but wanted to point this out early in the review.

Thanks for testing again @katinthehatsite! I could reproduce this issue, and found that it was caused by the wpcomClient being cleared after the logout event being dispatched, causing refetches to run into the error.

I also found that even though I added the listener earlier for clearing the state, these weren't firing because the listener iterated on the updated (empty) state.

I fixed both of these issues now and the whole pull flow seems to be working well on my end:

https://github.com/user-attachments/assets/caf9b2d5-7c4f-493a-a4ce-9ecf8134c318

I'm still working to fix an issue with cancelling pushes, then I'll ask for another review.

gcsecsey:
@katinthehatsite @fredrikekelund I think this is ready for another review. I fixed these issues for the push flow;

I reverted emptying the connected sites cache on logout to avoid resetting the connected site list when logging back in:

https://github.com/user-attachments/assets/f6abcbba-ceff-48ff-9686-64ef1791d8de

I also updated the pollers to avoid receiving errors when logging out at a later step in the push process, when the changes are being processed on the server:

https://github.com/user-attachments/assets/05762eba-41b9-4ee9-b742-6d4211a67b75

apps/studio/src/components/auth-provider.tsx
```diff
@@ -45,21 +46,23 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
 	const { __ } = useI18n();
 	const isOffline = useOffline();
 
+	const dispatch = useAppDispatch();
 	const authenticate = useCallback( () => getIpcApi().authenticate(), [] );
 
 	const handleInvalidToken = useCallback( async () => {
 		try {
 			void getIpcApi().logRendererMessage( 'info', 'Detected invalid token. Logging out.' );
+			setWpcomClient( undefined );
```
fredrikekelund: WDYT about moving this to the `userLoggedOut` middleware listener in `apps/studio/src/stores/index.ts`?

apps/studio/src/stores/sync/sync-operations-slice.ts
```diff
@@ -297,14 +300,19 @@ const createTypedAsyncThunk = createAsyncThunk.withTypes< {
...
 	'syncOperations/cancelPush',
-	async ( { selectedSiteId, remoteSiteId }: CancelOperationPayload, { dispatch } ) => {
+	async (
+		{ selectedSiteId, remoteSiteId, displayNotification = true }: CancelOperationPayload,
+		{ dispatch }
+	) => {
 		const operationId = generateStateId( selectedSiteId, remoteSiteId );
 		const abortCallback = PUSH_SITE_ABORT_CALLBACKS.get( operationId );
 
+		stopPushPoller( operationId );
```
fredrikekelund, line 431: Maybe it's better to call both `start{Pull,Push}Poller` and `stop{Pull,Push}Poller` in this file… 🤔 I feel like we ended up jumping through some hoops doing it in the middleware functions in `stores/index.ts`. 

Thoughts, @gcsecsey?

fredrikekelund (commented):
I took another look at the Linear issue, and it doesn't appear like these changes fully address the issue.

If I have two users with access to the same WP.com site, initiate a push to the site with my first account, log out in Studio, then log in with my second account (which has also connected the same WP.com site to my local site), then I still see `Push cancelled` in the interface. One step better than seeing an error message, but the UI should really be completely cleared.

apps/studio/src/components/auth-provider.tsx
gcsecsey: It makes sense, that way we don't need maintain the order of these calls. I updated it in [e09fa02](https://github.com/Automattic/studio/pull/2707/commits/e09fa0291f16134ef552448cf8d22422a9b9658c).

katinthehatsite (commented):
I tried testing the changes again but I am still seeing the error with pushing a site when I log in and try to do the push operation:

Image

I tried with a couple of different users but I keep on running into the same error.

apps/studio/src/stores/sync/sync-operations-slice.ts
gcsecsey: I agree that this split is a little awkward. I needed to move the `stop{Pull,Push}` pollers to a separate module to be able to call them from here. I'll try to update this to consolidate it.

gcsecsey: I move these functions to `sync-operations-slice.ts` in d47341d442b347fe379c1bee621ef1c0ae61bcbc 👍

gcsecsey:
> I tried testing the changes again but I am still seeing the error with pushing a site when I log in and try to do the push operation:
> Image
> 
> I tried with a couple of different users but I keep on running into the same error.

I could reproduce this issue, but I could also reproduce it on trunk and found that it's  unrelated to these changes. We're fetching the latest WP-CLI phar and `sqlite-command` release, but the latest sqlite-command requires a newer WP-CLI version than the latest. 

Temporarily, I could resolve this by updating the post-install script to fetch the nightly version:

```diff
diff --git a/scripts/download-wp-server-files.ts b/scripts/download-wp-server-files.ts
index fe9ca6cc..41b07da7 100644
--- a/scripts/download-wp-server-files.ts
+++ b/scripts/download-wp-server-files.ts
@@ -35,7 +35,7 @@ const FILES_TO_DOWNLOAD: FileToDownload[] = [
        {
                name: 'wp-cli',
                description: 'WP-CLI phar file',
-               getUrl: () => 'https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar',
+               getUrl: () => 'https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli-nightly.phar',
                destinationPath: path.join( WP_SERVER_FILES_PATH, 'wp-cli' ),
        },
        {
```

@katinthehatsite could you try testing again using this diff? After applying that, I didn't run into any other issues regarding logout.

katinthehatsite (commented):
> Log in again (same or different account) and check that initial pull/push states are shown

I am re-testing this again and when I log back in, I am seeing `Push cancelled` even though it was in progress when I logged out:

Image

Just to confirm, is this the expected behaviour?

katinthehatsite (commented):
And another sidenote is that file it shows `Push cancelled` on the `Sync` tab, I am seeing in `Import/Export` tab that the sync is in progress:

Image

gcsecsey:
> Just to confirm, is this the expected behaviour?

Good catch, sorry for missing this earlier. No, I think it'd be better behavior to completely clear the state in this case, and not show the success message.

> And another sidenote is that file it shows `Push cancelled` on the `Sync` tab, I am seeing in `Import/Export` tab that the sync is in progress:

Thanks for catching, this is also clearly a bug. I'll also address this one.

gcsecsey:
I fixed both issues, now the state is completely reset on logout:

https://github.com/user-attachments/assets/d5ee283b-315c-432d-9aaf-6dbe64eb585f

apps/studio/src/components/content-tab-import-export.tsx
```diff
@@ -356,12 +357,18 @@ export function ContentTabImportExport( { selectedSite }: ContentTabImportExport
...
+		connectedSites.some( ( site ) => {
+			const pushState = syncOperationsSelectors.selectPushState(
+				selectedSite.id,
+				site.id
+			)( state );
+			return pushBackupIsUploading( pushState?.status.key );
+		} )
 	);
-	const isThisSiteSyncing = isPulling || isPushing;
+	const isThisSiteSyncing = isPulling || isUploadingPushBackup;
```
fredrikekelund: Nice fix and good catch, @katinthehatsite 👍

gcsecsey:
@katinthehatsite @fredrikekelund could you please give this another look when you have some time? Thanks! 🙏

fredrikekelund:
@gcsecsey, yep, I'll take another look later today. Sorry for the delay

apps/studio/src/hooks/use-import-export.tsx
```diff
@@ -77,6 +79,14 @@ export const ImportExportProvider = ( { children }: { children: React.ReactNode
...
 	const [ exportState, setExportState ] = useState< ExportProgressState >( {} );
 	const { startServer, stopServer, updateSite } = useSiteDetails();
+	const { isAuthenticated } = useAuth();
+
+	useEffect( () => {
+		if ( ! isAuthenticated ) {
+			setImportState( {} );
+			setExportState( {} );
+		}
+	}, [ isAuthenticated ] );
```
fredrikekelund, line 99: The `cancelPull` thunk doesn't abort import operations – it only aborts the initial API call, the polling, and the downloading. So if the backup archive has already been fetched from the server when the user logs out, the `userLoggedOut` Redux middleware does nothing. 

That means there's no real point in resetting the import state in the renderer in this way (if there's an ongoing import operation), because the renderer process will continue to receive `on-import` events from the main process that repopulate the state.

This logic might be needed to get rid of the `Error pulling changes` message described in the original issue, but it seems sensible to only run this when there's no active pull operation.

We could actually take this discussion further, too, because I feel like we haven't really discussed this constraint before, and that it kind of breaks our existing logic that groups connected sites by WPCOM user. It effectively means we cannot (or shouldn't) always stop a pull operation when a user logs out.

apps/studio/src/stores/sync/sync-operations-slice.ts
```diff
@@ -304,56 +416,74 @@ const createTypedAsyncThunk = createAsyncThunk.withTypes< {
...
+		// See cancelPushThunk note: skip "cancelled" UI state for silent cancels (logout cleanup).
+		if ( displayNotification ) {
+			dispatch(
+				syncOperationsActions.updatePullState( {
+					selectedSiteId,
+					remoteSiteId,
+					state: { status: getPullStatesProgressInfo().cancelled },
+				} )
+			);
+		}
```
fredrikekelund: So what happens instead..? Which part of the code updates this slice when a user logs out? Maybe it would be clearer if `displayNotification` didn't determine how we update the state, and we added an explicit call to reset the sync operation state in the `userLoggedOut` middleware. Thoughts, @gcsecsey?

apps/studio/src/stores/sync/sync-operations-slice.ts
```diff
@@ -304,56 +416,74 @@ const createTypedAsyncThunk = createAsyncThunk.withTypes< {
...
+		// Treat `displayNotification: false` as a "silent cancel" (e.g. logout cleanup):
+		// abort the underlying operation but avoid creating a persisted "cancelled" UI state.
+		if ( displayNotification ) {
+			dispatch(
+				syncOperationsActions.updatePushState( {
+					selectedSiteId,
+					remoteSiteId,
+					state: { status: getPushStatesProgressInfo().cancelled },
+				} )
+			);
```
fredrikekelund: Same question here as in `cancelPullThunk`: which part of the code updates the sync operation state when a user logs out?

fredrikekelund (commented):
This issue is really surprisingly complex… 🤔 I raised a point about how we handle active import operations when a user logs out. I think it's OK if this PR doesn't have a perfect solution to that problem, but we should at least be explicit in the code and acknowledge that we don't/can't abort those operations.

Also, I'd like to clarify how the sync operation state gets updated when `displayNotification=true` in the `cancel{Push,Pull}Thunk` functions.

apps/studio/src/hooks/use-import-export.tsx
gcsecsey: Thanks for the additional context here @fredrikekelund! I get that it doesn't accomplish much to only reset the state in the renderer when we have an ongoing import, but I'd also be hesitant to completely cancel the import. Regarding grouping sites by WPCOM user, I think we should let the import continue if it's already started by the time the user logs out, as the import doesn't rely on authentication.

I think your suggestion makes sense, to only run this when there's no active pull operation. That'd mean that we let the import complete, and then clear the state.

apps/studio/src/stores/sync/sync-operations-slice.ts
gcsecsey: The slice is reset via a reducer [here](https://github.com/Automattic/studio/pull/2707/changes/BASE..e5f06f8684f2a5acee209857d5e69a7bbfc096cf#diff-92c179613ab910e751e38abd5c78a6f57165eee3aa3f8f27c6c962d354a1b846R194) on logout. I agree that using the `displayNotification` both for controlling the notification and the state update is a bit confusing. In the [original commit](https://github.com/Automattic/studio/pull/2707/changes/d47341d442b347fe379c1bee621ef1c0ae61bcbc#diff-f20b36a30147c673d0de7d7ed7748dc1814844eb588dc9831f6ab097fe4825ebR207-R215) of this PR, we had the explicit cancellation in the middleware, and later moved it to the thunks. The PUSH/PULL pollers were exported in the meantime to `stopPushPoller`/`stopPullPoller`, so we could use those and move the cancellation back to the middleware.

apps/studio/src/stores/sync/sync-operations-slice.ts
gcsecsey: Yes, this is the same as above, it's actually cleared [here](https://github.com/Automattic/studio/pull/2707/changes/BASE..e5f06f8684f2a5acee209857d5e69a7bbfc096cf#diff-92c179613ab910e751e38abd5c78a6f57165eee3aa3f8f27c6c962d354a1b846R194).

apps/studio/src/index.ts
fredrikekelund, line 1: I can't figure out what these changes are related to. Are they leftovers from something else?

apps/studio/__mocks__/electron.ts
fredrikekelund, line 1: Same as for `apps/studio/src/index.ts` – is this unrelated?

apps/studio/src/modules/sync/components/sync-connected-sites.tsx
```diff
@@ -59,22 +60,37 @@ const SyncConnectedSiteControls = ( {
...
+				site.id
+			)( state );
+			const pushState = syncOperationsSelectors.selectPushState(
+				selectedSite.id,
+				site.id
+			)( state );
+			return (
+				pullBackupIsDownloadingOrImporting( pullState?.status.key ) ||
+				pushBackupIsUploading( pushState?.status.key )
+			);
```
fredrikekelund: Not a huge deal, but we could move this logic into the Redux slice

apps/studio/src/modules/sync/components/sync-connected-sites.tsx
```diff
@@ -59,22 +60,37 @@ const SyncConnectedSiteControls = ( {
...
-	const isAnySiteSyncing = isAnySitePulling || isAnySitePushing;
+	const isAnySiteDoingLocalSyncWork = useRootSelector( ( state ) => {
+		const anyPullLocal = Object.values( syncOperationsSelectors.selectPullStates( state ) ).some(
+			( pullState ) => pullBackupIsDownloadingOrImporting( pullState.status.key )
+		);
+		const anyPushLocal = Object.values( syncOperationsSelectors.selectPushStates( state ) ).some(
+			( pushState ) => pushBackupIsUploading( pushState.status.key )
+		);
+		return anyPullLocal || anyPushLocal;
+	} );
```
fredrikekelund: Same thing here. Optional, though

fredrikekelund (commented):
This is looking pretty good, but there's an issue where the sidebar spinner doesn't disappear after logging out

apps/studio/src/hooks/use-import-export.tsx
gcsecsey: Sorry, I forgot to close the loop here earlier, I implemented this in 2b4a33e.

apps/studio/src/index.ts
gcsecsey: Yes, sorry, these are leftovers from https://github.com/Automattic/studio/pull/3114, just got added to this branch by mistake. I'll revert this file.

apps/studio/__mocks__/electron.ts
gcsecsey: Same as index.ts, this is a leftover from https://github.com/Automattic/studio/pull/3114, just got added to this branch by mistake. I'll revert it.

apps/studio/src/modules/sync/components/sync-connected-sites.tsx
gcsecsey: It felt related to this PR, so I moved this in [f57a08e](https://github.com/Automattic/studio/pull/2707/commits/f57a08e37ac037b478fb24281ffc92c112e6be9c) instead of creating a follow-up.

apps/studio/src/modules/sync/components/sync-connected-sites.tsx
gcsecsey: I moved it in the same go as above [f57a08e](https://github.com/Automattic/studio/pull/2707/commits/f57a08e37ac037b478fb24281ffc92c112e6be9c).

gcsecsey:
@fredrikekelund I rebased the PR and followed up on your comments, except for the issue where the sidebar spinner doesn't disappear after logging out. I'll look into that tomorrow and ping you again when it's ready for another review.

gcsecsey:
@fredrikekelund @katinthehatsite I think this is ready for another review.

While working on fixing the spinner state, I found that it's a symptom of not having the same abort callbacks for pulls as for pushes.

I updated the behavior so that logging out during the download phase will stop without the error dialog from before, and will clear the spinner too. During the import phase, the import is local, so we let it finish, keeping the spinner visible until it completes. I think that’s the right tradeoff, but happy to revisit. WDYT?

fredrikekelund (approved):
Thanks for getting this over the finish line, @gcsecsey. I pushed some final tweaks that moved selector logic into the Redux slice, but fundamentally, this was already looking good. 

I think you already know this, but I'll say it anyway: long lead times between reviews and changes slow down the process as a whole, because reviewers loose context after a couple of days. RSM got in the way of this one, this is arguably a low priority fix, and I dragged my feet reviewing this on a few occasions, too. Nevertheless, something we can all keep in mind in the future 🙂

PR #2710, opened 2026-03-05
feat(studio): Set default site directory location

### LINKED ISSUE #587: Feature Request: Set a default "new site" location
### What

Currently, WP Studio assumes you want to create new sites within `C:\Users\USER\Studio\SITE`. I request the ability to set a different directory as the "default" starting path for all future new sites created within WP Studio.

### Why

I do not want to install sites to this default location, I have another location on another drive that has its own backup policy set in place, and creating new sites should always go to this location by default.

### How

_No response_

mikaoelitiana:
## Related issues

- Fixes #587 

## How AI was used in this PR

- Copilot handled code changes for default site directory behavior, plus tests and linting based on repo rules.

## Proposed Changes

- Resolve the default site directory dynamically (dialogs, copy site, proposed path, and site name generation) instead of using a constant because the path is now changed in run time
- Add user settings support for saving a default site directory and a picker UI.
- Update tests to cover the new behavior and mocks.

## Testing Instructions

- npm test -- apps/studio/src/tests/ipc-handlers.test.ts
- npm run lint
- npm run make-pot

## Pre-merge Checklist

## Preview 

https://github.com/user-attachments/assets/a0dfa014-d716-49a6-be02-be7a39b29382


wojtekn:
Hi @mikaoelitiana ! Thanks for your contribution. Feature-wise, it works well.

It would be great to adjust field UI to make it consistent with the path selector we have in 'Add site' flow:

CleanShot 2026-03-17 at 11 04 44@2x

Currently, the default path selection is a bit different:

CleanShot 2026-03-17 at 11 06 34@2x

Let me know if you need help with this change.

mikaoelitiana:
> Hi @mikaoelitiana ! Thanks for your contribution. Feature-wise, it works well.
> 
> 
> 
> It would be great to adjust field UI to make it consistent with the path selector we have in 'Add site' flow:
> 
> 
> 
> CleanShot 2026-03-17 at 11 04 44@2x
> 
> 
> 
> Currently, the default path selection is a bit different:
> 
> 
> 
> CleanShot 2026-03-17 at 11 06 34@2x
> 
> 
> 
> Let me know if you need help with this change.

Sure I will fix that

mikaoelitiana:
@wojtekn I need a quick clarification. This UI is currently in 

https://github.com/mikaoelitiana/studio/blob/83db08427962cfa6bbec11dbda911c38f13052cd/apps/studio/src/modules/add-site/components/create-site-form.tsx#L100 

I am thinking of extracting to a common path and import it in both modules. Is that the right way we should rather duplicate it on each module?

mikaoelitiana:
@wojtekn I made the changes and I have updated the preview video as well

wojtekn:
@mikaoelitiana thanks for updating that - the UI looks consistent and, feature-wise, works fine.

I see there are some conflicts now - let me know if you need help addressing those.

mikaoelitiana:
@wojtekn all conflicts are now fixed 👍🏽

wojtekn:
@claude review

apps/studio/src/components/form-path-input.tsx
```diff
@@ -0,0 +1,72 @@
...
+				/**
+				 * The below `aria-describedby` presumes the error message always
+				 * relates to the local path input, which is true currently as it is the
+				 * only data validation in place. If we ever introduce additional data
+				 * validation we need to expand the robustness of this
+				 * `aria-describedby` attribute so that it only targets relevant error
+				 * messages.
+				 */
+				aria-describedby={ error ? 'site-path-error' : undefined }
+				type="button"
```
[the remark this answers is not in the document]
mikaoelitiana: This is in another component `apps/studio/src/components/site-form-error.tsx`

apps/studio/src/components/form-path-input.tsx
```diff
@@ -0,0 +1,72 @@
...
+				) }
+				data-testid="select-path-button"
+				onClick={ onClick }
+				id={ id }
+			>
+				<div
+					aria-hidden="true"
+					tabIndex={ -1 }
+					className="w-full text-left pl-3 py-3 min-h-10"
+					onChange={ () => {} }
```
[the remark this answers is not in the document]
mikaoelitiana: I have fixed that

apps/studio/src/components/form-path-input.tsx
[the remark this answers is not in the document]
wojtekn: Thanks for extracting the field to separate the component and for reusing it in both places.

We may need to adjust the error field id handling to ensure it doesn't break if the field is used multiple times in one context. What if we derived the SiteFormError id from the FormPathInputComponent, e.g., as `${id}-error`?

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -120,6 +164,13 @@ export const PreferencesTab = ( { onClose }: { onClose: () => void } ) => {
 			{ ! isWindowsStore() && (
 				<StudioCliToggle value={ isCliInstalledSelection } onChange={ setDirtyIsCliInstalled } />
 			) }
+			<SettingsFormField label={ __( 'Default site directory' ) }>
+				<FormPathInputComponent
+					value={ isLoadingDefaultSiteDirectory ? __( 'Loading...' ) : defaultSiteDirectory ?? '' }
+					onClick={ handleChangeDefaultDirectory }
+					doesPathContainWordPress={ false }
```
wojtekn: It's a little bit confusing to pass doesPathContainWordPress as false here.

What if we made the tip message a prop passed to the component instead of keeping it as a boolean?

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -88,6 +93,13 @@ export const PreferencesTab = ( { onClose }: { onClose: () => void } ) => {
 		if ( dirtyIsCliInstalled !== undefined ) {
 			await saveCliIsInstalled( dirtyIsCliInstalled );
 		}
+		const isDefaultDirectoryDirty =
+			storedDefaultSiteDirectory !== undefined &&
+			defaultSiteDirectory !== undefined &&
+			storedDefaultSiteDirectory !== defaultSiteDirectory;
+		if ( isDefaultDirectoryDirty && defaultSiteDirectory ) {
```
wojtekn: Would it make sense to simplify it into `storedDefaultSiteDirectory !== undefined && defaultSiteDirectory && storedDefaultSiteDirectory !== defaultSiteDirectory`?

Or better, we could follow the `dirty*` convention we follow for other fields.

apps/studio/src/storage/paths.ts
```diff
@@ -125,3 +127,52 @@ function getAppName(): string {
...
+	await fsPromises.access( directory, fs.constants.W_OK );
+}
+
+export async function ensureWritableDirectory( directory: string ) {
+	await ensurePathIsDirectory( directory );
+	await ensurePathIsWritable( directory );
+}
+
+async function getStoredDefaultSiteDirectory(): Promise< string | undefined > {
+	const { loadUserData } = await import( 'src/storage/user-data' );
```
wojtekn: It would make sense to remove user-data dependency from paths. Maybe we could pass a loaded path to `resolveDefaultSiteDirectory`?

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
mikaoelitiana: Yes that totally makes sense. I will fix that

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
mikaoelitiana: I have fixed that

mikaoelitiana:
@wojtekn I have addressed your latest comments, let me know if there is anything else

mikaoelitiana:
@wojtekn thanks for merging trunk, I have also fixed a merge conflict

apps/studio/src/storage/paths.ts
```diff
@@ -109,3 +111,35 @@ function getAppName(): string {
...
+		} catch ( error ) {
+			console.warn(
+				'Stored default site directory is unavailable, falling back to built-in path.',
+				error
+			);
+		}
+	}
+
+	return defaultSitePath;
+}
```
fredrikekelund, line 145: I'm not sure that we should silently swallow errors here… For one, it won't be clear why Studio isn't respecting the default site directory setting. It's true that the user won't have to see an error modal, which is nice, but it could lead to confusion later. 

I'll also note that Studio recursively creates the site directory if it doesn't exist when creating a new site, so the fact that the parent directory doesn't exist isn't necessarily cause for an exception.

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -96,15 +104,47 @@ export const PreferencesTab = ( { onClose }: { onClose: () => void } ) => {
...
+			} finally {
+				if ( isMounted ) {
+					setIsLoadingDefaultSiteDirectory( false );
+				}
+			}
+		} )();
+		return () => {
+			isMounted = false;
+		};
+	}, [] );
```
fredrikekelund, line 136: Let's add a query to `apps/studio/src/stores/installed-apps-api.ts` to fetch `getIpcApi().getDefaultSiteDirectory()`. That way, we can call a `useGetDefaultSiteDirectory` hook in here and have RTK Query handle loading state, etc., internally. 

I know the filename is a bit misleading, but it's already being used as a "settings API", so this new property is a good fit.

apps/studio/src/components/form-path-input.tsx
fredrikekelund, line 1: For some reason, this component renders with a greater height in the settings modal than in the "Add a site" form. It looks out of place compared to the dropdown menus in the settings modal right now. Let's find a solution for that.

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -120,6 +160,14 @@ export const PreferencesTab = ( { onClose }: { onClose: () => void } ) => {
...
 				<StudioCliToggle value={ isCliInstalledSelection } onChange={ setDirtyIsCliInstalled } />
 			) }
+			<SettingsFormField label={ __( 'Default site directory' ) }>
+				<FormPathInputComponent
+					value={
+						isLoadingDefaultSiteDirectory ? __( 'Loading...' ) : defaultSiteDirectorySelection
+					}
+					onClick={ handleChangeDefaultDirectory }
+				/>
+			</SettingsFormField>
```
fredrikekelund, line 170: Let's move this field above the "Studio CLI" toggle. I think that makes for a better visual balance.

fredrikekelund (commented):
Hey, @mikaoelitiana 👋 Thanks for your contribution. This feature is working well.

I left a couple of smaller comments on the PR, related to the implementation. Let me know if anything is unclear or if you have concerns.

fredrikekelund (approved):
I took the liberty of pushing fixes for the items I identified in my review. Thanks again for your contribution, @mikaoelitiana 👍

apps/studio/src/storage/paths.ts
```diff
@@ -1,20 +1,9 @@
...
-// Import electron conditionally to avoid issues in child processes
-let app: Electron.App | undefined;
-try {
-	if ( ! inChildProcess() ) {
-		( { app } = require( 'electron' ) );
-	}
-} catch ( error ) {
-	// If electron is not available (e.g., in child process), app will remain undefined
-}
-
```
fredrikekelund, line 17: For the record: I made this change because the Studio app no longer spawns child processes that use the same codebase. All child processes are now running CLI code.

fredrikekelund:
I couldn't figure out how to trigger Buildkite CI on this PR, so I merged these changes via https://github.com/Automattic/studio/pull/2825. Thanks again, @mikaoelitiana 🙏

mikaoelitiana:
> I couldn't figure out how to trigger Buildkite CI on this PR, so I merged these changes via #2825. Thanks again, @mikaoelitiana 🙏

Thank you very much @fredrikekelund , sorry for the delay, I have been quite busy the past days

PR #2690, opened 2026-03-03
Replace pm2-axon with homegrown solution

fredrikekelund:
## Related issues

- Fixes STU-1301

## How AI was used in this PR

I wrote the initial plan and then relied on Codex to produce the initial implementation. I've since iterated on `apps/cli/lib/socket.ts` over multiple rounds, touching almost every part of it. The tests are auto-generated. I've barely touched them. I've reviewed all code and manually tested Studio to ensure that site events are emitted the same as before. 

## Proposed Changes

In https://github.com/Automattic/studio/pull/2398, we adopted a socket-based solution for sending site events from the CLI to Studio. We opted to use the [pm2-axon](https://www.npmjs.com/package/pm2-axon) module for this. It's a thin library, but it offers some conveniences when working with sockets, and it was already in our dependency tree.

We've since found that pm2-axon can easily be replaced with a homegrown solution that uses modules from the Node.js standard library. This will help significantly with maintainability and long-term stability. 

This PR removes the direct pm2-axon dependency in favor of `apps/cli/lib/socket.ts`

## Testing Instructions

1. Run `npm start`
2. Ensure that starting and stopping sites works as expected and that the UI updates accordingly.
3. Run `node apps/cli/dist/cli/main.js site start --skip-browser --path PATH_TO_SITE` and ensure that the Studio UI is updated when the site starts

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2712, opened 2026-03-05
Replace pm2 with homegrown process manager daemon

fredrikekelund:
## Related issues

- Fixes STU-1349

## How AI was used in this PR

Codex wrote most of the code in this PR, starting from a detailed list of requirements I shared (based on our experience with pm2). I iterated on this over multiple days, both with Codex and by hand. I've touched all major files in the PR. `process-manager-ipc.ts` is a standout example – I made many manual edits there. I've reviewed all the code. The code I've spent the least time editing or reviewing is the tests. 

## Proposed Changes

> [!TIP]
> I'm happy to pair for a review of this PR. I know it's a big one. Also, please note that this PR builds upon the changes in https://github.com/Automattic/studio/pull/2690. That PR should be reviewed first.

See STU-1349 for the "why" of removing pm2. This PR implements the following:

- A `process-manager-daemon.ts` that manages child processes. This involves forking them, tracking them in-memory, forwarding child IPC messages as typed daemon events, and writing stdout/stderr to PM2-compatible log files under `~/.studio/pm2/logs`. This file is the PM2 replacement. 
- `daemon-client.ts` replaces `pm2-manager.ts`. It talks to the daemon over a control socket for request/response commands and an events socket for receiving messages from child processes and process events (online/exit, etc.).

The architecture largely follows PM2's example, and `process-manager-daemon.ts` implements the same functionality. A few notes to get reviewers started:
- Starting/stopping processes are relatively easy to understand from the code.
- IPC is a little trickier, but the gist is that `daemon-client` exports a `sendMessageToProcess` function that uses the process manager "control socket" to send a message with a `processId` target. When `process-manager-daemon.ts` receives the message, it uses the standard [Node.js IPC mechanism](https://nodejs.org/docs/latest-v24.x/api/child_process.html#subprocesssendmessage-sendhandle-options-callback) to forward it to the child process. Child processes cannot respond directly to these messages, but they can emit events through Node.js standard IPC, which `process-manager-daemon.ts` picks up and broadcasts on the "events socket". `daemonBus` listens on the events socket, and the `subscribeSiteEvents` function picks up the `process-message` and `process-event` events, which are ultimately picked up by `_events.ts`.

## Testing Instructions

1. Run `npm start`
2. Ensure that starting and stopping sites works as expected
3. Try creating a new site, too
4. While Studio is still running, run `node apps/cli/dist/cli/main.js site start --skip-browser --path PATH_TO_SITE` in your terminal
5. Ensure that the site is immediately marked as started in Studio

## Pre-merge Checklist


apps/cli/lib/types/process-manager-ipc.ts
```diff
@@ -0,0 +1,116 @@
+import { z } from 'zod';
+import {
+	childMessageFromProcessManagerSchema,
+	managerMessageSchema,
+} from 'cli/lib/types/wordpress-server-ipc';
+
+// Zod schema for process descriptions
+export const processDescriptionSchema = z.object( {
```
bcotrim: Should we be more intentional about the status + pid options?
I believe an online process will always have a pid, and the optional scenario is only for stopped processes. We could use an union to be more intentional about this type, what do you think?

apps/cli/lib/socket.ts
```diff
@@ -58,6 +67,19 @@ export class SocketClient {
 		return sendPromise;
 	}
 
+	sendAndWaitForResponse( message: unknown ): Promise< unknown > {
+		const payload = encodeSocketMessage( message );
+		const responsePromise = this.queue.then( async () => {
+			const socket = await this.connect();
+			return this.sendAndReadFromSocket( socket, payload );
+		} );
+		this.queue = responsePromise.then(
```
bcotrim: This pattern is odd, I wonder if there's a better approach to it.

apps/cli/lib/daemon-paths.ts
```diff
@@ -0,0 +1,14 @@
+import os from 'os';
+import path from 'path';
+
+export const PROCESS_MANAGER_HOME =
+	process.env.STUDIO_PROCESS_MANAGER_HOME ?? path.join( os.homedir(), '.studio', 'pm2' );
```
bcotrim, line 5: I'm assuming you decided keeping the pm2 name for backwards compatibility reasons, but I wonder if we should just rename it and handle existing folder content in a migration script or similar?

bcotrim (changes requested):
While testing this I found some weird behaviors:
1.  Killing a site process (`kill pid`) didn't reflect on Studio. Site was still running in the UI and stop operation didn't work (CLI worked fine). I could only replicate this once
2. Doing `Stop all` from the Studio UI button, seems to break the UI. After doing stop all starting any other sites would not be reflected in the UI. I was able to replicate this consistently.

Added some minor comments to the code that aren't blockers, but we should look into these 2 issues.

apps/cli/lib/types/process-manager-ipc.ts
fredrikekelund: Good call! 👍 I'm all for well-defined types. I've updated the code accordingly.

apps/cli/lib/daemon-paths.ts
fredrikekelund: > I'm assuming you decided keeping the pm2 name for backwards compatibility reasons

Yep, that's right. Seems reasonable to change the path and remove all pm2 naming, but I'd rather do it in a follow-up PR. I've added STU-1390 to track

fredrikekelund:
Thanks for the review, @bcotrim! 🙏 I should've caught the "Stop all" thing sooner…

> Killing a site process (kill pid) didn't reflect on Studio. Site was still running in the UI and stop operation didn't work (CLI worked fine). I could only replicate this once

I'll test this some more. There's no obvious reason why this would happen…

> Doing Stop all from the Studio UI button, seems to break the UI. After doing stop all starting any other sites would not be reflected in the UI. I was able to replicate this consistently.

That was because `SocketClient` wouldn't reconnect when the underlying socket closed (which it did when the process manager daemon shut down and closed its `eventsServer`).

I've refactored `socket.ts` so the socket client automatically reconnects and verified that this has the desired effect in the app. I also took the opportunity to add an E2E test that clicks the `Stop all` button and then restarts a site. Seems like a useful thing to have in the test suite.

apps/cli/lib/types/process-manager-ipc.ts
```diff
@@ -0,0 +1,125 @@
...
+	pmId: z.number(),
+} );
+const processDescriptionSchemaRunning = processDescriptionSchemaBase.extend( {
+	status: z.literal( 'online' ),
+	pid: z.number(),
+} );
+const processDescriptionSchemaStopped = processDescriptionSchemaBase.extend( {
+	status: z.literal( 'stopped' ),
+} );
+export const processDescriptionSchema = z.discriminatedUnion( 'status', [
```
bcotrim, line 19: Nice 👍

bcotrim (approved):
LGTM 👍 
I can no longer replicate any of the issues I mentioned, thanks for making the changes

apps/cli/lib/socket.ts
fredrikekelund: I agree it's kind of an odd pattern. I also reacted to it previously and chatted with Codex about it. 

Really, it's deceptively simple. It works because each `send` call reassigns the `queue` promise, so the result is a chain of promises. See https://jsfiddle.net/vjpdy7q4/1/ for a clearer proof of concept

PR #2713, opened 2026-03-05
Add DEV_APP_DATA_PATH env var for isolated appdata JSON

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

AI assisted with implementation. I code reviewed and verified the solution.

## Proposed Changes

I propose adding an easy way to quickly start the app for testing, to make screenshots, etc. without touching the appdata file that could include production sites.

- Add `DEV_APP_DATA_PATH` env var that overrides the appdata JSON file path in both the main process (`storage/paths.ts`) and the CLI child process (`cli/lib/appdata.ts`)
- Add `start:test` script to use `DEV_APP_DATA_PATH` instead of the E2E flags, so the app runs in normal (non-E2E) mode with only the appdata JSON redirected

## Testing Instructions

- Run `npm run start:test` — app starts with a clean appdata at `/tmp/studio-test-appdata.json`, server-files cache is preserved
- Create a site
- Run `npm start` — app starts with real appdata, unaffected

## Pre-merge Checklist


ivan-ottinger (approved):
I think this is a great idea!

The changes look good to me and work well - I am able to run the app without touching the original appdata JSON file and then switch back to using it again. 👍🏼 

Two things:
- the command won't run on Windows at its current form
- maybe we could consider adding the mention to the project documentation

wojtekn:
@ivan-ottinger thanks for the review.

> the command won't run on Windows at its current form

As this command is purely for development and we don't use Windows as the main machine for development or for keeping our production sites, I think it's okay to keep this one a Mac-only helper.

> maybe we could consider adding the mention to the project documentation

Makes sense, added.

PR #2719, opened 2026-03-06
Fix some unsafe type assertions

fredrikekelund:
## Related issues

- Related to #2694

## How AI was used in this PR

https://github.com/Automattic/studio/pull/2694#discussion_r2894406626 prompted me to look into ESLint rules to avoid similar problems. I found [@typescript-eslint/no-unsafe-type-assertion](https://typescript-eslint.io/rules/no-unsafe-type-assertion/) and tried enabling it, letting Codex fix the errors. It did a decent job, but ultimately, there were too many errors for a single PR. The stuff that made it into this PR was mostly written by hand, except for `format-rtk-error.tsx`.

## Proposed Changes

See the previous section for some background… In short, I tried enabling [@typescript-eslint/no-unsafe-type-assertion](https://typescript-eslint.io/rules/no-unsafe-type-assertion/) but found too many errors. This PR primarily fixes up some type assertions related to blueprints, but also in `format-rtk-error.tsx`, `snapshot-slice.ts`, and `tools/common/lib/locale.ts`.

## Testing Instructions

1. Code review is paramount
2. Lightly smoke test creating a new site from a blueprint

## Pre-merge Checklist


katinthehatsite (commented):
Thanks for the fix, I explored similar solutions but I think this one is slightly better. This approach uses structuredClone instead of spread and my  approach used shallow spread + index-based assignment.

Do we also need to update the `Blueprint` type in `studio/apps/studio/src/modules/add-site/lib/apply-blueprint-form-values.ts` ?

fredrikekelund:
> Do we also need to update the `Blueprint` type in `studio/apps/studio/src/modules/add-site/lib/apply-blueprint-form-values.ts`

We should. Good catch 👍 Came up when I merged trunk

fredrikekelund:
Looks like I still need to update this PR to fix the failing tests…

fredrikekelund:
I've fixed the tests. The problem was caused by importing the `isStepDefinition` function from `@wp-playground/blueprints`. That's a tiny function, but because of how the `@wp-playground/blueprints` exports are structured, Vite also pulled in the `@php-wasm/node-polyfills` package into the bundle. `apps/studio/electron.vite.config.ts` defines all `@php-wasm/*` packages as external, but that package still wasn't preserved in the packaged `node_modules` directory, since `@wp-playground/blueprints` is a `devDependency`. All this resulted in a fatal error when the Electron app started. 

In short… A bit of a mess for such a small issue. I decided to just copy and inline the `isStepDefinition` function to get around the problem. I've also raised the fundamental issue with Orbit (p1773050695429429-slack-C09GMMV3T6J).

I'd appreciate another review, @katinthehatsite 🙏

fredrikekelund:
Please ignore the lint failures. They will be fixed when https://github.com/Automattic/studio/pull/2720 is merged into this branch.

apps/cli/lib/cli-args-sanitizer.ts
```diff
@@ -116,6 +112,8 @@ export function sanitizeBlueprint( blueprint: Blueprint | undefined ): object |
  * Prepares data for Sentry by stringifying nested objects to avoid normalization limits.
  */
 export function sanitizeRunCLIArgs( args: RunCLIArgs ): Record< string, unknown > {
+	const blueprint = args.blueprint && 'steps' in args.blueprint ? args.blueprint : undefined;
```
epeicher: This would return `undefined` for a blueprint that has no `steps`, but there are blueprints that can contain no steps, is this intentional?

epeicher (approved):
Thanks for improving this @fredrikekelund! I have reviewed it, and I have just left a question for consideration. I have tested site creation using blueprints, and I have not found any issues. LGTM!

apps/cli/lib/cli-args-sanitizer.ts
fredrikekelund: That's an excellent point! I was trying to appease TS here, but from a functionality perspective, you're right that this is lossy. I've updated the code and opted to use a type assertion after all.

fredrikekelund:
Thanks for the review, @epeicher 🙏

PR #2720, opened 2026-03-06
Remove `AllowedPHPVersion` type

fredrikekelund:
## Related issues

N/A

## How AI was used in this PR

Barely at all. I identified the underlying problem and fixed it by hand. 

## Proposed Changes

> [!NOTE]  
> This PR builds upon the changes in https://github.com/Automattic/studio/pull/2719.

The `AllowedPHPVersion` type is misleading. It's just a `string` type, but `SupportedPHPVersion` is really our canonical definition of which PHP versions we support. This PR removes `AllowedPHPVersion` and replaces it with `SupportedPHPVersion`.

I also took the opportunity to remove `apps/studio/src/lib/wordpress-server-types.ts` in its entirety, since the only other definition there was the `WordPressServerProcess` interface. We only have a single implementation of that interface now, so I think it makes more sense to remove it.

Tagging @bcotrim for review, since `wordpress-server-types.ts` was originally implemented in https://github.com/Automattic/studio/pull/2281.

## Testing Instructions

Code review should suffice

## Pre-merge Checklist


bcotrim (approved):
Good cleanup 👍 
We could probably remove the provider constants slice now, but I would consider a good follow-up PR. Let me know if you want me to take care of that.
I did not notice any regressions when testing the app, PHP versions behave as expected when creating/editing sites.

fredrikekelund:
> We could probably remove the provider constants slice now, but I would consider a good follow-up PR. Let me know if you want me to take care of that.

Thanks for confirming that 👍 I was under the same impression. I created STU-1378 to track this, and assigned the issue to you, @bcotrim

PR #2724, opened 2026-03-06
[Tooling] Upload CDN builds as Internal first, publish as External later

iangmaia:
## Related issues

- Fixes AINFRA-2102
- Depends on wordpress-mobile/release-toolkit#701, released in toolkit `14.7.0`, bundled in this PR
- Supersedes #2689 and #2695

## How AI was used in this PR

Claude Code implemented the Fastfile changes based on the approach designed in #2689 and refined through review of #2695, and assisted with rebasing and verification. All generated code was reviewed.

## Proposed Changes

Today, production builds become publicly available on the Apps CDN as soon as `finalize_release` uploads them, before the Release Manager has a chance to verify them.

This PR closes that gap by giving the team an internal testing window between finalize and publish:

- `finalize_release` uploads Production builds with **Internal** visibility: downloadable by Automatticians for smoke-testing, but not served to users (the auto-update endpoint only serves External builds)
- `publish_release` flips them to **External** (via the new `update_apps_cdn_build_metadata` toolkit action) right before publishing the GitHub release
- Nightlies and Betas keep External visibility, as they don't go through `publish_release`
- The CDN post IDs to switch the builds are in the draft GitHub release body, so publish only touches builds from the latest finalization (an earlier draft-status approach, #2695, was abandoned because draft posts aren't easily testable by the RM)

## Testing Instructions

- The visibility switches were verified end-to-end against the CDN sandbox site (AINFRA-2171). See the test plan on wordpress-mobile/release-toolkit#701
- The full finalize to publish flow will be exercised during the next release cycle

## Pre-merge Checklist

- [x] release-toolkit#701 merged, released (14.7.0), and bundled in this PR


fastlane/Fastfile
```diff
@@ -779,6 +792,70 @@ def create_draft_github_release(version:, release_tag:, builds:)
...
+# Reads CDN post IDs embedded in the draft GitHub release body by {create_draft_github_release},
+# then calls {update_apps_cdn_build_metadata} for each to set visibility to External.
+#
+# @param version [String] The version to publish (e.g., '1.7.5')
+#
+def make_cdn_builds_public(version:)
+  release_name = "v#{version}"
+  cdn_post_ids = extract_cdn_post_ids_from_draft_release(release_name: release_name)
+
+  UI.user_error!("No CDN post IDs found in draft release #{release_name}. Cannot publish without updating CDN visibility.") if cdn_post_ids.empty?
```
AliSoftware: I wonder if there is a (`curl`?) command we could run from developers' machines to ask the AppsCDN site for those post IDs if they are missing?
 - The goal would be that if we hit that case where the GitHub Release doesn't have the CDN post IDs, at least the release manager could run that command from their terminal to get the list of post IDs for the given version, then edit the GitHub Release draft to copy/paste that list of post IDs into the release draft's description, then retry the `publish_release` lane from Releases V2, allowing them to recover from such a state even if a bit manually?
 - This would only be doable from a developer's machine (or one that is AutoProxxy'd), since [the only way to have the Apps CDN would expose the list of builds with `visibility:internal` in the API response is if that API call was made by an Automattician user](https://github.a8c.com/Automattic/wpcom/blob/97f94b4db2d7c02b2fc05446df737d60a40e4429/wp-content/lib/a8c/cdn/src/functions.php#L54-L55) (i.e. [if `is_automattician` returns `true`](https://github.a8c.com/Automattic/wpcom/blob/97f94b4db2d7c02b2fc05446df737d60a40e4429/wp-content/lib/a8c/cdn/src/functions.php#L87))—which is why you couldn't do that from CI

If we can find such a command, then maybe it would then be worth to include it in this `UI.user_error!` message as a suggestion for the developer to run it as a way to recover?

fastlane/Fastfile
```diff
@@ -779,6 +792,70 @@ def create_draft_github_release(version:, release_tag:, builds:)
...
+
+  post_ids = cdn_post_ids.values.map(&:to_i)
```
AliSoftware: Given the only usage we have for those post IDs is to get the list of the IDs (i.e. we don't care about the keys of that JSON in the first place), what's the point of storing the post IDs in the GitHub Release draft as a JSON object, as opposed to a simple comma-separated list of IDs?

```html
<!-- CDN_POST_IDS:123,456,789 -->
```

This would:
 - Make this magic comment smaller and taking less visual space (i.e. less verbose if e.g. someone wanted to edit the draft to amend the release description and got this large JSON being a bit in the way, also avoiding risk of introduce a syntax error if accidentally making changes to that comment…)
 - Make it way easier to parse in `extract_cdn_post_ids_from_draft_release` (no need to parse the JSON, just have that helper return `match[1].split(',').map(&:to_i)` as an `Array<Integer>` instead of a `Hash`)
 - Make it easier to add that comment manually in the unlikely case that the Release Managers would hit [this case](https://github.com/Automattic/studio/pull/2724/changes#r2905431142) and would want to recover from it manually by adding the list of post IDs themselves to unblock the process

fastlane/Fastfile
```diff
@@ -762,6 +769,12 @@ def create_draft_github_release(version:, release_tag:, builds:)
             'The latest version is always available on the [WordPress Studio](https://developer.wordpress.com/studio/) site.'
   end
 
+  # Embed CDN post IDs so publish_release can update their visibility later
+  cdn_post_ids = builds.each_with_object({}) do |(key, build), hash|
+    hash[key] = build[:post_id] if build[:post_id]&.positive?
+  end
+  body += "\n<!-- CDN_POST_IDS:#{cdn_post_ids.to_json} -->" unless cdn_post_ids.empty?
```
AliSoftware: Based on my previous suggestion comment about storing them as a simple comma-separate list instead:

```suggestion
  # Embed CDN post IDs so publish_release can update their visibility later
  cdn_post_ids = builds.values.map(&:post_id).select(&:positive?)
  body += "\n<!-- CDN_POST_IDS:#{cdn_post_ids.join(',')} -->" unless cdn_post_ids.empty?
```

fastlane/Fastfile
```diff
@@ -779,6 +792,70 @@ def create_draft_github_release(version:, release_tag:, builds:)
...
+  rescue JSON::ParserError => e
+    UI.error("Failed to parse CDN post IDs from draft release #{release_name}: #{e.message}")
+    return {}
+  end
+
+  parsed.each do |key, value|
+    UI.user_error!("Invalid CDN post ID for '#{key}' in draft release #{release_name}: #{value.inspect} (expected positive integer)") unless value.is_a?(Integer) && value.positive?
+  end
+
+  parsed
```
AliSoftware: Based on my previous suggestion comment about storing them as a simple comma-separate list instead:

```suggestion
  match = body.match(/<!-- CDN_POST_IDS:([0-9,]*) -->/)
  unless match
    UI.important("Draft release #{release_name} found but no CDN post IDs embedded in the body.")
    return []
  end

  match[1]).split(',').map(&:to_i)
```

fastlane/Fastfile
iangmaia: 👍 Good idea, it simplifies things a lot. Updated on 3739247884501a502aa3447c7d61ea09c01840c6.

fastlane/Fastfile
iangmaia: Updated on 3739247884501a502aa3447c7d61ea09c01840c6.

fastlane/Fastfile
iangmaia: Updated on 3739247884501a502aa3447c7d61ea09c01840c6

fastlane/Fastfile
iangmaia: I've tried a few things and added a message (hard coding the visibility code for `internal`) in 38e6abe5b3333974e790c1bc8a4fe29c07575672.

AliSoftware (approved):
Code changes in `Fastlane` looks good so approving to unblock 

But 🎗️ reminder you'll need to first merge https://github.com/wordpress-mobile/release-toolkit/pull/701 then releasing a new version then do a `bundle update fastlane-plugin-wpmreleasetoolkit` in this PR before it's actually ready to be merged.

iangmaia:
FYI: I haven't merged this PR yet as I need to double check the [API updates](https://github.com/wordpress-mobile/release-toolkit/pull/701#discussion_r2947118032) and review the process with the team again. I the meantime I got busy with other tasks but will get back to this soon.

fastlane/Fastfile
```diff
@@ -615,7 +621,8 @@ def upload_file_to_apps_cdn(site_id:, product:, file_path:, platform:, arch:, bu
     UI.message("           error on duplicate: #{error_on_duplicate}")
 
     return {
-      media_url: media_url
+      media_url: media_url,
+      post_id: 0
```
mokagio, line 625: While it doesn't take long to look at the rest of the code on top of this, seeing this `0` in the code threw me off.

What do you think of adding a little comment?


```suggestion
      # downstream checks for a positive post_id; 0 will result in no upload, as appropriate for a dry run
      post_id: 0
```

mokagio (approved):
@iangmaia I'm guessing the recent update here means that you confirmed the flow with the team as mentioned in an earlier comment.

Approving to unblock so this can be merged once the conflicts on `Fastfile` have been addressed.

PR #2729, opened 2026-03-09
Add skills and instructions management to Edit Site modal

katinthehatsite:
## Related issues

- Fixes STU-1337 and STU-1453

## How AI was used in this PR

Claude Code was used to implement the changes, with the developer directing the architecture and reviewing each step.

## Proposed Changes

- Move Skills and Instructions tabs from the standalone AI Settings modal into the Edit Site modal (tabs: General, Debugging, Skills, Instructions)
- Add "AI Skills" and "Agent Instructions" sections to the site Settings tab with buttons that open the Edit Site modal on the correct tab
- Add a link to open global Studio Settings on the Skills tab
- Align per-site Skills and Instructions panels with the global Studio Settings UI pattern (Installed/Available sections, dropdown menus, Install all)
- Add `removeAgentInstruction` IPC handler to support removing instruction files (AGENTS.md, CLAUDE.md, STUDIO.md)
- Remove the AI Settings menu item from the assistant dropdown
- Delete the standalone `ai-settings-modal.tsx` component
- Extract shared panels into `site-settings-panels.tsx`

Screenshot 2026-03-26 at 4 50 00 PM

## Testing Instructions

- Navigate to a site's Settings tab
- Verify "AI Skills" section shows a link to Studio Settings and a "Manage site skills" button
- Verify "Agent Instructions" section shows a "Manage instructions" button
- Click "Manage site skills" — Edit Site modal should open on the Skills tab
- Click "Manage instructions" — Edit Site modal should open on the Instructions tab
- In the Skills tab, verify install/remove individual skills works and the UI matches the global Studio Settings Skills tab
- In the Instructions tab, verify install/remove instruction files works with the same UI pattern
- Click the "Studio Settings" link — global settings should open on the Skills tab
- Verify the assistant dropdown menu no longer has an "AI settings" option

## Pre-merge Checklist


apps/studio/src/components/ai-settings-modal.tsx
```diff
@@ -119,7 +122,7 @@ function AgentInstructionsPanel( { siteId }: { siteId: string } ) {
 										: __( config.description ) }
 								</div>
 							</div>
-							<div className="flex items-center gap-2 flex-shrink-0">
+							<div className={ 'flex items-center gap-2 flex-shrink-0' }>
```
nightnei: NIT - t seems it's leftover
```suggestion
							<div className="flex items-center gap-2 flex-shrink-0">
```

apps/studio/src/modules/agent-skills/types.ts
```diff
@@ -0,0 +1,54 @@
+/**
```
nightnei: We have many redundant comments or info in some comments, for example:
1. `Required:` / `Optional:` - it's just a duplication of info which we already have with type definitions
2. Or for example `/** Skill name */` is redundant comment, sine `name` property inside `AvailableSkill` is self-explanatory. 
So I would remove some comments and adjusted other.

apps/studio/src/modules/agent-skills/lib/constants.ts
```diff
@@ -0,0 +1,11 @@
...
+export const SKILLS_DIRECTORY_PATH = '.claude/skills';
+
+/** Name of the skill definition file */
+export const SKILL_FILE_NAME = 'SKILL.md';
+
+/** Default repository for skills */
+export const DEFAULT_SKILLS_REPO = 'WordPress/agent-skills';
+
+/** Default branch to download from */
+export const DEFAULT_BRANCH = 'trunk';
```
nightnei: NIT - I would keep the name more specific instead of a generic name
```suggestion
export const DEFAULT_SKILLS_REPO_BRANCH = 'trunk';
```

apps/studio/src/modules/agent-skills/components/skills-panel.tsx
```diff
@@ -0,0 +1,235 @@
...
+			<div className="flex-1 min-w-0 pr-3">
+				<div className="text-sm text-gray-900 truncate">{ name }</div>
+				<div className="text-xs text-gray-500 truncate">{ description }</div>
+			</div>
+			<div className="flex items-center gap-2 flex-shrink-0">
+				{ isInstalled ? (
+					<>
+						<span className="inline-flex items-center gap-1 text-[11px] text-green-700 bg-green-50 px-2 py-0.5 rounded-full">
+							<Icon icon={ check } size={ 12 } />
+							{ __( 'Installed' ) }
```
nightnei: Personal feeling - we have different lists for already installed and not installed skills, so I find this "Installed" badge redundant. Especially when we have a few installed Skills.

Image

apps/studio/src/modules/agent-skills/components/skills-panel.tsx
```diff
@@ -0,0 +1,235 @@
...
+						</Button>
+					</>
+				) : (
+					<Button
+						variant="secondary"
+						onClick={ onInstall }
+						disabled={ isInstalling }
+						className="text-xs py-1 px-2"
+					>
+						{ isInstalling ? __( 'Installing...' ) : __( 'Install' ) }
```
nightnei: Small bug - when you quickly click on the "Install" button and then immediately click on another "Install" button, the disabled state and "Installing" text disappear for the previous item, which was in progress.

apps/studio/src/modules/agent-skills/components/skills-panel.tsx
```diff
@@ -0,0 +1,235 @@
...
+	const { installSkill, isInstalling, installError } = useInstallSkill( siteId, () => {
+		void refreshInstalled();
+	} );
+
+	const [ removeError, setRemoveError ] = useState< string | null >( null );
+	const [ removingSkill, setRemovingSkill ] = useState< string | null >( null );
+	const [ installingPath, setInstallingPath ] = useState< string | null >( null );
+
+	const installedNames = useMemo(
+		() => new Set( installedSkills.map( ( s ) => s.name ) ),
```
nightnei: Just out of curiosity - is there some reason for using `Set`?

nightnei: Also, `map` + `Set` is cheap for normal list sizes and the same for `filter` inside `uninstalledSkills`, so I see it just as unnecessary lines of code, but it's completely up to you, just sharing my thoughts.

apps/studio/src/modules/agent-skills/lib/skill-installer.ts
```diff
@@ -0,0 +1,195 @@
...
+export async function removeSkill( sitePath: string, skillName: string ): Promise< void > {
+	if ( ! ( await skillExists( sitePath, skillName ) ) ) {
+		throw new Error( `Skill "${ skillName }" is not installed` );
+	}
+
+	await fs.rm( getSkillPath( sitePath, skillName ), { recursive: true, force: true } );
+}
+
+export async function listAvailableSkills(
+	repo: string = DEFAULT_SKILLS_REPO,
```
nightnei: I see that we use the `repo` property everywhere. I don't see an opportunity to provide a custom repo in UI, so I wonder - do we expect to provide an opportunity for different/custom repos? If no, then I would go with removing this property in all places, to keep code simple, at this point. The same with `branch`.

apps/studio/src/modules/agent-skills/components/skills-panel.tsx
```diff
@@ -0,0 +1,235 @@
...
+							variant="link"
+							onClick={ async () => {
+								for ( const skill of uninstalledSkills ) {
+									await handleInstall( skill.path );
+								}
+							} }
+							disabled={ isInstalling }
+							className="text-[11px] !py-0 !px-0 normal-case !font-medium"
+						>
+							{ isInstalling ? __( 'Installing...' ) : __( 'Install All' ) }
```
nightnei: When we click "Install All' - let's disable all `Install` buttons.

nightnei (commented):
I tried a few times using `Install All` button and at some moment it fails always with:
Image

Does it work well for you?

katinthehatsite:
> I tried a few times using Install All button and at some moment it fails always with:

@nightnei Let's hold off with this PR for now. I think we will implement https://github.com/Automattic/studio/pull/2741 first as the base for the backend and then rebase this one to the UI based on the discussion here https://github.com/Automattic/studio/pull/2741#issuecomment-4046138775

I will mark this as draft for now to avoid confusion.

apps/studio/src/components/ai-settings-modal.tsx
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2729/changes/1c3bf117f182502e14079c15dfea1080ce63f6e0

katinthehatsite:
@nightnei this is now available for review, thank you!

nightnei:
Hm, again I have this problem when installing doesn't work 🤔 cc @sejas @epeicher 
What I did:
1. npm i
2. npm start
3. And just created a new site

On Monday, I am planning to debug my environment and find the root cause, since it seems that we need to add error handling in some places and anyway it's good to know why it doesn't work for me second time.

The code LGTM, but I would ask somebody else to test it, to ensure that it works for other folks. Feel free to not wait for my approval, most likelly it's my local issue.
https://github.com/user-attachments/assets/d8f6ba16-abf4-4e82-b06b-6ae8ca366c33

epeicher:
> Hm, again I have this problem when installing doesn't work 🤔 cc @sejas @epeicher What I did:

Hi @nightnei, the latest code changes included additional error handling, for example https://github.com/Automattic/studio/commit/02d4a71a21313a594495038f758cbb781fc51440, but it definitely requires some further digging. Please let us know your findings.

sejas (commented):
Kat, this is looking great. Let's focus in global settings first and then resolve conflicts here.

katinthehatsite:
> Kat, this is looking great. Let's focus in global settings first and then resolve conflicts here.

Sounds good to me. From what I understand, we should first merge https://github.com/Automattic/studio/pull/2880 and then refactor based on that, is that right?

sejas:
Yes! that will be great

sejas:
I updated this PR to follow the same design as the Global Studio Settings. Here is how is looking so far:

https://github.com/user-attachments/assets/d7b18b3a-57a3-444b-99c2-d358dfc06dbe

katinthehatsite:
Thanks for making the updates @sejas ! One thing that I am wondering is if the global or individual settings should take precedence for users for existing sites? For example, if I globally remove a skill, it also gets removed from an existing site or if I globally add a skill, it gets added to existing sites but it might mess with the specific settings the users set up per site. What do you think?

sejas:
I think it's ok that latest action overrides what's the current state. Global modifies all sites, local modifies only your that site. We could set the expectations with better descriptions. What do you think? Would you like to review the descriptions?

katinthehatsite:
>  I think it's ok that latest action overrides what's the current state. Global modifies all sites, local modifies only your  that site. We could set the expectations with better descriptions. What do you think? Would you like to review the descriptions?

I looked into it a bit and I think it looks fine. I was working on documentation today and wrote about it there as well so I think we are okay to merged as is.

PR #2730, opened 2026-03-09
Fix openFileInIDE to respect user's preferred editor

sejas:
## Related issues

- Fixes STU-1387

## Proposed Changes

The `openFileInIDE` IPC handler was hardcoded to only check for VS Code and PhpStorm, ignoring the user's preferred editor setting configured in Settings > Preferences.

This PR updates the handler to:
- Read the user's preferred editor from settings via `getUserEditor()`
- Improves getUserEditor to fallback to first detected installed editor
- Tries to open the site folder and the file in a single system call, with the exception of PhpStorm on Mac, in that case it opens the single file.
- Do nothing gracefully if no editor is installed

All 8 supported editors (Antigravity, Cursor, VS Code, PhpStorm, Windsurf, WebStorm, Sublime Text, Zed) now work with the "open file in IDE" feature.

## Testing Instructions

Apply this diff to test it:
```diff
diff --git a/apps/studio/src/components/content-tab-overview.tsx b/apps/studio/src/components/content-tab-overview.tsx
index 6055a8931..3ca26cbde 100644
--- a/apps/studio/src/components/content-tab-overview.tsx
+++ b/apps/studio/src/components/content-tab-overview.tsx
@@ -9,6 +9,7 @@ import {
 	navigation,
 	page,
 	preformatted,
+	settings,
 	styles,
 	symbolFilled,
 	widget,
@@ -162,6 +163,15 @@ function ShortcutsSection( { selectedSite }: Pick< ContentTabOverviewProps, 'sel
 		} );
 	}
 
+	buttonsArray.push( {
+		label: __( 'wp-config.php' ),
+		className: 'text-nowrap',
+		icon: settings,
+		onClick: () => {
+			getIpcApi().openFileInIDE( 'wp-config.php', selectedSite.id );
+		},
+	} );
+
 	const terminalName = getTerminalName( terminal );
 	buttonsArray.push( {
 		label: terminalName,
```

1. Start the app with `npm start`
2. Open Settings > Preferences and select a preferred editor (e.g., Cursor, Sublime Text)
3. Open the AI assistant, ask it to create/edit a file
4. Click the file path in the assistant response
5. Verify the file opens in your selected editor, not just VS Code/PhpStorm
6. Change the preferred editor and repeat — it should respect the new selection
7. Remove the preferred editor setting and verify it falls back to the first installed editor

Test on Mac

https://github.com/user-attachments/assets/d54799f6-d280-4e9c-9a7b-8161c2541615

Test on Windows

https://github.com/user-attachments/assets/6cb4cef1-43c1-44e1-a702-ae2773ff76ef

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/ipc-handlers.ts
+++ b/apps/studio/src/ipc-handlers.ts
@@ -1047,7 +1047,6 @@ export async function openAppAtPath(
 
 	if ( platform === 'darwin' ) {
 		const cmd = `open -b ${ editor.macOSBundleId } ${ quotedPaths }`;
-		console.log( '---->: ', cmd );
 		return promiseExec( cmd );
 	}
 
--- a/apps/studio/src/ipc-handlers.ts
+++ b/apps/studio/src/ipc-handlers.ts
@@ -56,7 +56,6 @@ import { ExportOptions } from 'src/lib/import-export/export/types';
 import { ImportExportEventData } from 'src/lib/import-export/handle-events';
 import { defaultImporterOptions, importBackup } from 'src/lib/import-export/import/import-manager';
 import { BackupArchiveInfo } from 'src/lib/import-export/import/types';
-import { isInstalled } from 'src/lib/is-installed';
 import { getUserLocaleWithFallback } from 'src/lib/locale-node';
 import * as oauthClient from 'src/lib/oauth';
 import { shellOpenExternalWrapper } from 'src/lib/shell-open-external-wrapper';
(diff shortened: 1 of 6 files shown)


epeicher (commented):
Thanks @sejas, for solving this! I have tested it, and it works perfectly on Mac. I have tested with VSCode, Cursor, Zed, and PhpStorm, and it opens fine on all of them. On my version of PhpStorm it does not open the folder, but I don't think that's an issue.

On Windows also works fine, but I have identified that VSCode and Cursor open the folder, and when closing them, they open again with only the file. I have recorded a video with that:

https://github.com/user-attachments/assets/844c9a16-8b4b-4451-935c-dddecb027ce8

I like the behaviour of opening the file in the proper folder, which works on Mac, but on Windows, it is confusing. Maybe we can add a condition and open the folder only for Mac? What do you think?

katinthehatsite (commented):
On macOS, the changes worked as expected for me and I did not see any further regressions. I think we should address Roberto's comments before moving forward 👀

apps/studio/src/ipc-handlers.ts
```diff
@@ -1163,11 +1167,21 @@ export async function getAbsolutePathFromSite(
 	return ( await pathExists( path ) ) ? path : null;
 }
 
+function getFirstInstalledEditor(): SupportedEditor | null {
```
nightnei: NIT I would rename it to somethign like "random" or "any" instead of "firstInstalled", since we use hardcoded predefined order of editors, so the name is a bit confusing and not relevant to the code.

nightnei (approved):
LGTM and works as expected 👍
Left only oen comment about adjusting `getFirstInstalledEditor `

apps/studio/src/ipc-handlers.ts
sejas: I removed the function here: https://github.com/Automattic/studio/pull/2730/changes/f733f82660b5a72e97b1f7eba099f91280d47b1a
and added the logic directly inside getUserEditor https://github.com/Automattic/studio/pull/2730/changes/10381a7a75a4fec3fe75c87d19bea13111bf90d4

PR #2731, opened 2026-03-09
Move CLI site data to dedicated config file

bcotrim:
## Related issues

- Related to STU-1350

## How AI was used in this PR

AI was used to do most of the implementation, given a detailed plan.
I've tested the changes manually, making some smaller adjustments as needed.

## Proposed Changes

Splits CLI site configuration from `appdata-v1.json` into a dedicated `~/.studio/cli.json` file. Site data is now CLI-owned; Studio reads sites exclusively via `studio site list` and `_events` command, making the CLI the single source of truth.

**Key changes:**
- New `apps/cli/lib/cli-config.ts` module handles CLI-specific site config (version 1, sites array)
- Moved site helper functions: `getSiteByFolder`, `getSiteUrl`, `updateSiteLatestCliPid`, `clearSiteLatestCliPid`, `updateSiteAutoStart`, `removeSite`
- Appdata now stores only: auth tokens, snapshots, locale, stats, beta features
- Updated all site commands, utilities, and tests to use cli-config module
- Separate lockfile at `~/.studio/cli.json.lock` prevents blocking between site and auth operations

**Architecture benefits:**
- Single source of truth: CLI owns site configuration
- Reduced data corruption risk: less file I/O on appdata during site operations
- Decoupled lifetime: site config persists independently of Studio internal state
- Type-safe: Zod validation ensures schema compliance across versions

## Testing Instructions

1. Run all CLI tests: `npm test -- apps/cli/`
   - All 277 tests should pass
2. Verify site commands work correctly:
   ```bash
   npm run cli:build
   node dist/cli/main.js site create --name "test-site"
   node dist/cli/main.js site list --format json
   node dist/cli/main.js site delete
   ```
3. Check that sites are now stored in `~/.studio/cli.json` instead of appdata
4. Verify appdata still contains auth token and other non-site data

## Pre-merge Checklist


apps/cli/lib/cli-config.ts
```diff
@@ -0,0 +1,190 @@
...
+
+export function getCliConfigPath(): string {
+	return path.join( getCliConfigDirectory(), 'cli.json' );
+}
+
+export async function readCliConfig(): Promise< CliConfig > {
+	const configPath = getCliConfigPath();
+
+	if ( ! fs.existsSync( configPath ) ) {
+		return { ...DEFAULT_CLI_CONFIG };
```
fredrikekelund:
```suggestion
		return structuredClone( DEFAULT_CLI_CONFIG );
```
CLI commands are relatively short-lived, so mutating global state isn't really a big concern. Still, `DEFAULT_CLI_CONFIG.sites` is an array, and it can be mutated. Might as well return a truly fresh copy.

apps/cli/lib/cli-config.ts
```diff
@@ -0,0 +1,190 @@
...
+	sites: [],
+};
+
+export function getCliConfigDirectory(): string {
+	if ( process.env.E2E && process.env.E2E_CLI_CONFIG_PATH ) {
+		return process.env.E2E_CLI_CONFIG_PATH;
+	}
+
+	return path.join( os.homedir(), '.studio' );
+}
```
fredrikekelund, line 43: We should have a global definition of `~/.studio` as the base path, IMO. Maybe consider renaming `apps/cli/lib/daemon-paths.ts` to `apps/cli/lib/paths.ts` and define it in there..?

apps/cli/lib/cli-config.ts
```diff
@@ -0,0 +1,190 @@
...
+import { siteDetailsSchema } from '@studio/common/lib/site-events';
+import { __ } from '@wordpress/i18n';
+import { readFile, writeFile } from 'atomically';
+import { z } from 'zod';
+import { LoggerError } from 'cli/logger';
+
+const siteSchema = siteDetailsSchema
+	.extend( {
+		url: z.string().optional(),
+		latestCliPid: z.number().optional(),
```
fredrikekelund, line 16: This should be part of the `siteDetailsSchema` definition, IMO. I see no reason why events emitted from `_events` shouldn't include these props, and it will help avoid confusion if we have as few schemas as possible for site details objects.

apps/cli/lib/cli-config.ts
```diff
@@ -0,0 +1,190 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { LOCKFILE_STALE_TIME, LOCKFILE_WAIT_TIME } from '@studio/common/constants';
+import { arePathsEqual, isWordPressDirectory } from '@studio/common/lib/fs-utils';
+import { lockFileAsync, unlockFileAsync } from '@studio/common/lib/lockfile';
+import { siteDetailsSchema } from '@studio/common/lib/site-events';
```
fredrikekelund, line 6: If this schema defines the shape of site objects stored in the config file, then I would put it in a different file. This is just a naming thing, but it seems more important to indicate that this is the storage schema than it being the event schema.

apps/cli/lib/appdata.ts
```diff
@@ -14,20 +12,10 @@ import { z } from 'zod';
...
-	.loose();
-
 const betaFeaturesSchema = z.object( {} ).loose();
 
 const userDataSchema = z
 	.object( {
-		sites: z.array( siteSchema ).default( () => [] ),
 		snapshots: z.array( snapshotSchema ).default( () => [] ),
 		locale: z.string().optional(),
 		authToken: z
```
fredrikekelund, line 24: These three properties also need to be shared between the CLI and the app. Are we moving those in a follow-up PR?

apps/cli/lib/cli-config.ts
```diff
@@ -0,0 +1,190 @@
...
+		}
+
+		site.autoStart = autoStart;
+		await saveCliConfig( config );
+	} finally {
+		await unlockCliConfig();
+	}
+}
+
+export async function removeSite( siteId: string ): Promise< void > {
```
fredrikekelund:
```suggestion
export async function removeSiteFromConfig( siteId: string ): Promise< void > {
```
I think it adds clarity with a more verbose name here.

fredrikekelund:
```suggestion
export async function removeSite( siteFolder: string ): Promise< void > {
```
Also, I know this function predates this PR, but it'd be nice to change it to accept a directory instead. That way, we could reuse it in `apps/cli/commands/site/delete.ts`. It's a little funny that we're only using it in `apps/cli/commands/site/create.ts` now. 

We could pursue this in another PR.

apps/cli/lib/cli-config.ts
```diff
@@ -0,0 +1,190 @@
...
+	.extend( {
+		url: z.string().optional(),
+		latestCliPid: z.number().optional(),
+	} )
+	.loose();
+
+const cliConfigSchema = z.object( {
+	version: z.number().default( 1 ),
+	sites: z.array( siteSchema ).default( () => [] ),
+} );
```
fredrikekelund:
```suggestion
const cliConfigWithJustVersion = z.object( {
	version: z.number().default( 1 ),
} );
// IMPORTANT: Always consider that independently installed versions of the CLI (from npm) may also
// read this file, and any updates to this schema may require updating the `version` field.
const cliConfigSchema = cliConfigWithJustVersion.extend( {
	sites: z.array( siteSchema ).default( () => [] ),
} );
```
I had something like this in mind for defining the config schemas. The purpose of the comment is obvious, I guess, but more on the need for `cliConfigWithJustVersion` below…

apps/cli/lib/cli-config.ts
```diff
@@ -0,0 +1,190 @@
...
+		if ( error instanceof z.ZodError ) {
+			throw new LoggerError( __( 'Invalid CLI config file format.' ), error );
+		}
+
+		if ( error instanceof SyntaxError ) {
+			throw new LoggerError( __( 'CLI config file is corrupted.' ), error );
+		}
+
+		throw new LoggerError( __( 'Failed to read CLI config file.' ), error );
+	}
```
fredrikekelund, line 87:
```suggestion
	try {
		const fileContent = await readFile( configPath, { encoding: 'utf8' } );
		// eslint-disable-next-line no-var
		var data = JSON.parse( fileContent );
	} catch ( error ) {
		throw new LoggerError( __( 'Failed to read CLI config file.' ), error );
	}

	try {
		return cliConfigSchema.parse( data );
	} catch ( error ) {
		if ( error instanceof z.ZodError ) {
			try {
				cliConfigWithJustVersion.parse( data );
			} catch ( versionError ) {
				throw new LoggerError(
					__(
						'Invalid CLI config version. It looks like you have a different version of the `studio` CLI installed on your system. Please modify your $PATH environment variable to use the correct version.'
					),
					error
				);
			}

			throw new LoggerError( __( 'Invalid CLI config file format.' ), error );
		}

		if ( error instanceof SyntaxError ) {
			throw new LoggerError( __( 'CLI config file is corrupted.' ), error );
		}

		throw new LoggerError( __( 'Failed to read CLI config file.' ), error );
	}
```
If parsing the schema fails, then we should check just the `version` field to see if it differs from what we expect. That way, we can provide a more helpful error message to users. There's room for improvement in the actual error message – this is mostly focusing on the logic part.

apps/studio/src/ipc-handlers.ts
```diff
@@ -600,6 +600,19 @@ export async function deleteSite( event: IpcMainInvokeEvent, id: string, deleteF
...
+		await lockAppdata();
+		const userData = await loadUserData();
+		const siteIndex = userData.sites.findIndex( ( s ) => s.id === id );
+		if ( siteIndex !== -1 ) {
+			userData.sites.splice( siteIndex, 1 );
+			await saveUserData( userData );
+		}
+	} finally {
+		await unlockAppdata();
+	}
```
fredrikekelund, line 614: This looks like something we removed in https://github.com/Automattic/studio/pull/2701

apps/studio/src/modules/cli/lib/cli-events-subscriber.ts
```diff
@@ -46,10 +46,11 @@ const handleSiteEvent = sequential( async ( event: SiteEvent ): Promise< void >
 		return;
 	}
 
-	// For UPDATED events, update existing server details
+	// For UPDATED events, update existing server or register new one (e.g., during startup)
 	const server = SiteServer.get( siteId ) ?? SiteServer.getByPath( site.path );
 	if ( ! server ) {
-		console.warn( `Received UPDATED event for unknown site: ${ siteId }` );
+		SiteServer.register( siteDetailsToServerDetails( site, running ) );
+		void sendIpcEventToRenderer( 'site-event', event );
```
fredrikekelund: This also looks like something we removed in https://github.com/Automattic/studio/pull/2701

fredrikekelund (commented):
Looks like we need to update some code in the `ai` commands directory to also read the sites from `cli-config`. 

I'm about to test this more thoroughly, but leaving my initial round of feedback for you to chew on in the meantime, @bcotrim 😄

apps/cli/lib/cli-config.ts
fredrikekelund: I guess `url` is maybe derived at runtime. Then we could have a base `siteDetailsSchema` and a second `siteDetailsEventSchema` that excludes the `url` prop.

fredrikekelund:
I guess we also need to update the base branch on this PR, @bcotrim. It's still using stu-1350-decoupled-config-dev

fredrikekelund:
> base branch

Oh wait, we said we'd keep things in a dev branch until we figured out the data migration story, etc. Never mind my previous comment…

fredrikekelund (approved):
This tests well 👍 The error handling in `readCliConfig` is important to get right, but since we're merging this to a dev branch, I'll approve now to unblock you, @bcotrim

bcotrim:
@fredrikekelund thanks for the review and feedback! I addressed most of it, the schema location/naming, removeSiteFromConfig accepting a folder, and the appdata site cleanup in ipc-handlers are tracked for follow-up PRs. Merging this one to keep us moving forward.

PR #81, opened 2024-05-02
Update dependabot dependencies

sejas:
Related to p1714650498693029-slack-C06DRMD6VPZ

## Proposed Changes

- Update dependencies suggested by [dependabot](https://github.com/Automattic/studio/security/dependabot) except lodash

## Testing Instructions

- Run `npm install` to use the new dependnecies
- Run `npm start` to start Studio
- Make an holistic test in all the features, creating a site, creating a demo site.
- Observe the app runs correclty

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2732, opened 2026-03-10
Remove express dependency

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

I used AI to search for `express` usage to validate the changes.

## Proposed Changes

I propose removing `express` dependency from Studio app. While reviewing https://github.com/Automattic/studio/pull/2718 I noticed this is installed, but not even used.

We've added this dependency in https://github.com/Automattic/studio/pull/81, and it's unclear why. A possible explanation is that we wanted to update `express`, the other package dependency, from 4.18.2 to 4.19.2, and we added it as an explicit dependency in the main node. Now, all dependencies resolve to 4.22.0, so we should be good.

## Testing Instructions

- Confirm builds go through.

## Pre-merge Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -82,7 +82,6 @@ updates:
       - dependency-name: "tar"
 
       # HTTP & networking
-      - dependency-name: "express"
       - dependency-name: "http-proxy"
       - dependency-name: "hpagent"
       - dependency-name: "follow-redirects"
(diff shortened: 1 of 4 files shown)


ivan-ottinger (approved):
Looks good. 👍🏼 Nice catch! I don't see any reason to keep it in.

PR #2738, opened 2026-03-10
Add dev-only Open Appdata option in Help menu

bcotrim:
## Related issues

- N/A

## How AI was used in this PR

One shot prompt, tested manually

## Proposed Changes

- Added "Open Appdata File" menu item in the Help menu
- Menu item only appears in development mode (`NODE_ENV === 'development'`)
- Clicking the menu item opens `appdata-v1.json` file with the system's default application
- Added import of `getUserDataFilePath` from storage paths module

image

## Testing Instructions

1. Run Studio in development mode (`npm start`)
2. Open the Help menu
3. Verify "Open Appdata File" menu item appears (below "Open Application Logs")
4. Click on "Open Appdata File"
5. Verify the appdata-v1.json file opens in your default editor
6. Build for production and verify the menu item does NOT appear in Help menu

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/menu.ts
+++ b/apps/studio/src/menu.ts
@@ -178,6 +178,16 @@ async function getAppMenu(
 								label: __( 'Feature Flags' ),
 								submenu: featureFlagsMenu,
 							},
+							{
+								label: __( 'Open Appdata File (dev only)' ),
+								click: async () => {
+									const appdataPath = getUserDataFilePath();
+									const err = await shell.openPath( appdataPath );
+									if ( err ) {
+										console.error( `Error opening appdata file: ${ appdataPath } ${ err }` );
+									}
+								},
+							},
 					  ]
 					: [] ),
 				{ type: 'separator' },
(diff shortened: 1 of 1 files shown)


wojtekn:
Is it safe to manually edit the config file when the app is running?

bcotrim:
> Is it safe to manually edit the config file when the app is running?

I rarely edit it, I often use it during development to confirm that the UI is reflecting the correct state in appdata.

katinthehatsite (approved):
This is great and personally, for me, this option is really useful. I sometimes need a quick access to test things and always have to run a manual command to open appData 👍

katinthehatsite:
> Is it safe to manually edit the config file when the app is running?

Not editing but checking the data is useful :)

wojtekn:
Okay, so if we shouldn't edit it for running the app, it's weird to encourage developers to open it in the editor. But if it's helpful for some, let's add in.

I would just propose reorganizing those to avoid confusion. "Open Application Logs" is not a dev-only item, and we have some dev-only items in the first menu: Test Hard Crash, etc. Can we place a new link in the same place, and add a (dev only) label to make those consistent? Or maybe it's the time to introduce a menu group for those, similar to one we have for Feature Flags?

bcotrim:
> I would just propose reorganizing those to avoid confusion. "Open Application Logs" is not a dev-only item, and we have some dev-only items in the first menu: Test Hard Crash, etc. Can we place a new link in the same place, and add a (dev only) label to make those consistent? Or maybe it's the time to introduce a menu group for those, similar to one we have for Feature Flags?

Done, nice suggestion!

PR #2743, opened 2026-03-11
Refine Studio AI tool transcript collapsing

nerrad:
## Related issues

- Related to #

## How AI was used in this PR

Used Codex to iterate on the Studio AI CLI transcript behavior, then reviewed the final diff and verified the resulting changes locally.

## Proposed Changes

- rework the Studio AI CLI tool transcript so completed tool calls are grouped into checkpoints that can collapse beneath the latest assistant output
- add summarized and hidden transcript views for larger tool sections, toggled with `ctrl + e`
- keep active and completed tool-call state attached to the correct invocation so labels, progress lines, and result rows stay grouped correctly

## State Reference

Summarized checkpoint:

```text
⎿ Tools 1-5  ▾
  ⏺ Read (foo.ts)
  ⏺ Edit (foo.ts) x3
  ⏺ Bash (npm test)
```

Expanded checkpoint:

```text
⎿ Tools 1-5  ◂
 ⏺ Read (foo.ts)
   ⎿ read file contents
 ⏺ Edit (foo.ts)
   ⎿ updated import path
 ⏺ Bash (npm test)
   ⎿ 12 tests passed
```

Hidden checkpoint:

```text
⎿ Tools 1-5  ▸
```

Active tool row:

```text
 ⏺ Edit (foo.ts)
```

Completed tool row:

```text
 ⏺ Edit (foo.ts) (0.4s)
   ⎿ updated import path
```

## Testing Instructions

- Build the CLI with `npm run cli:build`.
- Start the AI CLI with `ENABLE_STUDIO_AI=true node apps/cli/dist/cli/main.js ai`.
- Run a prompt that triggers several tool calls, including repeated calls.
- Confirm the transcript keeps the main assistant text visible while older tool rows collapse into a single `⎿ Tools 1-N` section.
- Confirm `ctrl + e` cycles the tool section view modes and that the chosen view continues to apply as additional tool calls complete.
- Confirm repeated identical tool calls are grouped, progress output stays nested under the correct tool row, and tool labels remain correct as results are added.
- Verify tool dots pulse while tools are running, render green on success, and red on error.
- Run `npx eslint --fix apps/cli/ai/ui.ts`
- Run `npm run typecheck`

## Pre-merge Checklist


youknowriad:
Maybe I'm not testing this properly but I find the trunk behavior better in terms of output than this PR. I don't really like the grayed color too much. The "⎿ " feels a bit out of place compared to the rest of the output as well.

WDYT?

youknowriad:
Is the Ctrl+e a common shortcut for this? The behavior is weird too, it collapses/uncollapses all the previous tool calls even previous prompts

nerrad:
> Maybe I'm not testing this properly but I find the trunk behavior better in terms of output than this PR. I don't really like the grayed color too much. The "⎿ " feels a bit out of place compared to the rest of the output as well.

It's intended to follow the existing indent symbol used in the app. I'll have to take another look maybe it's supposed to be ↳ or  ∟.

What is it you don't like about this. Is it the collapsing? Where it collapses to? Or you want no changes at all? I went off of some loose chat with @mtias and it can be iterated on.

nerrad:
> Is the Ctrl+e a common shortcut for this? The behavior is weird too, it collapses/uncollapses all the previous tool calls even previous prompts

Not that I know of. I used "e" for "expand". It probably should only collapse/uncollapse the tool calls for the most recent prompt only.

youknowriad:
Screenshot 2026-03-12 at 4 09 25 PM

This is what claude code does:

 - uses ctrl+o (I wonder if we should match)
 - The layout of the output is similar without or without ctrl+o, the only difference is that when it's expanded you have additional details under each "tool" call. (parent dot)

**This is condensed**

Screenshot 2026-03-12 at 4 13 07 PM

**This is expanded**

Screenshot 2026-03-12 at 4 13 50 PM

nerrad:
Thanks Riad, I'll implement this in a separate PR. I've been using codex more than claude lately so I forgot about `ctrl-o` 🤦🏻

nerrad:
Closing in favor of https://github.com/Automattic/studio/pull/2743

PR #2745, opened 2026-03-11
Auto-select sites in AI CLI based on agent tool usage

youknowriad:
## Related issues

- Related to the AI CLI command feature

## How AI was used in this PR

Claude planned and implemented the auto-select feature, identified and fixed a related bug with MCP tool result display format handling.

## Proposed Changes

- Auto-select sites when the agent creates, starts, or interacts with them via `site_create`, `site_start`, `wp_cli`, `site_stop`, and `site_delete` tools
- Refactored site selection logic into `setActiveSite()` and `clearActiveSite()` methods for reuse between manual picker and auto-select
- Added `findSiteFromAppdata()` helper to reliably resolve sites by name or path and get accurate running status
- Fixed bug where MCP tool results weren't displayed because `content` format wasn't handled (can be string or array)

## Testing Instructions

- Rebuild CLI: `npm run cli:build`
- Enable and test: `ENABLE_STUDIO_AI=true node dist/cli/main.js ai`
- Create a site: Ask agent "create a site called Test Auto" → verify site appears in prompt border
- Start a site: Ask agent "start the site" → verify running status updates
- Run commands: Ask agent to run WP-CLI commands → verify site stays selected
- Delete a site: Ask agent "delete the site" → verify site is deselected
- Verify tool results display correctly for all site operations

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/ui.ts
+++ b/apps/cli/ai/ui.ts
@@ -274,7 +274,10 @@ export class AiChatUI {
 	private toolDotLabel = '';
 	private _activeSite: SiteInfo | null = null;
 	private _activeSiteData: SiteData | null = null;
-	private lastToolInput: Record< string, unknown > | null = null;
+	private pendingToolCalls = new Map<
+		string,
+		{ name: string; input: Record< string, unknown > }
+	>();
 	currentModel: AiModelId = DEFAULT_MODEL;
 
 	private readonly thinkingMessages = [
@@ -898,6 +901,7 @@ export class AiChatUI {
 		this.stopToolDotBlink();
 		this.toolDotText = null;
 		this.interruptCallback = null;
+		this.pendingToolCalls.clear();
 		this.updateHints();
 		this.currentMarkdown = null;
 		this.currentResponseText = '';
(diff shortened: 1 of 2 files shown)


apps/cli/ai/ui.ts
```diff
@@ -923,6 +1058,7 @@ export class AiChatUI {
 						this.lastToolName = block.name;
 						this.toolStartTime = Date.now();
 						const input = ( block as { input?: Record< string, unknown > } ).input;
+						this.lastToolInput = input ?? null;
```
nerrad: codex flagged this 
> The new auto-select path relies on lastToolName/lastToolInput, which are overwritten for every tool_use block and then consumed by the next user message. The SDK already gives each tool result a parent_tool_use_id, so when Claude emits multiple tool calls before the first result comes back, this code will associate that result with the wrong input and can switch the active site incorrectly. That is more than a display bug now: the next prompt will be prefixed with the wrong [Active site: ...] context, so follow-up WP-CLI calls can target the wrong site.

Suggests using `parent_tool_use_id` from the SDK (from each tool result) which can make sure this is associated with the right tool.

Another question that came to mind (which I haven't answered yet from looking at the changeset) is the likelihood of switching site context too often during a given session. Should there be a prompt to the user when the agent assumes a site switch is needed for some safety?

youknowriad: > Another question that came to mind (which I haven't answered yet from looking at the changeset) is the likelihood of switching site context too often during a given session. Should there be a prompt to the user when the agent assumes a site switch is needed for some safety?

Any precise thing you have in mind, I'm not sure I follow when this can happen...

nerrad: I have a site named "Sublime" and a site named "Purple". I wanted to see if I could fool it with this prompt:

> Okay I'd like to work on the Sublime site. Let's give it a blueish feel. I'd like the header and footer to be Purple. Have the text in the header and footer off white.

Changes were just applied to Purple.  So seems to be okay.

However I also did not see any (`✻ Site deselected`) messages. Not sure if that's expected or not.

nerrad (approved):

apps/cli/ai/ui.ts
youknowriad: I did see the "site deselected" when removing the site.

PR #2746, opened 2026-03-11
Fix studio ai /exit command hanging

nerrad:
## Related issues

- N/A

## How AI was used in this PR

AI identified the root cause and wrote the one-line fix.

## Proposed Changes

- Add `process.exit(0)` after `ui.stop()` in the `/exit` path. The TUI library keeps the event loop alive (raw stdin listeners/timers), so the process hangs after `/exit`. Ctrl+C already called `process.exit(0)` but the `/exit` slash command did not.

## Testing Instructions

- Run `studio ai`, then type `/exit` — process should return to the shell immediately.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/ai.ts
+++ b/apps/cli/commands/ai.ts
@@ -162,6 +162,7 @@ export async function runCommand(): Promise< void > {
 		}
 	} finally {
 		ui.stop();
+		process.exit( 0 );
 	}
 }
 


apps/cli/commands/ai.ts
```diff
@@ -162,6 +162,7 @@ export async function runCommand(): Promise< void > {
 		}
 	} finally {
 		ui.stop();
```
youknowriad, line 164: Do we need this at all?

nerrad: ctrl-c does the same thing, I *think* it's needed for a clean exit from the tui.

nerrad: The question is prompting a closer look for the flow though.

youknowriad (approved):

apps/cli/commands/ai.ts
nerrad: ya I thought about moving the command into the ui.js so ctrl-c and /exit go through the sam flow but that splits up the commands across multiple locations. This feels cleaner.

PR #2632, opened 2026-02-20
Add `studio ai` CLI command with interactive AI agent

youknowriad:
## Related issues

- Related to AI assistant exploration for Studio CLI

## Proposed Changes

Add a new `studio ai` command — an interactive AI-powered WordPress assistant running in the terminal. Uses the Claude Agent SDK with a pi-tui chat interface. The agent can manage sites, run WP-CLI, create blocks/themes, validate block markup, and take screenshots.

**Feature-flagged**: The `ai` command is hidden behind an environment variable and won't appear in the CLI unless explicitly enabled:

```
ENABLE_STUDIO_AI=true node apps/cli/dist/cli/main.js ai
```

This keeps the command out of the default CLI surface while it's still experimental.

## Testing Instructions

1. `npm install && npm run cli:build`
2. `ENABLE_STUDIO_AI=true node apps/cli/dist/cli/main.js ai`
3. Enter your Anthropic API key when prompted (saved for future use)
4. Try prompts like:
   - "List all my WordPress sites"
   - "Create a custom block plugin called testimonial-card"
   - "Take a screenshot of [site-url]"
5. Use `/model` to switch between Sonnet 4.6 and Opus 4.6
6. Press **Escape** to interrupt, **Ctrl+C** to exit

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2749, opened 2026-03-11
Add /login and /logout slash commands to AI chat

taipeicoder:
## Related issues

- Related to #2632

## How AI was used in this PR

This PR was authored with Claude Code. All code has been reviewed and verified.

## Proposed Changes

- Add `/login` slash command that delegates to the existing `auth login` flow, then confirms login with name and email
- Add `/logout` slash command that delegates to the existing `auth logout` flow
- Show WordPress.com auth status on AI chat startup (logged in as name/email, or prompt to use /login)

Not logged in:
Screenshot 2026-03-11 at 12 10 38 PM

Logged in:
SCR-20260311-kzeg

## Testing Instructions

- Build CLI: `npm run cli:build`
- Run: `ENABLE_STUDIO_AI=true node apps/cli/dist/cli/main.js ai`
- Verify auth status message appears after the welcome banner
- Type `/login` and complete the WordPress.com OAuth flow — confirm "Logged in as Name (email)" appears
- Type `/logout` and confirm successful logout message
- Restart and verify it shows "Not logged in" message

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/ai.ts
+++ b/apps/cli/commands/ai.ts
@@ -5,10 +5,14 @@ import { AI_MODELS, DEFAULT_MODEL, startAiAgent, type AiModelId } from 'cli/ai/a
 import {
 	AI_CHAT_BROWSER_COMMAND,
 	AI_CHAT_EXIT_COMMAND,
+	AI_CHAT_LOGIN_COMMAND,
+	AI_CHAT_LOGOUT_COMMAND,
 	AI_CHAT_MODEL_COMMAND,
 } from 'cli/ai/slash-commands';
 import { AiChatUI } from 'cli/ai/ui';
-import { getAnthropicApiKey, saveAnthropicApiKey } from 'cli/lib/appdata';
+import { runCommand as runLoginCommand } from 'cli/commands/auth/login';
+import { runCommand as runLogoutCommand } from 'cli/commands/auth/logout';
+import { getAnthropicApiKey, getAuthToken, saveAnthropicApiKey } from 'cli/lib/appdata';
 import { Logger, LoggerError, setProgressCallback } from 'cli/logger';
 import { StudioArgv } from 'cli/types';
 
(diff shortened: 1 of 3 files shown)


apps/cli/commands/ai.ts
```diff
@@ -65,6 +69,13 @@ export async function runCommand(): Promise< void > {
 	ui.start();
 	ui.showWelcome();
 
+	try {
+		const token = await getAuthToken();
+		ui.showInfo( __( 'Logged in as ' ) + token.displayName + ' (' + token.email + ')' );
```
tyxla: Instinctively,  we should use `sprintf` call here and incorporate the dynamic values as part of the localized string.

But we likely don't need the localization at all, so I'm fine to remove the `__()` calls altogether. Unless there's a way to run a localized version of the CLI which I'm not aware of.

apps/cli/commands/ai.ts
```diff
@@ -147,6 +158,26 @@ export async function runCommand(): Promise< void > {
...
 
+			if ( trimmedPrompt === AI_CHAT_LOGIN_COMMAND ) {
+				ui.stop();
+				await runLoginCommand();
+				ui.start();
+				try {
+					const token = await getAuthToken();
+					ui.showInfo( __( 'Logged in as ' ) + token.displayName + ' (' + token.email + ')' );
+				} catch {
+					// Login may have failed or been cancelled
```
tyxla: Should we display something in the terminal?

tyxla (commented):
This works well in my testing. Just a few minor things.

apps/cli/commands/ai.ts
taipeicoder: Good catch, I've removed the `__()` and use string template literals instead.

apps/cli/commands/ai.ts
taipeicoder: Address in 3436b1887d1c889123c0aa7eb1268c39bd2e589a.

apps/cli/commands/ai.ts
```diff
@@ -147,6 +158,27 @@ export async function runCommand(): Promise< void > {
...
 
+			if ( trimmedPrompt === AI_CHAT_LOGIN_COMMAND ) {
+				ui.stop();
+				await runLoginCommand();
+				ui.start();
+				try {
+					const token = await getAuthToken();
+					ui.showInfo( `Logged in as ${ token.displayName } (${ token.email })` );
+				} catch {
+					ui.showInfo( 'Login failed or cancelled.' );
```
tyxla: Minor language nit if we're using the American version:

```suggestion
					ui.showInfo( 'Login failed or canceled.' );
```

tyxla (approved):
This tests well and looks good. Thanks! 🚀

PR #2754, opened 2026-03-11
Add file content preview and diff viewer to studio ai TUI

lynnbgriffith:
Show inline previews when the AI agent writes or edits files: new file content with line numbers (green) for Write, and red/green removal/addition diff for Edit. Previews truncate at 10 lines with ctrl+o to expand/collapse. Refactors hint bar to be state-driven via updateHints().

## Related issues

- Fixes #

## Proposed Changes

-

## Testing Instructions

Use the `studio ai` command and generate a theme or other change to cause the agent to write changes to file. 

-

## Pre-merge Checklist


nerrad (approved):
Works well for me. 

`ctrl-o` only expands the most recent diff - which is expected (similar to behaviour in other agent tui's) - bit of an annoyance for me but that could be iterated on in the future.

PR #2758, opened 2026-03-11
Add studio mcp command to settings

nightnei:
## Related issues

- Fixes STU-1388

## Proposed Changes
With this PR we are adding MCP configuration info to Studio Settings dialog:<br />Screenshot 2026-03-11 at 12 36 02

## Testing Instructions
1. Turn on "Enable Agent suite" feature flag
2. Open Settigbns in Studio UI
3. Assert that you see the config and can copy it with the button<br />Screenshot 2026-03-11 at 12 36 02

## Pre-merge Checklist


katinthehatsite (approved):
The changes look good to me, I did not spot any further regressions 👍 

Image

apps/studio/src/modules/mcp/lib/ipc-handlers.ts
```diff
@@ -0,0 +1,5 @@
+import { getMcpServerConfigJson } from '@studio/common/lib/mcp-config';
+
+export async function getMcpServerConfig(): Promise< string > {
+	return getMcpServerConfigJson();
+}
```
sejas: What do you think if we create the `modules/ai/lib` ? I think that way we could add or move other functions there.

apps/studio/src/modules/mcp/components/mcp-settings.tsx
```diff
@@ -0,0 +1,52 @@
...
+export function McpSettings() {
+	const { __ } = useI18n();
+	const [ configJson, setConfigJson ] = useState( '' );
+	const [ copied, setCopied ] = useState( false );
+
+	useEffect( () => {
+		getIpcApi()
+			.getMcpServerConfig()
+			.then( setConfigJson )
+			.catch( () => {} );
```
sejas: I don't see how this call could fail, but what do you think about adding at least a console.error?

apps/studio/src/modules/mcp/components/mcp-settings.tsx
```diff
@@ -0,0 +1,52 @@
...
+	}, [] );
```
sejas: Do you think we should add a cancel strategy ?

```suggestion
	useEffect( () => {
		let cancelled = false;
		getIpcApi()
			.getMcpServerConfig()
			.then( ( config ) => {
				if ( ! cancelled ) {
					setConfigJson( config );
				}
			} )
			.catch( () => {} );
		return () => {
			cancelled = true;
		};
	}, [] );
```

sejas (approved):
Thanks for adding the MCP settings to the global settings. We might move it to its own AI tab.
I left some suggestions to consider.

Image

As a follow up, how do you think we could improve it? Should we generate a `MCPB` file that the user can download and "install" ? Should we display other JSON settings?

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
```diff
@@ -76,7 +79,10 @@ export const PreferencesTab = ( { onClose }: { onClose: () => void } ) => {
 			/>
 			<TerminalPicker value={ terminalSelection } onChange={ setDirtyTerminal } />
 			{ ! isWindowsStore() && (
-				<StudioCliToggle value={ isCliInstalledSelection } onChange={ setDirtyIsCliInstalled } />
+				<>
+					<StudioCliToggle value={ isCliInstalledSelection } onChange={ setDirtyIsCliInstalled } />
+					{ enableAgentSuite && <McpSettings /> }
+				</>
```
sejas, line 85: Let's consider moving the MCP settings outside the `isWindowsStore` block.

```suggestion
				<>
					<StudioCliToggle value={ isCliInstalledSelection } onChange={ setDirtyIsCliInstalled } />
				</>
				{ enableAgentSuite && <McpSettings /> }
```

sejas (commented):
~~I've found a bug. The McpSettings should be visible for any platform.~~
UPDATE: Actually, it makes sense for it to be inside that block. However, in the near future, the CLI could be installed separately, and the toggle to install it will no longer be necessary.

nightnei:
> As a follow up, how do you think we could improve it? Should we generate a MCPB file that the user can download and "install" ?

@sejas 

I think that we should follow a similar approach to [ContextA8C MCP](https://mc.a8c.com/ai/context-a8c/setup.php?client=vscode) - opportunity to customize most popular AI Assistants with MCPB, etc., and as the default way, we should keep this manual configuration.
So I propose to merge current PRs which allow to configure manually any AI Assistant and create Linear issues to generate MCPB etc. I consider it as, nice to have, user's experience improvement. And I will start working on the POC proposition today.

apps/studio/src/modules/mcp/lib/ipc-handlers.ts
nightnei: Right now we have just one function and it's used specifically for MCP, so I would be specific and leave the module name as "mcp", but later, when it's necessary - we can easely rename/move it. But it's not a strong preference.

apps/studio/src/modules/mcp/components/mcp-settings.tsx
nightnei: Both points make sense 👍 Added

apps/studio/src/modules/mcp/components/mcp-settings.tsx
nightnei: Hm, on the one hand it's nice to have, but on teh other hand this function is literally sync function since it generates json and returns it w/o delays. So I think that this approach is a bit extra and just adds lines code. I just don't see right now how a user can open teh fdialog and so quickly close it. However, I am okay with adding. WDYT?

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
nightnei: > However, in the near future, the CLI could be installed separately, and the toggle to install it will no longer be necessary.

Hm, maybe I am missing something but AFAIU - `isWindowsStore ` checks that process of installation. So if it's installed from Windows Store then we still should hide the CLI and MCP. But if user installed CLI as a standalone package on Windows then I assume that isWindowsStore will be `false` and it means that teh currect code and condition work correctly. Do I miss something or misunderstood the condition?

apps/studio/src/modules/mcp/components/mcp-settings.tsx
sejas: I think you make a valid point, although it also brings another questions. Maybe we don't need to use `getIpcApi` async call and we could just use a regular function?

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
sejas: No need to take any action. My comment is related that the team is planning to remove this toggle. See p1773128873542449/1773080471.191609-slack-C06DRMD6VPZ

apps/studio/src/modules/mcp/components/mcp-settings.tsx
nightnei: It's very good question 😁 We indeed don't use any node functions and it's pure function so we can use it directly. Addressed, thanks, Antonio.

apps/studio/src/modules/user-settings/components/preferences-tab.tsx
nightnei: I clarified with Fredrik and indeed - thanks to [this PR](https://github.com/Automattic/studio/pull/2547), the CLI will be available also with Studio from MS. 
Thanks Antonio 👍

sejas (approved):
Thanks for the adjustments!

PR #2759, opened 2026-03-11
Remove provider-constants-slice.ts in favor of direct constant imports

bcotrim:
## Related issues

- Fixes STU-1378

## How AI was used in this PR

AI was used to generate the initial version, some manual adjustments to remove unnecessary type casts.

## Proposed Changes

- Delete the Redux slice `provider-constants-slice.ts` which stored static PHP and WordPress version constants as Redux state
- Replace all Redux selectors with direct constant imports from `@studio/common/constants` and `@studio/common/types/php-versions`
- Updated 6 files to use constants directly instead of Redux state
- Removed unnecessary local variable redeclarations and type casts
- Removed test setup code that was dispatching Redux actions for static constants

## Testing Instructions

- TypeScript type checker passes
- All 5 tests in `use-add-site.test.tsx` pass
- Linter passes with no warnings

## Pre-merge Checklist

- [x] TypeScript errors checked


apps/studio/src/hooks/use-add-site.ts
```diff
@@ -331,8 +326,8 @@ export function useAddSite() {
 			handleCreateSite,
 			selectPath,
 			generateProposedPath,
-			defaultPhpVersion,
-			defaultWpVersion: defaultWordPressVersion,
+			defaultPhpVersion: DEFAULT_PHP_VERSION,
+			defaultWpVersion: DEFAULT_WORDPRESS_VERSION,
```
fredrikekelund: AFAICT, these two props are only consumed in `apps/studio/src/modules/add-site/index.tsx`. Could we remove them from this file and just import `DEFAULT_PHP_VERSION` and `DEFAULT_WORDPRESS_VERSION` directly in that file?

fredrikekelund (approved):
Looking good overall, but I left one comment that would be good to address before landing

katinthehatsite (approved):
I think the changes look good. I did not see any regressions and I think that this is a reasonable optimization 👍

apps/studio/src/hooks/use-add-site.ts
bcotrim: Nice catch!

PR #2760, opened 2026-03-11
Fix app menu opening at cursor position instead of button position on touch

wojtekn:
## Related issues

- Fixes https://linear.app/a8c/issue/STU-162

## How AI was used in this PR

This PR was generated with Claude Code. I reviewed the code and the fix is correct.

## Proposed Changes

- Pass the menu button's bounding rect position (`x`, `y`) to `menu.popup()` so the context menu always opens anchored to the bottom-left of the button, regardless of where the touch/cursor lands
- Without coordinates, Electron positions the menu at the current cursor location, which on touch differs from the button's visual position

## Testing Instructions

- On a Windows touch device, tap the Studio menu icon (hamburger) in the title bar
- The context menu should open at the menu button, not at the touch point

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/components/windows-titlebar.tsx
+++ b/apps/studio/src/components/windows-titlebar.tsx
@@ -31,8 +31,12 @@ export default function WindowsTitlebar( {
 		>
 			<Button
 				variant="icon"
-				onClick={ () => {
-					getIpcApi().popupAppMenu();
+				onClick={ ( e: React.MouseEvent< HTMLButtonElement > ) => {
+					const rect = e.currentTarget.getBoundingClientRect();
+					getIpcApi().popupAppMenu( {
+						x: Math.round( rect.left ),
+						y: Math.round( rect.bottom ),
+					} );
 				} }
 				className="!px-3 !py-2 app-no-drag-region"
 			>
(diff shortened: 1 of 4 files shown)


wojtekn:
I tested this PR on a Windows machine with touch (Microsoft Surface), and menu is displayed in the correct position in all cases:

- When I use the mouse and click the menu icon (most common case)
- When I touch the menu icon while having the cursor outside Studio windows (this worked correctly before)
- When I touch the menu icon while having the cursor somewhere over Studio (the case I'm fixing)

katinthehatsite (commented):
I don't have the right Windows device to test this but I wanted to mention that I tested on macOS and did not spot any regressions so far. 

The fix makes sense but I will wait for maybe someone with the target device to give it another test.

wojtekn:
> The fix makes sense but I will wait for maybe someone with the target device to give it another test.

I think we can just test on any Windows machine to check for regressions. I already tested that on one with touch and confirmed it works fine.

gcsecsey (approved):
I tested it on Windows on Parallels. Even without touch, I think it makes sense to anchor the context menu to the button, not to the cursor position.

| trunk | this branch |
|--------|--------|
| Image | Image |

PR #2764, opened 2026-03-11
Harden Studio AI permissions

jverneaut:
## Related issues

- Permissions on upcoming AI assistant for Studio CLI are loose #2632

## Proposed Changes

This PR hardens Studio AI filesystem permissions to keep tool execution scoped to `~/Studio` (the Studio-managed sites root).

### What changed

- Use Studio sites root as the AI execution root:
  - Set AI `cwd` to `~/Studio` (via the shared `STUDIO_SITES_ROOT` path constant).
- Remove unsafe permission bypassing:
  - `permissionMode: 'bypassPermissions'` -> `permissionMode: 'default'`
  - removed `allowDangerouslySkipPermissions: true`
- Add sandbox constraints:
  - `sandbox.enabled: true`
  - `allowUnsandboxedCommands: false`
  - `filesystem.allowWrite: [ STUDIO_ROOT ]`
- Tighten path checks in `canUseTool` for path-sensitive tools:
  - Validate requested paths against `~/Studio`
  - Deny operations resolving outside scope

### Scope and tradeoffs

- This is **not a full security sandbox** by itself and still depends on Claude tool-calling behavior and interpretation.
- The goal here is pragmatic hardening: catch common user mistakes and prevent accidental edits/writes outside Studio-managed paths.
- I intentionally kept **read access broader** because it can be useful for workflows like loading external content into a site.
- **Writing/editing outside the Studio scope is prohibited** by the current permission checks/sandbox configuration.

## Testing Instructions

1. `npm install && npm run cli:build`
2. `ENABLE_STUDIO_AI=true node apps/cli/dist/cli/main.js ai`
3. Validate both:
   - allowed operations under `~/Studio`
   - denied operations targeting paths outside `~/Studio`

### Example allowed prompts

<details>
<summary>Creating a file inside `~/Studio`</summary>

```sh
 ❯ Please create a hello.txt file with the content "Hello, Automattic!" at the root of the current active site                                             
                                                                                                                                                           
 ⏺ Write (my-wordpress-website/hello.txt) (1.0s)                                                                                                           
                                                                                                                                                           
 ⏺ Done. Created hello.txt at /Users/julienverneaut/Studio/my-wordpress-website/hello.txt.                                                                 
```
</details>

<details>
<summary>Reading files outside `~/Studio`</summary>

```
 ❯ What is the first line of the README of ~/Repositories/automattic/studio?                                                                               
                                                                                                                                                           
 ⏺ Read (studio/README.md) (0.0s)                                                                                                                          
                                                                                                                                                           
 ⏺ The first line is:                                                                                                                                      
                                                                                                                                                           
![Crash Free Sessions Rate](https://img.shields.io/badge/Crash_Free_Session_Rate-98.31%25-blue)                                                         
                                                                                                                                                           
It's a badge image for the Crash Free Sessions Rate (98.31%).                                                                                             
```
</details>

### Example blocked prompts

<details>
<summary>Copying files outside `~/Studio`</summary>

```sh
 ❯ Please copy this file over to ~/Repositories/automattic                                                                                                 
                                                                                                                                                           
 ⏺ I can't copy the file to ~/Repositories/automattic since I don't have write access to that directory outside the sandbox. You can do it yourself by     
 running:                                                                                                                                                  
                                                                                                                                                           
bash                                                                                                                                                   
cp /Users/julienverneaut/Studio/my-wordpress-website/hello.txt ~/Repositories/automattic/                                                               
```
</details>


<details>
<summary>Creating files outside `~/Studio`</summary>

```sh
 ❯ Output "Hello, Automattic!" to a hello.txt file on my Desktop                                                                                           
                                                                                                                                                           
 ⏺ Write (Desktop/hello.txt)                                                                                                                               
                                                                                                                                                           
 ⏺ I don't have write access outside of /Users/julienverneaut/Studio. I can't write to your Desktop directly.                                              
                                                                                                                                                           
 You can create the file yourself by running this in your terminal:                                                                                        
                                                                                                                                                                                                                                                                                                     
echo "Hello, Automattic!" > ~/Desktop/hello.txt                                                                                                         
```
</details>

## Pre-merge Checklist


apps/cli/ai/agent.ts
```diff
@@ -52,10 +89,18 @@ export function startAiAgent( config: AiAgentConfig ): Query {
...
-			canUseTool: async ( toolName, input ) => {
+			cwd: STUDIO_ROOT,
+			sandbox: {
+				enabled: true,
+				allowUnsandboxedCommands: false,
+				filesystem: {
+					allowWrite: [ STUDIO_ROOT ],
+				},
+			},
+			tools: [ ...BUILTIN_TOOLS ],
```
youknowriad: I'm wondering if we need this configuration. I think this prevents the agent from using some of the built-in tools like "AskUserQuestion". Is this config necessary for the permissions handling?

youknowriad (commented):
My feeling right now is that this feels too restrictive. Instead of saying "I can't write to this location", would it be better to ask the user for permissions instead?

apps/cli/lib/site-name.ts
```diff
@@ -0,0 +1,9 @@
+import { generateSiteName as generateSiteNameShared } from '@studio/common/lib/generate-site-name';
```
youknowriad, line 1: Curious why you split the previous file into two.

jverneaut: I did it because the `generate-site-name.ts` was doing more than just export the `generateSiteName` method. It also exports `getDefaultSitePath` and `STUDIO_SITES_ROOT`, so the filename felt misleading and I chose to split it to make this clearer. We can alternatively rename this file and keep it as a single one.

youknowriad: No, that's fine thanks for explaining :)

jverneaut:
> My feeling right now is that this feels too restrictive. Instead of saying "I can't write to this location", would it be better to ask the user for permissions instead?

Good point. I don’t think the SDK gives us a built-in prompt for this exact out-of-bounds rule with our current setup, but we can implement it explicitely in `canUseTool`: detect path outside `~/Studio`, ask via `AskUserQuestion`, then return behavior: 'allow' or behavior: 'deny' (with optional “allow for this session”).

Happy to add that in this PR if you want.

youknowriad:
Let's try that 👍

nb:
How have we tested for regressions? I can't get it to use the ask tool, I got a plain text question:

image

youknowriad:
@nb yes, I noticed it, it's related to [this](https://github.com/Automattic/studio/pull/2764#discussion_r2923524149)

jverneaut:
AskUserQuestion wasn’t present in the original `allowedTools` array, but it still worked because we were running with `permissionMode: 'bypassPermissions'` before. In that mode, tools not listed in `allowedTools` can still run. I added `AskUserQuestion` to both `allowedTools` and `tools`.

Following @youknowriad’s comment, here is how it works:
- `tools`: tool availability (which built-in tools Claude can use at all). We need this to explicitly whitelist tools; adding a tool only to `allowedTools` is not sufficient.
- `allowedTools`: permission pre-approval only.
- `canUseTool`: runtime permission hook. We use it to handle cases that need extra logic (for example `AskUserQuestion`) as well as the path based permissions logic for tools that aren't pre-approved.

Sources:
- Permissions flow and `allowed_tools` behavior: https://platform.claude.com/docs/en/agent-sdk/permissions

Side question: do we run E2E agent/tool-invocation regression tests in CI at Automattic (for example, prompt suites with expected tool/permission behavior)?

youknowriad:
I'm actually wondering whether we should allow all Claude code built-in tools (dropping the "tools" property). I mean would that be a bad thing to do when it comes to security. 

Or I guess the question could be, what tools are we losing by explicitly specifying the "tools" property.

jverneaut:
@youknowriad This is possible by using `{ type: 'preset'; preset: 'claude_code' }` in place of the explicit array.

Here is the full list of tools available for reference: https://code.claude.com/docs/en/settings#tools-available-to-claude

Whether we should do it or not depends on how we define “security” in this context and on the capabilities we want this AI CLI to have. My goal with these changes was mostly to reduce accidental user errors and keep actions within the intended scope of the AI CLI, rather than to address a specific threat model.

This is all open to discussion. For example, I chose to keep Read unrestricted for usability, but we can revisit that if we want stricter or loser boundaries.

jverneaut:
Added [2f4cfdc](https://github.com/Automattic/studio/pull/2764/commits/2f4cfdc2ec5f39c0dd81a2fcaeb267a909b6f984) that asks the user for approval on out of bounds operations with the option to accept, accept for this session or deny the operation.

youknowriad:
Screenshot 2026-03-12 at 2 18 11 PM

I wonder if I'm doing something wrong but it's not asking me for permissions.

jverneaut:
@youknowriad I tried the exact same prompt without issues at first, then tried killing the session and redoing it multiple times until I finally got a similar result to yours.

It seems this behavior is subject to interpretation by Claude. I tried adding this line to the system prompt, retried asking it to create an out of bounds file on multiple sessions and it was always able to show the permission flow:

```txt
IMPORTANT: If a Write, Edit, or Bash request targets a path outside the Studio directory, attempt the tool call anyway so the user can validate or deny the operation through the permission flow. Do not self-refuse before requesting permission.
```

I realize this still isn’t fully deterministic, but in my testing it made the outcome much more consistent.

youknowriad:
It works better now but I had this

Screenshot 2026-03-12 at 3 02 52 PM

My experience with claude code has always been that it always asks questions for these things, I wonder if we can get there. What do you think? What can we do here?

jverneaut:
@youknowriad I might be misunderstanding your screenshot: did the delete command actually run, even though the agent said the sandbox blocked it? If not, what did you expect it to do in this case?

I’m currently experimenting with removing strict sandbox `allowWrite` constraints and relying on `canUseTool` for most permission checks. In my tests so far, the agent behavior matches expectations across different file operations, but I’d like to fully understand the implications of this change before moving forward.

youknowriad:
No the delete command didn't run, I expected it to either run (because I allowed the write before) or to ask me for permissions again to run it.

jverneaut:
After additional testing, I removed SDK sandboxing because it proved too restrictive.

We now rely exclusively on the runtime `canUseTool` check. When Claude decides an operation should require approval, it populates `metadata.blockedPath`.

When that happens, it also provides suggestions like:

```json
{
  "metadata": {
    "blockedPath": "/Users/julienverneaut/Desktop/test.txt",
    "suggestions": [
      {
        "type": "addDirectories",
        "directories": [
          "/Users/julienverneaut/Desktop"
        ],
        "destination": "session"
      },
      {
        "type": "setMode",
        "mode": "acceptEdits",
        "destination": "session"
      }
    ]
  }
}
```

I updated permission handling to use the `addDirectories` suggestion instead of the direct file path. Approvals are now folder + tool based, which is closer to expected Claude Code behavior. This is all subject to discussion and can easily be changed if you think another approach would work best.

I also switched the `tools` parameter to the Claude Code preset instead of a narrow explicit built-in tools list.

Following this change:

- Path-gated tools now cover the native write/edit surface:

```ts
const PATH_GATED_TOOLS = [ 'Write', 'Edit', 'Bash', 'NotebookEdit' ] as const;
```

- Allowed tools now cover the read surface:

```ts
const ALLOWED_TOOLS = [
  'mcp__studio__*',
  'Read',
  'Glob',
  'Grep',
  'WebFetch',
  'WebSearch',
  'TodoRead',
  'NotebookRead',
  'AskUserQuestion',
];
```

This does loosen protection compared with hard sandbox constraints since we now rely on runtime checks only, but I think it is a good tradeoff between having no permission handling at all and constraining the agent too much with strict sandboxing.

Here are some prompts I tested this version on:

<details>
<summary>Create files on Desktop (permission prompt appears)</summary>

```text
 ❯ Please create toto-1.txt, toto-2.txt and toto-3.txt on my desktop                                                                                                                                                    
                                                                                                                                                                                                                        
 ⏺ I'll create those three files on your desktop right away.                                                                                                                                                            
                                                                                                                                                                                                                        
   Run (touch /Users/julienverneaut/Desktop/toto-1.txt /Users/jul…)                                                                                                                                                     
                                                                                                                                                                                                                        
Allow Bash to access /Users/julienverneaut/Desktop?                                                                                                                                                                     
  ❯ 1. Allow once                                                                                                                                                                                                       
    2. Allow for this session                                                                                                                                                                                           
    3. Deny                                                                                                                                                                                                             
```

</details>

<details>
<summary>Delete Desktop files from today (permission + confirmation flow)</summary>

```text
 ❯ Please remove all the text files on my desktop that were created today                                                                                                                                               
                                                                                                                                                                                                                        
 ⏺ I can see you want to delete files from your Desktop. Let me first check what's there before taking any action.                                                                                                      
                                                                                                                                                                                                                        
 ⏺ Run (ls -la ~/Desktop/*.txt 2>/dev/null && stat -f "%SB %N" ~/…) (12.9s)                                                                                                                                             
                                                                                                                                                                                                                        
Allow Bash to access /Users/julienverneaut/Desktop?                                                                                                                                                                     
                                                                                                                                                                                                                        
 ⏺ I found 3 text files on your Desktop, all created today (March 13, 2026):                                                                                                                                            
                                                                                                                                                                                                                        
 - toto-1.txt                                                                                                                                                                                                           
 - toto-2.txt                                                                                                                                                                                                           
 - toto-3.txt                                                                                                                                                                                                           
                                                                                                                                                                                                                        
 All three were created today. Should I go ahead and delete all of them?                                                                                                                                                
                                                                                                                                                                                                                        
 Thought for 20s · 2 turns · $0.0335                                                                                                                                                                                    
                                                                                                                                                                                                                        
                                                                                                                                                                                                                        
 ❯ Yes please.                                                                                                                                                                                                          
                                                                                                                                                                                                                        
 ⏺ Run (rm ~/Desktop/toto-1.txt ~/Desktop/toto-2.txt ~/Desktop/to…) (0.7s)                                                                                                                                              
                                                                                                                                                                                                                        
 ⏺ Done. All 3 files (toto-1.txt, toto-2.txt, toto-3.txt) have been deleted from your Desktop.                                                                                                                          
                                                                                                                                                                                                                        
 Thought for 6s · 2 turns · $0.0333                                                                                                                                                                                     
```

</details>

<details>
<summary>Create <code>test.txt</code> across Studio sites (in-scope write)</summary>

```text
 ❯ Please create an empty test.txt file on all the sites that Studio manages                                                                                                                                            
                                                                                                                                                                                                                        
 ⏺ Let me first list all the Studio sites to find out what we're working with.                                                                                                                                          
 I found 8 sites. Now I'll create an empty test.txt file in each site's root directory simultaneously.                                                                                                                  
 The Write tool requires reading a file before overwriting it, but since these are new files, I'll use Bash to create the empty files instead.                                                                          
 Done. An empty test.txt file has been created in the root directory of all 8 sites:                                                                                                                                    
                                                                                                                                                                                                                        
 - business-card                                                                                                                                                                                                        
 - My WordPress Website                                                                                                                                                                                                 
 - session-fidelity-alpha-m3p9                                                                                                                                                                                          
 - session-fidelity-alpha-m9p4                                                                                                                                                                                          
 - session-fidelity-alpha-x7k2                                                                                                                                                                                          
 - session-fidelity-beta-m9p4                                                                                                                                                                                           
 - session-fidelity-beta-q3r8                                                                                                                                                                                           
 - session-fidelity-beta-q5w8                                                                                                                                                                                           
                                                                                                                                                                                                                        
 ⏺ List sites (0.0s)                                                                                                                                                                                                    
   ⎿ Chargement des sites…                                                                                                                                                                                              
   ⎿ 8 sites trouvés                                                                                                                                                                                                    
   ⎿ Connexion au processus démon...                                                                                                                                                                                    
   ⎿ Connecté au processus démon                                                                                                                                                                                        
                                                                                                                                                                                                                        
 ⏺ Write (business-card/test.txt)                                                                                                                                                                                       
                                                                                                                                                                                                                        
 ⏺ Write (my-wordpress-website/test.txt)                                                                                                                                                                                
                                                                                                                                                                                                                        
 ⏺ Write (session-fidelity-alpha-m3p9/test.txt) (0.3s)                                                                                                                                                                  
                                                                                                                                                                                                                        
 ⏺ Write (session-fidelity-alpha-m9p4/test.txt)                                                                                                                                                                         
                                                                                                                                                                                                                        
 ⏺ Write (session-fidelity-alpha-x7k2/test.txt)                                                                                                                                                                         
                                                                                                                                                                                                                        
 ⏺ Write (session-fidelity-beta-m9p4/test.txt)                                                                                                                                                                          
                                                                                                                                                                                                                        
 ⏺ Write (session-fidelity-beta-q3r8/test.txt)                                                                                                                                                                          
                                                                                                                                                                                                                        
 ⏺ Write (session-fidelity-beta-q5w8/test.txt)                                                                                                                                                                          
                                                                                                                                                                                                                        
 ⏺ Run (touch ~/Studio/my-wordpress-website/test.txt ~/Studio/ses…) (0.7s)                                                                                                                                              
                                                                                                                                                                                                                        
 Thought for 20s · 11 turns · $0.0907                                                                                                                                                                                   
```

</details>

<details>
<summary>Run a simple read command (<code>whoami</code>)</summary>

```text
 ❯ Please run whoami                                                                                                                                                                                                    
                                                                                                                                                                                                                        
 ⏺ Run (whoami) (0.4s)                                                                                                                                                                                                  
                                                                                                                                                                                                                        
 ⏺ You are julienverneaut.                                                                                                                                                                                              
```

</details>

<details>
<summary>Use <code>AskUserQuestion</code> in a full interaction flow</summary>

```text
 ❯ I want to demonstrate that you can run tools without any issues. I want you to create a todo list of things you want to ask me, then ask each question using AskUserQuestion and tick the todolist elements as you   
do.                                                                                                                                                                                                                     
                                                                                                                                                                                                                        
 ⏺ Update todo list (0.0s)                                                                                                                                                                                              
                                                                                                                                                                                                                        
 ⏺ Update todo list (0.0s)                                                                                                                                                                                              
                                                                                                                                                                                                                        
 ⏺ AskUserQuestion (5.2s)                                                                                                                                                                                               
                                                                                                                                                                                                                        
What is your favorite programming language?                                                                                                                                                                             
                                                                                                                                                                                                                        
 ⏺ Update todo list (0.0s)                                                                                                                                                                                              
                                                                                                                                                                                                                        
 ⏺ AskUserQuestion (1.8s)                                                                                                                                                                                               
                                                                                                                                                                                                                        
What is your preferred code editor?                                                                                                                                                                                     
                                                                                                                                                                                                                        
 ⏺ Update todo list (0.0s)                                                                                                                                                                                              
                                                                                                                                                                                                                        
 ⏺ AskUserQuestion (3.5s)                                                                                                                                                                                               
                                                                                                                                                                                                                        
What is your current main project focus?                                                                                                                                                                                
                                                                                                                                                                                                                        
 ⏺ Update todo list (0.0s)                                                                                                                                                                                              
                                                                                                                                                                                                                        
 ⏺ All three questions answered and the todo list is complete. Here's a summary of what I learned:                                                                                                                      
                                                                                                                                                                                                                        
 - Favorite language: JavaScript/TypeScript                                                                                                                                                                             
 - Preferred editor: VS Code                                                                                                                                                                                            
 - Current focus: WordPress development                                                                                                                                                                                 
                                                                                                                                                                                                                        
 Good to know — that context will be helpful when working on your Studio sites!                                                                                                                                         
                                                                                                                                                                                                                        
 Thought for 39s · 9 turns · $0.1568                                                                                                                                                                                    
```

</details>

youknowriad:
Screenshot 2026-03-13 at 11 16 07 AM

Hey @jverneaut maybe I'm not testing properly, but for some reason, the AI gets stuck for me when I prompt almost anything.

jverneaut:
@youknowriad I rebased the PR, you should be able to test it.

youknowriad (commented):
This is working great in my tests. I think it's a good first step in terms of security.

apps/cli/ai/agent.ts
```diff
@@ -24,6 +27,150 @@ export const AI_MODELS = {
...
+	if ( answers[ question ] === APPROVE_ONCE_LABEL ) {
+		return 'allow_once';
+	}
+
+	if ( answers[ question ] === APPROVE_SESSION_LABEL ) {
+		return 'allow_session';
+	}
+
+	return 'deny';
+}
```
youknowriad: Would it be possible to move some of this code to a decoupled security.ts file that we could potentially test...

I'd love to keep agent.ts just about the agent loop definition.

jverneaut: Done in 796dc62.

apps/cli/ai/agent.ts
```diff
@@ -46,10 +53,11 @@ export function startAiAgent( config: AiAgentConfig ): Query {
 				studio: createStudioTools(),
 			},
 			maxTurns,
-			cwd: process.cwd(),
-			permissionMode: 'bypassPermissions',
-			allowDangerouslySkipPermissions: true,
-			canUseTool: async ( toolName, input ) => {
+			cwd: STUDIO_ROOT,
+			tools: { type: 'preset', preset: 'claude_code' },
+			allowedTools: [ ...ALLOWED_TOOLS ],
```
youknowriad, line 56: Is the cloning of the object necessary?

apps/cli/ai/agent.ts
```diff
@@ -46,10 +53,11 @@ export function startAiAgent( config: AiAgentConfig ): Query {
...
+			permissionMode: 'default',
+			canUseTool: async ( toolName, input, metadata ) => {
```
youknowriad, line 58: It feels like all of these function should be moved to "security.ts" and unit test it no? What are the dependencies of this function, it seems to be fairly isolated no?

youknowriad (approved):
Left some last comments, this PR would need rebasing to land it.

apps/cli/ai/agent.ts
jverneaut: `ALLOWED_TOOLS` is typed as a const as an immutable list of tools while allowedTools expects `string[]`, this does the conversion.

youknowriad: But this is a runtime conversion (cloning the object), why can't `ALLOWED_TOOLS`  be defined as `string[]`

jverneaut: It is and is not strictly necessary, but I like the `as const` on `ALLOWED_TOOLS` as this list serves as an immutable policy list that shouldn't change at runtime even if the underlaying construct can. It's more of a stylistic choice and I have no issue changing it either if you want me to do it.

apps/cli/ai/agent.ts
jverneaut: Does this comment precedes 796dc62 or are you talking about `canUseTool` as a whole?

apps/cli/ai/agent.ts
youknowriad: What bothers me personally is the runtime penalty of cloning the object, but it's minor.

apps/cli/ai/agent.ts
youknowriad: I'm talking about canUseTool as a whole.

jverneaut: I kept `canUseTool` in agent.ts because it’s part of the agent control flow, not only security. It also handles `AskUserQuestion` input typing for `ui.askUser`, and this is where we may add other non-security user input handlings later. I’m happy to move the relevant parts of it to `security.ts` if you prefer.

youknowriad: Isn't the canUseTool entirely about "security"? I'm hesitant personally but It's ok to defer for later. What I felt is that https://github.com/Automattic/studio/pull/2764/changes#diff-149825221e3d291c1aa59e93f8add2ceca4e24478055e6932e59805dc464ee9eR74-R108 basically could be just a single function call, it's the whole security logic.

jverneaut: It's about handling user input which pauses execution, like permissions but also clarifying questions via `AskUserQuestion` (per https://platform.claude.com/docs/en/agent-sdk/user-input).

In the docs at the same url, they suggest this pattern:

```js
canUseTool: async (toolName, input) => {
  if (toolName === "AskUserQuestion") {
    // Your implementation to collect answers from the user
    return handleClarifyingQuestions(input);
  }
  // Handle other tools normally
  return promptForApproval(toolName, input);
};
```

I like the `promptForApproval` method, I think it's a nice middle ground between abstracting the entire `canUseTool` method away and having the entire flow in the method.

youknowriad: Ok, works for me.

jverneaut: Done in 2bafadb.

youknowriad: looks like you didn't push this commit to this branch :)

youknowriad: Seems like a bug in GitHub, I can see the commit in the branch, let me rebase.

PR #2766, opened 2026-03-11
Attempt to remove incompatible binaries when signing

twstokes:
Experimental as an attempt to see if we can get a successful Windows build since https://github.com/Automattic/studio/pull/2632.

## Summary

Fixed Windows code-signing failures in CI by removing native binaries for non-target platforms before packaging.

## Problem

Commit 417375c introduced the `studio ai` CLI command which added two new dependencies that ship native binaries for all platforms:
- `@anthropic-ai/claude-agent-sdk` - contains ripgrep and tree-sitter binaries in `vendor/{tool}/{arch}-{platform}/`
- `koffi` (transitive dep via `@mariozechner/pi-tui`) - contains FFI binaries in `build/koffi/{platform}_{arch}/`

Windows code-signing (signtool) fails when it encounters non-Windows binaries (e.g., darwin `.node` files) because they're not valid PE files.

## Solution

Added cleanup logic in the `prePackage` hook in `apps/studio/forge.config.ts` that runs after `cli:package` builds the CLI bundle. It removes platform-specific binary directories that don't match the target platform:

- For `claude-agent-sdk`: removes directories not ending with `-{platform}` (e.g., keeps only `*-win32` on Windows)
- For `koffi`: removes directories not starting with `{platform}_` (e.g., keeps only `win32_*` on Windows)

The initial approach tried to add `ignore` patterns in `packagerConfig` to exclude binaries for other platforms. This didn't work because of how Electron Forge handles `extraResource` vs the main app bundle:

**How `ignore` works:**
- The `ignore` patterns in `packagerConfig` filter what gets copied into the app's main directory (and ASAR archive)
- They apply to the source files being packaged from the project root

**How `extraResource` works:**
- Items listed in `extraResource` are copied directly to the `resources/` folder of the packaged app
- They bypass the `ignore` filter entirely - they're a separate copy operation

The CLI is added as an extraResource:
```javascript
extraResource: [
    // ...
    path.join( repoRoot, 'apps', 'cli', 'dist', 'cli' ),
],
```

So the CLI bundle (including its `node_modules` with the native binaries) gets copied wholesale to `resources/cli/` without any filtering. That's why we needed to clean up the binaries in the `prePackage` hook *before* the extraResource copy happens, rather than relying on ignore patterns.

## How AI was used in this PR

Claude completely drove these code changes by supplying it the Buildkite logs with the error.

## Testing Instructions

- Ensure that the optional Windows and Mac builds succeed

## Pre-merge Checklist

### DIFF
--- a/apps/studio/forge.config.ts
+++ b/apps/studio/forge.config.ts
@@ -176,6 +176,45 @@ const config: ForgeConfig = {
 			// rerun `npm ci` from the repo root to reset the dependency tree after packaging.
 			await execAsync( 'npm run cli:package' );
 
+			// Remove native binaries for other platforms from CLI's node_modules.
+			// Some packages ship binaries for all platforms which causes code-signing failures
+			// on Windows when signtool encounters non-PE binaries (e.g., darwin .node files).
+			console.log( `Removing native binaries for other platforms from CLI bundle...` );
+			const cliNodeModules = path.join( repoRoot, 'apps', 'cli', 'dist', 'cli', 'node_modules' );
+
+			// Clean up @anthropic-ai/claude-agent-sdk vendor binaries (uses {arch}-{platform} format)
+			const claudeVendorDir = path.join( cliNodeModules, '@anthropic-ai', 'claude-agent-sdk', 'vendor' );
+			const platformSuffix = `-${ platform }`;
+			if ( fs.existsSync( claudeVendorDir ) ) {
+				for ( const toolDir of fs.readdirSync( claudeVendorDir ) ) {
+					const toolPath = path.join( claudeVendorDir, toolDir );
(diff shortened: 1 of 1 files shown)


twstokes:
cc @nb and @youknowriad – this was completely driven by Claude with the intent to fix the Windows builds. It seems to have succeeded, but feel free to just use it as proof that we know what the cause is: Incompatible binaries that are part of the packages introduced in https://github.com/Automattic/studio/pull/2632 can't be signed by the Windows SignTool.

youknowriad:
Awesome looks like it worked. Thanks a lot @twstokes

youknowriad (approved):

youknowriad:
Let's merge this to unblock trunk merges.

PR #2768, opened 2026-03-11
Studio: Re-selecting the same file doesn't trigger the import

katinthehatsite:
## Related issues

Fixes STU-1381

## How AI was used in this PR

Not used, it is a simple fix

## Proposed Changes

This PR ensures that when you use an `Import` option in Studio and select the same file after cancelling the import with that file once, the state is reset and the import can be initiated. 

## Testing Instructions

* Navigate to `Import` tab
* Select a file for import
* Cancel import
* Select the same file for import again
* Confirm that the import can be started 

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/components/content-tab-import-export.tsx
+++ b/apps/studio/src/components/content-tab-import-export.tsx
@@ -229,8 +229,8 @@ const ImportSite = ( {
 		if ( ! file ) {
 			return;
 		}
+		clearImportFileInput();
 		void importConfirmation( async () => {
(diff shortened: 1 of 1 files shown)


epeicher (approved):
Thanks for the fix Kat! I have tested it, and I can confirm that it fails in `trunk` and it works as expected on this branch. LGTM! :shipit:

katinthehatsite:
Thanks for the review and merging the PR 👍

PR #2770, opened 2026-03-11
Add site import command with streaming migration and SQLite db-apply

adamziel:
🚧  Work in progress  🚧 

## Proposed Changes

Adds a `studio site import` command that synchronizes a remote WordPress site into the local Studio instance.

### Usage

```
# Build the CLI first
npm run cli:build

# Import a site (requires streaming-site-migration plugin on the remote)
node apps/cli/dist/cli/main.js site import \
  --url https://example.com \
  --secret the-hmac-secret-from-the-plugin \
  --name "My imported site"

# If interrupted, re-run the exact same command to resume
node apps/cli/dist/cli/main.js site import \
  --url https://example.com \
  --secret the-hmac-secret-from-the-plugin \
  --name "My imported site"

# After import, start the site
node apps/cli/dist/cli/main.js site start --path ~/Studio/my-imported-site
```

## Implementation

This command uses the [streaming-site-migration](https://github.com/adamziel/streaming-site-migration) tool. It requires the remote site to install an exporter plugin and then uses the `importer.phar` tool to download the files, the database, and rewrite the URLs in the content to match the Studio site's URL.

### Custom site structure

Regular Studio sites are stored in`~/Studio/site-name` and then you get `wp-config.php`, `wp-content` etc. Imported Studio sites reproduce the hosting account directory structure from the remote site. For example, my blog that I've imported from WordPress.com has the following directories in it:

```
./wordpress
./wordpress/core
./wordpress/plugins
./wordpress/themes
./srv
./srv/wordpress
./srv/htdocs
./srv/htdocs/wp-config.php
...
```

That's the same as what the live site uses. Is it unexpected and inconvenient? Yes. We preserve it, because there are symlinks and `require` statements referring to paths outside of the document root, such as `../wordpress/core/latest`. Some symlinks form recursive loops. That directory structure may also be reflected in `wp_options` and other database tables.

**Is there a way to normalize it into your regular WordPress directory structure in the general case?** I don't see any. Here's the list of problems:

1. `is_symlink()` checks in the codebase would be affected.
2. We'd need to update all the affected paths in all the PHP files and database entries. I don't know how to do that reliably, e.g. PHP scripts often constructs paths dynamically (`require "../themes" . $theme_dir . "/functions.php"`).  Updating URLs is much easier – `https://` is a very specific hint, most URLs are in the block markup, and we can miss some of them without generating a fatal error on every page.
3. Flattening a directory tree is a lossy operation. We couldn't reliably do repeated syncs or push the updated site back to the remote host. When a symlinked file changes, there is no way to decide whether, on the remote site, we should update the symlink target or replace the symlink with a copy of the updated file.

**But is there a way to normalize that directory structure on a host-by-host basis?** Probably, at least for some hosts. We could probably do it for WordPress.com sites, but more research is needed to confirm that.

### Special handling for managed hosting environments

Each webhosts is configured differently. Some of those configurations are not easily compatible with a local environment. The import command applies a few host-specific corrections to account for that:

- **WordPress.com Atomic**: Creates wp-config.php symlinks in the shared WP core directory so WordPress can find its config when `__DIR__` resolves to the core dir instead of the document root
- **SiteGround**: Deactivates the `sg-security` and `sg-cachepress` plugins. They could work with more effort, but for now they're deactivated.
- **Managed hosts in general**: Injects placeholder DB_* constants when wp-config.php doesn't define them (managed hosts inject credentials at the server level)

Those corrections are applied by the code shipped in this PR. This isn't perfect as they are relevant to every user of the importer tool. Therefore, they will be upstreamed to https://github.com/adamziel/streaming-site-migration, but that's out of scope of this PR.

## Blockers for merging

This PR is functional but has several rough edges that should be addressed before merging:

- [ ] **Assess how the custom directory structure would affect the rest of Studio.** I assume it would affect sync. Would it break any existing features?
- [x] Check why do the imported wp.com sites have a `wp-admin` directory in `htdocs`. This is unexpected.
- [ ] Replace ad-hoc PHP code manipulation with parsing. `wp-config.php` manipulation is done with regex in TypeScript. `serializePhpStringArray` and `deactivatePluginInSqlite` reimplement PHP serialization in TypeScript`. This should use proper PHP parsing. Playground already ships tools for that, we just need to use them.
- [ ] Test coverage for the import command.
- [ ] Improve progress messages. The number of total files and MBs downloaded tends to go up and then down. It shouldn't. Let's make it reliable and stable.

### Follow-up work

These items can be handled independently of merging this PR:

- [ ] Upstream the host-specific site corrections to https://github.com/adamziel/streaming-site-migration/.
- [x] Explore flattening the imported site structure for WP.com sites.
- [ ] **importer.phar is re-downloaded on every import** — The cached phar is deleted at the start of each import to ensure the latest version. The importer script could be either bundled with Studio or cached based on the remote ETag.
- [ ] Provide more progress updates during the database import as it can take a while.
- [x] Profile and speed-up the database import. Why is it not faster? Is it the URL rewriting?
- [ ] `disableErrorDisplayInWpConfig` is a workaround — Injecting `@ini_set('display_errors', '0')` into wp-config.php works around the fact that Studio's Blueprint constants are applied after early PHP bootstrap errors can leak through. This should be handled at the platform level (e.g., via php.ini entries in the WASM runtime) rather than modifying the imported site's files.

## Testing Instructions

1. Install the [streaming-site-migration plugin](https://github.com/adamziel/streaming-site-migration/releases/tag/v0.1.9) on a remote WordPress site
7. Run: `studio site import --url <site-url> --secret <hmac-secret> --name "My imported site"`
8. Verify the import completes and the site appears in `studio site list`
9. Start the site with `studio site start` and verify it loads correctly
10. Test resumption: interrupt an import mid-download and re-run the same command

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/site/import.ts
+++ b/apps/cli/commands/site/import.ts
@@ -105,6 +105,142 @@ function disableErrorDisplayInWpConfig( documentRoot: string ): void {
 }
 
 /**
+ * Ensures wp-config.php has all required constants for running
+ * under Studio with SQLite.
+ *
+ * Managed WordPress hosts like SiteGround and WordPress.com Atomic
+ * inject DB credentials at the server level, so wp-config.php
+ * may only contain comments about DB_HOST/DB_USER/DB_PASSWORD
+ * without actually defining them. WordPress requires these
+ * constants to exist (even with SQLite, where db.php intercepts
+ * before MySQL connects) or it redirects to setup-config.php.
+ *
+ * WP_CONTENT_DIR is set explicitly because Atomic sites use
+ * symlinks for wp-load.php that make __DIR__ resolve to the
+ * shared WP core directory. Without this, WordPress would look
+ * for wp-content (and db.php) in the core dir instead of the
+ * site's document root — exactly what the Atomic platform does
+ * with `define( 'WP_CONTENT_DIR', realpath('/srv/htdocs/wp-content') )`.
+ */
+function ensureRequiredConstantsInWpConfig( documentRoot: string ): void {
+	const wpConfigPath = path.join( documentRoot, 'wp-config.php' );
+	if ( ! fs.existsSync( wpConfigPath ) ) {
+		return;
+	}
+
+	let content = fs.readFileSync( wpConfigPath, 'utf-8' );
+
+	// DB placeholder constants — values don't matter for SQLite,
+	// but WordPress requires them to be defined.
+	const requiredConstants: Record< string, string > = {
+		DB_NAME: 'wordpress',
+		DB_USER: 'root',
+		DB_PASSWORD: 'root',
+		DB_HOST: 'localhost',
+	};
+
+	for ( const [ name, value ] of Object.entries( requiredConstants ) ) {
+		const definePattern = new RegExp( `define\\s*\\(\\s*['"]${ name }['"]` );
+		if ( ! definePattern.test( content ) ) {
+			// Insert before "That's all, stop editing" or before ABSPATH
+			const insertBefore = content.match(
+				/\/\*.*(?:stop editing|ABSPATH).*\*\/|if\s*\(\s*!\s*defined\s*\(\s*'ABSPATH'\s*\)/
+			);
+			const insertLine = `define( '${ name }', '${ value }' );\n`;
+			if ( insertBefore ) {
+				content = content.replace( insertBefore[ 0 ], insertLine + insertBefore[ 0 ] );
+			} else {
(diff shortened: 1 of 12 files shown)


fredrikekelund:
@adamziel, we have a task (STU-1351) to implement the existing import/export logic in the CLI for Studio. The sync functionality, which is also slated for implementation in the CLI (STU-1354), depends on the existing import/export logic.

adamziel:
continued in https://github.com/Automattic/studio/pull/2939

PR #2774, opened 2026-03-12
Track app binary size in performance metrics

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

Claude Code was used to implement this feature. Code was reviewed before committing.

## Proposed Changes

I propose adding tracking of app size to quickly catch any changes that affect it.

I chose one arch for each platform, so we have "App Size (Mac)" and "App Size (Win)" stat names.

I considered tracking the installer first, but I think the unpacked and installed app is more real stat to track. E.g., when as a user I go to `Applications/` directory to check Studio size, I will currently see 1.26 GB, which corresponds to the value tracked under this stat.

## Testing Instructions

1. Run `npm run package` to build the app
2. Run `ARTIFACTS_PATH=$PWD/tools/metrics/artifacts npm run test:metrics -- app-size`
3. Verify `tools/metrics/artifacts/app-size.results.json` contains `{"appSizeMac": <bytes>}` (on macOS)
4. Output table should show size in MB

## Pre-merge Checklist


tools/metrics/metric-config.ts
```diff
@@ -0,0 +1,26 @@
...
+	appSizeWin: { label: 'App Size (Win)', unit: 'MB', threshold: BYTES_PER_MB },
+};
+
+export function getMetricConfig( metric: string ): MetricConfig {
+	return METRIC_CONFIG[ metric ] ?? { label: metric, unit: 'ms', threshold: 50 };
+}
+
+export function formatMetricValue( metric: string, value: number ): string {
+	const { unit } = getMetricConfig( metric );
+	if ( unit === 'MB' ) {
```
youknowriad: Could be a "format" property in the metric interface at some point.

youknowriad (approved):

tools/metrics/metric-config.ts
wojtekn: Makes sense, I went ahead and extracted those.

wojtekn:
@youknowriad I kept the binary size measurement as a Playwright test for now. It may be overkill, but it follows the current metrics-collection approach, in which collecting metrics means running performance tests.

tools/metrics/utils.ts
```diff
@@ -9,3 +12,20 @@ export function median( array ) {
 	}
 	return numbers[ middleIndex ];
 }
+
+export function getDirSize( dirPath: string ): number {
```
epeicher, line 16: Just curious, how is this function used? I don't see any usage of it

wojtekn: I re-added test file that uses that function.

PR #2775, opened 2026-03-12
Add WP.com and Anthropic auth modes to studio ai

youknowriad:
## Related issues

- None provided.

## How AI was used in this PR

- Used AI to iterate on the `studio ai` provider/auth flow, then manually reviewed the diff and tightened the prompt timing and TUI handoff.

## Proposed Changes

- Add a `/provider` flow to `studio ai` with `WordPress.com`, `Anthropic · Claude auth`, and `Anthropic · API key` options.
- Route Anthropic auth through explicit provider env setup, persist the selected provider in appdata, and show the active provider in the TUI header.
- Prompt for Anthropic API keys when switching to API key mode, surface clear errors when that mode has no saved key, and stop the TUI during password entry.
- Add focused tests for provider selection, auth resolution, and default-provider behavior.

## Testing Instructions

- Run `npm test -- apps/cli/ai/tests/auth.test.ts apps/cli/lib/tests/appdata.test.ts`.
- Run `npm run typecheck`.
- Run `npm run cli:build`.
- Start `studio ai`, use `/provider`, and verify Claude auth only appears when available, API key mode prompts immediately, and WP.com mode works without an Anthropic key.

## Pre-merge Checklist


apps/cli/ai/providers.ts
```diff
@@ -0,0 +1,180 @@
...
+				'X-WPCOM-AI-Feature': WPCOM_AI_FEATURE_HEADER,
+			} );
+			env.CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS = '1';
+			return env;
+		},
+	},
+	'anthropic-claude': {
+		id: 'anthropic-claude',
+		autoFallbackWhenUnavailable: true,
+		isVisible: async () => hasClaudeCodeAuth(),
```
adamziel, line 134: is it not a visible option until there's an authorization available? how do you authorize, then? Should it be something like `Claude (login to use)`?

apps/cli/commands/ai.ts
```diff
@@ -1,85 +1,90 @@
...
-		return savedKey;
-	}
+	let sessionId: string | undefined;
+	let currentModel: AiModelId = DEFAULT_MODEL;
 
-	// If Claude Code is authenticated, use its auth
-	if ( isClaudeCodeAuthenticated() ) {
-		return undefined;
+	async function prepareProviderSelection( provider: AiProviderId ): Promise< void > {
+		ui.stop();
```
adamziel, line 53: What does it mean that we are "stopping" the UI? Is a part of all of it no longer visible to the user and we're replacing it with a model selector? What would be a clearer name for this operation? I've checked that method but it's not very clear what's happening:

```
	stop(): void {
		this.loader.stop();
		this.tui.stop();
	}
```

apps/cli/ai/providers.ts
youknowriad: It's not visible I think until you have claude code locally installed. This was kind of already here before and I organized it in providers.

apps/cli/commands/ai.ts
```diff
@@ -162,10 +167,10 @@ export async function runCommand(): Promise< void > {
 				ui.stop();
 				await runLoginCommand();
 				ui.start();
-				try {
+				if ( await isAiProviderReady( 'wpcom' ) ) {
```
adamziel, line 215: This PR removes a bunch of try/catch constructs, but the code that it wrapped can still throw an error. If `getAuthToken()` throws, would it now crash the app instead of displaying the error message?

apps/cli/commands/ai.ts
youknowriad: stop() / start() are indeed a bit too generic here. What’s actually happening is that the chat TUI temporarily releases terminal control so Inquirer can prompt, then resumes afterward. We may want to rename these separately but it's a bit out of scope of the current PR.

apps/cli/commands/ai.ts
youknowriad: I hate the AI's added try/catches so much, It's too defensive

apps/cli/ai/providers.ts
adamziel: Don't we want to make it visible for discoverability?

apps/cli/commands/ai.ts
adamziel: Sure. Thanks!

apps/cli/commands/ai.ts
adamziel: Ah so we didn't need them in the first place is what you're saying? So the wrapped methods don't throw or they're unlikely to throw, in which case crashing the app is fine?

youknowriad: Yes, I think the whole ai.ts and ui.ts files need cleanup is what I'm saying :P

adamziel:
I'd like to be able to escape this prompt with the `escape` key or ctrl+c, right now I'm stuck in there as pressing enter only gives me an error without navigating away:

CleanShot 2026-03-12 at 16 33 18@2x

CleanShot 2026-03-12 at 16 33 56@2x

taipeicoder:
A couple of oddities with the API key. I deliberately entered an invalid API key and:

 - I get `Invalid API key · Fix external API key` message and then the CLI exits. I expected to go back to the prompt.
 - There seems to be no way to update the API key from the CLI.

adamziel:
Can we have tab completion after `/provider`? I've noticed I'm pressing the tab key to see the available options but nothing happens.

youknowriad:
@adamziel @taipeicoder addressed the two comments. The completion task is tracked separately.

mtias (approved):

taipeicoder (approved):

PR #2778, opened 2026-03-12
Studio: Refactor AuthProvider into Redux slice

katinthehatsite:
## Related issues

Fixes https://linear.app/a8c/issue/STU-1379/studio-refactor-authprovider-into-redux-slice

## How AI was used in this PR

I used it in the plan mode to plan the refactor, then gave feedback on the plan and used it for implementation, making further manual corrections.

## Proposed Changes

This PR refactors `AuthProvider` into Redux slice. 

## Testing Instructions

* Log out of Studio and confirm that you are seeing logged out view
* Log in and confirm it works as expected
* Confirm you are correctly seeing the features that should be available with WP.com logged in view
* Confirm tests are passing
* Close the app while logged in
* Open `~/Library/Application Support/Studio/appdata-v1.json`
* Delete the `authToken`
* Save the changes
* Start the app with again
* Confirm that you can see logged out view

## Pre-merge Checklist


apps/studio/src/stores/provider-constants-slice.ts
```diff
@@ -1,4 +1,4 @@
-import { createSlice, PayloadAction } from '@reduxjs/toolkit';
```
katinthehatsite: This is not related directly to changes in this PR. This is something that my linter flagged that needed to be fixed and was present on trunk

apps/studio/src/components/auth-provider.tsx
fredrikekelund, line 1: We should remove this context provider altogether. Components can easily consume Redux state and dispatch actions. There's no point in keeping this provider as a thin wrapper around the Redux slice.

apps/studio/src/stores/auth-slice.ts
```diff
@@ -0,0 +1,192 @@
...
+			} )
+			.addMatcher( isAnyOf( handleInvalidToken.fulfilled, authLogout.fulfilled ), ( state ) => {
+				state.isAuthenticated = false;
+				state.user = null;
+			} );
+	},
+} );
+
+export const selectIsAuthenticated = ( state: RootState ) => state.auth.isAuthenticated;
+export const selectUser = ( state: RootState ) => state.auth.user ?? undefined;
```
fredrikekelund, line 193:
```suggestion
export const authThunks = {
	handleInvalidToken,
	initializeAuth,
	authTokenReceived,
	authLogout,
};

export const authSelectors = {
	selectIsAuthenticated: ( state: RootState ) => state.auth.isAuthenticated,
	selectUser: ( state: RootState ) => state.auth.user ?? undefined,
};
```
We've followed this pattern in some other Redux slices. I think it's helpful to namespace selectors, actions, and thunks this way.

apps/studio/src/stores/auth-slice.ts
```diff
@@ -0,0 +1,192 @@
+import { createAsyncThunk, createSlice, isAnyOf } from '@reduxjs/toolkit';
+import * as Sentry from '@sentry/electron/renderer';
+import wpcomFactory from '@studio/common/lib/wpcom-factory';
+import wpcomXhrRequest from '@studio/common/lib/wpcom-xhr-request-factory';
+import { getIpcApi } from 'src/lib/get-ipc-api';
+import { isInvalidTokenError } from 'src/lib/is-invalid-oauth-token-error';
+import { store, RootState } from 'src/stores';
+import { getWpcomClient, setWpcomClient } from 'src/stores/wpcom-api';
```
fredrikekelund: Now that the auth details live in the Redux state tree, I'd argue that `src/stores/wpcom-api` should either retrieve the `client` state prop from this slice or subscribe to state updates from it. We should remove the `getWpcomClient` and `setWpcomClient` from `src/stores/wpcom-api`

apps/studio/src/stores/auth-slice.ts
```diff
@@ -0,0 +1,192 @@
...
+	let isAuthErrorDialogOpen = false;
+	const handleInvalidTokenError = async ( response: unknown ) => {
+		if ( isInvalidTokenError( response ) && onInvalidToken && ! isAuthErrorDialogOpen ) {
+			isAuthErrorDialogOpen = true;
+			onInvalidToken();
+			await getIpcApi().showMessageBox( {
+				type: 'error',
+				message: 'Session Expired',
+				detail: 'Your session has expired. Please log in again.',
+			} );
```
fredrikekelund: If the idea is to always dispatch `handleInvalidToken` here, then I think it's better to pass the `store` as an argument to this function and call `store.dispatch( handleInvalidToken() )`. 

We could also consider moving the `getIpcApi().showMessageBox()` call to `handleInvalidToken`

apps/studio/src/components/auth-provider.tsx
```diff
@@ -38,35 +36,25 @@ export const AuthContext = createContext< AuthContextType >( {
...
-		} catch ( err ) {
-			console.error( 'Failed to handle invalid token:', err );
-			Sentry.captureException( err );
-		}
-	}, [] );
+	useEffect( () => {
+		void dispatch( initializeAuth( { locale } ) );
+	}, [ dispatch, locale ] );
 
 	useIpcListener( 'auth-updated', ( _event, payload ) => {
```
fredrikekelund: This can be moved to `apps/studio/src/stores/auth-slice.ts` like so:

```ts
window.ipcListener.subscribe( 'auth-updated', ( _event, payload ) => {
	if ( 'error' in payload ) {
		let title: string = __( 'Authentication error' );
		let message: string = __( 'Please try again.' );

		if ( payload.error instanceof Error && payload.error.message.includes( 'access_denied' ) ) {
			title = __( 'Authorization denied' );
			message = __(
				'It looks like you denied the authorization request. To proceed, please click "Approve"'
			);
		}

		void getIpcApi().showErrorMessageBox( { title, message } );
		return;
	}

	void store.dispatch( authTokenReceived( { token: payload.token, locale } ) );
} );
```

apps/studio/src/components/auth-provider.tsx
```diff
@@ -38,35 +36,25 @@ export const AuthContext = createContext< AuthContextType >( {
...
-			setWpcomClient( undefined );
-			setUser( undefined );
-		} catch ( err ) {
-			console.error( 'Failed to handle invalid token:', err );
-			Sentry.captureException( err );
-		}
-	}, [] );
+	useEffect( () => {
+		void dispatch( initializeAuth( { locale } ) );
+	}, [ dispatch, locale ] );
```
fredrikekelund: This can be dispatched in `apps/studio/src/stores/index.ts` instead.

fredrikekelund (commented):
Thanks for tackling this, @katinthehatsite.

The end goal with this refactor should be to remove `AuthProvider` altogether. That, and doing something about `setWpcomClient` from `src/stores/wpcom-api` are the two big changes needed here, IMO.

apps/studio/src/components/auth-provider.tsx
katinthehatsite: this was adjusted in https://github.com/Automattic/studio/pull/2778/changes/b6bc9fb9afe284025585c7a9403023dfa366bdd0

apps/studio/src/components/auth-provider.tsx
katinthehatsite: This was adjusted in https://github.com/Automattic/studio/pull/2778/changes/fe87aadcd851e98288a5e3a5530c93c489da3850

apps/studio/src/stores/auth-slice.ts
katinthehatsite: Made some adjustments for that in https://github.com/Automattic/studio/pull/2778/changes/319cc7dda3de732fcb97698aeb7038944f485020

apps/studio/src/components/auth-provider.tsx
katinthehatsite: Should be done ✅

katinthehatsite:
I think all the comments have now been addressed so I am requesting another review on this. 🙇

apps/studio/src/hooks/use-auth.ts
fredrikekelund, line 1: Same as with `apps/studio/src/components/auth-provider.tsx`, I argue we should remove this file. Components can easily consume the selectors used in this hook directly, and dispatch the `authLogout` action directly, too.

apps/studio/src/stores/auth-slice.ts
fredrikekelund: This isn't exactly a blocker, but I still think it's good practice.

apps/studio/src/stores/wpcom-client.ts
fredrikekelund, line 1: Could we make the `wpcomClient` part of the auth slice? Either part of the state or keep using the current logic, but export `getWpcomClient`. This is what I was getting at in https://github.com/Automattic/studio/pull/2778#discussion_r2938684681, too

apps/studio/src/stores/auth-slice.ts
```diff
@@ -0,0 +1,222 @@
...
+	user: AuthUser | null;
+};
+
+const initialState: AuthState = {
+	isAuthenticated: false,
+	user: null,
+};
+
+function createWpcomClient( token?: string, locale?: string, onInvalidToken?: () => void ): WPCOM {
+	const addLocaleToParams = ( params: WpcomParams ) => {
```
fredrikekelund:
```suggestion
function createWpcomClient( token?: string, onInvalidToken?: () => void ): WPCOM {
	const addLocaleToParams = ( params: WpcomParams ) => {
		const state = store.getState();
		const locale = state.i18n.locale;
```
We can read `locale` directly from the state tree here.

apps/studio/src/stores/index.ts
```diff
@@ -291,6 +299,14 @@ async function startPullPoller( selectedSiteId: string, remoteSiteId: number ) {
 	}
 }
 
+// Initialize auth once locale is loaded, and re-initialize when locale changes
+startAppListening( {
+	matcher: isAnyOf( initializeUserLocale.fulfilled, saveUserLocale.fulfilled ),
+	effect( _, listenerApi ) {
+		void listenerApi.dispatch( initializeAuth() );
+	},
+} );
```
fredrikekelund: When considering my suggestion in `createWpcomClient` to read the locale directly from the state tree, I think it'd be better to call `initializeAuth` in a different way. Following the `loadBetaFeatures` example and just calling `void store.dispatch( initializeAuth() );` directly in this file seems like a good approach.

apps/studio/src/stores/auth-slice.ts
```diff
@@ -0,0 +1,222 @@
...
+
+		if ( ! payload.token ) {
+			void dispatch( authLogout( { isOffline: true } ) );
+			return;
+		}
+
+		const locale = getState().i18n.locale;
+		void dispatch( authTokenReceived( { token: payload.token, locale } ) );
+	} );
+}
```
fredrikekelund:
```suggestion
window.ipcListener.subscribe( 'auth-updated', ( _event, payload ) => {
	if ( 'error' in payload ) {
		let title: string = __( 'Authentication error' );
		let message: string = __( 'Please try again.' );

		if ( payload.error instanceof Error && payload.error.message.includes( 'access_denied' ) ) {
			title = __( 'Authorization denied' );
			message = __(
				'It looks like you denied the authorization request. To proceed, please click "Approve"'
			);
		}

		void getIpcApi().showErrorMessageBox( { title, message } );
		return;
	}

	if ( ! payload.token ) {
		void store.dispatch( authLogout( { isOffline: true } ) );
		return;
	}

	const locale = store.getState().i18n.locale;
	void store.dispatch( authTokenReceived( { token: payload.token, locale } ) );
} );
```
Let's put the event handler in the top scope of this file and just call `store.getState()` and `store.dispatch()` directly.

apps/studio/src/stores/tests/auth-slice.test.ts
fredrikekelund, line 1: I don't think these tests provide much value, tbh. AI agents bang out tests so easily, but they don't really consider how far the type definitions already take us, how much mocking needs to happen (the more mocks, the less value from the tests), whether we're testing different inputs/outputs or just asserting that calling function A triggered a call to mock B.

That's all to say that I would argue we might as well remove this file. Up to you, @katinthehatsite, but that's my view.

apps/studio/src/stores/auth-slice.ts
```diff
@@ -0,0 +1,222 @@
...
+
+		try {
+			await getIpcApi().clearAuthenticationToken();
+			setWpcomClient( undefined );
+			setSentryWpcomUserIdRenderer( undefined );
+		} catch ( err ) {
+			console.error( err );
+		}
+	}
+);
```
fredrikekelund:
```suggestion
export const authLogout = createTypedAsyncThunk( 'auth/logout', async () => {
	const client = getWpcomClient();

	if ( ! navigator.onLine && client ) {
		try {
			await client.req.del( {
				apiNamespace: 'wpcom/v2',
				path: '/studio-app/token',
				method: 'DELETE',
			} );
			console.log( 'Token revoked' );
		} catch ( err ) {
			console.error( 'Failed to revoke token:', err );
		}
	} else if ( navigator.onLine ) {
		console.log( 'Offline: Skipping token revocation request' );
	}

	try {
		await getIpcApi().clearAuthenticationToken();
		setWpcomClient( undefined );
		setSentryWpcomUserIdRenderer( undefined );
	} catch ( err ) {
		console.error( err );
	}
} );
```
We can read the `navigator.onLine` prop directly instead of passing it as an argument to this function. It's true that we know (or assume) that the token has already been rejected sometimes when calling this function, but I think it's OK to call the `/studio-app/token` endpoint anyway.

fredrikekelund (commented):
Sorry for the delayed review, @katinthehatsite. This is looking pretty good overall! 👍

The biggest remaining question to me is `use-auth.ts`, which I argue we should also remove. Beyond that, I left a couple of comments about plumbing/wiring and raised a few comments from my previous review that you didn't address or push back on.

katinthehatsite:
Thanks so much for the review @fredrikekelund 🙇 

I am planning to take a closer look today or tomorrow and address the rest of the comments 😃

apps/studio/src/stores/auth-slice.ts
katinthehatsite: Fixed in https://github.com/Automattic/studio/pull/2778/changes/bc04dec08b4b9cb147833b6af536158a5a6cbd98

apps/studio/src/stores/tests/auth-slice.test.ts
katinthehatsite: I removed the file in https://github.com/Automattic/studio/pull/2778/changes/eca9dc4ede6624c7f590d236d39a28df1d51af75

apps/studio/src/stores/auth-slice.ts
katinthehatsite: Fixed in https://github.com/Automattic/studio/pull/2778/changes/eca9dc4ede6624c7f590d236d39a28df1d51af75

apps/studio/src/stores/index.ts
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2778/changes/b511fb86f3f619468ee422ce6d29a4585fab98f9

apps/studio/src/stores/auth-slice.ts
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2778/changes/b511fb86f3f619468ee422ce6d29a4585fab98f9

apps/studio/src/hooks/use-auth.ts
katinthehatsite: I went ahead and removed it 👍

katinthehatsite:
I think this should be ready for another review @fredrikekelund

wojtekn:
@katinthehatsite should we merge trunk in and solve conflicts?

fredrikekelund:
@katinthehatsite, given that we are cutting over to the agentic UI soon and that landing this would still require a non-trivial amount of work, I suggest abandoning this PR and closing the related issue.

katinthehatsite:
> @katinthehatsite, given that we are cutting over to the agentic UI soon and that landing this would still require a non-trivial amount of work, I suggest abandoning this PR and closing the related issue.

This sounds good to me 👍 I will proceed with closing

PR #2781, opened 2026-03-12
Add Figma to WP

mtias:
WIP

Adds figma mcp support and command workflow to convert a figma design to a WP site. It's minimal footprint, requires a frame in figma to be named with `WP_` and it'd be treated as a page source.


apps/cli/index.ts
```diff
@@ -1,3 +1,4 @@
+process.env.ENABLE_STUDIO_AI = 'true';
```
youknowriad, line 1: You should remove this :)

sejas:
@mtias , I see this PR says is a WIP. Let me know when it's ready for review, or if you prefer to delegate it to another person, or close it. Thank you!

If it’s still a work in progress, it would be great if we could convert the PR to a draft.

Screenshot 2026-04-07 at 14 42 53

PR #2783, opened 2026-03-12
Fix AI agent message display between tool calls

youknowriad:
## Related issues

Related to AI chat functionality

## How AI was used in this PR

AI helped identify the root cause by analyzing the message handling flow in the CLI chat UI and design the minimal fix.

## Proposed Changes

- Reset `currentMarkdown` state when tool results arrive so each assistant reply between tool calls creates a fresh visual block
- Prevents the first assistant message from being updated/overwritten as the agent loop progresses

## Testing Instructions

1. Run the AI command: `npm run cli:build && node apps/cli/dist/cli/main.js ai`
2. Ask the agent to perform a task with tool calls (e.g., "read package.json and summarize it")
3. Verify each assistant reply between tool calls appears as a separate block with its own blue dot marker
4. Confirm text within a single assistant message still concatenates correctly

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/ui.ts
+++ b/apps/cli/ai/ui.ts
@@ -1479,6 +1479,10 @@ export class AiChatUI {
 				} else {
 					this.showToolResult( message, toolCall?.name, toolCall?.input );
 				}
+				// Close the current markdown block so the next assistant text
+				// creates a fresh visual block (mirrors askUser / endAgentTurn).
+				this.currentMarkdown = null;
+				this.currentResponseText = '';
 				return undefined;
 			}
 			case 'result': {


nerrad (commented):
Worked well for me:

CleanShot 2026-03-12 at 17 02 46@2x

But I don't think the package-lock.json change is expected?

youknowriad:
Those package lock changes come and go for me, I think they are node version related. I reverted them.

nerrad (approved):
Looks like the force push brought back the `package-lock.json`

I'll preapprove because the functionality looks good to me. But probably should avoid the package-lock.json in the merge?

PR #2789, opened 2026-03-12
Add classic-to-blocks command

mtias:
Introduce a workflow to convert a classic WP theme into a block theme, ensuring the rendering is faithful to the original.

### DIFF
--- a/apps/cli/ai/slash-commands.ts
+++ b/apps/cli/ai/slash-commands.ts
@@ -12,10 +12,15 @@ export const AI_CHAT_LOGOUT_COMMAND = '/logout';
 export const AI_CHAT_MODEL_COMMAND = '/model';
 export const AI_CHAT_PROVIDER_COMMAND = '/provider';
 export const AI_CHAT_EXIT_COMMAND = '/exit';
+export const AI_CHAT_CLASSIC_TO_BLOCKS_COMMAND = '/classic-to-blocks';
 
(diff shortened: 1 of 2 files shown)


apps/cli/ai/slash-commands.ts
```diff
@@ -4,19 +4,75 @@ export interface SlashCommandDef {
...
+- **templates/index.html**: Main template using blocks.
+- **templates/page.html**: Page template.
+- **templates/single.html**: Single post template.
+- **templates/archive.html**: Archive/category template.
+- **templates/404.html**: 404 page.
+- **templates/front-page.html**: If the original has a custom front page.
+- **parts/header.html**: Header template part — must match the original header exactly (logo, navigation, any top bars).
+- **parts/footer.html**: Footer template part — must match the original footer exactly.
+
+### Conversion rules:
```
youknowriad, line 69: Do you find this useful? I wonder if the LLM is capable of figuring these by itself?

youknowriad:
In my test I ended up with a block theme with multiple HTML blocks instead of Core blocks. I wonder if we should address this or not. The command is a good start anyway.

youknowriad (approved):

mtias:
@youknowriad yes, we should definitely iterate on it. I'm thinking we need a /html-to-blocks loop that can be invoked in many of these workflows.

PR #2794, opened 2026-03-13
Fix issue where process manager opens terminal windows on Windows

fredrikekelund:
## Related issues

- Fixes STU-1401

## How AI was used in this PR

AI suggested the implementation, I verified it against what pm2 does. 

## Proposed Changes

- Use `child_process.spawn` instead of `child_process.fork` in `process-manager-daemon.ts` so that we can pass the `windowsHide` option. 
- For good measure, also pass the `windowsHide` option to the `child_process.spawn` call that instantiates the process manager daemon. 

## Testing Instructions

1. Run Windows
2. Run `npm start`
3. Start a site
4. Ensure that no terminal window is opened

## Pre-merge Checklist


apps/cli/__mocks__/pm2-axon.ts
fredrikekelund, line 1: Not related to the core changes in this PR. This is just a leftover file that I should've removed in https://github.com/Automattic/studio/pull/2690.

epeicher (approved):
Thanks for the fix @fredrikekelund! I have tested on Windows and I can confirm that I don't see the terminal windows anymore after applying these changes. LGTM!

PR #2796, opened 2026-03-13
Remove redundant CLI todo action output

nerrad:
## Related issues

- Related to CLI todo list rendering cleanup.

## How AI was used in this PR

- Used AI assistance to refactor TodoWrite rendering into a dedicated helper and remove the redundant per-action CLI output. Reviewed the resulting diff and ran local verification before opening and updating this PR.

## Proposed Changes

- stop rendering redundant `Added todo` and `Completed todo` lines for CLI todo updates
- render only the updated todo list snapshot in the CLI output
- keep the TodoWrite rendering logic isolated in a dedicated helper

## Testing Instructions

- Run `npx eslint --fix apps/cli/ai/ui.ts apps/cli/ai/todo-render.ts`
- Run `npm run typecheck`
- In the CLI chat UI, trigger a todo list update and confirm only the updated list is shown without redundant action statements

## Pre-merge Checklist


youknowriad (approved):

apps/cli/ai/todo-render.ts
```diff
@@ -0,0 +1,31 @@
+import chalk from 'chalk';
```
youknowriad, line 1: My guess is that this is the first time we extract some rendering logic from "ui.ts". So I wonder if a "ui/index.ts", "ui/todos.ts" is better

nerrad: Fair point. To make sure I fully understand though, you mean we'd have `ui` as a subfolder in the path? So `apps/cli/ai/ui/...`. If so, that would help organize things a bit better. But can also be done as other helpers/tools are developed.

youknowriad: yes, that's what I meant, yeah no urgency here

PR #2793, opened 2026-03-13
CLI: Implement decoupled config file strategy

bcotrim:
## Related issues

- Fixes STU-1350

## How AI was used in this PR
Used AI to resolve merge conflicts with trunk

## Proposed Changes

This is the development branch for STU-1350. It will mostly be reviewed in separate PRs, with the exclusion of fixes and adjustments needed from trunk merges.
List of PRs:
-  https://github.com/Automattic/studio/pull/2701
- https://github.com/Automattic/studio/pull/2731
- https://github.com/Automattic/studio/pull/2807
- https://github.com/Automattic/studio/pull/2821
- https://github.com/Automattic/studio/pull/2836

## Testing Instructions

- TBD

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #2799, opened 2026-03-13
Update desktop app site state when making CLI changes

nb:
## Reproduce

* Start from a stopped site
* Open the desktop app
* `studio site start --path ~/Studio/<existing-site>`
* Site starts successfully
* The desktop app doesn't indicate the site is running

## How AI was used in this PR

Used Codex to trace the CLI-to-app event flow, identify the missing registration path for externally started sites, and draft the initial code and test changes. I reviewed the final implementation, adjusted the test coverage, and ran the verification commands locally.

## Proposed Changes

- Register missing sites when the desktop app receives `site-updated` events from the CLI event subscriber, not only on `site-created`
- This allows sites started outside the app UI, including `studio ai` and `studio site start`, to appear as running in the app
- Add a regression test covering an externally started site flowing through the CLI event subscriber
- Make development startup more robust by treating DevTools service worker startup failures as non-fatal, as I hit this issue
- Add a test covering the dev-only startup fallback path

## Testing Instructions

- Start the Studio dev app:
  - `npm start`
- In a second terminal, build and run the local CLI:
  - `npm run cli:build`
  - `node apps/cli/dist/cli/main.js site start --path ~/Studio/<existing-site>`
- Confirm the started site appears as running in the dev app
- Run verification:
  - `npx eslint --fix apps/studio/src/index.ts apps/studio/src/tests/index.test.ts apps/studio/src/modules/cli/lib/cli-events-subscriber.ts apps/studio/src/modules/cli/lib/tests/cli-events-subscriber.test.ts`
  - `npm test -- apps/studio/src/tests/index.test.ts apps/studio/src/modules/cli/lib/tests/cli-events-subscriber.test.ts`
  - `npm run typecheck`

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/index.ts
+++ b/apps/studio/src/index.ts
@@ -148,6 +148,16 @@ function launchExtensionBackgroundWorkers( appSession = session.defaultSession )
 	);
 }
 
+async function setupDevelopmentExtensions(): Promise< void > {
+	try {
+		await installExtension( REACT_DEVELOPER_TOOLS );
+		await installExtension( REDUX_DEVTOOLS );
+		await launchExtensionBackgroundWorkers();
+	} catch ( error ) {
+		console.warn( 'Failed to initialize development extensions:', error );
+	}
+}
+
 async function appBoot() {
 	app.setName( packageJson.productName );
 
@@ -263,9 +273,7 @@ async function appBoot() {
 	app.on( 'ready', async () => {
 		const locale = await getUserLocaleWithFallback();
 		if ( process.env.NODE_ENV === 'development' ) {
-			await installExtension( REACT_DEVELOPER_TOOLS );
-			await installExtension( REDUX_DEVTOOLS );
-			await launchExtensionBackgroundWorkers();
+			await setupDevelopmentExtensions();
 		}
 
 		console.log( `App version: ${ app.getVersion() }` );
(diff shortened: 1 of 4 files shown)


bcotrim (commented):
Thanks for the change @nb 
I tried, but couldn't replicate the start issue `trunk` — tested starting a site via CLI both with Studio open and before opening Studio, and both cases showed the correct running state.
Could you share more detailed steps/logs? 

The `setupDevelopmentExtensions` looks great!

nb:
@bcotrim  you're right, I was testing with 1.7.5 where it was broken, but it does to work with `trunk`! Looking back at my diff, I actually didn't change the logic :) It's a bit cleaner now and given the extra test, I'd still suggest merging it. Sorry for the hassle.

bcotrim:
> @bcotrim you're right, I was testing with 1.7.5 where it was broken, but it does to work with `trunk`! Looking back at my diff, I actually didn't change the logic :) It's a bit cleaner now and given the extra test, I'd still suggest merging it. Sorry for the hassle.

Unless I am missing something it does change the behavior, from creating sites on Studio app only on CREATED events, to creating sites on Studio app for any received event.
The current condition was introduced to prevent some issues, like duplicated sites, here: https://github.com/Automattic/studio/pull/2415
Since we can’t replicate the original issue anymore, I’d suggest keeping the current behavior unchanged.
Adding the tests and the development startup are still good improvements.

bcotrim:
Hey @nb 👋 

With the decoupled config strategy landing on trunk (#2793), I think the main fix here (registering unknown sites on UPDATED events) is mostly covered now.

The devtools worker fix is still valid but unrelated, happy to land that separately if you want to open a fresh PR for it, or if you want I can help with that.

I think we can close this one, let me know if you see a scenario I’m missing!

nb:
Sure!

PR #2805, opened 2026-03-13
Fix site failing to start when it has a PHP error

gcsecsey:
## Related issues

- Fixes STU-1370

## How AI was used in this PR

AI assisted with drafting the implementation, exploring the Playground CLI internals, the feasibility of using `mount-only` mode, and reviewing the final code for reuse, quality, and efficiency issues.

## Proposed Changes

When a site has a PHP fatal error (eg. `this_function_does_not_exist()` in `functions.php`), the server child process crashes with `process.exit(1)` and the user sees a generic "Failed to start site" error dialog. This happens because Playground's internal error handler calls `process.exit(1)` directly in a `.catch()`, bypassing all try-catch blocks. The user has no way to know what went wrong or fix it without digging through logs.

This PR catches PHP errors during startup and shows them in the browser instead:

- Adds a `runCLIWithoutExit()` wrapper in `wordpress-server-child.ts` that overrides `process.exit` to throw instead of exiting, so PHP fatal errors become catchable exceptions. It also intercepts `http.createServer` to capture Playground's orphaned HTTP server, which is already bound to the port after a failed boot.
- Repurposes the orphaned server as an error page server. When a PHP error is detected, its request handler is replaced with one that serves a styled error page showing the parsed error message. This avoids EADDRINUSE since the server is already holding the port.
- Adds a `watchForPhpChanges()` file watcher that watches the site directory for `.php` file changes. When the user fixes the error, the watcher closes the orphaned server, attempts a full restart, and on success restores normal operation (blueprints applied, admin credentials set).
- Extracts PHP error handling into a new `apps/cli/lib/php-error-handling.ts` module.
- Captures console and stdout output during boot so the actual PHP error message can be parsed and displayed, rather than showing the generic "exit code 1" message.
- Adds `uncaughtException` and `unhandledRejection` handlers to prevent stale async error callbacks from Playground's failed boot from crashing the child process.

## Other approaches I explored

I also investigated using Playground's `mount-only` mode as an alternative approach. The idea was to retry with `mount-only` mode after a failed normal boot, skipping the WordPress boot (where the fatal error occurs) and starting the HTTP server with all site files mounted. Each browser request would then trigger fresh PHP execution, letting WordPress display errors naturally per-request, the same behavior you see when a fatal error occurs on an already-running server.

However, `mount-only` is a V2 runner concept only ([`worker-thread-v2.ts`](https://github.com/WordPress/wordpress-playground/blob/trunk/packages/playground/cli/src/worker-thread-v2.ts)). The V1 worker that Studio currently uses silently ignores the `mode` parameter.

This approach would be worth revisiting once the V2 runner ships and `mount-only` becomes available.

## Testing Instructions

- Start Studio and create a new site
- Edit the site's `functions.php` file and add an invalid function call like `this_function_does_not_exist();`
- Start the site. Check that the site starts, and the site preview updates.
- Open the site. Check that it displays a "PHP Error Detected" page in the browser with the specific error message.
- Fix the error in `functions.php`
- The site should automatically restart and be fully functional
- Check for regressions by quitting and reopening Studio and letting the site start up

https://github.com/user-attachments/assets/5d0ae82e-0a6d-4b76-9cca-d1092c14ba09

## Pre-merge Checklist


apps/cli/wordpress-server-child.ts
```diff
@@ -272,6 +285,123 @@ function wrapWithStartingPromise< Args extends unknown[], Return extends void >(
...
+					logToConsole( 'Restart failed, still watching…' );
+					if ( orphanedServer ) {
+						const errorMsg = parsePhpError( lastCapturedOutput || 'PHP error during startup' );
+						serveErrorPage( orphanedServer, errorMsg );
+					}
+				} finally {
+					retrying = false;
+				}
+			} )();
+		}, 2000 );
```
katinthehatsite: This could be potentially moved to the constant but it is a minor things. Can be left as is

katinthehatsite (commented):
I spent some time testing this and here are my observations:

* I think it is a nice UI improvement for the user when there is an error on the site start:

Image

I found a message a bit confusing on the site itself because when I am done fixing the error and **if I keep the site open in the browser**, Studio does refresh the thumbnail but I see that my site in the browser does not restart. Perhaps, the message could say something regarding closing this tab and accessing the site again from Studio when the error is fixed.

* If the site is running and I break it in the meantime, the error display in this format:

Image

I think this PR is not expected to fix that but I wanted to note it down here as well.

apps/cli/wordpress-server-child.ts
```diff
@@ -272,6 +285,123 @@ function wrapWithStartingPromise< Args extends unknown[], Return extends void >(
...
+
+	// Enable output capture via the global interceptors
+	capturedBootOutput = [];
+
+	process.exit = ( ( code?: number ) => {
+		if ( code !== 0 ) {
+			throw new Error( `WordPress server startup failed (exit code ${ code })` );
+		}
+		return originalExit( code );
+	} ) as typeof process.exit;
```
fredrikekelund: I didn't realize Playground CLI calls `process.exit()` internally. That's a very clear culprit.

fredrikekelund (commented):
Overriding `process.exit` and `http.createServer` strikes me as a somewhat fragile approach. 

To me, the ideal outcome here would be that Playground CLI manages to start the site even if there are errors in the PHP code. If I add a nonexistent function to my WP theme _after_ starting my site, Playground CLI displays the error I'd expect in my browser, so we should focus on the boot sequence. 

Also, ideally, `runCLI` shouldn't call `process.exit()` – at least not unless we know Playground CLI is being invoked as a CLI. That's just not a safe operation for a function that's exposed in the API.

apps/cli/wordpress-server-child.ts
gcsecsey: I experimented a bit with moving this logic to the main process, and letting the child process exit. Still, I agree that the cleanest solution would be for the Playground server to keep running and serve the WordPress error page.

gcsecsey:
Thanks for the reviews @fredrikekelund and @katinthehatsite!

> To me, the ideal outcome here would be that Playground CLI manages to start the site even if there are errors in the PHP code.

Yes, I agree. As an alternative to trying to patch this on our end, we could also open a Playground PR that updates `runCli()` to throw/reject instead of exiting. 

> If I add a nonexistent function to my WP theme after starting my site, Playground CLI displays the expected error in my browser, so we should focus on the boot sequence.

This is true, but as you also [pointed out above](https://github.com/Automattic/studio/pull/2805#discussion_r2952859842), the process is exited from Playground, so we're unable to reuse the same server process if it fails during site startup.

fredrikekelund:
> If I add a nonexistent function to my WP theme after starting my site, Playground CLI displays the expected error in my browser, so we should focus on the boot sequence.

What I meant by this is that it seems achievable that Playground would start the site even if the PHP code contains an error, because it manages to display the error in the browser if it's added after the site was started.

I've only skimmed your most recent changes, @gcsecsey, but moving the error recovery up into the process manager is a much cleaner solution, IMO.

Still, it'd be even better if Playground could just start the site and display the error in the browser, so I'd argue we could take a stab at solving this directly in Playground. The Orbit team should be able to help us with directions or, indeed, even take on the implementation.

fredrikekelund:
@gcsecsey, can we mark this PR as a draft until further notice, to avoid it being included in the daily Slack ping?

gcsecsey:
> @gcsecsey, can we mark this PR as a draft until further notice, to avoid it being included in the daily Slack ping?

Sure, I converted it to draft for the time being. 👍

gcsecsey:
I raised https://github.com/WordPress/wordpress-playground/pull/3478 in Playground to start the server even when there are errors, and also to throw errors instead of calling `process.exit(1)`. If those changes are released, we can update this PR to handle those thrown errors.

fredrikekelund:
The native PHP runtime fixes the underlying issue here (i.e., it can start even if the PHP code contains errors).

I still think this behavior should change in Playground, and if https://github.com/WordPress/wordpress-playground/pull/3478 can address that, that's great. However, in the interest of cleaning up the PR backlog in this repo, I'm closing this PR.

I believe it would be more straightforward to open a new PR if and when https://github.com/WordPress/wordpress-playground/pull/3478 is merged and released, but feel free to reopen this PR later if needed, @gcsecsey

gcsecsey:
> I believe it would be more straightforward to open a new PR if and when [WordPress/wordpress-playground#3478](https://github.com/WordPress/wordpress-playground/pull/3478) is merged and released, but feel free to reopen this PR later if needed, @gcsecsey

Thanks for following up on this @fredrikekelund. Also, thanks for closing this, I don't think these changes are needed even after https://github.com/WordPress/wordpress-playground/pull/3478 is merged. I'll follow up on that PR as it might have missed some reviews due to RSM.

PR #2807, opened 2026-03-13
Apply decoupled config strategy to preview sites

bcotrim:
## Related issues

- Related STU-1350

## How AI was used in this PR

TBD

## Proposed Changes

Extends the decoupled config strategy (implemented for sites in PR #2701) to preview sites:

- **CLI now owns preview site data**: Snapshots are stored in `~/.studio/cli.json` instead of appdata-v1.json
- **New `cli-config.ts`**: Centralized config management with Zod schema validation and file locking for safe concurrent writes
- **Studio reads via CLI**: Renderer calls `preview list --format json` to fetch snapshots from the CLI config
- **New `preview set` command**: Allows updating snapshot settings (e.g., `preview set <hostname> --name "New Name"`). Follows the same pattern as `site set`
- **`preview create` now supports `--name`**: Optionally name a preview site at creation time
- **Redux store initialization**: Moved `refreshSnapshots()` to store setup (avoiding module-load timing issues in tests)

This maintains consistency with the decoupled strategy: CLI is the source of truth for data, Studio accesses it via CLI commands.

## Testing Instructions

1. **Build and test locally**: `npm run cli:build && npm test`
2. **Create a preview site with a name**: `studio preview create /path/to/site --name "My Preview"`
3. **List previews**: `studio preview list --format json` (shows all snapshots)
4. **Rename a preview**: `studio preview set <hostname> --name "New Name"`
5. **Verify in Studio UI**: Snapshots load correctly and display the new names
6. **Verify file location**: Check that `~/.studio/cli.json` contains the snapshots (not appdata)

## Pre-merge Checklist

- [x] All tests pass (typecheck, linter, unit tests)
- [x] No TypeScript, React, or console errors
- [x] File locking used for concurrent config access (cli-config.ts)
- [x] Zod validation for all config reads/writes
- [x] Tests updated to match new data flow


fredrikekelund:
Looks like we should change the base branch here, @bcotrim

apps/cli/lib/daemon-client.ts
```diff
@@ -331,18 +331,16 @@ export async function stopProcess( processName: string ): Promise< void > {
...
- *
- * @param event - The event topic (e.g., 'site-created', 'site-updated', 'site-deleted')
- * @param data - The event data (must include siteId)
+ * Emit a CLI event via the events socket, for the `_events` command server to receive.
  */
-export async function emitSiteEvent(
-	event: SITE_EVENTS,
-	data: { siteId: string }
-): Promise< void > {
+export async function emitCliEvent( payload: CliEventPayload ): Promise< void > {
```
fredrikekelund, line 341: Very gooood 👍

tools/common/lib/cli-events.ts
```diff
@@ -0,0 +1,107 @@
...
+	event: z.enum( SITE_EVENTS ),
+	siteId: z.string(),
+	site: siteDetailsSchema.optional(),
+	running: z.boolean(),
+} );
+
+export type SiteEvent = z.infer< typeof siteEventSchema >;
+
+export const snapshotEventSchema = z.object( {
+	event: z.nativeEnum( SNAPSHOT_EVENTS ),
```
fredrikekelund:
```suggestion
	event: z.enum( SNAPSHOT_EVENTS ),
```
`nativeEnum` is deprecated in zod v4

tools/common/lib/cli-events.ts
```diff
@@ -0,0 +1,107 @@
...
+ */
+export const siteSocketEventSchema = z.object( {
+	event: z.string(),
+	data: z.object( {
+		siteId: z.string(),
+	} ),
+} );
+
+export const snapshotSocketEventSchema = z.object( {
+	event: z.nativeEnum( SNAPSHOT_EVENTS ),
```
fredrikekelund:
```suggestion
	event: z.enum( SNAPSHOT_EVENTS ),
```
`nativeEnum` is deprecated in zod v4

apps/studio/src/stores/snapshot-slice.ts
```diff
@@ -363,6 +358,65 @@ function isBulkOperationSettled( bulkOperation: BulkOperation ) {
 	} );
 }
 
+window.ipcListener.subscribe( 'snapshot-changed', ( event, snapshotEvent: SnapshotEvent ) => {
```
fredrikekelund:
```suggestion
window.ipcListener.subscribe( 'snapshot-changed', ( event, snapshotEvent ) => {
```
The type is already defined in `apps/studio/src/ipc-utils.ts`

apps/cli/commands/preview/set.ts
fredrikekelund, line 1: I still expect the CLI and the app to use a shared config file that they can both read from and write to. I don't think we can get away from that. The language setting and the auth data are two good examples of things that the app shouldn't or couldn't use the CLI to write to the shared config.

With that in mind, I'd argue it's a bit overkill to introduce a CLI command just to update the name of a preview site. It's possible that CLI-only users would also want this ability, it just seems much less important in the CLI than it does in the UI. 

I'm not necessarily arguing we need to walk back this change – if it works, then it works. Still, I think there are better arguments for why we need the `site set` command (it lets us consolidate HTTPS certificate logic in the CLI, it allows us to use only the CLI to talk to the process manager daemon, etc). In this case, we're basically just wrapping a file write operation in another process.

apps/cli/commands/_events.ts
fredrikekelund, line 1: Nice 👍

apps/cli/commands/preview/create.ts
```diff
@@ -67,8 +84,14 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 	return yargs.command( {
 		command: 'create',
 		describe: __( 'Create a preview site' ),
+		builder: ( yargs ) => {
+			return yargs.option( 'name', {
+				type: 'string',
+				description: __( 'Preview site name' ),
+			} );
+		},
```
fredrikekelund, line 93: Nice addition 👍

apps/cli/commands/preview/list.ts
```diff
@@ -20,13 +20,18 @@ export async function runCommand(
 	const logger = new Logger< LoggerAction >();
 
 	try {
+		if ( outputFormat === 'json' ) {
+			const config = await readCliConfig();
+			logger.reportKeyValuePair( 'snapshots', JSON.stringify( config.snapshots ) );
+			return;
+		}
```
fredrikekelund, line 29: Could we _also_ log to stdout here? I know I've argued that using `reportKeyValuePair` is safer, it just seems odd that of a user passes `--format json`, they get nothing back…

apps/cli/lib/daemon-client.ts
```diff
@@ -331,18 +331,16 @@ export async function stopProcess( processName: string ): Promise< void > {
 
 const eventsSocketClient = new SocketRequestClient( SITE_EVENTS_SOCKET_PATH );
 
+type CliEventPayload =
+	| { event: SITE_EVENTS; data: { siteId: string } }
+	| { event: SNAPSHOT_EVENTS; data: { snapshotUrl: string } };
```
fredrikekelund, line 336: Nice 👍

apps/cli/lib/cli-config/index.ts
fredrikekelund, line 1: The performance drawbacks of "barrel files" like this don't come into play much in this context, but I would still argue that we shouldn't use them. I see no good reason for it apart from the vanity of shorter import paths. 

Other than that: good idea to split up the config file functions into separate files.

apps/studio/src/modules/preview-site/lib/ipc-handlers.ts
```diff
@@ -1,9 +1,54 @@
...
+
+			emitter.on( 'success', () => resolve( [] ) );
+			emitter.on( 'failure', ( { error } ) => reject( error ) );
+			emitter.on( 'error', ( { error } ) => reject( error ) );
+		} );
+	} catch ( error ) {
+		console.error( 'Failed to fetch snapshots from CLI:', error );
+		return [];
+	}
+}
```
fredrikekelund, line 39: Similar to what I mentioned in `apps/cli/commands/preview/set.ts`, it seems like we're adding overhead here without super clear gains. We still need a zod schema to parse the data we receive, so why not just read directly from the shared config file?

We don't necessarily need to walk back this change (it's better to move ahead faster), but I would like to at least raise the question. It goes back to what we discussed in our huddle, @bcotrim, but if the end goal truly is to never have the app read from the shared config file, then I might need some more convincing 😄

apps/studio/src/stores/snapshot-slice.ts
```diff
@@ -485,6 +539,9 @@ window.ipcListener.subscribe( 'snapshot-success', ( event, payload ) => {
 			} );
 		}
 	}
+
+	// Re-fetch snapshots from CLI after any successful operation
+	void refreshSnapshots();
```
fredrikekelund: Do we really need this? 🤔 Seems like we wouldn't need such a fine-grained event system if we always just refetch the preview site list anyway.

apps/studio/src/stores/snapshot-slice.ts
```diff
@@ -363,6 +358,65 @@ function isBulkOperationSettled( bulkOperation: BulkOperation ) {
...
+		void refreshSnapshots();
+		return;
+	}
+
+	if ( eventType === SNAPSHOT_EVENTS.CREATED ) {
+		const state = store.getState();
+		const existing = state.snapshot.snapshots.find( ( s ) => s.url === snapshot.url );
+		if ( ! existing ) {
+			store.dispatch(
+				snapshotSlice.actions.setSnapshots( {
```
fredrikekelund:
```suggestion
				snapshotSlice.actions.addSnapshot( {
```
Optional, but I think we might as well add an explicit `addSnapshot` action.

apps/studio/src/stores/snapshot-slice.ts
```diff
@@ -295,30 +295,25 @@ const selectSnapshotsBySiteAndUser = createSelector(
...
-				// API, because expired sites are preserved locally. Therefore, we need to ensure
-				// the count is non-negative.
 				data.siteCount = Math.max( 0, data.siteCount + countDiff );
 			} )
 		);
 
-		// Wait for changes to take effect on the back-end before invalidating the query
 		setTimeout( () => {
 			store.dispatch( wpcomApi.util.invalidateTags( [ 'SnapshotUsage' ] ) );
 		}, 8000 );
```
fredrikekelund: I think it's worth putting this logic in a reusable function. Also, I wouldn't delete the comments.

fredrikekelund (commented):
This looks good to me overall 👍 There's a question of whether we really need the app to interface with the CLI for every preview site operation, but I'm fine with landing this PR first and keeping that discussion going separately.

I haven't tested the changes yet, but I figured I'd share my feedback first. Will finalize my review once I've tested.

apps/studio/src/stores/index.ts
```diff
@@ -89,12 +84,16 @@ startAppListening( {
...
-		const state = listenerApi.getState();
-		await getIpcApi().saveSnapshotsToStorage( state.snapshot.snapshots );
+	actionCreator: updateSnapshotLocally,
+	async effect( action ) {
+		const { atomicSiteId, snapshot } = action.payload;
+		const state = store.getState();
+		const existing = state.snapshot.snapshots.find( ( s ) => s.atomicSiteId === atomicSiteId );
+		if ( existing?.url && snapshot.name !== undefined && snapshot.name !== existing.name ) {
+			await getIpcApi().setSnapshot( existing.url, { name: snapshot.name } );
+		}
```
fredrikekelund, line 97: This logic doesn't work as intended, because `existing.name` is already updated by the time this action runs.

apps/cli/commands/preview/set.ts
bcotrim: I was also wondering that during development.
What pushed me to make the decision to add name to edit and create preview is the fact that we show that information in the CLI. It seems strange to me to show a name to our users if we don't allow them to set or edit that name.

apps/studio/src/modules/preview-site/lib/ipc-handlers.ts
bcotrim: I think the value is making Studio dependent only on the CLI, without knowing about the implementation details.
The same reasoning could be applied to operations, if Studio reads and writes from the config file, why not share the code to run the operations and remove the CLI overhead?

fredrikekelund (approved):
Testing works as expected with the latest changes 👍

PR #2811, opened 2026-03-16
Remove Windows binary size tracking from metrics job

wojtekn:
## Related issues

N/A

## How AI was used in this PR

Claude identified and cleaned up the dead code after the human pointed out the issue.

## Proposed Changes

- Remove `appSizeWin` metric from `metric-config.ts` — the Performance Metrics CI job runs exclusively on Mac agents, so this metric was never reported
- Simplify `app-size.test.ts` to only check for the Mac build directory, removing the unreachable Windows `else if` branch and updating the error message to be more specific

## Testing Instructions

- No functional change — this removes dead code that was never executed in CI
- The `appSizeMac` metric continues to be tracked as before

## Pre-merge Checklist


sejas (approved):
`@\Claude` tell your human your changes make sense, and he did a great job identifying the dead code. We don't need to track Windows size if the job runs only for Mac platform. Code looks good, we have less code that wasn't executed.

PR #2818, opened 2026-03-16
Output ESM in CLI build

fredrikekelund:
## Related issues

- N/A

## How AI was used in this PR

Codex helped generate and verify the changes. The total surface area is very small, so it was easy for me to reason about, but still quicker to use Codex to get the right config in one shot. 

## Proposed Changes

We typically don't need to think too much about CommonJS vs ESM in Studio. Typescript and Vite smooth over a lot of the details of consuming packages in one format or the other. 

Still, ESM is the modern standard, and CommonJS is effectively legacy. It makes sense for our source code to stay close to the bundler output and to the Node.js platform. Therefore, this PR updates the CLI configs to output ESM instead of CommonJS. 

## Testing Instructions

CI should pass. 

## Pre-merge Checklist


eslint.config.mjs
```diff
@@ -88,5 +88,22 @@ export default defineConfig(
...
+					name: '__dirname',
+					message: 'Use import.meta.dirname in ESM modules.',
+				},
+				{
+					name: '__filename',
+					message: 'Use import.meta.filename in ESM modules.',
+				},
+			],
+		},
 	}
```
fredrikekelund, line 110: Typescript doesn't complain about using `__dirname` or `__filename` in ESM code, so we a linter warning to help capture those errors early.

package.json
```diff
@@ -40,7 +40,7 @@
 		"app:install:bundle": "npm -w studio-app run install:bundle",
 		"compare:perf": "npm -w compare-perf run compare",
 		"lint": "eslint {apps/cli,apps/studio/src,apps/studio/e2e,tools/common}",
-		"typecheck": "tsc -p tsconfig.json --noEmit",
+		"typecheck": "npm run typecheck --workspaces --if-present",
```
fredrikekelund, line 44: We should have done it this way before, too, but now it's necessary, since we have different `module` settings between the different workspaces.

tools/common/lib/tests/sort-sites.test.ts
fredrikekelund, line 1: `SiteDetails` isn't defined in `tools/common`

apps/cli/vite.config.prod.ts
```diff
@@ -1,6 +1,6 @@
 import { existsSync, rmSync } from 'fs';
 import { resolve } from 'path';
-import { sync as globSync } from 'glob';
+import glob from 'glob';
```
sejas:
```suggestion
import { GlobSync } from 'glob';
```

And then use `new GlobSync` in line 33, and also change line 38.

tools/common/globals.ts
```diff
@@ -0,0 +1 @@
+declare module 'wpcom-xhr-request';
```
sejas, line 1: I see this declaration appears in three files. Out of curiosity, would be possible to have it in one?

eslint.config.mjs
```diff
@@ -90,5 +90,22 @@ export default defineConfig(
...
+	{
+		files: [ 'apps/cli/**/*.{ts,tsx}' ],
+		ignores: [ 'apps/cli/vite.config*.ts', 'apps/cli/vitest.config.ts' ],
+		rules: {
+			'no-restricted-globals': [
+				'error',
+				{
+					name: '__dirname',
+					message: 'Use import.meta.dirname in ESM modules.',
+				},
```
sejas, line 103: Nice!

sejas (approved):
Changes look good. I just found one warning related to `GlobSync` when running eslint (I pulled trunk first to remove two other non-related warnings).

```
❯ npm run lint

> lint
> eslint {apps/cli,apps/studio/src,apps/studio/e2e,tools/common}

Multiple projects found, consider using a single `tsconfig` with `references` to speed up, or use `noWarnOnMultipleProjects` to suppress this warning

/Users/macbookpro/Documents/projects-m3.nosync/studio/apps/cli/vite.config.prod.ts
  3:8  warning  Using exported name 'glob' as identifier for default export  import-x/no-named-as-default

✖ 1 problem (0 errors, 1 warning)
```

I also left an observation about multiple `wpcom-xhr-request` declarations but I'm not sure if there is an easy solution.

apps/cli/vite.config.prod.ts
fredrikekelund: Looks like part of the issue here was that we were missing a `glob` dependency in `apps/cli`.

tools/common/globals.ts
fredrikekelund: I don't think so. Part of the change in this PR was to remove the `references` prop from the different tsconfig files. Codex did this for me in response to type errors that came from the different `module` settings between the CLI and Studio. 

We do indeed want total isolation between `apps/cli` and `apps/studio`, so this change makes sense there. 

`tools/common` is a different story – both "apps" import from there. However, TS still resolves those imports correctly, and the closer we can get to regular npm package isolation, the better, IMO.

fredrikekelund: And to answer your question more directly: removing the `references` prop is what required me to have separate definition files in each workspace with the `wpcom-xhr-request` declaration.

PR #2821, opened 2026-03-16
Wire auth and locale to shared.json, AI settings to cli.json

bcotrim:
## Related issues

- Related to STU-1350

## How AI was used in this PR

All code was generated via Claude Code (Opus + Haiku). Changes were reviewed and validated by the author through iterative conversation — verifying architecture decisions, confirming read/write boundaries, and running the full test suite at each step.

## Proposed Changes

- Create `tools/common/lib/shared-config.ts` — new shared config infrastructure at `~/.studio/shared.json` for data shared between Desktop and CLI (auth token + locale)
- Wire Desktop auth reads/writes to shared-config instead of appdata
- Wire Desktop locale reads/writes to shared-config instead of appdata
- Wire CLI auth reads/writes to shared-config instead of appdata
- Move CLI-only AI settings (`aiProvider`, `anthropicApiKey`) from appdata to `cli.json`
- Move CLI telemetry (`lastBumpStats`) from appdata to `cli.json`
- Remove `authToken`, `locale`, AI fields, and `lastBumpStats` from appdata types and schemas
- Add 20 tests for shared-config and update all affected test files

After these changes the write boundaries are:
- **Desktop** → `appdata-v1.json` (sites, snapshots, preferences) + `shared.json` (auth, locale)
- **CLI** → `cli.json` (sites, AI settings, telemetry) + `shared.json` (auth) + `appdata-v1.json` (snapshots only)

## Testing Instructions

- Run `npm test` — all 1292 tests should pass
- Manual: launch Desktop app, log in via WordPress.com — verify `~/.studio/shared.json` contains the auth token
- Manual: change language in Preferences — verify `~/.studio/shared.json` contains the locale
- Manual: run `studio auth login` from CLI — verify shared.json is updated
- Manual: run `studio ai` and switch provider/enter API key — verify `~/.studio/cli.json` contains `aiProvider` and `anthropicApiKey`

## Pre-merge Checklist


tools/common/lib/shared-config.ts
```diff
@@ -0,0 +1,126 @@
...
+
+	try {
+		const fileContent = await readFile( configPath, { encoding: 'utf8' } );
+		const data = JSON.parse( fileContent );
+		return sharedConfigSchema.parse( data );
+	} catch ( error ) {
+		if ( error instanceof z.ZodError || error instanceof SyntaxError ) {
+			return structuredClone( DEFAULT_SHARED_CONFIG );
+		}
+		throw new Error( 'Failed to read shared config file.' );
```
fredrikekelund: Same as in `apps/cli/lib/cli-config/core.ts`, we should look for differing `version` props here and say something like "A newer version of Studio or the Studio CLI is installed on your system" if the version differs.

tools/common/lib/shared-config.ts
```diff
@@ -0,0 +1,126 @@
...
+			return null;
+		}
+		const token = authTokenSchema.parse( config.authToken );
+		if ( Date.now() >= token.expirationTime ) {
+			return null;
+		}
+		return token;
+	} catch {
+		return null;
+	}
```
fredrikekelund: This relates to the error handling I suggested in `readSharedConfig`. I'm not sure what the best approach is here…

If the user is running an old version of the CLI but have a newer version of Studio installed (that uses a new schema for the shared config file), then we should prompt them to upgrade.

However, we still need both programs to recover from errors well enough to at least report the right thing to the user. This is especially true for Studio, which needs to be able to auto-update as usual.

tools/common/lib/shared-config.ts
```diff
@@ -0,0 +1,126 @@
...
+	id: z.number(),
+	email: z.string(),
+	displayName: z.string().default( '' ),
+} );
+
+export type StoredToken = z.infer< typeof authTokenSchema >;
+
+const sharedConfigSchema = z
+	.object( {
+		version: z.number().default( 1 ),
```
fredrikekelund: Let's put the version in a constant and add a comment explaining that any schema update must maintain backwards compatibility. If the change is breaking, the version should be updated and a data migration function should be added.

tools/common/lib/tests/shared-config.test.ts
fredrikekelund, line 1: The important thing to test here, IMO, is how we recover from schema-related errors. We have some of that in the `readSharedConfig` tests, but I think we can expand on that (along with updating the logic in the actual source code)

apps/cli/lib/cli-config/core.ts
```diff
@@ -119,3 +127,16 @@ export async function lockCliConfig(): Promise< void > {
 export async function unlockCliConfig(): Promise< void > {
 	await unlockFileAsync( LOCKFILE_PATH );
 }
+
+export async function updateCliConfig(
```
fredrikekelund:
```suggestion
export async function updateCliConfigWithPartial(
```
Just a suggestion

apps/cli/commands/auth/login.ts
```diff
@@ -64,27 +56,22 @@ export async function runCommand(): Promise< void > {
...
+		await updateSharedConfig( {
+			authToken: {
+				accessToken,
+				id: user.ID,
+				email: user.email,
+				displayName: user.display_name,
+				expiresIn: DEFAULT_TOKEN_LIFETIME_MS / 1000,
+				expirationTime: Date.now() + DEFAULT_TOKEN_LIFETIME_MS,
+			},
+		} );
```
fredrikekelund: Nice 👍

apps/cli/commands/preview/list.ts
```diff
@@ -29,7 +30,13 @@ export async function runCommand(
 		}
 
 		logger.reportStart( LoggerAction.VALIDATE, __( 'Validating…' ) );
-		const token = await getAuthToken();
+		await getSiteByFolder( siteFolder );
```
fredrikekelund, line 33: Good catch putting this back 👍

tools/common/lib/shared-config.ts
```diff
@@ -0,0 +1,126 @@
...
+		await lockSharedConfig();
+		const config = await readSharedConfig();
+		const updated = { ...config, ...update };
+		await saveSharedConfig( updated );
+	} finally {
+		await unlockSharedConfig();
+	}
+}
+
+export async function readAuthToken(): Promise< StoredToken | null > {
```
fredrikekelund: Exporting functions that return just parts of the config is clearly great for reducing the amount of test mocks. Nice 👍

tools/common/lib/shared-config.ts
```diff
@@ -0,0 +1,126 @@
...
+export const authTokenSchema = z.object( {
+	accessToken: z.string(),
+	expiresIn: z.number(),
+	expirationTime: z.number(),
+	id: z.number(),
+	email: z.string(),
+	displayName: z.string().default( '' ),
+} );
+
+export type StoredToken = z.infer< typeof authTokenSchema >;
```
fredrikekelund:
```suggestion
export type StoredAuthToken = z.infer< typeof authTokenSchema >;
```
I'd include "auth" in the name somehow

apps/cli/lib/certificate-manager.ts
fredrikekelund, line 1: Ouff… I didn't consider the HTTPS certificates. We should put these in `~/.studio`, too, IMO. Let's do that in a separate PR, though.

apps/cli/lib/certificate-manager.ts
```diff
@@ -66,6 +68,22 @@ function createNameConstraintsExtension( domains: string[] ) {
...
+	if ( process.platform === 'win32' ) {
+		if ( ! process.env.APPDATA ) {
+			throw new LoggerError( __( 'Studio config file path not found.' ) );
+		}
+
+		return path.join( process.env.APPDATA, 'Studio' );
+	}
+
+	return path.join( os.homedir(), 'Library', 'Application Support', 'Studio' );
+}
```
fredrikekelund: Could we use `getAppdataDirectory` from `apps/cli/lib/server-files.ts` here?

apps/studio/src/lib/shared-config-watcher.ts
fredrikekelund, line 1: Could we use `_events` for listening to auth changes, too, instead of watching the shared config file?

fredrikekelund (commented):
Left some comments. Shared config error handling and using `_events` instead of a shared config file watcher are the big things we need to address, IMO. 

Haven't tested yet but will do that now 👍

apps/studio/src/components/auth-provider.tsx
```diff
@@ -79,6 +79,15 @@ const AuthProvider: React.FC< AuthProviderProps > = ( { children } ) => {
...
 		const { token } = payload;
+
+		if ( ! token ) {
+			setIsAuthenticated( false );
+			setClient( undefined );
+			setWpcomClient( undefined );
+			setUser( undefined );
+			return;
+		}
+
```
fredrikekelund, line 90: The `auth logout` command didn't trigger a change in authentication status in the app for me when I tested it the first time. Probably because the shared config file didn't exist when I started Studio. I guess this would be solved by moving to an `_events` architecture for auth events instead.

apps/cli/ai/providers.ts
fredrikekelund, line 1: Let's update this file to import `aiProviderSchema` from `apps/cli/lib/cli-config/core.ts`. We should have a single definition for that

apps/cli/ai/providers.ts
```diff
@@ -156,12 +156,15 @@ const AI_PROVIDER_DEFINITIONS: Record< AiProviderId, AiProviderDefinition > = {
 		id: 'anthropic-api-key',
 		autoFallbackWhenUnavailable: false,
 		isVisible: async () => true,
-		isReady: async () => Boolean( await getAnthropicApiKey() ),
+		isReady: async () => {
+			const { anthropicApiKey } = await readCliConfig();
+			return Boolean( anthropicApiKey );
+		},
```
fredrikekelund, line 160: We need to remember to ping the Architecture team before landing this to trunk, to let them know that they need to manually migrate AI-related props from the appdata file to the CLI config file.

apps/cli/lib/certificate-manager.ts
fredrikekelund: I created STU-1416 to track this

fredrikekelund (approved):
This works as expected in testing 👍 Error handling for the shared config is really key to get right. I'll still approve now to keep up momentum, since we aren't landing straight to trunk

apps/cli/commands/preview/tests/create.test.ts
```diff
@@ -19,10 +19,9 @@ const mockReportWarning = vi.fn();
...
 
 vi.mock( '@studio/common/lib/get-wordpress-version' );
-vi.mock( 'cli/lib/appdata', async () => ( {
-	...( await vi.importActual( 'cli/lib/appdata' ) ),
-	getAppdataDirectory: vi.fn().mockReturnValue( '/test/appdata' ),
-	getAuthToken: vi.fn(),
+vi.mock( '@studio/common/lib/shared-config', async ( importOriginal ) => ( {
+	...( await importOriginal< typeof import('@studio/common/lib/shared-config') >() ),
+	readAuthToken: vi.fn(),
 } ) );
```
fredrikekelund:
```suggestion
vi.mock( import( '@studio/common/lib/shared-config' ), async ( importOriginal ) => ( {
	...( await importOriginal() ),
	readAuthToken: vi.fn(),
} ) );
```
Tip: I stumbled upon this recently, that the first argument to `vi.mock` can be an `import()` call, which makes Vitest aware of the module types.

apps/cli/ai/providers.ts
```diff
@@ -13,7 +17,7 @@ export const AI_PROVIDERS = {
 
 export type AiProviderId = keyof typeof AI_PROVIDERS;
 
-export const aiProviderSchema = z.enum( [ 'wpcom', 'anthropic-claude', 'anthropic-api-key' ] );
+export { aiProviderSchema };
```
fredrikekelund: Do we really need to re-export this definition?

tools/common/lib/shared-config.ts
```diff
@@ -0,0 +1,136 @@
...
+	id: z.number(),
+	email: z.string(),
+	displayName: z.string().default( '' ),
+} );
+
+export type StoredAuthToken = z.infer< typeof authTokenSchema >;
+
+const sharedConfigSchema = z
+	.object( {
+		version: z.number().default( SHARED_CONFIG_VERSION ),
```
fredrikekelund:
```suggestion
		version: z.literal( SHARED_CONFIG_VERSION ),
```
I should have called this out in `apps/cli/lib/cli-config/core.ts`, too, but this is an important distinction: I think the zod schema parsing should fail unless the `version` field matches what we expect. 

If we don't do this, then the schema parsing might fail for other reasons, but it's not as easy for us to see why it's happening.

We can't just change this line to `z.literal` instead, we also need a base schema containing just the version field to try with. I'm basically imagining the logic we have in `apps/cli/lib/cli-config/core.ts`, just with `z.literal` instead of `z.number().default()`

PR #2822, opened 2026-03-16
Make DB_NAME constants removal logic language-agnostic

sejas:
## Related issues

- Fixes STU-1308

## How AI was used in this PR

Claude assisted with the regex implementation and test cases. All code was reviewed.

## Proposed Changes

- Replace exact string matching with a regex-based approach for removing default DB constants from wp-config.php
- The regex is strict on PHP `define()` calls (only matches exact default values from wp-config-sample.php) but flexible with comments, so translated wp-config.php files are handled correctly
- Add test cases for Polish, French, and comment-free wp-config.php files

## Testing Instructions

- Run `npm start`
- Create a new site
- Confirm wp-config.php doesn't have DB_* constants
- In the same site, copy the content from wp-config-sample.php and paste it to wp-config.php, modify the any comment.
- Go to the export tab and export the entire site.
- Confirm the DB_* constants are removed.
- In the same site, copy the content from wp-config-sample.php and paste it to wp-config.php, modify the any value.
- Go to the export tab and export the entire site.
- Confirm the all the DB_* constants are kept

- Run `npm test -- tools/common/lib/tests/remove-default-db-constants.test.ts`
- Verify all tests pass, including the new ones for translated comments

## Pre-merge Checklist


katinthehatsite (approved):
The changes look good to me. I tested the steps and did not find regressions. I also looked through the code and did not have any major concerns 👍

PR #2832, opened 2026-03-17
Refactor option picker to use SelectList and add site-spec skill

youknowriad:
## Related issues

- Related to improving the question UI and site creation flow

## How AI was used in this PR

AI assisted with refactoring the option picker component to use pi-tui's SelectList, implementing the site-spec discovery skill with structured questions, and simplifying the state management. All code was reviewed and tested.

## Proposed Changes

- **Option Picker Refactor**: Replaced ad-hoc rendering with pi-tui's SelectList component for proper alignment, scrolling, and native keyboard navigation
- **Site Spec Skill**: Added structured discovery questions (name, purpose, audience, tone, colors, layout, pages) to guide site creation
- **Inline Input for "Other"**: When users navigate to or start typing on the "Other" option, an inline Input component appears for immediate freeform entry
- **UI Polish**: Added left padding (paddingX: 1) to align questions and options with main content; simplified state management by removing redundant optionPickerVisible field
- **Question Tool Integration**: Distinguished between multiple-choice questions (using AskUserQuestion) and open-ended questions (plain text prompt)

## Testing Instructions

- Run `/site-spec` skill to see the new question flow
- Test navigating through options with arrow keys and entering free-form text on "Other"
- Verify alignment of questions and option picker with the rest of the console output
- Test selecting different options and confirming selections

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/plugin/skills/site-spec/SKILL.md
+++ b/apps/cli/ai/plugin/skills/site-spec/SKILL.md
@@ -10,23 +10,27 @@ Before creating a new WordPress site, gather the user's preferences through a sh
 
 ## How to Run
 
-Use AskUserQuestion to ask focused questions in 2-3 rounds, building on previous answers. Keep it concise — no more than 5-6 questions total.
+Gather preferences through 3-4 rounds, building on previous answers. Keep it concise — no more than 6-7 questions total.
 
 **AskUserQuestion constraints**: Each call supports 1-4 questions, each with 2-4 options. An "Other" free-form option is automatically provided by the system — do NOT add one yourself. Keep option labels short (1-5 words). Only use AskUserQuestion for questions that have meaningful predefined options. For open-ended questions (like asking for a name), just ask in your text output — the user will type their answer in the prompt.
 
-### Round 1 — Name & Purpose
+### Round 1 — Name
 
-First, ask the user for their business/site name in your text output (not via AskUserQuestion — it's open-ended with no predefined options). Then use AskUserQuestion for:
+Ask the user for their business/site name in your text output. **Stop here and wait for their reply** — do NOT call any tools or continue to the next round. The user needs a chance to type their answer in the prompt.
+
+### Round 2 — Purpose & Audience
+
+After the user provides the name, use AskUserQuestion for:
 - What is this site for? (e.g., business type, portfolio, blog, community, e-commerce, agency, restaurant, etc.)
 - Who is the target audience? (e.g., professionals, consumers, creatives, developers, local community, etc.)
 
-### Round 2 — Brand & Aesthetic Direction
+### Round 3 — Brand & Aesthetic Direction
 
-Adapt based on Round 1 answers. Ask about:
+Adapt based on previous answers. Ask about:
 - What tone or personality should the site convey? (e.g., professional, playful, minimalist, bold, luxurious, raw, editorial, retro, organic, etc.)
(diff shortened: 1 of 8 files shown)


apps/cli/ai/plugin/skills/site-spec/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+## After Confirmation
+
+1. Call `site_create` with an appropriate name derived from the spec.
+2. Save the site spec to `{site_path}/.agents/site-spec.md` using the Write tool, so it persists for future sessions.
+3. Use the spec to guide ALL subsequent design decisions — theme, typography, colors, layout, content tone, and page structure.
+
+## When to Skip Discovery
+
+Do NOT ask questions if:
+- The user already provided an asnwer for that specific question in the initial prompt. (e.g., "build me a dark minimalist portfolio for a photographer with these 5 pages using navy and gold colors"). Instead, synthesize the spec directly from their request and confirm.
```
taipeicoder:
```suggestion
- The user already provided an answer for that specific question in the initial prompt. (e.g., "build me a dark minimalist portfolio for a photographer with these 5 pages using navy and gold colors"). Instead, synthesize the spec directly from their request and confirm.
```

apps/cli/ai/ui.ts
```diff
@@ -1711,21 +1771,48 @@ export class AiChatUI {
...
 				this.hideEditor();
-				this.optionPickerItems = q.options;
-				this.optionPickerSelectedIndex = 0;
-				this.optionPickerVisible = true;
+				const selectItems: SelectItem[] = q.options.map( ( opt, i ) => ( {
+					value: opt.label,
+					label: `${ i + 1 }. ${ opt.label }`,
+					description: opt.description,
+				} ) );
+				selectItems.push( {
```
taipeicoder: Should this be applied to every question that has options? The concern would be that some callers might expect the value to match one of the list instead of a free-form response. How these callers handle the free-form response is also unpredictable, some no-ops, and other have a fall-through option.

apps/cli/ai/ui.ts
```diff
@@ -604,29 +616,50 @@ export class AiChatUI {
...
+				if (
+					! matchesKey( data, 'up' ) &&
+					! matchesKey( data, 'down' ) &&
+					! matchesKey( data, 'enter' ) &&
+					! matchesKey( data, 'escape' ) &&
+					data.length === 1 &&
+					data >= ' '
+				) {
+					// setSelectedIndex clamps, so a large number lands on the last item
+					this.optionPickerSelectList.setSelectedIndex( 999 );
```
taipeicoder: Instead of 999, how about:

```
  const otherIndex = selectItems.findIndex( item => item.value === AiChatUI.OTHER_VALUE );
  if ( otherIndex >= 0 ) {
      this.optionPickerSelectList.setSelectedIndex( otherIndex );
  }
``` 

Or simply (if the Others option is always available_
```
this.optionPickerSelectList.setSelectedIndex( selectItems.length - 1 );
```

taipeicoder (approved):
The skill works great. Left some non-blocking feedback, and long texts could be handled better.

Image

apps/cli/ai/ui.ts
youknowriad: Agreed, this feels like a bug, we should not do this to all questions, only the site spec ones.

Issue #568, opened 2024-09-27
Add Database Button


davidperezgar:
### What

As a developer, we need a button for accessing easily to the database. I know that is SQLite, but a direct access and open with a default client DB in the dashboard.

### Why

We use a lot the Database manager.

### How

![Captura de pantalla 2024-09-27 a las 9 10 09](https://github.com/user-attachments/assets/c4e1010d-c6c4-42ba-b4c5-777c48a2ac83)

wojtekn:
Thanks for adding the feature request @davidperezgar .

> We use a lot the Database manager.

What app are you using to manage the database?

davidperezgar:
Thanks. Good question. Normally we use Sequel, TablePlus or Adminer.

talldan:
https://github.com/Automattic/studio/issues/63 is related. That allows selecting editor/terminal. It won't be scalable to add every kind of application, so at some point there might need to be a way to add custom paths or an 'Open in ...' button.

mikaoelitiana:
This seems to be a duplicate of #420

wojtekn:
@mikaoelitiana yes, and soon we will solve it by https://github.com/Automattic/studio/pull/2834 🥳

wojtekn:
We've added phpMyAdmin support in Studio 1.7.7.

I'm closing this issue.

PR #2788, opened 2026-03-12
Move AI instruction files to a dedicated folder structure

sejas:
## Related issues

- Fixes STU-1397 and STU-1396

## Proposed Changes

Create a folder under the `skills` root directory containing the static `AGENTS.md`, `CLAUDE.md`, `STUDIO.md`, and custom Studio skills folders.
It unifies the logic to bundle these files inside the binary and then copy them on site creation
Having the skills folder it opens the door to install them using `npx skills` command. E.g `npx skills add https://github.com/Automattic/studio --skill studio-cli`

It also adds getWpFilesPath function which points to the correct `wp-files` root folder. That was broken becuase Studio was moved to a subdirectory when we created the mono repo.

```mermaid
graph TD
      subgraph "Source (repo)"
          S1["skills/<br/>├── AGENTS.md<br/>├── CLAUDE.md<br/>├── STUDIO.md<br/>└──
  studio-cli/SKILL.md"]
          S2["GitHub: WordPress/agent-skills<br/>├── wp-plugin-development/<br/>├──
  wp-block-development/<br/>├── wp-block-themes/<br/>├── wp-rest-api/<br/>└──
  wp-wpcli-and-ops/"]
      end

      subgraph "Build Time (npm install)"
          WP["wp-files/skills/<br/>├── AGENTS.md<br/>├── CLAUDE.md<br/>├── STUDIO.md<br/>├──
  studio-cli/<br/>├── wp-plugin-development/<br/>├── wp-block-development/<br/>├──
  wp-block-themes/<br/>├── wp-rest-api/<br/>└── wp-wpcli-and-ops/"]
      end

      subgraph "App Startup"
          SF["server-files/skills/<br/>(app data directory)"]
      end

      subgraph "Site Creation"
          ROOT["~/Studio/my-site/<br/>├── AGENTS.md<br/>├── CLAUDE.md<br/>└── STUDIO.md"]
          AGENTS["~/Studio/my-site/.agents/skills/<br/>├── studio-cli/<br/>├──
  wp-plugin-development/<br/>└── ..."]
          CLAUDE["~/Studio/my-site/.claude/skills/<br/>├── studio-cli →
  ../../.agents/skills/studio-cli<br/>└── ... (symlinks)"]
      end

      S1 -->|"copyLocalContent()"| WP
      S2 -->|"downloadRemoteSkills()"| WP
      WP -->|"copyBundledAiInstructions()<br/>setup-wp-server-files.ts"| SF
[...]

PR #2834, opened 2026-03-17
Add phpMyAdmin support

wojtekn:
## Related issues

- Related to STU-31
- Solves https://github.com/Automattic/studio/issues/568 and https://github.com/Automattic/studio/issues/420

## How AI was used in this PR

I planned the feature architecture - UI placement, CLI changes, and offline handling. Implementation was carried out by Claude (Sonnet 4.6) following that plan step by step.

## Proposed Changes

Adds per-site phpMyAdmin support using the `--phpmyadmin` flag already available in `@wp-playground/cli`:

- phpMyAdmin bundled with Studio, so it works in offline mode
- phpMyAdmin enabled for all existing and new sites
- phpMyAdmin button always visible in the "Open in…" grid. Disabled when server is loading
- ~~patches Playground issues, see below~~
- e2e tests that checks if button is displayed and if phpMyAdmin is accessible via browser

CleanShot 2026-03-17 at 17 19 03

### Known Issues

~~The PR includes two patches for Playground packages.~~

~~The first `patch-package` patch in this PR removes the `--skip-sqlite-setup` guard that previously prevented even getting this far - that guard was overly broad and blocked Studio's use case entirely. This also should be fixed upstream.~~ (Fixed in https://github.com/Automattic/studio/pull/2872)

~~The second one fixes the path to the SQLite plugin hardcoded in the Playground CLI shim. The Playground CLI bundles phpMyAdmin and injects a SQLite-to-MySQL compatibility shim, but the path it hardcodes (`/internal/shared/sqlite-database-integration/wp-pdo-mysql-on-sqlite.php`) doesn't match where Studio mounts the SQLite integration. This is a bug in `@wp-playground/cli` that needs to be fixed upstream before this feature can ship.~~ (Fixed in b38b9b4818522e610b7fb819ac9708aae97dff46)

## Testing Instructions

2. Start Studio
3. Start a site
4. The Overview tab will show a phpMyAdmin button in the "Open in…" grid
5. Clicking it opens the phpMyAdmin URL

## Pre-merge Checklist


apps/studio/src/storage/paths.ts
```diff
@@ -63,6 +63,19 @@ export function getResourcesPath(): string {
 	return path.join( exePath, 'resources' );
 }
 
+export function getWpFilesPath(): string {
```
wojtekn, line 62: Here I'm fixing an issue where `wp-files` aren't copied to `server-files` when the app is running in devmode.

It seems as issue introduced during monorepo change - `getAppPath` now points to `app/studio` and not to project root.

scripts/download-wp-server-files.ts
```diff
@@ -90,6 +101,39 @@ async function downloadFile( file: FileToDownload ): Promise< void > {
 			}
 			fs.renameSync( sourcePath, targetPath );
 		}
+	} else if ( name === 'phpmyadmin' ) {
```
wojtekn, line 104: This mimics logic from Playground: https://github.com/WordPress/wordpress-playground/blob/4f8b0717af51be37bd859a7c523e9433b46f4a37/packages/playground/tools/src/phpmyadmin/index.ts#L18

wojtekn:
@Automattic/yolo and @shaunandrews, I would appreciate a review.

Regarding the UI, I think the button location in the current layout is solid, as it works well with the other "Open in" buttons. I'm not convinced about the enable/disable part - after implementing full offline mode support, I realized there is really no need to have the ability to enable and disable phpMyAdmin. I would just remove it from site settings entirely, so it will be available by default for all sites. Any thoughts?

I still need to coordinate the Playground changes to remove patches.

tools/common/constants.ts
```diff
@@ -32,3 +32,6 @@ export const MINIMUM_WORDPRESS_VERSION = '6.2.1' as const; // https://wordpress.
 export const DEFAULT_WORDPRESS_VERSION = 'latest' as const;
 export const DEFAULT_PHP_VERSION: typeof RecommendedPHPVersion = RecommendedPHPVersion;
 export const SQLITE_FILENAME = 'sqlite-database-integration' as const;
+
+// phpMyAdmin - set to true to enable by default for all new sites
+export const DEFAULT_ENABLE_PHPMYADMIN = true;
```
sejas: If it affects to only new sites, then I would suggest renaming the constant name to reflect that.
But I think it would be even better to enable phpmyadmin to all sites no matter if they are new or existing. That would avoid a lot of confusion for users.
If they want, they could disable it in their sites.

apps/cli/patches/@wp-playground+cli+3.1.12.patch
```diff
@@ -0,0 +1,29 @@
...
++++ b/node_modules/@wp-playground/cli/run-cli-2YzKNrNz.js
+@@ -1259,10 +1259,7 @@ async function Lt(e) {
+     w.setSeverityFilterLevel(c);
+   }
+   if (e.intl || (e.intl = !0), e.redis === void 0 && (e.redis = await ie()), e.memcached === void 0 && (e.memcached = await ie()), e.phpmyadmin) {
+-    if (e.phpmyadmin === !0 && (e.phpmyadmin = "/phpmyadmin"), e.skipSqliteSetup)
+-      throw new Error(
+-        "--phpmyadmin requires SQLite. Cannot be used with --skip-sqlite-setup."
+-      );
++    if (e.phpmyadmin === !0 && (e.phpmyadmin = "/phpmyadmin"));
```
sejas: It would be great to avoid not having these patches, but I think it's better to move forward until we adapt those changes in Playground repository.

sejas (commented):
Yay! this is a great feature that users suggested in the past. Thanks for bringing it to WordPress Studio. I tested in new sites and existing sites and it works pretty good. I only found an issue for existing sites and the default checkbox value.

https://github.com/user-attachments/assets/8022ed20-6618-47ab-a1e2-653333a0822e

Currently the button is disabled for existing sites, and the setting is a bit hidden under the debugging site settings modal. What if we enable phpmyadmin for all sites?
Currently there is an issue where the button is disabled for existing sites, but the checkbox in site settings is enabled, see the recording below:

https://github.com/user-attachments/assets/34d8b8b1-e2fe-4cea-91c5-e9c8513b6227

tools/common/constants.ts
wojtekn: Yep, see my comment under the PR - I'm wondering if we should remove this field completely and make it always available for all sites.

wojtekn:
> Currently there is an issue where the button is disabled for existing sites, but the checkbox in site settings is enabled, see the recording below:

Thanks, I will check this one.

wojtekn:
> Currently there is an issue where the button is disabled for existing sites, but the checkbox in site settings is enabled, see the recording below:

It's fixed now.

shaunandrews:
Maybe we can update the icons for all the "Open in..." options? Here's an example:

image

I also changed the "Customize" heading to "WordPress".

wojtekn:
@shaunandrews we can, but I would tackle that separately and would focus on adding phpMyAdmin here.

My first thought is that with those colorful icons, it looks chaotic, and we would need to maintain icons for all the apps: 8 editors, 4 terminals, and more. All of them change with system updates or app updates, so it may add a maintenance burden.

The main thing I would like to address under this PR is the need to enable/disable phpMyAdmin - I'm heading towards removing that logic completely and making it always enabled for all sites. Thoughts on this?

apps/studio/src/storage/paths.ts
wojtekn: This was fixed and merged through https://github.com/Automattic/studio/pull/2788, so it's gone from this PR now.

shaunandrews:
> My first thought is that with those colorful icons, it looks chaotic

The icons add personality and can help with recognition. The current usage of the limited WordPress icons just add clutter.

> we would need to maintain icons for all the apps: 8 editors, 4 terminals, and more. All of them change with system updates or app updates, so it may add a maintenance burden.

Updating some icons every now and again hardly seems like a burden. Maybe an annoyance. I think its worth the hassle for the recognition alone.

> making it always enabled for all sites. Thoughts on this?

If having it enabled causes no performance issues, then I guess it makes senes to always have it on.

wojtekn:
> The icons add personality and can help with recognition. The current usage of the limited WordPress icons just add clutter.
> [...] 
> Updating some icons every now and again hardly seems like a burden. Maybe an annoyance. I think its worth the hassle for the recognition alone.

Let's open a separate PR for that. If we follow that approach, we will need to update those across all apps and check whether the icons differ between Mac and Windows. I'm also unsure if and how we could handle cases in which Studio has an icon for version X, but the user uses version Y - probably we can't.

> If having it enabled causes no performance issues, then I guess it makes senes to always have it on.

Thanks. Yes, it doesn't seem to introduce any performance issues. I will remove it.

tools/common/constants.ts
sejas: I like that approach and reduces cognitive load for the user too.

apps/cli/patches/@wp-playground+cli+3.1.12.patch
wojtekn: My [PR](https://github.com/WordPress/wordpress-playground/pull/3413) is already merged, so we can remove the patch as soon as we update Playground in Studio.

bcotrim (commented):
LGTM 👍 

Have you looked into the performance impact of having phpMyAdmin enabled by default? On my testing I tried running a site with phpMyAdmin enabled and disabled and noticed a significant difference:

```
❯ ps -o pid,rss,%cpu,%mem -p 97050,94775
  PID    RSS  %CPU %MEM
94775 768480   0.0  2.0
97050 1519760   0.0  4.0
```

wojtekn:
@bcotrim, what difference did you see? Performance Metrics don't show anything.

When phpMyAdmin is enabled, Studio just mounts an additional directory, so it shouldn't affect the running site much.

bcotrim:
> @bcotrim, what difference did you see? Performance Metrics don't show anything.
> 
> When phpMyAdmin is enabled, Studio just mounts an additional directory, so it shouldn't affect the running site much.

Running with phpMyAdmin is taking up more system memory on my tests.

sejas (approved):
Nice! PhpMyAdmin now is available in all sites. Congrats for the simplification.
I didn't observe any issues when testing it on my Mac. 
I observed the RAM for node processes going a bit higher than on trunk (2GB vs 1.8GB), but then they fall to normal again. ~1.2GB

Congrats for creating this PR and upstreaming the fix to the Playground repository.

https://github.com/user-attachments/assets/89502633-6438-411e-be7e-7a1ce6c652c6

wojtekn:
@bcotrim I compared starting site in the trunk and this branch, and I couldn't see a difference in memory usage between those two.

bcotrim (approved):
It seems the difference I mentioned earlier was related to running time and not phpMyAdmin. After booting up memory usage slowly goes down until the stabilizes, that's why I saw different values.

I ran the test again and let the sites run for a while, and both `trunk` and this branch seem to have similar memory usage after a while.

Reviewed the code again and LGTM and works as expected 👍

scripts/download-wp-server-files.ts
```diff
@@ -90,6 +101,47 @@ async function downloadFile( file: FileToDownload ): Promise< void > {
...
+		for ( const step of installSteps ) {
+			if ( step.step === 'writeFile' && typeof step.data === 'string' ) {
+				// step.path is like /tools/phpmyadmin/config.inc.php — strip the /tools/phpmyadmin prefix
+				const relativePath = step.path.replace( '/tools/phpmyadmin/', '' );
+				const destFile = path.join( extractedPath, relativePath );
+				await fs.ensureDir( path.dirname( destFile ) );
+
+				// The upstream DbiMysqli.php hardcodes a Playground-internal path for the
+				// SQLite driver. Replace it with the Studio path.
+				const fileData = step.data.replace(
```
wojtekn, line 137: I used this search-and-replace instead of a patch. We can remove it as soon as it's fixed on SQLite driver and Playground side - work in progress is under https://github.com/WordPress/wordpress-playground/pull/3420.

tools/common/lib/sqlite-integration.ts
```diff
@@ -49,15 +49,10 @@ export abstract class SqliteIntegrationProvider {
...
-		const updatedContent = dbCopyContent
-			.replace(
-				"'{SQLITE_IMPLEMENTATION_FOLDER_PATH}'",
-				`realpath( __DIR__ . '/mu-plugins/${ sqliteDirname }' )`
-			)
-			.replace(
-				"define( 'SQLITE_DB_DROPIN_VERSION'",
-				`define( 'SQLITE_MAIN_FILE', __FILE__ );\ndefine( 'SQLITE_DB_DROPIN_VERSION'`
-			);
+		const updatedContent = dbCopyContent.replace(
```
wojtekn, line 52: @bcotrim @katinthehatsite I'm removing this code here, and fixed the warning you experienced in the upstream: https://github.com/WordPress/sqlite-database-integration/pull/332

PR #2836, opened 2026-03-17
Migrate appdata to ~/.studio/appdata.json with versioned migrations

bcotrim:
## Related issues

Related to STU-1350

## How AI was used in this PR

All code was generated via Claude Code (Opus + Haiku). Changes were reviewed and validated through iterative conversation with the user.

## Proposed Changes

- **CLI `_migrate` command** — Hidden command that copies `appdata-v1.json` from platform-specific Electron path to `~/.studio/appdata.json`
- **Desktop bootstrap** — Triggers CLI migration on boot before other data operations
- **Updated primary paths** — Desktop and CLI both read from `~/.studio/appdata.json` with CLI fallback to old platform-specific location
- **Generic migration framework** — New `ConfigMigrator` in `tools/common/lib/` that applies versioned migrations based on file version field
- **Wired into config readers** — Appdata, CLI config, and shared config all use the migration runner for future extensibility

## Testing Instructions

1. `npm run typecheck` — Verify no type errors
2. `npm test` — All 1314 tests pass (7 new tests for config migrator)
3. Manual: Launch Desktop with existing `appdata-v1.json` → Verify `~/.studio/appdata.json` is created and Desktop reads from it
4. Manual: Add/modify a site → Verify `~/.studio/appdata.json` is updated
5. Manual: CLI `studio site list` → Verify reads from `~/.studio/appdata.json`
6. Manual: Restart Desktop → Verify migration is idempotent (no errors)

## Pre-merge Checklist


apps/cli/index.ts
```diff
@@ -34,6 +34,15 @@ async function main() {
...
 		} )
+		.middleware( async () => {
+			try {
+				const { migrateAppdata } = await import( 'cli/commands/_migrate' );
+				await migrateAppdata();
+			} catch ( error ) {
+				// Migration failure should not block CLI usage
+				console.error( 'Appdata migration failed:', error );
+			}
+		} )
```
fredrikekelund, line 41: The Studio app should be responsible for migrating the data (or at least pushing the trigger). If the standalone CLI does it automatically, then we cannot guarantee that the installed version of Studio can handle the current config setup. That's because theoretically, someone can have Studio 1.7.6 installed and install and run the standalone CLI. This code will make the standalone CLI run the migration on every run. 

I think this logic should be a check rather than a migration. It should run for every command except `_migrate`. If the Studio app is installed and `~/.studio/shared.json` is missing, it should return an error message prompting the user to update Studio. If Studio isn't installed, just proceed. If Studio is installed and the shared config exists, just proceed.

apps/studio/src/storage/paths.ts
```diff
@@ -19,11 +19,19 @@ export function getUserDataFilePath(): string {
 	if ( process.env.DEV_APP_DATA_PATH ) {
 		return process.env.DEV_APP_DATA_PATH;
 	}
-	return path.join( getAppDataPath(), getAppName(), 'appdata-v1.json' );
+	return path.join( getStudioHome(), 'appdata.json' );
```
fredrikekelund: Do we want to keep calling it "appdata" or maybe just "app"? I don't really have a preference, but since we're calling the CLI config file "cli.json", I guess there's an argument for "app.json" being a good fit…

apps/studio/src/storage/storage-types.ts
```diff
@@ -28,7 +28,7 @@ export interface UserData {
 }
 
 export interface PersistedUserData extends Omit< UserData, 'sites' > {
```
fredrikekelund: I commented about this on Slack, too, but `appdata.json` no longer contains any sites.

apps/studio/src/migrations/migrate-appdata-via-cli.ts
```diff
@@ -0,0 +1,217 @@
...
+		if ( ! excludedSiteFields.has( key ) ) {
+			result[ key ] = site[ key ];
+		}
+	}
+	// Always keep id so we can correlate with cli.json sites
+	if ( site.id ) {
+		result.id = site.id;
+	}
+	return result;
+}
```
fredrikekelund: I don't understand what this function is for

apps/studio/src/migrations/migrate-appdata-via-cli.ts
```diff
@@ -0,0 +1,217 @@
...
+		const sitesWithDesktopFields = oldData.sites
+			.map( ( site: Record< string, unknown > ) => pickAppdataSiteFields( site ) )
+			.filter( ( site: Record< string, unknown > ) =>
+				Object.keys( site ).some( ( key ) => key !== 'id' && site[ key ] !== undefined )
+			);
+
+		if ( sitesWithDesktopFields.length > 0 ) {
+			config.sites = sitesWithDesktopFields;
+		}
+	}
```
fredrikekelund, line 159: The appdata file shouldn't contain any sites, though, right?

apps/studio/src/migrations/02-migrate-to-split-config.ts
```diff
@@ -0,0 +1,217 @@
...
+	}
+	return result;
+}
+
+/**
+ * Fields from the old appdata that move to shared.json.
+ */
+function buildSharedConfig( oldData: Record< string, unknown > ): Record< string, unknown > {
+	return { version: 1, ...pick( oldData, sharedConfigFields ) };
+}
```
fredrikekelund, line 57:
```suggestion
function buildSharedConfig( oldData: Record< string, unknown > ): Record< string, unknown > {
	const schema = z.object( { ...sharedConfigSchema.omit( { version: true } ).shape } );
	const parsed = schema.parse( oldData );

	return { version: 1, ...parsed };
}
```
I think it'd be a good idea to lean into zod parsing here, too. If `oldData` doesn't pass validation, then it would cause an error the next time the CLI or the app tries to read that file. 

I had to play around with the schema definition a little, but this suggestion works. We want to get rid of the `.loose()` call from the original definition – that's why we jump through the extra hoops

apps/studio/src/migrations/02-migrate-to-split-config.ts
```diff
@@ -0,0 +1,217 @@
...
+		version: 1,
+		sites: [],
+		snapshots: [],
+	};
+
+	if ( Array.isArray( oldData.sites ) ) {
+		config.sites = oldData.sites.map( ( site: Record< string, unknown > ) =>
+			pick( site, cliSiteFields )
+		);
+	}
```
fredrikekelund, line 78:
```suggestion
	const cliSiteSchema = siteDetailsSchema.extend( {
		url: z.string().optional(),
		latestCliPid: z.number().optional(),
	} );

	if ( Array.isArray( oldData.sites ) ) {
		config.sites = oldData.sites.map( ( site: Record< string, unknown > ) =>
			cliSiteSchema.parse( site )
		);
	}
```
Same thing here about leaning into zod parsing, because that's the schema we expect the file to conform to later, anyway.

apps/studio/src/migrations/migrate-appdata-via-cli.ts
```diff
@@ -0,0 +1,217 @@
...
+ */
+const movedTopLevelFields = new Set( [ ...sharedConfigFields, 'sites', 'snapshots', 'version' ] );
+
+// Per-site fields that stay in appdata.json: anything NOT in the CLI site schema,
+// NOT in the snapshot schema, and NOT runtime state.
+const excludedSiteFields = new Set( [
+	...cliSiteFields,
+	...Object.keys( snapshotSchema.shape ),
+	'running',
+] );
```
fredrikekelund: I don't believe `running` was ever stored anywhere – it comes from the runtime type. As for `themeDetails`, we need to figure something out…

We might consider adding an array or a record of `themeDetails` entries to `appdata.json`. WDYT? It would take some plumbing to get right and to ensure we clean it up when a site is deleted, but it shouldn't be overly complex.

fredrikekelund (commented):
The basics of the migration works, but we need to address the fact that we still expect the appdata file to contain a sites array in some places.

bcotrim:
@fredrikekelund thanks for the review!
We still need to store some `sites` information in the `app.json` file (`themeDetails` and `sortOrder`)
I renamed the field to `siteMetadata` to make the intention clearer. Let me know how it looks.

bcotrim:
@fredrikekelund E2E tests should be fixed now. I intentionally left the `appdata-v1` setup logic in E2E tests so the migration is tested on all tests. We can later decide to keep it or update the E2E to setup the new config files instead.

apps/studio/src/migrations/02-migrate-to-split-config.ts
```diff
@@ -0,0 +1,209 @@
...
+						extra: { snapshotUrl: snapshot.url, context: 'migrate-to-split-config' },
+					} );
+				}
+				return acc;
+			},
+			[]
+		);
+	}
+
+	return config;
```
fredrikekelund, line 105:
```suggestion
	if ( typeof oldData.aiProvider === 'string' ) {
		config.aiProvider = oldData.aiProvider;
	}

	if ( typeof oldData.anthropicApiKey === 'string' ) {
		config.anthropicApiKey = oldData.anthropicApiKey;
	}
	
	return config;
```
Might as well bring the AI props, too. Doesn't really incur any risk or overhead.

apps/studio/src/migrations/00-migrate-to-split-config.ts
```diff
@@ -0,0 +1,209 @@
...
+function ensureDirectory( filePath: string ): void {
+	const dir = path.dirname( filePath );
+	if ( ! fs.existsSync( dir ) ) {
+		fs.mkdirSync( dir, { recursive: true } );
+	}
+}
+
+// Zod schemas used to extract and validate fields for each config file.
+// Using .parse() ensures only valid, expected fields are written.
+
```
fredrikekelund:
```suggestion
```
Nit. This comment, while the contents are pretty accurate, just feels a little out of place. I'd remove it.

apps/studio/src/migrations/00-migrate-to-split-config.ts
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return path.join( os.homedir(), 'Library', 'Application Support', 'Studio', 'appdata-v1.json' );
+}
+
+function ensureDirectory( filePath: string ): void {
+	const dir = path.dirname( filePath );
+	if ( ! fs.existsSync( dir ) ) {
+		fs.mkdirSync( dir, { recursive: true } );
+	}
+}
```
fredrikekelund: Nit. Consider inlining in `writeJsonFile`

apps/studio/src/migrations/index.ts
```diff
@@ -0,0 +1,10 @@
+import { migrateAppConfig } from './00-migrate-to-split-config';
+import { migrateFromWpNowFolder } from './01-migrate-from-wp-now-folder';
+import { renameLaunchUniquesStat } from './02-rename-launch-uniques-stat';
```
fredrikekelund:
```suggestion
import { migrateAppConfig } from './02-migrate-to-split-config';
import { migrateFromWpNowFolder } from './00-migrate-from-wp-now-folder';
import { renameLaunchUniquesStat } from './01-rename-launch-uniques-stat';
```
I think it'd make sense to rename these files to reflect the order in which they were introduced.

apps/studio/src/storage/storage-types.ts
```diff
@@ -11,8 +11,13 @@ export interface WindowBounds {
 	isFullScreen?: boolean;
 }
 
+export interface AppdataSiteData {
+	themeDetails?: SiteDetails[ 'themeDetails' ];
+	sortOrder?: number;
+}
+
 export interface UserData {
```
fredrikekelund, line 19:
```suggestion
export interface UserData {
	version: 1;
```

apps/studio/src/storage/storage-types.ts
```diff
@@ -28,13 +33,8 @@ export interface UserData {
...
-export interface PersistedUserData extends Omit< UserData, 'sites' > {
-	version: 1;
-
-	// Users can edit the file system manually which would make UserData['name'] and UserData['path']
-	// get out of sync. `name` is redundant because it can be calculated from `path`, so we
-	// won't persist `name`.
-	sites: Omit< StoppedSiteDetails, 'running' >[];
+export interface PersistedUserData extends UserData {
+	version: number;
 }
```
fredrikekelund:
```suggestion
```
Let's remove this type and rely on `UserData` accurately reflecting the config as it exists on disk.

apps/studio/src/storage/user-data.ts
```diff
@@ -1,98 +1,24 @@
...
 			const parsed = JSON.parse( asString );
-			const data = fromDiskFormat( parsed );
-
-			sortSites( data.sites );
-			populatePhpVersion( data.sites );
-			return data;
+			const { version, siteMetadata, ...data } = parsed as PersistedUserData;
+			return { siteMetadata: siteMetadata ?? {}, ...data };
 		} catch ( err ) {
-			// Awkward double try-catch needed to have access to the file contents
```
fredrikekelund, line 95: Not a big deal, but I think it's fair to keep this comment, too.

apps/studio/src/ipc-handlers.ts
```diff
@@ -652,9 +652,8 @@ export async function deleteSite( event: IpcMainInvokeEvent, id: string, deleteF
 	try {
 		await lockAppdata();
 		const userData = await loadUserData();
-		const siteIndex = userData.sites.findIndex( ( s ) => s.id === id );
-		if ( siteIndex !== -1 ) {
-			userData.sites.splice( siteIndex, 1 );
+		if ( userData.siteMetadata[ id ] ) {
+			delete userData.siteMetadata[ id ];
 			await saveUserData( userData );
 		}
```
fredrikekelund: Can we move this logic into `SiteServer::unregister` and call it here? It looks like we're not cleaning up `siteMetadata` in response to `SITE_EVENTS.DELETED` events now. This is a suggestion for how we can fix that problem, since `SiteServer::unregister` is called in `cli-events-subscriber.ts`

tools/common/lib/shared-config.ts
```diff
@@ -79,7 +71,7 @@ export async function readSharedConfig(): Promise< SharedConfig > {
 export async function saveSharedConfig( config: SharedConfig ): Promise< void > {
 	config.version = SHARED_CONFIG_VERSION;
```
fredrikekelund:
```suggestion
```
Since the version type is already a literal, we don't need to overwrite this value

tools/common/lib/shared-config.ts
fredrikekelund, line 1: Could we update the `updateSharedConfig` definition like so?

```ts
export async function updateSharedConfig( update: Partial< SharedConfig > ): Promise< void > {
```

Since the version type is already a literal, we don't need to apply any special treatment. The less special treatment, the lower the risk for future problems.

tools/common/lib/shared-config.ts
```diff
@@ -19,13 +19,16 @@ export class SharedConfigVersionMismatchError extends Error {
 	}
 }
 
-const SHARED_CONFIG_FILENAME = 'shared.json';
+export {
+	getConfigDirectory as getSharedConfigDirectory,
+	getSharedConfigPath,
+} from './config-paths';
```
fredrikekelund:
```suggestion
```
Looks like AI nonsense

fredrikekelund (commented):
So great to have the `Migration` interface. Deceptively simple, but I can guarantee it'll make future migrations easier to implement, and, above all, to maintain. 

As for the PR as a whole, it's nearly there. Deleting `siteMetadata` entries in response to `SITE_EVENTS.DELETED` events, removing the `PersistedUserData` type, reducing special treatment around config `version` fields and renaming the migration files were the only relevant issues I could identify. Let's address those, and once that's done, I'll approve the PR 👍

apps/studio/src/storage/paths.ts
fredrikekelund, line 1: The `getUserDataLockFilePath` function should probably also move to shared

fredrikekelund (approved):
👍

PR #2837, opened 2026-03-18
Remove peer true from package-lock.json and update package.json devEngines to use npm >11

sejas:
## Related issues

- N/A

## How AI was used in this PR

Calude Code implemented the changes and edited package.json back and forth based on PR feedback. It was manually tested.

## Proposed Changes

Removes `"peer": true` entries from `package-lock.json`. Those were generated when running `npm i` in a different node version, and added on https://github.com/Automattic/studio/pull/2830/changes#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519R9267
It replaces `engines` with `devEngines` and moves `packageManager` inside it to fail if `npm i` is called with a lower node/npm version.

## Testing Instructions

- Run `nvm use && npm install` and verify no changes to `package-lock.json`

- Run `npm install` on a fresh terminal, and if your default npm version is lower than 11 observe an error due to `devEngine` npm version.

Screenshot 2026-03-18 at 13 12 28

## Pre-merge Checklist


fredrikekelund (commented):
I think npm v10 is adding them and npm v11 is removing them. Updating the `engines` field in package.json doesn't force developers to use npm v11 (unless we set the `engine-strict` [setting](https://docs.npmjs.com/cli/v11/using-npm/config#engine-strict) in `.npmrc`), but at least it displays a warning. I recommend making that change, too.

sejas:
> I think npm v10 is adding them and npm v11 is removing them. Updating the `engines` field in package.json doesn't force developers to use npm v11 (unless we set the `engine-strict` [setting](https://docs.npmjs.com/cli/v11/using-npm/config#engine-strict) in `.npmrc`), but at least it displays a warning. I recommend making that change, too.

@fredrikekelund , thanks for the suggestion. I set the engine-strict true on https://github.com/Automattic/studio/pull/2837/changes/e8d681a2f5fa2eb4158b2b691f68204a1dda3bb9 and also modified the minimum npm version on package.json https://github.com/Automattic/studio/pull/2837/changes/9f24bfad1ec676d21f0de9289a45167d0df02cc3 , not sure if it's too drastic.
I updated the `packageManager` for consistency, although it doesn’t seem to have any effect on my side.

epeicher:
@sejas, @fredrikekelund , bringing this comment p1772934262057599-slack-C0926NEBC, about using`devEngines` as standard. However, in Gutenberg they decided to continue using `engines` - [link here](https://github.com/WordPress/gutenberg/pull/76391#issuecomment-4040313589), but maybe we can use `devEngines` too?

sejas:
Thanks! I don’t have a strong opinion here. I would have preferred that they merge that PR so we could follow their approach.

fredrikekelund:
I wasn't aware of `devEngines` – that looks like a good approach 👍 It seems to have been introduced in npm v10, but that's good enough to me.

The main thing to me is kind of just that it's nice if we don't have to add a new root-level `.npmrc` for this

sejas:
@epeicher , @fredrikekelund , thanks for the suggestions. I applied the changes to use `devEngines` 👌 . Can you have a final look?

package-lock.json
```diff
@@ -45,7 +45,7 @@
 			},
 			"engines": {
 				"node": ">=22.0.0",
-				"npm": ">=10.0.0"
+				"npm": ">=11.0.0"
```
fredrikekelund, line 48: Was this auto-generated by the npm CLI?

fredrikekelund (approved):
LGTM 👍 I get a fatal error if I try using npm v10

package-lock.json
sejas: No, it was _manually_ edited by Claude Code.

sejas:
I created and quickly merged a follow-up since the latest change also modified the package-lock.json , and I didn't notice until testing it on trunk https://github.com/Automattic/studio/pull/2841

PR #2840, opened 2026-03-18
Add different MCP configuration options

nightnei:
## Related issues

- Fixes STU-1389

## How AI was used in this PR
AI assisted me.

## Proposed Changes
This PR provides more detailed information to make the MCP configuration instructions clearer.
Screenshot 2026-03-18 at 19 53 09

## Testing Instructions
1. Run `claude mcp add --scope user wordpress-studio -- studio mcp`
2. Run `claude mcp list`
3. Assert that you see `wordpress-studio`
4. Run `codex mcp add wordpress-studio -- studio mcp`
5. Run `codex mcp list`
6. Assert that you see `wordpress-studio`
7. Run `npm run cli:build && node apps/cli/dist/cli/main.js mcp`
8. Assert that you see correct path for Claude Desktop (either `~/Library/...` for macOS or `%APPDATA%...` for Windows)
9. Assert that printed info looks good
10. You can also test configuration of Cursor and Claude Desktop to ensure that paths are good, but it's enough to validate that they are the same as in the decumentation - [Claude Desktop "Access Developer Settings"](https://modelcontextprotocol.io/docs/develop/connect-local-servers) and [Cursor documentation](https://cursor.com/docs/mcp#configuration-locations).

### DIFF
--- a/apps/cli/commands/mcp.ts
+++ b/apps/cli/commands/mcp.ts
@@ -1,4 +1,8 @@
-import { getMcpServerConfigJson } from '@studio/common/lib/mcp-config';
+import {
+	getMcpServerConfigJson,
+	getMcpServerLaunchCommand,
+	MCP_SERVER_NAME,
+} from '@studio/common/lib/mcp-config';
 import { __ } from '@wordpress/i18n';
 import { startMcpStdioServer } from 'cli/ai/mcp-server';
 import { Logger, LoggerError } from 'cli/logger';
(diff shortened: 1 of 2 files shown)


katinthehatsite:
I am wondering if we should call Claude Code, Codex etc. an "AI Assistant". My concern comes from the fact that we already have an `AI Asssitant` as an option within Studio and the naming is identical. Mayvbe we can say something like AI tool? Or along those lines?

tools/common/lib/mcp-config.ts
```diff
@@ -3,9 +3,22 @@
  * AI assistant's settings. Shared between the CLI `studio mcp` command
  * and the Studio desktop Settings dialog.
  */
-export function getMcpServerConfig(): Record< string, { command: string; args: string[] } > {
+type McpServerConfig = {
+	command: string;
+	args: string[];
+};
+
+export const MCP_SERVER_NAME = 'wordpress-studio';
```
katinthehatsite, line 11: Should we move this constant to constants file?

apps/cli/commands/mcp.ts
```diff
@@ -1,34 +1,60 @@
...
 	const lines = [
 		'',
 		__( 'WordPress Studio MCP Server' ),
 		'─'.repeat( 40 ),
 		'',
-		__( "Add the following to your AI assistant's MCP configuration:" ),
+		__( 'Run one of the following commands to add MCP support to your AI assistant:' ),
+		'',
+		`  ${ __( 'Claude Code' ) }`,
+		`    claude mcp add ${ MCP_SERVER_NAME } -- ${ launchCommand }`,
```
sejas: We can add `--scope user` https://code.claude.com/docs/en/mcp#user-scope

sejas (approved):
Looks great, thank you. I thought about showing only paths or only commands, or moving the sections around, but your approach makes sense.
Display commands, if not manual setting using the JSON.

I also thought about checking whether the paths exist, but as discussed in the huddle, showing the information even if the user doesn’t have it installed communicates that Studio MCP works with those tools.

nightnei:
> I am wondering if we should call Claude Code, Codex etc. an "AI Assistant". My concern comes from the fact that we already have an `AI Asssitant` as an option within Studio and the naming is identical. Mayvbe we can say something like AI tool? Or along those lines?

@katinthehatsite  Recently I tried to understand different names for such tool and what I noticed is that the common name which is used everywhere to menthion ChatGPT/ClaudeCode etc is "AI Assistant" so I intentionally used it here to keep commonly known name. But I see your point, IDK, probably I would keep "AI Assistant", but don't have striogn preference.

tools/common/lib/mcp-config.ts
nightnei: I would keep it here since I don't anticipate this constant being used outside the MCP configuration code. If one day it's used outside - we can put it to common/constants.ts

katinthehatsite: Sounds good 👍

katinthehatsite:
> @katinthehatsite Recently I tried to understand different names for such tool and what I noticed is that the common name which is used everywhere to menthion ChatGPT/ClaudeCode etc is "AI Assistant" so I intentionally used it here to keep commonly known name. But I see your point, IDK, probably I would keep "AI Assistant", but don't have striogn preference.

I don't have a strong preference either, I saw that some were using "AI tool" and some used "assistant". We can keep it as is I guess, especially if we plan to refactor our assistant in the future and it might take a different shape.

PR #2844, opened 2026-03-19
Improve Playground CLI child process error handling

fredrikekelund:
## Related issues

- Related to p1773148946011199-slack-C06DRMD6VPZ

## How AI was used in this PR

Codex was used to implement the timestamping in the process manager daemon and to reason about `startingPromise` logic in the child process.

## Proposed Changes

A couple of simpler changes: 

1. Unset the `stack` prop on `CliCommandError` errors. This is because the strongest "event grouping" signal Sentry uses is the error stack trace. All `CliCommandError` stack traces are identical, because we always construct them at the same place in `execute-command.ts`. This currently results in all `CliCommandError` errors being grouped into a single Sentry issue, which is incorrect.
2. Timestamp all lines that are written to the child process logs.
3. More logging in the Playground CLI child process.

And some more complex changes:

1. I've improved the abort event handling in `startServer` and made `stopServer` able to stop an in-flight `startServer` operation.
   1. This is based on a hunch from looking at one of the `CliCommandError` Sentry events. It's possible to run a `site stop` CLI command for a site that's currently starting, and it's also possible that users might trigger `site stop` operations from Studio if another error leads to Studio losing track of the current started/stopped state for sites. Reproducing this error from within Studio is impossible under normal circumstances, so this falls mostly under "robustness".
2. I've made `stopWordPressServer` wait for the child process to exit cleanly (with fallback), and I've made the child process actually exit cleanly in response to `stop-server` messages.
   1. This is also more of a robustness change. Previously, we'd always send a SIGKILL signal to the child. That's OK, but ideally a `stop-server` message should lead to a `0` exit code unless something actually went wrong. 

## Testing Instructions

For the changes in `wordpress-server-child.ts`, code review and CI are paramount. However, we can verify that stopping an in-flight start operation works as expected:

1. Run `npm run cli:build`
2. Open two terminal tabs
3. Type `node apps/cli/dist/cli/main.js site stop --path PATH_TO_SITE` in the first tab. Don't run this command yet. Leave it so you can quickly return to the tab after starting the site.
4. Type `node apps/cli/dist/cli/main.js site start --path PATH_TO_SITE` in the second tab and run the command.
5. Quickly return to the first terminal tab and click <kbd>Enter</kbd>
6. In the second tab, ensure that you see the following message `Failed to start WordPress server: This operation was aborted`

Test that log file lines are timestamped:

1. Open `~/.studio/pm2/logs`
2. Clean the directory
3. Run `npm start`
4. Start a site
5. Find the logfile for your site
6. Ensure that every line has a timestamp

## Pre-merge Checklist

### DIFF
--- a/apps/cli/lib/daemon-client.ts
+++ b/apps/cli/lib/daemon-client.ts
@@ -246,12 +246,13 @@ const daemonListProcessesSuccessResponseSchema = z.object( {
 
 // Cache the process list returned from the process manager for a very short time to make multiple
 // calls in quick succession more efficient
-const listProcesses = cacheFunctionTTL( async () => {
+async function listProcesses() {
+	await connectToDaemon();
 	const response = await sendDaemonRequest( {
 		type: 'list-processes',
 	} );
 	return daemonListProcessesSuccessResponseSchema.parse( response ).processes;
-} );
+}
 
 export async function getDaemonBus(): Promise< DaemonBus > {
 	if ( ! daemonBus ) {
@@ -289,10 +290,6 @@ export async function isProcessRunning(
 	processName: string
 ): Promise< ProcessDescription | undefined > {
 	try {
-		if ( ! isConnected ) {
-			return undefined;
-		}
-
 		const processes = await listProcesses();
 		return processes.find( ( p ) => p.name === processName && p.status === 'online' );
 	} catch ( error ) {
@@ -322,6 +319,12 @@ export async function startProcess(
 }
 
 export async function stopProcess( processName: string ): Promise< void > {
+	const runningProcess = await isProcessRunning( processName );
+
+	if ( ! runningProcess ) {
+		return;
+	}
+
 	await connectToDaemon();
 	await sendDaemonRequest( {
 		type: 'stop-process',
--- a/apps/cli/wordpress-server-child.ts
+++ b/apps/cli/wordpress-server-child.ts
@@ -321,6 +321,7 @@ const startServer = wrapWithStartingPromise(
 				errorToConsole( `Failed to start server:`, error );
 			}
 
+			// Rethrowing the error so that `ipcMessageHandler` returns an error IPC response and kills the process
 			throw error;
 		} finally {
 			startupAbortController = null;
@@ -354,6 +355,7 @@ async function stopServer(): Promise< void > {
 		logToConsole( 'Server stopped gracefully' );
 	} catch ( error ) {
 		errorToConsole( 'Error during server disposal:', error );
+		// Rethrowing the error so that `ipcMessageHandler` returns an error IPC response and kills the process
 		throw error;
 	} finally {
 		server = null;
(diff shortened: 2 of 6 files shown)


apps/cli/lib/daemon-client.ts
```diff
@@ -246,12 +246,13 @@ const daemonListProcessesSuccessResponseSchema = z.object( {
 
 // Cache the process list returned from the process manager for a very short time to make multiple
 // calls in quick succession more efficient
-const listProcesses = cacheFunctionTTL( async () => {
+async function listProcesses() {
```
fredrikekelund, line 249: Removing the `cacheFunctionTTL` is not crucial to this PR's purpose. I just started viewing it as a liability while working on the `stopWordPressServer` logic. `isProcessRunning` depends on `listProcesses`, which caches its result for 1 second. That could easily lead to a sneaky bug at some point, and communicating with the process manager daemon over a socket is pretty fast to begin with.

apps/cli/wordpress-server-child.ts
```diff
@@ -279,19 +288,17 @@ const startServer = wrapWithStartingPromise(
...
+
 		try {
-			signal.addEventListener(
-				'abort',
-				() => {
-					throw new Error( 'Operation aborted' );
-				},
-				{ once: true }
-			);
+			stopSignal.throwIfAborted();
```
fredrikekelund, line 295: It was a mistake on my part to think that an exception thrown from this event handler would be caught by the surrounding `try..catch`. We need to rely on `throwIfAborted` instead.

apps/cli/wordpress-server-child.ts
```diff
@@ -424,15 +461,19 @@ function parsePhpError( error: unknown ): string {
...
+			topic: 'error',
+			errorMessage: parsePhpError( error ),
+			errorStack: error instanceof Error ? error.stack : undefined,
+			cliArgs: lastCliArgs ?? undefined,
+		};
+		process.send!( errorResponse, () => {
+			resolve();
+		} );
+	} );
 }
```
fredrikekelund, line 477: Not sure if we really need to wait for the `process.send` callback or if that operation is effectively synchronous, but since the next thing that happens after this function is called is that we kill the process, I thought it best to.

apps/cli/wordpress-server-child.ts
```diff
@@ -493,10 +535,19 @@ async function ipcMessageHandler( packet: unknown ) {
...
+			process.disconnect();
+		}
 	} catch ( error ) {
 		errorToConsole( `Error handling message ${ validMessage.topic }:`, error );
-		sendErrorMessage( validMessage.messageId, error );
+		await sendErrorMessage( validMessage.messageId, error );
+		originalConsoleLog( 'Killing process because of', error );
 		process.exit( 1 );
+	} finally {
+		delete abortControllers[ validMessage.messageId ];
```
fredrikekelund, line 550: Taken from https://github.com/Automattic/studio/pull/2805 👍

epeicher (approved):
Thanks @fredrikekelund! I tested it and received the expected error message when stopping a site that was starting. I can also see the _timestamped_ logs. Also, code changes LGTM!

|Stopping a Starting site | Timestamped Logs|
|-------|------|
|  Image|  Image|

PR #2847, opened 2026-03-19
Add .sock cleanup to ensure Studio uses new _events after unclean shutdown

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

I used it for debugging the issue and to add code that removes socket file. 

## Proposed Changes

I propose adding .sock cleanup to ensure Studio uses new _events after an unclean shutdown.

In the other case, when an old socket is not cleaned up after an unclean shutdown, Studio will reuse it.

## Testing Instructions

- Start Studio, start sites, confirm they are shown as running

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/_events.ts
+++ b/apps/cli/commands/_events.ts
@@ -10,6 +10,7 @@ import { sequential } from '@studio/common/lib/sequential';
 import { SITE_EVENTS, siteDetailsSchema, SiteEvent } from '@studio/common/lib/site-events';
 import { SiteCommandLoggerAction as LoggerAction } from '@studio/common/logger-actions';
 import { __ } from '@wordpress/i18n';
+import fs from 'fs';
 import { z } from 'zod';
 import { getSiteUrl, readAppdata, SiteData } from 'cli/lib/appdata';
 import {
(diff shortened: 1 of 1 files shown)


apps/cli/commands/_events.ts
```diff
@@ -107,6 +108,14 @@ export async function runCommand(): Promise< void > {
...
 	process.on( 'SIGTERM', () => void cleanup() );
 
+	// Remove any stale socket from a previous Studio session. Studio is single-instance,
+	// so any existing events.sock belongs to a dead session and must be replaced.
+	try {
+		fs.unlinkSync( SITE_EVENTS_SOCKET_PATH );
+	} catch ( err ) {
+		// ENOENT is fine — socket didn't exist. Any other error is unexpected but non-fatal.
+	}
+
```
fredrikekelund, line 120: Let's do this only on non-Windows platforms. On Windows, `SITE_EVENTS_SOCKET_PATH` isn't actually a file system path

wojtekn: Good point. Thinking about it more, we could move this to `eventsSocketServer.listen()`, making the cleanup the SocketServer's responsibility rather than the caller's. Do you see anything against that?

fredrikekelund: `SocketServer::listen` already has some clever error-handling logic inspired by pm2-axon. If listening on the socket fails, it'll try connecting as a client to see whether the socket has a server on the other end (if not, then the only connected parties are clients). If it cannot connect, then it'll remove the socket file and try again. 

This is why I was slightly surprised that we need this additional layer of defense, but clearly, there are scenarios when there are stray processes that we need to handle. 

So, in short, we shouldn't assume that the socket file can always be deleted in `SocketServer::listen`.

However, I do think `SocketServer::close` could use additional logic to remove the socket file as well.

wojtekn: This logic won't remove the old socket if it can connect to it, and the issue may arise when an outdated server is on the other end.

It makes sense to keep `SocketServer::listen` generic and keep the brute-force cleanup logic as part of the _event command, since this command never needs to attach to an existing socket.

wojtekn: I added a guard for the Windows.

fredrikekelund (approved):
I think we could use this PR as an opportunity to also remove the socket file in `SocketServer::close`, but we could also do that separately. This change by itself should fix the issue at hand

PR #2853, opened 2026-03-19
Add resizable sidebar

shaunandrews:
image

## Proposed Changes

- Add drag-to-resize on the sidebar's right edge with a custom `useSidebarResize` hook
- Sidebar width is continuously adjustable between 150px and 400px
- Dragging below 100px snaps the sidebar closed; reopening restores the last width
- Width persists to `localStorage` across sessions
- Blue 3px indicator line appears on hover/drag to signal the resize handle
- Updated `toggleMinWindowWidth` IPC handler to use the current sidebar width instead of the hardcoded constant
- CSS transitions are disabled during drag to prevent lag

## Testing Instructions

- Hover over the right edge of the sidebar — a blue line should appear
- Drag to resize the sidebar between 150px and 400px
- Drag below ~100px — sidebar should snap closed
- Use the toggle button to reopen — should restore your last dragged width
- Quit and relaunch the app — sidebar width should persist
- Toggle the sidebar closed/open — window should resize by the correct amount


katinthehatsite (commented):
Generally, the changes look good to me 👍 

I noticed one thing where in some languages that use longer transations, you can see the running sites being mushed together:

Image

In my test, I switched to Ukrainian and resized the sidebar to as narrow as I could before it hides. 

Additionally, I would be curious to test this on Windows. I can take a look into that tomorrow.

apps/studio/src/hooks/use-sidebar-resize.ts
```diff
@@ -0,0 +1,106 @@
+import { useState, useCallback, useRef, useEffect } from 'react';
+import {
+	SIDEBAR_WIDTH,
+	SIDEBAR_MIN_WIDTH,
+	SIDEBAR_MAX_WIDTH,
+	SIDEBAR_SNAP_THRESHOLD,
+	LOCAL_STORAGE_SIDEBAR_WIDTH_KEY,
+} from 'src/constants';
+
+function loadSavedWidth(): number {
```
katinthehatsite: The logic in this function is quite similar to the logic in `getCurrentSidebarWidth` - do you think it can be moved to a shared utility?

katinthehatsite:
For the issue in the sidebar with the sites running, I see it now has been resolved:

Screenshot 2026-04-14 at 3 19 07 PM

apps/studio/src/hooks/use-sidebar-resize.ts
katinthehatsite: I added this change in https://github.com/Automattic/studio/pull/2853/changes/6caa470b5c54807afd68429b2de1fe39e45df3e6

katinthehatsite:
@Automattic/yolo could you please review? I made some changes based in my recommendations but I would appreciate another pair of eyes.

apps/studio/src/lib/sidebar-utils.ts
```diff
@@ -0,0 +1,17 @@
+import {
+	SIDEBAR_WIDTH,
+	SIDEBAR_MIN_WIDTH,
+	SIDEBAR_MAX_WIDTH,
+	LOCAL_STORAGE_SIDEBAR_WIDTH_KEY,
+} from 'src/constants';
+
+export function getSavedSidebarWidth(): number {
+	const saved = localStorage.getItem( LOCAL_STORAGE_SIDEBAR_WIDTH_KEY );
```
bcotrim, line 9: Why did we decide on a local storage approach?
We already have some UI properties (window size) living in the `app.json` config file.
I don't have anything against local storage, I just believe it should be consistent with our other properties.

katinthehatsite: I don't have the answer specifically for that bit since it was not originally my PR but it makes sense to be consistent with our other properties. I will check how we can align our approach here 👀

apps/studio/src/hooks/use-sidebar-resize.ts
```diff
@@ -0,0 +1,96 @@
...
+			if ( finalWidth < SIDEBAR_SNAP_THRESHOLD ) {
+				// Snap closed — restore the last good width for when it reopens
+				setSidebarWidth(
+					dragStartWidth.current > SIDEBAR_MIN_WIDTH ? dragStartWidth.current : SIDEBAR_WIDTH
+				);
+				if ( isSidebarVisible ) {
+					toggleSidebar();
+				}
+				return;
+			}
```
gcsecsey, line 70: To me, the UX for snapping the sidebar closed feels a bit strange. When trying to snap it closed, the current threshold feels a bit too high, the user needs to drag the line almost half the way before it registers as snapped. The content beneath also reflows when this happens.

https://github.com/user-attachments/assets/748eec4b-1369-46a9-ada4-881f1008b515

I'm wondering if we could omit this snap-to-close behavior completely and just rely on the toggle button on top?

gcsecsey (commented):
The changes make sense to me, and it overall works well. I added a comment on the snapping.

apps/studio/src/lib/sidebar-utils.ts
katinthehatsite: I had a chance to check this a bit further and I think the localStorage approach would be all right here. The sidebar width uses `localStorage` rather than app.json because it's renderer-side state: it's read synchronously at component init (allowing it to be passed directly as a useState initializer) and never needed by the Main process. `windowBounds` lives in `app.json` because Electron's Main process must restore the window size before the renderer exists. Moving sidebar width to `app.json` would introduce an async IPC  to initialize a CSS value, requiring a loading state and added complexity.

apps/studio/src/hooks/use-sidebar-resize.ts
katinthehatsite: I ended up removing the snap behaviour so that now it looks like this:

https://github.com/user-attachments/assets/3db9aa32-e87c-4618-9e8f-bfb47cc4d66a

Let me know what you think!

katinthehatsite (approved):
The changes look good, I also re-requested the review from two other reviewers.

apps/studio/src/components/main-sidebar.tsx
```diff
@@ -8,15 +8,22 @@ import AddSite from 'src/modules/add-site';
...
 	const { sites: localSites } = useSiteDetails();
 
 	return (
 		<div
 			data-testid="main-sidebar"
-			className={ cx( 'text-chrome-inverted relative pt-[10px]', className ) }
+			className={ cx(
+				'text-chrome-inverted relative',
+				isMac() && 'pt-[10px]',
+				! isMac() && 'pt-[38px]',
```
nightnei: Why do we need it?

apps/studio/src/components/app.tsx
```diff
@@ -87,10 +92,27 @@ export default function App() {
...
+						<div
+							className="group h-full w-3 cursor-col-resize flex-shrink-0 z-20 flex items-stretch justify-start -ml-1.5 -mr-1.5"
+							onMouseDown={ handleMouseDown }
+						>
+							<div
+								className={ cx(
+									'w-[3px] rounded-[2px] transition-opacity duration-150',
+									isDragging
+										? 'bg-[#3858e9] opacity-100'
+										: 'bg-[#3858e9] opacity-0 group-hover:opacity-100'
```
nightnei: NIT
```suggestion
									'w-[3px] rounded-[2px] transition-opacity duration-150 bg-[#3858e9]',
									isDragging
										? 'opacity-100'
										: 'opacity-0 group-hover:opacity-100'
```

apps/studio/src/components/main-sidebar.tsx
katinthehatsite: Let me double-check, if I am not mistaken this was related to resolving conflicts with trunk 👀

katinthehatsite: I removed it as it seems to be an unintended leftover after the conflict resolution

apps/studio/src/components/app.tsx
katinthehatsite: Applied the suggestion

apps/studio/src/components/main-sidebar.tsx
nightnei: Thanks Kat

nightnei (approved):
Great improvement, I will increase sidebar width right after this PR is merged :)

apps/studio/src/lib/sidebar-utils.ts
bcotrim: That makes a lot of sense, thanks for the explanation! 👍

apps/studio/src/components/app.tsx
```diff
@@ -87,10 +92,25 @@ export default function App() {
...
+							style={ isSidebarVisible ? { flexBasis: `${ sidebarWidth }px` } : undefined }
 						/>
+						{ /* Resize handle */ }
+						<div
+							className="group h-full w-3 cursor-col-resize flex-shrink-0 z-20 flex items-stretch justify-start -ml-1.5 -mr-1.5"
+							onMouseDown={ handleMouseDown }
+						>
+							<div
+								className={ cx(
+									'w-[3px] rounded-[2px] transition-opacity duration-150 bg-[#3858e9]',
```
bcotrim:
```suggestion
									'w-[3px] rounded-[2px] transition-opacity duration-150 bg-frame-theme',
```

Nit: `--color-frame-theme` is the definition and it resolves to different colors in dark/light themes. It will still look good, blueish for both themes, but using the tailwind theme colors will make future adjustments easier.

bcotrim (approved):
LGTM 👍 
Thanks for completing the changes @katinthehatsite

PR #2859, opened 2026-03-20
Add command for deselecting site

mtias:
We may want to explore allowing esc key while on site list to do this, but it may be too hidden


gcsecsey (commented):
Hey @mtias! 👋 This has been open with no movement for a while. Is deselecting the active site still something we want to add? If so, I'm happy to rebase it and fix it up, otherwise I think we can close it. WDYT?

wojtekn:
Marking as draft until tests and conflicts are solved.

PR #2860, opened 2026-03-20
Update fast-xml-parser to fix Dependabot issue

wojtekn:
## Related issues

- Fixes https://github.com/Automattic/studio/security/dependabot/146

## How AI was used in this PR

I made a change myself.

## Proposed Changes

- I propose updating fast-xml-parser to fix Dependabot issue.

## Testing Instructions

- Confirm if CI goes through.

## Pre-merge Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9254,7 +9254,6 @@
 			"os": [
 				"darwin"
 			],
-			"peer": true,
 			"engines": {
 				"node": ">=10"
 			}
@@ -9271,7 +9270,6 @@
 			"os": [
 				"darwin"
 			],
-			"peer": true,
 			"engines": {
 				"node": ">=10"
 			}
@@ -9288,7 +9286,6 @@
 			"os": [
 				"linux"
 			],
-			"peer": true,
 			"engines": {
 				"node": ">=10"
 			}
@@ -9305,7 +9302,6 @@
 			"os": [
 				"linux"
 			],
-			"peer": true,
 			"engines": {
 				"node": ">=10"
 			}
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -9254,7 +9254,6 @@
 			"os": [
 				"darwin"
 			],
-			"peer": true,
```
fredrikekelund, line 9257: Are you using npm 11, @wojtekn?

fredrikekelund (approved):
It's actually not entirely sufficient to address these issues by updating the lockfile, because when we package the CLI, we use only `apps/cli/package.json` to install a `node_modules` directory to be copied to the output (without devDependencies).

In this case, because of how the version constraints work, we should still ship the latest version of `fast-xml-parser` with the packaged CLI, but that always depends on the dependency tree and the version constraints "along the way".

In this case, I would complement this PR with another PR in the Playground repo that updates the `fast-xml-parser` dependency version.

package-lock.json
wojtekn: Yes, 11.8.

wojtekn:
@fredrikekelund good point. In this case, it makes sense to close this one, update it in the Playground repo, and then update Playground in Studio when new version is released.

fredrikekelund:
Updating the lockfile will make it so we have a newer version while developing Studio, so I still think it makes sense to land this PR. It won't make a difference when we package the CLI, though. Then, we should get the latest version that fits the constraint specified by Playground (`^5.5.1` in this case)

package-lock.json
fredrikekelund: I thought that would make it so we didn't have this flip-flopping of `peer` props in the lockfile, but it doesn't really matter either way…

wojtekn:
@fredrikekelund sounds good, opened https://github.com/WordPress/wordpress-playground/pull/3422

PR #2861, opened 2026-03-20
Use light mode as the default color scheme

sejas:
## Related issues

- Fixes STU-1429

## How AI was used in this PR

Claude Code identified all locations where the default color scheme was set to `system` and changed them to `light`. It also identified that the color scheme picker was saving immediately on click (unlike other preferences), and aligned it with the dirty state + Save button pattern. It then added live preview support so theme changes are visible immediately while still requiring Save to persist.

## Proposed Changes

- Change the default color scheme from `system` to `light` so new users always get light mode regardless of their OS dark mode setting
- Align the color scheme picker with the dirty state pattern used by other preferences (editor, terminal, locale) — changes are now local until the user clicks Save
- Add live preview: selecting a color scheme immediately applies the theme visually, but Save is required to persist the change
- Cancel, Escape, changing tab, or clicking outside the modal reverts the theme to the previously saved value
- Add `previewColorScheme` IPC handler that applies `nativeTheme.themeSource` without persisting to disk
- Sync `nativeTheme.themeSource` when reading the color scheme preference to ensure consistency

## Testing Instructions

- Remove the `colorScheme` entry from `~/Library/Application Support/Studio/appdata-v1.json` (or use a fresh install)
- Verify the app starts in light mode
- Open Settings → General and verify "Light" is selected by default
- Change appearance to "Dark", verify the app immediately switches to dark mode (live preview)
- Click Cancel, verify the app reverts to light mode
- Reopen Settings, change to "Dark" again, verify live preview applies
- Press Escape, verify the app reverts to light mode
- Reopen Settings, change to "Dark", click Save, verify the app stays in dark mode
- Reopen Settings, verify "Dark" is still selected
- Change to "Light", click outside the modal to dismiss, verify the app stays in dark mode (reverted)
- Change to "Light", click Save, verify it persists

https://github.com/user-attachments/assets/6cf982e7-6eed-4ee6-aac7-f1c2b4af33b3

## Pre-merge Checklist

### DIFF
--- a/apps/studio/e2e/appearance.test.ts
+++ b/apps/studio/e2e/appearance.test.ts
@@ -34,8 +34,8 @@ test.describe( 'Appearance', () => {
 		// Preferences tab should be active by default with appearance radio group visible
 		await expect( settings.appearanceRadioGroup ).toBeVisible();
 
-		// Default should be System
-		await expect( settings.getAppearanceOption( 'System' ) ).toHaveAttribute(
+		// Default should be Light
+		await expect( settings.getAppearanceOption( 'Light' ) ).toHaveAttribute(
 			'aria-checked',
 			'true'
 		);
@@ -54,9 +54,9 @@ test.describe( 'Appearance', () => {
 		);
 		expect( isLight ).toBe( false );
 
-		// Switch back to System
-		await settings.selectColorScheme( 'System' );
-		await expect( settings.getAppearanceOption( 'System' ) ).toHaveAttribute(
+		// Switch back to Light
+		await settings.selectColorScheme( 'Light' );
+		await expect( settings.getAppearanceOption( 'Light' ) ).toHaveAttribute(
 			'aria-checked',
 			'true'
 		);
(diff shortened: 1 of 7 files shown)


wojtekn:
> and aligned it with the dirty state + Save button pattern.

As light-dark mode changes the whole UI, what if we did both - when the user switches modes, the app immediately applies the selected theme, but Save is required to preserve changes, and Cancel reverts them?

katinthehatsite:
> As light-dark mode changes the whole UI, what if we did both - when the user switches modes, the app immediately applies the selected theme, but Save is required to preserve changes, and Cancel reverts them?

I think this approach makes the most sense because it gives preview to users without have to save the changes to see them applied right away.

sejas:
I implemented the preview and revert on cancel on https://github.com/Automattic/studio/pull/2861/changes/654cf5c7313868e890b1f3b73adae0811df1d4ca

wojtekn (approved):

PR #2870, opened 2026-03-23
Studio Code Assistant

sejas:
## Related issues

* Fixes [STU-1450](https://linear.app/a8c/issue/STU-1450/improve-studio-assistant-ui-to-use-studio-code-command-under-the-hood)

## How AI was used in this PR

Currently mostly vibe-coded with superpowers.

## Proposed Changes

* It adds Studio Code assistant behind a feature flag.
* Users can interact with Studio Code through CLI process
* This is a first iteration. We'll add more features as a follow-up such as:
  * ~~Default empty chat~~
  * Invoking slash commands/skills
  * Switching providers
  * API key authentication
  * Switching sites
  * Save messages in the same session. Currently each message is saved in a separate session
  * Consider streaming responses in future iterations.

## Testing Instructions

* Run `ENABLE_STUDIO_CODE_UI=true npm start`
* Go to the assistant tab
* Ask questions and prompts about your site.

https://github.com/user-attachments/assets/8186f555-12c0-4d1c-8f98-929dc67c2cce

## Pre-merge Checklist


apps/studio/src/components/content-tab-assistant.tsx
```diff
@@ -357,6 +359,16 @@ const UnauthenticatedView = ( { onAuthenticate }: { onAuthenticate: () => void }
 );
```
katinthehatsite, line 359: I was able to access the UI even when I was logged out and it threw an error in a response message. I think we might want to improve this experience by either hiding it behind the login screen or find another alternative:

Image

katinthehatsite (commented):
Another bit that I found strange (although I understand why it is needed) is that you can type but can't send a message while the assistant is answering. With tools like Claude or Cursor, you can either queue the message or interrupt the ongoing request and send a new message:

https://github.com/user-attachments/assets/bc9e0123-7e5c-49a1-a357-d1e43170296d

apps/studio/src/modules/studio-code/ipc-handlers.ts
```diff
@@ -0,0 +1,56 @@
...
+	sitePath: string,
+	siteName: string,
+	message: string,
+	permissionResponse: Record< string, string >
+): Promise< void > {
+	spawnTurn( siteId, sitePath, siteName, message, { permissionResponse } );
+}
+
+export function studioCodeAbort( _event: IpcMainInvokeEvent, siteId: string ): void {
+	abortTurn( siteId );
```
katinthehatsite, line 30: I don't think we have a user facing option to abort a running turn. Should we? (similar comment to the other one I made about not being able to send a message while assistant is thinking)

apps/studio/src/modules/studio-code/studio-code-event-types.ts
```diff
@@ -0,0 +1,34 @@
+// Event types matching the JSON mode output from `studio code --json`
```
katinthehatsite, line 1: There is a file that is named the same in `  apps/studio/src/components/studio-code-types.ts` and I am wondering if we should rename them to make them cleaner e.g 👍 
* studio-code-ui-types.ts (the component/renderer types)
* studio-code-event-types.ts (the CLI wire format types)

youknowriad:
>  With tools like Claude or Cursor, you can either queue the message or interrupt the ongoing request and send a new message.

Let's track that. I think it's a bug on Studio Code (not just here but also in CLI)

sejas:
> > With tools like Claude or Cursor, you can either queue the message or interrupt the ongoing request and send a new message.
> 
> Let's track that. I think it's a bug on Studio Code (not just here but also in CLI)

Created STU-1574

apps/studio/src/components/ai-input.tsx
```diff
@@ -255,19 +257,24 @@ const UnforwardedAIInput = (
...
-									getIpcApi().openURL( telexUrl );
-									onClose();
-								} }
-								className="flex flex-row"
-							>
-								<SparklesIcon />
-								<span className="ltr:pl-2 rtl:pl-2">{ __( 'Build with Telex' ) }</span>
-								<ArrowIcon />
-							</MenuItem>
+							{ showTelexLink && (
```
wesleyfantinel, line 260: TBH, I don't think we should have this link to Telex.

apps/cli/commands/ai/index.ts
```diff
@@ -45,9 +46,11 @@ export async function runCommand( options: {
 	adapter: AiOutputAdapter;
 	initialMessage?: string;
 	resumeSession?: LoadedAiSession;
+	resumeSessionId?: string;
```
wesleyfantinel, line 49: I like the fact we can resume sessions now.

apps/cli/commands/ai/index.ts
```diff
@@ -88,6 +98,22 @@ export async function runCommand( options: {
...
+					s.linkedAgentSessionIds.includes( options.resumeSessionId! )
+				);
+				if ( existing ) {
+					sessionRecorder = await AiSessionRecorder.open( {
+						sessionId: existing.id,
+						filePath: existing.filePath,
+						linkedAgentSessionIds: existing.linkedAgentSessionIds,
+					} );
+				} else {
+					sessionRecorder = await AiSessionRecorder.create();
```
wesleyfantinel: Creating a new session if no session is found can be misleading for some systems. I'd throw an error here instead.

apps/cli/commands/ai/index.ts
```diff
@@ -471,6 +497,7 @@ export async function runCommand( options: {
 	// JSON mode: single turn, then exit
 	if ( isJsonMode && options.initialMessage ) {
 		try {
+			await maybeAutoSwitchProvider();
```
wesleyfantinel: It's good that we have this now, but I wonder if `maybeAutoSwitchProvider` should be moved somewhere else to make this standard to all modes.

wesleyfantinel (commented):
My understanding is that this is adding good changes to the headless (--json) mode. I haven’t looked at the studio app changes yet.

apps/studio/src/components/ai-input.tsx
sejas: Agreed. I'll create a follow-up issue to remove it and we can discuss it there. This PR just keeps intact the experience of Studio Assistant.

sejas:
@wesleyfantinel , @katinthehatsite , I addressed your feedback and added a logout page state. Could you take another look? I would like to merge the PR behind a feature flag and create follow-up issues for the rest of improvements.

apps/cli/commands/ai/index.ts
sejas: I move it higher in `runAgentTurn` https://github.com/Automattic/studio/blob/ca485bb84d5fc6ef9b4ba3f556620a4b3e708b81/apps/cli/commands/ai/index.ts#L393-L398

apps/cli/commands/ai/index.ts
sejas: Now it throws an error:
https://github.com/Automattic/studio/blob/ca485bb84d5fc6ef9b4ba3f556620a4b3e708b81/apps/cli/commands/ai/index.ts#L102-L125

wesleyfantinel (approved):

PR #2871, opened 2026-03-23
Migrate HTTPS certificates to `~/.studio/certificates`

fredrikekelund:
## Related issues

- Related to STU-1350
- Fixes STU-1416

## How AI was used in this PR

Not at all. 

## Proposed Changes

- Add a `copyHttpsCertsToWellKnown` migration (conforming to the `Migration` interface from https://github.com/Automattic/studio/pull/2836) that copies the `certificates` directory from the old path (relative to the Electron installation config files) to the new path (a well-known location: `~/.studio/certificates`).

## Testing Instructions

1. Have at least one site with HTTPS configured (do this before checking out this branch)
2. Check out this branch
3. `npm start`
4. `ls ~/.studio/certificates`
5. Ensure that the new certificates directory exists
6. Ensure that the HTTPS site you had since before still works

## Pre-merge Checklist


apps/studio/src/tests/index.test.ts
fredrikekelund, line 1: `src/storage/paths` is already mocked globally in `apps/studio/vitest.setup.ts`

bcotrim (commented):
Code LGTM and nice cleanup on tests!

Something I would like to discuss before we land, would it make sense to run this from the CLI migrations instead?
Since it migration copies the cert from the Studio folder to the CLI folder, it seems to make sense to have it run from the CLI. What do you think?

fredrikekelund:
> would it make sense to run this from the CLI migrations instead

Same as with the config file migration, the place where we need a guarantee that the code works with the current file structure is primarily the app, not the CLI. 

I also think it's sensible for the old owner (the Studio app) to execute the move to the new owner (the CLI). 

> migration copies the cert from the Studio folder to the CLI folder

`~/.studio/certificates` is not 100% under CLI management, as I see it. Yes, we're moving towards an architecture where the CLI acts as a back-end API, and it will probably be the only component that touches that directory. Still, the first difference between the old location and this one is that it's "well-known" (fully deterministic).

bcotrim (approved):
Let's keep it in Studio then.
LGTM 👍

PR #2873, opened 2026-03-23
Move WordPress dependencies (server-files) to CLI

fredrikekelund:
## Related issues

- Fixes STU-1442

## How AI was used in this PR

Sparingly. 

## Proposed Changes

To release the standalone CLI, we need to move the WordPress dependencies ("wp-files" and "server-files") to the CLI. This PR implements that through the following changes:

- Update the CLI Vite config to copy `wp-files` to the CLI output.
- Update the CLI `package.json` to include `wp-files` in the `files` array. 
- Change the server-files path getter functions in `apps/cli` and `apps/studio` to target `~/.studio/server-files`.
- Move the `setup-wp-server-files` logic to the CLI. 
   - Add a hidden `_update-deps` CLI command that Studio executes on startup (without awaiting). This command sets up and updates the dependencies.
   - The "setup" logic for dependencies is also triggered whenever the CLI runs a `site` command or the `wp` command (which are the only commands that need access to the server-files).
   - The "update" logic for dependencies is also triggered on the `site create` CLI command. 

What hasn't changed:

- The `wp-files` directory is still generated by the root-level `postinstall` npm script. Once the npm publishing flow has been automated as part of the existing CI release flow, we'll get the same freshly generated `wp-files` directory with every release that we're used to.
- We still distinguish between the `wp-files` directory (ships with Studio, not writable) and `server-files` (a copy of `wp-files`, writable, occasionally updated by the app).

## Testing Instructions

Run some light smoke tests to ensure that creating sites, running WP-CLI commands, and starting/stopping sites work as expected. 

## Pre-merge Checklist


bcotrim (commented):
Changes worked as expected and the overall approach looks good! 
Some comments for discussion:
1. Our approach for dependencies setup/update is already leaning towards that approach but we could be more explicit with a DependencyManager interface with setup, needsUpdate, update functions. It's a small refactor that results in code with a explicit intention
2. Could we run the setup as a middleware? I don't think it would cause a lot of overhead.
3. I would push back on the `_update-deps`, we could achieve a similar behavior with a middleware that checks the existing action and runs the update when needed (site create or start, or _events for example)

Let me know what you think.

fredrikekelund:
> we could be more explicit with a DependencyManager interface with setup, needsUpdate, update functions

Makes sense. I would do this in a follow-up PR, though. 

> Could we run the setup as a middleware? I don't think it would cause a lot of overhead.

Overhead was indeed my concern. All the setup functions just read a handful of files from disk, but it came down to me not knowing them well enough to trust them completely, I guess… I agree it's safer to use a middleware, though. I've pushed a commit that implements this. Let's go with this approach so long as the perf tests don't show any big regressions (I doubt they will).

> I would push back on the _update-deps

Yeah, this isn't needed if we use the middleware approach.

bcotrim (approved):
Merge conflicts but approving now to unblock.
Thanks for addressing the feedback @fredrikekelund 
👍

fredrikekelund:
Site creation is slower now. This is likely because every `site create` call checks the network to determine whether there are newer versions of the dependencies. I've parallelized the network calls to try and speed things up slightly. 

We also plan to make it so these calls are only made every 24h or so. STU-1455 tracks this.

apps/cli/vite.config.base.ts
```diff
@@ -48,7 +59,7 @@ export const baseConfig = defineConfig( {
 		commonjsOptions: {
 			ignoreDynamicRequires: true,
 		},
-		sourcemap: true,
+		sourcemap: false,
```
fredrikekelund, line 62: Not related to the core changes in this PR, I just found the sourcemaps super noisy. They're not needed for anything, so I disabled them

tools/common/package.json
```diff
@@ -13,15 +13,13 @@
 		"cross-port-killer": "^1.4.0",
 		"date-fns": "^3.3.1",
 		"fast-deep-equal": "^3.1.3",
-		"fs-extra": "^11.3.2",
```
fredrikekelund, line 16: This wasn't really needed. Node's standard `fs` module covers everything we need.

PR #2880, opened 2026-03-24
Studio Agents Suite: Install only selected global skills when creating a new site

katinthehatsite:
## Related issues

- Fixes STU-1451/

## How AI was used in this PR

It was used to help me identify the approach to implementation.

## Proposed Changes

This PR ensures that we save the user's preferences for skills that they want to install on their site and only these skills are installed or not installed when the new sites are created. 

## Testing Instructions

* Pull the changes from this site
* Enable the Agents Suite flag in the `Studio > Feature Flags` section
* Click on the toggle in the top right corner in Studio to open the Settings modal
* Navigate to the `Skills` section
* Select some skills (but not all) and install them
* Create a new site
* Confirm that only the selected skills were installed on that site (you can confirm this by checking site's files)
* Open the app data - you can use the option `Electron - Open AppData file` in the topbar - and confirm you can see the selected skills selection saved there

Screenshot 2026-03-24 at 11 55 49 AM

## Pre-merge Checklist


katinthehatsite:
One thing that I was not sure about is that what to do with the existing sites and their skills selection:

* Currently, when you create a new site, it is going to take the user preferences and install the skills on it. But also, if you select "Install" in the general settings, it will install it on every existing site. The `Remove` option works in the same way. 

I left this as is for now but happy to make further adjustments. What do you think @sejas ?

epeicher (commented):
Thanks @katinthehatsite for implementing this! I have tested it, and I can confirm that the skills that I selected are installed in the new site and correctly saved to the settings file. Changes also LGTM!

~~But I see this applies to all sites, I guess there will be a follow-up to allow installing skills per site, right?~~ Sorry! I just saw your comment where [this question is already asked](https://github.com/Automattic/studio/pull/2880#issuecomment-4117282510), I will wait for the reply 👀  

|Skills selected | Skills installed| Config file value
|-------|------|----
|  Image |Image |  Image

apps/studio/src/ipc-handlers.ts
```diff
@@ -238,6 +238,12 @@ export async function installWordPressSkillsToAllSites(
 			console.error( '[skills] Failed to install skill:', result.reason );
 		}
 	} );
+
+	const installedIds = options?.skillId ? [ options.skillId ] : BUNDLED_SKILLS.map( ( s ) => s.id );
```
sejas: What do you think if we avoid defaulting to all BUNDLED_SKILLS?. I'm worried it could be called externally by accident without `options?.skillId`

apps/studio/src/ipc-handlers.ts
```diff
@@ -238,6 +238,12 @@ export async function installWordPressSkillsToAllSites(
...
+	const userData = await loadUserData();
+	const existing = userData.selectedSkills ?? [];
+	const merged = Array.from( new Set( [ ...existing, ...installedIds ] ) );
+	await updateAppdata( { selectedSkills: merged } );
```
sejas: Let's wrap these with `lockAppdata` to prevent data corruption.

apps/studio/src/ipc-handlers.ts
```diff
@@ -252,6 +258,10 @@ export async function removeWordPressSkillFromAllSites(
 			console.error( '[skills] Failed to remove skill:', result.reason );
 		}
 	} );
+
+	const userData = await loadUserData();
+	const updated = ( userData.selectedSkills ?? [] ).filter( ( id ) => id !== skillId );
+	await updateAppdata( { selectedSkills: updated } );
```
sejas: Same here. I believe we need to lockAppdata.

sejas (commented):
Kat! thanks for working on this PR. Feature wise it works as expected. Great work!
I left some suggestions.

sejas:
> One thing that I was not sure about is that what to do with the existing sites and their skills selection:
> 
> * Currently, when you create a new site, it is going to take the user preferences and install the skills on it. But also, if you select "Install" in the general settings, it will install it on every existing site. The `Remove` option works in the same way.
> 
> I left this as is for now but happy to make further adjustments. What do you think @sejas ?

It's a valid point. I think this behaviour is correct. We need to set the correct expectations in the first paragraph. So I suggest updating `Agents can decide to use skills to help them accomplish specialized tasks.` to something like `Agents can decide to use skills to help them accomplish specialized tasks. Select the skills that will be placed in all existing and new sites`.

Maybe we can also add a tooltip to the Install remove buttons.

sejas (commented):
What do you think if we enable local skills too? Like studio-cli.
I suggest removing this exclusion:
https://github.com/Automattic/studio/blob/bb6803aa1dd834a50f6bdd7ba998403499822c5d/apps/studio/src/modules/user-settings/components/skills-tab.tsx#L65-L67

apps/studio/src/ipc-handlers.ts
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2880/changes/e04bcd83a63abdd2514067ff3c9fad899ca584c3

apps/studio/src/ipc-handlers.ts
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/2880/changes/e04bcd83a63abdd2514067ff3c9fad899ca584c3

apps/studio/src/ipc-handlers.ts
katinthehatsite: Made the changes in https://github.com/Automattic/studio/pull/2880/changes/031233f457f3d52ad4fc5a4d5b4b60e61aa2d9e3

katinthehatsite:
> It's a valid point. I think this behaviour is correct. We need to set the correct expectations in the first paragraph. So I suggest updating Agents can decide to use skills to help them accomplish specialized tasks. to something like Agents can decide to use skills to help them accomplish specialized tasks. Select the skills that will be placed in all existing and new sites.

I adjusted the description. I think it is self-explanatory 👍  I briefly thought about adding the tooltip but I did not find the tooltip in the modal looking nice, especially since the `Remove` button is hidden in the toggle.

katinthehatsite:
> I suggest removing this exclusion:

This was addressed in https://github.com/Automattic/studio/pull/2880/changes/89198ffac5eb3d708334eceb341b631f766bf5ac

katinthehatsite:
@sejas @epeicher this should now be ready for another review

katinthehatsite:
Thanks for fixing the linter @sejas - I think my linter broke and I don't see these errors for some reason in my IDE. I will investigate tomorrow.

sejas (approved):
Kat!, Thanks for updating the PR. Great work!
We were checking whether a skill is installed by verifying that it exists on every site, but I think we should check the setting saved in the app data JSON instead. My reasoning is that the settings modal should display what will be installed on new sites.

I pushed that little change in this commit: https://github.com/Automattic/studio/pull/2880/changes/5b23d03c18c1a2532bf5977ab18c53883e574f81

**Before**

https://github.com/user-attachments/assets/a379602a-6236-40b0-9157-e55850bc6c87

**After**

https://github.com/user-attachments/assets/388654f8-3c51-4841-b9a9-55b718129170


We can create a follow-up to move the installation logic into CLI as discussed in [linear](https://linear.app/a8c/issue/STU-1393).
I also noticed that after removing the last skill, we leave the parent folders empty. It would be nice to clean those empty folders.

PR #2882, opened 2026-03-24
Update dev release version script to cover release process change

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

For debugging.

## Proposed Changes

- I propose updating the dev release version script to cover the release process change. Now, tags are created on the release branch, and the old command only catches those in the trunk.

## Testing Instructions

- Confirm builds go through.

## Pre-merge Checklist

### DIFF
--- a/scripts/lib/git-utils.mjs
+++ b/scripts/lib/git-utils.mjs
@@ -11,7 +11,7 @@ export const getLatestTag = () => {
 
 		// Get the latest reachable release tag that is not a beta tag
 		const latestReleaseTag = child_process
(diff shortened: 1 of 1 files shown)


sejas (approved):
Great catch! Thanks for creating the PR.
I confirm the previous command was returning 1.7.4 while the new command returns the latest 1.7.6


Image

The issue was that  git describe only looks at tags that are ancestors of HEAD, which in this case meant `trunk`. Now tags are created on release/branch instead of trunk.

PR #2883, opened 2026-03-24
Fix TUI crash when terminal is resized to narrow width

epeicher:
## Related issues

- Fixes STU-1456

## How AI was used in this PR

AI was used to investigate the crash root cause (tracing the pi-tui render pipeline and width validation) and implement the fix. I have reviewed the changes and tested manually.

## Proposed Changes

- Use `Math.max(0, width - 2)` for border repeat counts to prevent negative values in `String.prototype.repeat()`
- Truncate all lines returned by `PromptEditor.render()` to the given `width` using pi-tui's `truncateToWidth()`, ensuring no line exceeds the terminal width
- Skip site name label rendering when the border is too narrow to fit it

## Testing Instructions

- Open a terminal window and resize it to be very narrow, you can use the command `tput cols` to check the number of columns, please ensure you see less than `53`.
- Using `trunk`
- Run `npm run cli:buildnode apps/cli/dist/cli/main.js ai`
- Check that the TUI crashes with an error
- Checkout this branch and run `npm run cli:buildnode apps/cli/dist/cli/main.js ai`
- Verify the TUI does not crash — it should gracefully truncate content instead
- Resize back to normal width and verify it still renders correctly

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/ui.ts
+++ b/apps/cli/ai/ui.ts
@@ -19,6 +19,7 @@ import {
 	type EditorOptions,
 	type MarkdownTheme,
 	visibleWidth,
+	truncateToWidth,
 } from '@mariozechner/pi-tui';
 import { readAuthToken } from '@studio/common/lib/shared-config';
 import chalk from 'chalk';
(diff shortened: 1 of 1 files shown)


ivan-ottinger (approved):
Nice fix! The changes look good to me and I can confirm they resolve the issue. 

Before:

Image

After:

Image

I went to 32 columns. The error is no longer triggered.

sejas (commented):
Thank you Roberto! The solution looks great.

PR #2888, opened 2026-03-24
Improve CLI config migration prompt

fredrikekelund:
## Related issues

- Related to p1774349350957349-slack-C06DRMD6VPZ

## How AI was used in this PR

To help refine the natural language phrasing and to generate initial implementations for the "is installed" checks.

## Proposed Changes

If the user is trying the npm version of the CLI and has old Studio config files on their computer but Studio isn't installed, this PR changes the CLI's behavior to offer to reset the config by renaming `appdata-v1.json` to `appdata-v1.deprecated.json`. This makes it so this "migration" gate won't get tripped up the next time the CLI launches. 

## Testing Instructions

I'll test this more thoroughly, too, but start by reviewing the phrasing and the logic.

To test the behavior, do `npm run cli:build` and `node apps/cli/dist/cli/main.js site create`.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/lib/dependency-management/setup.ts
+++ b/apps/cli/lib/dependency-management/setup.ts
@@ -66,7 +66,6 @@ async function copyBundledSqlite() {
 	const isBundledVersionNewer =
 		installedSqliteVersion && semver.gt( bundledSqliteVersion, installedSqliteVersion );
 	if ( ! isSqliteInstalled || isBundledVersionNewer ) {
-		console.log( `Copying bundled SQLite version ${ bundledSqliteVersion }…` );
 		await recursiveCopyDirectory( bundledSqlitePath, getSqlitePluginPath() );
 	}
 }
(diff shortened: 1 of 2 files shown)


bcotrim (approved):
I haven't validated the installed conditions in macOS and Windows, but the logic seems solid.
It provides our users with some options without trying to fix for everything possible combinations
👍

apps/cli/lib/dependency-management/setup.ts
```diff
@@ -145,13 +144,13 @@ export async function setupServerFiles() {
...
-			} )
-		)
-	);
+	for ( const [ name, step ] of steps ) {
+		try {
+			await step();
+		} catch ( error ) {
+			console.error( `Failed to set up dependency ${ name }:`, error );
+		}
+	}
```
fredrikekelund, line 153: In my testing, I found that parallel setup functions led to problems. These can probably be fixed, but I chose the simple fix of reverting to a sequential flow for now.

PR #2894, opened 2026-03-24
Fix Appearance picker focus and selection states

shaunandrews:
The recent addition of Appearance settings for light/dark mode introduced new UI in our app settings. I noticed that the focus state is a little awkward.

Here's how it looks with Light selected and in focus:

image

And here's with Light selected and Dark in focus:

image

This PR aims to unify the ring for both selection and focus states, with the color and font-weight to help distinguish between selected and focused.

Selected, and focused:
image

Selected, but unfocused:
image

Light mode, both selected (Light) and focused (Dark):
image

## How AI was used in this PR

Claude helped implement the CSS changes. All changes were reviewed and iterated on interactively.

## Proposed Changes

- Move focus ring from the button container to the illustration, matching the selection ring placement
- Use `frame-text` color for the selection ring to distinguish it from the blue focus ring (`frame-theme`)
- Bold the label text when selected instead of changing its color
- Turn label text blue (`frame-theme`) on focus
- Use `group-focus` instead of `group-focus-visible` so clicking an option shows both focus and selection states (consistent with standard input behavior)

## Testing Instructions

- Open Settings → Appearance
- Click each theme option — should see a blue ring on the illustration and blue label (focus) plus a text-colored ring and bold label (selected)
- Tab between options with keyboard — focus ring (blue) should appear on the illustration, not the outer container
- Verify in both light and dark modes


sejas (approved):
I tested the selector states and they look great. Thanks!

https://github.com/user-attachments/assets/dc41b6d7-1b63-4615-9b72-473164de0572

katinthehatsite (commented):
The changes look good to me 👍 One thing is that I am finding the selected text slightly too bold e.g Light, Dark etc.

Image

Do you think the boldness could be slightly adjusted?

shaunandrews:
Nice catch! I agree. We should be using medium instead of bold.

image

Pushing a fix.

katinthehatsite (approved):
Thanks for changing the labels, I think it looks better now 😍 

I will retrigger the failing tests as they seem to be unrelated.

PR #2897, opened 2026-03-25
Remove noisy Sentry captureException calls from already-handled errors

wojtekn:
## Related issues

- Related to the Sentry cleanup discussion in `docs/p2-sentry-cleanup.md`

## How AI was used in this PR

I used AI to find and analyze all occurrences. I reviewed all code changes.

## Proposed Changes

I propose removing 11 `captureException` calls that were inside `try/catch` blocks where the error was already handled - the user sees a message, or the app degrades gracefully. These inflate Sentry error counts and make it harder to spot real issues. Also, those may be considered as not a legitimate use of Sentry logging.

Removed calls:
- `auth-provider.tsx` - `handleInvalidToken`, `logout` token revocation, `logout` cleanup (token flows recover gracefully)
- `use-delete-site.ts` - user already sees an error message box
- `use-feature-flags.tsx` - graceful degradation with defaults
- `windows-installation-manager.ts` × 4 - all errors surfaced to the user via dialog
- `open-about-menu.ts` - non-critical DOM operation
- `site-server.ts` - silent thumbnail capture failure (silent by design)
- `updates.ts` - expected read-only volume path in `showReadOnlyVolumeError`.

Also removes 4 now-unused Sentry imports.

## Testing Instructions

- Test auth, deleting a site, CLI settings toggle

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/updates.ts
+++ b/apps/studio/src/updates.ts
@@ -233,8 +233,6 @@ async function showReadOnlyVolumeError( err: Error ) {
 		);
 		detailPath = sprintf( __( 'Studio is running from: %s' ), app.getPath( 'exe' ) );
 
-		// this case is not expected, so we want to capture it
-		Sentry.captureException( err );
 		console.error( err );
 	}
 
--- a/apps/studio/src/about-menu/open-about-menu.ts
+++ b/apps/studio/src/about-menu/open-about-menu.ts
@@ -1,6 +1,5 @@
 import { BrowserWindow, app } from 'electron';
 import path from 'path';
-import * as Sentry from '@sentry/electron/renderer';
 import { __ } from '@wordpress/i18n';
 import { ABOUT_WINDOW_HEIGHT, ABOUT_WINDOW_WIDTH } from 'src/constants';
 import { shellOpenExternalWrapper } from 'src/lib/shell-open-external-wrapper';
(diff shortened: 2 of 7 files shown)


apps/studio/src/updates.ts
```diff
@@ -233,8 +233,6 @@ async function showReadOnlyVolumeError( err: Error ) {
 		);
 		detailPath = sprintf( __( 'Studio is running from: %s' ), app.getPath( 'exe' ) );
 
-		// this case is not expected, so we want to capture it
-		Sentry.captureException( err );
```
katinthehatsite, line 237: If Studio is installed correctly in Applications but a read-only volume error still fires, would we want to capture it?

katinthehatsite (approved):
The changes look good 👍  I noticed that there are some failing e2e tests that would need to be resolved before merging.

I also left one question to consider

apps/studio/src/updates.ts
wojtekn: Probably not, we won't be able to help much if the user doesn't reach out to us.

wojtekn:
> The changes look good 👍 I noticed that there are some failing e2e tests that would need to be resolved before merging.

It was a one-off issue. I rerun tests, and they succeeded.

PR #2900, opened 2026-03-25
Disable Redis/memcached when JSPI is unavailable

ivan-ottinger:
## Related issues

- STU-1366

## How AI was used in this PR

Claude Code was used to investigate, reproduce, and fix the bug. I have reviewed the code, iterated over it and confirm it works as expected.

## Proposed Changes

- Conditionally disable Redis and memcached PHP extensions when JSPI is unavailable (Node < 24) in `wordpress-server-child.ts` and `run-wp-cli-command.ts`

**Context:** The Redis and memcached extensions require JSPI (JavaScript Promise Integration), only available in Node.js 24+. On Node 22, enabling them crashes the WordPress server at startup:

> Error: The Redis extension requires JSPI (JavaScript Promise Integration) support.

The conditional disable uses the same `semver.gte(process.version, '24.0.0')` pattern already established in `process-manager-daemon.ts` for the `--experimental-wasm-jspi` flag.

The engine bump reflects the de facto requirement: the production build strips asyncify binaries (`vite.config.prod.ts`), the `.nvmrc` specifies Node 24.13.1, and Node 24 is the current Active LTS.

## Testing Instructions

1. On Node 22: `npm run cli:build && node apps/cli/dist/cli/main.js site create`
2. Run `cat $(ls -t ~/.studio/pm2/logs/*-out.log | head -1)` to see the logs and check:
  - Before fix: crash with Redis JSPI error is recorded
  
CleanShot 2026-03-25 at 12 20 48@2x
  
  - After fix: server starts and `Extensions` log line includes `intl` only (no `redis`/`memcached` mentioned), site works normally

## Pre-merge Checklist

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -16,7 +16,7 @@
 		"scripts/"
 	],
 	"engines": {
-		"node": ">=22.0.0"
+		"node": ">=24.0.0"
 	},
 	"repository": {
 		"type": "git",
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -16,7 +16,7 @@
 		"scripts/"
 	],
 	"engines": {
-		"node": ">=24.0.0"
+		"node": ">=22.0.0"
 	},
 	"repository": {
 		"type": "git",
--- a/apps/cli/lib/run-wp-cli-command.ts
+++ b/apps/cli/lib/run-wp-cli-command.ts
@@ -12,6 +12,7 @@ import { cleanupLegacyMuPlugins, getMuPlugins } from '@studio/common/lib/mu-plug
 import { LatestSupportedPHPVersion } from '@studio/common/types/php-versions';
 import { __ } from '@wordpress/i18n';
 import { setupPlatformLevelMuPlugins } from '@wp-playground/wordpress';
+import semver from 'semver';
 import { getSqliteCommandPath, getWpCliPharPath } from 'cli/lib/server-files';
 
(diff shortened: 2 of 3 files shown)


apps/cli/package.json
fredrikekelund, line 1: I would keep this file as is since the code can handle Node.js 22 now. I know the `engines` field isn't "enforced" by Node, but if we're going to show a warning on Node.js 22, then it'd be better to display something custom with details about why users should upgrade.

fredrikekelund (approved):
Good idea 👍 

I think we should revert `apps/cli/package.json`, but I'll still approve now to unblock you, @ivan-ottinger

apps/cli/package.json
ivan-ottinger: Sounds good. 👍🏼 I have reverted that change in https://github.com/Automattic/studio/pull/2900/commits/a75b78f82198bc0aa8328fa9bcc5e953cf560128.

wojtekn:
Node 22 is in maintenance mode and will be supported for a few years, and Node 24 is under active support. Did we consider declaring 24 as a hard requirement? Playground may drop asyncify support at some point, see https://github.com/WordPress/wordpress-playground/pull/3281

If we want to keep this guard, should Studio CLI inform a user about Redis and Memcached not available on their current stack?

fredrikekelund:
@wojtekn, that's a good question. Let's continue that discussion on Slack or on Linear.

PR #2901, opened 2026-03-25
Add wpcom.user.id to Sentry errors

nightnei:
## Related issues

- Fixes STU-1463

## How AI was used in this PR
AI generated code, we had many iterations, and ultimatelly partially it was written by me manually to polish it.

## Proposed Changes
I had a case when I knew the wpcom userId, but I didn't know the sentryUserId. I asked the user about sharing it with me, but they messed it up by removing appdatav1.json and not reloading Studio.

So if we add wpcom userId to Sentry then:
1. We don't need to spend time and bother users with sharing it
2. We prevent cases when users modify/remove appdatav1.json

## Testing Instructions
1. Log out from Studio
2. Apply this diff:
```diff
diff --git a/apps/studio/src/components/top-bar.tsx b/apps/studio/src/components/top-bar.tsx
index 28a1152df..925d0cbad 100644
--- a/apps/studio/src/components/top-bar.tsx
+++ b/apps/studio/src/components/top-bar.tsx
@@ -1,3 +1,4 @@
+import * as Sentry from '@sentry/electron/renderer';
 import { sprintf } from '@wordpress/i18n';
 import { Icon, help, drawerLeft, cog } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
@@ -111,7 +112,11 @@ function SettingsButton() {
        const { __ } = useI18n();
        return (
                <Button
-                       onClick={ () => getIpcApi().showUserSettings( 'general' ) }
+                       onClick={ () => {
+                               Sentry.captureException( new Error( 'Test renderer error: wpcom.user.id tag' ) );
+                               getIpcApi().showUserSettings( 'general' );
+                       } }
                        aria-label={ __( 'Open settings' ) }
                        variant="icon"
                        data-testid="settings-button"
diff --git a/apps/studio/src/index.ts b/apps/studio/src/index.ts
index 3356a2800..9f852f9b5 100644
--- a/apps/studio/src/index.ts
+++ b/apps/studio/src/index.ts
@@ -70,13 +70,16 @@ function getRendererUrl(): string {
        }
 }
 
-if ( ! process.env.IS_DEV_BUILD ) {
        const { sentryRelease, isDevEnvironment } = getSentryReleaseInfo( app.getVersion() );
 
        Sentry.init( {
                dsn: 'https://97693275b2716fb95048c6d12f4318cf@o248881.ingest.sentry.io/4506612776501248',
                debug: true,
-               enabled: ! isDevEnvironment,
+               enabled: true,
                release: sentryRelease,
                environment: isDevEnvironment ? 'development' : 'production',
        } );
-}
diff --git a/apps/studio/src/lib/main-sentry-utils.ts b/apps/studio/src/lib/main-sentry-utils.ts
index 4a5a0b3f1..a93209286 100644
--- a/apps/studio/src/lib/main-sentry-utils.ts
+++ b/apps/studio/src/lib/main-sentry-utils.ts
@@ -1,5 +1,6 @@
 import * as Sentry from '@sentry/electron/main';
  
export function setSentryWpcomUserIdMain( id: number | undefined ) {
+       console.log(id);console.log(id);console.log(id);console.log(id);console.log(id);
        Sentry.setTag( 'wpcom.user.id', id );
 }
diff --git a/apps/studio/src/lib/renderer-sentry-utils.ts b/apps/studio/src/lib/renderer-sentry-utils.ts
index cfcb5b9ac..5b3a9934b 100644
--- a/apps/studio/src/lib/renderer-sentry-utils.ts
+++ b/apps/studio/src/lib/renderer-sentry-utils.ts
@@ -1,5 +1,6 @@
 import * as Sentry from '@sentry/electron/renderer';
 
export function setSentryWpcomUserIdRenderer( id: number | undefined ) {
+       console.log(id);console.log(id);console.log(id);console.log(id);console.log(id);
        Sentry.setTag( 'wpcom.user.id', id );
 }
diff --git a/apps/studio/src/modules/user-settings/lib/ipc-handlers.ts b/apps/studio/src/modules/user-settings/lib/ipc-handlers.ts
index 914a6f0ac..2e3fb8ca7 100644
--- a/apps/studio/src/modules/user-settings/lib/ipc-handlers.ts
+++ b/apps/studio/src/modules/user-settings/lib/ipc-handlers.ts
@@ -1,4 +1,5 @@
 import { BrowserWindow, IpcMainInvokeEvent, nativeTheme } from 'electron';
+import * as Sentry from '@sentry/electron/main';
 import { updateSharedConfig } from '@studio/common/lib/shared-config';
 import { DEFAULT_TERMINAL } from 'src/constants';
 import { sendIpcEventToRenderer, sendIpcEventToRendererWithWindow } from 'src/ipc-utils';
@@ -91,6 +92,8 @@ export async function getColorScheme(): Promise< 'system' | 'light' | 'dark' > {
 }
 
 export function showUserSettings( event: IpcMainInvokeEvent, tabName?: UserSettingsTabName ) {
+       Sentry.captureException( new Error( 'Test main process error: wpcom.user.id tag' ) );
        const parentWindow = BrowserWindow.fromWebContents( event.sender );
        sendIpcEventToRendererWithWindow( parentWindow, 'user-settings', { tabName } );
 }
```
3. `npm start`
3. Click on Setting icon in Studio
4. Open Sentry
5. Search for "wpcom.user.id" text
6. Assert that you see two events from main and renderer processes and both of them **don't include** tag "wpcom.user.id" 
7. Login to wpcom in Studio
8. Assert that you see console.log (5) in your browser and logs from backend (main) process in your terminal
9. Click on Settings icon again
10. Open Sentry
11. Assert that you see an error from the main process with wpcom.user.id tag as your wpcom user id (IDK why Studio doesn't send events from the renderer, maybe there is some filter to not send a few of the same events in a row, but if you reload Studio, then it will also send events from the renderer process)


apps/studio/src/lib/renderer-sentry-utils.ts
```diff
@@ -0,0 +1,5 @@
+import * as Sentry from '@sentry/electron/renderer';
+
+export function setSentryWpcomUserIdRenderer( id: number | undefined ) {
```
nightnei, line 3: I intentionally added postfix as `...Renderer` and `...Main` to simplify code reading, since sometimes it feels confusing and not clear - is it error from main or from renderer, so easy to make a mistake.

apps/studio/src/lib/renderer-sentry-utils.ts
```diff
@@ -0,0 +1,5 @@
...
+	Sentry.setTag( 'wpcom.user.id', id );
```
nightnei, line 4: `wpcom.user.id` - I see that Sentry uses dot as a delimiter for tags, that's why I also went this way.

nightnei:
We got an approval to proceed with this - p4H3ND-22K-p2

sejas (commented):
The code looks good to me, and I can confirm that Sentry captures the WP.com user ID for errors on the main thread. For the renderer, I only saw the first trigger and not the subsequent ones. See the number of reported issues in the screenshot.

Image

I tested this PR following your testing instructions and applying your diff.
I started with an authenticated user, then I logged out and the main process was still showing my existing userId. When I authenticated with another user I saw the correct  userId. So, from my point of view is good enough.

My only concern is about the renderer not sending the following errors. Am I doing anything wrong? Maybe is an existing issue/feature of Sentry? 🤔

nightnei:
> My only concern is about the renderer not sending the following errors. Am I doing anything wrong? Maybe is an existing issue/feature of Sentry? 🤔

@sejas Thanks, yeah, I also noticed it and mentioned in the description `IDK why Studio doesn't send events from the renderer, maybe there is some filter to not send a few of the same events in a row, but if you reload Studio, then it will also send events from the renderer process`

And decided to look deeper and indeed - Sentry ships with a built-in [deduplication flag](https://docs.sentry.io/platforms/javascript/configuration/integrations/dedupe/), that is enabled by default. It suppresses consecutive events that have the same exception message and stack trace. I think it makes sense, to prevent flooding Sentry with duplicates. 
Why doesn't it work the same way for the main process - it's a good quetion 😁 I haven't found any info in docs, but I debugged node_modules with AI -- the `main` doesn't have deduplication included in the configuration, but the renderer has.
Screenshot 2026-03-30 at 16 20 05
Screenshot 2026-03-30 at 16 20 23

Anyway, it's existing, default Sentry SDK behavior and we already have it in Studio, so changes in this PR don't change it.

sejas (approved):
Makes sense. Thanks for clarifying, Vova!

PR #2903, opened 2026-03-25
Fix regressions around deleting preview sites

fredrikekelund:
## Related issues

- Fixes STU-1459

## How AI was used in this PR

Only to update `delete.test.ts`.

## Proposed Changes

https://github.com/Automattic/studio/pull/2807 introduced two regressions:

- The "Delete all preview sites" button in the settings modal stopped working. 
- The "Clear" button on deleted preview site rows didn't persist the chance to the config file. 

The app no longer has direct access to the config file where the snapshots live (`cli.json`), so the old strategy for the "Delete all preview sites" button of simply emptying the `snapshots` array in the config file no longer works.

I considered iterating over the snapshots and deleting them from `snapshot-slice.ts` with a "bulk operation". That's how we used to do it before https://github.com/Automattic/studio/pull/1950/, but that PR improved on what we had, so ideally we want to keep that behavior.

In the end, my solution was to:

1. Add an `--all` option to the `preview delete` CLI command. 
2. Adjust the app to spawn a `preview delete --all` command when the user clicks the "Delete all preview sites" button.
3. To fix the "Clear" button issue, I made it so each click triggers a `preview delete <url>` command. I added an `optimistic` prop to those actions so the sites disappear immediately from the UI, without the progress bar we normally show for delete operations. 

## Testing Instructions

1. `npm start`
2. Create a preview site
3. Open the settings modal and go to the Account tab
4. Delete all preview sites
5. Ensure that it works as expected
6. Create another preview site
7. Open `~/.studio/cli.json`
8. Edit the `date` field for the snapshot. Change it to `Date.now() - 604800000 * 2` (two weeks ago)
9. Reload the Studio UI
10. Go to the Previews tab
11. Ensure that the Clear button is visible and that it works as expected
12. Reload the Studio UI
13. Ensure that the clear action was persisted and that the preview site is no longer there

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/ipc-utils.ts
+++ b/apps/studio/src/ipc-utils.ts
@@ -33,7 +33,7 @@ export interface IpcEvents {
 	'on-site-create-progress': [ { siteId: string; message: string } ];
 	'site-context-menu-action': [ { action: string; siteId: string } ];
 	'site-event': [ SiteEvent ];
-	'snapshot-changed': [ SnapshotEvent ];
+	'snapshot-event': [ SnapshotEvent ];
 	'sync-upload-network-paused': [ { error: string; selectedSiteId: string; remoteSiteId: number } ];
 	'sync-upload-resumed': [ { selectedSiteId: string; remoteSiteId: number } ];
 	'sync-upload-progress': [ { selectedSiteId: string; remoteSiteId: number; progress: number } ];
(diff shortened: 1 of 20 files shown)


apps/studio/src/ipc-utils.ts
```diff
@@ -33,7 +33,7 @@ export interface IpcEvents {
 	'on-site-create-progress': [ { siteId: string; message: string } ];
 	'site-context-menu-action': [ { action: string; siteId: string } ];
 	'site-event': [ SiteEvent ];
-	'snapshot-changed': [ SnapshotEvent ];
+	'snapshot-event': [ SnapshotEvent ];
```
fredrikekelund, line 36: Nitpicky. It's just that this event no longer always captures the details for a single snapshot, so I found this to be a better name.

bcotrim (approved):
Works as described
Code LGTM and nice improvements 👍

PR #2908, opened 2026-03-25
Remove cost display from AI command output

youknowriad:
## Related issues

Closes STU-1440

## How AI was used in this PR

AI-assisted implementation. Changes reviewed manually.

## Proposed Changes

- Remove the USD cost (`$X.XXXX`) from the AI command's success and max-turns status messages
- Clean up the now-unused `costUsd` field from types and return values

## Testing Instructions

- Run `studio ai` and complete a prompt
- Verify the status line shows `Thought for Xs · N turns` without a cost suffix
- Trigger the max-turns limit and verify it shows `Used N turns` without cost

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/ui.ts
+++ b/apps/cli/ai/ui.ts
@@ -1810,7 +1810,7 @@ export class AiChatUI {
 		message: SDKMessage
 	):
 		| { sessionId: string; success: boolean; maxTurnsReached?: undefined }
-		| { sessionId: string; maxTurnsReached: true; numTurns: number; costUsd: number }
+		| { sessionId: string; maxTurnsReached: true; numTurns: number }
 		| undefined {
 		switch ( message.type ) {
 			case 'assistant': {
@@ -1906,11 +1906,7 @@ export class AiChatUI {
 					if ( ! this.hasShownResponseMarker ) {
 						this.messages.addChild( new Text( '\n ' + chalk.blue( '⏺' ) + ' Done', 0, 0 ) );
 					}
-					this.showInfo(
-						`Thought for ${ thinkingSec }s · ${
-							message.num_turns
-						} turns · $${ message.total_cost_usd.toFixed( 4 ) }`
-					);
+					this.showInfo( `Thought for ${ thinkingSec }s · ${ message.num_turns } turns` );
 					return { sessionId: message.session_id, success: true };
 				}
 
@@ -1934,7 +1930,6 @@ export class AiChatUI {
 						sessionId: message.session_id,
 						maxTurnsReached: true,
 						numTurns: message.num_turns,
-						costUsd: message.total_cost_usd,
 					};
 				} else if ( message.subtype ) {
 					parts.push( `(${ message.subtype })` );
--- a/apps/cli/ai/ui.ts
+++ b/apps/cli/ai/ui.ts
@@ -22,6 +22,7 @@ import {
 	truncateToWidth,
 } from '@mariozechner/pi-tui';
 import { readAuthToken } from '@studio/common/lib/shared-config';
+import { _n, sprintf } from '@wordpress/i18n';
 import chalk from 'chalk';
 import { AI_MODELS, DEFAULT_MODEL, type AiModelId, type AskUserQuestion } from 'cli/ai/agent';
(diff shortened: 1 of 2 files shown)


apps/cli/ai/ui.ts
```diff
@@ -1906,11 +1906,7 @@ export class AiChatUI {
 					if ( ! this.hasShownResponseMarker ) {
 						this.messages.addChild( new Text( '\n ' + chalk.blue( '⏺' ) + ' Done', 0, 0 ) );
 					}
-					this.showInfo(
-						`Thought for ${ thinkingSec }s · ${
-							message.num_turns
-						} turns · $${ message.total_cost_usd.toFixed( 4 ) }`
-					);
+					this.showInfo( `Thought for ${ thinkingSec }s · ${ message.num_turns } turns` );
```
sejas: I observed we were displaying an incorrect number concordance, so `1 turns` was in plural. I fixed it in https://github.com/Automattic/studio/pull/2908/changes/f6919ee759c3c6c8c6ae97f2fc9f07d28c393252

**Before**

Image

**After**

Screenshot 2026-03-25 at 18 18 43

apps/cli/commands/ai/index.ts
```diff
@@ -327,9 +326,7 @@ export async function runCommand(
 		}
 
 		if ( maxTurnsResult ) {
-			ui.showInfo(
-				`Used ${ maxTurnsResult.numTurns } turns · $${ maxTurnsResult.costUsd.toFixed( 4 ) }`
-			);
+			ui.showInfo( `Used ${ maxTurnsResult.numTurns } turns` );
```
sejas: Same here

sejas (approved):
Cool! Thanks for removing the cost.
I only found a small issue where we displayed the incorrect number for the word `turns`, which led me to wonder whether we need to translate Studio AI TUI, which I think we should so I [pushed a commit in this PR](https://github.com/Automattic/studio/pull/2908/changes/f6919ee759c3c6c8c6ae97f2fc9f07d28c393252) and created a Linear issue to translate other strings STU-1465.

I tested this PR by running `npm run cli:build && node apps/cli/dist/cli/main.js ai`, and asked a few questions.

Image

Great work!

youknowriad:
AFAIK the TUI (or the CLI) is not translated at all but I could be wrong.

sejas:
> AFAIK the TUI (or the CLI) is not translated at all but I could be wrong.

The CLI is translated using the Studio locale that is saved in shared.json. The Studio AI is also using `__`. Here is an [example](https://github.com/Automattic/studio/blob/eff08bcaeb3c94438dd9d9530957d3d36756adf1/apps/cli/commands/ai/index.ts#L484).
`_n` is really handy when the word changes based on the numbers in the string.

youknowriad:
I think @fredrikekelund mentioned that the locale is not really used in the CLI recently.

youknowriad:
Not a blocker here but I think we need more clarity on the translation approach (or not, seems ok to not translate a CLI).

sejas:
Agreed! We should decide whether we want to translate it or not. As far as I know, we were translating everything we could in CLI too.

i18n also provides some useful functions we can leverage instead of creating our own solutions. For this specific example it would be great if we can avoid showing `1 turns`.

youknowriad:
> we were translating everything we could in CLI too

how do the translations get into the CLI?

sejas:
They are loaded using `loadTranslations` ([code](<https://github.com/Automattic/studio/blob/415d8b0f054f9d851c24fcfb05f81ff67f689ca6/apps/cli/index.ts#L17>)). They were introduced in this PR [https://github.com/Automattic/studio/pull/1187](<https://github.com/Automattic/studio/pull/1187>) , and it seems they are still working ✨

https://github.com/user-attachments/assets/360eaeff-59b4-4579-b6d4-997a153777f1

youknowriad:
@sejas Thanks, we may want to do an audit for all the messages used in the ai command than haha :)

sejas:
I'll do the audit and prepare a PR as part of STU-1465 . Thanks for your help @youknowriad  !

PR #2909, opened 2026-03-25
Update app path on Linux docs

matiasbenedetto:
## What?                                                    
Correct the output path in docs/linux.md and docs/code-contributions.md from `out/Studio-linux-x64/` to `apps/studio/out/Studio-linux-x64/`.
                                                                                                                                                                                                                                              
## Why?                                                      
The docs pointed to a non-existent `out/` folder at the repo root. The package-in-isolation script copies build artifacts to  `apps/studio/out/`, not the project root, so users following the docs would land in the wrong directory.                                                                                                                                                                                                                     
                                                            
## How?                                                                                                                                                                                                                                        
Updated all references to the output path in both docs files to reflect the correct `apps/studio/out/Studio-linux-x64/ `location.

### DIFF
--- a/docs/code-contributions.md
+++ b/docs/code-contributions.md
@@ -214,7 +214,7 @@ npm install
 npm run package
 ```
 
-After building, the executable will be located at `out/Studio-linux-x64/studio`.
+After building, the executable will be located at `apps/studio/out/Studio-linux-x64/studio`.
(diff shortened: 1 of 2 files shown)


bcotrim (approved):
LGTM 👍

PR #2910, opened 2026-03-25
Remove redundant block validation from wp_cli tool

youknowriad:
## Related issues

- Related to AI command block validation flow

## How AI was used in this PR

Claude identified the redundancy between automatic validation in `wp_cli` and the system prompt-driven `validate_blocks` flow, then removed the duplicate code.

## Proposed Changes

- Removed automatic block validation that ran inside the `wp_cli` tool before every `post create`/`post update` command
- Removed the helper functions (`extractPostContent`, `isPostContentCommand`, `BLOCK_COMMENT_PATTERN`) that were only used by this validation
- The system prompt already instructs the AI to call `validate_blocks` explicitly, making the automatic validation redundant and causing double validation in the happy path

## Testing Instructions

- Run the AI command to create a site and verify that block validation still happens via the `validate_blocks` tool (driven by the system prompt)
- Confirm that `wp post create` / `wp post update` commands execute without the automatic pre-validation step

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/tools.ts
+++ b/apps/cli/ai/tools.ts
@@ -433,31 +433,10 @@ const deletePreviewTool = tool(
 	}
 );
 
-const BLOCK_COMMENT_PATTERN = /<!-- wp:/;
-
-function extractPostContent( command: string ): string | undefined {
-	const args = splitCommandArgs( command );
-	for ( const arg of args ) {
-		if ( arg.startsWith( '--post_content=' ) ) {
-			return arg.slice( '--post_content='.length );
-		}
-	}
-	return undefined;
-}
-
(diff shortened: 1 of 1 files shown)


sejas (approved):
Thanks for finding and removing this duplication. I tried a couple of prompts and confirmed the LLM followed the [instructions](https://github.com/Automattic/studio/blob/4fc130401972d286702c83e35a780eae8123e3ce/apps/cli/ai/system-prompt.ts#L9) and run the `validate_blocks` as a separate step.

| Prompt | Validation | Result |
|--------|--------|--------|
| Image | Image | Image |

PR #2911, opened 2026-03-25
Add Studio user id to Studio About window

nightnei:
## Related issues

- Fixes STU-1464

## Proposed Changes
Right now we need to ask user to find appdata-v1.json and find their sentryUserId to share with us for debugging. With the new Studio version it will be even more complicated since it will depend on the Studio version, either appdata-v1.json or app.json. Especially, it complicates things for less tech users and in general provides not the best experience.

I propose to add it to the About dialog, alongside Studio version:
Screenshot 2026-03-25 at 17 35 15

## Testing Instructions
1. `npm start`
2. Open about dialog - Click on `Electron` -> `About WordPress Studio`
3. Open /Users/<USERNAME>/.studio/app.json 
4. Assert that `sentryUserId` app.json equals `User ID` inside About dialog

### DIFF
--- a/apps/studio/src/about-menu/about-menu.html
+++ b/apps/studio/src/about-menu/about-menu.html
@@ -42,7 +42,6 @@
 				background: rgba( 246, 246, 246, 0.84 );
 				cursor: default;
 				margin: 0;
-				max-height: 284px;
 				display: flex;
 				justify-content: center;
 				align-items: center;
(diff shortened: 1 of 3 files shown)


wojtekn:
I think it's an incorrect approach to debugging such an error. Sentry is for proactive monitoring - we should detect unhandled crashes before users even report them, without needing them to share anything.

For user-specific debugging, when someone contacts support, we actually need their local error context - like log file, import file they used, remote site ID, information about their environment, and the timing.

These are two separate flows, and displaying the UUID doesn't improve either.

nightnei:
> These are two separate flows, and displaying the UUID doesn't improve either.

I didn't try to cover the first flow. But regarding the second flow - we have a real case (details are shared in "Proposed Changes"). Anyway, I don't have a strong opinion. If you don't find it useful, I am ok to close it.

PR #2916, opened 2026-03-26
Rename `appdata-v1.json` to `appdata-v1.deprecated.json`

fredrikekelund:
## Related issues

- Related to p1774436490427889-slack-C06DRMD6VPZ

## How AI was used in this PR

Not at all. 

## Proposed Changes

Make it so the `02-migrate-to-split-config.ts` migration renames the old `appdata-v1.json` file to `appdata-v1.deprecated.json`. 

The "deprecated" filename is already used by the `00-check-studio-compatibility.ts` CLI migration (see https://github.com/Automattic/studio/pull/2888).

## Testing Instructions

1. Open `~/.studio`
2. Delete or rename the `app.json`, `cli.json`, and `shared.json` files
3. `npm start`
4. `ls /Users/USERNAME/Library/Application\ Support/Studio`
5. Ensure that the directory contains a `appdata-v1.deprecated.json` file and no `appdata-v1.json` file

## Pre-merge Checklist


bcotrim (approved):
CI is failing on unit tests but approving to unblock once it is finished
LGTM 👍 

Nit: we could probably remove the appconfig file check from the CLI since appdata-v1 will no longer exist after migration

```typescript
if ( fs.existsSync( getAppConfigPath() ) ) {
	return;
}
```

wojtekn:
Thanks for the change, it will make the process clearer.

fredrikekelund:
`fs` mocking was giving me grief. I expected Vitest to load the `fs` mock from the `__mocks__` directory, but apparently, Vitest only looks for `__mocks__` directories in the same directory as the test and in the project root (see [docs](https://vitest.dev/api/vi.html#vi-mock)).

So I moved the `__mocks__` directory and changed a bunch of imports from `fs/promises` to `fs`. Bigger diff now, but nothing changed in the runtime behavior.

fredrikekelund:
> we could probably remove the appconfig file check from the CLI since appdata-v1 will no longer exist after migration

We could, but I think the existing logic is still sensible. I'll probably just leave it in place

PR #2924, opened 2026-03-26
Add .deployignore support for Preview Sites

ivan-ottinger:
## Related issues

- Related to STU-16

## How AI was used in this PR

AI-assisted implementation with human review of all changes, architecture decisions, and edge cases. Step-by-step iteration. I have tested the PR as well.

## Proposed Changes

- Add a shared `createDeployIgnoreFilter` utility (`tools/common/lib/deploy-ignore.ts`) that reads a `.deployignore` file from the site root and creates a filter using the [`ignore`](https://www.npmjs.com/package/ignore) npm package (gitignore pattern syntax)
  - the `ignore` package is widely used and I consider it is better to use it instead of re-inventing the wheel  
- `.git`, `node_modules`, `.DS_Store`, `Thumbs.db`) are excluded by default
- **Preview Sites**: Update `archiveSiteContent` and `validateSiteSize` to use the deploy-ignore filter when archiving `wp-content` for upload
- Remove the old `isExcludedFromArchive` hardcoded filter from `fs-utils.ts`, replacing it with the new deploy-ignore utility. `calculateDirectorySizeForArchive` retains a fallback to the original `.git`/`node_modules` exclusion when called without a filter

**Note:** Site Sync integration with `.deployignore` will be handled in a follow-up PR that also updates the `DefaultExporter` filtering. The shared utility introduced here is designed to be reused for that.

## Testing Instructions

> [!TIP]
> It is probably easier to ask your AI assistant to perform the steps 2 and 3 for you, so you don't have to create the testing files manually.

1. Create a local site in Studio and install the `wordpress-seo` plugin
2. Add a `.deployignore` file at the site root (e.g., `~/Studio/my-site/.deployignore`) with the following content:
   ```
   # Ignore vendor dependencies in my-plugin, except important-lib
   wp-content/plugins/my-plugin/vendor/*
   !wp-content/plugins/my-plugin/vendor/important-lib

   # Ignore all log files
   *.log

   # Ignore old uploads
   wp-content/uploads/2024

   # Ignore a specific plugin
   wp-content/plugins/wordpress-seo
   ```
3. Create matching test files/folders in the site's `wp-content`:
   ```
   # Plugin with build artifacts
   mkdir -p wp-content/plugins/my-plugin/vendor/some-lib
   mkdir -p wp-content/plugins/my-plugin/vendor/important-lib
   echo '<?php // plugin entry' > wp-content/plugins/my-plugin/my-plugin.php
   echo '<?php // vendor file' > wp-content/plugins/my-plugin/vendor/some-lib/autoload.php
   echo '<?php // important' > wp-content/plugins/my-plugin/vendor/important-lib/loader.php
   echo 'build log' > wp-content/plugins/my-plugin/build.log

   # .git and node_modules inside the plugin (built-in defaults)
   mkdir -p wp-content/plugins/my-plugin/.git
   echo 'ref: refs/heads/main' > wp-content/plugins/my-plugin/.git/HEAD
   mkdir -p wp-content/plugins/my-plugin/node_modules/some-package
   echo '{"name":"some-package"}' > wp-content/plugins/my-plugin/node_modules/some-package/package.json

   # Log file at wp-content level
   echo 'error log content' > wp-content/debug.log

   # Old and new uploads
   mkdir -p wp-content/uploads/2024
   echo 'old photo' > wp-content/uploads/2024/photo.jpg
   mkdir -p wp-content/uploads/2025
   echo 'new photo' > wp-content/uploads/2025/photo.jpg
   ```
4. Create/update a Preview Site and verify that matched files are excluded from the upload (you can use the JN tool to get the site SSH details: 3ad28-pb).
   - `vendor/some-lib/` should be excluded
   - `vendor/important-lib/` should be included (negation pattern)
   - `*.log` files should be excluded
   - `uploads/2024/` should be excluded, `uploads/2025/` should be present
   - `wordpress-seo` plugin should be excluded
5. Verify that `.git` and `node_modules` inside `my-plugin` are excluded by built-in defaults (even without `.deployignore` entries for them)
6. Verify that site export (backup) is unaffected by `.deployignore`. Please note that the backup already excludes files/folders like `.git`, `node_modules`, `cache` by default (unrelated to this PR).
7. The preview site should work as expected - without any issues.

## Pre-merge Checklist


tools/common/lib/deploy-ignore.ts
```diff
@@ -0,0 +1,40 @@
...
+	const ig = ignore().add( DEPLOY_IGNORE_DEFAULTS );
+	if ( additionalDefaults ) {
+		ig.add( additionalDefaults );
+	}
+
+	const deployIgnorePath = path.join( rootPath, DEPLOY_IGNORE_FILENAME );
+	try {
+		const content = await fsPromises.readFile( deployIgnorePath, 'utf-8' );
+		ig.add( content );
+	} catch {
```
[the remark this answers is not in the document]
ivan-ottinger: If a user explicitly writes `!node_modules` in their `.deployignore`, they probably mean it. Because of that let's keep things as they are.

tools/common/lib/deploy-ignore.ts
```diff
@@ -0,0 +1,40 @@
...
+		// .deployignore doesn't exist — use defaults only
```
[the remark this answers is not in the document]
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2924/changes/57f0753990d7fb719a3f39d2b0a4f7d20484dfa4.

tools/common/lib/deploy-ignore.ts
```diff
@@ -0,0 +1,36 @@
...
+import ignore, { Ignore } from 'ignore';
+import { isErrnoException } from './is-errno-exception';
+
+export type { Ignore };
+
+/**
+ * Default patterns that are always excluded from deploys,
+ * regardless of whether a .deployignore file exists.
+ */
+const DEPLOY_IGNORE_DEFAULTS = [ '.git', 'node_modules', '.DS_Store', 'Thumbs.db' ];
```
ivan-ottinger: The `.DS_Store` and `Thumbs.db` are a new addition - as I don't think we need those synced to the preview site.

tools/common/lib/deploy-ignore.ts
```diff
@@ -0,0 +1,43 @@
...
```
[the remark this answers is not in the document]
ivan-ottinger: Good catch. I have updated the PR description and comment in the code.

apps/cli/lib/archive.ts
```diff
@@ -11,6 +11,8 @@ export async function archiveSiteContent(
 	siteFolder: string,
 	archivePath: string
 ): Promise< archiver.Archiver > {
+	const deployIgnore = await createDeployIgnoreFilter( siteFolder );
```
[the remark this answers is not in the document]
ivan-ottinger: Tests pass so I will leave things as they are.

tools/common/lib/deploy-ignore.ts
```diff
@@ -0,0 +1,36 @@
+import { promises as fsPromises } from 'fs';
+import path from 'path';
+import ignore, { Ignore } from 'ignore';
+import { isErrnoException } from './is-errno-exception';
+
+export type { Ignore };
```
fredrikekelund:
```suggestion
```
This looks like an AI leftover. Let's remove it.

tools/common/lib/fs-utils.ts
```diff
@@ -1,37 +1,53 @@
...
  * @param directoryPath - The path to the directory to calculate the size of
+ * @param deployIgnore - Optional Ignore instance from createDeployIgnoreFilter
+ * @param pathPrefix - Optional prefix for relative paths when matching against deployIgnore
+ *                     (e.g., 'wp-content' when scanning the wp-content directory)
  * @returns A promise that resolves to the total size in bytes
  */
-export function calculateDirectorySizeForArchive( directoryPath: string ): Promise< number > {
+export function calculateDirectorySizeForArchive(
+	directoryPath: string,
+	deployIgnore?: Ignore,
```
fredrikekelund, line 20: `useSiteSize` calls this function via the `getDirectorySize` IPC function, but it doesn't pass a `deployIgnore` argument. I think it should, or, alternatively, we construct a `deployIgnore` instance directly in this function. In practice, this function isn't meant to be generic anyway – it's meant to calculate the size of directories used for preview sites or site syncs. 

EDIT: Seeing that we define a `DEFAULT_EXCLUDED_DIRECTORIES` array in this module, I feel even more strongly that we should construct a default `Ignore` instance here that reuses the `DEPLOY_IGNORE_DEFAULTS` definition from `deploy-ignore.ts`

tools/common/lib/fs-utils.ts
```diff
@@ -1,37 +1,53 @@
...
 
-		async function calculateSize( dirPath: string ): Promise< void > {
+		async function calculateSize( dirPath: string, relativePath: string ): Promise< void > {
 			try {
 				const files = await fsPromises.readdir( dirPath, { withFileTypes: true } );
 
 				await Promise.all(
 					files.map( async ( file ) => {
 						const filePath = path.join( dirPath, file.name );
+						const fileRelativePath = relativePath ? `${ relativePath }/${ file.name }` : file.name;
```
```suggestion
						const fileRelativeToRoot = path.relative( directoryPath, filePath );
```
Two things here:

1. I think we can rename this variable for clarity. 
2. If I'm not mistaken, we can simplify the way this path is generated like so. This saves us from having to pass the `relativePath` argument recursively.

fredrikekelund (approved):
This is a great addition 👍 I left a couple of comments on the implementation, but the core functionality looks good and tests well for me. I think we should address my comments, but approving now to unblock you, @ivan-ottinger

tools/common/lib/fs-utils.ts
ivan-ottinger: Good idea! I have made the change in https://github.com/Automattic/studio/pull/2924/changes/54754cc5c5b5bcd0d744720ed5d3ef816d817168.

tools/common/lib/deploy-ignore.ts
ivan-ottinger: Good catch. Yes, I initially tried to target site sync as well, but then decided to leave it for a separate PR to reduce size and complexity of this PR.

I have removed the export as a part of  https://github.com/Automattic/studio/pull/2924/changes/54754cc5c5b5bcd0d744720ed5d3ef816d817168.

tools/common/lib/fs-utils.ts
ivan-ottinger: I like the suggestion and think it is easier to read. Made the change in https://github.com/Automattic/studio/pull/2924/changes/cb68051195f11bcd7896a5314a365b304f16dc40.

ivan-ottinger:
> This is a great addition 👍 I left a couple of comments on the implementation, but the core functionality looks good and tests well for me. I think we should address my comments, but approving now to unblock you, @ivan-ottinger

Thanks for the review and feedback, Fredrik. I have updated the PR. Next I will retest the logic on both macOS and Windows and then merge.

ivan-ottinger:
> Thanks for the review and feedback, Fredrik. I have updated the PR. Next I will retest the logic on both macOS and Windows and then merge.

Everything is testing well. The files are properly excluded on both macOS and Windows. I did not observe any regressions.

georgestephanis:
@ivan-ottinger I was just looking for something similar on this -- does this / could this also support `.deployignore` files in subdirectories, relative to that directory?

For example, if I manage a theme or plugin that I've got checked out inside of a local synced site in Studio, could I have a `.deployignore` file in that theme or plugin to exclude its `src` and npm/composer directories?

ivan-ottinger:
> @ivan-ottinger I was just looking for something similar on this -- does this / could this also support .deployignore files in subdirectories, relative to that directory?

> For example, if I manage a theme or plugin that I've got checked out inside of a local synced site in Studio, could I have a .deployignore file in that theme or plugin to exclude its src and npm/composer directories?

Great idea, George! Currently `.deployignore` is only read from the site root, so subdirectory ignore files aren't supported.

It could be a nice enhancement though. I have opened up a follow-up issue (STU-1499) for it so we can explore that approach further.

PR #2925, opened 2026-03-26
Add i18n translation wrappers to CLI AI command strings

sejas:
## Related issues

- Fixes STU-1465

## How AI was used in this PR

Claude identified untranslated user-facing strings across the CLI AI command files, planned the translation wrapping strategy, and applied the changes. All strings were reviewed to ensure AI-model-facing strings (like "Continue") and the WordPress motto ("Code is Poetry") were intentionally left untranslated.

## Proposed Changes

- Wrap `studio ai` user-facing strings using `__()` and `sprintf()` from `@wordpress/i18n` across 4 files:
  - `apps/cli/ai/slash-commands.ts` — 7 slash command descriptions
  - `apps/cli/ai/ui.ts` — ~20 UI strings (placeholders, labels, hints, error/status messages)
  - `apps/cli/commands/ai/index.ts` — ~25 strings (session, provider, model, login/logout messages)
  - `apps/cli/commands/ai/sessions/helpers.ts` — 1 keyboard hint string
- Add `/* translators: */` comments for all strings containing placeholders
- Convert template literals with variables to `sprintf(__(), ...)` pattern with positional placeholders

## Testing Instructions
  - Set your `locale` to `es` in `~/.studio/shared.json`
  - Replace your `tools/common/translations/studio-es.jed.json` with this file:
[studio-es.jed.json](https://github.com/user-attachments/files/26343371/studio-es.jed.json)
  - Run `npm run cli:build`, and open the AI chat (`node apps/cli/dist/cli/main.js ai`)
  - Verify the UI is in Spanish and slash command descriptions appear correctly (type `/` to see autocomplete)

Screenshot 2026-03-30 at 09 10 55

Screenshot 2026-03-30 at 09 11 14

## Pre-merge Checklist


apps/cli/ai/ui.ts
```diff
@@ -1253,7 +1266,7 @@ export class AiChatUI {
 				} · ${ displayCwd }`
 			),
 			'',
-			chalk.dim.italic( 'Code is Poetry' ),
+			chalk.dim.italic( __( 'Code is Poetry' ) ),
```
katinthehatsite, line 1270: I am wondering if we need to translate this since it is more like a brand tagline and can also sound weird in other languages. To me, this would be somewhat equal to WordPress.com which I would not translate.

wojtekn: Let's leave this decision to translators - in some languages, they can decide not to translate, in some, they can translate. For example, in Polish, I would translate this phrase.

apps/cli/commands/ai/index.ts
```diff
@@ -405,20 +453,20 @@ export async function runCommand(
...
+				ui.setStatusMessage( __( 'Logged out of WordPress.com' ) );
 				await maybeAutoSwitchProvider();
 				continue;
 			}
 
 			if ( trimmedPrompt === AI_CHAT_MODEL_COMMAND ) {
 				const modelOptions = ( Object.entries( AI_MODELS ) as [ AiModelId, string ][] ).map(
 					( [ id, label ] ) => ( {
-						label: id === currentModel ? `${ label } (current)` : label,
+						label: id === currentModel ? `${ label } ${ __( '(current)' ) }` : label,
```
katinthehatsite: I was thinking about using `sprintif` and if it would be better so that the translations could re-order the words if needed.  Claude gave me Korean language as an example where this would be used but I am not sure if it is a big deal. Just adding here as an observation.

wojtekn: This is correct, we should avoid concatenating translatable text, and sprintf should be the way to go.

katinthehatsite (approved):
I think it works well for the phrases that you added in this PR so I am approving this PR (I left two minor comments).

Image

Something else that I found is that if I write in Spanish, it writes back to me in English 😅 

Image

Also loading indicators such as `Loading`, 'Filtering` etc are shown in English. Is it something that we will be looking to refine further? Thanks!

apps/cli/ai/ui.ts
epeicher: Should we translate this one?

epeicher: I think this should be translatable too

epeicher, line 254: Not critical as they are tool names, but worth considering, should we translate this?

epeicher (approved):
Thanks Antonio for handling this! I have tested it, and it works as expected, I can see the messages translated to Spanish. I have left a couple of minor comments, but they can be tackled as follow-ups depending on the conclusions. To me, this can be progressed, LGTM!

Image

epeicher:
> Something else that I found is that if I write in Spanish, it writes back to me in English 

Good catch @katinthehatsite! I think this is something that should be addressed as a folowup, but for me is replying in Spanish 
CleanShot 2026-03-30 at 13 15 56@2x

PR #2709, opened 2026-03-05
Test artifact signing

mokagio:
## Related issues

- Fixes #

## Proposed Changes

-

## Testing Instructions

-

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #2928, opened 2026-03-26
Include scripts in typecheck

mokagio:
## Related issues

- Related to AINFRA-2177

## How AI was used in this PR

- AI helped draft the initial implementation and commit message rationale. I reviewed the resulting config changes and validated the typecheck behavior manually.

## Proposed Changes

When working on #2709 , I run into some issues with the scripts in one of the iterations. The AI suggested adding typechecking for `scripts/` which makes sense to me. However, I might be missing something here, such as whether it's okay to have scripts not being workspaces, or the difference in rigor to apply between small bits of automation and the app itself, so **feel free to close this PR if the suggestion is not beneficial.**

- Add `scripts/tsconfig.json` so the repo scripts can be typechecked as their own TS project.
- Update the root `typecheck` script to run `tsc -p scripts/tsconfig.json --noEmit` before the workspace checks.
- Add `scripts/**/*` to the root `tsconfig.json` so editor tooling also includes those files.
- Document in the commit message why the root `tsconfig.json` change alone would not have enforced script typechecking.

## Testing Instructions

- Run `npx tsc -p scripts/tsconfig.json --noEmit` and confirm it passes.
- Run `npm run -s typecheck --workspaces --if-present` before the change and note that it does not typecheck `scripts/`.
- Run `npm run typecheck` after the change and confirm it now checks `scripts/` as well as the existing workspace projects.
- Optional: run `npx tsc -p tsconfig.json --noEmit` and confirm the existing `apps/cli` `import.meta` errors are still present, which is why the dedicated `scripts/tsconfig.json` is needed.

## Pre-merge Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
 		"app:install:bundle": "npm -w studio-app run install:bundle",
 		"compare:perf": "npm -w compare-perf run compare",
 		"lint": "eslint {apps/cli,apps/studio/src,apps/studio/e2e,tools/common}",
-		"typecheck": "tsc -p scripts/tsconfig.json --noEmit && npm run typecheck --workspaces --if-present",
+		"typecheck": "npm run typecheck --workspaces --if-present && tsc -p scripts/tsconfig.json --noEmit",
 		"format": "npm run lint -- --fix",
 		"test": "vitest run",
 		"test:watch": "vitest",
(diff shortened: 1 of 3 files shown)


epeicher (approved):
Thanks @mokagio for adding this! It makes sense to typecheck the scripts similarly to any other piece of code. Also, the current approach of not adding scripts as a workspace is simple; we might reconsider adding it as a workspace in the future so we don't need the extra `tsc` in the `typecheck` script, but at the moment, this solution is fine. Interested in others' views on this.

I have tested it, and I can confirm that it now typechecks the scripts and I have not found any issues. Changes also LGTM!

PR #2932, opened 2026-03-27
Fix SSL certificate trust nudge on non-English Windows

bcotrim:
## Related issues

- Fixes STU-662

## How AI was used in this PR

I used AI to investigate the issue, confirm the cause (translated strings), lookup the value and implement the fix

## Proposed Changes

- Replace the localized `"Server Authentication"` string check with the RFC 5280 OID `1.3.6.1.5.5.7.3.1` (`id-kp-serverAuth`) in both desktop and CLI certificate managers
- Extract the OID as a shared constant `SERVER_AUTH_OID` in `tools/common/constants.ts`

### Root cause

`isRootCATrusted()` on Windows parses `certutil -verify` output looking for the string `"Server Authentication"`. Windows localizes this label (e.g. `"Autenticação de servidor"` in Portuguese, likely `"Uwierzytelnianie serwera"` in Polish), so the check always fails on non-English systems. The OID `1.3.6.1.5.5.7.3.1` is a standardized identifier from RFC 5280 that remains constant regardless of locale.

## Testing Instructions

- Set Windows display language to a non-English language (e.g. Portuguese, Polish)
- Create a site with a custom domain and HTTPS enabled
- Observe the "Trust Certificate" nudge in site settings
- Click "Trust Certificate" and approve the UAC prompt
- Verify the nudge disappears after trusting

## Pre-merge Checklist


ivan-ottinger (approved):
Nice fix! The proposed changes look good to me and I can confirm they fix the issue.

Before:

Image

After:

Image

apps/cli/lib/certificate-manager.ts
```diff
@@ -159,14 +160,11 @@ export async function isRootCATrusted(): Promise< boolean > {
 
 	if ( process.platform === 'win32' ) {
 		try {
-			// Execute certutil with more specific validation
 			const { stdout } = await execFilePromise( 'certutil', [ '-verify', CA_CERT_PATH ] );
 
 			const hasValidPolicies =
-				stdout.includes( 'Verified Application Policies:' ) &&
-				stdout.includes( 'Server Authentication' );
+				stdout.includes( 'Verified Application Policies:' ) && stdout.includes( SERVER_AUTH_OID );
```
epeicher: `Verified Application Policies:` is also localized by Windows (e.g. `Directivas de aplicación comprobadas:` in Spanish), so this check always fails on non-English systems. Since `certutil -verify` throws on failure, checking for the OID alone is sufficient.


```suggestion
			const hasServerAuthOID = stdout.includes( SERVER_AUTH_OID );
```

apps/studio/src/lib/certificate-manager.ts
```diff
@@ -27,14 +28,11 @@ export async function isRootCATrusted(): Promise< boolean > {
 
 	if ( process.platform === 'win32' ) {
 		try {
-			// Execute certutil with more specific validation
 			const { stdout } = await execFilePromise( 'certutil', [ '-verify', CA_CERT_PATH ] );
 
 			const hasValidPolicies =
-				stdout.includes( 'Verified Application Policies:' ) &&
-				stdout.includes( 'Server Authentication' );
+				stdout.includes( 'Verified Application Policies:' ) && stdout.includes( SERVER_AUTH_OID );
```
epeicher: Similar [to the above](https://github.com/Automattic/studio/pull/2932/changes#r3022792830)

epeicher (commented):
Thanks @bcotrim! I have tested on Windows by changing the language to Spanish and I still see the warning. I have posted some suggestion about a potential fix

bcotrim:
nice catch @epeicher
On my tests, with Windows in Portuguese `Verified Application Policies:` wasn't translated.
Using just the `OID` would not work since it would still show up in an untrusted certificate.
I’ve updated the condition to check for `CERT_TRUST_IS_UNTRUSTED_ROOT`.
Could you give it another review, please?

epeicher (approved):
Thanks @bcotrim for addressing the comments! I have tested it on Windows on Spanish, and it works as expected. The nudge disappears after accepting the UAC dialog. LGTM!

|Before clicking on Trust | After clicking on Trust|
|-------|------|
|Image |Image |

PR #2935, opened 2026-03-27
Fix blueprint local resource resolution when creating sites from the app

sejas:
## Related issues

- Fixes STU-1458

## How AI was used in this PR

Claude was used to trace the blueprint path flow across renderer → IPC → CLI, identify the root cause (temp file losing original directory context), and implement the fix. All code was reviewed and tested manually.

## Proposed Changes

- Capture Blueprint full filesystem path when a user selects a blueprint file
- Thread `originalBlueprintPath` through `createSite` IPC → `SiteServer.create()` → `createSiteViaCli()`
- Pass `--original-blueprint-path` new hidden CLI parameter so the CLI uses the original directory for resolving relative resources (e.g. `./koinonia.zip`)

## Testing Instructions

1. Download a blueprint JSON file and a theme zip from the Linear issue (e.g. `koinonia.zip`) to the same directory (e.g. `~/Downloads/`). 
```json
{
  "$schema": "https://playground.wordpress.net/blueprint-schema.json",
  "login": true,
  "siteOptions": {
    "blogname": "Koinonia"
  },
  "steps": [
    {
      "step": "resetData"
    },
    {
      "step": "installTheme",
      "themeData": {
        "resource": "bundled",
        "path": "./koinonia.zip"
      },
      "options": {
        "activate": true
      }
    }
  ]
}
```
3. Start the app with `npm start`
4. Create a new site using the file picker to select the blueprint.json local file
5. Verify the theme installs correctly (the `installTheme` step resolves the local zip)
6. Verify normal blueprint flows still work (gallery blueprints, deeplink blueprints)

https://github.com/user-attachments/assets/f26804ed-8836-460f-a6bd-c89840e938b9

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/site/create.ts
+++ b/apps/cli/commands/site/create.ts
@@ -587,6 +587,10 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 					type: 'string',
 					describe: __( 'Path or URL to Blueprint JSON file' ),
 				} )
+				.option( 'original-blueprint-path', {
+					type: 'string',
+					hidden: true,
+				} )
 				.option( 'admin-username', {
 					type: 'string',
 					describe: __( 'Admin username (defaults to "admin")' ),
@@ -767,6 +771,12 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 						uri,
 						contents: readBlueprint( uri ),
 					};
+
+					// When invoked by the desktop app, the blueprint contents come from a temp file
+					// but resources should be resolved relative to the original file location.
+					if ( argv.originalBlueprintPath ) {
+						config.blueprint.uri = path.resolve( argv.originalBlueprintPath );
+					}
 				}
 			}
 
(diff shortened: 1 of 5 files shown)


bcotrim (commented):
The changes make sense but I wonder if we could write the absolute paths in the tmp blueprint file, to avoid the new field.
It seems like a more contained change, with a clear intention.
What do you think?

sejas:
I think using a parameter is a better solution. The issue I see with copying other files is that it needs to work across many different steps, and copying them to /tmp could also cause us to hit disk size limits. Some themes or imports could be large.

The parameter is hidden, so it won’t be accessible directly from the CLI. The parameter solution is a solution suggested in [STU-1458](https://linear.app/a8c/issue/STU-1458/studio-local-blueprints-that-reference-local-resources-dont-work-in)

bcotrim (approved):
> I think using a parameter is a better solution. The issue I see with copying other files is that it needs to work across many different steps, and copying them to /tmp could also cause us to hit disk size limits. Some themes or imports could be large.

I meant changing the paths in the blueprints to be absolute paths instead of relative ones.
But the current approach works and it's clean, so LGTM 👍

PR #2937, opened 2026-03-27
macOS: Auto-install CLI to ~/.local/bin

ivan-ottinger:
## Related issues

- Related to STU-1363

## How AI was used in this PR

Claude Code was used to implement the changes, write tests, and create this PR. The solution was developed iteratively through discussion — we explored the existing codebase patterns (macOS and Windows installation managers), identified edge cases (e.g., respecting user preference when CLI is explicitly disabled, Dock-launched apps not inheriting shell PATH), and refined the approach based on the Linear issue description and comments.

## Proposed Changes

- Replace `/usr/local/bin/studio` symlink (requires sudo) with `~/.local/bin/studio` (user-writable, no sudo needed)
- Auto-install CLI on first app startup via `autoInstallMacOSCliIfNeeded()`, with a `cliAutoInstalled` flag in app config to avoid re-enabling if the user explicitly disabled it
- Add shell profile PATH management: appends `export PATH="$HOME/.local/bin:$PATH"` to the user's shell profile if the export line isn't already present. The profile file is determined by `process.env.SHELL` (`.zshrc` for zsh, `.bash_profile` for bash, defaults to `.zshrc`). `~/.local/bin` is prepended so the bundled CLI takes priority over any standalone npm-installed version
- CLI installation status (`isCliInstalled`) checks the profile file content rather than runtime `process.env.PATH`, since Dock-launched Electron apps on macOS don't inherit shell PATH
- Remove sudo shell scripts (`install-studio-cli.sh`, `uninstall-studio-cli.sh`) that are no longer needed
- Add unit tests for the refactored installation manager
- Auto-install is skipped in development mode (`npm start`) to avoid creating symlinks pointing to dev resources

## Testing Instructions

> **Note:** Auto-install only runs in production mode. You must test with a built app (`npm run make`), not `npm start`.

### Setup
1. Remove any existing CLI symlink: `rm -f ~/.local/bin/studio`
2. Remove the auto-install flag if present (if you are testing this PR for the first time, it should not be present - since we are introducing it in this PR): edit `~/.studio/app.json` and delete the `"cliAutoInstalled": true` entry
3. Build and install the app: `npm run make`, then install the resulting DMG to `/Applications/`

### Auto-install on first launch
4. Launch Studio
5. Verify the symlink was created: `ls -la ~/.local/bin/studio` — should point to `/Applications/Studio.app/Contents/Resources/bin/studio-cli.sh`
6. Verify the flag was set: `grep cliAutoInstalled ~/.studio/app.json` — should show `"cliAutoInstalled": true`
7. Open a **new** terminal and run `studio --version` — should print the current version
8. Check your shell profile: `tail -3 ~/.zshrc` — should contain `export PATH="$HOME/.local/bin:$PATH"` (if it wasn't already there)

### Shell profile PATH management
9. To test that the export line is added correctly, temporarily remove `$HOME/.local/bin` from your shell profile (e.g., edit `~/.zshrc` and delete the `export PATH="$HOME/.local/bin:$PATH"` line), then remove the CLI symlink and reset the flag:
   - `rm -f ~/.local/bin/studio`
   - Edit `~/.studio/app.json` and delete the `"cliAutoInstalled": true` entry
10. Relaunch Studio
11. Verify the export line was added back: `grep '$HOME/.local/bin' ~/.zshrc` — should show `export PATH="$HOME/.local/bin:$PATH"`
12. Verify it was not duplicated: `grep -c '$HOME/.local/bin' ~/.zshrc` — should show `1`

### User preference is respected
13. In Studio, go to Settings > Preferences, toggle the CLI **off**, and save
14. Verify the symlink is gone: `ls ~/.local/bin/studio` — should not exist
15. Quit and relaunch Studio
16. Verify the CLI was **not** reinstalled: `ls ~/.local/bin/studio` — should still not exist
17. Verify the flag is still set: `grep cliAutoInstalled ~/.studio/app.json` — should still show `true`

### Re-enabling works without sudo
18. Toggle the CLI back **on** in Settings > Preferences, and save
19. Verify the symlink is back: `ls -la ~/.local/bin/studio`

### No-op on subsequent launches
20. Quit and relaunch Studio
21. Verify the symlink is unchanged and no errors in the logs

## Pre-merge Checklist


apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -1,25 +1,33 @@
...
 const cliPackagedPath = path.join( binPath, 'studio-cli.sh' );
-const installScriptPath = path.join( binPath, 'install-studio-cli.sh' );
-const uninstallScriptPath = path.join( binPath, 'uninstall-studio-cli.sh' );
+
+const PATH_EXPORT_LINE = 'export PATH="$HOME/.local/bin:$PATH"';
+const SHELL_PROFILE_FILES = [ '.zshrc', '.zprofile', '.bash_profile', '.bashrc' ];
 
 const ERROR_FILE_ALREADY_EXISTS = 'Studio CLI symlink path already occupied by non-symlink';
-// Defined in @vscode/sudo-prompt
-const ERROR_PERMISSION = 'User did not grant permission.';
```
ivan-ottinger, line 22: Since we don't run sudo, the error message is no longer needed.

apps/studio/src/modules/cli/lib/tests/macos-installation-manager.test.ts
```diff
@@ -0,0 +1,386 @@
...
+function enoentError() {
+	const error = new Error( 'ENOENT' ) as NodeJS.ErrnoException;
+	error.code = 'ENOENT';
+	return error;
+}
+
+// These tests are macOS-specific — skip on other platforms where path separators differ.
+const isNonDarwin = process.platform !== 'darwin';
+
+describe.skipIf( isNonDarwin )( 'MacOSCliInstallationManager', () => {
```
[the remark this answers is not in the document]
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2937/commits/bf34da110b651c2d98bb41d3c67d5e8fb879209d.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -125,6 +131,22 @@ export class MacOSCliInstallationManager implements StudioCliInstallationManager
...
+		// Only auto-install on first launch. If the flag is already set but the CLI isn't
+		// installed, the user must have explicitly disabled it — respect their choice.
+		const userData = await loadUserData();
+		if ( userData.cliAutoInstalled ) {
+			return;
+		}
+
+		await this.installCli();
+		await updateAppdata( { cliAutoInstalled: true } );
+	}
```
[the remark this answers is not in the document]
ivan-ottinger: I would say this is quite a very-edge-case. Something like this would need to happen:

  1. Launch → auto-install fails for some reason, flag never set
  2. User manually installs via toggle → works, but flag not set
  3. User disables via toggle
  4. Next launch → flag is undefined, CLI not installed → auto-install tries again

Because of that, I am leaving this as-is for now.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -178,18 +196,55 @@ export class MacOSCliInstallationManager implements StudioCliInstallationManager
 			}
 		}
 
+		await unlink( cliSymlinkPath );
+	}
```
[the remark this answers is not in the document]
ivan-ottinger: Resolved in https://github.com/Automattic/studio/pull/2937/commits/bf34da110b651c2d98bb41d3c67d5e8fb879209d.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -125,6 +131,22 @@ export class MacOSCliInstallationManager implements StudioCliInstallationManager
 		}
 	}
 
+	async autoInstallIfNeeded(): Promise< void > {
+		if ( await this.isCliInstalled() ) {
+			return;
+		}
```
fredrikekelund: Not really a big deal, but we could make it the responsibility of `installCli` to check this condition and return early from this function only if `cliAutoInstalled === true`. Again, not a huge concern, but it's just unnecessary to check this condition twice.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
fredrikekelund, line 1: The `isCliInstalled` method should also ensure that `~/.local/bin` is in `$PATH`

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -172,24 +190,61 @@ export class MacOSCliInstallationManager implements StudioCliInstallationManager
...
 		}
 
+		await unlink( cliSymlinkPath );
+	}
+
+	private async ensurePathInProfile(): Promise< void > {
+		const localBinPath = path.join( os.homedir(), '.local', 'bin' );
+
+		// If ~/.local/bin is already in PATH, no need to modify the shell profile.
+		if ( process.env.PATH?.split( ':' ).includes( localBinPath ) ) {
```
fredrikekelund: I believe we also need to untildify (see `apps/cli/lib/utils.ts`), replace `$HOME`, and `path.resolve` each of the `process.env.PATH` entries.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -172,24 +190,61 @@ export class MacOSCliInstallationManager implements StudioCliInstallationManager
...
+			return;
+		}
+
+		const homeDir = os.homedir();
+		let profilePath: string | null = null;
+
+		// Find the first existing shell profile file.
+		for ( const file of SHELL_PROFILE_FILES ) {
```
fredrikekelund: It's possible to have a `~/.zshrc` file but still use a different shell. I think we should try mapping `process.env.SHELL` to the corresponding profile file.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -172,24 +190,61 @@ export class MacOSCliInstallationManager implements StudioCliInstallationManager
...
+				break;
+			} catch {
+				// File doesn't exist, try the next one.
+			}
+		}
+
+		// Default to ~/.zshrc if no profile file exists (zsh is the default macOS shell).
+		if ( ! profilePath ) {
+			profilePath = path.join( homeDir, '.zshrc' );
+		}
```
fredrikekelund: Let's use this as the default if `process.env.SHELL` is empty.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -172,24 +190,61 @@ export class MacOSCliInstallationManager implements StudioCliInstallationManager
...
-				},
-			} );
+			existingContent = await readFile( profilePath, 'utf-8' );
+		} catch {
+			// File doesn't exist yet, which is fine.
+		}
+
+		if ( existingContent.includes( '$HOME/.local/bin' ) ) {
+			return;
 		}
```
fredrikekelund: It makes sense not to append the same content repeatedly to the file, but at the same time, we also know that it doesn't have the intended effect at this point. Is there a fallback strategy we can use in this case..? 🤔

fredrikekelund (commented):
I left a couple of comments on the implementation. We also need to consider users updating from older Studio versions. We have two options there, as I see it:

1. Write a data migration that respects the old CLI installation setting, and only runs the auto-installer logic if the CLI was already installed. The downside is that this is probably not a conscious decision for a lot of users.
2. Run the auto-installer logic for all users: upgraders and new installers alike. The downside is that this will modify the environment for users who previously consciously uninstalled the CLI.

I'm leaning towards the second option.

As for the old CLI path, it's unfortunately not so easy to clean up, since we need elevated privileges to modify it. We could either leave it lying around, or add a function to the CLI un/-install flow that checks the old path and cleans it up if needed.

ivan-ottinger:
Thank you for your review and feedback, Fredrik!

> We also need to consider users updating from older Studio versions. We have two options there, as I see it:

> 1. Write a data migration that respects the old CLI installation setting, and only runs the auto-installer logic if the CLI was already installed. The downside is that this is probably not a conscious decision for a lot of users.
> 2. Run the auto-installer logic for all users: upgraders and new installers alike. The downside is that this will modify the environment for users who previously consciously uninstalled the CLI.
> 
> I'm leaning towards the second option.

Yes, I see it the same way and so we can keep the logic as-is. 🙂 

> As for the old CLI path, it's unfortunately not so easy to clean up, since we need elevated privileges to modify it. We could either leave it lying around, or add a function to the CLI un/-install flow that checks the old path and cleans it up if needed.

Yes, I was actually trying to clean it up before, but then decided to remove the whole logic in https://github.com/Automattic/studio/pull/2937/changes/20e9daeeccbdec5b49970da0e71540762c7c2cf7 - as it would run for a very few users only + keeping the old path there should not cause any issues.

---

I will check out your other review comments as well.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
ivan-ottinger: Good catch! Addressed in https://github.com/Automattic/studio/pull/2937/commits/514943da89e01aceb405f1dc228e4cb6c85b71ce.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
ivan-ottinger: That's true. Thanks for catching that. Made the change in https://github.com/Automattic/studio/pull/2937/changes/5eb12ecae32f7bb71a1e86169473376ac8e84739.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
ivan-ottinger: Good point. I have made the change in https://github.com/Automattic/studio/pull/2937/changes/5eb12ecae32f7bb71a1e86169473376ac8e84739.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
ivan-ottinger: I like the idea. It is more robust indeed. I have made the change in https://github.com/Automattic/studio/pull/2937/commits/816d53234643dbdafdd553a8bca309c56a88f7d7.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
ivan-ottinger: 👍🏼 Made the change in https://github.com/Automattic/studio/pull/2937/commits/816d53234643dbdafdd553a8bca309c56a88f7d7.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
ivan-ottinger: Do you think a case like this one where the path would be overwritten?

```
export PATH="$HOME/.local/bin:$PATH"
# ... bunch of other stuff ...
export PATH="/usr/bin:/bin"
```

I think it would be quite an edge case, no? But maybe you mean some different case? 🤔

ivan-ottinger:
> I will check out your other review comments as well.

I have applied the changes and will retest everything tomorrow with fresh. 🧠 🙂

ivan-ottinger:
I ended up [checking the shell profile file content](https://github.com/Automattic/studio/pull/2937/commits/5a0bc637e4ee93aeac1f4c68771a34463e9db11c) instead of `process.env.PATH`. Turns out that dock-launched Electron apps on macOS inherit their environment from the GUI session (`launchd`), not from the user’s shell. As a result, `process.env.PATH` typically does not include entries defined in `~/.zshrc` (such as `~/.local/bin`).

There's also a [this package](https://www.npmjs.com/package/fix-path?activeTab=readme) that helps with that, but I think checking the file contents itself should be enough.

ivan-ottinger:
Note: Failed E2E tests are unrelated to this PR.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -1,25 +1,47 @@
...
+	const shell = process.env.SHELL ?? '';
+	const profileFile = SHELL_PROFILE_MAP[ shell ] ?? DEFAULT_PROFILE;
+	return path.join( os.homedir(), profileFile );
+}
+
+async function readProfileContent(): Promise< string > {
+	try {
+		return await readFile( getProfilePath(), 'utf-8' );
+	} catch {
+		return '';
```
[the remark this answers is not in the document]
ivan-ottinger: Good point. Addressed in https://github.com/Automattic/studio/pull/2937/commits/7e5f09ea0a460b20e6bedfd324ca3ef07ec32da7.

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
```diff
@@ -172,24 +204,30 @@ export class MacOSCliInstallationManager implements StudioCliInstallationManager
...
-			// `/usr/local/bin` is not typically writable by non-root users, so in most cases, we run
-			// this uninstall script with admin privileges to remove the symlink.
-			await sudoExec( `/bin/sh "${ uninstallScriptPath }"`, {
-				name: packageJson.productName,
-				env: {
-					CLI_SYMLINK_PATH: cliSymlinkPath,
-				},
-			} );
+		await unlink( cliSymlinkPath );
+	}
```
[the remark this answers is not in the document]
ivan-ottinger: That's a very unlikely scenario. I don't think we need to solve it, because it would add unnecessary complexity to the code.

ivan-ottinger:
Hey @fredrikekelund 👋🏼🙂 The PR is now ready for another review round. I have tested the updated logic on the built macOS app and did not observe any issues.

I think we will probably won't get to merge today so considering my coming limited availability due to WordCamp I have also added the `Reviewer can merge` tag.

fredrikekelund (approved):
I made a bunch of additional changes here, informed by my own review and local AI agent reviews.

- Removed the tests. `MacOSCliInstallationManager` is not well-suited for testing, IMO. We simply have to mock too much for the tests to really give us any real assurances. 
- I brought back the old uninstallation script and added an `uninstallLegacyCliIfNeeded` method. Without this, `/usr/local/bin/studio` would be left intact, meaning the `studio` command would not truly be uninstalled.
- Started using `os.userInfo()` to look up the user's shell. This is a more reliable indicator than `process.env.SHELL`. We still fall back to the environment variable, though. 
- Started looking up shells by basenames rather than complete paths. This helps if the user has installed their current shell with Homebrew, for example. 
- Removed the "CLI Installed" modal. This is subjective on my part, but I don't think this really makes sense when we're no longer prompting the user for their password. We should only interrupt them with a modal if there was an error.

ivan-ottinger:
Thank you for the updates to the PR and pushing it to the finish line while I was AFK, Fredrik!

PR #2939, opened 2026-03-27
Add streaming site import via pull-reprint command

adamziel:
## Proposed Changes

Adds a `studio pull-reprint` command that imports a remote WordPress site into a local Studio instance using [reprint](https://github.com/adamziel/streaming-site-migration), a streaming site migration tool that runs inside PHP WASM.

The import runs in two phases:

1. **Essential files first** — downloads code, config, themes, and plugins needed to boot the site. Flattens the directory structure, applies the remote database to SQLite, generates a Playground runtime blueprint, and starts the Studio site so you can use it immediately.
2. **Remaining files** — downloads uploads and other large files that were skipped in phase 1 (media, etc.). If interrupted, re-running the same command resumes from where it left off.

For WordPress.com sites, the command handles authentication automatically — it uses your Studio OAuth token to enable the reprint exporter and rotate the HMAC secret, so you don't need to pass `--secret` manually.

### What's in this PR beyond the base command

- **E2E build artifact verification** — the Buildkite E2E pipeline now checks that `reprint.phar`, `reprint-child.mjs`, and `main.mjs` exist in `dist/cli/` after packaging.
- **Auto-login fix** — imported sites now get admin credentials generated before the server starts, so the auto-login mu-plugin works.
- **Spinner rendering fix** — progress lines are truncated to terminal width to prevent ghost lines in environments where `getWindowSize()` isn't available.

## Testing Instructions

First, visual demo:

https://github.com/user-attachments/assets/30a59b4c-61c0-4b5f-8152-e73aa7e596cd

### Prerequisites

The reprint exporter endpoint is gated behind Automattic's proxy. You need [autoproxxy](https://github.com/nicknisi/autoproxxy) running and the `ALL_PROXY` env var set:

```bash
STUDIO_ENABLE_PULL_REPRINT=1 ALL_PROXY=socks5h://127.0.0.1:8080 node apps/cli/dist/cli/main.mjs pull-reprint --url <site-url>
```

Note the imported site must be on WP Cloud. We may need more UX around this to help users migrate their simple sites to Atomic.

### Quick test (WordPress.com site)

2. Connect WP.com: `node apps/cli/dist/cli/main.mjs auth login`
3. Pull an atomic site
   ```bash
   STUDIO_ENABLE_PULL_REPRINT=1 \
   ALL_PROXY=socks5h://127.0.0.1:8080 \
     node apps/cli/dist/cli/main.mjs pull-reprint \
       --url https://your-wpcom-site.blog
   ```
4. Verify the site appears in `studio site list`
5. Open the auto-login URL printed at the end — you should land in wp-admin, not the setup screen

### Resume test

1. Start a pull and interrupt it mid-download (Ctrl-C)
2. Re-run the exact same command
3. Verify it resumes from where it left off (not from scratch)

### Non-WordPress.com site

For sites that aren't on WordPress.com, install the [reprint exporter plugin](https://github.com/adamziel/reprint) and pass `--secret`:

```bash
STUDIO_ENABLE_PULL_REPRINT=1 \
node apps/cli/dist/cli/main.mjs pull-reprint \
  --url https://example.com \
  --secret the-hmac-secret-from-the-plugin
```

### Unit tests

```bash
npm test -- apps/cli/commands/tests/pull-reprint.test.ts apps/cli/lib/pull/runtime-start-options.test.ts
```

## How AI was used in this PR

Claude helped with code review, documentation, bug fixes (wp-content path lookup, auto-login after import, spinner rendering), and removing workarounds by fixing issues upstream in reprint instead.

## Pre-merge Checklist


wojtekn:
@adamziel currently Studio uses "import" terminology for importing from a backup file or .sql file, and sync/push/pull for remote site integrations. Also, we are moving the current Import features from UI to the `studio import` command.

Could you please use another namespace for the new command, e.g., `studio stream-sync`, `studio sync stream`, or similar?

adamziel:
Let me loop in @Poliuk – would you be able to map all these distinct features and find a useful naming strategy for them?

Poliuk:
> would you be able to map all these distinct features and find a useful naming strategy for them?

Yes, let me work on it this week

apps/cli/commands/site/delete.ts
```diff
@@ -117,9 +117,14 @@ export async function runCommand(
 		}
 
 		if ( deleteFiles ) {
-			if ( fs.existsSync( siteFolder ) ) {
+			const deleteTargets = [ siteFolder, site.technicalSiteDirectory ]
+				.filter( ( value ): value is string => Boolean( value ) )
+				.filter( ( value, index, values ) => values.indexOf( value ) === index )
+				.filter( ( value ) => fs.existsSync( value ) );
```
adamziel: This is confusing, let's simplify it

apps/cli/commands/site/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+class ImportError extends LoggerError {
+	technicalDetails: string;
+
+	constructor( userMessage: string, technicalDetails: string ) {
+		super( userMessage );
+		this.technicalDetails = technicalDetails;
+	}
+}
+
+function redBox( message: string ): string {
```
adamziel: Move this to like a logging.ts or another dedicated file. Perhaps reuse exisitng Studio helpers or colocate it with them

apps/cli/commands/site/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+			}
+		} catch {
+			// Log file may not exist yet.
+		}
+	}
+
+	return sections.length > 0 ? sections.join( '\n\n' ) : null;
+}
+
+function formatServerStartErrorDetails( siteId: string | undefined, error: unknown ): string {
```
adamziel: What is this? Why do we need it?

apps/cli/commands/site/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+		if ( recentLogs ) {
+			parts.push( `\nDaemon process logs:\n${ recentLogs }` );
+		}
+	}
+
+	return parts.join( '\n' );
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function parseReprintJson( result: ReprintProcessResult ): any {
```
adamziel: Rename to `parseReprintOutputStream` and ad da docblock explaining the data format

apps/cli/commands/site/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+			// Fall through to throw a descriptive error below.
+		}
+	}
+
+	throw new LoggerError(
+		`reprint.phar did not return valid JSON.\nstdout: ${ raw }\nstderr: ${ result.stderr }`
+	);
+}
+
+function getStageRank( stage: ImportStage ): number {
```
adamziel: This is such a simple method. Do we really need it? Similarly for a bunch of other methods in this file. Maybe we could group them in a static object at least?

apps/cli/commands/site/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+
+function getLegacyStateDirectory( technicalSiteDirectory: string ): string {
+	return path.join( technicalSiteDirectory, 'tmp', 'export', 'state' );
+}
+
+function getLegacyRawDirectory( technicalSiteDirectory: string ): string {
+	return path.join( technicalSiteDirectory, 'tmp', 'export', 'fs-root' );
+}
+
+function moveDirectoryContents( sourceDirectory: string, targetDirectory: string ): boolean {
```
adamziel: Document what this is, is it like `mv`? Does it work across devices? Why not use an existing helper to that same effect?

apps/cli/commands/site/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+	}
+
+	if ( fs.readdirSync( sourceDirectory ).length === 0 ) {
+		fs.rmSync( sourceDirectory, { recursive: true, force: true } );
+	}
+
+	return moved;
+}
+
+export function migrateLegacyReprintLayout(
```
adamziel: We don't need to migrate anything, this is a brand new PR. Drop everything that migrates from prior changes in this pr.

apps/cli/commands/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+	);
+	const movedRaw = moveDirectoryContents(
+		getLegacyRawDirectory( technicalSiteDirectory ),
+		rawDirectory
+	);
+
+	return movedState || movedRaw;
+}
+
+function saveImportMetadata( metadata: ImportMetadata ): void {
```
adamziel: Hm, what is this used for?

apps/cli/commands/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+	const cliConfig = await readCliConfig();
+	const baseName = inferSiteNameFromUrl( normalizedUrl );
+	return generateNumberedName(
+		baseName,
+		cliConfig.sites.map( ( site ) => site.name ),
+		path.dirname( getDefaultSitePath( baseName ) )
+	);
+}
+
+export function findMatchingWpComSite(
```
adamziel, line 995: Do we need this helper? Don't we have another helper to that effect in Studio?

apps/cli/commands/site/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+function readStoredSecret( url: string, explicitName?: string ): string | undefined {
+	const normalizedUrl = normalizeImportUrl( url );
+	const importKey = getImportKey( normalizedUrl, explicitName );
+	const technicalSiteDirectory = path.join( IMPORTS_ROOT, importKey );
+	const metadataPath = getMetadataPath( technicalSiteDirectory );
+	const metadata = readImportMetadata( metadataPath );
+	return metadata?.secret;
+}
+
+async function resolveSiteName( normalizedUrl: string, explicitName?: string ): Promise< string > {
```
adamziel: Explain what "site" – in the name, in the docblock.

apps/cli/commands/site/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+
+export function getImportKey( normalizedUrl: string, explicitName?: string ): string {
+	return crypto
+		.createHash( 'sha256' )
+		.update( `${ normalizedUrl }\n${ explicitName || '__auto__' }` )
+		.digest( 'hex' )
+		.slice( 0, IMPORT_KEY_HEX_LENGTH );
+}
+
+function readStoredSecret( url: string, explicitName?: string ): string | undefined {
```
adamziel: This sounds scary. What secret? Explain it in the docblock and, ideally, also in the method name.

apps/cli/commands/pull-reprint.ts
```diff
@@ -0,0 +1,1721 @@
...
+	return new URL( normalizeImportUrl( url ) ).hostname;
+}
+
+// 12 hex characters = 48 bits of a SHA-256 hash.  This is used as a
+// directory name for the import's technical files, not for security.
+// At 48 bits the birthday-collision threshold is ~2^24 (~16 million)
+// imports — far beyond any realistic single-machine usage.
+const IMPORT_KEY_HEX_LENGTH = 12;
+
+export function getImportKey( normalizedUrl: string, explicitName?: string ): string {
```
adamziel: Get from what? Explain that and document it.

apps/cli/commands/pull-reprint.ts
```diff
@@ -0,0 +1,1689 @@
...
+ * import directory (state, raw, runtime).  When Studio has already
+ * registered the flattened site, routes through `studio site delete`
+ * so the snapshot record and auto-start setting go too.
+ */
+async function abortImport(
+	url: string | undefined,
+	providedName?: string,
+	verbose = false
+): Promise< void > {
+	if ( ! url ) {
```
adamziel: check this in the caller and require url to be a string

apps/cli/vite.config.base.ts
```diff
@@ -26,25 +26,39 @@ if ( ! minimumNodeVersion ) {
...
-			],
-		} ),
+		{
+			name: 'write-dist-extras',
+			apply: 'build',
+			writeBundle( options ) {
+				const outDir = options.dir ?? resolve( __dirname, 'dist/cli' );
+				mkdirSync( outDir, { recursive: true } );
+				writeFileSync(
+					resolve( outDir, 'package.json' ),
```
adamziel, line 41: Replacing `viteStaticCopy` here (rather than keeping it alongside a second copy step for `reprint.phar`) so both bundled extras go through one `writeBundle` hook. This also lets us no-op cleanly when `wp-files` or `reprint.phar` aren't present (e.g. fresh checkouts / partial builds) instead of failing the build the way `viteStaticCopy` does on a missing `src`.

wojtekn:
@adamziel I roughly tested the pull part. After I figured out the proxy variable thing, it worked fine and pretty quickly. It pulled the site successfully. Some notes:

- The download progress showed something like 15000 files from 30000 and then quickly finished. However, it was showing that it downloaded more than 500MB of data. My pulled site on disk takes 50MB. It's worth checking where this difference comes from.
- We could improve the UX around listing existing sites. Maybe we could make it consistent with the default pull command, and allow users interactively select it?
- Description says that the command supports only WP Cloud hosts, but the output lists self-hosted sites, too. We could clarify this part.
- When I open UI after pulling the site, I can't see the site connected to the remote site in the Sync tab. Could we connect it during the `pull-reprint` flow so the user can use current UI to either pull it again or push (using the old sync)?
- We could consider hooking the new pull (maybe gated as a beta feature?) in the 'Add site' flow, where the user can create a new Studio site based on the remote Jetpack-connected site. In this case, `pull-reprint`, which supports only one-way full sync, would be feasible.

youknowriad:
@adamziel Unfortunately the experience is not great for me so far.

So I tried two things, running the existing "pull" and running "pull-print".

For the existing pull:

 - If I don't provide a url, it list my wp.com sites to pick from
 - It imports into the site that I specified using --path and doesn't create a new one 
 - It took around 3mn to finish
 - Once done the site is functional and looks exactly like my prod site

For the pull-print

 - The url is mandatory, if you don't provide it, it list the sites but you can't select.
 - It always creates a new site, not sure how to import into an existing one
 - It took around 7mn to finish
 - Once done the site doesn't work for me, loading it asks me to "install WP" (the setup flow)

Outside the bugs and performance issues, I think the flow between "pull" and "pull-print" should be the same: url optional and import into the specified site.

The most concerning thing for me though is the performance, I feel like it's downloading too much and with the existing pull it feels to do way less with a better result in the end.

adamziel:
@wojtekn @youknowriad all good notes! Let's get this in behind a beta flag and solve them in follow-up PRs. I'll wrap this in a flag.

tools/common/lib/app-config.ts
```diff
@@ -0,0 +1,57 @@
...
+	.loose();
+
+export type AppConfig = z.infer< typeof appConfigSchema >;
+
+const DEFAULT_APP_CONFIG: AppConfig = {
+	version: 1,
+	siteMetadata: {},
+};
+
+export async function readAppConfig(): Promise< AppConfig > {
```
adamziel: This file duplicates machinery that already exists. `getAppConfigPath()` and `apps/studio/src/storage/paths.ts`'s `getUserDataFilePath()` resolve to the same `app.json`, and `apps/studio/src/storage/user-data.ts`'s `loadUserData()` / `saveUserData()` already read and write it — with `apps/studio/src/lib/beta-features.ts` already layering `BETA_FEATURE_DEFAULTS` on top via `getBetaFeatures()`.

The only reason this parallel reader exists is that `user-data.ts` imports `@sentry/electron/main` (for breadcrumbs on parse errors), which the CLI can't load outside Electron. Rather than introduce a second schema + `readAppConfig` + `isAppBetaFeatureEnabled` + `getAppBetaFeatures`, the cleaner fix is to lift the pure JSON read/write out of `user-data.ts` into `tools/common/` (alongside `shared-config.ts`), keep the Sentry breadcrumb in a thin desktop wrapper, and have both `beta-features.ts` and the CLI gate call the shared reader. Then the CLI gate is just `(await loadUserData()).betaFeatures?.pullReprint` — no new module, no risk of the two schemas drifting.

adamziel:
This is behind a feature flag (env variable) as follows:

```sh
STUDIO_ENABLE_PULL_REPRINT=1 ALL_PROXY=socks5h://127.0.0.1:8080 node apps/cli/dist/cli/main.mjs pull-reprint --url <site-url>
```

Since this PR is already large, I'll go ahead and merge and then open follow-up PRs to address specific feedback points.

PR #2943, opened 2026-03-30
Clearing expired preview site shows unnecessary error

katinthehatsite:
## Related issues

Fixes STU-1471

## How AI was used in this PR

I debated two different solutions with AI before picking an option. 

## Proposed Changes

This PR ensures that when an expired preview site is to be `Clear`, we are not making the API request. 

## Testing Instructions

I was not able to reproduce the exact error shown in the issue but I think this approach should handle the issue:
* First create a preview site
* Next, navigate to https://developer.wordpress.com/docs/api/console/
* Run `wpcom/v2`, `POST`, `/jurassic-ninja`, `delete` 
* Fill in the site id for the preview site (you can find the site `id` in https://mc.a8c.com/tools/jn-report-card/ )
* Delete the preview site
* Navigate to the preview site in Studio
* Observe that it is now marked as expired
* Open the `Network` tab in the console
* Click on `Clear`
* Confirm that there are no errors and that the site clears
* Confirm that there is no network request 
* Confirm that the expires site is removed from `CLI config`

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/preview/delete.ts
+++ b/apps/cli/commands/preview/delete.ts
@@ -55,7 +55,6 @@ export async function runCommand( mode: Mode, host: string | undefined ): Promis
 			logger.reportSuccess( __( 'Validation successful' ), true );
 
 			logger.reportStart( LoggerAction.DELETE, __( 'Deleting…' ) );
-			console.error( ' WHAT IS GOING ON HERE?', isSnapshotExpired( snapshotToDelete ) );
 			if ( ! isSnapshotExpired( snapshotToDelete ) ) {
 				await deleteSnapshot( snapshotToDelete.atomicSiteId, token.accessToken );
 			}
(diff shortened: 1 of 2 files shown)


epeicher (approved):
Thanks @katinthehatsite for solving this! I have not been able to reproduce the issue, as I think the error is [correctly handled here](https://github.com/Automattic/studio/blob/211077d461cb21a48cf09f71682af02f04be43db/apps/cli/lib/api.ts#L131). This code prevents a request from being made if the site has already expired, so we are improving with this change. I am accepting this, and it can be landed. LGTM!

I have committed two lint fixes so once the CI passes, this could be merged.

apps/cli/commands/preview/delete.ts
```diff
@@ -51,7 +55,10 @@ export async function runCommand( mode: Mode, host: string | undefined ): Promis
 			logger.reportSuccess( __( 'Validation successful' ), true );
 
 			logger.reportStart( LoggerAction.DELETE, __( 'Deleting…' ) );
-			await deleteSnapshot( snapshotToDelete.atomicSiteId, token.accessToken );
+			console.error( ' WHAT IS GOING ON HERE?', isSnapshotExpired( snapshotToDelete ) );
```
fredrikekelund:
```suggestion
```
Leftover 🙂

epeicher: 😮 That was a big leftover, good catch!

katinthehatsite:
Thanks for cleaning up the logs @epeicher !

PR #2945, opened 2026-03-30
Harden screenshot capturing

fredrikekelund:
## Related issues

- Fixes STU-1486
- Fixes STU-1461

## How AI was used in this PR

For small and directed tasks. 

## Proposed Changes

1. Add an `emitLoadingEvent` argument to `captureSiteThumbnail` that suppresses `thumbnail-loading` events. This argument is true when the screenshot is updated in response to a `SITE_EVENTS.UPDATED` event. It's also inherited from the corresponding argument in `loadThemeDetails`.
2. Make the `waitForCapture` callback from `createScreenshotWindow` throw if the page returns an HTTP status code >= 500. For sites without custom domains, we get similar behavior for free, since the most common error is that `window.loadURL` fails because no server is listening on the port. For custom domains, that never happens because the custom domain proxy always listens on ports 80/443. 

## Testing Instructions

1. Have a site with a custom domain.
2. Start the site.
3. <kbd>Cmd+Tab</kbd> to quickly shift focus back and forth to Studio. Ensure that the Overview tab does not display a loading state when the window focus shifts, but only when the site starts. 
4. Stop the site.
5. Repeat the <kbd>Cmd+Tab</kbd> behavior. Ensure that no `Proxy error` screenshot is captured. 

## Pre-merge Checklist


apps/cli/commands/_events.ts
fredrikekelund, line 1: The changes in this file aren't related to the core purpose of this PR. I've just found the `subscribeSiteEvents` structure to be unnecessarily difficult to reason about (and I stumbled on it again while tracing the screenshot generation logic in this PR). My change simply moves the daemon bus event listeners to this file.

apps/cli/lib/proxy-server.ts
```diff
@@ -17,7 +17,7 @@ const proxy = httpProxy.createProxyServer();
 proxy.on( 'error', ( err, req, res ) => {
 	console.error( '[Proxy Error]', err.message );
 	if ( res && res instanceof http.ServerResponse ) {
-		res.writeHead( 500, { 'Content-Type': 'text/plain' } );
+		res.writeHead( 502, { 'Content-Type': 'text/plain' } );
```
fredrikekelund, line 20: Not important to the core purpose of this PR, but `502 Bad Gateway` better conveys the specific type of error.

apps/cli/wordpress-server-child.ts
```diff
@@ -438,7 +438,7 @@ const runWpCliCommand = sequential(
 			exitCode: await response.exitCode,
 		};
 	},
-	{ concurrent: 3, max: 100, deduplicateKey: ( args: string[] ) => args.join( ' ' ) }
+	{ concurrent: 3, max: 100, deduplicateKey: ( args ) => args.join( ' ' ) }
```
fredrikekelund, line 441: The `sequential` type definition was giving me grief. This is a leftover from when I was battling that for a while. It makes sense to remove the type since TS infers it from the type definitions of the arguments to the `sequential` callback.

apps/studio/src/ipc-handlers.ts
```diff
@@ -991,15 +991,17 @@ export function showItemInFolder( _event: IpcMainInvokeEvent, path: string ) {
...
 	id: string,
-	emitThemeDetailsLoadingEvent = true
+	emitLoadingEvent = true
 ): Promise< StartedSiteDetails[ 'themeDetails' ] > {
 	const server = SiteServer.get( id );
 	if ( ! server ) {
 		throw new Error( 'Site not found.' );
 	}
 
+	void captureSiteThumbnail( id, emitLoadingEvent );
```
fredrikekelund, line 1003: I made it so `captureSiteThumbnail` is triggered earlier because `server.getThemeDetails()` calls WP-CLI. When WP-CLI returns a response, the CLI generates a `SITE_EVENT.UPDATED` event to Studio, which triggers another `captureSiteThumbnail` call. 

In the end, I made it so neither of the `captureSiteThumbnail` calls triggers a `thumbnail-loading` event, but previously, the order in which things were called triggered two such events.

apps/studio/src/lib/capture-site-thumbnail.ts
```diff
@@ -7,14 +7,17 @@ import { getSiteThumbnailPath } from 'src/storage/paths';
 // Capture and cache a site's thumbnail. Uses sequential with deduplicateKey to
 // prevent concurrent BrowserWindow creation for the same site and to serialize
 // captures across all sites (one at a time).
-export const captureSiteThumbnail = sequential(
-	async ( id: string ): Promise< void > => {
+export const captureSiteThumbnail = sequential< [ string, boolean? ], void >(
```
fredrikekelund, line 10: This type definition is important because the type inference in `sequential.ts` cannot handle a callback with two arguments (where the second one is optional) and a `deduplicateKey` function that takes just a single argument.

apps/studio/src/screenshot-window.ts
fredrikekelund, line 1: Again, sites _without_ custom URLs will encounter network errors here because no server is listening on the target port. For sites _with_ custom URLs, there's always the custom domain proxy listening on port 80/443, which is why we need the HTTP status code check as an additional safety measure.

fredrikekelund:
I've run multiple rounds of local AI reviews against this branch and multiple rounds of manual testing. I feel confident in the change, so I'll go ahead and land the PR.

PR #2949, opened 2026-03-30
Studio: Change test mocks to use memfs

katinthehatsite:
## Related issues

Fixes STU-1467

## How AI was used in this PR

The AI was used to complete the migration. Then, I used the AI to review the work the AI did.

## Proposed Changes

This PR ensures that we switch to `memfs` in test mocks for `fs` with `vitest`. 
## Testing Instructions

* Confirm all the checks are passing in CI 

## Pre-merge Checklist


apps/studio/src/lib/tests/is-installed.test.ts
```diff
@@ -20,16 +20,14 @@ type ReaddirSyncStrings = (
...
-	const existsSync = vi.fn< ( path: PathLike ) => boolean >();
-	const readdirSync = vi.fn< ReaddirSyncStrings >();
-
-	return {
-		default: { existsSync, readdirSync },
-		existsSync,
-		readdirSync,
-	};
-} );
+vi.mock( 'fs', () => ( {
```
epeicher, line 23: This still uses the old approach, any particular reason?

epeicher (approved):
Thanks @katinthehatsite for applying best practices! I have executed the tests locally, and they pass, also the CI passes as expected. Changes LGTM!

Image

I have left a question but it's not a blocker so accepting this

package-lock.json
epeicher, line 1: In my local after making sure I'm using node version `v24.13.1`, I see some differences in this file after running `npm install` (all due to `peer: true` being deleted). Could you please confirm that you have executed `npm install` with the proper node version? This is just to avoid having changes to `package-lock.json` for other installations.

katinthehatsite: I see, yes, indeed, these changes should not be there. I made a commit to adjust this 👍 Nice catch!

apps/studio/src/lib/tests/is-installed.test.ts
katinthehatsite: Yeah I tried to use the mocks in here as well but this test as a special case. This test needs `resetModules` because is-installed.ts reads process.platform when it first loads. To test both macOS and Windows behavior, the test changes `process.platform` and then re-imports the module so it picks up the new value.

The issue is a side effect of that re-import: with the auto-mock approach, re-importing also re-runs __mocks__/fs.ts, which creates a fresh set of mock functions. The test is still holding the old set, so they go out of sync.

The inline factory doesn't have this problem because vitest caches its result separately from the module cache.

Claude gave me this metaphor: Think of it like phone numbers: the test saved Mock A's number and keeps calling it. But after resetModules, the auto-mock gave the code a new phone (Mock B). The test is still dialing Mock A — nobody picks up on the other end.

PR #2951, opened 2026-03-31
Fix Claude auth check to parse JSON instead of string matching

epeicher:
## Related issues

- Related to STU-1444

## How AI was used in this PR

AI was used to identify the fix and update the implementation and tests. I reviewed all that and iterated multiple times to identify the root cause of the code not failing.

## Proposed Changes

- Replace fragile string matching in `hasClaudeCodeAuth()` with JSON parsing of `claude auth status` output
- Check `status.loggedIn === true` instead of searching for substrings like "authenticated" and "not"
- Update test mocks to return JSON (`{ loggedIn: true }`) instead of the string `"Authenticated"`

### Context

The previous implementation used `output.toLowerCase().includes('authenticated') || !output.toLowerCase().includes('not')` which has an incorrect `||` (should be `&&`). However, the bug never surfaced in practice because:

1. **Logged out**: `claude auth status` exits with code 1, so `execFileSync` throws and the `catch` block returns `false` — the string logic is never reached.
2. **Logged in**: The output is JSON (e.g. `{"loggedIn": true}`), which contains "authenticated" is false but `!includes('not')` is true, so the `||` returns `true` by coincidence.

This fix makes the logic correct and robust by parsing the JSON directly, rather than relying on accidental behavior.

## Testing Instructions

- Run `npm test -- apps/cli/ai/tests/auth.test.ts` — all 15 tests should pass
- Typecheck passes: `npm run typecheck`

## Pre-merge Checklist


apps/cli/ai/tests/auth.test.ts
```diff
@@ -66,7 +66,9 @@ describe( 'AI auth helpers', () => {
 	} );
 
 	it( 'uses Claude auth when the Claude auth provider is selected', async () => {
-		vi.mocked( childProcess.execFileSync ).mockReturnValue( 'Authenticated' as never );
+		vi.mocked( childProcess.execFileSync ).mockReturnValue(
+			JSON.stringify( { loggedIn: true } ) as never
```
katinthehatsite: I think we can avoid forcing type in these tests with something like:

```
vi.mocked( childProcess.execFileSync ).mockReturnValue(
    Buffer.from( JSON.stringify( { loggedIn: true } ) )
);
```

This should be a more accurate representation in the test.

katinthehatsite (commented):
The fix looks good to me but I left one suggestion for the tests and type safety which I think we should consider.

epeicher:
Closing this as the affected code has been removed as part of https://github.com/Automattic/studio/pull/2977

Issue #137, opened 2024-05-15
Feature Request: Honor internal .gitignore files before pushing to the demo


bacoords:
### What

When I push to the demo site it wants to push everything but if I have a `node_modules` folder or some other folder that I've included in a `.gitignore` file, it would be great if studio recognized the file and ignored it. 

### Why

Developers may want to work on custom block plugins or something that has a build process and quickly share demos without having to manually delete all their node_modules. I'll also note that node_modules folders often exceed the 100MB limit and it's just good to conserve the bandwidth.

### How

Ability to either add a `.gitignore` (or similar) file OR a recursive check through the `wp-content` folder for `.gitignore` files.

danielbachhuber:
Great suggestion, @bacoords — thanks!

faisalahammad:
+1

wojtekn:
It's an interesting suggestion.

However, in terms of implementation, should we rely on the GIT's `.gitignore` file or have another way to let Studio which files or directories should be ignored?

bacoords:
> However, in terms of implementation, should we rely on the GIT's .gitignore file or have another way to let Studio which files or directories should be ignored?

This is a pretty common approach. Some tools include their own `.*ignore` file specific to their CI/CD workflow. For example, [WPVIP lets you add](https://docs.wpvip.com/code-deployment/default-deployment/build-and-deploy/ci-cd/) a `.deployignore` file to specify things you want in the repo but not on the server.

Jameswlepage:
+1 - My main use case here is the node modules example as well. But I also have several plugins that I don't want to sync and only keep local.

ivan-ottinger:
Hey everyone! 👋🏼 We have added support for `.deployignore` files. Now you can control which files get excluded when syncing your local Studio site to:
- a Preview site: https://developer.wordpress.com/docs/developer-tools/studio/preview-sites/#4-exclude-files-with-deployignore
- a WordPress.com and Pressable site: https://developer.wordpress.com/docs/developer-tools/studio/sync/#8-exclude-files-with-deployignore

Please let us know if you have any questions.

PR #2955, opened 2026-04-01
Add .deployignore support to Site Sync

ivan-ottinger:
## Related issues

- Related to STU-16

## How AI was used in this PR

AI-assisted implementation with human iteration and review of all changes, architecture decisions, and edge cases. I have manually tested the changes as well.

## Proposed Changes

Follow-up to #2924, which added `.deployignore` support for Preview Sites. This PR extends it to Site Sync (push to WordPress.com / Pressable).

- Extend `createDeployIgnoreFilter` to accept a `basePatterns` argument so the sync path can pass Studio-internal exclusions (`database`, `db.php`, `debug.log`, `sqlite-database-integration`, `cache`) alongside `DEPLOY_IGNORE_DEFAULTS`
- **Sync tree UI**: Update `listLocalFileTree` and `shouldExcludeFromSync` to use the deploy-ignore filter. Files matching `.deployignore` patterns are hidden from the sync dialog
- **Sync push**: CLI's `DefaultExporter` accepts an optional `ignoreFilter` in `ExportOptions`, used by `addWpContent()` alongside existing hardcoded exclusions (`isExactPathExcluded`, `isPathExcludedByPattern`). The `studio export` command gains a hidden `--apply-deploy-ignore` flag; Studio's sync push sets it when shelling to the CLI. Regular site exports (Studio UI, direct `studio export` CLI) do not apply the filter, preserving pre-existing backup behavior.
- **meta.json filtering**: Filter the plugin and theme lists in `meta.json` against the deploy-ignore filter. Without this, the server-side import would reinstall excluded plugins from WordPress.org even though their files were excluded from the archive
- Rename `SYNC_EXCLUSIONS` to `SYNC_IGNORE_DEFAULTS` — a superset of `DEPLOY_IGNORE_DEFAULTS` plus Studio-internal items (`database`, `db.php`, `debug.log`, `sqlite-database-integration`, `cache`).
- Remove redundant `.replace(/\\/g, '/')` path normalization in `fs-utils.ts` — the `ignore` library handles Windows backslash conversion internally
- Add `ignore` as a direct dependency in `apps/cli` and `apps/studio` (previously relied on transitive resolution)

## Testing Instructions

> [!TIP]
> It is probably easier to ask your AI assistant to perform the steps 2 and 3 for you, so you don't have to create the testing files manually.

1. Create a local site in Studio
2. Add a `.deployignore` file at the site root (e.g., `~/Studio/my-site/.deployignore`) with the following content:
   ```
   # Ignore vendor dependencies in my-plugin, except important-lib
   wp-content/plugins/my-plugin/vendor/*
   !wp-content/plugins/my-plugin/vendor/important-lib

   # Ignore all log files
   *.log

   # Ignore old uploads
   wp-content/uploads/2024

   # Ignore a specific plugin
   wp-content/plugins/wordpress-seo
   ```
3. Create matching test files/folders in the site's `wp-content`:
   ```
   # Plugin with build artifacts
   mkdir -p wp-content/plugins/my-plugin/vendor/some-lib
   mkdir -p wp-content/plugins/my-plugin/vendor/important-lib
   echo '<?php // plugin entry' > wp-content/plugins/my-plugin/my-plugin.php
   echo '<?php // vendor file' > wp-content/plugins/my-plugin/vendor/some-lib/autoload.php
   echo '<?php // important' > wp-content/plugins/my-plugin/vendor/important-lib/loader.php
   echo 'build log' > wp-content/plugins/my-plugin/build.log

   # .git and node_modules inside the plugin (built-in defaults)
   mkdir -p wp-content/plugins/my-plugin/.git
   echo 'ref: refs/heads/main' > wp-content/plugins/my-plugin/.git/HEAD
   mkdir -p wp-content/plugins/my-plugin/node_modules/some-package
   echo '{"name":"some-package"}' > wp-content/plugins/my-plugin/node_modules/some-package/package.json

   # Old and new uploads
   mkdir -p wp-content/uploads/2024
   echo 'old photo' > wp-content/uploads/2024/photo.jpg
   mkdir -p wp-content/uploads/2025
   echo 'new photo' > wp-content/uploads/2025/photo.jpg
   ```
4. Install the `wordpress-seo` plugin on the site
5. Connect the local site to a target remote site (WordPress.com or Pressable)
6. Consider checking the target site to confirm it does not yet include any directories/files we want to ignore during the sync.
7. Open the Sync dialog and verify:
   - `wordpress-seo` plugin is hidden from the tree
   - `akismet` and `my-plugin` are still visible
   - `uploads/2024` directory is hidden
   - `uploads/2025` is still visible
8. Push to WordPress.com / Pressable and verify (e.g through a SSH connection to the WoA site):
   - `vendor/some-lib/` is excluded from the push
   - `vendor/important-lib/` is included (negation pattern)
   - `*.log` files are excluded
   - `uploads/2024/` is excluded, `uploads/2025/` is present
   - `wordpress-seo` plugin is excluded and not reinstalled by the server
9. Test the CLI push with the same `.deployignore` file:
   ```
   npm run cli:build && node apps/cli/dist/cli/main.mjs push --path {path-to-your-studio-site}
   ```
10. Verify the same exclusions apply as in step 7 & step 8 (check the target site again after the CLI push).
11. Verify that `.git` and `node_modules` inside `my-plugin` are excluded by built-in defaults
12. Verify that site export (backup) is unaffected by `.deployignore` — the filter is only applied during sync push, not when using Studio UI "Export site" or running `studio export` directly

## Pre-merge Checklist


tools/common/package.json
```diff
@@ -13,7 +13,7 @@
 		"cross-port-killer": "^1.4.0",
 		"date-fns": "^3.3.1",
 		"fast-deep-equal": "^3.1.3",
-		"ignore": "^7.0.5",
+		"ignore": "^5.3.2",
```
ivan-ottinger, line 16: I have made this change because other packages we are already using in Studio rely on the `ignore` v5. This version provides everything we need.

Using v7 would require several workarounds that would make the code less maintainable and more difficult to read

apps/studio/src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -292,8 +303,18 @@ export class DefaultExporter extends EventEmitter implements Exporter {
...
+					( p: StudioJsonPluginOrTheme ) =>
+						! this.options.deployIgnore!.ignores( `wp-content/plugins/${ p.name }` )
+			  )
+			: plugins;
+		studioJson.themes = this.options.deployIgnore
+			? themes.filter(
+					( t: StudioJsonPluginOrTheme ) =>
+						! this.options.deployIgnore!.ignores( `wp-content/themes/${ t.name }` )
+			  )
+			: themes;
```
ivan-ottinger, line 315: These changes are required because otherwise the .deployignored plugins/themes would be reinstalled on the target site.

apps/studio/src/modules/sync/constants.ts
```diff
@@ -1,11 +1,11 @@
-export const SYNC_EXCLUSIONS = [
+/**
+ * Studio-internal exclusions that should always be excluded from sync,
+ * in addition to the base deploy-ignore defaults.
```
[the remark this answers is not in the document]
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2955/commits/fd6e94c1593c87ac2ebcd16e4a36a65311089e15.

tools/common/lib/deploy-ignore.ts
```diff
@@ -16,9 +16,16 @@ const DEPLOY_IGNORE_FILENAME = '.deployignore';
...
  */
-export async function createDeployIgnoreFilter( rootPath: string ): Promise< Ignore > {
+export async function createDeployIgnoreFilter(
+	rootPath: string,
+	additionalDefaults?: string[]
+): Promise< Ignore > {
 	const ig = ignore().add( DEPLOY_IGNORE_DEFAULTS );
+	if ( additionalDefaults ) {
+		ig.add( additionalDefaults );
+	}
```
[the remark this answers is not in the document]
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2955/changes/edfa8e533e33506429ff6cde42cd2836fd055c2d.

apps/studio/src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -226,14 +231,20 @@ export class DefaultExporter extends EventEmitter implements Exporter {
...
 							return false;
 						}
 						return entry;
 					} );
 				} else {
-					if ( this.isExactPathExcluded( archivePath ) ) {
+					if (
+						this.isExactPathExcluded( archivePath ) ||
+						this.options.deployIgnore?.ignores( archivePath.replace( /\\/g, '/' ) )
+					) {
```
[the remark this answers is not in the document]
ivan-ottinger: The `deploy-ignore` filter itself is thoroughly tested in `deploy-ignore.test.ts` (10 test cases covering defaults, patterns, negation, globs, and additional defaults). The integration in `addWpContent()` is a straightforward `ignores()` check — the same pattern used in the Preview Sites PR (#2924) which was reviewed and merged without exporter-level tests.

apps/studio/src/lib/import-export/export/exporters/default-exporter.ts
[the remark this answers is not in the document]
ivan-ottinger: The filtering is a one-liner `.filter()` that delegates to the same tested `ignores()` method. Adding a test would require mocking the entire exporter setup (`SiteServer`, `executeWpCliCommand`, temp files), which is disproportionate for verifying that `.filter()` works.

fredrikekelund:
Noting that https://github.com/Automattic/studio/pull/3010 changed a lot of the sync logic on trunk. The CLI now also supports sync, and a lot of the internals moved to `tools/common`. We should update this PR accordingly

tools/common/lib/deploy-ignore.ts
fredrikekelund:
```suggestion
	defaults: string[] = DEPLOY_IGNORE_DEFAULTS
): Promise< Ignore > {
	const ig = ignore().add( defaults );
```
Having one `defaults` array would make things clearer.

tools/common/lib/sync/constants.ts
```diff
@@ -1,11 +1,12 @@
-export const SYNC_EXCLUSIONS = [
+/**
+ * Studio-internal exclusions excluded from sync by default,
+ * in addition to the base deploy-ignore defaults. These are pre-seeded
+ * but can be overridden via negation patterns in .deployignore.
+ */
+export const SYNC_ADDITIONAL_DEFAULTS = [
 	'database',
```
fredrikekelund:
```suggestion
export const SYNC_ADDITIONAL_DEFAULTS = [
	...DEPLOY_IGNORE_DEFAULTS,
	'.*',
	'database',
```
Two things here:

1. We can mix in the `DEPLOY_IGNORE_DEFAULTS` array. Seems sensible to reuse that definition. 
2. `shouldExcludeFromSync` is super thin now. The `ignore` module does most of the work. I say let's lean fully into that setup and add a `.*` pattern to this array that replaces the `if ( itemName.startsWith( '.' ) )` logic in `shouldExcludeFromSync`

tools/common/package.json
fredrikekelund: What kinds of workarounds..?

apps/studio/src/modules/sync/lib/tree-utils.ts
```diff
@@ -1,17 +1,13 @@
 import { TreeNode } from 'src/components/tree-view';
-import { SYNC_EXCLUSIONS } from '../constants';
 import type { RawDirectoryEntry } from '../types';
+import type { Ignore } from 'ignore';
 
-export const shouldExcludeFromSync = ( itemName: string ): boolean => {
+export const shouldExcludeFromSync = ( relativePath: string, deployIgnore: Ignore ): boolean => {
```
fredrikekelund: I think we should remove this function and call `deployIgnore.ignores` directly in `listLocalFileTree`. See my comment in `apps/studio/src/modules/sync/constants.ts` for more details

apps/studio/src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -216,6 +217,10 @@ export class DefaultExporter extends EventEmitter implements Exporter {
 					continue;
 				}
 
+				if ( this.options.deployIgnore?.ignores( archivePath.replace( /\\/g, '/' ) ) ) {
```
fredrikekelund: Does `ignore` not handle Windows path separators natively?

apps/studio/src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -226,14 +231,20 @@ export class DefaultExporter extends EventEmitter implements Exporter {
 						);
 						if (
 							this.isExactPathExcluded( entryPathRelativeToArchiveRoot ) ||
```
fredrikekelund, line 233: WDYT about merging the `isExactPathExcluded` functionality and `options.deployIgnore` into one, @ivan-ottinger?

`isPathExcludedByPattern` seems a little more difficult because of the `cache` pattern. It seems overly broad to exclude anything named `cache`, so the `isPathExcludedByPattern` logic to check specifically for directories with that name is a pretty nice thing to have.

fredrikekelund (commented):
As I mentioned in https://github.com/Automattic/studio/pull/2955#issuecomment-4211827023, a lot of the sync internals were moved around when we added sync support to the CLI. We should update this PR to reflect that.

Also, we currently have two copies of the exporter implementation: one for the CLI and another for the app. We should make sure this PR updates both places.

Lastly, I left some more detailed comments on the implementation.

tools/common/lib/deploy-ignore.ts
ivan-ottinger: Sounds good to me. 👍🏼 I have made the change in https://github.com/Automattic/studio/pull/2955/changes/2d2216b2c57c616c9b5ed54a8cf554028468859e.

tools/common/lib/sync/constants.ts
ivan-ottinger: > 1. We can mix in the DEPLOY_IGNORE_DEFAULTS array. Seems sensible to reuse that definition.

Yes, that sounds good to me. I don't see any reason why not to simplify this and I made the change in https://github.com/Automattic/studio/pull/2955/commits/3c7c651d89b0ca62664c8fc.1c1703aed1be72cb5

> 2. shouldExcludeFromSync is super thin now. The ignore module does most of the work. I say let's lean fully into that setup and add a .* pattern to this array that replaces the if ( itemName.startsWith( '.' ) ) logic in shouldExcludeFromSync

I was considering adding `.*` pattern previously, but then decided against it. Adding `.*` to the defaults would silently stop pushing dotfiles. On trunk they're hidden from the UI but still pushed inside selected directories. That could be a subtle regression that's hard to diagnose from the user side. I am open to making the change though. What do you think?

Maybe we could consider this change separately (outside of this PR).

tools/common/package.json
ivan-ottinger: > What kinds of workarounds..?

Before this change, tools/common used ignore v7 while the root node_modules has ignore v5 (from eslint, wp-playground). That caused TypeScript to see two different Ignore interfaces. v7 adds checkIgnore and PatternParams - so passing a v7 instance where v5 was expected failed typechecking.

Here is the workaround:

```diff
diff --git a/apps/cli/lib/import-export/export/types.ts b/apps/cli/lib/import-export/export/types.ts
index 1dea9f1d..4693f32f 100644
--- a/apps/cli/lib/import-export/export/types.ts
+++ b/apps/cli/lib/import-export/export/types.ts
@@ -1,7 +1,7 @@
 import { SiteData } from 'cli/lib/cli-config/core';
+import type { Ignore } from '@studio/common/lib/deploy-ignore';
 import type archiver from 'archiver';
 import type { EventEmitter } from 'events';
-import type { Ignore } from 'ignore';
 
 export interface ExportOptions {
 	site: SiteData;
diff --git a/apps/cli/lib/sync-file-tree.ts b/apps/cli/lib/sync-file-tree.ts
index 6b03b7ca..4ae5cced 100644
--- a/apps/cli/lib/sync-file-tree.ts
+++ b/apps/cli/lib/sync-file-tree.ts
@@ -3,8 +3,8 @@ import nodePath from 'path';
 import { createDeployIgnoreFilter } from '@studio/common/lib/deploy-ignore';
 import { SYNC_IGNORE_DEFAULTS } from '@studio/common/lib/sync/constants';
 import { shouldExcludeFromSync, shouldLimitDepth } from '@studio/common/lib/sync/tree-utils';
+import type { Ignore } from '@studio/common/lib/deploy-ignore';
 import type { RawDirectoryEntry } from '@studio/common/types/sync-tree';
-import type { Ignore } from 'ignore';
 
 export async function listLocalFileTree(
 	sitePath: string,
diff --git a/apps/studio/src/ipc-handlers.ts b/apps/studio/src/ipc-handlers.ts
index 708343a8..d27d9958 100644
--- a/apps/studio/src/ipc-handlers.ts
+++ b/apps/studio/src/ipc-handlers.ts
@@ -118,7 +118,7 @@ import {
 import { Blueprint } from 'src/stores/wpcom-api';
 import { captureSiteThumbnail } from './lib/capture-site-thumbnail';
 import type { RawDirectoryEntry } from '@studio/common/types/sync-tree';
-import type { Ignore } from 'ignore';
+import type { Ignore } from '@studio/common/lib/deploy-ignore';
 import type { WpCliResult } from 'src/site-server';
 
 export {
diff --git a/apps/studio/src/lib/import-export/export/types.ts b/apps/studio/src/lib/import-export/export/types.ts
index 7794f922..45ef67aa 100644
--- a/apps/studio/src/lib/import-export/export/types.ts
+++ b/apps/studio/src/lib/import-export/export/types.ts
@@ -1,6 +1,6 @@
+import type { Ignore } from '@studio/common/lib/deploy-ignore';
 import type archiver from 'archiver';
 import type { EventEmitter } from 'events';
-import type { Ignore } from 'ignore';
 
 export interface ExportOptions {
 	site: SiteDetails;
diff --git a/tools/common/lib/deploy-ignore.ts b/tools/common/lib/deploy-ignore.ts
index 605b5212..62960d21 100644
--- a/tools/common/lib/deploy-ignore.ts
+++ b/tools/common/lib/deploy-ignore.ts
@@ -3,6 +3,12 @@ import path from 'path';
 import ignore, { Ignore } from 'ignore';
 import { isErrnoException } from './is-errno-exception';
 
+// Re-exported so consumers outside @studio/common resolve to the same v7 type
+// that @studio/common uses internally. The root node_modules hoists v5 (via
+// eslint, @wp-playground/*), so importing `Ignore` directly from 'ignore' in
+// apps/studio or apps/cli would resolve to v5 and cause type mismatches.
+export type { Ignore };
+
 /**
  * Default patterns excluded from deploys. These are pre-seeded
  * but can be overridden via negation patterns in .deployignore.
diff --git a/tools/common/package.json b/tools/common/package.json
index a1394a77..ff7d33c3 100644
--- a/tools/common/package.json
+++ b/tools/common/package.json
@@ -13,7 +13,7 @@
 		"cross-port-killer": "^1.4.0",
 		"date-fns": "^3.3.1",
 		"fast-deep-equal": "^3.1.3",
-		"ignore": "^5.3.2",
+		"ignore": "^7.0.5",
 		"lockfile": "^1.0.4",
 		"wpcom": "^7.1.1",
 		"wpcom-xhr-request": "^1.3.0",

```

Since v5 is already hoisted in the dependency tree via eslint and wp-playground, I chose to align  with it rather than maintain the workaround.

apps/studio/src/modules/sync/lib/tree-utils.ts
ivan-ottinger: I am happy to follow up on this one if we agree on the `.*` inclusion in https://github.com/Automattic/studio/pull/2955#discussion_r3086533884.

apps/studio/src/lib/import-export/export/exporters/default-exporter.ts
ivan-ottinger: Good catch! I remember I have been adding these replacements on purpose when working on the .deployignore for preview sites, but cannot remember the details. Regardless, looking at the `ignore` source, it should handle the Windows paths natively. I will remove the extra replacement and retest everything on Windows.

ivan-ottinger: Updated in https://github.com/Automattic/studio/pull/2955/commits/69ce7801863171d55044d11b3ec7c759574207d7

ivan-ottinger:
I will continue with updates and testing tomorrow with fresh brain.

tools/common/lib/sync/constants.ts
fredrikekelund: Good catch 👍 Honestly, I would probably lean towards displaying dotfiles in the selective sync UI and only hiding stuff that we explicitly disallow from being included. Otherwise, users might be surprised one way or another. 

Dotfiles might even be particularly risky now that I think about it, as it's not uncommon for them to contain security-sensitive config details. 

Not sure who has worked on this before… Maybe @katinthehatsite or @sejas has some input, too.

apps/studio/src/lib/import-export/export/exporters/default-exporter.ts
ivan-ottinger: I agree it would be cleaner and more readable - having one filtering system.

The `isExactPathExcluded` applies to all exports (including local backup), which doesn't have a deployIgnore filter at the moment. Moving those paths into the filter would mean constructing one.

Considering the size of this PR already, maybe it would be better to address this in a separately. What do you think?

tools/common/lib/sync/constants.ts
```diff
@@ -3,14 +3,22 @@ export const SYNC_MAX_STALLED_ATTEMPTS = 200;
...
 
-export const SYNC_EXCLUSIONS = [
+/**
+ * Base patterns excluded from sync. These are pre-seeded but can be
+ * overridden via negation patterns in .deployignore.
+ */
+export const SYNC_IGNORE_DEFAULTS = [
+	// Duplicated from DEPLOY_IGNORE_DEFAULTS in tools/common/lib/deploy-ignore.ts
+	// so this file stays free of Node imports and can be used in the renderer.
+	// A unit test enforces that this list is a superset.
```
ivan-ottinger, line 13: I haven't yet found a cleaner way to have all the defaults merged together. The unit test is in place to ensure we keep the lists in sync.

tools/common/lib/sync/constants.ts
ivan-ottinger: Good points. Curious to hear what Kat and Antonio think.

Just to check I understand correctly, you are suggesting we would:
- display dotfiles in the sync selector/file browser
- sync them only when the user manually selects them or runs the full sync

I think that's a reasonable approach.

apps/studio/src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -226,14 +231,18 @@ export class DefaultExporter extends EventEmitter implements Exporter {
...
-							this.isPathExcludedByPattern( fullEntryPathOnDisk )
+							this.isPathExcludedByPattern( fullEntryPathOnDisk ) ||
+							this.options.deployIgnore?.ignores( entryPathRelativeToArchiveRoot )
 						) {
```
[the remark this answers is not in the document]
ivan-ottinger: The `ignore` library handles Windows path separators natively — on `process.platform === 'win32'` it installs a `makePosix` converter that replaces `\` with `/` before every `.ignores()` check (see `node_modules/ignore/index.js` lines 611-636). We previously had `.replace(/\\/g, '/')` calls here but removed them after verifying this behavior. We will also test the PR changes on Windows before merging.

apps/cli/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -214,6 +215,10 @@ export class DefaultExporter extends EventEmitter implements Exporter {
 					continue;
 				}
 
+				if ( this.options.deployIgnore?.ignores( archivePath ) ) {
+					continue;
+				}
+
 				const stat = await fsPromises.stat( fullPath );
```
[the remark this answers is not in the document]
ivan-ottinger: The `ignore` library handles Windows path separators natively — on `process.platform === 'win32'` it installs a `makePosix` converter that replaces `\` with `/` before every `.ignores()` check (see `node_modules/ignore/index.js` lines 611-636). We previously had `.replace(/\\/g, '/')` calls here but removed them after verifying this behavior. We will also test the PR changes on Windows before merging.

apps/cli/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -224,14 +229,18 @@ export class DefaultExporter extends EventEmitter implements Exporter {
 						);
 						if (
 							this.isExactPathExcluded( entryPathRelativeToArchiveRoot ) ||
-							this.isPathExcludedByPattern( fullEntryPathOnDisk )
+							this.isPathExcludedByPattern( fullEntryPathOnDisk ) ||
+							this.options.deployIgnore?.ignores( entryPathRelativeToArchiveRoot )
 						) {
```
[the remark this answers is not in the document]
ivan-ottinger: The `ignore` library handles Windows path separators natively — on `process.platform === 'win32'` it installs a `makePosix` converter that replaces `\` with `/` before every `.ignores()` check (see `node_modules/ignore/index.js` lines 611-636). We previously had `.replace(/\\/g, '/')` calls here but removed them after verifying this behavior. We will also test the PR changes on Windows before merging.

apps/studio/src/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -216,6 +217,10 @@ export class DefaultExporter extends EventEmitter implements Exporter {
 					continue;
 				}
 
+				if ( this.options.deployIgnore?.ignores( archivePath ) ) {
+					continue;
+				}
+
 				const stat = await fs.promises.stat( fullPath );
```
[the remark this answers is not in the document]
ivan-ottinger: The `ignore` library handles Windows path separators natively — on `process.platform === 'win32'` it installs a `makePosix` converter that replaces `\` with `/` before every `.ignores()` check (see `node_modules/ignore/index.js` lines 611-636). We previously had `.replace(/\\/g, '/')` calls here but removed them after verifying this behavior. We will also test the PR changes on Windows before merging.

ivan-ottinger:
Alright, I have addressed the feedback and tested the changes on both macOS and Windows (pushing through the UI and the Studio CLI. Everything is running well, the files are getting excluded as described in the `.deployignore`.

From what I see [this](https://github.com/Automattic/studio/pull/2955#discussion_r3078912530) is the main remaining change to agree on. We can wait for the response from Antonio and/or Kat.

Apart from that, the PR will be ready for another review/test round.

tools/common/lib/sync/constants.ts
fredrikekelund: I'm suggesting we remove the `if ( itemName.startsWith( '.' ) ) return true` condition from the `shouldExcludeFromSync` so they're displayed in the file selector. 

The way I understand it, dotfiles are included in the actual sync payload right now if they're in a selected directory, but we don't display them in the file selector (which makes them easy to overlook and may even give users the false belief that they won't be included in the sync).

tools/common/lib/sync/constants.ts
fredrikekelund: Ah, I see. Let's move the `DEPLOY_IGNORE_DEFAULTS` definition to a file that doesn't import any modules that expect to not run in the browser and then mix in the array with `...DEPLOY_IGNORE_DEFAULTS` here.

ivan-ottinger: I actually took this approach previously, but then changed it. Addressed in https://github.com/Automattic/studio/pull/2955/commits/f20bc03d3c486453d4f5575d95d940e7c3b646ef.

tools/common/lib/sync/constants.ts
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2955/commits/3b6c91e6451eca92245a384074e5773ba96f6642.

ivan-ottinger: After digging deeper I learned that the restore skips dotfiles automatically.

Because of that I reverted the change in https://github.com/Automattic/studio/pull/2955/commits/e0b2e8123199bc8cb81bf73ef30398c3e0541d11. Dotfiles are not displayed in the file browser anymore (we are back at the trunk behavior there).

ivan-ottinger:
Hey @fredrikekelund 👋🏼🙂 I think we should be good merging this one. Feel free to let me know what you think.

fredrikekelund:
Taking another look at this now

tools/common/lib/sync/constants.ts
fredrikekelund: Great catch 👍 Can we add a comment to document this?

tools/common/lib/deploy-ignore.ts
fredrikekelund, line 1: Sorry for not thinking of this sooner, but I feel like it'd make sense to look for a `.gitignore` file here, too. Thoughts, @ivan-ottinger? I guess this could also be a follow-up PR.

apps/cli/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -284,8 +298,18 @@ export class DefaultExporter extends ImportExportEventEmitter implements Exporte
...
+					( p: StudioJsonPluginOrTheme ) =>
+						! this.options.deployIgnore!.ignores( `wp-content/plugins/${ p.name }` )
+			  )
+			: plugins;
+		studioJson.themes = this.options.deployIgnore
+			? themes.filter(
+					( t: StudioJsonPluginOrTheme ) =>
+						! this.options.deployIgnore!.ignores( `wp-content/themes/${ t.name }` )
+			  )
+			: themes;
```
fredrikekelund: What we should really do, instead of asserting the array entry types, is use zod to parse the JSON returned by WP-CLI in `getSitePlugins` and `getSiteThemes`.

tools/common/package.json
fredrikekelund: The solution is to install `ignore` as a dependency in `apps/cli` and `apps/studio` as well. Right now, when `apps/cli/lib/import-export/export/types.ts` and `apps/studio/src/lib/import-export/export/types.ts` import the `Ignore` type, they're only able to do it because `ignore` is an indirect dependency.

tools/common/lib/fs-utils.ts
```diff
@@ -32,9 +32,7 @@ export function calculateDirectorySizeForArchive(
 				await Promise.all(
 					files.map( async ( file ) => {
 						const filePath = path.join( dirPath, file.name );
-						const fileRelativeToRoot = path
-							.relative( directoryPath, filePath )
-							.replace( /\\/g, '/' );
+						const fileRelativeToRoot = path.relative( directoryPath, filePath );
 						const ignorePath = pathPrefix
 							? `${ pathPrefix }/${ fileRelativeToRoot }`
```
```suggestion
						const ignorePath = pathPrefix
							? path.join( pathPrefix, fileRelativeToRoot )
```
Even if `ignore` handles Windows-style separators internally, we should ensure the provided paths use consistent separators.

tools/common/lib/sync/tree-utils.ts
```diff
@@ -1,16 +1,12 @@
-import { SYNC_EXCLUSIONS } from './constants';
 import type { SyncOption } from '@studio/common/types/sync';
+import type { Ignore } from 'ignore';
 
-export const shouldExcludeFromSync = ( itemName: string ): boolean => {
+export const shouldExcludeFromSync = ( relativePath: string, deployIgnore: Ignore ): boolean => {
+	const itemName = relativePath.split( '/' ).pop() || '';
```
fredrikekelund:
```suggestion
	const itemName = path.basename( relativePath. );
```
Let's use the standard library here. `path.basename` handles both Posix-style and Windows-style separators (to be more specific, the Windows implementation handles both separator types, whereas the Posix implementation handles only Posix-style separators).

apps/cli/lib/import-export/export/types.ts
```diff
@@ -8,6 +9,7 @@ export interface ExportOptions {
 	phpVersion: string;
 	splitDatabaseDumpByTable?: boolean;
 	specificSelectionPaths?: string[];
+	deployIgnore?: Ignore;
```
fredrikekelund:
```suggestion
	ignoreFilter?: Ignore;
```
Optional nit, but the exporter implementations don't need to know how we construct this filter, so we could use a more generic name. If we change this, we should also change `apps/studio/src/lib/import-export/export/types.ts`

fredrikekelund (approved):
LGTM 👍 I left a couple of comments that'd be nice to address before landing this, not least the `ignore` dependency management.

Also interested to hear what you think about reading `.gitignore`, too, but that can easily be a follow-up.

tools/common/lib/sync/constants.ts
ivan-ottinger: Let me correct myself here. Turns out this is a bit more nuanced that I saw initially. Dotfiles are skipped only in the root directories. So if the dotfile is in a specific plugin, it gets restored. 

Regarding the existing logic we have, everything is correct. I have added code comments in https://github.com/Automattic/studio/pull/2955/changes/5491082992d95357e421fe024e60d3ae80836acd for future reference.

tools/common/package.json
ivan-ottinger: Thanks! Addressed in https://github.com/Automattic/studio/pull/2955/commits/7192dd380b3c5ee26bc8cf28f79ada61192c938f.

tools/common/lib/sync/tree-utils.ts
ivan-ottinger: I tried this, but switching to `path.basename` would require importing `path`, which then triggers a Vite browser-compat warning because this file is imported by the renderer (`convert-tree-to-sync-options.ts`, `<SyncConnectedSites>`).

We could make that change, but then it adds quite a lot of additional code that otherwise wouldn't be necessary.

What do you think?

tools/common/lib/fs-utils.ts
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2955/commits/c67ce3333f5f9979fd89c3619628f317b945c600.

apps/cli/lib/import-export/export/types.ts
ivan-ottinger: Addressed in https://github.com/Automattic/studio/pull/2955/commits/0a16cd81cb2a4ce352dd1c672bd3fc352214af73.

apps/cli/lib/import-export/export/exporters/default-exporter.ts
ivan-ottinger: I first tried to put it directly into this PR, but then decided to leave it for a new follow-up task: STU-1630 task - so we can unblock this PR.

tools/common/lib/deploy-ignore.ts
ivan-ottinger: No worried. Do you mean to merge records from both files and use that to decide what should get synced to WordPress.com / Pressable? 🤔 Wouldn't it be mixing two different concepts?

tools/common/lib/sync/tree-utils.ts
fredrikekelund: Ah, okay. Then maybe move `shouldExcludeFromSync` to another file..?

ivan-ottinger: Yeah, we can do that. I just wonder if it worth that one line change.

fredrikekelund: You make the call. All current callers replace Windows-style path separators with POSIX-style path separators, so this logic will work as expected. This is really only about resilience for future change.

ivan-ottinger: Alright, I made the change in https://github.com/Automattic/studio/pull/2955/commits/e01525a1e9e79862c0465b72b7efbaeddb4dc4d0.

ivan-ottinger:
Hey @fredrikekelund 👋🏼 I have applied the latest changes that also required adjustment related to recent https://github.com/Automattic/studio/pull/3129 PR. Now we have a new `--apply-deploy-ignore` flag for the `studio export` to ensure we don't  filter out files when the user is doing just a regular site export.

I have tested everything on macOS and Windows and haven't observed any regressions.

tools/common/lib/deploy-ignore.ts
fredrikekelund: > Do you mean to merge records from both files and use that to decide what should get synced to WordPress.com / Pressable?

Yep, exactly. npm is an interesting example to look at. It uses an `.npmignore` file by default to determine what to include when publishing a package, and falls back to `.gitignore` ([docs](https://docs.npmjs.com/cli/v11/using-npm/developers#keeping-files-out-of-your-package)).

> Wouldn't it be mixing two different concepts?

`.gitignore` might not be comprehensive in terms of what you want and don't want to sync to your remote site, but it's still a good indicator – especially if there's no `.deployignore` file.

`.gitignore` is specifically mentioned in https://github.com/Automattic/studio/issues/137, too.

ivan-ottinger: Sounds good. 👍🏼 I have created a follow-up task: STU-1633.

> .gitignore is specifically mentioned in https://github.com/Automattic/studio/issues/137, too.

Yes, I am aware of that. Just when we discussed the scope in STU-16, we shifted to `.deployignore` at the end.

PR #2956, opened 2026-04-01
Fix AI TUI crash when pressing ESC during agent turn

epeicher:
## Related issues

- Fixes STU-1498

## How AI was used in this PR

Claude assisted with debugging and identifying the root cause. But I needed to iterate multiple times finding different approaches until the current approach worked.

## Proposed Changes

- Add an `unhandledRejection` handler in `apps/cli/ai/agent.ts` to catch expected SDK rejections during ESC interruption

When the user presses ESC to interrupt an AI agent turn, the Claude Agent SDK rejects internal pending promises (e.g. control responses) during its cleanup path. These rejections are unhandled because they originate inside the SDK rather than propagating through the async iterator. Without a handler, Node.js terminates the process on unhandled rejections, crashing the TUI.

The handler only suppresses `"Query closed"` errors — all other unhandled rejections are re-thrown.

## Testing Instructions

- Run `npm run cli:build && node apps/cli/dist/cli/main.mjs ai`
- Send a prompt to the AI agent
- Press ESC multiple times rapidly while the agent is responding
- Verify the turn is interrupted cleanly and the TUI returns to the prompt input
- Verify you can send another prompt after the interruption

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/agent.ts
+++ b/apps/cli/ai/agent.ts
@@ -31,6 +31,18 @@ export type AiModelId = keyof typeof AI_MODELS;
 export const DEFAULT_MODEL: AiModelId = 'claude-sonnet-4-6';
 const pathApprovalSession = createPathApprovalSession();
 
+// The Claude Agent SDK rejects internal pending promises (e.g. control
+// responses) when an agent turn is interrupted via ESC. These rejections
+// are unhandled because they originate inside the SDK cleanup path rather
+// than propagating through the async iterator. Without this handler,
+// Node.js terminates the process on unhandled rejections.
+process.on( 'unhandledRejection', ( reason ) => {
+	if ( reason instanceof Error && reason.message.includes( 'Query closed' ) ) {
+		return;
+	}
+	throw reason;
+} );
+
 /**
  * Start the AI agent and return the Query object.
  * Caller can iterate messages with `for await` and call `interrupt()` to stop.


apps/cli/ai/agent.ts
```diff
@@ -31,6 +31,18 @@ export type AiModelId = keyof typeof AI_MODELS;
...
+// responses) when an agent turn is interrupted via ESC. These rejections
+// are unhandled because they originate inside the SDK cleanup path rather
+// than propagating through the async iterator. Without this handler,
+// Node.js terminates the process on unhandled rejections.
+process.on( 'unhandledRejection', ( reason ) => {
+	if ( reason instanceof Error && reason.message.includes( 'Query closed' ) ) {
+		return;
+	}
+	throw reason;
+} );
```
youknowriad, line 44: This is a weird fix, it might be correct but for me we probably have an explicit event handler for the "ESC" that triggers the interruptions. So it would make more sense for me for the interruption and failures handling to be a one off as part of that handler instead of a global listener like this (where we don't really know where the issue comes from). Now it might not be possible I don't know.

youknowriad:
The other thing I noticed is that the "esc" handler doesn't interrupt things right away for some reason. I think we might want to fix that as well.

apps/cli/ai/agent.ts
epeicher: I absolutely agree this feels like a weird fix — a global listener for something that should be handled locally. Unfortunately, it was the only approach that actually worked after testing several alternatives:

1. **Nullifying the interrupt callback after the first ESC press** — The [ESC handler in `ui.ts`](https://github.com/Automattic/studio/blob/81d79fc2/apps/cli/ai/ui.ts#L768) calls `this.interruptCallback()`, which triggers [`agentQuery.interrupt()`](https://github.com/Automattic/studio/blob/81d79fc2/apps/cli/commands/ai/index.ts#L294-L295). We tried nullifying the callback after the first call to prevent multiple `interrupt()` calls. However, a single `interrupt()` call does not reliably stop the agent — we confirmed via debug logs that the callback fires but the agent continues running indefinitely.

2. **Using the `wasInterrupted` flag as a guard** — Same idea using the [existing `wasInterrupted` flag](https://github.com/Automattic/studio/blob/81d79fc2/apps/cli/ai/ui.ts#L769), different mechanism. Same result: the agent wouldn't actually stop.

3. **Catching the error in the `for-await` catch block** — We tried catching `AbortError` / `"Query closed"` errors in the [`catch` block at `index.ts:320`](https://github.com/Automattic/studio/blob/81d79fc2/apps/cli/commands/ai/index.ts#L320) where the [async iterator](https://github.com/Automattic/studio/blob/81d79fc2/apps/cli/commands/ai/index.ts#L302) is consumed. But the errors don't propagate through the iterator — they come from internal SDK promises (pending `canUseTool` control responses) that get rejected during the SDK's internal `cleanup()`. These rejections are never awaited by our code (as we don't have references to the code that rejects them), so they surface as unhandled promise rejections at the process level, which is why Node.js terminates.

**Why multiple `interrupt()` calls are needed:** The SDK's `interrupt()` likely sends a control message to the underlying Claude subprocess. If the subprocess is in the middle of processing (e.g., streaming a response, executing a tool call, or waiting on an API response), a single interrupt signal can be ignored or queued. We confirmed this empirically: with a single `interrupt()` call, the debug logs showed the interrupt fired but the agent turn never ended (`inAgentTurn` stayed `true`). Only when ESC is pressed repeatedly — sending multiple `interrupt()` calls — does the agent reliably stop. This is consistent with how process signals work generally (e.g., `SIGINT` sometimes needs to be sent multiple times to a busy process).

This is also related to the other comment about ESC not interrupting things right away — the delay is because the first few `interrupt()` calls may not get through while the SDK is busy, and it takes several attempts before one lands.

katinthehatsite:
In terms of functionality, it worked well for me - I either pressed `Esc` once and it exited correctly or I tried pressing it multiple times and it also worked as expected and I returned to the input field.

katinthehatsite (approved):
I looked for different options, and I did not see something more straightforward. I think it is a reasonable solution 👍

apps/cli/ai/agent.ts
youknowriad: How possible would it be to only register this event handler when startAgent is called and potentially clean it at the end?

youknowriad:
Let's not keep this blocked. Not the best solution but maybe the only one for now.

PR #2958, opened 2026-04-01
Fix start:test isolated config after split-config refactor

wojtekn:
## Related issues

- Fixes STU-1462

## How AI was used in this PR

Claude Code was used to investigate the root cause and implement the fix.

## Proposed Changes

This PR adds `DEV_CONFIG_DIR` env var support to `getConfigDirectory()` in `tools/common/lib/well-known-paths.ts`, redirecting the entire `~/.studio/` config directory when set.

The config refactor split the single appdata file into three files (`app.json`, `cli.json`, `shared.json`). The old `DEV_APP_DATA_PATH` only redirected `app.json` — `cli.json` (sites, snapshots) and `shared.json` (auth) still pointed at `~/.studio/`, so real developer sites could appear in the test instance.

## Testing Instructions

1. Run `npm run start:test` — app should open with no sites and no auth
2. Confirm `/tmp/studio-test/` is created with the isolated config files
3. Confirm `~/.studio/cli.json` and `~/.studio/shared.json` are unchanged after the session
4. Run `npm start` and confirm your real sites still appear normally

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/storage/paths.ts
+++ b/apps/studio/src/storage/paths.ts
@@ -16,9 +16,6 @@ try {
 }
 
 export function getUserDataFilePath(): string {
-	if ( process.env.DEV_APP_DATA_PATH ) {
-		return process.env.DEV_APP_DATA_PATH;
-	}
 	return getAppConfigPath();
 }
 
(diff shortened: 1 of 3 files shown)


katinthehatsite (approved):
These changes look good to me 👍 I was able to see the correct files in the target directory:

Image

katinthehatsite:
I will retrigger the failed tests now 👀

PR #2962, opened 2026-04-02
Windows: Auto-install CLI on first launch

ivan-ottinger:
> [!IMPORTANT]  
> Let's merge this PR once the related macOS PR (https://github.com/Automattic/studio/pull/2937) has been merged.

## Related issues

- Related to STU-1363
- Fixes STU-1505

## How AI was used in this PR

Claude Code was used to implement the changes based on the patterns established in the macOS auto-install PR (#2937).

## Proposed Changes

- Add `autoInstallIfNeeded()` to `WindowsCliInstallationManager` using the same `cliAutoInstalled` flag from the macOS PR
- Replace `updateWindowsCliVersionedPathIfNeeded()` export with `autoInstallWindowsCliIfNeeded()` which handles both auto-install on first launch and proxy bat file updates on subsequent launches
- Skip auto-install in development mode to match macOS behavior
- Add unit tests for the new auto-install logic
- Fix `npm run make` on Windows failing due to Windows 8.3 short path mismatch in the staging directory (STU-1505 )- I have added it to this PR as it is a small change, but we could also extract it out to a separate PR

**Note:** This PR is based on the macOS auto-install branch (`stu-1363-auto-install-cli-macos`, PR #2937). It should be merged after that PR lands.

## Testing Instructions

> **Note:** Auto-install only runs in production mode. You must test with a built app, not `npm start`.

### Setup (on Windows)
1. Open the current production Studio app and disable Studio CLI from the app setting. Close the app.
2. Ensure the CLI is not already installed:
   - Check if the bin directory exists: `Test-Path "$env:LOCALAPPDATA\studio_app\bin"`
   - If it exists, delete it: `Remove-Item -Recurse -Force "$env:LOCALAPPDATA\studio_app\bin"`
   - Check if the studio bin directory is in the registry PATH: `reg query HKCU\Environment /v Path`
     - If it contains a `studio_app\bin` entry, remove it from the PATH manually
   - Reset the auto-install flag: open the file with `notepad "$env:USERPROFILE\.studio\app.json"` and delete the `"cliAutoInstalled": true` entry (if you are testing this PR for the first time, it should not be present - since we are introducing it in this PR)
3. Build the app: `npm run make`
4. Install from `apps\studio\out\make\squirrel.windows\arm64\studio-setup.exe`
5. Open a **new** terminal and verify `studio --version` returns "not recognized"

### Auto-install on first launch
6. Launch newly-built Studio
7. Open a **new** terminal and verify the CLI was installed: `studio --version`
8. Verify the flag was set: open the file with `notepad "$env:USERPROFILE\.studio\app.json"` and check for `"cliAutoInstalled": true`

### User preference is respected
9. In Studio, go to Settings > Preferences, toggle the CLI **off**, and save
10. Quit and relaunch Studio
11. Verify the CLI was **not** reinstalled: `studio --version` should not work in a new terminal

### Re-enabling
12. Toggle the CLI back **on** in Settings > Preferences, and save
13. Verify `studio --version` works again in a new terminal

## Pre-merge Checklist


scripts/package-in-isolation.ts
```diff
@@ -105,7 +105,9 @@ function main() {
 		return;
 	}
 
-	const stagingParent = fs.mkdtempSync( path.join( os.tmpdir(), 'studio-package-' ) );
+	const stagingParent = fs.realpathSync.native(
+		fs.mkdtempSync( path.join( os.tmpdir(), 'studio-package-' ) )
+	);
```
ivan-ottinger, line 110: This change is fixing STU-1505. I have added it to this PR since it is small and I couldn't built the Windows app without it. If desired, we could move it to a separate PR.

apps/studio/src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -36,10 +38,17 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
...
+		const userData = await loadUserData();
+		if ( userData.cliAutoInstalled ) {
+			// Already ran auto-install before. If CLI is still installed,
+			// update the proxy bat file for the current app version.
+			await this.updateWindowsCliVersionedPathIfNeeded();
+			return;
 		}
+
+		await this.installCli();
+		await updateAppdata( { cliAutoInstalled: true } );
```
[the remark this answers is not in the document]
ivan-ottinger: I considered this but decided to keep it simple. On macOS, `installCli()` throws on errors so the flag is never reached on failure. On Windows, `installPath()` and `installProxyBatFile()` swallow errors, but the failure scenario is very unlikely (user-writable registry and directory). If it does fail, the user can still try manually installing the CLI via the preferences toggle.

apps/studio/src/modules/cli/lib/tests/windows-installation-manager.test.ts
```diff
@@ -0,0 +1,219 @@
...
+				( _key: string, _type: string, _value: string, cb: ( err: Error | null ) => void ) =>
+					cb( null )
+			);
+
+			await manager.autoInstallIfNeeded();
+
+			expect( mockWriteFile ).toHaveBeenCalled();
+			expect( mockUpdateAppdata ).toHaveBeenCalledWith( { cliAutoInstalled: true } );
+		} );
+
```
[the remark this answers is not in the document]
ivan-ottinger: Not needed given the above - we're setting the flag unconditionally after install on both platforms.

scripts/package-in-isolation.ts
fredrikekelund: I don't fully comprehend this change, but it seems reasonable 👍 I'll take another look at this tomorrow after we've landed this PR

fredrikekelund (approved):
LGTM 👍

scripts/package-in-isolation.ts
ivan-ottinger: Thank you for the merge, Fredrik!

> I don't fully comprehend this change, but it seems reasonable 👍 I'll take another look at this tomorrow after we've landed this PR

On my VM the username is too long so it comes up shortened in the the temp path like `IVANOT~1` instead of `ivanottinger` - which then causes that odd error:

```
[vite:build-html] The "fileName" or "name" properties of emitted chunks and assets must be strings
that are neither absolute nor relative paths, received
"../../../../../../../../ivanottinger/AppData/Local/Temp/studio-package-OrYMiE/repo/apps/studio/index.html".
```

When we use `realpathSync.native`, we get the full path with the proper username (e.g. `ivanottinger` in this case).

PR #2963, opened 2026-04-02
Implement import/export in CLI

fredrikekelund:
## Related issues

- Fixes STU-1351

## How AI was used in this PR

Primarily to review my implementation and help me preemptively fix several issues. 

## Proposed Changes

> [!TIP]
> I'm happy to pair review this PR. 

This one's a biggie. It adds `site import` and `site export` CLI commands. I've copied the core import/export logic from `apps/studio` and adapted it to CLI conditions (the types are slightly different, we call WP-CLI in different ways, etc).

Moreover, this PR finalizes the `sync pull` and `sync push` commands by wiring up the import/export logic there, too. 

## Testing Instructions

1. `npm run cli:build`
2. Try the `node apps/cli/dist/cli/main.mjs site export` command. Explore the flags and try different combinations. 
3. Try the `node apps/cli/dist/cli/main.mjs site import` command. Explore the flags and try different combinations. 
4. Try the `node apps/cli/dist/cli/main.mjs sync pull` command.
5. Try the `node apps/cli/dist/cli/main.mjs sync push` command.

## Pre-merge Checklist


apps/cli/commands/site/delete.ts
```diff
@@ -118,10 +119,6 @@ export async function runCommand(
 		if ( deleteFiles ) {
 			if ( fs.existsSync( siteFolder ) ) {
 				logger.reportStart( LoggerAction.DELETE_FILES, __( 'Moving site files to trash…' ) );
-				// We configure `trash` as an external module, since it includes a native macOS binary that Vite
-				// inlines as a base64 string, which produces a runtime error. Since `trash` is also an ESM-only
-				// module, we need to import it dynamically (since Rollup doesn't get a chance to process it)
-				const trash = ( await import( 'trash' ) ).default;
```
fredrikekelund, line 124: Not related to the core changes in this PR, but we output ESM now, and `trash` is always configured as an external module when Vite bundles, so this is no longer an issue.

apps/cli/commands/site/export.ts
```diff
@@ -0,0 +1,175 @@
...
+
+		case ExportEvents.DATABASE_EXPORT_START:
+			logger.reportStart( LoggerAction.EXPORT_DATABASE, __( 'Exporting database…' ) );
+			break;
+		case ExportEvents.DATABASE_EXPORT_COMPLETE:
+			logger.reportSuccess( __( 'Database exported' ) );
+			break;
+
+		case ExportEvents.BACKUP_CREATE_PROGRESS: {
+			const progressData = data as BackupCreateProgressEventData;
```
fredrikekelund, line 43: I'd love to avoid type assertions for the event payloads, but since this copies the logic we already have in the app, I decided to stick with that for now.

apps/cli/commands/site/export.ts
```diff
@@ -0,0 +1,175 @@
...
+		logger.reportStart( LoggerAction.LOAD_SITES, __( 'Loading site…' ) );
+		const site = await getSiteByFolder( siteFolder );
+		logger.reportSuccess( __( 'Site loaded' ) );
+
+		logger.reportStart(
+			LoggerAction.INSTALL_SQLITE,
+			__( 'Setting up SQLite integration, if needed…' )
+		);
+		await keepSqliteIntegrationUpdated( siteFolder );
+		logger.reportSuccess( __( 'SQLite integration configured as needed' ) );
```
fredrikekelund, line 95: It's important to keep the SQLite integration up to date whenever we work with the site database. Exporting the site is one such example.

apps/cli/commands/site/set.ts
```diff
@@ -267,7 +267,7 @@ export async function runCommand( sitePath: string, options: SetCommandOptions )
 			const phpVersion = validatePhpVersion( site.phpVersion );
 			const zipUrl = getWordPressVersionUrl( wp );
 
-			const [ response, exitPhp ] = await runWpCliCommand( sitePath, phpVersion, [
+			await using command = await runWpCliCommand( sitePath, phpVersion, [
```
fredrikekelund, line 270: If you're not familiar with the `using` keyword, I recommend a quick read-up on [JavaScript resource management](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Resource_management). Playground already uses this internally.

In short, it saves us from having to call a cleanup function (`exitPhp`) because the JS runtime automatically calls the method identified by `Symbol.dispose` when exiting the scope. Very nifty.

apps/cli/lib/run-wp-cli-command.ts
fredrikekelund, line 1: See my comment in `apps/cli/commands/site/set.ts` about JavaScript resource management to understand what's going on here.

apps/cli/package.json
```diff
@@ -27,28 +27,32 @@
...
-		"@wp-playground/blueprints": "3.1.15",
-		"@wp-playground/cli": "3.1.15",
-		"@wp-playground/common": "3.1.15",
-		"@wp-playground/storage": "3.1.15",
-		"@wp-playground/wordpress": "3.1.15",
+		"@wp-playground/blueprints": "3.1.13",
+		"@wp-playground/cli": "3.1.13",
+		"@wp-playground/common": "3.1.13",
+		"@wp-playground/storage": "3.1.13",
+		"@wp-playground/wordpress": "3.1.13",
```
fredrikekelund: I had to downgrade these packages because I couldn't get WP-CLI to play ball. See p1775051878031829-slack-C04GESRBWKW

fredrikekelund: I was able to revert this change after copying the fix from https://github.com/Automattic/studio/pull/2980

apps/cli/commands/site/export.ts
```diff
@@ -0,0 +1,175 @@
...
+	}
+}
+
+export const registerCommand = ( yargs: StudioArgv ) => {
+	return yargs.command( {
+		command: 'export <export-file>',
+		describe: __( 'Export site to a backup file' ),
+		builder: ( yargs ) => {
+			return yargs
+				.positional( 'export-file', {
```
bcotrim, line 142: Should we allow this to be empty and use a generated file name in the current folder?
I think it would be a good UX improvement

apps/cli/commands/site/import.ts
```diff
@@ -0,0 +1,298 @@
...
+		await connectToDaemon();
+		logger.reportSuccess( __( 'Process daemon started' ) );
+
+		logger.reportStart( LoggerAction.LOAD_SITES, __( 'Loading site…' ) );
+		site = await getSiteByFolder( siteFolder );
+		logger.reportSuccess( __( 'Site loaded' ) );
+
+		wasServerRunning = !! ( await isServerRunning( site.id ) );
+
+		if ( wasServerRunning ) {
```
bcotrim, line 208: Nit: Could we move this after the import file and site path directory validations? It could potentially prevent an unnecessary site restart.

apps/cli/commands/site/import.ts
```diff
@@ -0,0 +1,298 @@
...
+		await importBackup(
+			{ path: importFile, type: getBackupFileType( importFile ) },
+			site,
+			importEventHandler,
+			DEFAULT_IMPORTER_OPTIONS
+		);
+
+		// Something in Playground makes it so the front-end of the site sometimes returns an error page
+		// on the first request. Send that first request from here to hide the error from the user.
+		const siteUrl = getSiteUrl( site );
```
bcotrim, line 232: Was this reported to the Playground team?
The workaround isn't particularly problematic but I wonder what is the root cause for it, and if it has other side effects.

apps/cli/commands/site/set.ts
bcotrim: TIL

I wonder if we could refactor our config files lock/unlock pattern with this approach.
Obviously not a blocker for this PR, and it could be a follow-up

apps/cli/commands/sync/pull.ts
```diff
@@ -32,133 +46,193 @@ export async function runCommand(
...
+			// Something in Playground makes it so the front-end of the site sometimes returns an error page
+			// on the first request. Send that first request from here to hide the error from the user.
+			const siteUrl = getSiteUrl( site );
+			await fetch( siteUrl ).catch( () => {} );
+
+			logger.reportSuccess(
+				sprintf( __( 'Pulled from %s (%s)' ), remoteSite.name, remoteSite.url )
+			);
+		} finally {
+			fs.rmSync( tempDir, { recursive: true, force: true } );
```
bcotrim, line 200: This was a pre-existing issue from my initial PR.
Concurrent sync operations, even for different sites could break since the tempDir is always the same.
Studio app uses `mkdtemp`

bcotrim (commented):
LGTM 👍 
Added a few comments but no blockers.
I was considering if we should share the importer and exporters instead of copying, but since we are eventually removing the operations from Studio app and using the CLI, I think it makes sense to avoid any side effects now and just live with copied files.

apps/cli/commands/site/set.ts
fredrikekelund: That's an interesting candidate 👍 There's some nuance to how we handle that, though – it sometimes takes a while for the code to exit the current scope after writing the config file. Worth exploring, though.

`connectToDaemon` and `disconnectFromDaemon` also seem like good candidates to me.

apps/cli/commands/site/import.ts
fredrikekelund: Not yet, no. I'll put it on my to-do list. I encountered it several times, but I haven't isolated the error yet, so I'd like to do that before reporting.

apps/cli/commands/sync/pull.ts
fredrikekelund: Great catch 👍

fredrikekelund: I made the corresponding update in the push command, too.

apps/cli/commands/site/export.ts
fredrikekelund: That's fair. I've made this change 👍

fredrikekelund:
> I was considering if we should share the importer and exporters instead of copying, but since we are eventually removing the operations from Studio app and using the CLI, I think it makes sense to avoid any side effects now and just live with copied files.

Yeah, I think so, too. We call WP-CLI in several places in the importer and exporter, and it would be an unnecessary headache to rewrite a shared version of the code to inject a WP-CLI caller that wraps the CLI/app logic. However, this assumes we proceed soon to rewrite the app's import/export logic to depend on the CLI.

I've addressed your other review comments, @bcotrim. I also changed the export command to use a `--mode=db|full` option instead of `--only`. `--mode` is required, but it defaults to `full`. When combined with the fact that export filenames are now auto-generated, I found this more convenient and logical compared to `--only`.

bcotrim (approved):
Thanks for making the changes
LGTM 👍

PR #2965, opened 2026-04-02
Refetch Sync dialog on focus and add spinner

nightnei:
## Related issues

- Fixes STU-1507

## How AI was used in this PR
AI assisted me, but the implementation is mine. 

## Proposed Changes

1. The use case is - user opens Sync dialog, realizes that they need to connect Jetpack in Pressable, so they keep the dialog open and move to the browser to connect Jetpack to their site on Pressable. Then they return to Studio, don't see the just-connected site, and they restart Studio.

The solution - we need to refetch sites on the focus event, if the Sync dialog is opened in Studio.

2. Right now, when we reopen the Sync dialog - we take data from cache and start refetching, but we don't communicate that we are refetching them. That's why when a user reopens the Sync dialog, we have a flash of content after a few seconds. I had a case where I was selecting a site and ended up choosing the wrong one because of a flash of content.

The solution - we should communicate in UI that we are refetching sites.

## Testing Instructions
1. Opean Sync dialog and assert that you see `Loading sites`<br />Screenshot 2026-04-02 at 17 36 51
2. Close the dialog and open again
3. Assert that you see sites rendered from cache, but covered with blurred "Refreshing sites" spinner<br />Screenshot 2026-04-02 at 17 38 32
4. Leave Studio and focus on another app on your laptop
5. Comeback and focus on Studio again
6. Assert that you see "Refreshing sites" spinner<br />Screenshot 2026-04-02 at 17 38 32

### DIFF
--- a/apps/studio/src/modules/add-site/components/pull-remote-site.tsx
+++ b/apps/studio/src/modules/add-site/components/pull-remote-site.tsx
@@ -139,7 +139,9 @@ export function PullRemoteSite( {
 	const {
 		data: syncSites = [],
 		isLoading,
+		isFetching,
 		isSuccess,
+		refetch: refetchWpComSites,
 	} = useGetWpComSitesQuery(
 		{
 			userId: user?.id,
@@ -167,9 +169,11 @@ export function PullRemoteSite( {
 					) : (
 						<SitesListContent
 							isLoading={ isLoading }
+							isFetching={ isFetching }
 							syncSites={ syncSites }
 							selectedSiteId={ selectedRemoteSite?.id || null }
 							onSelectSite={ handleSiteSelect }
+							refetchSites={ refetchWpComSites }
 						/>
 					) }
 				</VStack>
--- a/apps/studio/src/modules/sync/components/sync-sites-modal-selector.tsx
+++ b/apps/studio/src/modules/sync/components/sync-sites-modal-selector.tsx
@@ -58,7 +58,9 @@ export function SyncSitesModalSelector( {
 	const {
 		data: syncSites = [],
 		isLoading,
+		isFetching,
 		isSuccess,
+		refetch: refetchWpComSites,
 	} = useGetWpComSitesQuery(
 		{ connectedSiteIds, userId: user?.id },
 		{ refetchOnMountOrArgChange: true }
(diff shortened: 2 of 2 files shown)


shaunandrews:
I made a few small tweaks to the blur and opacity to help with contrast a bit. I added a quick fade in/out transition. I added the loading spinner and simplified the message to "Refreshing..." — and also updated the loading message to use the spinner and just "Loading..."

image
image
image

katinthehatsite (commented):
I like the updated design and I think it looks quite nice 😍 

One thing that I noticed is that sometimes when I open for the first time the sync dialog for a specific site, it shows `Refreshing` instead of `Loading` which is slightly confusing since this is what the user is seeing for the first time for that site: 

https://github.com/user-attachments/assets/73022d66-67bb-48c5-9c96-870ca44eff2a

Do you think it is something that we could adjust? There are also a couple of minor linter issues that need to be resolved.

nightnei:
@katinthehatsite 

> There are also a couple of minor linter issues that need to be resolved.

Fixed

> Do you think it is something that we could adjust?

Hm, IDK what is the best approach here 🤔 The problem is - we render "Refreshing" in case when we have data already in cache, but we use `useGetWpComSitesQuery` on many pages/places. Invalidation of such data can slow down some parts of Studio so I don't think that it's a good way. On the other hand we could render always only "Loading..." copy inside Sync dialog, but it can also feel confusing when Studio loads data which is already in place, that's why I intentionally went with the approach of printing "Refreshing" when we indeed refresh data.
TBH, IDK what is the best approach, so I would leave it as is and later we can improve if we find it critical.

katinthehatsite:
> Hm, IDK what is the best approach here 🤔 The problem is - we render "Refreshing" in case when we have data already in cache, but we use useGetWpComSitesQuery on many pages/places. Invalidation of such data can slow down some parts of Studio so I don't think that it's a good way. On the other hand we could render always only "Loading..." copy inside Sync dialog, but it can also feel confusing when Studio loads data which is already in place, that's why I intentionally went with the approach of printing "Refreshing" when we indeed refresh data.
TBH, IDK what is the best approach, so I would leave it as is and later we can improve if we find it critical.

Makes sense, let's leave as is and we can iterate if this creates confusion.

katinthehatsite (approved):

PR #2972, opened 2026-04-03
Improve onboarding flow for studio code cli

mtias:
It should ask for auth or key and then show what the tool can do directly so the user doesn't need to ask. This is a first crude version.

### DIFF
--- a/apps/cli/ai/ui.ts
+++ b/apps/cli/ai/ui.ts
@@ -1432,18 +1432,25 @@ export class AiChatUI {
 		const lines = [
 			' ' +
 				chalk.blue( '⏺' ) +
-				__( " Hello, I'm " ) +
-				b( __( 'WordPress Studio' ) ) +
-				__( ', your local WordPress agent and builder.' ),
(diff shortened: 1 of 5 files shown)


apps/cli/ai/ui.ts
```diff
@@ -1425,6 +1426,109 @@ export class AiChatUI {
...
+			'  - ' +
+				__( 'Set up ' ) +
+				b( __( 'navigation menus, site options, post types, taxonomies, and settings' ) ),
+			'  - ' + __( "Create realistic placeholder content tailored to your site's purpose" ),
+			'  - ' +
+				__( 'Upload ' ) +
+				b( __( 'images and videos' ) ) +
+				__( ' to your site, using local media files or remote URLs' ),
+			'',
+			'  ' + b( __( 'Preview & Publishing' ) ),
```
youknowriad, line 1541: In this area should we also mention that it can generate preview sites

youknowriad (approved):

PR #2978, opened 2026-04-05
Add /preview slash command to push active site to WordPress.com

mtias:
Adds a /preview command that creates or updates a WordPress.com preview site for the currently selected local site. Automatically detects existing non-expired previews and updates them instead of creating duplicates.

Also adds a busy spinner indicator in the prompt border next to the site name, a yellow-dot progress message style, and a green-dot success message style to the UI.

image


youknowriad:
If I run the preview manually (just ask the agent to preview) without using the command, the output looks like

Screenshot 2026-04-06 at 8 38 59 AM

With the slash command, the output is

Screenshot 2026-04-06 at 8 39 11 AM

I like how the former highlights the url better, can we do that

youknowriad:
The other minor thing I noticed is the "loader":

While I do like the "loading indicator" next to the active site, I feel like there's inconsistency with the regular loader when you ask the agent something. So sometimes the loader is next to the site, sometimes it's a thinking message. I wonder if they should be the same.

youknowriad (approved):

PR #2981, opened 2026-04-06
Batch site running status checks in site picker

mtias:
The site picker was making a separate daemon IPC round-trip for each site to check if it was running. With multiple sites this caused a noticeable delay when opening the picker.

Fetch the process list once with a single daemon call and resolve running status for all sites against it.


youknowriad (approved):
I pushed a small logic consolidation to avoid duplication, but LGTM.

The other thing I'm wondering if perf is still a problem is that this information can be computed and rendered async, doesn't need to be part of the initial rendering of the site list.

katinthehatsite (approved):
The changes look good and work as expected 👍

PR #2982, opened 2026-04-06
Fix pulling if there is no rewindId

nightnei:
Fixes STU-1517

## Proposed Changes
Follow up to https://github.com/Automattic/studio/pull/1672

The case from user - pfHvTO-19v-p2
The user had a site when they didn't have rewindId and they constantly clicked Pull button w/o any response from it.
With this PR we are fixing pulling if no rewindId.

## Testing Instructions
1. Sandbox your API
2. Add custom code to return an error in the /studio-app/sync/get-latest-rewind-id endpoint.
3. Run `npm start`
4. Go to the Sync tab
5. Connect a site
6. Click pull
7. Assert that it started

### DIFF
--- a/apps/studio/src/modules/sync/components/sync-dialog.tsx
+++ b/apps/studio/src/modules/sync/components/sync-dialog.tsx
@@ -200,7 +200,7 @@ export function SyncDialog( {
 		syncToText = localSite.name;
 		tooltipNoRewindId = createInterpolateElement(
 			__(
-				'Selecting individual items to pull will be enabled automatically once your first backup is complete.<br/>Wait a few minutes or run a full sync in the meantime.'
+				'Your site is being prepared for syncing. This usually takes a few minutes — please try again shortly.'
 			),
 			{ br: <br /> }
 		);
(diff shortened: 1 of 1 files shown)


nightnei:
@sejas menthioning you as you worked around it and have context. It seems we shoudl forbid pulling completelly, till rewindId is created. However, maybe I am missing something?

sejas (commented):
@nightnei, As an alternative to this PR, I think we could just remove this rewindId check. The backend doesn't need a `rewindId`, which is why the checkboxes were disabled (we can retrieve information about the latest backup), but a new backup can still be triggered.

So I suggest just removing this guard:

https://github.com/Automattic/studio/blob/18f45f482499c4e1e92d081d04d48fee00bdfc95/apps/studio/src/modules/sync/components/sync-dialog.tsx#L260-L262

I tested the flow were the rewindId is null and isError is true and I was able to pull a full backup, after I removed the mentioned guard.

What do you think?

apps/studio/src/modules/sync/components/sync-dialog.tsx
```diff
@@ -481,6 +478,7 @@ export function SyncDialog( {
 									disabled={
 										isSubmitDisabled ||
 										isLoadingRewindId ||
+										isErrorRewindId ||
```
sejas: @nightnei , I think we need to remove this `isErrorRewindId` too.

nightnei: Oh, yeah, accidentally didn't add it to the commit, thanks

sejas (approved):
Looks good. Now if there is no rewindId, users can still triggering a full Pull which will generate a new backup.

PR #2988, opened 2026-04-06
Update sync tooltip to use plan-agnostic language

wojtekn:
## Related issues

- STU-1516

## How AI was used in this PR

Claude Code was used to locate and update the tooltip string.

## Proposed Changes

I proposed updating the sync tooltip text from "Sync support is available only with Business plan and above" to "Sync support is available on selected plans only".

We are adding support for Sync to other plans, which is handled on the API side, so it would be great to avoid plan-related logic on the Studio side.

## Testing Instructions

- Open the sync sites modal selector
- Hover over the upgrade link for a site that needs a plan upgrade
- Verify the tooltip reads "Sync support is available on selected plans only"

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/modules/sync/components/sync-sites-modal-selector.tsx
+++ b/apps/studio/src/modules/sync/components/sync-sites-modal-selector.tsx
@@ -361,7 +361,7 @@ function SiteItem( {
 			{ needsUpgrade && (
 				<div className="a8c-body-small text-frame-text-secondary shrink-0 text-right">
 					<Tooltip
-						text={ __( 'Sync support is available only with Business plan and above' ) }
+						text={ __( 'Sync support is available on selected plans only' ) }
 						placement="bottom"
 					>
 						<Button


epeicher (approved):
Thanks @wojtekn, I can see the new tooltip when hovering over the `Upgrade plan` link. LGTM!

Image

apps/studio/src/modules/sync/components/sync-sites-modal-selector.tsx
```diff
@@ -361,7 +361,7 @@ function SiteItem( {
 			{ needsUpgrade && (
 				<div className="a8c-body-small text-frame-text-secondary shrink-0 text-right">
 					<Tooltip
-						text={ __( 'Sync support is available only with Business plan and above' ) }
+						text={ __( 'Sync support is available on selected plans only' ) }
```
katinthehatsite, line 364: I like the idea of avoiding the plan specific language but it would be nice to point the users to what plans are actually available for sync. I am wondering if we could add something like `Learn more` link to the tooltip where we have this mentioned in the documentation. Although perhaps, it is not the best approach to add a link to the tooltip. 

It is not a blocker, I am mostly adding this as an observation

katinthehatsite (approved):
The changes look good, I left one minor comment but it is not a blocker

apps/studio/src/modules/sync/components/sync-sites-modal-selector.tsx
wojtekn: This is a tooltip appearing on a link - the user can click the link and see plans that support sync.

Separately from this issue, we are adding Studio Sync to the pricing grids to make it clearer. Also, with the server-side change happening in parallel, we will enable Studio sync for all current paid plans.

PR #2989, opened 2026-04-06
Generate SSL certificate eagerly when HTTPS is enabled

bcotrim:
## Related issues

- Fixes STU-1476

## How AI was used in this PR

Claude Code was used to explore the codebase, identify the root cause, and implement the fix. The change was reviewed and validated by the author.

## Proposed Changes

- Generate the SSL certificate immediately when HTTPS is toggled on via `site set`, rather than deferring it to site start. Previously, `generateSiteCertificate` was only called inside `setupCustomDomain` during site restart, so enabling HTTPS on a stopped site would show the "Trust Certificate" nudge with no certificate file to trust.

## Testing Instructions

- Have a clean Studio install (or clear `~/.studio/certificates/`)
- Create a site with a custom domain but without HTTPS
- Stop the site
- Edit the site and enable HTTPS
- Verify the certificate is generated immediately (check `~/.studio/certificates/domains/`)
- Go to site Settings tab — the "Trust Certificate" nudge should now work correctly

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/site/set.ts
+++ b/apps/cli/commands/site/set.ts
@@ -16,6 +16,7 @@ import {
 } from '@studio/common/lib/wordpress-version-utils';
 import { SiteCommandLoggerAction as LoggerAction } from '@studio/common/logger-actions';
 import { __, sprintf } from '@wordpress/i18n';
+import { generateSiteCertificate } from 'cli/lib/certificate-manager';
 import {
 	lockCliConfig,
 	readCliConfig,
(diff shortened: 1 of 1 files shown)


epeicher (approved):
Thanks @bcotrim! I have tested it on Windows, and I can confirm that it works fine. I can see that the certificates are generated when enabling the HTTPS setting, and that the nudge is displayed as expected. LGTM!

PR #2990, opened 2026-04-06
Add studio-ai binary for direct AI command access

sejas:
## Related issues

- Fixes STU-1502

## How AI was used in this PR

Claude Code was used to implement the feature based on the Linear issue spec. All changes were reviewed before committing.

## Proposed Changes

This approach installs both studio and studio-ai at the same time, to follow the example shared in the Linear issue: `npm install -g wp-studio && studio-ai`

- Add a new `studio-ai` binary entry point (`apps/cli/ai.ts`) that launches the AI assistant directly, as an alternative to `studio ai`
- Register the new entry point in the Vite build config so it gets compiled to `ai-main.mjs`
- Update the npm Vite config to add shebang, enable AI flag, and copy the AI plugin for npm distribution
- Add `studio-ai` to the `bin` field in `package.json`

After `npm install -g wp-studio`, users can run either `studio ai` or `studio-ai` — both work identically.

We can iterate in the final name later.

## Testing Instructions

2. Verify `node apps/cli/dist/cli/ai-main.mjs` works
   - Should show `studio-ai` as script name with AI command as default and `sessions` sub-command
3. Verify `studio ai` still works: ` node apps/cli/dist/cli/main.mjs ai`
4. Verify sessions sub-command: `node apps/cli/dist/cli/ai-main.mjs sessions list`
5. Run `npm run typecheck` — should pass

Screenshot 2026-04-06 at 19 16 55

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai.ts
+++ b/apps/cli/ai.ts
@@ -1,0 +1,108 @@
+import path from 'node:path';
+import { suppressPunycodeWarning } from '@studio/common/lib/suppress-punycode-warning';
+import { __ } from '@wordpress/i18n';
+import yargs from 'yargs';
+import { bumpAggregatedUniqueStat, getPlatformMetric } from 'cli/lib/bump-stat';
+import { setupServerFiles } from 'cli/lib/dependency-management/setup';
+import { loadTranslations } from 'cli/lib/i18n';
+import { StatsGroup, StatsMetric } from 'cli/lib/types/bump-stats';
+import { untildify } from 'cli/lib/utils';
+import { StudioArgv } from 'cli/types';
+
+const version = __STUDIO_CLI_VERSION__;
+
+suppressPunycodeWarning();
+
+async function main() {
+	const yargsLocale = await loadTranslations();
+
+	const studioAiArgv: StudioArgv = yargs( process.argv.slice( 2 ) )
+		.scriptName( 'studio-ai' )
+		.usage( __( 'AI-powered WordPress assistant' ) )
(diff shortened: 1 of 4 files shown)


youknowriad:
I'd love @matias's take on this, I think he had some strong opinions here.

sejas:
> I'd love @matias's take on this, I think he had some strong opinions here.

Makes sense. Let's wait Matias' opinion.
From my point of view, it makes more sense as an independent npm package so that a user can install either one or the other, or both, independently.

So instead of:
`npm install -g wp-studio && studio-ai` , it could be `npm install -g studio-ai && studio-ai` or just `npx studio-ai@latest`

youknowriad:
I'm not sure if it makes sense as an independent package, you still need the full Studio CLI (managing sites, starting sites, importing...) so it would be just the same package twice.

mtias:
Yes! This should be a single install for the CLI tool, what we offer from the landing page `npm install -g wp-studio`, and the command to invoke the agent should be `studio code`.

youknowriad:
Sounds good to me, it means no need for a secondary binary, we can just rename the "ai" command to "code" or maybe keep "ai" as an alias to "code"

sejas:
Perfect! That sounds good to me too. We can use `studio code` as the primary subcommand. We can keep the `ai` alias too 👍

sejas:
I created the PR renaming the command to `studio code`: https://github.com/Automattic/studio/pull/3002

PR #2991, opened 2026-04-06
Add multisite warning to push dialog for WordPress.com targets

bcotrim:
## Related issues

- Fixes STU-1377

## How AI was used in this PR

Implementation was AI-assisted. All code was reviewed and iterated on manually.

## Proposed Changes

- Add multisite detection utility that reads `wp-config.php` for the `MULTISITE` constant
- Add `getIsMultisite` IPC handler and `useIsMultisite` React hook (same pattern as `getWpVersion`/`useGetWpVersion`)
- Show a non-blocking warning in the push dialog when pushing a multisite site to WordPress.com (Pressable is excluded since it supports multisite)
- Refactor push dialog warnings into a collapsible "N warnings — Show/Hide" section to keep the UI clean when multiple warnings are present
- Replace the semi-transparent blurred bottom bar with an opaque background for better readability

| Push (hidden) | Push | Preview Sites |
|--------|--------|--------|
| image | image | image |

## Testing Instructions

- Import a blueprint with `"step": "enableMultisite"` to create a multisite site
- Connect the site to a WordPress.com site
- Open the Push dialog — verify the multisite warning appears in a collapsible warnings section
- Click "Hide" to collapse warnings, click "Show" to expand — verify the toggle works and padding adjusts
- If a version mismatch warning is also present, verify both warnings show with a "2 warnings" label
- Connect to a Pressable site instead — verify the multisite warning does NOT appear
- Open a Pull dialog — verify no warnings section appears
- Run `npm test -- tools/common/lib/tests/is-multisite.test.ts` — all 7 tests pass

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/modules/sync/components/sync-dialog.tsx
+++ b/apps/studio/src/modules/sync/components/sync-dialog.tsx
@@ -289,7 +289,7 @@ export function SyncDialog( {
 			padding += 80;
 		}
 		if ( warningCount > 0 ) {
-			padding += warningsExpanded ? 80 * warningCount + 30 : 30;
+			padding += warningsExpanded ? 70 * warningCount + 30 : 20;
 		}
 		return padding;
 	};
@@ -469,7 +469,7 @@ export function SyncDialog( {
 								) }
 								<Button
 									variant="link"
-									className="!p-0 !h-auto text-[12px] [&.is-link]:text-frame-text-secondary [&.is-link]:hover:text-frame-theme"
+									className="!p-0 !h-auto text-[12px] leading-[16px] [&.is-link]:text-frame-text-secondary [&.is-link]:hover:text-frame-theme"
 									onClick={ () => setWarningsExpanded( ! warningsExpanded ) }
 								>
 									{ warningsExpanded ? __( 'Hide' ) : __( 'Show' ) }
--- a/apps/studio/src/modules/sync/components/sync-dialog.tsx
+++ b/apps/studio/src/modules/sync/components/sync-dialog.tsx
@@ -1,6 +1,12 @@
-import { SelectControl, Notice, __experimentalHeading as Heading } from '@wordpress/components';
+import {
+	Icon,
+	SelectControl,
+	Notice,
+	__experimentalHeading as Heading,
+} from '@wordpress/components';
 import { createInterpolateElement } from '@wordpress/element';
 import { sprintf, __ } from '@wordpress/i18n';
+import { cautionFilled } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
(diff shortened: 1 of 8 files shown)


apps/studio/src/modules/sync/components/sync-dialog.tsx
```diff
@@ -444,14 +455,49 @@ export function SyncDialog( {
...
+									warningCount
 								) }
-							</p>
-						</Notice>
+								<Button
+									variant="link"
+									className="!p-0 !h-auto text-[12px] [&.is-link]:text-frame-text-secondary [&.is-link]:hover:text-frame-theme"
+									onClick={ () => setWarningsExpanded( ! warningsExpanded ) }
+								>
+									{ warningsExpanded ? __( 'Hide' ) : __( 'Show' ) }
```
katinthehatsite, line 475: I am not sure if it is me but this is looking slightly misaligned for me by a couple of pixels:

Image

e.g. Hide and Show seem slightly higher than the warning

katinthehatsite (commented):
Thanks for all the changes - I like that there is an option to hide the warnings so that they  are not taking too much space. Should we slightly drop the space below the database when the modal is open?

Image

I am also wondering - do you know if multisites could create unexpected issues for the preview sites?

katinthehatsite:
Another question that I forgot to add to the review: I am currently seeing that when you open the modal, it shows by default the warnings and then you can hide them. Should we hide them by default so that they don't initially take too much space and then the user can open the later?

bcotrim:
Thanks for the review @katinthehatsite
I've made some changes to the layout and added the same warning on preview sites (it already existed for mismatched PHP/WordPress versions)
Can you take another look, please?

> Another question that I forgot to add to the review: I am currently seeing that when you open the modal, it shows by default the warnings and then you can hide them. Should we hide them by default so that they don't initially take too much space and then the user can open the later?

It was intentional, since it could be an operation that breaks the production site, I think having them open by default increases the chance our users won't miss it.
What do you think?

katinthehatsite:
> It was intentional, since it could be an operation that breaks the production site, I think having them open by default increases the chance our users won't miss it.
What do you think?

I don't have a strong opinion as I can see the benefits either way. I will give it another review 👀

katinthehatsite (approved):
I have retested it and I think it looks and works better. Also thanks for adding it to the preview sites 👍 

I am not sure what is the expected design for this though. It might be worth confirming this with design. On my side, I don't have any further comments

sejas (approved):
Works as expected. I see the warning when pushing to WPcom but not to Pressable. I also see a warning when creating a preview site.

| Push WPcom | Push Pressable | Preview site |
|--------|--------| --- |
|  Image | Image | Image |

PR #3000, opened 2026-04-07
More efficient dependency management in CLI

fredrikekelund:
## Related issues

- Related to p1775567333482379/1775561944.426079-slack-C06DRMD6VPZ

## How AI was used in this PR

With directed tasks and for multiple rounds of reviews. 

## Proposed Changes

https://github.com/Automattic/studio/pull/2873 moved the site dependencies ("server files") so they're managed by the CLI instead of by Studio. Most of the `wp-files` ➡️ `server-files` copying logic remained the same, but a big difference is that `setupServerFiles` is now called on every CLI invocation. Previously, it was only called when Studio started (and maybe when starting a site). This meant that `wp-files/phpmyadmin`, `wp-files/sqlite-command`, `wp-files/latest/languages`, and `wp-files/skills` were copied every time the CLI ran, which is quite expensive.

This PR adds version-comparison logic to `copyBundledSqliteCommand` and `copyBundledPhpMyAdmin`. `copyBundledTranslations` and `copyBundledLanguagePacks` get an rsync-like algorithm that recursively compares the source and target directories, looking for new files, and then comparing filesize+mtime. `copyBundledWpCli` and `copyBundledTranslations` get an equivalent algorithm, but we compare just a single file.

I've also updated `recursiveCopyDirectory` to preserve timestamps when copying files, and to use copy-on-write for additional performance benefits (when the file system supports it).

Lastly, the Vite build now preserves the file timestamps in the `wp-files` directory it copies to the `dist` directory. Without this change, the dependency setup logic would always copy the files again after running a CLI build. 

`setupServerFiles` takes ~55ms on this branch, compared to ~520ms on trunk. In other words, a 10x speedup. This overhead applies to every single CLI invocation, so there's still room for improvement here. The lowest hanging fruit would be to generate some kind of version file for the `wp-files/latest/languages` directory that we can use for comparison instead of having to iterate over the complete directory. This would be much faster. 

## Testing Instructions

1. `npm run cli:build`
2. `rm ~/.studio/server-files/language-packs/pt_BR.l10n.php`
3. `node apps/cli/dist/cli/main.mjs site list`
4. `ls ~/.studio/server-files/language-packs/pt_BR.l10n.php`
5. Ensure the files exists again

## Pre-merge Checklist

### DIFF
--- a/apps/cli/lib/dependency-management/setup.ts
+++ b/apps/cli/lib/dependency-management/setup.ts
@@ -18,6 +18,49 @@ import { areDirectoriesDifferentBySizeAndMtime } from './utils';
 import { getWordPressVersionFromInstallation, updateLatestWordPressVersion } from './wordpress';
 import { updateLatestWpCliVersion } from './wp-cli';
 
+type VersionReader = () => Promise< semver.SemVer | null >;
+
+async function copyBundledDependencyIfNewerOrMissing( {
+	sourceDirectoryPath,
+	targetDirectoryPath,
+	readSourceVersion,
+	readTargetVersion,
+}: {
+	sourceDirectoryPath: string;
+	targetDirectoryPath: string;
+	readSourceVersion: VersionReader;
+	readTargetVersion: VersionReader;
+} ) {
+	if ( ! fs.existsSync( sourceDirectoryPath ) ) {
+		return;
+	}
+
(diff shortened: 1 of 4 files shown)


bcotrim (commented):
Code LGTM, nice cleanup 👍

PR #3001, opened 2026-04-07
Fix copy for php and wp version mismatch

nightnei:
## Proposed Changes

We should not mention wordpress.com in the version mismatch notice (Sync dialog), since it's used for both wpcom and Pressable.

The new notice:
Screenshot 2026-04-07 at 15 51 38

## Testing Instructions
Visual review should suffice

### DIFF
--- a/apps/studio/src/modules/sync/components/sync-dialog.tsx
+++ b/apps/studio/src/modules/sync/components/sync-dialog.tsx
@@ -451,7 +451,7 @@ export function SyncDialog( {
 						<Notice status="warning" isDismissible={ false } className="mb-4">
 							<p data-testid="push-version-mismatch-notice">
 								{ __(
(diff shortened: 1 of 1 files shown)


apps/studio/src/modules/sync/components/sync-dialog.tsx
```diff
@@ -448,7 +451,7 @@ export function SyncDialog( {
 						<Notice status="warning" isDismissible={ false } className="mb-4">
 							<p data-testid="push-version-mismatch-notice">
 								{ __(
-									'Your Studio site is using a different WordPress or PHP version than your WordPress.com site. The remote site will keep on using the newest supported versions.'
+									'Your Studio site is using a different WordPress or PHP version than your remote site. The remote site will keep using the newest supported versions.'
```
nightnei, line 454: Removing "on" was recommended by AI, however, I find both options good.

katinthehatsite (approved):
Nice catch!

PR #3005, opened 2026-04-08
Strip unnecessary fields in wpcom_request response

epeicher:
## Related issues

- Fixes STU-1524

## How AI was used in this PR

AI assisted with debugging the root cause (identifying that `plan.features` was 63K+ chars) and writing the compact extraction logic.

## Proposed Changes

- **System prompt guidance**: Add a hint instructing the agent to use `_fields` on wp/v2 listing endpoints (templates, posts, pages, etc.) to request only lightweight fields and exclude heavy ones like `content`. This is the generic solution for large responses.
- **Special case for `plan.features`**: Add a `compactResponse()` function that strips `plan.features` (60K+ chars) from `/sites/{id}` responses, keeping only `product_id`, `product_slug`, `product_name_short`, `expired`, and `is_free`. This can't be solved via query params because the v1.1 API doesn't support sub-field filtering (e.g. `fields=plan.product_slug`). The function includes a clear comment explaining this is a special case, not a pattern to follow.

## Testing Instructions

- Apply these changes and run `npm run cli:build`. Then start studio ai by running `node apps/cli/dist/cli/main.mjs ai`
- Connect to a remote WordPress.com site by pressing down arrow and then right arrow
- Ask the agent a question that triggers `wpcom_request` (e.g., "What's the current name, tagline, and active theme of this site?")
- Verify the response succeeds without exceeding the MCP output limit, i.e. you don't see the error: `Error: result (96,907 characters) exceeds maximum allowed tokens. Output has been saved to…`
- Ask the agent to modify a block theme template on a remote site → verify it uses wp/v2 endpoints
- Ask the agent for design changes on a free WP.com site → verify it explains the paid plan limitation

|Before | After|
|-------|------|
|CleanShot 2026-04-08 at 13 17 00@2x| CleanShot 2026-04-08 at 13 14 41@2x|

## Pre-merge Checklist


apps/cli/ai/wpcom-tools.ts
```diff
@@ -3,6 +3,56 @@ import wpcomFactory from '@studio/common/lib/wpcom-factory';
...
+ *    heavy per-item fields (content, _links) so the agent can identify items and
+ *    fetch individual ones for full content.
+ */
+function compactResponse( result: ApiResponse ): ApiResponse {
+	if ( ! result || typeof result !== 'object' ) {
+		return result;
+	}
+
+	// Case 1: plan.features can be 60K+ chars — keep only essential plan properties
+	if ( ! Array.isArray( result ) && result.plan?.features ) {
```
youknowriad: For me this seems too specific to a given endpoint, Is there a way to apply a generic fix instead (like overriding the token limit or telling the agent to use the minim required properties...). I don't believe we should have specifics like that because it will be hard to scale and adapt as the endpoints evolve...

epeicher: Good point about scalability. I looked into a few alternatives:

**Prompt-based**: Tell the agent to always use `_fields` (wp/v2) or `fields` (v1.1) to request only what it needs. Works for most endpoints but doesn't fix the `plan` case. Even with `fields=ID,name,description,URL,plan`, the `plan` field alone is 63K+ because `features` is deeply nested and the API doesn't support sub-field filtering like `fields=plan.product_slug`. Plus the agent might not always follow the instruction.

**Increasing the token limit**: Would work but the LLM ends up processing 60K+ chars of `plan.features` that it never reads. The system prompt already hardcodes what each plan tier can do, so the agent only looks at `product_slug` and `is_free`. Same story with array listings where `content` and `_links` get ignored when the agent is just browsing. Feels wasteful to bump the token budget for data that gets thrown away.

**On the coupling concern**: I don't think these endpoint shapes will change that much. `plan.features` has had this structure for a long time, and `content`/`_links` are standard WP REST API fields. But I get the concern about hardcoding endpoint-specific stuff.

**Proposal**: Add prompt guidance so the agent uses `_fields`/`fields` by default, and keep `compactResponse()` as a safety net for what field filtering can't solve (mainly `plan.features`).

That said, if you'd rather avoid the coupling I'm also fine with just increasing the token limit and updating the prompt to use fields.  What do you think?

youknowriad: If it's just one specific case, I'm fine with it if we add a good comment to explain that it's a special case and not a pattern to follow, with the fields guidance for the generic case.

Do you think the fields guidance will solve all the other cases?

Curious about @lezama's thoughts as well as he was working on alternative solutions.

epeicher: I have applied the suggestions as part of [c457f53](https://github.com/Automattic/studio/pull/3005/commits/c457f53285405c5a502871d1e6e16bb946efdcad) and tested different scenarios, and I have not found the error. The suggested changes seem to do the trick. 

@lezama, the approach of trimming the output [suggested here](https://github.com/Automattic/studio/pull/3011/changes#diff-8bc987fb45f0d99a9ad64c2c77c6fb430ec58bd3d1be1a3a300004f5e70f4a79R121) is an alternative option that can be applied on top of this changes so we ensure the error is not displayed to the user. In some testing, I have found that the limit is not exact so we can add some extra margin to prevent the error.

apps/cli/ai/wpcom-tools.ts
```diff
@@ -3,6 +3,39 @@ import wpcomFactory from '@studio/common/lib/wpcom-factory';
...
+ * Claude Code's MCP output limit (~100k chars). The API doesn't support sub-field
+ * filtering (e.g. `fields=plan.product_slug`), so we can't solve this via query
+ * params. The agent only needs a few plan properties to gate features since the
+ * system prompt hardcodes what each plan tier can do.
+ *
+ * This is NOT a pattern to follow for other endpoints. For general large responses,
+ * the system prompt instructs the agent to use `_fields` (wp/v2) or `fields` (v1.1)
+ * query params to request only the properties it needs.
+ */
+function compactResponse( result: ApiResponse ): ApiResponse {
```
sejas: I wonder if we should use a more specific function name, like `stripPlanFeatures`.

apps/cli/ai/wpcom-tools.ts
```diff
@@ -3,6 +3,39 @@ import wpcomFactory from '@studio/common/lib/wpcom-factory';
...
+/**
+ * Special case: the WP.com /sites/{id} endpoint returns a `plan` object whose
+ * `features` sub-field alone is 60K+ characters, which pushes the tool result past
+ * Claude Code's MCP output limit (~100k chars). The API doesn't support sub-field
+ * filtering (e.g. `fields=plan.product_slug`), so we can't solve this via query
+ * params. The agent only needs a few plan properties to gate features since the
+ * system prompt hardcodes what each plan tier can do.
+ *
+ * This is NOT a pattern to follow for other endpoints. For general large responses,
+ * the system prompt instructs the agent to use `_fields` (wp/v2) or `fields` (v1.1)
```
sejas, line 17: Is this accurate? I didn’t see the system prompt mention `fields` for v1.1. Should we add that information to the system prompt?

sejas (approved):
I think it makes sense to remove plan features if they are 60K!. I left a couple of suggestions.

apps/cli/ai/wpcom-tools.ts
epeicher: I agree, I have renamed it as part of e21a84e

apps/cli/ai/wpcom-tools.ts
epeicher: Great catch! I have updated the system prompt as part of [e21a84e](https://github.com/Automattic/studio/pull/3005/commits/e21a84e42b63e758cb1dca1dbedd7fbbf57db7f2) as I identified that sometimes `fields` were failing. I have added a suggestion to use both  `_fields` for `wp/v2` or `fields`  for `v1.1` to reduce the response size. I have tested multiple scenarios and they all work fine. I have tested:
- What's the current name, tagline, and active theme of this site?
- Redesign my site to be more polished
- Can you update the `coming-soon` template to have a light background instead of the current one?

And I have not obtained the error

apps/cli/ai/wpcom-tools.ts
```diff
@@ -106,7 +141,8 @@ export function createWpcomToolDefinitions( token: string, siteId: number ) {
 						break;
 				}
 
-				return textResult( JSON.stringify( result, null, 2 ) );
+				const compacted = stripOversizedFields( result );
+				return textResult( JSON.stringify( compacted, null, 2 ) );
```
lezama: I think it’s better to use `JSON.stringify(result)` instead of `JSON.stringify(compacted, null, 2)`, since the added indentation increases token usage. I learned this in https://github.com/Automattic/studio/pull/3011. I think we can close that PR in favor of this one?

epeicher: Thanks for the suggestion @lezama! I will apply that

> I learned this in https://github.com/Automattic/studio/pull/3011. I think we can close that PR in favor of this one?

I think so, and we can apply the learnings on #3011 on follow-ups if required.

epeicher: Hi @lezama, I went ahead and merged this. You'll want to merge trunk into https://github.com/Automattic/studio/pull/3005 to pick up the changes and continue from there or maybe just close it  as discussed

PR #3016, opened 2026-04-08
Two-phase site building with a new skill to convert html blocks to core Gutenberg blocks

epeicher:
## Related issues

- Related to STU-1439
- Supersedes #2976

## How AI was used in this PR

Claude was used to implement the two-phase workflow, create the blockify skill, update the system prompt, and remove the deterministic HTML block checker. All code was reviewed, iterated on, and tested by the author. 

## Proposed Changes

- **New `blockify` skill** (`apps/cli/ai/plugin/skills/blockify/SKILL.md`): A dedicated, user-invokable skill that handles the full block conversion workflow — reads back all content (posts + template parts), plans conversion element-by-element, rewrites with native blocks, validates markup, and does a final visual check. Includes block pattern reference, conversion mapping table, and explicit rules to prevent over-conservative skipping (e.g., always decompose sections rather than keeping them entirely as `core/html`).
- **Updated system prompt** (`apps/cli/ai/system-prompt.ts`): Kept trunk's linear workflow and added the `blockify` skill as the final conversion step. Dropped the redundant "Check the misuse of HTML blocks" step since `blockify` supersedes it, removed the intermediate `validate_blocks` call in favour of a single final run after conversion as a safety guard, and moved the `take_screenshot` visual verification to run after `blockify` so any regressions introduced during conversion are caught.

## Testing Instructions

- Apply this changes and run `npm run cli:build` and `node apps/cli/dist/cli/main.mjs ai`
- Build a site with a prompt that a user would add, focused on the website and potential styles, but not necessarily specifying technical details like (Gutenberg blocks), an example prompt could be "Build a one page site for a luxury furniture brand called Mikado, known for its precision and high quality manufactoring. Use simple Icons for company logos that are clients (Nasa, SpaceX, Uber, Visa, Grab, bose, Discover, dji, Nikon, craftsman, sony). Use light colors", but I encourage you to use a different one.
- Verify that the generated site uses Gutenberg blocks instead of `core/html` wrappers. For that, you can open `<site-domain>/wp-admin/site-editor.php`, click on the page and check the `Document overview` 
CleanShot 2026-04-08 at 18 51 38@2x
Additionally, you can try to edit any of the sections and check if you can update the text.

  
## Pre-merge Checklist


apps/cli/ai/system-prompt.ts
```diff
@@ -112,10 +112,10 @@ For any request that involves a WordPress site, you MUST first determine which s
 Then continue with:
 
 1. **Get site details**: Use site_info to get the site path, URL, and credentials.
-2. **Plan the design**: Before writing any code, review the site spec (from the site-spec skill) and the Design Guidelines below to plan the visual direction — layout, colors, typography, spacing.
+2. **Plan the design and block structure**: Before writing any code, review the site spec (from the site-spec skill) and the Design Guidelines below to plan the visual direction — layout, colors, typography, spacing. Also plan how each section maps to Gutenberg blocks: which sections use \`core/group\`, where to use \`core/columns\`, which text is \`core/heading\` vs \`core/paragraph\`, etc. Refer to the Block Content Guidelines for the correct markup patterns. Do NOT default to \`core/html\` — compose in native blocks from the start.
```
youknowriad: What I don't like much is the coupling of the system prompt to the core blocks. Aren't the validation changes sufficient.

I'm also wondering why you used a single tool for both kind of validation (block validation and html blocks) instead of using separate tools. I'm not opinionated here as long as the solution works though.

apps/cli/ai/system-prompt.ts
```diff
@@ -166,11 +166,104 @@ const REMOTE_DESIGN_GUIDELINES = `## Design capabilities by plan
 
 const LOCAL_CONTENT_GUIDELINES = `## Block content guidelines
 
-- Only use \`core/html\` blocks for:
-	- Inline SVGs
-	- \`<form>\` elements and interactive inputs
-	- Animation/interaction markup with no block equivalent (marquee, cursor)
-	- A single \`<script>\` block at the bottom of the page for JS
+**CRITICAL — Think in blocks, not HTML.** When writing page/post content or template parts, you MUST compose content using Gutenberg block markup from the start. Do NOT write raw HTML sections and wrap them in \`<!-- wp:html -->\`. Instead, use the block patterns below to build every section.
```
youknowriad: Did you find that the changes to the system prompt here are necessary? Can't we just rely on the tool for validation and the agent should be capable to fix things properly based on the errors thrown by the tool (like we do for regular block validation).

epeicher: That was my initial approach and didn't work well, the agent was in a loop that took around 45 minutes where:
- Initially created mainly html blocks
- Used the tool to replace the html blocks by other core blocks
- Refined the design and replaced the core blocks by html blocks 
- Validated again using the tool and replacing back to core blocks

The final result included a mixture of Groups and html blocks nested. I recorded a video with that first approach, you can skip to the relevant parts and see the loop 3ae9e-pb, for example on the minute 7:20 the agent had validated al the html blocks to use core blocks but it continues until minute 46 to complete the site.

Then I changed the system prompt to emphasize as much as possible using blocks, that improved the usage of blocks

youknowriad: I see, but doesn't this mean that the validation tool is now unnecessary (so basically your skill suggestion)?

apps/cli/ai/system-prompt.ts
epeicher: That's definitely a concern I have too, [shared here also](https://github.com/Automattic/studio/pull/2976#issuecomment-4208110701). The underlying problem  seems to be that the agent tends to use html blocks all the time. Also in [this comment](https://github.com/Automattic/studio/pull/3016#discussion_r3056857854)  I explain the issue I found with using only the tool.

I am going to experiment using external skills as an alternative or a more generic hint that doesn't couple that much with block implementations

epeicher: > I'm also wondering why you used a single tool for both kind of validation (block validation and html blocks) instead of using separate tools. I'm not opinionated here as long as the solution works though.

This was updated because during testing the AI agent reliably called `validate_blocks` on every piece of block content but consistently forgot to call `check_html_blocks` as a separate step

youknowriad:
One idea @mtias brings up all the time as well, is that maybe instead of asking our system prompt to generate the design in one step, we should split it into two steps: First generate the design in raw HTML + CSS, Second convert the HTML into blocks.

apps/cli/ai/system-prompt.ts
epeicher: Yes, it is a bit redundant now, but I found that the agent still uses it and updates some blocks. I am going to continue testing, and if I don't find that the tool helps, I will remove it

epeicher:
> One idea @mtias brings up all the time as well, is that maybe instead of asking our system prompt to generate the design in one step, we should split it into two steps: First generate the design in raw HTML + CSS, Second convert the HTML into blocks.

I like that approach, as I had success in converting one of the testing sites from html to core blocks in a different session. I am going to iterate over that 👍

mtias:
Yeah, two clean stages is worth a shot because there are various flows in which design > html+css is more straightforward for the model than design > blocks.

1. Two-pass workflow in the system prompt — Explicitly split site building into "design pass" (HTML/CSS is fine, get the visuals right) and "block conversion pass" (convert everything to core blocks, validate). Right now the prompt mixes both concerns and the agent often stops after the design looks good.
2. Validation gate — The validate_blocks tool exists but the agent treats it as optional. We can make the system prompt stricter: "After writing any template or page content, you MUST run validate_blocks. If any core/html blocks are found that could be core blocks, convert them before proceeding."
3. A post-build audit step — Similar to how /classic-to-blocks is meant to work as a dedicated conversion pass, we could add an automatic "blockify" step at the end of every site build. The system prompt would say: "After building the site and taking screenshots, run a final block audit: list all core/html blocks and convert each one to the appropriate core block equivalent."

epeicher:
@mtias, @youknowriad, I have applied the [suggestion here](https://github.com/Automattic/studio/pull/3016#issuecomment-4213538972) of a two-pass approach and it works great. 

I have created the following site https://rarandalopez-mxqjv-studio.wp.build/ with the following prompt:

>Build a one page site for a luxury furtniture brand called "Auran Blocks 6", known for its precision and high quality manufactoring. Use elegant icons for company logos that are clients (Nasa, SpaceX, Uber, Visa, Grab, bose, Discover, dji, Nikon, craftsman, sony). Use light colors

and the result looks nice, and most of the components are editable

CleanShot 2026-04-10 at 18 00 12@2x

I have created an external skill that I reference from the system prompt, and I have also tested it stand-alone by converting another previously created site with a "Can you blockify this?" prompt. This also worked great.

Please let me know your view on the changes

chubes4:
Ideally, this is something that could be handled server-side without needing an AI skill, which will burn more tokens. 

See: https://github.com/WordPress/gutenberg/issues/13163

If we defined core block transforms in PHP, then the agent could simply provide HTML while Gutenberg handles the conversion internally. 

Similar to the client-side block conversion that already exists.

epeicher:
@chubes4, good point, a server-side approach using PHP block transforms would be more efficient and deterministic. That’s worth exploring as a follow-up.

The current solution with the blockify skill is a pragmatic step forward: it improves the status quo by producing editable native blocks for the user, rather than leaving everything as core/html. I’d say we could iterate on the underlying mechanism (e.g., leveraging Gutenberg’s block transforms server-side and optimizing the usage of tokens) in a subsequent iteration. Would be interested to hear others’ thoughts on this as well

youknowriad:
I'm testing this. Here are the things that I noticed:

 - The phase 2 did convert most of the blocks successfully. 
 - The phase 2 broke some very subtle things (for instance the buttons had a double background). I wonder if we can add a validation step the phase2 and tell it something like: "Validate that the block conversion didn't cause design regressions, or compare with the screenshots taken after phase 1" or something like that.
 - Not everything that can be converted was converted, I still had some small divs and text that was not converted. (A fairly small number though). Not sure if we can do something here but it's acceptable for me.

The other thing that strikes me is that I expected the two phases to be something like:

 - Build a static html/css representation (so like a static site)
 - Transform it to a block theme with actual blocks.

but it was more

 - Build a WP site and theme without using blocks.
 - Transform the site and pages to blocks.

Maybe not that important though.

epeicher:
Thanks for testing @youknowriad! 

I am going to work on `The phase 2 broke some very subtle things` as I have also noticed that on some of my tests.

I will also work on the other points, but I agree they could be less important than the design mistakes that phase 2 currently introduces.

apps/cli/ai/system-prompt.ts
```diff
@@ -112,11 +108,18 @@ For any request that involves a WordPress site, you MUST first determine which s
...
+3. **Build the site**: Write theme files (style.css, functions.php, templates, template parts) and create pages/posts with wp_cli. During this phase, write content as clean HTML — you may use \`core/html\` blocks or raw HTML freely. Focus entirely on getting the visuals right. All CSS goes in style.css as usual. The \`wp_cli\` tool takes literal arguments, not shell commands: never use shell substitution or shell syntax such as \`$(cat file)\`, backticks, pipes, redirection, environment variables, or host temp-file paths to provide post content. Pass the literal content directly in \`--post_content=...\`, make \`--post_content\` the final argument in the command, and Studio will rewrite large content to a virtual temp file automatically.
+4. **Visual review**: Use take_screenshot to capture the site's landing page on desktop and mobile and verify the design visually on both viewports. Check for wrong spacing, alignment, colors, contrast, borders, hover styles, and other visual issues. Fix any issues found. Pay particular attention to the navigation menu and the CTA buttons. Iterate until the design is excellent.
```
youknowriad: I wonder if this is needed. Right now the agent is really slow, it takes too much screenshots all the time. I wonder if one "validation" phase at the end is better perf wise.

epeicher: I have measured the site creation time of the current changes (as at commit [ea7817c](https://github.com/Automattic/studio/pull/3016/commits/ea7817cd8239d82b0fc66d50e389446db4e46d76)) and they are much longer than the version on trunk (~ 18 minutes vs ~ 8 minutes on `trunk`). If on `trunk` we apply the `blockify` skill, it conversts successfully the blocks in ~ 5 minutes so in total it would be ~ 13:30 minutes. Given that, I have reverted the changes to be as similar as trunk as possible with the latest step being the `blockify` and the validations and it seems to improve, now it takes ~ 14 minutes.

I am going to continue doing some testing but the latest changes improve the multiple loops

epeicher: I have refined the system prompt to be:
- Similar to `trunk` version
- Added a step to transform html to core blocks
- Validate as the last step

The time it has taken in some of my tests is ~ 12 minutes, which is longer than `trunk`, which takes ~ 8 minutes, but we are adding a step.

This is ready for another 👀

chubes4:
This draft moves block conversion to PHP: https://github.com/Automattic/studio/pull/3309

One shot, full static site, imported into a custom block theme. No need to use validate_blocks, teach the AI block markup, or force screenshots. This can significantly speed up generation.

epeicher:
> This draft moves block conversion to PHP: https://github.com/Automattic/studio/pull/3309

Thanks @chubes4, that looks promising! I’m closing this PR in favor of #3267.

@youknowriad, you might want to take a look at draft #3309 as well.

PR #3018, opened 2026-04-08
Studio: Filter out deprecated warnings from fatal errors

katinthehatsite:
## Related issues

Related to STU-1529

## How AI was used in this PR

It was used to identify the solution after the fix raised in https://github.com/Automattic/wp-cli-sqlite-command/pull/24 uncovered the issue.

## Proposed Changes

PHP 8.5 triggers a deprecation notice from `react/promise` inside  `wp-cli.phar`. Studio treats any stderr output as a fatal export failure, so the notice was causing the export to fail even though the actual export succeeded. This PR ensures that we filter out the deprecation warning from the actual fatal error:

```
Error invoking remote method 'exportSiteForPush': Error: Database export      
  failed: PHP Deprecated: Case statements followed by a semicolon (;) are       
  deprecated, use a colon (:) instead in                                        
  phar:///tmp/wp-cli.phar/vendor/react/promise/src/functions.php on line 369    
 ``` 

This would ideally need to be fixed upstream but I think this fix should be sufficient for now.                                                         

## Testing Instructions

* Ensure that you have a newly created site with PHP version 8.5
* First, you will need to apply the instructions from https://github.com/Automattic/wp-cli-sqlite-command/pull/24 to uncover this issue
* Once you see that the push fails due to the deprecation warning, pull the changes from this branch
* Navigate to the `Sync` tab
* Connect a WP.com site
* Do the push process
* Confirm that the push process succeeds 

## Pre-merge Checklist


fredrikekelund (commented):
We now have two importer/exporter implementations: one in `apps/cli` and another in `apps/studio`. 

In `apps/cli`, we only throw if WP-CLI returned a non-zero exit code, which I believe is the correct way to handle errors there.

I say let's do the same thing here.

katinthehatsite:
> I say let's do the same thing here.

Checking now 👀

katinthehatsite:
@fredrikekelund the issue should now be addressed, let me know if you have any further comments.

fredrikekelund (approved):
LGTM 👍 I'd probably make the equivalent change in `apps/studio/src/lib/import-export/export/exporters/default-exporter.ts`, too. Full-site exports already work when I test it with a PHP 8.5 site, but this change would ensure correctness.

katinthehatsite:
> LGTM 👍 I'd probably make the equivalent change in apps/studio/src/lib/import-export/export/exporters/default-exporter.ts, too. Full-site exports already work when I test it with a PHP 8.5 site, but this change would ensure correctness.

Should now be done 👍

PR #3026, opened 2026-04-09
Fix sidebar top padding on Windows

nightnei:
## Proposed Changes
I don't know why we didn't notice it before, but on Windows we have huge gap at the top of main sidebar. I checked - we always had it. I propose to align it with macOS, since it feels like something is broken and I don't see a reson to have different padding top.

Screenshot 2026-04-09 at 14 08 04

Fix sidebar top padding on Windows

## Testing Instructions
Run Studio on MacOS and Windows and assert that top padding looks good
Screenshot 2026-04-09 at 13 57 43
Screenshot 2026-04-09 at 14 08 46

### DIFF
--- a/apps/studio/src/components/main-sidebar.tsx
+++ b/apps/studio/src/components/main-sidebar.tsx
@@ -16,12 +16,7 @@ export default function MainSidebar( { className }: MainSidebarProps ) {
 	return (
 		<div
 			data-testid="main-sidebar"
-			className={ cx(
-				'text-chrome-inverted relative',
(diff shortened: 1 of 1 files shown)


epeicher (approved):
Thanks @nightnei for fixing this! I have tested it, and it looks better now. LGTM!

|Before | After|
|-------|------|
| Image|  Image|

katinthehatsite (approved):
I checked for regressions for macOS and the changes look good for me there 👍

PR #2850, opened 2026-03-19
Fix: CLI doesn't generate site thumbnails

gcsecsey:
## Related issues

- Fixes STU-1380

## How AI was used in this PR

Claude was used to iterate on an initial plan, then to carry out the changes.

## Proposed Changes

- Create a shared thumbnail capture function `captureSiteThumbnail` used by both the CLI events subscriber and the desktop UI
- Remove duplicate thumbnail logic from `loadThemeDetails` and delegate to `captureSiteThumbnail`
- Add deduplication logic to prevent concurrent BrowserWindow creation when multiple code paths capture the same site's thumbnail
- Update tests

## Testing Instructions

1. Build CLI with `npm run cli:build`
2. Start the app with `npm start`
3. In a new terminal, create a site via CLI
  ```
  node apps/cli/dist/cli/main.js site create --path ~/Studio/cli-thumb-test
  ```
4. In the desktop app, check that the site appears with a thumbnail
5. Create a new site on the desktop app and check that thumbnail generation still works as expected

| trunk | this branch |
|--------|--------|
| image | image | 

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #3028, opened 2026-04-09
Fix site thumbnail not displaying correctly for newly created sites

fredrikekelund:
## Related issues

- Fixes STU-1541

## How AI was used in this PR

To help quickly implement a fix after I had diagnosed the issue. 

## Proposed Changes

Since https://github.com/Automattic/studio/pull/2850, Studio generates site thumbnails much more frequently. This leads to a race condition when creating a new site, causing Studio to display `Preview unavailable` instead of the site thumbnail. This PR fixes that. 

More details: when the `thumbnailData` value changes rapidly in `ContentTabOverview`, the first base64-encoded image doesn't have time to load before the value changes again. This triggers an `error` event, which removes the `` from the DOM and replaces it with the text `Preview unavailable`. When the `` isn't in the DOM, it doesn't load, and so subsequent `thumbnailData` changes never make the image appear. 

This PR fixes the issue by simply resetting the `isThumbnailError` state whenever `thumbnailData` changes.

## Testing Instructions

1. `npm start`
2. Create a new site
3. Ensure that the site thumbnail displays correctly

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/modules/cli/lib/cli-events-subscriber.ts
+++ b/apps/studio/src/modules/cli/lib/cli-events-subscriber.ts
@@ -66,9 +66,6 @@ const handleSiteEvent = sequential( async ( event: SiteEvent ): Promise< void > 
 	}
 
 	void sendIpcEventToRenderer( 'site-event', event );
-	if ( running ) {
-		void captureSiteThumbnail( siteId, false );
-	}
 } );
 
 let subscriber: ReturnType< typeof executeCliCommand > | null = null;
--- a/apps/studio/src/components/content-tab-overview.tsx
+++ b/apps/studio/src/components/content-tab-overview.tsx
@@ -15,7 +15,7 @@ import {
 	widget,
 } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { ArrowIcon } from 'src/components/arrow-icon';
 import { ButtonsSection, ButtonsSectionProps } from 'src/components/buttons-section';
 import { useSiteDetails } from 'src/hooks/use-site-details';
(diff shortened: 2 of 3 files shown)


katinthehatsite:
Just a sidenote: it seems that there is a wrong Linear issue linked in the PR description.

katinthehatsite (commented):
Currently, what I am seeing on my end is a brief flash of `Preview unavailable` and then the preview appears:

https://github.com/user-attachments/assets/f26bec30-7270-4847-aba9-f6009c289508

I am wondering if we could make the loading state persist until it loads completely. If there is no straightforward solution, we could merge then as it is anyway an improvement in user experience

fredrikekelund:
> it seems that there is a wrong Linear issue linked in the PR description.

Updated 👍

> what I am seeing on my end is a brief flash of Preview unavailable and then the preview appears

Hmm… I'll take a look. I'm sure we can find a relatively simple solution.

apps/studio/src/modules/cli/lib/cli-events-subscriber.ts
```diff
@@ -66,9 +66,6 @@ const handleSiteEvent = sequential( async ( event: SiteEvent ): Promise< void >
 	}
 
 	void sendIpcEventToRenderer( 'site-event', event );
-	if ( running ) {
-		void captureSiteThumbnail( siteId, false );
-	}
```
fredrikekelund, line 71: This change fixes the issue you identified in https://github.com/Automattic/studio/pull/3028#pullrequestreview-4088298284, @katinthehatsite. We don't need to capture screenshots for every `UPDATED` site event – it's enough to do it on `CREATED` events. 

Studio already triggers a new screenshot capture when the window is re-focused and when starting a server.

gcsecsey (approved):
Thanks for working on a fix for this @fredrikekelund! 🙌

With the most recent changes, I see no flickering and the thumbnails are displaying as described. 👍 

https://github.com/user-attachments/assets/c9a909c4-af74-435e-886e-c8b15384cc56

PR #3036, opened 2026-04-09
Disable autoStart when server crashes on startup

bcotrim:
## Related issues

- Fixes STU-1418

## How AI was used in this PR

## Proposed Changes

- Move `updateSiteAutoStart()` in the CLI `site stop` command to run **before** the `isServerRunning` check, so the autoStart flag is always persisted even when the server isn't running (e.g., it crashed before fully starting)
- Update `SiteServer.stop()` to always reset `running` and `autoStart` in memory, regardless of whether the server was running or the stop command threw an error

**Root cause:** When an `autoStart=true` server crashed during startup, the renderer called `stopServer()` which ran `site stop`. The CLI stop command checked `isServerRunning()` first — since the server crashed before starting, it returned early without updating `autoStart` to `false`. On next launch, the same crash repeated.

## Testing Instructions

1. Start two sites via CLI, then stop all with `--auto-start` flag to simulate Studio quit
2. Break one site (e.g., add a mu-plugin with `ini_set('memory_limit', '1M'); while(true) { $arr[] = str_repeat('x', 1024*1024); }`)
3. Launch Studio — the broken site should fail to auto-start with an error dialog
4. Check `~/.studio/cli.json` — the broken site should now have `autoStart: false`
5. Quit and relaunch Studio — the broken site should **not** attempt to auto-start again
6. Verify healthy sites still auto-start correctly

## Pre-merge Checklist


gcsecsey (approved):
Thanks @bcotrim for working on this! I think this is a nice quality of life improvement! 🙌 

Changes look good to me, and I confirm that the site with the error is not attempted to be started after the error. 👍 

The `autostart` flag is set to false after the error, and it's kept false after subsequent restarts:
Image

PR #3031, opened 2026-04-09
Add PromptFoo evaluation framework for Studio Code agent

lezama:
## Summary

Adds a [PromptFoo](https://www.promptfoo.dev/) evaluation framework for the Studio Code AI agent. A custom eval runner hooks into `startAiAgent()` to test the full agent loop — tool calls, permission flows, and output quality.

## How it works

```
npm run eval → promptfoo → eval-runner.ts → startAiAgent() → structured JSON → assertions
```

The runner returns raw data (tool calls, tool results, text segments, permission questions). All assertion logic stays in the YAML config.

A custom grader provider calls Claude Haiku via the WP.com AI proxy using Studio's existing auth token — no separate API key needed for local development.

## Test cases

| Test | What it checks | Type |
|------|---------------|------|
| **identity** | Agent identifies as WordPress Studio AI | LLM judge |
| **site-creation** | `site_create` tool called and succeeds | deterministic |
| **security** | Agent asks permission before writing outside `~/Studio` | deterministic |

## Changes

- **`apps/cli/ai/eval-runner.ts`** — Runner that hooks into `startAiAgent()`, captures tool calls/results/text/questions, returns JSON
- **`apps/cli/vite.config.dev.ts`** — eval-runner as dev-only Vite entry point (not in production builds)
- **`eval/grader-provider.mjs`** — PromptFoo grader that calls Claude Haiku via WP.com proxy for LLM judge assertions
- **`eval/promptfoo.config.yaml`** — 3 test cases with deterministic + LLM judge assertions
- **`eval/README.md`** — Usage docs
- **`package.json`** — `npm run eval` and `npm run eval:view` scripts

## Usage

```bash
studio auth login
npm run eval               # builds CLI, runs all 3 tests (~30s)
npm run eval -- -n 1       # run first test only
npm run eval:view          # view results in browser
```

## Follow-up

Daily CI workflow with Slack notifications — shipping separately once testable on trunk.

## Testing Instructions

1. `studio auth login`
2. `npm run eval`
3. Verify 3/3 tests pass
4. `npm run eval:view` to inspect results

## Quality gates

- [x] ESLint: pass
- [x] TypeScript: pass
- [x] Tests: 70/70 pass
- [x] Eval suite: 3/3 pass (32s)

PR #3042, opened 2026-04-09
Add daily eval CI workflow with Slack notifications

lezama:
## Summary

Adds a daily agent evaluation step to the existing Studio Buildkite pipeline. Runs automatically only on scheduled builds, otherwise it's behind a manual opt-in button. 

Stacked on #3031 — merge that first.

## Changes

- **`.buildkite/pipeline.yml`** — Eval step gated on `build.source == 'schedule'` or `EVAL_PIPELINE=true`
- **`.buildkite/commands/run-eval.sh`** — Installs deps, builds CLI, runs eval, posts results to Slack
- **`apps/cli/ai/eval-runner.ts`** — Skip provider resolution when `ANTHROPIC_API_KEY` is already set (needed in CI where there's no Studio auth)

## Setup (after merge)

An admin adds a schedule in the [Studio pipeline settings](https://buildkite.com/automattic/studio/settings/schedules):

- **Cron:** `0 6 * * *`
- **Branch:** `trunk`
- **Env vars:** `EVAL_SLACK_CHANNEL=C08LZ79MKG9`

`ANTHROPIC_API_KEY` and `SLACK_TOKEN` are already available from the pipeline's existing secrets. To trigger manually: new build with env var `EVAL_PIPELINE=true`.

## Slack notification

```
✅ Agent eval: 3/3 passed
• agent identifies as WordPress Studio AI: ✅
• agent calls site_create and it succeeds: ✅
• agent asks permission before writing outside ~/Studio: ✅
View build
```


wojtekn:
@lezama let's use Buildkite for any actions that build the app, and use GH actions only for basic stuff like static analysis, PR gardening, auto-merges, etc.

lezama:
@wojtekn updated to use buildkite, but needs lots of review, not sure what I am doing.

gcsecsey (commented):
Hey @lezama 👋 I'm checking in on this one since it has been idle for a bit. I found STU-1417, which has the broader goal of having an eval framework so we can track Studio AI quality over time, so this approach makes sense to me.

I tested this on Buildkite ([build](https://buildkite.com/automattic/studio/builds/16919#019e88dc-1f5d-45f6-9749-4b385cedb3b4)). I created a new build on this branch, using these env vars:
```
EVAL_PIPELINE=true
EVAL_TEST_FILTER=1
```

`EVAL_PIPELINE=true` makes the new Buildkite step to run, and `EVAL_TEST_FILTER=1` runs only one eval case, to make the test quicker and cheaper. I also omitted setting `EVAL_SLACK_CHANNEL` to  avoid sending a notification for now. 

The new `Agent Evaluation` step was picked up correctly, dependencies installed, and the CLI built, and `eval/results.json` was uploaded.

However, the eval itself did not actually run successfully. PromptFoo reported provider errors for all test cases: `Could not find any valid files in the command: node ../apps/cli/dist/cli/eval-runner.mjs`

Image

So it seems that the Buildkite wiring mostly works, but the eval command is being run from the wrong directory.

lezama:
hey @gcsecsey , thanks for the reminder about this one!

I gave it a pass and rebase to bring it up to date, also pushed [Add optional manual trigger to run the agent eval on a PR](https://github.com/Automattic/studio/pull/3042/commits/3415f92314c72043b177101124a7d9a01a8fcb8a)

In my local testing we need https://github.com/Automattic/studio/pull/3682 for tests to pass.

gcsecsey:
Thanks @lezama for updating this, I merged in the trunk changes and also noticed that we're using the wrong paths, so I also pushed a fix for that. 

The newly added Agent Evaluation step now runs, but fails on each test case to call the provider. I think we might be missing an `ANTHROPIC_API_KEY` env var that we check for in [eval-runner.ts](https://github.com/Automattic/studio/pull/3042/changes#diff-f4adcad8f173569fbbce4a53c390f90093863a5900f6e49822dc1b423350a836). Do you know maybe how we can add it to Buildkite?

.buildkite/commands/run-eval.sh
```diff
@@ -0,0 +1,66 @@
...
+' "$RESULTS_FILE")
+
+if curl -sf -o /dev/null -X POST https://slack.com/api/chat.postMessage \
+    -H "Authorization: Bearer $SLACK_TOKEN" \
+    -H 'Content-Type: application/json; charset=utf-8' \
+    --data-binary "$payload"; then
+  echo "Sent to $EVAL_SLACK_CHANNEL"
+else
+  echo "Warning: Slack notification failed"
+fi
```
iangmaia, line 75: Just a thought, not sure if it would be a worthwhile improvement: I wonder if this couldn't be replaced by a Fastlane action call (since we have utils to send slack messages). At the same time, that would require a `install_gems` and using the Ruby runtime, which this script saves so 🤷

.buildkite/commands/run-eval.sh
```diff
@@ -0,0 +1,66 @@
...
+  fi
+  eval_args+=( -n "$EVAL_TEST_FILTER" )
+fi
+npx promptfoo@0.121.4 eval -c scripts/eval/promptfoo.config.yaml "${eval_args[@]}"
+
+echo '--- :slack: Send Slack notification'
+if [[ -z "${EVAL_SLACK_CHANNEL:-}" || -z "${SLACK_TOKEN:-}" || ! -f "$RESULTS_FILE" ]]; then
+  echo "Skipping Slack notification (missing channel/token/results)"
+  exit 0
+fi
```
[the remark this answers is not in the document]
gcsecsey: Good catch — this defeats the whole point, since a failing eval is exactly when we want the ping. Fixed in 7bd31c3: capture the eval exit code instead of letting set -e abort, post to Slack, then exit "$eval_status" at the end so the step still goes red on failure.

.buildkite/commands/run-eval.sh
```diff
@@ -0,0 +1,66 @@
...
+  [.results.results[] |
+    "• " + (.testCase.description // .vars.caseId // "unknown") + ": " +
+    (if .success then "✅" else "❌" end)
+  ] | join("\n") as $lines |
+  (($s.successes + $s.failures + $s.errors) | tostring) as $total |
+  (if $s.failures == 0 and $s.errors == 0
+    then ":white_check_mark: Agent eval: " + ($s.successes | tostring) + "/" + $total + " passed"
+    else ":x: Agent eval: " + ($s.failures | tostring) + "/" + $total + " failed"
+  end) as $header |
+  (if $s.failures == 0 and $s.errors == 0 then "#36a64f" else "#e01e5a" end) as $color |
```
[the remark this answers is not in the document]
gcsecsey: Nice catch. Fixed in 8592071 — the failure header now uses $s.failures + $s.errors, so an errors-only run reads 3/3 failed instead of 0/3 failed.

.buildkite/commands/run-eval.sh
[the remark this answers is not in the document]
gcsecsey: Agreed. Fixed in 244f1bd: dropped curl -f, capture the response body and check .ok (logging .error when false), so a 200 with {"ok":false} no longer reports as sent.

apps/cli/ai/eval-runner.ts
```diff
@@ -134,21 +134,20 @@ async function runEval( input: EvalRunnerInput ) {
...
+		phaseTimingsMs.resolve_initial_provider_ms = Date.now() - phaseStartedAt;
+
+		phaseStartedAt = Date.now();
+		aiProvider = ( await resolveUnavailableAiProvider( aiProvider ) ) ?? aiProvider;
+		phaseTimingsMs.resolve_unavailable_provider_ms = Date.now() - phaseStartedAt;
+
+		phaseStartedAt = Date.now();
+		Object.assign( env, await resolveAiEnvironment( aiProvider ) );
+		phaseTimingsMs.resolve_ai_environment_ms = Date.now() - phaseStartedAt;
+	}
```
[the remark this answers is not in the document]
gcsecsey: Fixed in 2b787f1: the bypass now requires both `ANTHROPIC_API_KEY` and a CI marker (CI/BUILDKITE), so local npm run eval keeps going through normal provider resolution  even when a dev has the key set.

.buildkite/pipeline.yml
```diff
@@ -248,6 +248,30 @@ steps:
           - x64
           - arm64
 
+  # Optional manual trigger to run the agent eval on a PR. On scheduled builds
+  # (and EVAL_PIPELINE runs) this gate is skipped — a skipped dependency counts
+  # as passed — so the eval below runs automatically. On a PR it shows a button
+  # to opt in.
+  - input: ":test_tube: Run agent eval?"
```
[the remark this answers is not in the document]
gcsecsey: The behavior is intentional, I updated the PR description to match.

.buildkite/commands/run-eval.sh
gcsecsey: I think it's better to keep the curl version for now, to avoid pulling in `install_gems` + the Ruby runtime just for sending the Slack message.

gcsecsey:
Hey @iangmaia 👋 thanks for the review! I think we have a blocker with the env vars, and I'm hoping you can help.

When I [tested this](https://buildkite.com/automattic/studio/builds/17794#019edb1c-769e-4d22-b277-427e58b4325f) on Buildkite, the step builds and runs fine, but every eval case fails on the provider call. My guess is that `ANTHROPIC_API_KEY` isn't in the environment.

Is there a mechanism we can use to expose `ANTHROPIC_API_KEY` and `SLACK_TOKEN` to the job as an env var? If not, could you add these as Buildkite secrets?

.buildkite/commands/run-eval.sh
```diff
@@ -0,0 +1,78 @@
...
+eval_status=0
+npx promptfoo@0.121.4 eval -c scripts/eval/promptfoo.config.yaml "${eval_args[@]}" || eval_status=$?
+
+echo '--- :slack: Send Slack notification'
+if [[ -z "${EVAL_SLACK_CHANNEL:-}" || -z "${SLACK_TOKEN:-}" || ! -f "$RESULTS_FILE" ]]; then
+  echo "Skipping Slack notification (missing channel/token/results)"
+  exit "$eval_status"
+fi
+
+RUN_URL="${BUILDKITE_BUILD_URL:-https://buildkite.com}"
```
mokagio, line 33: Might it be cleaner to fail if `BUILDKITE_BUILD_URL` is not available? It's a core expectation of the CI environment after all.

.buildkite/commands/run-eval.sh
mokagio: > to avoid pulling in install_gems + the Ruby runtime just for sending the Slack message.

I'll push back on this. Given the script runs on the `mac` queue, which already has Ruby available, I think the advantage of delegating the Slack plumbing to fastlane would be worth while: less custom code for us to maintain.

This sounds like a refactor, though, no need to block the PR for it.

.buildkite/pipeline.yml
```diff
@@ -248,6 +248,30 @@ steps:
...
+    key: agent_eval
+    depends_on: input-eval
+    agents:
+      queue: mac
+    command: bash .buildkite/commands/run-eval.sh
+    plugins: [$CI_TOOLKIT_PLUGIN, $NVM_PLUGIN]
+    artifact_paths:
+      - scripts/eval/results.json
+    timeout_in_minutes: 25
+    if: build.source == 'schedule' || build.env('EVAL_PIPELINE') == 'true' || (build.branch != 'trunk' && build.tag !~ /^v[0-9]+/)
```
mokagio, line 273: If the intention is to run this step on a schedule, then with this setup where the only call is in `pipeline.yml`, the schedule would require running the whole pipeline "only" to get the eval result. 

Is this acceptable?

The alternative is to have a dedicated pipeline, something like `agent_eval_pipeline.yml` with a step like this one and no `if` conditions, and configure the schedule to run that pipeline.

The downside would be that we'd need to duplicate the step between `agent_eval_pipeline.yml` and `pipeline.yml` (with the `pipeline.yml` version having all of the `if` checks it currently has minus `build.source == 'schedule'` and possibly `build.env('EVAL_PIPELINE') == 'true'`.

mokagio: > The downside would be that we'd need to duplicate the step between agent_eval_pipeline.yml and pipeline.yml (with the pipeline.yml version having all of the if checks it currently has minus build.source == 'schedule' and possibly build.env('EVAL_PIPELINE') == 'true'.

Thinking about it, but I'm not 100% sure, given the `depends_on: input-eval`, we might do entirely away with the `if` in this step, and rely entirely on the conditions defined in the `input-eval` step.

mokagio:
> Is there a mechanism we can use to expose ANTHROPIC_API_KEY and SLACK_TOKEN to the job as an env var? If not, could you add these as Buildkite secrets?

@gcsecsey yes there is. Let's sync out of GitHub about how.

(Update: Internal ref p1782452433458179-slack-C06DRMD6VPZ)

.buildkite/pipeline.yml
gcsecsey: > If the intention is to run this step on a schedule, then with this setup where the only call is in pipeline.yml, the schedule would require running the whole pipeline "only" to get the eval result.

I'm not sure about the motivation for this change, but I'm not sure it's imperative that we run this on a schedule, but to run this often enough to catch any regressions. @lezama please correct me if I'm wrong.

Would it make sense to run this step automatically for builds from trunk? That'd ensure it runs at least daily.

PR #3050, opened 2026-04-10
Upgrade archiver to v7 to fix glob and inflight deprecation warnings

wojtekn:
## Related issues

- Fixes deprecation warnings for `glob@8` and `inflight` in the CLI and Studio dependency trees

## How AI was used in this PR

This PR was created with Claude Code assistance.

## Proposed Changes

- Upgrades `archiver` from `^6.0.2` to `^7.0.1` in both `apps/cli` and `apps/studio`
- Upgrades `@types/archiver` from `^6.0.4` to `^7.0.0` in both
- `archiver@7` pulls in `archiver-utils@5` which uses `glob@10` instead of `glob@8`, eliminating the `glob` and `inflight` deprecation warnings
- Drops `apps/studio/patches/@types+archiver+6.0.4.patch` — no longer needed
- Fixes type errors introduced by `@types/archiver@7` (applied consistently in both apps):
  - `followSymlinks` was removed from `CoreOptions` in the types (still supported at runtime) — fixed with `as archiver.ArchiverOptions` cast at call sites
  - `ProgressData` named import broken with `export=` module style — fixed by importing the `archiver` namespace type and using `archiver.ProgressData`
  - `vi.mocked(archiver)` cast updated to work with the new type shape in both test files

## Testing Instructions

- Run `npm test` — archive/export tests should pass
- Build the CLI: `npm run cli:build` and verify no deprecation warnings about `glob` or `inflight` appear

## Pre-merge Checklist


epeicher (commented):
Thanks @wojtekn for updating this! I think as part of this upgrade we need to test if the zips generated include `symLinks` and in that case we can remove the patch-package for archiver@6.0.2.

I have also found one warning when installing, not sure if it's related to this

Image

wojtekn:
@epeicher this PR already removes the types patch, but not the library patch one. I removed the other one now.

The follow-symlink still works. The source site:

```
% ls -l
total 16
drwxr-xr-x@ 17 USER staff   544 Apr 10 10:08 akismet
-rw-r--r--@  1 USER  admin  2671 Apr 10 10:08 hello.php
-rw-r--r--@  1 USER admin    28 Apr 10 10:08 index.php
lrwxr-xr-x@  1 USER staff    47 Apr 10 12:39 studio-demo-site-companion -> /Users/USER/Sites/studio-demo-site-companion
```

And exported package:

```
% ls -l studio-backup-my-sparkly-website-2026-04-10-12-39-40/wp-content/plugins 
total 16
drwxr-xr-x  18 USER staff   576 Apr 10 12:39 akismet
-rw-r--r--   1 USER staff  2671 Apr 10 08:08 hello.php
-rw-r--r--   1 USER staff    28 Apr 10 08:08 index.php
drwxr-xr-x   9 USER staff   288 Jul 16  2025 studio-demo-site-companion
```

youknowriad (approved):
Approving, I didn't see warnings when building but I'm honestly not 100% that I'm testing properly.

sejas (approved):
I got an error when exporting a site, but I was able to reproduce the same error on trunk , so it seems not related to this PR.

Image
Image

`npm run cli:build` doesn't show the library errors, although neither does on trunk. I think we could try it in next beta 🤞 .

wojtekn:
Thanks for reviews @youknowriad @epeicher and @sejas !

The error is caused by SQLite issue - I forgot to mention that we need to comment out the database export part to test it.

fredrikekelund:
@wojtekn, the `followSymlinks` option was something custom that @epeicher implemented (see https://github.com/archiverjs/node-archiver/pull/810). This PR seems to have removed both the `@types/archiver` and `archiver` patches, which makes me think we lost this behavior…

epeicher:
@fredrikekelund I think that is the default behaviour of that version (after a quick _visual_ look at the source code of the [transient dependency](https://github.com/archiverjs/node-zip-stream/blob/master/index.js)) and @wojtekn [has tested that](https://github.com/Automattic/studio/pull/3050#issuecomment-4222932356) so I think we are safe. If the generated zip includes symlinked files and folders, the patch is not required anymore.

But we should remove [the leftovers](https://github.com/Automattic/studio/blob/c9ed8c1537227fcfa1d9aadfd87718d0a70f36a3/apps/cli/lib/archive.ts#L20) if that's the case

fredrikekelund:
I just tested this and did not get the same results. My export contains a symlink, not the resolved contents of that symlink

fredrikekelund:
This does expose an issue in the previous implementation, though, which is that we needed an `archiver` patch for `apps/cli`, too

PR #3003, opened 2026-04-07
Fix dark mode scrollbar

bcotrim:
## Related issues

- Fixes STU-1426

## How AI was used in this PR

AI identified the root cause and wrote the CSS fix. Changes were visually verified by the developer before/after.

## Proposed Changes

> [!NOTE]
> I wasn't able to replicate the issue mentioned in Linear, but I still improved the layout of the scrollbar. 

- Add `color-scheme: dark` to the dark mode media query so the browser renders native UI elements (scrollbars, form controls) with dark styling
- Replace hardcoded color values in `.sites-scrollbar` with existing CSS custom properties (`--color-frame-scrollbar-thumb`, `--color-frame-scrollbar-border`, `--color-frame-scrollbar-thumb-hover`) for consistency with other scrollbar styles

| trunk | now |
|--------|--------|
| image | image | 

## Testing Instructions

- Switch Studio to dark mode (Settings > Appearance > Dark, or set macOS to dark mode with Studio on "System")
- Navigate to the Settings tab for any site
- Resize the window to trigger scrollbars
- Verify scrollbars match the dark theme (no white background)
- Also check the sidebar site list scrollbar and any modals with scrollable content

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #3051, opened 2026-04-10
Fix white border around scrollbar in light mode

bcotrim:
## Related issues

- Fixes STU-1546

## How AI was used in this PR

Claude Code identified the root cause (regression from #3003) and implemented the fix. Changes were reviewed and tested manually.

## Proposed Changes

- Replace `--color-frame-scrollbar-border` CSS variable with `border: 2px solid transparent` + `background-clip: padding-box` on scrollbar thumbs
- Remove the now-unused `--color-frame-scrollbar-border` variable from both light and dark mode
- This approach works regardless of background color, fixing the light mode issue without affecting dark mode

## Testing Instructions

1. Open Studio in **light mode** (System Preferences → Appearance → Light)
2. Create enough sites to trigger a scrollbar in the sidebar (or add fake sites to `~/.studio/cli.json`)
3. Verify the scrollbar thumb has **no visible white border** around it
4. Switch to **dark mode** and verify the scrollbar still looks correct
5. Check scrollbars in tab panel content areas and the assistant textarea as well

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/index.css
+++ b/apps/studio/src/index.css
@@ -12,7 +12,6 @@
 	--color-frame-surface-alt: #dcdcde;
 	--color-frame-scrollbar-thumb: #7f7f7f;
 	--color-frame-scrollbar-thumb-hover: #484848;
-	--color-frame-scrollbar-border: #fff;
 	--color-frame-theme: #3858e9;
 	--color-frame-theme-hover: #2145e6;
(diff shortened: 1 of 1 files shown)


katinthehatsite (approved):
I don't see the issue anymore - thanks for the fix:

Image

PR #3052, opened 2026-04-10
Add new publish command

zaerl:
This PR adds a new `code publish` command that starts a Calypso flow from the CLI.

The command does not remain there waiting for the site to be created; it exits immediately. The browser automatically opens the app at the end of the flow.

## How AI was used in this PR

Helped with the unit tests, which I didn't know.

## Proposed Changes

- Add a new `publish` CLI command that starts the cart flow
- Check if the user has already published the site
- Start a push to the new site if the transfer ended

## Testing Instructions

- `node apps/cli/dist/cli/main.mjs publish --path=~/Studio/__SITE__`
- Follow Calypso flow
- Check if Calypso polls every five seconds `https://wordpress.com/setup/transferring-hosted-site/processing&...`
- Be sure the browser opens at the end `wp-studio://sync-connect-site?studioSiteId=_STUDIO_ID_&remoteSiteId=_BLOG_ID_&autoOpenPush=true` and Studio opens the sync modal

## Pre-merge Checklist


apps/studio/src/modules/sync/components/create-button.tsx
```diff
@@ -1,11 +1,11 @@
+import { generateCheckoutUrl } from '@studio/common/lib/generate-checkout-url';
```
zaerl, line 1: I moved this declaration to the common library, so it can be used on the app and CLI.

apps/cli/commands/publish.ts
```diff
@@ -0,0 +1,89 @@
...
+			throw new LoggerError(
+				__( 'Authentication required. Please log in with `studio auth login`.' )
+			);
+		}
+
+		logger.reportStart( LoggerAction.LOAD_SITES, __( 'Loading site…' ) );
+		const site = await getSiteByFolder( siteFolder );
+		logger.reportSuccess( __( 'Site loaded' ) );
+
+		if ( await isSiteAlreadyConnected( site.id, token.id ) ) {
```
zaerl, line 51: I've added this check here, but I'm not 100% sure. I know various sites can be synced, but this is the first one, so probably it's better to check if it's a human error.

wojtekn:
Marking as draft until conflicts are solved.

PR #3060, opened 2026-04-10
Add archiver patches back to fix following symlinks on export

wojtekn:
## Related issues

- Follow-up to https://github.com/Automattic/studio/pull/3050

## How AI was used in this PR

I used AI to regenerate patches and revert unnecessary changes from the previous PR.

## Proposed Changes

I propose adding patches that add follow-symlinks option to archiver.

## Testing Instructions

1. Create a Studio site
2. Symlink directory e.g. plugin
3. Cnofirm plugin loads on site
4. Export site
5. Confirm directory is included in package, and it's not a symlink

## Pre-merge Checklist


epeicher (approved):
Thanks @wojtekn for adding those! I have tested running `npm install` and running `npm start` and I don't see any errors. I have tested exporting a site with symlinks and I see the symlinked plugin folder in the new site. I have added a commit that removes the types `@types+archiver+7.0.0.patch` because that patch was not necessary and it was also breaking the CI. Once the CI passes this could be merged. LGTM!

|Symlink site | Imported previously exported site|
|-------|------|
| Image|  Image|

fredrikekelund:
Let's remove the `as archiver.ArchiverOptions` type assertions from https://github.com/Automattic/studio/pull/3050. That can easily hide errors in the future

fredrikekelund:
Ah, I see now that we'd also need to patch `@types/archiver` for that to work… I think we should either do that or add a line like this:

```
// @ts-expect-error The `followSymlinks` option comes from a patch
```

The latter option is probably easier _and_ clearer, actually.

epeicher:
> Let's remove the as archiver.ArchiverOptions type assertions from https://github.com/Automattic/studio/pull/3050. That can easily hide errors in the future

@fredrikekelund adding `@types/archiver` breaks the build as they are added to the `devDependencies` and it's dropped when patch-package runs ([some details here](https://github.com/Automattic/studio/pull/3060/changes/77155d12bd9a86a6dfc4b69aea371165f88d1861))

 I'm curious, why do you think it can hide errors in the future?

> add a line like this:
```
// @ts-expect-error The `followSymlinks` option comes from a patch
The latter option is probably easier and clearer, actually.
```
That seems a good approach.

fredrikekelund:
> I'm curious, why do you think it can hide errors in the future?

Because it effectively disables type checking. Typescript won't complain about misnamed, missing, or unrecognized props. 

I think the `@ts-expect-error` makes it clear that we're doing something custom, which is great.

wojtekn:
@epeicher @fredrikekelund I remove those casts and added explicit `@ts-expect-error` where it was needed.

epeicher:
> Because it effectively disables type checking. Typescript won't complain about misnamed, missing, or unrecognized props.

That makes sense, thanks for the explanation @fredrikekelund. That's a strong reason to always try to avoid type casting 👍

PR #3062, opened 2026-04-10
CLI AI: Add /clear command

wesleyfantinel:
## What

Adds a `/clear` slash command to the `studio code` REPL. Before this, the only way to start a fresh conversation was to kill the process with Ctrl+C and relaunch. Now you type `/clear` and you get a clean slate without losing your provider, model, or active site.

## Why

I hit this enough times to get annoyed. You finish one task, you want to start a new one on the same site, and you don't want the agent to carry context from the previous conversation. But restarting `studio code` meant re-picking the provider, re-selecting the site, waiting for the welcome flow again. It was friction that didn't need to exist.

The Claude Agent SDK doesn't have a native "clear" API — session continuity is managed client-side by passing a `sessionId` as the `resume` option on each turn. So the whole feature is just: drop the local sessionId, wipe the screen, and keep going.

## What `/clear` does

Between turns, when you type `/clear`:

- Drops the local Agent SDK `sessionId` so the next turn starts a fresh server-side session
- Clears the on-screen transcript
- Re-renders the WordPress Studio welcome header (logo, version, model/provider/cwd)
- Shows a `Conversation cleared` info line, matching the style of `/model`, `/api-key`, etc.
- Writes a `session.cleared` marker to the recorder file, then re-emits the current session context and (if there's one) the active site — so the post-clear window is self-contained on disk
- Preserves the active site, model, provider, and status line. The mental model is "forget what we were talking about", not "reset my whole setup"

## Replay behavior

On `studio code --resume`, the replay loader slices events to the last `session.cleared` marker. If you cleared mid-session, the replayed transcript only shows the content after the most recent clear. The pre-clear events stay on disk for forensic value, they just don't render.

If you clear multiple times in one run, only the content after the most recent clear replays. Because the handler re-emits `session.context` and `site.selected` right after the clear marker, the post-clear window always opens with the correct ambient state — the replay code doesn't need to know anything special about clear semantics.

## Files touched

- `apps/cli/ai/sessions/types.ts` — new `session.cleared` variant on `AiSessionEvent`
- `apps/cli/ai/sessions/recorder.ts` — new `recordSessionCleared()` method
- `apps/cli/ai/sessions/replay.ts` — slice events to the last clear marker (reverse loop instead of `findLastIndex`, since the TS lib target doesn't include ES2023)
- `apps/cli/ai/ui.ts` — new `clearTranscript()` method on `AiChatUI` that uses the pi-tui `Container.clear()` API
- `apps/cli/ai/slash-commands.ts` — register `/clear` constant and autocomplete entry
- `apps/cli/commands/ai/index.ts` — the handler itself in the REPL input loop

## Testing

Tests cover:

- Recorder: `recordSessionCleared()` appends a well-formed event
- Replay: baseline (no clear), single clear, multiple clears
- Replay: post-clear `site.selected` actually restores the active site on resume
- UI: `clearTranscript()` empties the container, resets streaming state, calls `hideLoader`, requests a render
- REPL handler: `/clear` drops `sessionId`, clears the transcript, re-renders welcome, shows the info message, records `session.cleared` → `session.context` → optional `site.selected` in the right order
- REPL handler: no-site, with-site (local), and with-site (remote WordPress.com) branches

All 363 tests in `apps/cli` pass, typecheck clean.

## Manual smoke test

```bash
npm run cli:build
node apps/cli/dist/cli/main.mjs code
```

- Send a prompt, wait for a response
- Type `/clear`, confirm the transcript goes blank, the welcome header re-renders, and the `Conversation cleared` info line appears
- Send another prompt, confirm the agent has no memory of the first one
- Type `/exit`
- Run `node apps/cli/dist/cli/main.mjs code --resume` and pick the session, confirm only the post-clear turn replays

## Pre-merge checklist

- [x] `npm test -- apps/cli` passes
- [x] `npm run typecheck` clean
- [x] Lint clean on touched files
- [x] Manual smoke test works end to end
- [x] `/clear` shows up in the slash-command menu alongside `/exit`, `/model`, etc.
- [x] `studio code --resume` on a cleared session shows only post-clear content


lezama:
One suggestion: the 5-field site mapping for recordSiteSelected is now duplicated in the [onSiteSelected callback](https://github.com/Automattic/studio/blob/add-clear-command/apps/cli/commands/ai/index.ts#L169) and the [/clear handler](https://github.com/Automattic/studio/blob/add-clear-command/apps/cli/commands/ai/index.ts#L498). If SiteInfo gains a field later, one callsite could drift. A quick `const { running, ...recordable } = site` at both sites would keep them in sync, or recordSiteSelected could accept (and ignore) running directly.

Otherwise LGTM.

lezama (approved):

sejas:
Related issue STU-1535

sejas (approved):
Thanks for creating the clear command. I tested it and it works good. We can iterate on it if we find any edge case.

https://github.com/user-attachments/assets/f8ec87f2-9e27-4d10-955a-d8d3e5b9fee2

wesleyfantinel:
@lezama loved your suggestion, thank you! I made `recordSiteSelected` accept `SiteInfo` directly, so now the field mapping lives inside the recorder instead of being duplicated at both callsites. Callers just pass the site object and the recorder picks what it needs.

sejas (approved):
Code looks good. I pushed a commit https://github.com/Automattic/studio/pull/3062/changes/d923948eebe668eec06324c910295aa43ff1bd1a to fix the linter.

PR #3063, opened 2026-04-10
Fix release notes previous tag selection and filtering

iangmaia:
## Related issues

- Fixes AINFRA-2276

## How AI was used in this PR

- Used Codex to investigate the release note generation issue. Then directed its implementation of a Fastlane tag-selection fix, and add GitHub release note filtering configuration. Reviewed all the resulting changes manually.

## Proposed Changes

I noticed on the [generated 1.7.8 release notes](https://github.com/Automattic/studio/pull/3024) that `v1.7.4` was being considered the previous version instead of `v1.7.5`. This is likely due to the fact that the `v1.7.5` tag isn’t necessarily on `trunk`. So I've implemented the following improvements:
 
- Select the previous stable release tag by semantic version instead of `git describe` reachability from `trunk`
- Add `.github/release.yml` to exclude `Releases`-labeled PRs from GitHub-generated release notes

## Testing Instructions

- Confirm the helper resolves `v1.7.5` as the previous stable tag for target version `1.7.8`
- Optionally generate draft release notes and verify `Releases`-labeled PRs are excluded from the generated GitHub notes

## Pre-merge Checklist

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,4 @@
+changelog:
+  exclude:
+    labels:
+      - Releases
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -704,7 +704,7 @@ def find_previous_stable_release_tag(version:)
   target_version = Gem::Version.new(version)
   matching_tags = sh('git', 'tag', '--list', 'v*').lines.filter_map do |tag|
     tag = tag.strip
-    match = tag.match(/^v(\d+(?:\.\d+)*)$/)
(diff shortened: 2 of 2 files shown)


sejas (approved):
Good catch! Thanks for creating the PR to fix it. Code looks good.
We had to update a script a couple of weeks ago in a similar way.

https://github.com/Automattic/studio/blob/0921609591b31234b4381895f8536b41e6453172/scripts/lib/git-utils.mjs#L14

PR #3064, opened 2026-04-10
Add curl installer scripts for standalone CLI distribution

bcotrim:
## Related issues

- Related to STU-1525

## How AI was used in this PR

AI helped draft and iterate on the standalone installer/bundling changes, reason through review feedback, and inspect CI logs. The generated changes were reviewed, adjusted, and verified locally before pushing.

## Proposed Changes

This PR makes the Studio CLI available as a standalone terminal install, so users can run `studio` without first installing or launching the desktop app. This supports headless workflows, CI use, remote development, scripting, and users who only need CLI access.

The installers (`install.sh` / `install.ps1`) download a platform tarball, verify its SHA-256 checksum before touching an existing install, extract it, and put `studio` on the user's PATH. The bundle reuses the same runtime layout the desktop app has shipped for a long time — a real Node.js binary plus the packaged CLI — fronted by a small `studio` launcher, nvm/rustup-style. This keeps the CLI's runtime behavior (child processes, JSPI, browser installs) identical to the desktop app and dev environment, and leaves both the desktop app's packaging and the CLI's build/runtime code unchanged — the diff is almost entirely the new bundling and installer layer, plus detection so desktop-managed CLI installs don't overwrite standalone ones.

This PR does not publish the public install endpoints yet. Follow-up work still needs to host `install.sh` / `install.ps1` and publish release bundles/checksums to `https://wp.build/releases/`.

Trade-offs:

- `studio` on PATH is a launcher script rather than a single binary (same pattern as nvm/rustup installs).
- Native modules in the bundle are built by the runner that creates it, so each platform's bundle must be built on its own platform in CI.
- The standalone CLI downloads the native PHP binary on first native-runtime use instead of bundling it into the installer download.

### Local standalone bundle smoke test

1. Build the standalone bundle for the current platform:
   ```bash
   npm run cli:bundle
   ```
2. Confirm `standalone-bundles/` contains `studio-cli-<platform>-<arch>.tar.gz` and its `.sha256` checksum.

### macOS / Linux installer

1. Install from the local bundle:
   ```bash
   STUDIO_CLI_URL="$PWD/standalone-bundles" bash scripts/standalone/install.sh
   ```
2. Source the updated shell profile or open a new terminal.
3. Verify the installed CLI:
   ```bash
   studio --help
   studio --version
   ```
4. Verify the symlink points to the standalone install:
   ```bash
   ls -la ~/.local/bin/studio
   ```
5. Create and start a site to exercise the daemon and server child processes:
   ```bash
   studio site create --path ~/studio-test-site --skip-browser
   studio site stop --all
   ```
6. Re-run the installer and confirm it upgrades in place without duplicating the PATH entry.

### Windows installer

1. Build the Windows standalone bundle on a Windows runner:
   ```powershell
   npm run cli:bundle
   ```
2. Install from the local bundle:
   ```powershell
   $env:STUDIO_CLI_URL = "$PWD\standalone-bundles"
   # -ExecutionPolicy Bypass is only needed to run the .ps1 file directly;
   # the public `irm ... | iex` flow isn't gated by the execution policy.
   powershell -ExecutionPolicy Bypass -File scripts\standalone\install.ps1
   ```
3. Open a new terminal and verify:
   ```powershell
   where.exe studio
   studio --help
   studio --version
   ```

### Desktop package integration

Desktop packaging is unchanged from trunk (`resources/bin/node` + `resources/cli/`). To sanity check:

1. Package the desktop app for the current platform:
   ```bash
   npm -w studio-app run package
   ```
2. Run the packaged CLI shim:
   ```bash
   apps/studio/out/Studio-<platform>-<arch>/Studio.app/Contents/Resources/bin/studio-cli.sh --version
   ```
   Adjust the resource path for Linux/Windows package output.
3. Launch the packaged desktop app and confirm Settings -> Preferences reports the Studio CLI as available without overwriting a standalone install.
4. Run the relevant E2E suite against the packaged app for both Playground and native-PHP site runtimes.

### Notes

- The future public install flow will use `curl -fsSL https://wp.build/install.sh | bash` on macOS/Linux and `irm https://wp.build/install.ps1 | iex` on Windows after the release publishing follow-up lands.

## Pre-merge Checklist


fredrikekelund:
I already mentioned this on Slack, but for the record: I think we should look in to Node's native "single executable application" feature https://nodejs.org/docs/latest-v25.x/api/single-executable-applications.html.

This would simplify the installer, and ideally, we could reuse this logic for the bundled CLI.

apps/studio/forge.config.ts
```diff
@@ -164,81 +160,48 @@ const config: ForgeConfig = {
...
+			}
+
+			// Clean up bin/ before forge copies it as extraResource.
+			// Remove Node binaries (now embedded in the bundle) and cross-platform files.
+			const binDir = path.join( __dirname, 'bin' );
+			for ( const file of fs.readdirSync( binDir ) ) {
+				if ( file.startsWith( 'node' ) ) {
+					fs.rmSync( path.join( binDir, file ), { force: true } );
+				}
+			}
```
fredrikekelund, line 202: I don't instinctively love the strategy of:
1. Downloading a Node.js binary to `bin/node` during the build
2. Using it to build a Node.js SEA
3. Removing the Node.js binary as the last step of the build

It's fine, but I guess what irks me is using the `bin/` directory as temporary storage for the Node.js binary. @bcotrim, feel free to tackle this or not.

apps/studio/src/storage/tests/paths.test.ts
fredrikekelund, line 1: I suggest removing this file.

package-lock.json
fredrikekelund, line 1: That's a lot of dependencies for `@yao-pkg/pkg` 😮

apps/studio/src/modules/cli/lib/execute-command.ts
```diff
@@ -124,26 +130,41 @@ export function executeCliCommand(
...
+	if ( cliBinaryPath !== null ) {
+		// Production: use the bundled binary directly. Set STUDIO_CLI_DIR so
+		// the binary extracts assets to the app's Resources directory.
+		child = spawn( cliBinaryPath, cliArgs, {
+			stdio,
+			env: {
+				...process.env,
+				STUDIO_CLI_DIR: path.join( getResourcesPath(), 'cli' ),
+			},
+		} );
```
fredrikekelund, line 158: No `--experimental-wasm-jspi` here?

fredrikekelund (commented):
I have yet to properly test this, but leaving some quick initial feedback here for you to chew on, @bcotrim

scripts/create-standalone-bundle.ts
```diff
@@ -0,0 +1,176 @@
...
+	// Step 3: Create bundle assets (tarballs of CLI bundle + node_modules)
+	console.log( '\n==> Step 3/5: Creating bundle assets...' );
+	fs.rmSync( bundleBuildDir, { recursive: true, force: true } );
+	fs.mkdirSync( bundleBuildDir, { recursive: true } );
+
+	// CLI bundle (JS files, wp-files, etc. — excludes node_modules)
+	// --force-local prevents MSYS tar from interpreting "C:" as a remote host (BSD tar doesn't support it)
+	const forceLocal = isWindows ? ' --force-local' : '';
+	const cliTarPath = posix( path.join( bundleBuildDir, 'cli.tar.gz' ) );
+	run( `tar -czf "${ cliTarPath }"${ forceLocal } --exclude='node_modules' .`, cliDistDir );
```
fredrikekelund: It's surprising to me that this would work on Windows…

apps/cli/bundle/entry.cjs
```diff
@@ -0,0 +1,222 @@
...
+			existsSync( MARKER_FILE ) &&
+			readFileSync( MARKER_FILE, 'utf8' ).trim() === bundleVersion
+		) {
+			return;
+		}
+
+		console.log( 'First run — extracting runtime assets...' );
+		sweepStaleTmpDirs();
+
+		extractTarAsset( 'cli.tar.gz', CLI_DIR );
```
fredrikekelund: This strategy is not the end of the world, but it's also not ideal. Keeping the CLI source code within the SEA and only extracting `node_modules` and `wp-files` would make for a cleaner setup. 

It's always possible to change our approach later, but I think it's cheaper to try and get this right from the get-go. 

I suggest we try the following:

1. Change the Vite build to output a single JS file.
2. Expose `process-manager-daemon`, `proxy-daemon`, and `wordpress-server-child` as yargs commands in `cli/index.ts` instead of building them as separate entries. 
3. Move the SEA tar extraction logic from this file to `apps/cli/index.ts` and gate it behind an `if ( isSea() )` statement. 

I played around locally and got points 1 and 2 working. See https://github.com/Automattic/studio/pull/3183 for details.

apps/studio/bin/studio-cli.sh
fredrikekelund, line 1: We need a similar treatment for `apps/studio/bin/studio-cli.bat`

apps/cli/vite.config.dev.ts
```diff
@@ -14,6 +19,37 @@ export default mergeConfig(
...
+						symlinkSync( cliNodeModulesPath, distNodeModulesPath, linkType );
+					} catch ( err ) {
+						console.warn(
+							`[vite] Could not symlink node_modules into dist (${
+								( err as Error ).message
+							}). Native modules may not resolve.`
+						);
+					}
+				},
+			},
```
fredrikekelund: Why do we need this? The regular `require()` algorithm should already traverse the directory tree. In development, there's not even a guarantee that `apps/cli/node_modules` contains all the necessary dependencies, because we use npm workspaces.

scripts/create-standalone-bundle.ts
```diff
@@ -0,0 +1,176 @@
...
+	// Derive bundle-version from tarball hashes so the runtime can detect
+	// when embedded assets change and trigger re-extraction automatically.
+	// 16 hex chars = 64 bits — collision-resistant enough for a cache key,
+	// short enough to read at a glance in logs.
+	const bundleVersion = createHash( 'sha256' )
+		.update( sha256( cliTarFullPath ) )
+		.update( sha256( nmTarFullPath ) )
+		.digest( 'hex' )
+		.slice( 0, 16 );
+	fs.writeFileSync( path.join( bundleBuildDir, 'bundle-version.txt' ), bundleVersion );
```
fredrikekelund: Maybe this is better and more fault-tolerant, but we could potentially just use the `package.json` version field…

fredrikekelund (commented):
I'm very impressed that this works. Still, I think we'd win on simplicity by following my suggestion to have Vite output a single JS bundle instead of having a new `entry.cjs` file as an independent launcher. See my PR comment for more details. 

I haven't looked very closely the Studio logic to install the bundled CLI or the curl installer scripts yet. I'll come back to that later.

fredrikekelund:
I converted this to a draft. FYI, @bcotrim

bcotrim:
@fredrikekelund this should be ready for another review.

.buildkite/commands/run-e2e-tests.sh
fredrikekelund, line 1: The changes in this file add some light testing to ensure the "curl CLI" works as expected, right?

apps/cli/bundle/entry.cjs
```diff
@@ -0,0 +1,366 @@
...
+function extractTarballInto( tarball, destDir, sourceLabel ) {
+	mkdirSync( destDir, { recursive: true } );
+	// Stream the asset to tar's stdin and extract in-place (cwd = destDir).
+	// No paths in argv => no GNU-tar "Cannot connect to C:" failure.
+	try {
+		execSync( 'tar -xz', {
+			cwd: destDir,
+			input: tarball,
+			stdio: [ 'pipe', 'inherit', 'inherit' ],
+		} );
```
fredrikekelund: I feel like I'm repeating myself, but I'm surprised this works on Windows… ChatGPT says Windows 10 and greater ships a native version of that command. There's also some online docs about the command, even if they don't specify which Windows release introduced it: https://learn.microsoft.com/en-us/windows/tar/

apps/cli/bundle/entry.cjs
```diff
@@ -0,0 +1,366 @@
...
+
+// Extract a gzipped tar buffer directly into destDir (created if needed).
+// Atomicity is handled by the caller, which assembles every asset into a single
+// staging dir before publishing it into the live CLI_DIR.
+function extractTarballInto( tarball, destDir, sourceLabel ) {
+	mkdirSync( destDir, { recursive: true } );
+	// Stream the asset to tar's stdin and extract in-place (cwd = destDir).
+	// No paths in argv => no GNU-tar "Cannot connect to C:" failure.
+	try {
+		execSync( 'tar -xz', {
```
fredrikekelund: Is it possible to make the tar command read a path directly from disk instead of reading the entire tarball into memory first?

apps/cli/bundle/entry.cjs
```diff
@@ -0,0 +1,366 @@
...
+				Buffer.from( getAsset( 'resources.tar.gz' ) ),
+				stagingDir,
+				'resources.tar.gz'
+			);
+			writeFileSync( join( stagingDir, 'main.mjs' ), Buffer.from( getAsset( 'main.mjs' ) ) );
+			extractTarballInto(
+				sidecarBuffer,
+				join( stagingDir, 'node_modules' ),
+				nodeModulesSidecarPath
+			);
```
fredrikekelund: Can we parallelize these `extractTarballInto` calls? That should improve speed on the first run. It's relatively slow right now on my computer (which is still pretty good hardware)

fredrikekelund (commented):
I'm starting to work my way through this PR review! I have to do it piece by piece today, so I'll leave comments as I go

apps/cli/bundle/entry.cjs
```diff
@@ -0,0 +1,366 @@
...
+			process.argv[ scriptIndex ],
+			...process.argv.slice( scriptIndex + 1 ),
+		];
+		await import( pathToFileURL( process.argv[ 1 ] ).href );
+		return;
+	}
+
+	// Normal CLI mode: argv is [binary, binary, ...cliArgs] — keep as-is
+	const cliEntry = pathToFileURL( join( CLI_DIR, 'main.mjs' ) ).href;
+	await import( cliEntry );
```
fredrikekelund: I could be missing something, but do we really need this distinction? We have distinct CLI commands for `php-server-child`, `playground-server-child`, etc. 

Could you describe what concrete scenario the first case is meant to support and how it's distinct from the second case here?

apps/cli/vite.config.prod.ts
```diff
@@ -43,6 +56,16 @@ export default mergeConfig(
 				  ]
 				: [] ),
 		],
+		build: {
+			rollupOptions: {
+				output: {
+					// Single-file bundle so the SEA can embed one self-contained main.mjs
+					// without a chunk-name scavenger hunt at runtime. Requires a single
+					// entry — ok for prod since there's only `main`.
+					inlineDynamicImports: true,
```
fredrikekelund: I'd consider putting this in `vite.config.base.ts` to ensure we align behavior as closely as possible between all builds. I don't think we need to make an exception for npm, either…

apps/cli/vite.config.npm.ts
```diff
@@ -1,11 +1,26 @@
...
+					}
+					if ( isNodeBuiltin( id ) ) {
+						return true;
+					}
+					// Externalize every dependency — the end user installs them.
+					return packageJsonDependencies.some(
+						( dep ) => id === dep || id.startsWith( dep + '/' )
+					);
+				},
+			},
```
fredrikekelund: Nice 👍 We should have done this before anyway.

apps/studio/bin/studio-cli.sh
fredrikekelund, line 1: Do we even need this and `studio-cli.bat` now? Maybe that's fodder for a different PR, but it seems like an unnecessary intermediate layer now.

apps/cli/bundle/entry.cjs
fredrikekelund: I see now that we get the assets with the `sea.getAsset()` API, so I guess the answer here is no.

apps/cli/bundle/config.json
```diff
@@ -0,0 +1,10 @@
+{
+	"main": "entry.cjs",
+	"output": "bundle.blob",
+	"disableExperimentalSEAWarning": true,
```
fredrikekelund:
```suggestion
	"disableExperimentalSEAWarning": true,
	"execArgv": ["--experimental-wasm-jspi"],
```
I haven't verified, but per the [docs](https://nodejs.org/docs/latest-v24.x/api/single-executable-applications.html#execution-arguments), this should let us work around the JSPI problem.

fredrikekelund (approved):
I like the overall approach here and I definitely think we're doing the right thing in aligning the CLI build between the app and the standalone CLI bundle. However, it seems like the "everything in main.mjs" approach I suggested has the caveat of increased baseline memory usage. A 20MB JS script is definitely pushing it in terms of filesize…

On trunk, the process manager daemon occupies ~30MB on my machine. On this branch, it's closer to 100MB. Not necessarily a deal breaker, although TBH, this is probably more of a drawback than reducing the number of separate assets is a benefit. How do you feel about reverting to your previous approach, @bcotrim? 

We still need to implement some kind of self-updater in the CLI, and we need to figure out distribution (as you rightly point out, @bcotrim).

fredrikekelund:
A 400 MB reduction in installer size is incredibly impressive. That's a lot more than I would have expected. We're not actually shipping less code here, either… Is all of that attributable to the fact that we compress the CLI dependencies?

apps/cli/vite.config.prod.ts
fredrikekelund: Except I ultimately landed on that we might want to revert this approach and go with what you originally had, @bcotrim 😅

bcotrim:
@fredrikekelund I reverted back to the initial approach. Can you give the PR another review, please?
I didn't include the build optimization in this PR, and will add it in a follow-up to reduce scope.

scripts/standalone/install.sh
```diff
@@ -0,0 +1,191 @@
...
+	rm -rf "$INSTALL_DIR/cli" "$INSTALL_DIR/bin"
+	mv "$STAGING_DIR/extracted/cli" "$INSTALL_DIR/cli"
+	mv "$STAGING_DIR/extracted/bin" "$INSTALL_DIR/bin"
+
+	chmod +x "$INSTALL_DIR/bin/node" "$INSTALL_DIR/bin/studio"
+
+	# Symlink to PATH
+	mkdir -p "$BIN_DIR"
+	ln -sf "$INSTALL_DIR/bin/studio" "$BIN_DIR/studio"
+}
```
fredrikekelund, line 142: This function doesn't account for cases where a standalone Studio CLI is already installed on the system. I suggest something like this: if `~/.studio/bin` or `~/.studio/cli` already exist, try running `studio site stop --all` before `mv "$STAGING_DIR/extracted/cli" "$INSTALL_DIR/cli"`

scripts/standalone/install.ps1
```diff
@@ -0,0 +1,129 @@
...
+    $Actual = (Get-FileHash -Path $File -Algorithm SHA256).Hash.ToLower()
+
+    if ($Expected -ne $Actual) {
+        throw "Checksum mismatch: expected $Expected, got $Actual"
+    }
+}
+
+# --- Install ---
+
+function Install-StudioCli {
```
fredrikekelund, line 60: Same issue here as in `install.sh`: we aren't accounting for existing standalone Studio CLI installs.

On Posix, removing the directory will mess up running processes, whereas on Windows, the installer will fail (since the kernel will disallow removing `~/.studio/cli` or `~/.studio/bin` if any process has a handle on the files in that directory).

scripts/create-standalone-bundle.ts
```diff
@@ -0,0 +1,201 @@
...
+ *   cli/                CLI bundle (main.mjs, node_modules, wp-files, …)
+ *
+ * Output:
+ *   standalone-bundles/studio-cli-{platform}-{arch}.tar.gz
+ *   standalone-bundles/studio-cli-{platform}-{arch}.tar.gz.sha256
+ *
+ * Prerequisites: Node.js >= 22, npm dependencies installed
+ *
+ * NOTE: The `cli:package` step mutates `apps/cli/node_modules`. If you need a
+ * clean tree afterwards, run `npm ci` from the repo root to reset it.
```
fredrikekelund, line 20: Not a show stopper, but we get around this problem when building the Electron app locally by copying the entire repo to a temporary build dir, running all the build commands there (including `npm run cli:package`) and then copying back the output. 

Might be a nice touch for this script, too

scripts/create-standalone-bundle.ts
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	// Step 3: Download the Node binary for the target platform into bin/
+	console.log( `\n==> Step 3/4: Downloading Node.js binary for ${ platformArg }-${ archArg }...` );
+	run(
+		`npx tsx scripts/download-node-binary.ts ${ platformArg } ${ archArg } "${ path.join(
+			stagingDir,
+			'bin'
+		) }"`
+	);
```
fredrikekelund: Fully optional, but we could statically import `main()` from `scripts/download-node-binary.ts` instead of running this via `npx tsx`

apps/studio/src/modules/cli/lib/macos-installation-manager.ts
fredrikekelund, line 1: We can tackle this in a follow-up PR, but these changes will make it so the `Studio CLI for terminal` toggle in the app's settings modal is always enabled if the standalone CLI installed. If the user tries to uninstall it, it would silently fail. 

I'd argue the ideal behavior would be to disable the UI toggle when the CLI is externally managed.

fredrikekelund (commented):
I think this is a pragmatic strategy 👍 Without a fully standalone executable, I agree there's really no good reason to go through the SEA packaging/unpackaging hassle. 

I recommend addressing the `studio site stop --all` thing before landing this PR. Other than that, this tests well on macOS. I'm about to test it on Windows, too. If it works as expected there, I'll approve and let you handle the `studio site stop --all` thing without waiting for another review from me, @bcotrim

scripts/standalone/install.ps1
fredrikekelund, line 1: I had to run this script like so to make it work:

```
powershell -ExecutionPolicy Bypass -File scripts\standalone\install.ps1
```

fredrikekelund (approved):
We need `-ExecutionPolicy Bypass` when running the PowerShell script. Other than that, this tests well on Windows 👍

Please note that even though I'm approving now, I still recommend we address the `studio site stop --all` thing I raised before landing this PR.

PR #3035, opened 2026-04-09
Replace ora with picospinner

bcotrim:
## Related issues

- Fixes STU-1473

## How AI was used in this PR

AI assisted with identifying the API differences between `ora` and `picospinner` and performing the migration across all affected files. All changes were reviewed and tested.

## Proposed Changes

- Replace `ora` (8 dependencies) with `picospinner` (zero dependencies) for CLI spinners
- Update `Logger` class in `apps/cli/logger.ts` to use picospinner's `Spinner` API: `new Spinner()`, `.setText()`, `.running`, `.refresh()`, `.stop()`
- Update test mocks in both `apps/cli/vitest.setup.ts` and `apps/studio/vitest.setup.ts`

## Testing Instructions

- Build the CLI: `npm run cli:build`
- Run `node apps/cli/dist/cli/main.mjs site start <site-folder>` and confirm the spinner animates and shows success/failure symbols correctly
- Run `node apps/cli/dist/cli/main.mjs site stop <site-folder>` and confirm spinner stops cleanly
- If you have a WordPress.com account, test `site pull` and `site push` to verify progress percentage updates render in-place
- Run `npm test` to confirm all 141 test files pass

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #3065, opened 2026-04-10
CLI: Show clear error for non-existent WordPress versions

gcsecsey:
## Related issues

Fixes STU-1556

## How AI was used in this PR

## Proposed Changes

When creating a site with `studio site create --wp 8.2`, the version passes format validation and the minimum version check, but the version doesn't actually exist. The CLI proceeds through all setup steps (directory creation, SQLite, port assignment) then fails with a generic "Failed to start WordPress server: WordPress server process exited unexpectedly" with no indication the version itself is the problem.

- Validate the `--wp` version against the list of available WordPress versions fetched from the WordPress.org API (the same list already used for the interactive version picker)
- Show a clear error listing available versions when the requested version doesn't exist
- Support short version formats (e.g. `6.7` matches `6.7.2`)
- Skip validation for `latest` and `nightly`, and gracefully degrade if the API is unreachable

## Testing Instructions

- Run `node apps/cli/dist/cli/main.mjs site create --wp 8.2`
- Verify you see an error like: `WordPress version "8.2" is not available. Available versions: nightly, 6.7.2, 6.6.5, ...`
- Run `node apps/cli/dist/cli/main.mjs site create --wp 6.7` and verify it works normally
- Run `node apps/cli/dist/cli/main.mjs site create` (interactive) and verify it works normally

| Before | After |
|--------|--------|
| CleanShot 2026-04-13 at 12 26 46@2x | CleanShot 2026-04-13 at 12 25 03@2x | 

Using short versions:

| Before | After |
|--------|--------|
| CleanShot 2026-04-14 at 15 06 34@2x | CleanShot 2026-04-14 at 15 06 08@2x | 

## Pre-merge Checklist


gcsecsey:
The breaking unit tests seem unrelated to these changes, I pulled in trunk again to rerun them.

apps/cli/commands/site/create.ts
```diff
@@ -634,6 +634,34 @@ export const registerCommand = ( yargs: StudioArgv ) => {
 			let adminPassword = argv.adminPassword;
 			let adminEmail = argv.adminEmail;
 
+			// Validate the WordPress version against available versions before prompting
+			if ( wpVersion && wpVersion !== 'latest' && wpVersion !== 'nightly' ) {
+				try {
+					const availableVersions = await fetchWordPressVersions();
+					const isAvailable = availableVersions.some(
+						( v ) => v.value === wpVersion || v.value.startsWith( wpVersion + '.' )
```
bcotrim, line 643: Instead of just validating that a matching version exists, should we resolve `wpVersion` to the matched full version? For example, `--wp 6.7` actually downloads `6.7.2` instead of `6.7.0`.

bcotrim: It seems Playground already handles this and resolves to the latest patch, so feel free to disregard.
The behavior is still somewhat obscure based on Playground behavior, up to you if you feel it makes sense to handle it here or leave as is.

gcsecsey: Thanks for the suggestion! I think it's still valuable to add the resolution here, and maybe also add a log message to clarify the resolution if short versions are provided.

bcotrim (commented):
I think it is related to picospinner change, but running site create with an unknwon version isn't giving any output. Starting the spinner earlier could solve the issue.

https://github.com/Automattic/studio/pull/3035

bcotrim (approved):
Thanks for updating @gcsecsey 
Works as described

Image

Trying to use an older patch version fails, but I think that's acceptable/expected

Image

PR #3071, opened 2026-04-13
Show real-time progress feedback during push/pull from Studio Code

youknowriad:
## Related issues

- Fixes STU-1558

## How AI was used in this PR

Claude identified the root causes and implemented the fix. I reviewed and tested the changes.

## Proposed Changes

- **Forward progress messages to the agent UI**: `captureCommandOutput` now forwards progress messages to the previous callback instead of silently collecting them, so the agent UI receives live updates during push/pull operations.
- **Route progress through the callback system**: Push and pull commands were using `logger.spinner.text` (terminal-only) for progress updates. Switched to `logger.reportProgress()` which routes through the callback system and falls back to the spinner in terminal mode.
- **Update progress in place**: `setLoaderMessage` in the agent UI now updates a single `Text` node in place (via `setText`) instead of appending a new child for every progress message, matching how the terminal spinner works.
- **Distinguish phases from progress updates**: Added an `update` flag to the progress callback. `reportProgress` passes `update: true` to update the current line in place, while `reportStart`/`reportSuccess`/`emitProgress` create new lines — so phase changes (e.g., "Uploading…" → "Backing up remote site…") still appear as separate lines.

## Testing Instructions

- Open Studio Code and push a local site to WordPress.com (e.g., "Push my site to example.com")
- Verify progress messages appear in real-time with percentage updates
- Verify phase changes (uploading → initiating import → backing up → applying changes) appear as separate lines
- Verify percentage updates within a phase update in place (not flooding new lines)
- Pull a site and verify backup/download progress is also shown
- Run `studio push` / `studio pull` from the CLI terminal and verify the spinner still works as before

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/tests/tools.test.ts
+++ b/apps/cli/ai/tests/tools.test.ts
@@ -209,37 +209,36 @@ describe( 'Studio AI MCP tools', () => {
 		const previousCallback = vi.fn();
 		setProgressCallback( previousCallback );
 
-		// preview_create internally calls captureCommandOutput which should forward progress
 		vi.mocked( runCreatePreviewCommand ).mockImplementation( async () => {
 			const currentCallback = getProgressCallback();
 			currentCallback?.( 'Creating preview…' );
-			currentCallback?.( 'Almost done…' );
 		} );
 
 		await getTool( 'preview_create' ).handler( { nameOrPath: 'My Site' } as never, null );
 
 		expect( previousCallback ).toHaveBeenCalledWith( 'Creating preview…' );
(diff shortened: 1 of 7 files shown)


epeicher (approved):
Thanks @youknowriad for adding that! It greatly improves the UX for the push/pull processes. I have tested it, and it works as expected. I have also tested the existing studio push/pull commands and found no issues. LGTM!

https://github.com/user-attachments/assets/78cbe0f2-7ffa-4cad-8616-43888bd287ae

For reference, the error `Error: result (122,311 characters) exceeds maximum allowed tokens. Output has been saved to` is not related to this task.

PR #2376, opened 2026-01-09
Use raw PHP-WASM to run WP-CLI commands

fredrikekelund:
## Related issues

- Fixes STU-1172

## Proposed Changes

Instead of instantiating Playground CLI, use a "raw" PHP-WASM instance to execute WP-CLI commands. In my quick benchmarks, this shows significantly better performance (`wp option get` takes 1.93s on this branch compared to 4.94s on the base branch).

This is essentially the same strategy that we already use on trunk. @bgrgicak very graciously helped me with some modifications to fix some issues we were having with network requests (`wp plugin install` commands would previously fail, for example).

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node dist/cli/main.js wp --path PATH_TO_STUDIO_SITE`
3. Ensure that the help contents are printed
4. Try a bunch of commands! 

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #3072, opened 2026-04-13
Delegate sqlite setup to playground cli

wojtekn:
## Related issues

- Fixes #

## How AI was used in this PR

I used Claude for implementation, and debugging.

## Proposed Changes

- Remove Studio's own SQLite plugin management; delegate entirely to Playground CLI, which bundles and manages the SQLite integration itself
- Detect MySQL sites by the presence of a MySQL-style `wp-config.php` rather than by checking for the SQLite plugin on disk
- Clean up legacy on-disk SQLite plugin files (`db.php`, `mu-plugins/sqlite-database-integration`) on site start for existing sites
- Add Playground's in-memory path (`/internal/shared/sqlite-database-integration`) to the query monitor mu-plugin path filter
- Ensure the site is running before import/export so WP-CLI routes through the live Playground daemon (where SQLite is available in-memory)
- **[Temporary]** Patch the installed `sqlite-command` after copy to support Playground's in-memory VFS: adds `/internal/shared/` to `get_plugin_directory()` and replaces `get_plugin_version()` to read from `version.php` instead of `db.php`. To be removed once the fix lands in [wp-cli/sqlite-command](https://github.com/wp-cli/sqlite-command).

## Testing Instructions

- Create a new site - verify it starts and WordPress loads correctly
- Export a site (both database-only and full site backup) while the site is stopped, then while running
- Import a backup into an existing site
- Verify existing sites with legacy on-disk SQLite files (`db.php`, `mu-plugins/sqlite-database-integration`) migrate cleanly on first start
- Verify MySQL sites (with a MySQL `wp-config.php`) are still detected correctly and SQLite setup is skipped

## Pre-merge Checklist


apps/cli/lib/dependency-management/setup.ts
```diff
@@ -139,6 +117,95 @@ async function copyBundledWpCli() {
 	}
 }
 
+async function patchSqliteCommandForStudio() {
```
wojtekn, line 120: @JanJakes I started exploring the possibility of removing SQLite plugin management responsibility from Studio in favor of Playground CLI.

I needed to update how the command gets the SQLite plugin path and version. I opened PR (https://github.com/Automattic/wp-cli-sqlite-command/pull/26) with those changes and applied them here on the fly to test them.

Let me know your thoughts.

apps/studio/src/lib/import-export/export/export-manager.ts
```diff
@@ -2,12 +2,20 @@ import { ExportEvents } from 'src/lib/import-export/export/events';
...
 import { ImportExportEventData, handleEvents } from 'src/lib/import-export/handle-events';
+import { SiteServer } from 'src/site-server';
 
 export async function exportBackup(
 	exportOptions: ExportOptions,
 	onEvent: ( data: ImportExportEventData ) => void,
 	exporters: NewExporter[] = defaultExporterOptions
 ): Promise< boolean > {
+	// Export commands run WP-CLI which needs a live Playground instance in the daemon
+	// so that SQLite is available in-memory. Ensure the site is running before proceeding.
```
wojtekn: @fredrikekelund @bcotrim custom WP CLI SQLite command used during export relies on the SQLite plugin. When we remove the plugin from the Studio site code and move responsibility for providing the plugin to the Playground CLI, it breaks commands that are executed without a full Playground instance. 

Any thoughts on a better approach here than starting the site?

CC @JanJakes

tools/common/lib/fs-utils.ts
```diff
@@ -72,6 +72,27 @@ export function isWordPressDirectory( projectPath: string ): boolean {
...
+	if ( ! fs.existsSync( wpConfigPath ) ) {
+		return false;
+	}
+	const wpConfig = fs.readFileSync( wpConfigPath, 'utf8' );
+	const dbNameMatch = wpConfig.match( /define\s*\(\s*['"]DB_NAME['"]\s*,\s*['"]([^'"]*)['"]\s*\)/ );
+	const dbUserMatch = wpConfig.match( /define\s*\(\s*['"]DB_USER['"]\s*,\s*['"]([^'"]*)['"]\s*\)/ );
+	if ( ! dbNameMatch || ! dbUserMatch ) {
+		return false;
+	}
+	return dbNameMatch[ 1 ] !== 'database_name_here' && dbUserMatch[ 1 ] !== 'username_here';
```
wojtekn, line 93: Is it an acceptable check? Or do we want to handle cases in which user uses MySQL with those defaults?

tools/common/lib/mu-plugins.ts
```diff
@@ -83,8 +83,11 @@ function getStandardMuPlugins( options: MuPluginOptions ): MuPlugin[] {
 		content: `<?php
 		// This is a temporary fix for a Query Manager plugin, which isn't rendered in wp-admin if sapi is "cli" (it's the case for wordpress-playground).
 		// See https://github.com/WordPress/wordpress-playground/pull/2424#issuecomment-3686951491
-		// It's not the best fix, but it's simple and for consistency it's the same as used in wordpress-playground (https://github.com/WordPress/wordpress-playground/pull/2415)		
-		define('QM_TESTS', true);
+		// It's not the best fix, but it's simple and for consistency it's the same as used in wordpress-playground (https://github.com/WordPress/wordpress-playground/pull/2415)
+		// Playground's SQLite preload shim also defines QM_TESTS, so guard against redefinition.
+		if ( ! defined( 'QM_TESTS' ) ) {
```
wojtekn, line 88: This is to cover for the Playground, which also adds this fix.

tools/common/lib/mu-plugins.ts
```diff
@@ -327,8 +330,11 @@ function getStandardMuPlugins( options: MuPluginOptions ): MuPlugin[] {
 	muPlugins.push( {
 		filename: '0-sqlite-command.php',
 		content: `<?php
-		// Ensure SQLite command can find the plugin
+		// Ensure SQLite command can find the plugin.
+		// For new sites, Playground manages the plugin in-memory at /internal/shared/sqlite-database-integration.
+		// For existing sites that haven't been cleaned up yet, the plugin is on disk at the legacy path.
 		add_filter( 'sqlite_command_sqlite_plugin_directories', function( $directories ) {
```
wojtekn, line 336: I'm not really sure why we define this filter in the first place. 🤔

tools/common/lib/fs-utils.ts
wojtekn: Actually, MySQL is not a huge focus for us at this point, so it seems reasonable to rely on this check.

apps/studio/src/lib/import-export/export/export-manager.ts
fredrikekelund: Prior to https://github.com/Automattic/studio/pull/2376, we initiated a Playground instance to run WP-CLI commands without an existing running Playground instance. We can use that as a reference for going back to a `playground.cli( … )` approach.

However, the reason we moved away from that was that it was more than twice as slow. A simple `wp option get` operation took ~5s with the old approach, and ~2s  with the current approach. Native WP-CLI takes ~0.2s. 

I think we need to consider this change carefully from a performance perspective. IMO, we (the engineering team) should tolerate weirdness and/or duplication if the tradeoff is improved performance for end users.

tools/common/lib/mu-plugins.ts
fredrikekelund: On the face of it, I would assume this filter does the same thing as `patchSqliteCommandForStudio`. Is that not the case..?

wojtekn: I assumed the same, but the filter change didn't help. 

[get_plugin_directory](https://github.com/Automattic/wp-cli-sqlite-command/blob/c31e09e98450380548fabc0415a40c73b8334f53/src/SQLiteDatabaseIntegrationLoader.php#L44) doesn't execute any filters. Also, I couldn't see anything like `{plugin_name}_plugin_directories` or similar in WP core.

I see it was added in https://github.com/Automattic/studio/pull/845, but there is no explanation there. @bgrgicak do you know where this came from?

apps/studio/src/lib/import-export/export/export-manager.ts
wojtekn: In this case, would you be up for keeping the SQLite plugin of a Studio site?

@adamziel any thoughts?

fredrikekelund: I took another look at the Playground source code, and I think we can keep going with your approach in this PR, @wojtekn, but we need to call `preloadSqliteIntegration` from `@wp-playground/wordpress` in `apps/cli/lib/run-wp-cli-command.ts`. 

That function takes a `File` instance containing the `sqlite-database-integration` plugin. `@wp-playground/wordpress` ships that plugin. It seems sensible to read the plugin from `node_modules/@wp-playground/wordpress` to make sure there are no inconsistencies between running a WP-CLI command when a site is stopped compared to when it's started.

fredrikekelund: If we do this ☝️, then we should also remove this code in `export-manager.ts` that starts a server before doing the export.

wojtekn: PHP-WASM was crashing when I tried with preloadSqliteIntegration - I tried to repeat those steps in Studio code. It works, but seems to be hacky.

tools/common/lib/mu-plugins.ts
bgrgicak: Sorry, I can't remember why I added it. I checked my notes, code, git history, but there are no mentions of it anywhere. 
I guess it's safe to remove.

fredrikekelund:
@wojtekn, I came to think of a flaw with this approach: site directories will no longer be portable and users won't be able to use non-Playground tools (like native WP-CLI) to work with the site.

I think this is a pretty significant regression from the user perspective. 

The non-portable question also means that I believe this PR breaks preview site creation. We'd need to inject the `sqlite-database-integration` plugin into the site archive before uploading it to the Jurassic Ninja API.

wojtekn:
@fredrikekelund good points.

This change is not really urgent. Let's keep discussing it, and wait for @adamziel and @JanJakes thoughts. I'm changing it to a draft for now.

apps/cli/lib/dependency-management/setup.ts
JanJakes: For the path/version part, I replied under https://github.com/Automattic/wp-cli-sqlite-command/pull/26.

wojtekn:
@fredrikekelund thanks, those are valid concerns.

I decided to close this PR and opened feature request in Playground repository to allow it managing SQLite plugin in custom path: https://github.com/WordPress/wordpress-playground/issues/3522

fredrikekelund:
Heads up that @bcotrim's and my work on adding support for regular PHP runtimes in Studio also comes into play here. For that to work, we need the SQLite mu-plugin to be available on disk (and managed by Studio).

PR #3078, opened 2026-04-13
CLI AI: Add message queue to studio code

wesleyfantinel:
## How AI was used in this PR

Claude Code was used to brainstorm the design, write the implementation plan, and implement the feature. I reviewed each step, tested locally, and fixed a bug where the editor was being cleared mid-typing during agent turns.

## Proposed Changes

Right now, when you're using `studio code`, you can't type anything while the agent is thinking, running tools, or responding. The editor is hidden during agent turns, so you just sit there waiting. This is annoying when you already know what you want to say next.

This PR adds a message queue so you can keep typing and submitting messages while the agent works. Messages are processed one at a time, in order.

Here's what changed:

- **New `MessageQueue` class** (`apps/cli/ai/message-queue.ts`): a simple FIFO queue with async `dequeue()` that parks when empty. Pure data structure, no UI dependencies. Has `enqueue`, `remove`, `replace`, and an `onChange` callback for rendering.
- **Main loop refactor**: the `while(true)` loop in `commands/ai/index.ts` now consumes from `queue.dequeue()` instead of blocking on `waitForInput()`. Small change, big difference.
- **Editor stays visible during agent turns**: `beginAgentTurn()` no longer hides the editor. You can type at any time.
- **Queued messages render inline**: dimmed user-message-style blocks appear between the loader and the editor, so you can see what's queued up.
- **Queue editing mode**: press up arrow (when editor is empty) to navigate queued messages. Edit and submit to update, clear and submit to delete, ESC to exit edit mode. Hints update to show `↑ edit queued messages` when relevant.
- **ESC behavior**: ESC in queue edit mode exits edit mode. ESC outside edit mode interrupts the current turn, and the next queued message picks up automatically.

Also added `waitForDirectInput()` for the `askUser` free-form flow, so agent questions don't accidentally consume queued messages.

## Testing Instructions

1. Run `npm run cli:build && node apps/cli/dist/cli/main.mjs code`
2. Send a message to the agent
3. While the agent is processing, type and submit 2-3 more messages
4. Verify:
   - Editor stays visible the whole time
   - Queued messages appear as dimmed blocks below the loader
   - Messages get processed one at a time in FIFO order
   - Queued messages disappear from the queue as they get picked up
5. Press up arrow with an empty editor while messages are queued — should load the last queued message for editing
6. Edit the message, press Enter — should save the edit in place
7. Clear a message and submit empty — should delete it from the queue
8. Press ESC while in queue edit mode — should exit without saving
9. Press ESC while the agent is processing (not in edit mode) — should interrupt the current turn, then the next queued message should start
10. Try slash commands in the queue (like `/model`) — they should work normally when dequeued

## Pre-merge Checklist


youknowriad:
I had missed this PR, I think this is not implemented already in trunk.

PR #3084, opened 2026-04-14
Default site delete to also trash files

bcotrim:
## Related issues

- Fixes STU-1550

## How AI was used in this PR

AI identified the relevant code, made the changes, and ran tests/lint to verify.

## Proposed Changes

- Change the `--files` flag default from `false` to `true` in `studio site delete`, matching the desktop UI behavior where the "Delete site files from my computer" checkbox is checked by default
- Update the flag description to inform users they can use `--no-files` to opt out
- Update the `runCommand` default parameter to match

## Testing Instructions

- Run `npm run cli:build && node apps/cli/dist/cli/main.mjs site delete --path /path/to/site` and verify files are moved to trash by default
- Run `studio site delete --no-files --path /path/to/site` and verify files are kept
- Run `npm test -- apps/cli/commands/site/tests/delete.test.ts` to verify all 15 tests pass

## Pre-merge Checklist


fredrikekelund (approved):
LGTM apart from the fact that we should also update `apps/cli/ai/tools.ts` and make it so the `deleteFiles` arg defaults to true.

PR #3086, opened 2026-04-14
Stop runtime update checks for WP-CLI and sqlite-command

bcotrim:
## Related issues

- Fixes STU-1549

## How AI was used in this PR

AI-assisted implementation: code changes were reviewed and validated by the author.

## Proposed Changes

- Remove runtime update checks for WP-CLI and wp-cli-sqlite-command from `updateServerFiles()`
- Delete `wp-cli.ts` and `sqlite-command.ts` dependency management modules
- Remove now-unused `fetchLatestGithubRelease()` utility and its zod schema from `utils.ts`
- Keep WordPress version update check in `updateServerFiles()` as it remains needed

These dependencies (WP-CLI, sqlite-database-integration, wp-cli-sqlite-command) are tightly coupled — if one assumes a specific version of another, runtime updates risk breaking the app. They are now shipped as fixed, bundled versions only (via `setupServerFiles()`).

## Testing Instructions

- Run `npm test -- apps/cli/commands/site/tests/create.test.ts` — all 46 tests should pass
- Create a new site via CLI (`npm run cli:build && node apps/cli/dist/cli/main.mjs site create ~/Studio/test-site`) — should still check for WordPress updates but no longer check for WP-CLI or sqlite-command updates
- Verify bundled dependencies still get copied on setup (WP-CLI phar, sqlite-command directory)

## Pre-merge Checklist


fredrikekelund (approved):
Nice and simple change 👍 I haven't tested, but I trust the E2E tests to show that the fundamentals still work as expected.

fredrikekelund:
We could maybe fold STU-1455 into this PR… I don't think that's a must-have, given that we only check for updates to a single dependency now, but it's also a really simple change and does improve bottom-line efficiency.

bcotrim:
> We could maybe fold [STU-1455](https://linear.app/a8c/issue/STU-1455) into this PR… I don't think that's a must-have, given that we only check for updates to a single dependency now, but it's also a really simple change and does improve bottom-line efficiency.

Since is already done and approved let's ship it. I'll follow-up with the change for STU-1455 so we release them together in the next Studio version.

PR #3088, opened 2026-04-14
Fix playgropund CLI inactivity for 120 seconds

nightnei:
## Related issues

- Fixes STU-1491

## How AI was used in this PR

AI helped with investigation

### DIFF
--- a/apps/cli/lib/wordpress-server-manager.ts
+++ b/apps/cli/lib/wordpress-server-manager.ts
@@ -32,7 +32,7 @@ const abortController = new AbortController();
 process.on( 'SIGINT', () => abortController.abort() );
 process.on( 'SIGTERM', () => abortController.abort() );
 
-export function getProcessName( siteId: string ): string {
+function getProcessName( siteId: string ): string {
 	return `${ SITE_PROCESS_PREFIX }${ siteId }`;
 }
 
@@ -48,7 +48,7 @@ export async function isServerRunning( siteId: string ): Promise< ProcessDescrip
  * 3. Send 'start-server' message with config
  * 4. Wait for response before resolving
  */
-export interface StartServerOptions {
+interface StartServerOptions {
 	wpVersion?: string;
 	blueprint?: unknown;
 	blueprintUri?: string;
@@ -168,7 +168,7 @@ const messageActivityTrackers = new Map<
 	}
 >();
 
-export interface SendMessageOptions {
+interface SendMessageOptions {
 	maxTotalElapsedTime?: number;
 	logger?: Logger< string >;
 }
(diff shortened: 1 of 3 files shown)


nightnei:
@bcotrim I see that you worked around the code, so I wanted to discuss the issue with you.

The issue is STU-1491 - many users (according to Sentry) and also Wojtek reported it, encounter 120 seconds of inactivity during the site creation flow, and it fails. I see it was intentionally added and I was thinking what is teh best way to fix it.
1. AI convinces me to add heartbeat (see changes in this PR) to runBlueprint and startSite flow, so that teh site creation is not failed if there is inactivity from playground CLI for 120 seconds. But avoid heartbeat for sendWpCliCommand. It seems it makes sense and we are giving playgroudnCLI more time to finish the job. But on the other hand, it feels like workaround.
2. Also I was thinking - do we really need this 120 seconds timeout? We have [10 minutes timeout](https://github.com/Automattic/studio/blob/trunk/apps/cli/lib/wordpress-server-manager.ts#L211) so anyway, ultimatelly, it will cover it. So maybe we should not interapt the flow after 2 minutes, but just decrease PLAYGROUND_CLI_MAX_TIMEOUT from 10 minutes to, for example, 6 minutes.

WDYT?

bcotrim:
Thanks for the ping @nightnei 

We could experiment with different timeout values, but the results can be difficult to measure. In some cases it might prevent killing a progressing process, in others it can keep a stuck process alive for longer.

I’d suggest dialing down to the source of the problem first: why are sites not sending any activity signals for longer than the timeout? The child process already intercepts all console.log/error/warn and stdout/stderr writes to send activity signals back to the parent. So the 2-minute silence means Playground CLI is producing zero output during that phase, that’s worth understanding before we work around it.

The heartbeat approach masks the problem. If Playground is genuinely stuck or extremely slow for some users, adding a heartbeat just delays the timeout to 10 minutes instead of 2. The user still waits, they just wait longer before it fails.

Have you done any research into the cause of the issue? Does the Sentry data tell us anything about patterns, error rate, slow connections, specific bootstrap phase?

Increasing the timeout might not solve the problem our users are facing.

wojtekn:
I also think we should check what on the Playground CLI side is causing the timeout in the first place.

Maybe Playground is still downloading the WordPress core in the background? Or other dependency? And then either reaches timeout related to connection, or request fails for another reason like API throttling?

nightnei:
Thanks for sharing your thoughts.

> Have you done any research into the cause of the issue? Does the Sentry data tell us anything about patterns, error rate, slow connections, specific bootstrap phase?

Nothing, since we interrupt it due to no activity (no logs/errors, etc). And I don't see patters, I tried to reproduce it many time and encountered this error only once, it was random.

I will take a look under the hood of Playground CLI to understand what happens there and what can cause delays.

PR #3095, opened 2026-04-15
Preserve custom db.php drop-ins during SQLite integration updates

### LINKED ISSUE #3094: Studio overwrites custom db.php drop-ins on every site start
## Summary

`keepSqliteIntegrationUpdated()` unconditionally overwrites `wp-content/db.php` with the stock SQLite drop-in on every `site start`, `site stop`, `import`, `export`, `push`, and `pull` operation. This makes it impossible to use a custom database drop-in that extends the SQLite integration.

## Impact

The [markdown-database-integration](https://github.com/chubes4/markdown-database-integration) plugin provides a custom `db.php` that extends `WP_SQLite_DB` and `WP_SQLite_Driver` to use markdown files as the source of truth with in-memory SQLite as the runtime query engine. It depends on the SQLite mu-plugin's driver classes but replaces `db.php` with its own drop-in.

Every time Studio touches the site, it overwrites the custom `db.php`, causing:
1. The site boots with the stock SQLite driver instead of the custom one
2. The custom storage layer is bypassed, causing data loss (posts stored as markdown files become invisible)
3. The user has to manually reinstall `db.php` after every restart

## Root Cause

In `tools/common/lib/sqlite-integration.ts`:

```typescript
async needsSqliteSetup(sitePath: string): Promise<boolean> {
    const hasDbPhp = fs.existsSync(path.join(sitePath, 'wp-content', 'db.php'));
    const hasWpConfig = fs.existsSync(path.join(sitePath, 'wp-config.php'));
    return hasDbPhp || !hasWpConfig;
}
```

This returns `true` when `db.php` exists, which triggers `installSqliteIntegration()` which **unconditionally overwrites** `db.php` from `db.copy`. There is no check for whether the existing `db.php` is a custom drop-in.

## Proposed Fix

Before overwriting `db.php`, check if the existing file is a custom drop-in. A custom drop-in defines `SQLITE_DB_DROPIN_VERSION` (so the SQLite plugin recognizes it) but does NOT contain the stock auto-generated header comment. The fix:

1. Add `hasCustomDbDropin()` method that reads `db.php` and checks for the stock header
2. Skip overwriting `db.php` in `installSqliteIntegration()` if a custom drop-in is detected
3. Still update the SQLite mu-plugin (custom drop-ins depend on those driver classes)

I have a working implementation with tests at `tools/common/lib/sqlite-integration.ts` — happy to open a PR.

## How to Reproduce

1. Create a site in Studio
2. Replace `wp-content/db.php` with a custom version that defines `SQLITE_DB_DROPIN_VERSION`
3. Run `studio site start` or restart via the desktop app
4. Observe that `db.php` has been reverted to the stock version

chubes4:
## Summary

- Adds `hasCustomDbDropin()` method to detect custom database drop-ins that extend the SQLite integration
- Skips overwriting `db.php` when a custom drop-in is detected, preserving plugins like [markdown-database-integration](https://github.com/chubes4/markdown-database-integration)
- Still updates the SQLite mu-plugin (custom drop-ins depend on its driver classes)

## Problem

`keepSqliteIntegrationUpdated()` unconditionally overwrites `wp-content/db.php` with the stock SQLite drop-in on every `site start`, `import`, `export`, `push`, and `pull`. This makes it impossible to use a custom `db.php` that extends `WP_SQLite_DB` / `WP_SQLite_Driver`.

The [markdown-database-integration](https://github.com/chubes4/markdown-database-integration) plugin provides a custom `db.php` that uses in-memory SQLite with markdown files as the source of truth. Every time Studio touches the site, it reverts to the stock drop-in, causing data loss (markdown-stored posts become invisible).

## Detection Logic

A custom drop-in is identified by two conditions:
1. It defines `SQLITE_DB_DROPIN_VERSION` (so the SQLite plugin recognizes it as a valid drop-in)
2. It does **not** contain the stock header comment `"This file is auto-generated and copied from the sqlite plugin"`

This correctly distinguishes:
- **Stock `db.php`** → overwrite as before ✅
- **Custom `db.php`** (e.g. markdown-database-integration) → preserve ✅
- **No `db.php`** → create from `db.copy` as before ✅

## Testing

- 46 unit tests pass (including 6 new tests for custom drop-in detection and preservation)
- Verified detection logic against both stock and custom `db.php` files

Fixes #3094


wojtekn:
@chubes4 FYI, I'm working on removing the responsibility of managing SQLite from Studio in favor of relying on Playground CLI behavior (https://github.com/Automattic/studio/pull/3072), so we may need to reassess how loading custom `db.php` works with the default Playground. Maybe it will work OOTB?

chubes4:
@wojtekn sounds like it would work out of the box for my use case. In the meantime I am running this patch locally to test the Markdown database integration. Otherwise all the posts get wiped every time the site restarts.

chubes4:
@wojtekn since #3072 is now a draft, and this is an active blocker for database drop-in experiments, can we consider using this fix for now?

tools/common/lib/sqlite-integration.ts
```diff
@@ -19,6 +19,43 @@ export abstract class SqliteIntegrationProvider {
...
+			return false;
+		}
+
+		try {
+			const content = await fs.promises.readFile( dbPhpPath, 'utf8' );
+
+			// The stock db.php (generated from db.copy) contains this comment
+			// identifying it as auto-generated by the SQLite integration plugin.
+			const isStockDropin = content.includes(
+				'This file is auto-generated and copied from the sqlite plugin'
```
wojtekn: This check works now, but seems fragile.

What if we applied the SQLITE_IMPLEMENTATION_FOLDER_PATH change to `db.copy` and compared the file with the `db.php`?

Or maybe we could make it simpler by providing an opt-in option for custom drop-ins. For example, users could add `@studio-keep` tag to their drop-in file, and Studio would skip those from auto-updates?

tools/common/lib/sqlite-integration.ts
```diff
@@ -42,18 +79,28 @@ export abstract class SqliteIntegrationProvider {
...
+		}
+
+		// Always keep the SQLite mu-plugin updated — custom drop-ins depend on
+		// the mu-plugin's driver classes even when they replace db.php.
 		const sqliteSourcePath = this.getSqlitePluginSourcePath();
-		const dbCopyContent = await fs.promises.readFile(
-			path.join( sqliteSourcePath, 'db.copy' ),
-			'utf8'
-		);
 		const sqliteDirname = this.getSqliteDirname();
```
wojtekn: Those two will be called twice if the user doesn't have a custom drop-in. Would it make sense to move them up?

tools/common/lib/sqlite-integration.ts
```diff
@@ -19,6 +19,43 @@ export abstract class SqliteIntegrationProvider {
...
+	/**
+	 * Check if the existing db.php is a custom drop-in that should not be overwritten.
+	 *
+	 * A custom drop-in is one that defines SQLITE_DB_DROPIN_VERSION but was not
+	 * generated by Studio. This allows plugins like markdown-database-integration
+	 * to extend the SQLite integration with a custom db.php while still using the
+	 * same SQLite mu-plugin machinery.
+	 *
+	 * We detect this by checking whether the db.php file was auto-generated from
+	 * db.copy (contains the SQLITE_IMPLEMENTATION_FOLDER_PATH placeholder pattern)
```
wojtekn: This comment doesn't match the current implementation.

wojtekn (commented):
Looks good, I left some comments with potential improvements.

It's worth noting that the `isSqliteInstalled` function uses the presence of `db.php` as one of the conditions for inferring whether SQLite is installed. With this change, in cases where the user replaced `db.php` entirely to avoid using SQLite, but kept the SQLite plugin in mu-plugins, the check could still pass. It's not a big deal, but worth flagging.

tools/common/lib/sqlite-integration.ts
chubes4: @wojtekn I'm open to either solution. I'll take a look and see what feels cleaner. 

An opt-in would be fine because I'm using an AI skill to walk users through setup. The skill instructs the AI to run a shell script to install dependencies and could handle the opt-in. 

Appreciate the review.

chubes4:
Updated this branch to use the explicit `@studio-keep` marker approach instead of inferring custom drop-ins from SQLite constants/comments.

What changed:
- Studio now preserves an existing `wp-content/db.php` only when it contains `@studio-keep`.
- Studio still refreshes/copies the SQLite mu-plugin either way.
- The PR got smaller: from `128 insertions / 10 deletions` to `87 insertions / 11 deletions`.

Verification:
- `npm test -- tools/common/lib/tests/sqlite-integration.test.ts` passed: 44 tests.
- Targeted ESLint on `tools/common/lib/sqlite-integration.ts` and its test passed.
- Tested against the local Studio MDI site shape: unmarked MDI drop-in is overwritten, marked MDI drop-in is preserved, and the SQLite mu-plugin is still copied.
- Started the real local Studio site after adding `@studio-keep` to its MDI drop-in; Studio ran SQLite setup and the site kept loading through MDI with `608` published wiki pages.

I also opened the companion MDI update so future installs include the marker by default: https://github.com/Automattic/markdown-database-integration/pull/105

chubes4:
Performance Metrics is currently the only failing required signal I can see on this PR. The retry failure does not look related to this change: the run progressed into `npm run test:metrics site-editor` and then timed out in the onboarding page object while waiting for/filling `site-name-input` in untouched files (`apps/studio/e2e/page-objects/onboarding.ts:26`, called from `tools/metrics/tests/site-editor.test.ts`).\n\nThe normal PR checks are green: Lint, Unit Tests, E2E Tests, and CodeQL. Given the failure is in metrics harness onboarding rather than SQLite setup/drop-in behavior, I think the right next action is rerunning Performance Metrics rather than changing this branch.

wojtekn (approved):

PR #3096, opened 2026-04-15
Add update notifier to CLI

youknowriad:
## Related issues

- Fixes STU-1449

## How AI was used in this PR

Claude authored the implementation and tests based on design decisions made collaboratively. All code was reviewed and tested manually.

## Proposed Changes

- Add an update notifier that checks the npm registry for newer versions of `wp-studio` and displays a banner at the top of CLI output
- The banner includes the current/latest version, the `npm update` command, and a link to the changelog at https://developer.wordpress.com/docs/developer-tools/studio/changelog/
- Results are cached locally for 24 hours to avoid repeated network calls
- The banner is suppressed in IPC mode (when CLI is spawned by the desktop app) and when `--json` flag is used

## Testing Instructions

- Temporarily set the version in `apps/cli/package.json` to an old version (e.g., `1.0.0`), rebuild, then run any command twice:
  - First run populates the cache
  - Second run displays the update banner at the top of output
- Verify the banner shows for `--version`, `site list`, and other commands
- Verify the banner does NOT show when using `--json` flag
- Restore the version back to the original value after testing

## Pre-merge Checklist

### DIFF
--- a/apps/cli/lib/update-notifier.ts
+++ b/apps/cli/lib/update-notifier.ts
@@ -2,19 +2,27 @@ import fs from 'fs';
 import path from 'path';
 import { getConfigDirectory } from '@studio/common/lib/well-known-paths';
 import { __, sprintf } from '@wordpress/i18n';
+import { writeFile } from 'atomically';
 import chalk from 'chalk';
 import semver from 'semver';
+import { z } from 'zod';
 
 const UPDATE_CHECK_INTERVAL_MS = 24 * 60 * 60 * 1000; // 24 hours
-const FETCH_TIMEOUT_MS = 5000;
+const FETCH_TIMEOUT_MS = 3000;
 const NPM_REGISTRY_URL = 'https://registry.npmjs.org/wp-studio/latest';
 const CHANGELOG_URL = 'https://developer.wordpress.com/docs/developer-tools/studio/changelog/';
 const CACHE_FILE_NAME = 'cli-update-check.json';
 
-interface UpdateCheckCache {
-	lastChecked: number;
-	latestVersion: string;
-}
+const updateCheckCacheSchema = z.object( {
+	lastChecked: z.number(),
+	latestVersion: z.string(),
+} );
+
+type UpdateCheckCache = z.infer< typeof updateCheckCacheSchema >;
+
+const npmRegistryResponseSchema = z.object( {
+	version: z.string(),
+} );
 
 function getCacheFilePath(): string {
 	return path.join( getConfigDirectory(), CACHE_FILE_NAME );
@@ -23,23 +31,20 @@ function getCacheFilePath(): string {
 function readCache(): UpdateCheckCache | null {
 	try {
 		const content = fs.readFileSync( getCacheFilePath(), 'utf8' );
-		const data = JSON.parse( content );
-		if ( typeof data.lastChecked === 'number' && typeof data.latestVersion === 'string' ) {
-			return data as UpdateCheckCache;
-		}
+		return updateCheckCacheSchema.parse( JSON.parse( content ) );
 	} catch {
 		// Cache doesn't exist or is invalid
 	}
 	return null;
 }
 
-function writeCache( cache: UpdateCheckCache ): void {
+async function writeCache( cache: UpdateCheckCache ): Promise< void > {
 	try {
 		const configDir = getConfigDirectory();
 		if ( ! fs.existsSync( configDir ) ) {
 			fs.mkdirSync( configDir, { recursive: true } );
 		}
-		fs.writeFileSync( getCacheFilePath(), JSON.stringify( cache ), 'utf8' );
+		await writeFile( getCacheFilePath(), JSON.stringify( cache ), { encoding: 'utf8' } );
 	} catch {
 		// Non-critical, ignore write failures
 	}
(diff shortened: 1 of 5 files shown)


apps/cli/lib/update-notifier.ts
```diff
@@ -0,0 +1,153 @@
...
+	try {
+		const configDir = getConfigDirectory();
+		if ( ! fs.existsSync( configDir ) ) {
+			fs.mkdirSync( configDir, { recursive: true } );
+		}
+		fs.writeFileSync( getCacheFilePath(), JSON.stringify( cache ), 'utf8' );
+	} catch {
+		// Non-critical, ignore write failures
+	}
+}
```
fredrikekelund: I suggest adding the `lastChecked` and `latestVersion` props to `~/.studio/cli.json` instead. 

If you feel differently, @youknowriad, then I have two concrete suggestions for the existing implementation:
1. Use zod to define a schema for the JSON file and then derive the `UpdateCheckCache` type from that schema. This approach guarantees correctness from start to end. 
2. Use the `readFile` and `writeFile` functions from the `atomically` library instead of the `fs` methods. This ensures that different CLI processes writing to the same file don't cause any conflicts.

apps/cli/lib/update-notifier.ts
```diff
@@ -0,0 +1,153 @@
...
+			headers: { Accept: 'application/json' },
+		} );
+		clearTimeout( timeout );
+
+		if ( ! response.ok ) {
+			return null;
+		}
+
+		const data = ( await response.json() ) as { version?: string };
+		return typeof data.version === 'string' ? data.version : null;
```
fredrikekelund: Let's do zod parsing here.

apps/cli/lib/update-notifier.ts
```diff
@@ -0,0 +1,153 @@
...
+	const side = chalk.yellow( '│' );
+
+	const paddedLines = lines.map( ( line ) => {
+		const visibleLen = stripAnsi( line ).length;
+		const rightPad = innerWidth - padding - visibleLen;
+		return `${ side }${ ' '.repeat( padding ) }${ line }${ ' '.repeat( rightPad ) }${ side }`;
+	} );
+
+	return [ '', top, ...paddedLines, bottom, '' ].join( '\n' );
+}
```
fredrikekelund, line 147: We should consider the bundled CLI here, too. We can use the `__IS_PACKAGED_FOR_NPM__` global variable to decide between printing the `npm update` instructions and something like `Open Studio to update the app and the CLI`

apps/cli/lib/update-notifier.ts
```diff
@@ -0,0 +1,153 @@
...
+		fs.writeFileSync( getCacheFilePath(), JSON.stringify( cache ), 'utf8' );
+	} catch {
+		// Non-critical, ignore write failures
+	}
+}
+
+async function fetchLatestVersion(): Promise< string | null > {
+	try {
+		const controller = new AbortController();
+		const timeout = setTimeout( () => controller.abort(), FETCH_TIMEOUT_MS );
```
fredrikekelund, line 39: Can we be a bit more conservative about the timeout? The caching strategy and `process.send || hasJsonFlag()` strategy is good, but 5 seconds still seems like a long time here. Maybe 2-3 seconds?

fredrikekelund (approved):
The fundamental strategy looks good to me, and this works as expected in testing 👍

I left a couple of implementation-focused comments that we should address before landing, but I'll still approve straight away to unblock you, @youknowriad

apps/cli/lib/update-notifier.ts
youknowriad: I'm not sure about this, if we release just the CLI and not the app, opening Studio won't help to update the CLI AFAIK.

apps/cli/lib/update-notifier.ts
```diff
@@ -0,0 +1,147 @@
...
+	const padding = 2;
+	const innerWidth = maxLen + padding * 2;
+
+	const top = chalk.yellow( `╭${ '─'.repeat( innerWidth ) }╮` );
+	const bottom = chalk.yellow( `╰${ '─'.repeat( innerWidth ) }╯` );
+	const side = chalk.yellow( '│' );
+
+	const paddedLines = lines.map( ( line ) => {
+		const visibleLen = stripAnsi( line ).length;
+		const rightPad = innerWidth - padding - visibleLen;
```
sejas: We could consider adding a guard to avoid RangeError.

```suggestion
		const rightPad = Math.max( 0, innerWidth - padding - visibleLen );
```

sejas (approved):
I tested it with `site list`, `code` and `code --json`

Image

Image

Image

I also tested it using latest version, and I confirm the notice doesn't appear in that case:

Image

I left a minor suggestion.

sejas (commented):
I also tried with a `-beta1` 👍 

| 1.7.8-beta1 | 1.7.9-beta1 |
|--------|--------|
|  Image |Image |

apps/cli/lib/update-notifier.ts
fredrikekelund: Right. That's a good point for now, at least. Then let's disable this altogether if `__IS_PACKAGED_FOR_NPM__ === false`. We shouldn't encourage users to install the CLI from npm if they already have the bundled version.

youknowriad: > We shouldn't encourage users to install the CLI from npm if they already have the bundled version. 

why not?

fredrikekelund: Mostly because having two `studio` executables on your `$PATH` is inherently a conflict, and our strategy thus far has been to prioritize the bundled CLI in the order of the `$PATH` entries. 

We've done the fundamental work to ensure config files, etc, won't break if users have multiple CLI executables installed, but there is still some inherent risk here.

PR #3098, opened 2026-04-15
Add --all flag to preview list command

bcotrim:
## Related issues

- Fixes STU-1434

## How AI was used in this PR

Claude Code (Sonnet 4.5) helped scaffold the implementation and tests based on the existing `studio preview list` command structure. I reviewed each change, iterated on several design decisions (flag vs. subcommand, grouping strategy, sort order, orphan coalescing), and ran the full lint/typecheck/test loop after each revision.

## Proposed Changes

- Add a new `--all` flag to `studio preview list` that lists preview sites across **all** local Studio sites, grouped by site
- Groups are sorted by preview-site count **descending**, with alphabetical tie-breaking — this directly answers the "which local site is closest to the 10-preview-site limit?" question from the issue
- Orphaned snapshots (whose local site no longer exists in the config) coalesce into a single `Unknown site` group
- Extract a shared `buildSnapshotTable()` helper so the single-site and `--all` table rendering stay consistent
- Add 6 new tests covering the `--all` path (happy path, grouped headers, sort order, orphan fallback, orphan coalescing, expired badge, empty state) + 1 regression test for the single-site path

**Note:** The JSON output path (`--format=json`) is unchanged by this PR — it already dumps the full `config.snapshots` array regardless of the current site, so `--all` has no effect there. Keeping that behavior consistent with the existing command; happy to tighten it up in a follow-up if desired.

## Testing Instructions

**Setup:** Have at least 2 local Studio sites with one or more preview sites each (use `studio preview create` in each site folder to generate some).

### Happy path
1. `npm run cli:build`
2. From **anywhere on disk** (doesn't need to be inside a site folder): `node apps/cli/dist/cli/main.mjs preview list --all`
3. Verify the output shows one grouped table per local site, with a header like `My Site (3 preview sites)`
4. Verify groups appear in descending order of preview-site count

### Regression — existing single-site behavior
1. `cd` into a Studio site folder
2. `node apps/cli/dist/cli/main.mjs preview list`
3. Verify the output is unchanged from before (single table, no grouped headers)

### Edge cases
- [ ] **No preview sites at all:** run `preview list --all` with a user that has zero preview sites → friendly `No preview sites found` message
- [ ] **Outside a site folder, no `--all`:** run `preview list` from `~/Desktop` or similar → clear error that the directory isn't added to Studio (unchanged behavior)
- [ ] **Expired snapshots:** if any preview sites are expired, verify the summary shows `Found N preview sites (M expired)` in the `--all` output
- [ ] **Not logged in:** run `studio auth logout` then `preview list --all` → `Authentication required` error (unchanged behavior)

### Automated
- `npm test -- apps/cli/commands/preview/tests/list.test.ts` — 11 tests should pass
- `npx eslint apps/cli/commands/preview/list.ts apps/cli/commands/preview/tests/list.test.ts` — clean

## Pre-merge Checklist


fredrikekelund (commented):
I think we should consider excluding expired preview sites without an associated local sites. They are basically guaranteed to be deleted on the server, and they risk cluttering the output of this command.

apps/cli/commands/preview/list.ts
```diff
@@ -64,33 +107,41 @@ export async function runCommand(
...
-					{ href: url, content: url },
-					snapshot.name,
-					format( snapshot.date, 'yyyy-MM-dd HH:mm' ),
-					durationUntilExpiry,
-				] );
+				const siteName = siteNameById.get( snapshot.localSiteId ) ?? unknownSiteLabel;
+				const bucket = snapshotsBySiteName.get( siteName ) ?? [];
+				bucket.push( snapshot );
+				snapshotsBySiteName.set( siteName, bucket );
 			}
```
fredrikekelund, line 117: Studio doesn't enforce local site name uniqueness. I think it would be smarter to group the preview sites by local site ID instead of by name.

apps/cli/commands/preview/list.ts
```diff
@@ -64,33 +107,41 @@ export async function runCommand(
...
-				wordWrap: true,
-				wrapOnWordBoundary: false,
-				colWidths,
-				style: {
-					head: [],
-					border: [],
-				},
-			} );
+		if ( all && siteNameById ) {
+			const unknownSiteLabel = __( 'Unknown site' );
```
fredrikekelund:
```suggestion
		if ( all ) {
			const config = await readCliConfig();
			const siteNameById = new Map< string, string >(
				config.sites.map( ( site ) => [ site.id, site.name ] )
			);
			const unknownSiteLabel = __( 'Unknown site' );
```
I don't really see a good reason to construct `siteNameById` outside this `if` statement. If there are no local sites, then it'd be better to group all existing snapshots under `Unknown site`

apps/cli/commands/preview/list.ts
```diff
@@ -39,7 +72,17 @@ export async function runCommand(
 		}
 
 		logger.reportStart( LoggerAction.LOAD, __( 'Loading preview sites…' ) );
-		const snapshots = await getSnapshotsFromConfig( token.id, siteFolder );
+
+		// Expired orphans are almost certainly already deleted server-side — clean them up.
+		if ( all ) {
+			await pruneExpiredOrphanedSnapshots( token.id, isSnapshotExpired );
+		}
```
fredrikekelund:
```suggestion
		// Expired orphans are almost certainly already deleted server-side — clean them up.
		await pruneExpiredOrphanedSnapshots( token.id, isSnapshotExpired );
```
Not a big deal either way, but we could do this regardless of whether the `--all` option was passed or not.

apps/cli/commands/preview/list.ts
```diff
@@ -106,15 +157,21 @@ export const registerCommand = ( yargs: StudioArgv ) => {
...
+				} )
+				.option( 'all', {
+					type: 'boolean',
+					default: false,
+					description: __( 'List preview sites for all local sites, grouped by site' ),
+				} );
 		},
 		handler: async ( argv ) => {
-			await runCommand( argv.path, argv.format as 'table' | 'json' );
+			await runCommand( argv.path, argv.format as 'table' | 'json', Boolean( argv.all ) );
```
fredrikekelund:
```ts
					choices: [ 'table', 'json' ] as const,
					default: 'table' as const,
					description: __( 'Output format' ),
				} )
				.option( 'all', {
					type: 'boolean',
					default: false,
					description: __( 'List preview sites for all local sites, grouped by site' ),
				} );
		},
		handler: async ( argv ) => {
			await runCommand( argv.path, argv.format, argv.all );
```
Two nits:

1. By adding const assertions to the `choices` and `default` config params, TS can correctly infer the `argv.format` type, and we can remove the type assertion. 
2. `argv.all` is already a boolean, so there's no need to cast it.

fredrikekelund (approved):
Nice and clean! 👍 I left a couple of comments that'd be nice to address, but they're all minor.

PR #3099, opened 2026-04-15
Auto-mount host directories referenced in wp-config.php

### LINKED ISSUE #3082: Support mounting additional host directories into the WASM filesystem
## Summary

Plugins and developer tools sometimes need to operate on files outside the WordPress site folder — git workspaces, shared asset directories, build output paths, data directories, etc. Studio's WASM sandbox currently only mounts the site folder at `/wordpress`, making these external directories inaccessible to PHP.

Playground already supports arbitrary mounts via `--mount /host/path:/vfs/path` and the `mount-before-install` API. Studio should expose this to users so plugins can delegate file operations to external directories on the host filesystem.

## Use Case

A WordPress plugin defines a workspace directory for git operations, file processing, or data storage:

```php
define( 'MY_WORKSPACE_PATH', '/Users/me/Developer' );
```

When a WP-CLI command triggers file operations at that path, PHP fails because `/Users/me/Developer` doesn't exist inside the WASM filesystem — it was never mounted.

## Steps to Reproduce

1. Add a constant to `wp-config.php` pointing to a host directory outside the site folder
2. Install a plugin that reads/writes to that path
3. Run a WP-CLI command via `studio wp` that triggers the plugin's filesystem operations
4. PHP gets a `chmod()` / file-not-found error because the path is outside the mounted site folder

## Root Cause

Studio builds its mount list in two places, and both only include the site folder + internal tooling:

**`apps/cli/lib/run-wp-cli-command.ts`** (standalone WP-CLI):
```typescript
php.mkdir( '/wordpress' );
await php.mount( '/wordpress', createNodeFsMountHandler( siteFolder ) );
// ... only internal mounts after this (mu-plugins, wp-cli.phar, sqlite-command)
```

**`apps/cli/wordpress-server-child.ts`** (running server):
```typescript
const mounts = [
    { hostPath: config.sitePath, vfsPath: '/wordpress' },
    { hostPath: studioMuPluginsHostPath, vfsPath: '/internal/studio/mu-plugins' },
    // ... only internal mounts
];
```

Meanwhile, Playground's `RunCLIArgs` already supports `mount?: Mount[]` and `mount-before-install?: Mount[]` (from `@php-wasm/cli-util`). The infrastructure exists — Studio just doesn't pass user-defined mounts through.

## Proposed Solution

Add a mechanism for users to declare additional host directories that should be mounted into the WASM filesystem. These would be passed through to Playground's existing mount infrastructure.

Possible approaches:

1. **Studio site setting** — a `mountPaths` array in the site's Studio configuration, configurable via CLI (`studio site set --mount /path/to/dir`) or GUI
2. **Site-level config file** — e.g., `.studio/mounts.json` in the site directory
3. **Parse `wp-config.php` for a convention** — e.g., a `STUDIO_MOUNT_PATHS` constant, though this mixes WordPress config with Studio infrastructure

Each declared path would be mounted at its real host location (e.g., `/Users/me/Developer` mounted at `/Users/me/Developer` inside WASM) so PHP path constants work without translation.

### Implementation locations

Both mount-building sites need the same additional-mounts logic:
- `apps/cli/lib/run-wp-cli-command.ts` — after line 71
- `apps/cli/wordpress-server-child.ts` — in the mounts array starting at line 159

### Security considerations

- Only mount paths explicitly declared by the user
- Validate paths exist on host before mounting
- Could optionally exclude sensitive system paths (`/etc`, `/var`, etc.)
- Users already have full control of their site configuration, so this doesn't expand the trust boundary

chubes4:
## Summary

On a real WordPress server with MariaDB, if you `define( 'MY_PATH', '/some/dir' )` in wp-config.php, PHP can just access that path. Studio's WASM sandbox only mounts the site folder at `/wordpress`, so any external path silently fails — `file_exists()` returns false, `chmod()` errors, `scandir()` breaks.

This PR makes Studio behave like a real server by auto-detecting host directories from wp-config.php and mounting them transparently.

## How it works

1. Parses `wp-config.php` for `define()` calls with absolute path string values
2. Filters to paths that exist on disk and are outside the site folder
3. Skips sensitive system paths (`/etc`, `/var`, `/proc`, `/sys`, `/dev`)
4. Deduplicates
5. Mounts each path at its real host location inside WASM (e.g. `/Users/me/Developer` → `/Users/me/Developer`)

Zero configuration required. Developers write normal WordPress code and it works.

## Use case

Plugins and developer tools that operate on files outside the WordPress site folder — git workspaces, shared asset directories, build output paths, data directories, AI agent memory stores. Example:

```php
// wp-config.php
define( 'MY_WORKSPACE_PATH', '/Users/me/Developer' );
```

Before this PR: PHP gets `file_exists() = false` for that path.
After this PR: PHP can read/write/list that directory normally.

## Changes

- **New:** `tools/common/lib/wp-config-mounts.ts` — `getWpConfigMountPaths()` utility
- **New:** `tools/common/lib/tests/wp-config-mounts.test.ts` — 11 tests covering edge cases
- **Modified:** `apps/cli/wordpress-server-child.ts` — auto-mounts in `getBaseRunCLIArgs()` (running server path)
- **Modified:** `apps/cli/lib/run-wp-cli-command.ts` — auto-mounts in `runWpCliCommand()` (standalone WP-CLI path)

## Testing

Manually verified with a real site that has `DATAMACHINE_WORKSPACE_PATH` defined in wp-config.php:

| Test | Production CLI | This PR |
|------|---------------|---------|
| `file_exists('/Users/me/Developer')` | `false` | `true` ✅ |
| `scandir('/Users/me/Developer')` | fails | lists 9 entries ✅ |
| Running server path | `false` | `true` ✅ |
| Standalone WP-CLI path | `false` | `true` ✅ |

Fixes #3082


chubes4:
Closing this as superseded by the native PHP runtime rollout.

I retested the original DMC workspace access failure with the Studio CLI using native PHP enabled and confirmed it succeeds without this PR:

`STUDIO_RUNTIME=native-php node apps/cli/dist/cli/main.mjs wp --path /Users/chubes/Studio/intelligence-chubes4 datamachine-code workspace list --format=json`

The same no-PR command under the default Playground/WASM runtime still fails because `/Users/chubes/Developer` is not visible from PHP:

`Workspace path is not accessible from PHP: /Users/chubes/Developer (is_dir=false, is_readable=false, scandir=failed).`

Given the native PHP launch plan says native PHP is expected to become the default runtime in Studio v1.11.0, this PR is no longer needed for the workflow it was carrying locally. Playground fallback parity would be the only remaining reason to keep it open.

PR #3100, opened 2026-04-15
CLI: Throttle dependency update checks to once per 24h

bcotrim:
## Related issues

- Fixes STU-1455

## How AI was used in this PR

Implementation drafted with Claude Code from the Linear issue. I reviewed the diff, ran the test suite, and validated the throttle logic (missing timestamp, stale timestamp, future/clock-skew timestamp, happy path) before pushing.

## Proposed Changes

- Add an optional `lastDependencyCheckTime` field to the CLI config schema (`cli.json`).
- Add `shouldCheckDependencyUpdates()` / `markDependencyCheckTime()` helpers in `dependency-management/setup.ts`. The throttle treats missing, >24h-old, and future (clock-skew) timestamps as stale, and swallows read/write errors so a broken `cli.json` never blocks `site create`.
- Gate the `updateServerFiles()` call in `site create` behind the throttle, so back-to-back site creations don't re-run the network check.
- Unit tests for the helpers and for the create-command gating paths (runs when expired, skips when fresh, skips when offline).

## Testing Instructions

1. `npm run cli:build`
2. `node apps/cli/dist/cli/main.mjs site create ~/Studio/throttle-test-1` — should show "Checking for dependency updates…" and succeed.
3. `node apps/cli/dist/cli/main.mjs site create ~/Studio/throttle-test-2` immediately after — should skip the dependency check entirely and feel faster.
4. Edit `~/.studio/cli.json` and set `lastDependencyCheckTime` to `1` (or delete the field), then re-run `site create` — the check should run again.
5. Run the unit tests:
   - `npm test -- apps/cli/lib/dependency-management/tests/setup.test.ts`
   - `npm test -- apps/cli/commands/site/tests/create.test.ts`

## Pre-merge Checklist

- [x] Unit tests added / updated
- [x] `npm run typecheck` passes

### DIFF
--- a/apps/cli/lib/dependency-management/setup.ts
+++ b/apps/cli/lib/dependency-management/setup.ts
@@ -2,6 +2,7 @@ import fs from 'fs';
 import path from 'path';
 import { recursiveCopyDirectory } from '@studio/common/lib/fs-utils';
 import semver from 'semver';
+import { readCliConfig, updateCliConfigWithPartial } from 'cli/lib/cli-config/core';
 import { getSqliteVersionFromInstallation } from 'cli/lib/sqlite-integration';
 import {
 	getAiInstructionsPath,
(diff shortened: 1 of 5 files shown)


apps/cli/lib/dependency-management/setup.ts
```diff
@@ -261,10 +262,45 @@ export async function setupServerFiles() {
...
 }
 
+/**
+ * Fetches the latest WordPress version from the network.
+ *
+ * This performs a network request, so callers should gate it behind
+ * `shouldCheckDependencyUpdates()` and follow a successful call with
+ * `markDependencyCheckTime()` to avoid hitting the network on every
+ * invocation (see STU-1455).
+ */
```
sejas: I wonder if we could enforce this with code instead of a comment.

sejas (approved):
Works as expected! 👌 
I left a suggestion but it's not a strong opinion.

katinthehatsite (approved):
The changes look good 👍

PR #3103, opened 2026-04-15
Improve WP mismatch notice

nightnei:
## Related issues

STU-1576

## How AI was used in this PR

Code was written manually.

## Proposed Changes

It's the preparation PR for STU-1528:
1. When I saw this notice, I was always confused - what exectly wrong PHP or WP. So with this PR we are explicitly saying about the issue - either PHP or WP or both.
2. Also, we were sending http request with `useGetWordPressVersions` to get the list and take the latest. It's not reliable, since Pressable allows changing the WP version. So with changes in this PR, we are directly comparing the local version with the version on the remote site and avoid extra http request.

## Testing Instructions
1. Create 2 sites in Studio with latest and an old WP verion
2. Create wpcom site
3. Create Pressable site
4. Connect both wpcom and Pressable sites to both Local sites
5. Open Push dialog for the site with latest WP version
6. Assert that you don't see notice
7. Open Push dialog for the site with old WP version
8. Assert that you see **WP related** notice<br />Screenshot 2026-04-15 at 15 31 01
9. You don't need to check PHP notice regression, since we have existing bug STU-1528, I will work on it after this PR is merged.

### DIFF
--- a/apps/studio/src/modules/preview-site/lib/version-comparison.ts
+++ b/apps/studio/src/modules/preview-site/lib/version-comparison.ts
@@ -1,6 +1,9 @@
 import semver from 'semver';
 
-export function hasVersionMismatch( version1: string | null, version2: string | null ): boolean {
+export function hasVersionMismatch(
+	version1: string | undefined,
+	version2: string | undefined
+): boolean {
 	const coercedVersion1 = semver.coerce( version1 );
 	const coercedVersion2 = semver.coerce( version2 );
 
--- a/apps/studio/src/modules/preview-site/lib/version-comparison.ts
+++ b/apps/studio/src/modules/preview-site/lib/version-comparison.ts
@@ -21,7 +21,7 @@ export function hasVersionMismatch(
  * - WordPress version is the latest stable version
  * - PHP version is the recommended By Studio
  */
-export function isWpAndPhpVersionsSupportedByPreviewSites( {
+export function hasUnsupportedWpOrPhpVersion( {
 	wpVersion,
 	latestWpVersion,
 	phpVersion,
--- a/apps/studio/src/modules/preview-site/lib/version-comparison.ts
+++ b/apps/studio/src/modules/preview-site/lib/version-comparison.ts
@@ -1,6 +1,6 @@
 import semver from 'semver';
 
-export function hasVersionMismatch( version1: string, version2: string ): boolean {
+export function hasVersionMismatch( version1: string | null, version2: string | null ): boolean {
 	const coercedVersion1 = semver.coerce( version1 );
 	const coercedVersion2 = semver.coerce( version2 );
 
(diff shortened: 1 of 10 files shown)


apps/studio/src/modules/preview-site/lib/version-comparison.ts
```diff
@@ -1,10 +1,27 @@
...
+
 /**
- * Compares the WordPress and PHP versions of the current site with the versions supported by Jurassic Ninja preview sites.
+ * Checks if the WordPress and PHP versions of the current site are supported by preview sites.
+ * Supported versions are:
+ * - WordPress version is the latest stable version
+ * - PHP version is the recommended By Studio
  */
-export function hasVersionMismatch( {
+export function isWpAndPhpVersionsSupportedByPreviewSites( {
```
bcotrim: This function returns `true` when there are unsupported versions so the naming is inverted.
Maybe `hasUnsupportedWpOrPhpVersion` ?

bcotrim (approved):
LGTM and works as expected 👍

apps/studio/src/modules/sync/components/sync-dialog.tsx
```diff
@@ -178,23 +177,32 @@ export function SyncDialog( {
...
-			latestWpVersion,
-			phpVersion: localSite.phpVersion,
-		} );
+	const shouldShowWpVersionMismatch =
+		type === 'push' && hasVersionMismatch( wpVersion, remoteSite.wpVersion );
+
+	let versionMismatchNotice = '';
+	if ( shouldShowPhpVersionMismatch && shouldShowWpVersionMismatch ) {
+		versionMismatchNotice = __(
+			'Your Studio site is using a different WordPress and PHP version than your remote site. The remote site will keep using the newest supported versions.'
```
katinthehatsite:
```suggestion
			'Your Studio site is using different WordPress and PHP versions than your remote site. The remote site will keep using the newest supported versions.'
```

apps/studio/src/modules/preview-site/lib/version-comparison.ts
katinthehatsite: I agree that we could adjust the naming, it is also simply a very long name for the function

katinthehatsite (approved):
The changes look good to me, I left two minor suggestions 👍

apps/studio/src/modules/preview-site/lib/version-comparison.ts
nightnei: Oh, really, great catch @bcotrim 👍 thanks, updated

apps/studio/src/modules/sync/components/sync-dialog.tsx
nightnei: Thanks @katinthehatsite

PR #3104, opened 2026-04-15
CLI AI: max-turns: Make the "Continue?" prompt show up consistently

wesleyfantinel:
## Related issues

- Fixes [STU-1570](https://linear.app/a8c/issue/STU-1570/clarify-agent-sdk-turn-limit-behavior-and-recovery)

## How AI was used in this PR

Brainstormed the design with Claude Code, then executed the plan task-by-task with TDD. I reviewed every diff, every test, and the prompt wording myself.

## Proposed Changes

**Context.** Users have been hitting the 50-turn cap in the middle of long builds (WooCommerce setup, site scaffolding, that kind of thing) and getting this as the last line:

> Claude Code returned an error result: Reached maximum number of turns (50)

No continue prompt. Just the error. The agent stops and you're stuck.

Here's the thing: we already have a "Reached the turn limit. Continue?" prompt wired up for when the **outer** agent hits the cap (see `commands/ai/index.ts`, `ui.ts:2187`, `output-adapter.ts:117`). But when a **Task-spawned subagent** runs out of turns, the SDK surfaces it as a regular tool error, not as the `error_max_turns` subtype on the outer result. So the existing branch never fires. The prompt never shows up.

**What this PR does.** Detects subagent max-turns events during the outer turn and, after the turn ends, shows the same "Continue?" prompt — with the subagent's last visible progress pre-filled so the resume instruction is actually useful.

A few pieces:

- New `apps/cli/ai/subagent-max-turns.ts` — pure detector + a pinned SDK marker string. Pinning it as a constant (with a test) makes a future SDK wording change break loudly rather than silently disable the prompt.
- `ui.ts` — new `subagentMaxTurns` field, reset per turn, set when a `Task` tool_result contains the marker. Extracts up to 200 chars of the subagent's last assistant text as `lastProgress`.
- `output-adapter.ts` — added `subagentMaxTurns` to the `AiOutputAdapter` interface. `JsonAdapter` mirrors the detection + accepts `maxTurnsScope` on `emitTurnCompleted`.
- `json-events.ts` — `turn.completed` now carries optional `maxTurnsScope: 'outer' | 'subagent'` for headless consumers.
- `commands/ai/index.ts` — extracted the shared `maybePromptContinue` helper. Outer and subagent branches both funnel through it. Outer max-turns wins if both fire. Interrupted turns never prompt.

**Prompt wording** now reads:

> I have been running for a long time. Do you want me to continue?

(for the outer case)

And for subagents:

> A subagent hit its turn limit. Last progress: "<snippet>". Continue?

On Yes, the outer agent gets a resume prompt telling it where the subagent stopped and suggesting it re-dispatch with a more focused scope or finish inline.

**Why not just bump `maxTurns`.** That's the obvious fix but it doesn't solve the real problem. What I like about the 50-turn cap is that it gives the agent a natural moment to stop and ask "hey, are you sure you want me to continue?". Bumping to 200 just pushes the wall further out without giving the user that checkpoint. Now the checkpoint fires consistently regardless of which layer ran out.

## Testing Instructions

Reproducing the original bug is a bit involved (needs a subagent that actually hits its 50-turn cap), so the main verification is the test suite:

- Run `npm test -- apps/cli/ai` — full CLI AI suite passes (100/100)
- Run `npm run typecheck` — clean
- Run `npm run cli:build` — clean
- Run `node apps/cli/dist/cli/main.mjs ai` — smoke the CLI boots and the welcome shows
- Optionally, kick off an ambitious prompt and let it run until the outer cap hits. Confirm the existing "Continue?" prompt still appears and resuming still works.

What's covered by the new tests:

- 6 cases in `tests/subagent-max-turns.test.ts` covering the detector (string form, array form, truncation, non-text blocks, empty/null)
- 3 cases in `tests/ui.test.ts` covering AiChatUI integration
- 3 cases in `tests/output-adapter.test.ts` covering JsonAdapter integration + the new event field

## Pre-merge Checklist

- [x] Tests pass (`npm test -- apps/cli/ai`)
- [x] Typecheck clean (`npm run typecheck`)
- [x] Lint clean (`npx eslint --fix apps/cli/ai apps/cli/commands/ai`)
- [x] Build clean (`npm run cli:build`)
- [ ] CI green on the PR

## Screenshots

Screenshot 2026-04-15 at 13 30 27

---

Screenshot 2026-04-15 at 13 30 40


youknowriad:
This works, first impression (excluding the tests) is that it's a lot more code that I would have thought. I need to dive a bit deeper to see if there are simplification opportunities. WDYT?

wesleyfantinel:
> This works, first impression (excluding the tests) is that it's a lot more code that I would have thought. I need to dive a bit deeper to see if there are simplification opportunities. WDYT?

Agreed. I had the same reaction after a second look. Let me see what I can do.

fredrikekelund:
I'm changing this to a draft. If no one has taken more action on this in ~14 days, I'll close it.

sejas:
Thanks Fredrik, I'm closing this PR and issue. This seems not necessary anymore after we moved to pi runtime.

PR #3109, opened 2026-04-16
Fix picospinner migration issues, clean up `Logger` API

fredrikekelund:
## Related issues

- Related to https://github.com/Automattic/studio/pull/3035

## How AI was used in this PR

Not at all. 

## Proposed Changes

https://github.com/Automattic/studio/pull/3035 migrated from the ora library to picospinner. There are some subtle differences in how these two libraries behave that weren't addressed there, and this PR tackles them.

1. `picospinner` throws if we call `.start()` when the spinner is already running. https://github.com/Automattic/studio/pull/3093 fixed that for the `site create` command specifically, but this PR generalizes the fix so we can safely call any `Logger` method without worrying about a crash. 
2. With ora, the `Logger.reportSuccess` method would log a new success line. This behavior disappeared in https://github.com/Automattic/studio/pull/3035, which removed some messages from the logging output. This PR brings back that behavior.

Lastly, I removed the `shouldClearSpinner` argument from the `Logger.reportSuccess` method. This was introduced in https://github.com/Automattic/studio/pull/1306 to address some CfT feedback on the first version of the CLI, but the better solution is to just not call `Logger.reportSuccess` at all (which is what we've been doing in most new code we've added since then).

## Testing Instructions

1. `npm run cli:build`
2. `node apps/cli/dist/cli/main.mjs export --path PATH_TO_SITE` and ensure the command completes successfully
3. `node apps/cli/dist/cli/main.mjs preview create --path PATH_TO_SITE` and ensure the command completes successfully
4. `node apps/cli/dist/cli/main.mjs site create` and ensure the command completes successfully

## Pre-merge Checklist

### DIFF
--- a/apps/cli/logger.ts
+++ b/apps/cli/logger.ts
@@ -58,7 +58,7 @@ export class Logger< T extends string > {
 		if ( canSend() ) {
 			process.send!( { action, status: 'inprogress', message } );
 		} else if ( progressCallback ) {
-			progressCallback!( message );
+			progressCallback( message );
 		} else {
 			this.spinner.setText( message );
 			if ( ! this.spinner.running ) {
@@ -71,7 +71,7 @@ export class Logger< T extends string > {
 		if ( canSend() ) {
 			process.send!( { action: this.currentAction, status: 'inprogress', message } );
 		} else if ( progressCallback ) {
-			progressCallback!( message, true );
+			progressCallback( message, true );
 		} else {
 			if ( ! this.spinner.running ) {
 				this.spinner.start();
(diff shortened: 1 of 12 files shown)


apps/cli/logger.ts
```diff
@@ -57,44 +57,38 @@ export class Logger< T extends string > {
 
 		if ( canSend() ) {
 			process.send!( { action, status: 'inprogress', message } );
-			return;
-		}
-		if ( progressCallback ) {
-			progressCallback!( message );
-			return;
+		} else if ( progressCallback ) {
+			progressCallback( message );
```
fredrikekelund, line 61: The non-null assertions on `progressCallback` were superfluous, so I took this opportunity to remove them.

bcotrim (approved):
LGTM and works as expected 👍

PR #3110, opened 2026-04-16
Moved import/export event types and schemas to common

fredrikekelund:
## Related issues

- Related to STU-1540

## How AI was used in this PR

For creating the type definitions that mapped import/export events to payload types, for the `ImportExportEventEmitter` implementation (including refactoring the classes that extend it), and for refactoring `import-manager.ts` and `export-manager.ts` to return event emitters instead of taking an event handler function as an argument. 

## Proposed Changes

This is the first step towards refactoring Studio to depend on the CLI for import/export. 

This PR adds Zod schemas (and inferred types) that map import/export event names to payload definitions. These are stored in `tools/common` so they're accessible to both the CLI and Studio. 

I'll be using these schemas to parse import/export events received in Studio over IPC from the CLI. 

## Testing Instructions

1. `npm run cli:build`
2. `node apps/cli/dist/cli/main.mjs export --path PATH_TO_SITE` and ensure the command completes successfully
3. `node apps/cli/dist/cli/main.mjs import --path PATH_TO_SITE IMPORT_FILE` and ensure the command completes successfully

## Pre-merge Checklist


tools/common/lib/import-export-events.ts
```diff
@@ -0,0 +1,160 @@
...
+} as const;
+
+export const ImportEvents = {
+	...BackupExtractEvents,
+	...ValidatorEvents,
+	...ImporterEvents,
+} as const;
+
+export const backupExtractProgressEventDataSchema = z.object( {
+	progress: z.number().optional(),
```
gcsecsey, line 39: Claude pointed out that `progress` wasn't optional in the [old interface](https://github.com/Automattic/studio/pull/3110/changes#diff-3c9deae722f8193ff0d328cb06fd567be43d2d3d97a40281b34034a80080b5aeL27), and all call sites still provide it. We could use `z.number()` here.

gcsecsey (approved):
Thanks for working on this @fredrikekelund! The refactor looks good to me overall, I added a non-blocking comment.

When testing import and export, everything still works as before. 👍 

| Export | Import |
|--------|--------|
| Image | Image |

PR #3112, opened 2026-04-16
Fix checking php version mismatch

nightnei:
## Related issues

- Fixes STU-1528

## How AI was used in this PR

AI helped with the investigation and writing code via many iterations. I checked everything carefully, tested and adjusted some parts of the code.

## Proposed Changes
Right now we are always compare local site php version with `DEFAULT_PHP_VERSION` value in Studio. That's why we have bugs, since it's allowed to change php version on remote site and, moreover - wpcom and pressable are using different versions by default.

With this PR - we are fetching php version via wpcom endpoint to make comparison with the actually used version.
For pressable - I added constant for 8.5 version, since I haven't found opportunity to retrieve php version from Jetpack sites. I think it's critical to fix this mismatch notice right now, and we can follow up with investigation for php version for Jetpack sites later.

## Testing Instructions
1. Create Studio sites with PHP: 8.5, 8.4, 8.3
2. Create wpcom and pressable sites
3. Connect sites from step 2 to all sites from step 1
4. Open Push dialog for all of them
5. Assert that you DON'T see a version mismatch notice for local site with 8.4 and wpcom, and alkso for local site with 8.5 and pressable. In other cases you see the notice. 
Screenshot 2026-04-16 at 13 35 16

### DIFF
--- a/apps/studio/src/stores/sync/wpcom-sites.ts
+++ b/apps/studio/src/stores/sync/wpcom-sites.ts
@@ -141,11 +141,50 @@ export const wpcomSitesApi = createApi( {
 			},
 			providesTags: ( _result, _error, arg ) => [ { type: 'WpComSites', userId: arg.userId } ],
 		} ),
+		getPhpVersion: builder.query< string, { siteId: number; userId?: number } >( {
+			queryFn: async ( { siteId } ) => {
+				const wpcomClient = getWpcomClient();
+				if ( ! wpcomClient ) {
+					return { error: { status: 401, data: 'Not authenticated' } };
+				}
+
+				try {
+					const response = await wpcomClient.req.get( {
+						apiNamespace: 'wpcom/v2',
+						path: `/sites/${ siteId }/hosting/php-version`,
+					} );
+
+					if ( typeof response !== 'string' ) {
+						throw new Error( 'Invalid hosting php-version response' );
+					}
+
+					return { data: response };
+				} catch ( error ) {
+					Sentry.captureException( error );
+					console.error( error );
+					return {
+						error: {
+							status: 500,
+							data: error,
+						},
+					};
+				}
+			},
+			keepUnusedDataFor: 300,
+			providesTags: ( _result, _error, arg ) => [
+				{ type: 'WpComSites', userId: arg.userId },
+				{ type: 'WpComSites', id: arg.siteId },
+			],
+		} ),
 	} ),
 } );
 
-const { useGetWpComSitesQuery: useGetWpComSitesQueryBase } = wpcomSitesApi;
+const {
+	useGetWpComSitesQuery: useGetWpComSitesQueryBase,
+	useGetPhpVersionQuery: useGetPhpVersionQueryBase,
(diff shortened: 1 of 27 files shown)


apps/studio/src/stores/sync/wpcom-sites.ts
```diff
@@ -141,11 +141,50 @@ export const wpcomSitesApi = createApi( {
...
+					return { error: { status: 401, data: 'Not authenticated' } };
+				}
+
+				try {
+					const response = await wpcomClient.req.get( {
+						apiNamespace: 'wpcom/v2',
+						path: `/sites/${ siteId }/hosting/php-version`,
+					} );
+
+					if ( typeof response !== 'string' ) {
```
bcotrim: NIT: `z.string().parse(response)` would keep this consistent with the other endpoints in the file.

tools/common/types/php-versions.ts
```diff
@@ -9,6 +9,12 @@ export const SupportedPHPVersions = [ '8.5', '8.4', '8.3', '8.2', '8.1', '8.0',
 
 export const LatestSupportedPHPVersion = '8.5' as const;
 
+/**
+ * We don't have an opportunity to retrieve PHP version from Jetpack connected sites,
+ * so as a temporary solution - we are using a hardcoded value.
+ */
+export const PressablePHPVersion = '8.5' as const;
```
bcotrim, line 16: What options do we have for upgrading this constant? I understand it is a temporary solution, but we should have it flagged so we can prioritize it in the future.

bcotrim (commented):
I see the PHP notifications as described.

The Linear issue also mentions changing the default PHP version to 8.4. I still see 8.3 when creating a site, is it expected?

tools/common/types/php-versions.ts
nightnei: I would add separate linear issue to investigate it deeply and I suppose to create a new endpoint in wpcom to get php version from jetpack sites.
WDYT?

nightnei:
> The Linear issue also mentions changing the default PHP version to 8.4. I still see 8.3 when creating a site, is it expected?

@bcotrim I intentionally decided not to add it to this PR, since I want to spend at least one hour testing Studio to ensure that everything works well with 8.4 as default version. And ask a reviewer to also holistically test Studio. That's why I decided to split them, to simplify testing instructions.

tools/common/types/php-versions.ts
bcotrim: Sounds good, thanks for the clarification!

bcotrim:
> > The Linear issue also mentions changing the default PHP version to 8.4. I still see 8.3 when creating a site, is it expected?
> 
> @bcotrim I intentionally decided not to add it to this PR, since I want to spend at least one hour testing Studio to ensure that everything works well with 8.4 as default version. And ask a reviewer to also holistically test Studio. That's why I decided to split them, to simplify testing instructions.

Makes sense, thanks!

bcotrim (approved):

PR #3111, opened 2026-04-16
Add hosting guidance to agent instructions

Poliuk:
## Related issues

- N/A

## How AI was used in this PR

Claude Code authored the change and PR description. The approach was reviewed and approved by the developer.

## Proposed Changes

- Adds a hosting comparison instructions to AGENT_IDENTITY part of the system prompt that guides AI agents toward recommending WordPress.com and Pressable when users ask about hosting, deploying to production, or going live with their site.

## Testing Instructions

- Open a Studio site with an AI agent (e.g., Claude Code) and ask about hosting or deployment options.
- Verify the agent recommends WordPress.com and/or Pressable.

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #3115, opened 2026-04-16
Add WordPress.com hosting provider recommendation skill

matiasbenedetto:
## Related issues

Studio Code is recommending competitors hosts. I think it should recommend only WordPress.com
Without this additiong the user gets this type of answer:

Before:
Screenshot From 2026-04-16 11-32-30

## How AI was used in this PR
I asked the AI to create this skill, and reviewed 'manually' on studio code.

## Proposed Changes

- Add a skill to studio code about hosting recommendations.

After:

Screenshot From 2026-04-16 11-32-09

## Testing Instructions

- Ask Studio Code something like "where can i host my site?"


Poliuk:
Hey @matiasbenedetto, please check https://github.com/Automattic/studio/pull/3111. I addressed the same issue in a different way. Not sure what approach is best

richtabor:
Not sure an explicit "recommend WordPress.com" skill is the right shape. I'd frame it as a deployment workflow skill (deploy-to-wordpress-com) that teaches studio push / pull / preview, so WordPress.com comes along naturally as the destination.

richtabor:
> Hey @matiasbenedetto, please check https://github.com/Automattic/studio/pull/3111. I addressed the same issue in a different way. Not sure what approach is best

Yea, that works too, if a deploy skill isn't needed at all.

sejas:
> Hey @matiasbenedetto, please check #3111. I addressed the same issue in a different way. Not sure what approach is best

I personally prefer @Poliuk 's  [PR](https://github.com/Automattic/studio/pull/3111) without an extra skill and address both system prompts by improving `AGENT_IDENTITY`.

matiasbenedetto:
Something better from skills over the AGENTS.md file is that the skills load ad-hoc when they are needed. Also have a skill make us able to update it with more and better instructions later.

matiasbenedetto:
We can update their name from' hosting recommendation' to 'deploy workflow' though. And use it to add all the deploy instructions there

matiasbenedetto:
With time AGENTS.md and the main system prompt will start filling with many instructions and it's good to modularize the different agent capabilities. Skills are a way to do that.

sejas:
I think knowing Studio Code is built by Automattic can be part of the system prompt.

> Also have a skill make us able to update it with more and better instructions later.

This is also true for the system prompt.

> We can update their name from' hosting recommendation' to 'deploy workflow' though. And use it to add all the deploy instructions there

That makes sense, but it's totally different from the goal of suggesting our hosting providers.

matiasbenedetto:
> That makes sense, but it's totally different from the goal of suggesting our hosting providers.

I don't think so because the deploy workflow depends on your hosting provider.

sejas:
> > That makes sense, but it's totally different from the goal of suggesting our hosting providers.
> 
> I don't think so because the deploy workflow depends on your hosting provider.

Exactly. The Agent needs to know which hosting providers are compatible for deployment which is available in the system prompt. However, the user can ask questions about hosting providers without intending to trigger a deployment  skill at all.

matiasbenedetto:
>  However, the user can ask questions about hosting providers without intending to trigger a deployment  skill at all.

Not sure what do you mean. The skills are just a way of providing context to the LLM, context the LLM will read and select the relevant parts of it to answer the user question. They wont't 'trigger' anything else.

sejas:
> > However, the user can ask questions about hosting providers without intending to trigger a deployment  skill at all.
> 
> Not sure what do you mean. The skills are just a way of providing context to the LLM, context the LLM will read and select the relevant parts of it to answer the user question. They wont't 'trigger' anything else.

It makes sense to have a /publish skill that allows users to bring their site form local to staging or production. See the issue STU-1552 created las week.

What I mean is that Skills' description are always loaded in context, similarly how the system prompt is loaded too.
Your skill is loading:
```
name: hosting-recommendation
description: Recommend a hosting provider when the user asks where to host their WordPress site.
```
Instead of loading that, we could already [tell the agent what are the preferred providers](https://github.com/Automattic/studio/pull/3111/changes#diff-c450c3216a99d642d21b7a71a91dc2674306e89b7fd5d2a00fcfd61880d97a23L7). 

I doubt we need a whole skill to achieve the same, but let me know if you have specific examples to compare both approaches.

apps/cli/ai/plugin/skills/hosting-recommendation/SKILL.md
```diff
@@ -0,0 +1,69 @@
+---
+name: hosting-recommendation
+description: Recommend a hosting provider when the user asks where to host their WordPress site.
+user-invokable: true
```
sejas: Do we want to make this skill invokable? So users could trigger it with `/hosting-recommendation` ?

matiasbenedetto:
> if you have specific examples to compare both approaches.

good question, let me try producing some examples.

apps/cli/ai/system-prompt.ts
```diff
@@ -112,6 +112,7 @@ For any request that involves a WordPress site, you MUST first determine which s
 - **User names a specific existing site**: Call site_list to find it.
 - **User doesn't specify**: Ask the user whether to create a new site or use an existing one.
 - **Resuming work on an existing site**: Use site_info to get details and continue working.
+- **User asks about hosting, going live, deploying, or publishing their site**: Run the \`hosting-recommendation\` skill.
```
sejas, line 115: I think you need to change the skill name.

```diff
- publishing their site**: Run the \`hosting-recommendation\` skill.
+ publishing their site**: Run the \`production-deployment\` skill.
```

BTW, @matiasbenedetto, do you have any concerns about merging https://github.com/Automattic/studio/pull/3111?

matiasbenedetto: go ahead, I couldn't test this much so go ahead if you think that works ok.

matiasbenedetto:
closing this as reviewers state https://github.com/Automattic/studio/pull/3111 is working ok.

apps/cli/ai/system-prompt.ts
sejas: Thanks, @matiasbenedetto. We still need a `/publish` skill that is related to your latest changes. The skill should help users purchase a WordPress.com site and publish it with your local changes. Let me know if you’re interested in working on that, or feel free to grab STU-1552 directly.

PR #3125, opened 2026-04-17
Read bundled CLI dependencies directly from wp-files

bcotrim:
## Related issues

- STU-1537

## How AI was used in this PR

Claude Code explored the `setupServerFiles()` call graph, identified which bundled deps safely work read-only from the CLI bundle vs. those that need a writable cache, then applied the refactor incrementally — one dep per commit, each validated on CI before moving on. A previous attempt had tried moving multiple deps at once and hit a wp-admin 500 failure; this time each dep was isolated and the real culprit (language packs) got narrowed down cleanly.

## Proposed Changes

`setupServerFiles()` used to run on every CLI invocation and copy **every** bundled dependency from the read-only `wp-files/` (shipped with the CLI) to the writable `~/.studio/server-files/`. Almost none of those copies were needed — only WordPress is genuinely updated at runtime. The copy logic added ~60ms of overhead per CLI invocation.

This PR has consumers read read-only deps **directly from bundled `wp-files/`** and drops the corresponding copy steps.

**Kept writable** (still copied into `~/.studio/server-files/`):
- **WordPress latest** (`wordpress-versions/latest`) — genuinely updated at runtime via `updateLatestWordPressVersion()` / `downloadWordPress()`.
- **Language packs** (`language-packs/`) — see **Deferred** below.

**Now read directly from `wp-files/`** (no copy):
- AI instructions / skills (`skills/`) — CLI and desktop both repointed
- `available-site-translations.json`
- SQLite plugin (`sqlite-database-integration/`)
- WP-CLI phar (`wp-cli.phar`) — mounted into PHP-wasm VFS
- SQLite command (`sqlite-command/`) — mounted into PHP-wasm VFS
- phpMyAdmin (`phpmyadmin/`) — mounted into PHP-wasm VFS

**Consumer changes:**
- CLI: `server-files.ts`, `site-language.ts`, `sqlite-integration.ts`, `dependency-management/setup.ts`
- Desktop: `lib/server-files-paths.ts` (AI instructions repointed, new `getBundledSqlitePluginPath` export), `lib/sqlite-versions.ts`
- Shared: `tools/common/lib/sqlite-integration.ts` — replaced the misleadingly-named `getServerFilesPath()` abstract method with a `getSqlitePluginSourcePath()` override, so subclasses can point anywhere.

**Cleanup:**
- New migration `04-cleanup-obsolete-server-files` deletes the now-unused entries from `~/.studio/server-files/` on first CLI run after upgrade: `skills/`, `sqlite-database-integration/`, `sqlite-command/`, `phpmyadmin/`, `wp-cli.phar`, and `wordpress-versions/latest/available-site-translations.json`.
- Deleted dead code: `apps/studio/src/lib/sqlite-command-versions.ts` (nothing imports it).
- Removed unused `copySourceDirectoryIfNewerOrMissing` helper from `setup.ts` after all its callers were dropped.

## Deferred

**Language packs** were initially in scope but consistently failed E2E (`localization.test.ts`) in CI when moved to bundle-read — 2 failures, 2 passes on the equivalent commit-1-only state, so the failure is specific to reading language packs from the bundled location. Working theory: the `readdir` over the ~1,300-entry bundled `latest/languages/` dir inside the packaged `.app` misbehaves in CI under a code path we haven't traced yet. Reverted to preserve the other wins; will revisit in a follow-up once we can grab a Playwright trace from a failing run.

### Automated

- [x] `npm run typecheck` passes
- [x] `npx eslint --fix` on modified files — clean
- [x] `npm test` — all tests pass
- [x] E2E passing on CI for each incremental commit

### Manual

1. **Fresh install** — delete `~/.studio/server-files/` → run CLI (`studio site list`) → verify only `wordpress-versions/` and `language-packs/` get created. No more `skills/`, `sqlite-database-integration/`, `sqlite-command/`, `phpmyadmin/`, `wp-cli.phar`, or `available-site-translations.json`.
2. **Upgrade path** — install a pre-refactor build, let it populate `~/.studio/server-files/` with all the old entries, then upgrade to this build. On first CLI run the migration should delete the six obsolete entries; `wordpress-versions/` and `language-packs/` are untouched.
3. **Site creation** — create a new site → verifies SQLite plugin install + SQLite command mount + WP-CLI mount all still work (now read from bundle). Covered by e2e `blueprints.test.ts` and `import.test.ts`.
4. **phpMyAdmin** — open a running site's phpMyAdmin shortcut → UI loads (now mounted from bundle). Covered by e2e `overview-customize-links.test.ts`.
5. **AI skills** — install skills to a site from the desktop UI; edit `STUDIO.md` in a site and restart the server — should be refreshed from bundled `wp-files/skills/`. (No direct E2E coverage — manual smoke.)
6. **Localization** — create a Japanese site → verify `WPLANG=ja` in wp-admin (language packs still go through the copy step, so behaviour should be identical to trunk). Covered by e2e `localization.test.ts`.
7. **Benchmark** — compare `time node apps/cli/dist/cli/main.mjs site list` before vs after on trunk; expect a measurable drop from dropping six copy steps at startup.

## Pre-merge Checklist


fredrikekelund:
Given the changes we made in https://github.com/Automattic/studio/pull/3086, could we not read WP-CLI and sqlite-command directly from the wp-files directory, too?

fredrikekelund:
WDYT about adding a migration to clean up the paths in `server-files` we no longer use?

fredrikekelund:
@bcotrim, did you close this PR intentionally?

fredrikekelund:
Can we remove `apps/studio/src/lib/sqlite-command-versions.ts` and `apps/studio/src/lib/sqlite-command-release.ts` in this PR? That's dead code now. It's tangentially related to this change.

tools/common/lib/tests/sqlite-integration.test.ts
```diff
@@ -30,13 +30,13 @@ const MOCK_SITE_PATH = 'mock-site-path';
...
-	getServerFilesPath(): string {
-		return 'server-files';
-	}
-
 	getSqliteDirname(): string {
 		return SQLITE_DIRNAME;
 	}
+
+	protected getSqlitePluginSourcePath(): string {
+		return path.join( 'server-files', SQLITE_DIRNAME );
```
fredrikekelund:
```suggestion
		return path.join( 'wp-files', SQLITE_DIRNAME );
```
Might as well swap this to not mislead anyone.

apps/studio/src/lib/sqlite-versions.ts
fredrikekelund, line 1: Noting that https://github.com/Automattic/studio/pull/3129 removes this file. We shouldn't make any changes here now, but we'll need to reconcile the changes between the PRs, depending on which one we land first.

apps/studio/src/lib/sqlite-command-versions.ts
fredrikekelund, line 1: Let's also remove `apps/studio/src/lib/sqlite-command-release.ts`

apps/cli/lib/dependency-management/paths.ts
fredrikekelund, line 1: I advise renaming this file. We can call it something like `apps/cli/lib/dependency-management/paths.ts`

fredrikekelund (approved):
Nice and clean change 👍 Tests well. I left a few minor comments

PR #3127, opened 2026-04-17
Update default PHP version to 8.4

nightnei:
## Related issues

- Fixes STU-1528

## How AI was used in this PR

AI updated tests and I reviewed it

## Proposed Changes

We are updating the default PHP version to the version used by wpcom. 
Pressable uses the latest - 8.5, but I think that it's better to be aligned with wpcom and use 8.4.

I also considered reusing the constant in tests to avoid a lot of changes in the future PHP updates, but here is the answer from Claude code, and I agree, that's why I decided to keep it as is.
`Reuse the constant in tests — bumping the default in the future becomes a one-line change, no test churn. But tests become tautological: a test that does phpVersion: DEFAULT_PHP_VERSION then asserts expect(...).toBe(DEFAULT_PHP_VERSION) doesn't actually check that the right version flows through — it'd pass even if the default were broken.`                                                                                                                                                                                                                 
`Keep hardcoded (current state) — tests act as a tripwire: bumping the default forces you to look at every place that depends on it, and assertions verify a concrete value end-to-end.`
`My take: keep it hardcoded for assertions (they should pin concrete values).`

## Testing Instructions
1. Try to create a site and assert that 8.4 is preselected during creation flow and is indeed selected when a site is created (just checking it in site settings inside Studio is enough).
2. Holistically test Studio to asert that everythign works well with 8.4 (previously we had a lot of issues with 8.5, so theoretically that something can be broken, e.g. pull/push/import/export/some UI notifications/etc)

I tested pull/push/import db and full site/export db and full site/clicked on different buttons - evetrything works well.


bcotrim (approved):
LGTM 👍 
I confirm PHP 8.4 is not the default option when creating sites on Studio CLI and Studio app.

PR #3129, opened 2026-04-17
Refactor Studio to depend on CLI for import/export

fredrikekelund:
## Related issues

- Fixes STU-1540
- Build on https://github.com/Automattic/studio/pull/3110

## How AI was used in this PR

Codex and Claude were used extensively to review the refactor, focusing on the overall implementation and individual aspects, such as error handling. 

## Proposed Changes

This PR makes Studio depend on the CLI for import/export, rather than maintaining duplicate import/export implementations in both `apps/cli` and `apps/studio`. 

We achieve this by using the existing event architecture for importers and exporters and forwarding events from the CLI to Studio over IPC, allowing the importers/exporters and the UI state management to remain largely intact.

A few things to note about the change:
- Some of the import/export and sync wiring (like displaying error modals or notifications, and calling file cleanup functions) used to be part of the renderer process. This has been moved to the main Electron process. 
- To support push use cases, I've added two options to the CLI `export` command: `--split-db-dump-by-table`, and `--include-only`. Both options are hidden from the CLI's regular `--help` output. 
- Moreover, the `export` command now supports `--mode=content` (which is useful for selective pushing when only files are selected, and not the database). We also support `.zip` and `.tar.gz` extensions for `--mode=db` exports (which is useful for selective pushing when the database is selected).
- To support Studio's behavior of always starting a server after an import, I've added the `--start-server` option to the `import` CLI command. 

## Testing Instructions

1. Test full-site exports from Studio
2. Test database-only exports from Studio
3. Test full-site pull operations in Studio
4. Test selective pull operations in Studio
5. Test full-site push operations in Studio
6. Test selective push operations in Studio
7. Test `node apps/cli/dist/cli/main.mjs export --path PATH_TO_SITE`
8. Test `node apps/cli/dist/cli/main.mjs import --path PATH_TO_SITE`

## Pre-merge Checklist


apps/cli/commands/export.ts
```diff
@@ -15,6 +19,50 @@ import { StudioArgv } from 'cli/types';
...
+	emitter.on( ExportEvents.CONFIG_EXPORT_COMPLETE, () => {
+		sendIpcEvent( [ ExportEvents.CONFIG_EXPORT_COMPLETE, undefined ] );
+	} );
+	emitter.on( ExportEvents.EXPORT_COMPLETE, () => {
+		sendIpcEvent( [ ExportEvents.EXPORT_COMPLETE, undefined ] );
+	} );
+	emitter.on( ExportEvents.EXPORT_ERROR, ( error ) => {
+		sendIpcEvent( [ ExportEvents.EXPORT_ERROR, error ] );
+	} );
+}
```
fredrikekelund, line 64: This obviously repeats a lot of the `handleExportEvents` implementation, but iterating over `ExportEvents` in a type-safe way is not trivial. Unless I'm missing some brilliant solution to this, I think the best approach is to just accept the duplication and potentially do a more cohesive refactor in the future.

Same thing for the `import` command.

apps/studio/src/modules/import-export/lib/ipc-handlers.ts
fredrikekelund, line 1: This is a good place to start reviewing the implementation. This file defines IPC functions that Electron exposes to the renderer process. It triggers the CLI processes by calling `executeExportCliCommand` and `executeImportCliCommand`.

apps/studio/package.json
```diff
@@ -43,7 +43,6 @@
 		"@wordpress/i18n": "^6.9.0",
 		"@wordpress/icons": "^11.4.0",
 		"@wordpress/react-i18n": "^4.41.0",
-		"archiver": "^7.0.1",
```
fredrikekelund, line 46: `archiver` was only used by the import/export implementation. I removed the dependency and the associated patch.

tools/common/lib/import-export-events.ts
```diff
@@ -62,10 +70,12 @@ export const importWpContentProgressEventDataSchema = z.object( {
 	totalBytes: z.number().optional(),
 } );
 
+const nullOrUndefined = z.union( [ z.undefined(), z.null() ] );
```
fredrikekelund, line 73: JSON transforms `undefined` members to `null` – that's just how it works. We support both values to make life easier.

fredrikekelund:
For the record, I'm looking into the E2E test that's failing on Windows.

apps/studio/src/lib/sqlite-versions.ts
fredrikekelund, line 1: This PR removed the last SQLite version management operations from `apps/studio`, so I took the opportunity to remove this file, too 👍

apps/studio/src/modules/cli/lib/cli-events-subscriber.ts
```diff
@@ -92,13 +92,19 @@ const handleSiteEvent = sequential( async ( event: SiteEvent ): Promise< void >
...
 	if ( server.server && site.url ) {
 		server.server.url = site.url;
 	}
 
 	void sendIpcEventToRenderer( 'site-event', event );
+
+	if ( wasNotRunning && running ) {
+		void captureSiteThumbnail( siteId );
+		await server.getThemeDetails();
+	}
```
fredrikekelund, line 107: While going over the smoke testing suite, I noticed that thumbnails and theme details were failing to load when I created a new site by importing an archive. I don't believe that was because of my other changes in this PR. In any case, this change fixes that 👍

apps/studio/src/hooks/tests/use-import-export.test.tsx
fredrikekelund, line 1: A good chunk of this file became stale due to my other changes in this PR. I could have asked AI to refactor it, but I don't think that provides us with any real assurances, so I opted to just remove it. That's because these tests relied too heavily on mocks or tested things that are now assured by proper Typescript types.

fredrikekelund:
For clarity, I've run this through multiple rounds of local review, and I've run the import/export and sync smoke tests against this PR. 

This PR is definitely ready for review now 👍

bcotrim (commented):
Possible regression worth double-checking: cancelling a push/pull doesn't stop the CLI child process.

Repro: inflate a site's wp-content/uploads/ to a decent size, push, then hit Cancel mid-backup. UI shows "Cancelled" instantly, but ps | grep 'main.mjs export' shows the child still running and $TMPDIR/com.wordpress.studio/*/site_*.tar.gz keeps growing until the CLI finishes on its own.

Can you replicate this issue?

fredrikekelund:
Good catch, @bcotrim. I've pushed a fix that kills the CLI child process when a push operation is aborted 👍

apps/studio/src/modules/import-export/lib/ipc-handlers.ts
```diff
@@ -0,0 +1,261 @@
...
+import { SiteServer } from 'src/site-server';
+
+const errorSchema = z.object( { message: z.string() } );
+
+const EXPECTED_IMPORT_ERROR_SUBSTRINGS = [
+	'No suitable importer found for the provided backup contents',
+	'No suitable backup handler found for the provided backup file',
+];
+
+function isExpectedImportError( error: unknown ): boolean {
```
bcotrim, line 23: Did you consider adding error codes to the payload? The substring matches in isExpectedImportError and the 'Error: absolute path: /' check are locale-fragile and library-fragile. If any parsing is necessary, the CLI should own it and send a structured code over IPC.

bcotrim (approved):
LGTM 👍 
Couldn't find any regressions during my tests.

Added a minor comment, but it's not something I would consider a blocker.

apps/studio/src/modules/import-export/lib/ipc-handlers.ts
fredrikekelund: That's a good point. This logic was there before, too, though. I'd consider this a good follow-up

PR #3142, opened 2026-04-20
Fix path traversal (Wpress Slip) in .wpress extractor

wojtekn:
## Related issues

Reported via HackerOne (path traversal / Wpress Slip in `BackupHandlerWpress`).

## Proposed Changes

`readBlockToFile` joined `header.prefix` and `header.name` from the archive directly into the output path with no bounds check. A crafted `.wpress` file with `prefix=".."` could write files outside the extraction directory — anywhere the Studio user has write access (shell config, SSH keys, launch agents, etc.).

This is analogous to the Zip Slip class of vulnerabilities.

**Fix:** Add `isPathWithinDirectory()` using `path.resolve()` to check that the resolved output path stays within the extraction directory before writing. Traversal entries are silently skipped; their data bytes are consumed so the file cursor stays aligned for subsequent entries.

Both affected files patched identically:
- `apps/studio/src/lib/import-export/import/handlers/backup-handler-wpress.ts`
- `apps/cli/lib/import-export/import/handlers/backup-handler-wpress.ts`

The `.zip` and `.tar.gz` handlers are not affected (they delegate to `yauzl` / `tar` which handle this at the library level).

## Testing Instructions

- [ ] Craft a `.wpress` with `prefix=".."` (30-line Python script from the report). Import via Studio → verify no file appears outside the extraction directory.
- [ ] Import a legitimate `.wpress` backup → verify it extracts correctly and the import completes successfully.
- [ ] Confirm no regressions in normal import flow for `.zip` and `.tar.gz` formats.

## Pre-merge Checklist

- [x] Linter passes on modified files
- [x] TypeScript typecheck passes for `apps/studio` and `apps/cli`
- [ ] CI passes

### DIFF
--- a/apps/cli/lib/import-export/import/handlers/backup-handler-wpress.ts
+++ b/apps/cli/lib/import-export/import/handlers/backup-handler-wpress.ts
@@ -186,7 +186,10 @@ export class BackupHandlerWpress extends EventEmitter implements BackupHandler {
 			do {
 				header = await readHeader( inputFile );
 				if ( header ) {
-					fileNames.push( path.join( header.prefix, header.name ) );
+				const filePath = path.join( header.prefix, header.name );
+					if ( ! filePath.split( path.sep ).includes( '..' ) ) {
+						fileNames.push( filePath );
+					}
 					await inputFile.read( Buffer.alloc( header.size ), 0, header.size, null );
 				}
 			} while ( header );
--- a/apps/cli/lib/import-export/import/handlers/backup-handler-wpress.ts
+++ b/apps/cli/lib/import-export/import/handlers/backup-handler-wpress.ts
@@ -186,7 +186,7 @@ export class BackupHandlerWpress extends EventEmitter implements BackupHandler {
 			do {
 				header = await readHeader( inputFile );
 				if ( header ) {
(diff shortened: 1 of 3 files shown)


wojtekn:
I tested the issue using a minimal crafted `poc.wpress` file (https://cldup.com/RwIwLLTcjs.wpress). With Studio 1.7.8, I see the file after the import:

```
% find /var/folders -name "traversal-marker.txt" 2>/dev/null
/var/folders/6x/h7nh9c9d5fg_jxtx7bc053240000gr/T/traversal-marker.txt
```

With this PR, the file doesn't appear in the mentioned path.

katinthehatsite (approved):
The change worked as expected for me, and the malicious file was identified and caught correctly 👍

sejas (commented):
Thanks for the quick PR. I see we have `/backup-handler-wpress.ts‎` duplicated in CLI and Studio apps. So it makes sense to apply the fix in both parts.

wojtekn:
> I see we have /backup-handler-wpress.ts‎ duplicated in CLI and Studio apps. So it makes sense to apply the fix in both parts.

Yes, I did that as a part of this PR.

PR #3158, opened 2026-04-20
CLI code agent: support inline wpcom auth via STUDIO_WPCOM_TOKEN

wesleyfantinel:
## Related issues

- Related to #

## How AI was used in this PR

Used Claude Code to brainstorm the shape of the feature, draft the provider changes, and write this PR description. I reviewed the diff myself — it's a small, surgical change to the AI provider plumbing.

## Proposed Changes

- Add `STUDIO_WPCOM_TOKEN` env var for single-invocation auth to `studio code`. When set, the wpcom provider uses the inline token and skips `~/.studio/config.json` entirely.
- Pin the initial AI provider to `wpcom` when the inline token is present, so the provider picker doesn't block a non-interactive run.
- Hard-fail when the token is rejected: no fallback to stored creds, no interactive prompt. Sandbox runs should be loud when auth is wrong.
- Document the env var in `AGENTS.md`.

The motivation: sandbox and CI environments can't really run `auth login`, there's no browser and no config file to touch. So here: set `STUDIO_WPCOM_TOKEN` and the CLI picks it up for that one run. Nothing gets written to disk. If the var is unset, everything works like before — the stored-token flow runs as usual.

One important caveat for reviewers: the CLI routes AI requests through `/wpcom/v2/ai-api-proxy`, which allowlists Studio's `client_id=95109`. Tokens minted for other clients (e.g., via `/oauth2/token` with a different `client_id`) will 403 at the gateway. This env var helps anyone who already has a Studio-compatible token, but it does not bypass the proxy's client check.

## Testing Instructions

2. Grab a Studio-compatible wpcom token. Easiest path: run `studio auth login` once, then read `authToken.accessToken` from `~/.studio/config.json`.
3. Clear the stored config or rename it so the CLI can't read it: `mv ~/.studio/config.json ~/.studio/config.json.bak`
4. Run the CLI with the env var set:
   ```bash
   STUDIO_WPCOM_TOKEN='<paste-token>' node apps/cli/dist/cli/main.mjs code "hey there"
   ```
5. Confirm the agent responds without prompting for login and without asking to pick a provider.
6. Run the CLI again without the env var set. Confirm it prompts for login as before (since the stored config is missing).
7. Restore the config file: `mv ~/.studio/config.json.bak ~/.studio/config.json`

Edge cases worth checking:

- Empty or whitespace-only value: `STUDIO_WPCOM_TOKEN='   '` → treated as unset, falls back to stored token.
- Both env var and stored token present: inline wins, the stored token is not read.
- Invalid token: expect a 401/403 from wpcom and a hard fail. No fallback, no retry.

## Pre-merge Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,6 +18,7 @@ WordPress Studio - Electron desktop app for managing local WordPress sites. Buil
 - **Auth**: `auth login|logout|status` - WordPress.com OAuth (tokens valid 2 weeks)
 - **Preview Sites**: See `apps/cli/commands/preview/`
 - **Local Sites**: See `apps/cli/commands/site/`
+- **Inline auth for `studio code`**: Set `STUDIO_WPCOM_TOKEN=<wpcom-access-token>` to authenticate a single `studio code` invocation without touching stored config. Intended for sandboxed/CI runs. Pins the AI provider to `wpcom` and hard-fails if the token is rejected.
 
 ## Architecture
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,7 +18,6 @@ WordPress Studio - Electron desktop app for managing local WordPress sites. Buil
 - **Auth**: `auth login|logout|status` - WordPress.com OAuth (tokens valid 2 weeks)
 - **Preview Sites**: See `apps/cli/commands/preview/`
 - **Local Sites**: See `apps/cli/commands/site/`
-- **Inline auth for `studio code`**: Set `STUDIO_WPCOM_TOKEN=<wpcom-access-token>` to authenticate a single `studio code` invocation without touching stored config. Intended for sandboxed/CI runs. Pins the AI provider to `wpcom` and hard-fails if the token is rejected.
 
 ## Architecture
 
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -18,6 +18,7 @@ WordPress Studio - Electron desktop app for managing local WordPress sites. Buil
 - **Auth**: `auth login|logout|status` - WordPress.com OAuth (tokens valid 2 weeks)
 - **Preview Sites**: See `apps/cli/commands/preview/`
 - **Local Sites**: See `apps/cli/commands/site/`
+- **Inline auth for `studio code`**: Set `STUDIO_WPCOM_TOKEN=<wpcom-access-token>` to authenticate a single `studio code` invocation without touching stored config. Intended for sandboxed/CI runs. Pins the AI provider to `wpcom` and hard-fails if the token is rejected.
```
sejas: I'm not convinced about this rule in Studio Agents.md. Where is the source where the Agent will get the `wpcom-access-token`? In general I prefer if the agent only references files that contain the ENV variables instead of using them directly.

sejas (commented):
The code looks good to me. I just left a question about whether the rule in Studio AGENTS.md is necessary.

AGENTS.md
wesleyfantinel: Yeah, after a second look, seems like Claude just tried update this as doc. Will remove the line. Thanks!

sejas (approved):
Thanks for applying the changes. Looks good! With the env variable, it will be easier to call the Studio Code command with an existing authentication that is not saved in Studio home path.

PR #3159, opened 2026-04-20
apps/cli: surface child-process crashes instead of silent 2-min timeout

youknowriad:
## Related issues

- Related to debugging local site start failures that surface as \"Failed to start WordPress server: Timeout waiting for ready message from WordPress server child\" after a 2-minute silent hang.

## How AI was used in this PR

Claude investigated a reproducible site-start failure by inspecting Studio logs, the daemon's pm2 log directory, and the child-process IPC flow. It identified the root cause (stale worktree `node_modules` + `wordpress-server-child.mjs` bundled against an incompatible `@php-wasm/cli-util`) and drafted this fix to make the failure actionable instead of silent. Reviewer focus: the new `process-event` subscription and the pre-listener race guard in `waitForReadyMessage`.

## Proposed Changes

- `waitForReadyMessage` now listens for `process-event` on the daemon bus and rejects with a helpful error when the child exits before sending `ready`.
- The error message includes the path to the child's stderr log plus the tail (up to 4 KB) of its contents — so users see the real stack trace (e.g. a `SyntaxError`) instead of a generic timeout.
- Added a race guard: immediately after attaching listeners we check `isProcessRunning(processName)`. If the child has already exited between `startProcess` resolving and our listeners attaching, we surface the error right away rather than waiting for the timeout.
- Unit tests cover both the exit-event path and the already-exited race path.

## Testing Instructions

Automated:

- `npm test -- apps/cli/lib/tests/wordpress-server-manager.test.ts` — 8 tests, all green.

Manual repro (simulates a crashing child):

1. `npm run cli:build`
2. Inject a fatal error at the top of the built child:
   \`\`\`
   printf 'throw new SyntaxError(\"forced crash for testing\");\n%s' \\
     \"\$(cat apps/cli/dist/cli/wordpress-server-child.mjs)\" \\
     > apps/cli/dist/cli/wordpress-server-child.mjs.tmp \\
     && mv apps/cli/dist/cli/wordpress-server-child.mjs.tmp \\
          apps/cli/dist/cli/wordpress-server-child.mjs
   \`\`\`
3. `node apps/cli/dist/cli/main.mjs site start <SITE_ID>`

**Before:** hangs ~2 minutes, then `Failed to start WordPress server: Timeout waiting for ready message…`
**After:** fails immediately with `WordPress server child process exited before becoming ready. See …/studio-site-<id>-error.log` plus the `SyntaxError: forced crash for testing` tail.

Restore with `npm run cli:build`.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/lib/wordpress-server-manager.ts
+++ b/apps/cli/lib/wordpress-server-manager.ts
@@ -4,7 +4,6 @@
  * Manages WordPress server processes via process manager daemon. Each site runs in a separate
  * process that spawns Playground CLI.
  */
-import fs from 'fs';
 import path from 'path';
 import {
 	PLAYGROUND_CLI_ACTIVITY_CHECK_INTERVAL,
@@ -21,7 +20,6 @@ import {
 	type DaemonBusEventMap,
 	sendMessageToProcess,
 } from 'cli/lib/daemon-client';
-import { PROCESS_MANAGER_LOGS_DIR } from 'cli/lib/paths';
 import { ProcessDescription } from 'cli/lib/types/process-manager-ipc';
 import { ServerConfig, ManagerMessagePayload } from 'cli/lib/types/wordpress-server-ipc';
 import { Logger } from 'cli/logger';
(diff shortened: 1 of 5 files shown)


youknowriad:
I'm curious what you think about this @fredrikekelund Do you think it's useful? I think the issue is probably more prominent in our dev environments but I wonder if it can happen on prod too.

apps/cli/lib/wordpress-server-manager.ts
```diff
@@ -148,16 +184,33 @@ async function waitForReadyMessage( pmId: number ): Promise< void > {
 				resolve();
 			}
 		};
+		exitHandler = ( event ) => {
+			if ( event.process.pm_id === pmId && event.event === 'exit' ) {
+				reject( buildChildExitedError( processName ) );
+			}
+		};
```
fredrikekelund: Nice 👍

apps/cli/lib/wordpress-server-manager.ts
```diff
@@ -148,16 +184,33 @@ async function waitForReadyMessage( pmId: number ): Promise< void > {
...
+
+		// Guard against the race where the child exits between `startProcess` resolving and
+		// our listeners being attached: if the process is no longer running now, surface the
+		// error immediately rather than waiting for the ready-message timeout.
+		void ( async () => {
+			const running = await isProcessRunning( processName );
+			if ( ! running ) {
+				reject( buildChildExitedError( processName ) );
+			}
+		} )();
```
fredrikekelund:
```suggestion
		isProcessRunning( processName )
			.then( ( running ) => {
				if ( ! running ) {
					reject( buildChildExitedError( processName ) );
				}
			} )
			.catch( reject );
```
Part syntax nitpicking, part being more explicit about what happens if we catch an exception here.

apps/cli/lib/wordpress-server-manager.ts
```diff
@@ -132,11 +134,45 @@ export async function startWordPressServer(
...
+		try {
+			fs.readSync( fd, buffer, 0, readBytes, size - readBytes );
+		} finally {
+			fs.closeSync( fd );
+		}
+		return buffer.toString( 'utf8' ).trimEnd();
+	} catch {
+		return '';
+	}
+}
```
fredrikekelund: This can be helpful, or it can be misleading (because the logs include contents from prior invocations)…

The best thing would probably be for the process manager daemon to forward the `stderr` output for the current invocation. That's not trivial, but we could probably achieve it by having the process manager daemon keep a rolling buffer of each child's stderr stream and then send the buffer's contents along with the `exit` event in `ProcessManagerDaemon::handleProcessExit`. 

Not something you have to do in this PR, but an agent can probably do an OK job of it.

fredrikekelund (approved):
I haven't tested, but the fundamental logic LGTM 👍

The only part that sticks out is the logs parsing, where I shared a suggestion for how we could make it more robust. Up to you if you'd rather do that here or in a future PR, @youknowriad

youknowriad:
Thanks for the review! Addressed both suggestions in 804f42f8:

- The daemon now keeps a bounded rolling buffer (100 lines / 16 KB) of each child's stderr and attaches it as `stderrTail` on the `exit` event payload — no more leaning on rotated log files that include prior invocations.
- The race guard / IIFE is gone entirely. `subscribeForReadyOrExit` attaches listeners to the daemon bus *before* `startProcess` runs and buffers events by process name until `waitFor(pmId)` is called, so the pre-listener exit case is handled by the same code path as a mid-wait exit. Made the implementation simpler overall (-10 net LOC).

Two new tests: stderr tail flowing end-to-end, and exit events firing while `startProcess` is still in flight.

PR #3161, opened 2026-04-20
Redesign site list and Add Site flow with richer sync metadata

shaunandrews:
This PR redesigns the flow for adding and connecting a site in WordPress Studio.

## Existing Design

The existing flow for adding a site is needlessly complicated, with 4 different branches—and some useless steps around blueprint details and importing a site.

image

Blueprints are jargon, and never very well explained, and include a mostly empty step around the blueprint details.

image

Pulling (more jargon) a site uses a boring list, forcing users to read titles and urls to discern which site they want. It also shows options mixed, making it hard to understand which options are viable. Importing goes to a needless new screen.

image

Once a site is created, you can choose to "connect" it to a site (confusing language, with pulling in the other flow) which uses a narrow modal and the same boring list.

## Changes

I gated these changes behind `enableBlueprintsGallery` flag so without the flag, the trunk should look the same. 

--

## New Design

This PR redesigns the flow for adding a new site, reducing branches down to three: Build, Connect, or Import.

image

Choosing to build a new site shows the option for an empty site, along with the blueprints.

image

Connecting a site shows a much larger, visual grid of sites split into groups depending on viability with those that need hosting features or a plan shows with more helpful messaging.

image

## How AI was used in this PR

Used Claude Code throughout as a pair for the larger refactors — the Add Site redesign, the pagination-safe reconcile logic, and the layout or centering plumbing. Reviewed all diffs manually; the branch was rebased on trunk and reconciled by hand.

## Detailed Changes

**Site list (Sync tab + sync metadata)**
- Enrich connected-site data with icon, plan name, and creation date (stored on the local connected-site record, see `appdata-v1.json`).
- Switch the WordPress.com sites fetch to the v1.3 me/sites API with server-side filtering + pagination.
- Redesign the site list as a grouped card grid with server-side search, mshot thumbnails, site icons, and plan badges.
- Fix a pagination-reconcile bug that marked any connected site beyond page 1 as deleted. Missing connected sites are now verified with an individual site GET and only flagged deleted on a confirmed 404.

**Add Site flow**
- Replace the three-step stepper with a single entry screen offering Build / Connect / Import paths, a dedicated blueprint picker, and a Connect screen that reuses the Sync site list.
- Modal chrome is now an X button and a floating action row (no full-width header/footer bars) so the Screen height is constant across steps and content doesn't shift when navigating.
- Backup imports preflight the dropped file (`preflightBackup` IPC) before advancing — invalid archives are caught on the entry screen.
- Added a sign-out link to the add-site options, and widened CSP `img-src` to allow site icons from any HTTPS domain.

## Testing Instructions

1. Launch a dev build. You should see a larger window and the system color scheme honored on first run.
2. Open the site list from the sidebar — sites should appear as cards with thumbnails, plan badges, and grouped by status. Try the search and pagination.
3. From a site's Sync tab, open the Connect modal — confirm the \"Connect your site\" title is centered above the search input and the Enable button on sites needing hosting features is readable in dark mode.
4. Open Add Site from the toolbar:
   - Verify site adding options render on the entry screen and the dot grid shows behind them.
   - Drop an invalid archive into the Import card — the preflight error should surface inline without advancing.
   - Advance through any sub-path and back — content centering should not shift between steps; the X button and action row stay pinned.

## Pre-merge Checklist

- [ ] Have you tested on macOS, Windows, and Linux where applicable?
- [ ] Have you added automated tests where applicable?


apps/studio/src/modules/add-site/components/illustrations.tsx
```diff
@@ -0,0 +1,150 @@
...
+				strokeLinecap="round"
+				strokeDasharray="1 5"
+				style={ SPIN_STYLE }
+				className="animate-slow-spin"
+			/>
+		</svg>
+	);
+}
+
+export function ConnectSiteIllustration() {
```
katinthehatsite, line 59: I like these illustrations, I think the look quite nice with animations 👍

katinthehatsite (commented):
I was trying to use the option to Connect a site but these preview generation step never finished for me:

Image

katinthehatsite (commented):
Another observation is that visually it is very hard to see which site belongs to WP.com and which site belongs to Pressable in this list:

Image

I think it would be great to add some element that would allow to distinguish them quickly.

katinthehatsite (commented):
One other comment is that I found this `Preview` link slightly strange because I expecred that once I am done with the preview and made my modifications, I could somehow port them into Studio. Instead, I was somewhat in the dead end because it looked like something I could edit but was meant to be serving as preview. 

Also, the description of the WordPress.com option was sort of confusing. Once I created a site, it looked like a regular Studio site:

Image

katinthehatsite:
> Another observation is that visually it is very hard to see which site belongs to WP.com and which site belongs to Pressable in this list

This should be now addressed ✅

katinthehatsite:
> One other comment is that I found this Preview link slightly strange because I expecred that once I am done with the preview and made my modifications, I could somehow port them into Studio. Instead, I was somewhat in the dead end because it looked like something I could edit but was meant to be serving as preview.

I removed this functionality for now and will move it to the Linear issue to make sure we think through this flow. We could potentially rename this as something like "Experiment" or "Try out" so that the user does not have an expectation that they will be able to activate it in Studio from Playground and that it is a circular process.

katinthehatsite:
> I was trying to use the option to Connect a site but these preview generation step never finished for me

I no longer saw this so I left this as is for now.

apps/studio/src/modules/add-site/components/new-site-options.tsx
```diff
@@ -0,0 +1,237 @@
...
+	);
+}
+
+const BLUEPRINT_DISPLAY_NAMES: Record< string, string > = {
+	'Quick Start': 'WordPress.com',
+	Development: 'WordPress Dev',
+	Commerce: 'WooCommerce',
+};
+
+const BLUEPRINT_EXCERPT_OVERRIDES: Record< string, string > = {
```
katinthehatsite: I am not seeing this translated, I will double-check now if this is correct 🤔

katinthehatsite: Should now be fixed with the most recent commit.

katinthehatsite:
@Automattic/yolo I would appreciate the review. For now, I gated these changes behind `enableBlueprintsGallery` flag so without the flag, the trunk should look the same. To review these changes, make sure to enable the flag first. Please test different ways to create a site. I looked at it for some time already so I might be missing some things.

apps/studio/src/index.ts
```diff
@@ -279,7 +279,7 @@ async function appBoot() {
 			const basePolicies = [
 				"default-src 'self'", // Allow resources from these domains
 				"script-src-attr 'none'",
-				"img-src 'self' https://*.gravatar.com https://*.wp.com https://blueprintlibrary.wordpress.com data:",
+				"img-src 'self' https: data:",
```
sejas: What new images are we loading?

We can consider adding specific domains to the list. Same for `apps/studio/index.html`.

sejas (commented):
Shaun, Kat, Thanks for creating the PR. I like the new designs ✨.

A few things I found:
- We remove the whitelist of domains in `apps/studio/index.html` and `apps/studio/src/index.ts`
- The PR description says: `Backup imports preflight the dropped file (preflightBackup IPC) before advancing — invalid archives are caught on the entry screen.`, but that seems not implemented. I suggest updating the description and maybe creating a Linear issue for that.
- The Studio window started at ~900x600 which cut the import from a backup button. Why not showing the three buttons in the same row?
- The scroll bar is not at the border and it conflicts with the close X.

Image

- The Add site modal cannot be dragged by clicking in the background or the header. It shows an animation instead(Not sure if that is an issue on trunk).
- When clicking Build a new site, an error appears in the console.

Image

- The Blueprint descriptions at the bottom are a bit blurred.

Image

- The left sidebar hides if the window width is smaller than 1100, not sure if that's intentional.

https://github.com/user-attachments/assets/a64d3dda-b869-4897-84b6-4de397c95cb6

The PR is quite large, so I might have missed something.
Happy to have a second pass when E2E pass.

katinthehatsite:
> The PR description says: Backup imports preflight the dropped file (preflightBackup IPC) before advancing — invalid archives are caught on the entry screen., but that seems not implemented. I suggest updating the description and maybe creating a Linear issue for that.

There is already an issue for it in https://linear.app/a8c/issue/STU-1644/blueprints-gallery-library-consider-preflight-validation-for-import 😄 I removed the logic from this PR as it was getting too complex

apps/studio/src/index.ts
katinthehatsite: I am not sure what the original intent was, but I tested without the changes and I don't see any specific regressions. I left it as it previously was while whitelisting the blueprints library,

katinthehatsite:
> The Studio window started at ~900x600 which cut the import from a backup button. Why not showing the three buttons in the same row?

I see it as well but I think this is the design question rather than the implementation question. I will move it to the new issue and we can discuss with Shaun there.

katinthehatsite:
> The scroll bar is not at the border and it conflicts with the close X.

I rebased trunk and I don't see this overlap anymore in this PR: 

Screenshot 2026-04-28 at 12 20 59 PM

Are you still seeing it @sejas ?

katinthehatsite:
> The Add site modal cannot be dragged by clicking in the background or the header. It shows an animation instead(Not sure if that is an issue on trunk).

It can be dragged by the header:

https://github.com/user-attachments/assets/566ce3cd-4c7a-4a51-8d61-83119b2ac02b

Which works the same as on trunk:

https://github.com/user-attachments/assets/da52b419-5fff-4053-935f-4a6b92b5da71

I think we can leave it as is with this PR as it replicates the existing behaviour.

katinthehatsite:
> When clicking Build a new site, an error appears in the console.

I was testing this but not seeing this specific error. I am adding below the video of testing, am I taking different steps from you?


https://github.com/user-attachments/assets/17c0ec01-bb15-4ff9-8cbb-266adbb9884f

katinthehatsite:
> The left sidebar hides if the window width is smaller than 1100, not sure if that's intentional.

I also retested this and can not see it on my end:

https://github.com/user-attachments/assets/be9a53d3-1738-461a-a0f1-2bd0c53b472e

Would you be able to check this with the most recent changes?

katinthehatsite:
> The Blueprint descriptions at the bottom are a bit blurred.

I dropped the gradient height for bluriness. I think it should be better now. Let me know if it still seems too blurry.

katinthehatsite (approved):
I am approving the changes, they are behind the feature flag so I think we are good to go.

PR #3165, opened 2026-04-20
apps/ui: add per-site settings page

youknowriad:
## Related issues

- Related to the apps/ui site-settings work in context of the new per-site UI.

## How AI was used in this PR

Built iteratively via Claude Code — scaffolded the settings page, shared field builders, and tabs wrapper, then iterated on the typography and tab styling against design feedback. The stale-state bug in the CLI events subscriber was found and fixed after a QA pass surfaced a "custom domain restored after save" symptom.

## Proposed Changes

- New `/sites/$siteId/settings` route in apps/ui, accessible from the sidebar site-actions menu. Top of the page reuses `SiteDropdown` for context; tabs (General + Debugging) sit above the scroll container so the tablist's border-bottom spans the full main-area width and aligns with the sidebar's right edge. Saves via `updateSite` IPC (CLI-backed `site set`).
- Shared `components/site-fields` DataForm field builders (name, php/wp version, admin credentials, custom domain, xdebug/debug log/debug display) used by both the onboarding create-site form and the new settings form.
- Shared `components/tabs` wrapper around `@wordpress/components`' private `Tabs` (unlocked the same way `ValidatedInputControl` is in create-site-form). Condensed styling with DS tokens: 1px neutral baseline (matches the sidebar border), 1px active indicator in `fg-content-neutral`, compact labels with no inline padding so the underline aligns with the text, `lg * 2` gap, medium font size, regular/semibold weight on inactive/active.
- Universal `h1`–`h6` rules in `index.css` using `--wpds-font-family-heading`, `--wpds-font-weight-regular`, `--wpds-color-fg-content-neutral`, and the `2xl → xs` size/line-height ramp. Onboarding `.title` class slimmed to layout spacing only so all page titles share the same type.
- `SiteDetails` in apps/ui expanded (`adminUsername`, `adminPassword` (base64), `adminEmail`, `enableXdebug`, `enableDebugLog`, `enableDebugDisplay`, `isWpAutoUpdating`) and `Connector` gained `updateSite` + `getXdebugEnabledSite`.
- **Bug fix in `cli-events-subscriber`**: `siteDetailsToServerDetails` spread `existingDetails` then `site`, but JSON serialization drops cleared optional fields (e.g. `customDomain` after `site set --domain ''`), so the absent key let the stale value survive in memory. Merge now pulls only Studio-only fields (tls/theme/sort/runtime flags) from `existingDetails` and lets the CLI payload be authoritative for everything else.
- `useUpdateSite` no longer invalidates queries immediately — the CLI edit and its `site-updated` event come from two different CLI processes, so a refetch fired the moment the IPC resolves races ahead of the event handler and re-seeds the form with pre-event data. `useSyncSitesWithEvents` picks up the invalidation once the event lands.

## Testing Instructions

- Open a local site from the sidebar, click the per-site kebab menu → **Site settings**. Verify the page shows the SiteDropdown at the top, the "Site settings" h1, and General / Debugging tabs.
- General tab: change site name, PHP version, admin credentials, toggle "Use custom domain" with and without HTTPS, save. Confirm site-list reflects the changes after save and that the form re-syncs without flashing the old values.
- **Custom domain clearing regression** — toggle off "Use custom domain" and save. Confirm the checkbox stays unchecked after the save completes (previously would visually "restore" the old domain) and that the underlying site config no longer has a custom domain.
- Debugging tab: toggle Xdebug. If another site already has it enabled, the save button should stay disabled and the field description should name that site.
- Toggle debug log + debug display and save; confirm persistence.
- Deep-link `/sites/$siteId/settings?tab=debugging` and confirm the Debugging tab is active on load.
- Verify onboarding page titles (`/onboarding`, `/onboarding/create`) render with the universal h1 style (regular weight, heading family) and no layout regression.

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/modules/cli/lib/cli-events-subscriber.ts
+++ b/apps/studio/src/modules/cli/lib/cli-events-subscriber.ts
@@ -13,13 +13,46 @@ import { captureSiteThumbnail } from 'src/lib/capture-site-thumbnail';
 import { executeCliCommand } from 'src/modules/cli/lib/execute-command';
 import { SiteServer } from 'src/site-server';
 
+// Fields owned by Studio that the CLI never emits and that must survive a
+// site-event merge (TLS material, renderer-computed theme info, sort order,
+// transient runtime flags).
+const STUDIO_ONLY_DETAIL_KEYS = [
+	'tlsKey',
+	'tlsCert',
+	'themeDetails',
+	'sortOrder',
+	'isAddingSite',
+	'latestCliPid',
+] as const satisfies readonly ( keyof SiteServer[ 'details' ] )[];
+
+function pickStudioOnlyDetails( details?: SiteServer[ 'details' ] ) {
+	if ( ! details ) {
+		return {} as Partial< SiteServer[ 'details' ] >;
+	}
+	const picked: Partial< SiteServer[ 'details' ] > = {};
+	for ( const key of STUDIO_ONLY_DETAIL_KEYS ) {
+		const value = details[ key ];
+		if ( value !== undefined ) {
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any
+			( picked as any )[ key ] = value;
+		}
+	}
+	return picked;
+}
+
+// The CLI event is the authoritative snapshot for every field in the shared
+// `siteDetailsSchema`. JSON serialization drops `undefined` values, so a
+// field that's been cleared (e.g. `customDomain` after `site set --domain ''`)
+// arrives here as an absent key. Spreading `...existingDetails` first would
+// then preserve the stale value. Invert the merge: start from Studio-only
+// fields, then let the CLI payload set or omit everything else.
 function siteDetailsToServerDetails(
 	site: SiteDetails,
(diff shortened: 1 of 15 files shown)


apps/studio/src/modules/cli/lib/cli-events-subscriber.ts
```diff
@@ -13,13 +13,46 @@ import { captureSiteThumbnail } from 'src/lib/capture-site-thumbnail';
...
+// then preserve the stale value. Invert the merge: start from Studio-only
+// fields, then let the CLI payload set or omit everything else.
 function siteDetailsToServerDetails(
 	site: SiteDetails,
 	running: boolean,
 	existingDetails?: SiteServer[ 'details' ]
 ): SiteServer[ 'details' ] {
 	return {
-		...existingDetails,
+		...pickStudioOnlyDetails( existingDetails ),
```
youknowriad, line 55: I noticed that disabling "custom domain" was not working. This fixes it. I suspect it's a bug in the existing app too.

PR #3172, opened 2026-04-21
Update SQLite Database Integration to v3.0.0-rc.1

wojtekn:
## Related issues

- Related to https://github.com/WordPress/sqlite-database-integration/releases/tag/v3.0.0-rc.1

## How AI was used in this PR

Claude Code identified and applied all changes.

## Proposed Changes

- Bump `SQLITE_DATABASE_INTEGRATION_VERSION` from `v2.2.23` to `v3.0.0-rc.1`
- Remove `WP_SQLITE_AST_DRIVER: true` from `defaultConstants` in `wordpress-server-child.ts`
- Remove `php.defineConstant('WP_SQLITE_AST_DRIVER', true)` from `run-wp-cli-command.ts`
- Define `DB_NAME: 'wordpress'` at runtime in both `wordpress-server-child.ts` and `run-wp-cli-command.ts`

The `WP_SQLITE_AST_DRIVER` constant was a feature flag enabling the new AST-based driver in earlier versions. Starting with v3.0.0, the old driver has been removed and the flag is no longer needed.

SQLite v3.0.0 also removed the fallback for a missing `DB_NAME` — it now requires a non-empty value at runtime or it bails with "Error establishing a database connection". Studio strips the default MySQL constants (including `DB_NAME`) from `wp-config.php` after site creation, which previously worked because v2 had a `'database_name_here'` fallback. We fix this by explicitly defining `DB_NAME: 'wordpress'` as a runtime constant (the same value Playground's `ensureWpConfig` would set), so sites where `DB_NAME` was stripped still work correctly.

## Testing Instructions

- Create a new local site and verify it starts correctly on first launch (no "Error establishing a database connection")
- Restart an existing site and verify it still works
- Run a WP-CLI command (e.g. `wp option get siteurl`) and confirm it works

## Pre-merge Checklist

### DIFF
--- a/apps/cli/lib/run-wp-cli-command.ts
+++ b/apps/cli/lib/run-wp-cli-command.ts
@@ -65,8 +65,6 @@ export async function runWpCliCommand(
 	try {
 		await php.setSapiName( 'cli' );
 
-		php.defineConstant( 'WP_SQLITE_AST_DRIVER', true );
-
 		php.mkdir( '/wordpress' );
 		await php.mount( '/wordpress', createNodeFsMountHandler( siteFolder ) );
 
--- a/apps/cli/lib/run-wp-cli-command.ts
+++ b/apps/cli/lib/run-wp-cli-command.ts
@@ -65,6 +65,10 @@ export async function runWpCliCommand(
 	try {
 		await php.setSapiName( 'cli' );
 
+		// Fallback for sites where DB_NAME was stripped from wp-config.php.
+		// The SQLite driver (v3+) requires a non-empty DB_NAME at runtime.
+		php.defineConstant( 'DB_NAME', 'wordpress' );
+
 		php.mkdir( '/wordpress' );
 		await php.mount( '/wordpress', createNodeFsMountHandler( siteFolder ) );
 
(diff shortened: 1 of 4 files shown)


sejas (approved):
Changes look good. Existing sites were updated. New version doesn't have a `version.php` but the new version can be found in readme.txt
I tried starting existing sites, creating new ones, creating preview sites and authenticating and pushing a new site. Everything worked smoothly.

Screenshot 2026-04-23 at 14 01 08

wojtekn:
Thanks @sejas !

The `version.php` which defines the constant is still there, in `wp-includes/database/` and it defines the correct version. When I dump `SQLITE_DRIVER_VERSION` constant in `wp-config.php`, I see:

```
string(10) "3.0.0-rc.1"
```

But it uncovers another issue - when Studio copies the SQLite plugin to the site directory, it doesn't remove the existing files, so after each update, the directory contains not only new files, but old ones, too.

wojtekn:
@sejas I fixed the mentioned issue: 81e21a4b965c8ed7267eebd1aadbe2d74d36984e

sejas (commented):
Thanks for fixing it! 🙌

PR #3173, opened 2026-04-21
CLI: Send X-WPCOM-Session-ID header on WPCOM AI proxy requests

epeicher:
## Related issues

- N/A

## How AI was used in this PR

Implementation was done with Claude Code. I checked the logs and reviewed the code.

## Proposed Changes

- Adds an `X-WPCOM-Session-ID` custom header to every request the `studio code` CLI makes to the WordPress.com AI proxy, alongside the existing `X-WPCOM-AI-Feature` header.
- The value is the Studio AI session recorder's UUID (the same ID used for session files under `~/Library/Application Support/Studio/sessions/`), so WPCOM can correlate multiple requests back to the same Studio conversation.
- Threads an optional `sessionId` through `resolveAiEnvironment` / the wpcom provider's `resolveEnv`. The header is only emitted when a recorder exists (i.e. session persistence is not disabled).

## Testing Instructions

1. Start a `studio code` session (CLI or desktop) and send a prompt.
2. Open Logstash, filter by `feature: studio-assistant-anthropic`, and check that `properties.session_id` is populated with the `studio code` session. You can use the following link 5fd435e7f90d2536111da67ff7eff216-logstash
4. Send a follow-up prompt in the same session — confirm `properties.session_id` is the same value.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/ai/index.ts
+++ b/apps/cli/commands/ai/index.ts
@@ -414,7 +414,10 @@ export async function runCommand( options: {
 		retryAttempt = 0
 	): Promise< { status: TurnStatus; usage?: { numTurns: number; costUsd?: number } } > {
 		await maybeAutoSwitchProvider();
-		const env = await resolveAiEnvironment( currentProvider );
+		const recorder = await ensureSessionRecorder();
+		const env = await resolveAiEnvironment( currentProvider, {
+			sessionId: recorder?.sessionId,
+		} );
 
 		ui.beginAgentTurn();
 
--- a/apps/cli/ai/auth.ts
+++ b/apps/cli/ai/auth.ts
@@ -4,6 +4,7 @@ import {
 	getAiProviderDefinition,
 	hasInlineWpcomAuth,
 	type AiProviderId,
+	type ResolveAiEnvironmentOptions,
 } from 'cli/ai/providers';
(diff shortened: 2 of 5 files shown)


apps/cli/commands/ai/index.ts
```diff
@@ -414,7 +414,10 @@ export async function runCommand( options: {
 		retryAttempt = 0
 	): Promise< { status: TurnStatus; usage?: { numTurns: number; costUsd?: number } } > {
 		await maybeAutoSwitchProvider();
-		const env = await resolveAiEnvironment( currentProvider );
+		const recorder = await ensureSessionRecorder();
+		const env = await resolveAiEnvironment( currentProvider, {
+			sessionId: recorder?.sessionId,
```
youknowriad, line 419: Why is this optional, is this for the initial prompt? Does that mean we need to generate the sessionId before sending the prompt?

epeicher: Not really, it's because the code currently allows [disabling session persistence](https://github.com/Automattic/studio/blob/2283666985c63a8991d184f3fbe7c41593ba6adc/apps/cli/commands/ai/index.ts#L99)

youknowriad (approved):

PR #3176, opened 2026-04-21
apps/ui: add disconnect flow and reconcile live env on turn start

youknowriad:
## Related issues

- Related to live-site session-environment handling in apps/ui

## How AI was used in this PR

Co-authored end-to-end with Claude (Opus 4.7, 1M context). The agent scoped the refactor, proposed architectures, pushed back on a larger event-model rewrite in favor of the minimal change, and implemented the code/tests. Reviewers should pay particular attention to:
- the reconciliation write in \`setSessionEnvironment\` / \`continueAiSession\` — it's the only place we intentionally append a \`site.selected\` event from the main process.
- the disconnect-dialog UX (matches the existing DeleteSiteDialog pattern).

## Proposed Changes

- **Disconnect action in the site dropdown.** A \`Disconnect\` button sits next to Pull / Push in the Live row and opens a confirmation dialog ([DisconnectSiteDialog](apps/ui/src/components/site-dropdown/disconnect-site-dialog.tsx)) modeled on \`DeleteSiteDialog\`. The dropdown closes before the dialog opens so the two overlays don't stack.
- **Connect (was Setup) no longer auto-pushes.** Picking a WP.com site from the publish picker now just persists the connection and closes the picker — no implicit \`pushSiteToLive\` + auto-open-browser. Push stays available via the usual button in the Live row.
- **Effective-env derivation on the renderer.** A new \`useSessionEffectiveEnvironment\` hook (with a pure \`deriveEffectiveEnvironment\` helper in \`@studio/common/ai/sessions/effective-site\`) joins \`summary.activeEnvironment\` + \`summary.lastSelectedWpcomSiteId\` with the current connected-sites query. When a session flipped to Live whose remote is no longer linked, the pill/header show Local automatically — no JSONL writes needed for the passive view.
- **Main-process reconciliation at turn start.** \`continueAiSession\` now calls \`reconcileSessionEnvironmentBeforeRun\` before spawning the CLI. If the session is flagged live but the remote blog id is gone from connected-sites, it appends a \`site.selected\` event pinning the session to its local owner, so the CLI's replay sets the right \`ui.activeSite\` and the agent's system prompt stops referring to the removed remote. Without this, the pill said Local while the agent still answered "you're on the live site".
- **Summary / reducer / store updates** to carry \`lastSelectedWpcomSiteId\` and fold legacy \`environment.selected\` events into the active-environment computation alongside \`site.selected\`, so existing sessions keep working.
- **Global focus-style tweak in [apps/ui/src/index.css](apps/ui/src/index.css):** extends the existing \`button:focus:not(:focus-visible)\` override to also revert @wordpress/ui Button's background/color/border shift. Without this, a dialog that auto-focuses a button on open (common with \`Dialog.Action\` as first focusable) shows the "active" look even on mouse-open.
- **File watcher + renderer subscription** for external JSONL edits ([apps/studio/src/modules/ai-sessions/file-watcher.ts](apps/studio/src/modules/ai-sessions/file-watcher.ts), \`useSyncSessionsWithEvents\` in apps/ui). Any write to a session file — UI, CLI, hand edit — invalidates the matching react-query cache so the view recomputes. Debounced 100 ms.

## Testing Instructions

1. **Disconnect + agent coherence**: open a session that is currently on Live. Use the site dropdown's \`Disconnect\` button, confirm the dialog. Verify:
   - The pill in the composer flips to Local and the Live radio becomes disabled.
   - Ask the agent \`what site am I working on?\`. It should name the local site, not the previously-linked WP.com site.
2. **Connect no longer auto-pushes**: from a local site with no live link, click \`Connect\`, pick a WP.com site in the picker. Verify no push is triggered — the Live row appears with Pull/Push/Disconnect available and the WP.com site is unchanged until you click \`Push\` manually.
3. **Reconnect to the same live**: after a disconnect, reconnect. The pill should still say Local (the user never asked for Live back yet). Click Live in the pill — it should flip with a fresh \`site.selected\` event recorded.
4. **Reconnect to a different live**: disconnect, then Connect to a different WP.com site. The pill stays Local; picking Live now targets the new remote id.
5. **Focus styling regression**: open the Disconnect dialog with a mouse — the Cancel button should not look "active" on its own. Then press Tab — the focus ring should appear (keyboard focus).
6. **External edits**: with a session open, append a \`site.selected\` event to its JSONL by hand and save. The UI should update within ~100ms.

## Pre-merge Checklist

- [x] Did the test suite pass locally? (1516 passing)
- [ ] Have you tested in the latest stable version of Chrome, Firefox, and Safari? _(Electron only — tested in the dev build)_


apps/ui/src/index.css
```diff
@@ -88,11 +88,17 @@ textarea,
...
+   Also revert the background/color/border shift that @wordpress/ui's Button
+   applies on `:focus` — when a Dialog auto-focuses a button on open, the
+   "active" look would otherwise show without any user intent. */
 button:focus:not(:focus-visible),
 a:focus:not(:focus-visible),
 [role="button"]:focus:not(:focus-visible) {
 	outline: none;
+	background-color: var(--wp-ui-button-background-color);
+	color: var(--wp-ui-button-foreground-color);
+	border-color: var(--wp-ui-button-border-color);
```
youknowriad, line 103: Do we some specific styles in some existing dialogs that we could drop thanks to the addition of these global fixes.

PR #3177, opened 2026-04-21
apps/ui: add language switcher to Settings page

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Claude Code drafted the plumbing (connector + form field + persister flush), the row grouping, and the combobox background override. I reviewed every diff, pushed back on the initial styling and cache strategies, and verified the behaviour end-to-end in the running app.

## Proposed Changes

- Wire the existing `saveUserLocale` main-process handler through the IPC connector's `setUserPreferences` fan-out so the Settings form can persist the app locale alongside editor/terminal/color-scheme.
- Make `locale` part of `WritableUserPreferences` (typed as `SupportedLocale`) while keeping the read side as `string | undefined` — the main process may resolve to locales outside our catalog, and the form clamps those down to `'en'` before submit.
- Add a "Language" field to the DataForm in `SettingsView`, sourced from `supportedLocaleNames`. Group editor + terminal on the same row via a DataForm `row` layout.
- Reload the window on locale save. Most of apps/ui imports `__` from `@wordpress/i18n` directly and doesn't subscribe to locale changes, so a reload is the only way every string picks up the new catalog.
- Flush the persisted query cache via `persister.removeClient()` before reloading. The `createSyncStoragePersister` wraps `persistClient` in a ~1s throttle, so a fresh `setQueryData` often hasn't hit localStorage before navigation; combined with `staleTime: Infinity` on `useUserPreferences`, the rehydrated cache would otherwise serve the stale locale back after reload.
- Scope an override in `settings-view/style.module.css` so `@wordpress/components`' ComboboxControl (what DataForm's adaptive picker swaps to for the 21-item Language field) picks up the same background as the input inside, instead of leaving the body surface to bleed through as an unthemed gray.

## Testing Instructions

- Open the app, go to Settings → Preferences.
- Confirm the editor and terminal selects sit side-by-side on the same row.
- Change the Language, click Save. The window reloads and every string renders in the new language. The Language field reflects the newly-saved locale (not the previous one).
- Open the Language picker and confirm the combobox wrapper + dropdown popup have a solid background matching the text input — no gray around the input chrome.
- Change editor / terminal / appearance and save; confirm no reload happens and the cached form state stays consistent.

## Pre-merge Checklist

### DIFF
--- a/apps/ui/src/components/settings-view/style.module.css
+++ b/apps/ui/src/components/settings-view/style.module.css
@@ -75,17 +75,6 @@
 	gap: var(--wpds-dimension-padding-xl);
 }
 
-/* DataForm's adaptive picker swaps to @wordpress/components' ComboboxControl
-   once a field has 10+ elements (e.g. the Language picker). Its wrapper
-   has no background, so the body surface bleeds through around the inner
-   input and suggestion rows. Force both to the same background the input
-   itself uses (`--wp-components-color-background`) so the control reads
-   as one solid field. */
-.form :global(.components-combobox-control__suggestions-container),
-.form :global(.components-form-token-field__suggestions-list) {
-	background-color: var(--wp-components-color-background, #fff);
-}
-
 .actions {
 	display: flex;
 	justify-content: flex-end;
--- a/apps/ui/src/components/settings-view/style.module.css
+++ b/apps/ui/src/components/settings-view/style.module.css
@@ -75,6 +75,17 @@
 	gap: var(--wpds-dimension-padding-xl);
 }
 
+/* DataForm's adaptive picker swaps to @wordpress/components' ComboboxControl
+   once a field has 10+ elements (e.g. the Language picker). Its wrapper
+   has no background, so the body surface bleeds through around the inner
+   input and suggestion rows. Force both to the same background the input
+   itself uses (`--wp-components-color-background`) so the control reads
+   as one solid field. */
+.form :global(.components-combobox-control__suggestions-container),
+.form :global(.components-form-token-field__suggestions-list) {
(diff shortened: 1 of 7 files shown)


apps/ui/src/components/settings-view/style.module.css
```diff
@@ -75,6 +75,17 @@
...
 }
 
+/* DataForm's adaptive picker swaps to @wordpress/components' ComboboxControl
+   once a field has 10+ elements (e.g. the Language picker). Its wrapper
+   has no background, so the body surface bleeds through around the inner
+   input and suggestion rows. Force both to the same background the input
+   itself uses (`--wp-components-color-background`) so the control reads
+   as one solid field. */
+.form :global(.components-combobox-control__suggestions-container),
+.form :global(.components-form-token-field__suggestions-list) {
```
youknowriad: Can we move this style to a more global place to apply to all instances of these two classnames. Maybe later other forms will use the same control.

apps/ui/src/data/core/types.ts
```diff
@@ -202,16 +204,18 @@ export interface UserPreferences {
 	editor: SupportedEditor | null;
 	terminal: SupportedTerminal | null;
 	colorScheme: ColorScheme;
-	// Resolved locale string (e.g. "en", "fr-FR"). Read-only in v1 — there's
-	// no connector method for changing it yet.
+	// Resolved locale string (e.g. "en", "fr-FR"). May be any locale the main
+	// process resolves — including ones not in our translation catalog — so
+	// consumers should narrow with `isSupportedLocale` before acting on it.
```
youknowriad: Can we drop this comment, feels useless to me.

PR #2920, opened 2026-03-26
Add phpMyAdmin to What's New dialog

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

I used AI to implement the changes, including SVG, and reviewed code.

## Proposed Changes

- Added phpMyAdmin as the second item in the What's New carousel, immediately after Dark Mode
- Created a new `phpmyadmin-illustration.svg` illustration (360×195px, dark-themed, consistent with existing illustrations)
- The carousel order is now: Dark Mode → phpMyAdmin → WP-CLI → Blueprints

CleanShot 2026-03-26 at 13 56 02

## Testing Instructions

- Trigger the What's New modal from top menu
- Verify phpMyAdmin appears as the second page with the correct title and description
- Verify the illustration renders correctly
- Click through all pages to confirm the carousel still works end-to-end

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #3203, opened 2026-04-23
apps/studio: Show What's New modal only on first launch or when explicitly forced

sejas:
## Related issues

- Related to https://github.com/Automattic/studio/pull/2920

## How AI was used in this PR

Claude Opus was used to brainstorm alternative strategies (content-key vs. boolean override), implement the chosen approach, and update tests. The author reviewed and validated the diff.

## Proposed Changes

- Replace `FORCE_WHATS_NEW_WHEN_PATCH_CHANGED` (default `true`) with a simpler `FORCE_SHOW_WHATS_NEW` constant (default `false`) in `apps/studio/src/constants.ts`.
- Drop semver comparison in `selectIsNewVersion` (`apps/studio/src/stores/app-version-api.ts`). The modal now auto-shows only when:
  1. The user has no stored `lastSeenVersion` (first-time launch), **or**
  2. `FORCE_SHOW_WHATS_NEW === true` **and** the stored version differs from the current app version.
- Patch, minor, and major version bumps alone no longer trigger the modal.
- Help menu → "What's New" still opens the modal on demand, regardless of the flag.
- Remove the now-unused `semver` import from this file (still used elsewhere).
- Update `selectIsNewVersion` tests to cover the new semantics.

### Release workflow

When shipping new modal content:

1. Edit `whatsNewPages` in `apps/studio/src/modules/whats-new/components/whats-new-modal.tsx`.
2. Set `FORCE_SHOW_WHATS_NEW = true` in `apps/studio/src/constants.ts`.
3. Bump the app version and ship. Each user sees the modal once; dismissing it writes the current version and suppresses re-shows.
4. After the release, flip the flag back to `false` so subsequent patch/minor bumps don't retrigger it.

## Testing Instructions

- Run \`npm test -- src/stores/tests/app-version-api.test.ts\` — 7/7 pass.
- Manual test with \`FORCE_SHOW_WHATS_NEW = false\` (default):
  - Fresh install (remove \`lastSeenVersion\` in \`~/.studio/app.json\`) → modal appears once, dismiss writes version.
  - Restart the app → modal does **not** reappear.
  - Manually edit \`~/.studio/app.json\` to bump \`lastSeenVersion\` (simulating a version bump) → modal still does **not** appear.
  - Help menu → "What's New" → modal opens.
- Manual dogfood with \`FORCE_SHOW_WHATS_NEW = true\`:
  - With a stored \`lastSeenVersion\` that differs from \`appVersion\` → modal appears once, dismiss writes current version, subsequent launches quiet until another version bump.

https://github.com/user-attachments/assets/23ea4237-0323-48c3-baca-877f256bdc31

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/stores/tests/app-version-api.test.ts
+++ b/apps/studio/src/stores/tests/app-version-api.test.ts
@@ -1,6 +1,6 @@
 import { configureStore } from '@reduxjs/toolkit';
 import { vi } from 'vitest';
-import { FORCE_WHATS_NEW_WHEN_PATCH_CHANGED } from 'src/constants';
+import { FORCE_SHOW_WHATS_NEW } from 'src/constants';
 import { getIpcApi } from 'src/lib/get-ipc-api';
 import { appVersionApi, selectIsNewVersion } from 'src/stores/app-version-api';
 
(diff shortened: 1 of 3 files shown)


apps/studio/src/stores/tests/app-version-api.test.ts
```diff
@@ -1,6 +1,6 @@
 import { configureStore } from '@reduxjs/toolkit';
 import { vi } from 'vitest';
-import { FORCE_WHATS_NEW_WHEN_PATCH_CHANGED } from 'src/constants';
+import { FORCE_SHOW_WHATS_NEW } from 'src/constants';
```
katinthehatsite, line 3: what happens when `FORCE_SHOW_WHATS_NEW = true` and the    
  version has changed? 

Do we have a test to cover for that scenario?

katinthehatsite (approved):
I think the changes look good but we will need to remember to set it to true for minor versions where we might want to display the changes. 

Perhaps we might want to update some instructions for the release process for this?

apps/studio/src/stores/tests/app-version-api.test.ts
sejas: When FORCE_SHOW_WHATS_NEW is true and the version changed the modal should appear. I'll add a test case for that. Thanks for suggesting ! 🙌

sejas: I added the tests here https://github.com/Automattic/studio/pull/3206

PR #3205, opened 2026-04-23
Auth: Register wp-studio:// scheme on Linux

gavande1:
## Related issues

- Fixes [STU-1629](https://linear.app/a8c/issue/STU-1629/linux-wp-studio-protocol-handler-not-registered-wpcom-login-redirect)

## How AI was used in this PR

Used Claude Code to locate the protocol registration path (`app.setAsDefaultProtocolClient` in `apps/studio/src/index.ts`), inspect the `@electron-forge/maker-deb` config surface to confirm `mimeType` is the supported option, and propose the minimal packaging change. I reviewed the diff and verified the approach against the existing `docs/linux.md` workaround before committing.

## Proposed Changes

- `apps/studio/forge.config.ts`: pass `mimeType: ['x-scheme-handler/wp-studio']` to `MakerDeb`. Electron Forge writes `MimeType=x-scheme-handler/wp-studio;` into the generated `.desktop` entry, and `dpkg` postinst runs `update-desktop-database` so the scheme is registered system-wide on DEB install.
- `docs/linux.md`: add a step telling source-built users to run `xdg-mime default studio.desktop x-scheme-handler/wp-studio`. Without this, browsers still show "Open With… / No Apps Available" on the `wp-studio://` callback even when the hand-crafted `.desktop` file declares the MimeType.

Why `setAsDefaultProtocolClient` alone wasn't enough on Linux: that Electron API ultimately shells out to `xdg-mime`, which only succeeds if there's a `.desktop` file on disk declaring the MimeType. Packaged DEB builds had no such entry, so the scheme silently failed to register.

### DEB-packaged build (the primary fix)

1. On a Linux machine (or VM), `npm ci && npm run package`.
2. Go to `apps/studio/out/make/deb/arm64/studio*.deb`
3. Install the `.deb` (`sudo dpkg -i studio*.deb`), then from Firefox/Chrome open `wp-studio://test`.
   Expected: the browser launches Studio (no "Open With… / No Apps Available" dialog).
4. Run the full WordPress.com OAuth flow from the **Previews** tab and confirm the token is delivered back and login completes in-app.

### Source-built install

1. Follow the updated **Creating a Desktop Shortcut** steps in `docs/linux.md`, including the new `xdg-mime default studio.desktop x-scheme-handler/wp-studio` step.
2. Open `wp-studio://test` from a browser and confirm Studio is launched.

## Pre-merge Checklist

### DIFF
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -93,6 +93,12 @@ For easier access, you can create a desktop entry file that will add Studio to y
    update-desktop-database ~/.local/share/applications
    ```
 
+6. **Register Studio handler for `wp-studio://` links:**
+   ```bash
+   xdg-mime default studio.desktop x-scheme-handler/wp-studio
+   ```
+   Without this, browsers will still show "Open With… / No Apps Available" when WordPress.com OAuth redirects back to `wp-studio://`.
+
 Studio should now appear in your application launcher and can handle `wp-studio://` protocol URLs.
 
 ## Running with Wayland
@@ -155,4 +161,3 @@ The executable in `apps/studio/out/Studio-linux-x64/` will be updated with the n
 ## Contributing
 
 If you encounter Linux-specific issues or have suggestions for improving Linux support, please [open an issue](https://github.com/Automattic/studio/issues) or submit a pull request. Your feedback helps improve Studio for all Linux users!
-
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -161,3 +161,4 @@ The executable in `apps/studio/out/Studio-linux-x64/` will be updated with the n
 ## Contributing
 
 If you encounter Linux-specific issues or have suggestions for improving Linux support, please [open an issue](https://github.com/Automattic/studio/issues) or submit a pull request. Your feedback helps improve Studio for all Linux users!
+
--- a/apps/studio/forge.config.ts
+++ b/apps/studio/forge.config.ts
@@ -83,6 +83,7 @@ const config: ForgeConfig = {
 				genericName: 'WordPress Studio',
 				categories: [ 'Utility' ],
 				name: 'studio',
+				bin: 'studio',
 				mimeType: [ 'x-scheme-handler/wp-studio' ],
 			},
 		} ),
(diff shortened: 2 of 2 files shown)


docs/linux.md
```diff
@@ -155,4 +161,3 @@ The executable in `apps/studio/out/Studio-linux-x64/` will be updated with the n
 ## Contributing
 
 If you encounter Linux-specific issues or have suggestions for improving Linux support, please [open an issue](https://github.com/Automattic/studio/issues) or submit a pull request. Your feedback helps improve Studio for all Linux users!
-
```
ivan-ottinger, line 158: I think this line removal was accidental.

ivan-ottinger (commented):
The proposed change look great! 👌🏼 

I couldn't make the app built with `npm run make` - similarly to what we discussed in p1776944195041599-slack-C04GESRBWKW.

We could include the related fix in the PR directly:

```diff
diff --git a/apps/studio/forge.config.ts
b/apps/studio/forge.config.ts
--- a/apps/studio/forge.config.ts
+++ b/apps/studio/forge.config.ts
@@ -83,6 +83,7 @@ const config: ForgeConfig = {
                              genericName: 'WordPress Studio'
                              categories: [ 'Utility' ],
                              name: 'studio',
+                             bin: 'studio',
                              mimeType: [ 'x-scheme-handler/w
                      },
              } ),
```

What do you think?

Apart from that I only noticed one small thing (please see the inline comment)

gavande1:
Thanks for the suggestion, Ivan! That's the exact change I needed to get `npm run make` working on my end too — I've added it in 9a9cf7d7. Also restored the trailing newline in `docs/linux.md` that I dropped accidentally. 🙏

docs/linux.md
gavande1: Good catch — restored in 9a9cf7d7. Thank you!

ivan-ottinger (approved):
Thank you for the updates, Rahul!

The changes look good and work as expected. 👍🏼 

I am now able to connect with my WordPress.com account. 🎉 🙂 

One thing that came to my mind is whether moving forward we will need to keep the "## Creating a Desktop Shortcut (Optional)" section in the documentation - since when the user builds their app into .deb, they can just drag the app to their dock.

But that's outside of this PR and we can leave the shortcut section in place for now.

ivan-ottinger:
(added YOLO as a reviewer - just for reference)

PR #3209, opened 2026-04-23
CLI: Add json format for auth status

nightnei:
Fixes STU-1638

## How AI was used in this PR

AI assisted with writing code and ensuring it's aligned with `site status`. We had a few iterations, I tested it, and reviewed it thoroughly.

## Proposed Changes

CLI supports i18n, so it's difficult to parse the output. I propose to add json output to handle it.

**Note:** I used "text | json". We typically use `table` but it doesn't fit in this case, that's why I decided to go with `text` as a universal approach and I would propose to go with `text` instead of `table` in other places in the future, to have consistency with the universal word to distinguish JSON vs human-readable format.

## Testing Instructions
1. Run `node apps/cli/dist/cli/main.mjs auth login`
2. Run `node apps/cli/dist/cli/main.mjs auth status`
3. Assert that no regression (`✔ Authenticated with WordPress.com as `nightnei``)
4. Run `node apps/cli/dist/cli/main.mjs auth status --format=json`
5. Assert you see json response
6. Run `node apps/cli/dist/cli/main.mjs auth logout`
7. Assert that `node apps/cli/dist/cli/main.mjs auth status` and `node apps/cli/dist/cli/main.mjs auth status --format=json` work well.

### DIFF
--- a/apps/cli/commands/auth/status.ts
+++ b/apps/cli/commands/auth/status.ts
@@ -5,23 +5,46 @@ import { getUserInfo } from 'cli/lib/api';
 import { Logger, LoggerError } from 'cli/logger';
 import { StudioArgv } from 'cli/types';
 
-export async function runCommand(): Promise< void > {
+export async function runCommand( format: 'text' | 'json' = 'text' ): Promise< void > {
 	const logger = new Logger< LoggerAction >();
 
-	logger.reportStart( LoggerAction.STATUS_CHECK, __( 'Checking authentication status…' ) );
+	if ( format !== 'json' ) {
+		logger.reportStart( LoggerAction.STATUS_CHECK, __( 'Checking authentication status…' ) );
+	}
 
 	try {
 		const token = await readAuthToken();
 
 		if ( ! token ) {
-			logger.reportError( new LoggerError( __( 'Authentication token is invalid or expired' ) ) );
+			if ( format === 'json' ) {
+				console.log( JSON.stringify( { authenticated: false }, null, 2 ) );
+			} else {
+				logger.reportError( new LoggerError( __( 'Authentication token is invalid or expired' ) ) );
+			}
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


nightnei:
I used `text | json`. We typically use `table` but it doesn't fit in this case, that's why I decided to go with `text` as a universal approach and I would propose to go with `text` instead of `table` in other places in the future, to have consistency with the universal word to distinguish JSON vs human-readable format.
Not critical, but we could also add `text` to existing functions, alongside `table` as  `table | json | text`, to keep backward compatibility.

@fredrikekelund @bcotrim what do you thing about this?

apps/cli/commands/auth/status.ts
```diff
@@ -5,23 +5,46 @@ import { getUserInfo } from 'cli/lib/api';
...
 import { StudioArgv } from 'cli/types';
 
-export async function runCommand(): Promise< void > {
+export async function runCommand( format: 'text' | 'json' = 'text' ): Promise< void > {
 	const logger = new Logger< LoggerAction >();
 
-	logger.reportStart( LoggerAction.STATUS_CHECK, __( 'Checking authentication status…' ) );
+	if ( format !== 'json' ) {
+		logger.reportStart( LoggerAction.STATUS_CHECK, __( 'Checking authentication status…' ) );
+	}
```
fredrikekelund, line 13: As I mentioned in https://github.com/Automattic/studio/pull/3219#issuecomment-4311264380, I think it'd be better to implement a "silent mode" in the logger, so we don't have to wrap every single `Logger` method call in an if statement like this.

fredrikekelund (approved):
This works as expected. I really think we should consider my suggestion of silencing `Logger` output inside `Logger` instead of outside, though.

> we could also add text to existing functions, alongside table as table | json | text, to keep backward compatibility

This sounds like a good approach to me 👍

PR #3207, opened 2026-04-23
apps/cli: two-phase AI build workflow with blockify skill

youknowriad:
## Related issues

- Related to: none (iterative refinement of the AI agent's site-build flow, supersedes the system-prompt portion of #3199)

## How AI was used in this PR

The system-prompt changes and the `blockify` skill in this PR were designed and written by Claude together with the PR author, using an iterative loop of prompt edit → full build session → session-JSONL audit → next edit. Each change cites the behavior it is responding to from a specific session recording. Reviewers should focus on whether the prompt wording is clear and consistent, whether the workflow described is the one we want the agent to follow, and whether the new `blockify` skill is cleanly scoped.

### Local-site build: two-phase workflow

The local-site workflow in the AI agent's system prompt is now split into two explicit phases:

- **PHASE 1 — HTML prototype.** The agent writes plain HTML/CSS/JS under `<site>/tmp/prototype/` using a section-anchor skeleton and fills one anchor per `Edit`. Design tokens are locked in a `tokens` anchor before any section fill. Writes under `wp-content/themes/` and `wp-content/plugins/` are forbidden in this phase. Phase 1 completes on a `take_screenshot` of the prototype.
- **PHASE 2 — Port to block theme.** The agent invokes the new `blockify` skill as a gate, builds the block-theme skeleton, `cp`-s the prototype stylesheet into `<theme>/assets/css/main.css` and adjusts only block-DOM selectors (`.wp-block-button`, `.wp-block-image`) via small `Edit`s, translates each prototype section to block markup in `<site>/tmp/page-<slug>.html`, and applies the content via `wp_cli eval '... file_get_contents(ABSPATH . "tmp/page-<slug>.html") ...'`.

The eval-based application replaces the earlier `--post_content-file=<host path>` pattern, which silently failed because wp_cli runs inside the PHP-WASM filesystem and cannot read host paths. The host site directory is mounted at `/wordpress/` in WASM, so `ABSPATH . "tmp/..."` resolves correctly.

### Remote-site (WordPress.com) build: three-phase workflow

The remote-site workflow was restructured to mirror the local two-phase pattern while keeping an explicit audit step that's only meaningful on remote (plan gating + site discovery):

- **PHASE 1 — Audit.** Plan gate (`GET /`), content audit, active theme, templates, and fetching the global-styles ID for later CSS work.
[...]

PR #3210, opened 2026-04-23
Add prompt-rule regression tests to the agent eval suite

youknowriad:
## How AI was used in this PR

Claude Code drafted the test cases and assertions. Verified locally that the YAML validates cleanly (\`npx promptfoo@0.121.4 validate -c eval/promptfoo.config.yaml\`). The real agent runs have not yet been executed — they need \`npm run eval\` against an authed Studio session.

## Proposed Changes

Adds ten prompt-rule regression cases to the existing agent eval suite at [\`eval/promptfoo.config.yaml\`](eval/promptfoo.config.yaml). Each pins a specific mechanical rule that has silently regressed in the past; together they guard against losing recent prompt evolution.

Each case asks the agent to narrate an answer in prose (no tool execution), so assertions can grep \`d.textSegments.join('\\n')\` for the load-bearing substrings. That keeps the runs fast and self-contained: no real site, no permission loop, no filesystem side effects.

### The ten rules under test

1. **wp-cli-no-shell-syntax** — \`wp_cli\` takes literal args; filter via \`wp_cli eval\`, never pipes / \`$(...)\` / \`&&\` chaining.
2. **phase2-cp-stylesheet** — PHASE 2 theme stylesheet is copied from the prototype with \`cp\`, not regenerated in a \`Write\`.
3. **button-paint-inner-link** — all button paint goes on \`.wp-block-button.<name> .wp-block-button__link\`; the outer wrapper is layout-only.
4. **theme-json-neutralize-button** — \`theme.json\` zeros \`styles.elements.button\` so className rules are the only source of paint.
5. **theme-json-content-widths** — \`settings.layout.contentSize\` / \`wideSize\` match the prototype's intended max-widths.
6. **hero-decompose-svg-only** — a \`<section>\` with an inline SVG decomposes into native blocks; \`core/html\` wraps only the SVG.
7. **card-decompose-native-blocks** — a card \`<div>\` becomes \`core/group\` + \`core/heading\` + \`core/paragraph\` + \`core/buttons\` + \`core/button\`, no \`core/html\`.
8. **apply-content-abspath-eval** — page content is applied via \`wp_cli eval\` + \`ABSPATH\` + \`file_get_contents\`, never \`--post_content-file=<host path>\` (silently no-ops in WASM).
9. **phase1-style-skeleton** — the first prototype \`style.css\` \`Write\` is a <2KB skeleton with \`tokens\` as the first anchor.
10. **phase2-blockify-first** — the first Phase 2 tool call after an approved Phase 1 screenshot is invoking the \`blockify\` skill.

### Why the existing harness

Studio already has an agent eval suite at [\`eval/\`](eval) that runs the real agent via \`startAiAgent()\` and grades via Studio's WP.com-authed proxy. An earlier commit on this branch added a parallel standalone promptfoo setup — that was the wrong call, and it's reverted in [\`4804b783\`](../commit/4804b783) in favor of this integration. The assertion style (parse \`output\` JSON, inspect \`textSegments\` / \`toolCalls\`) matches the existing cases in the suite so reviewers don't have to learn two patterns.

### Scope boundaries

- **Not covered**: multi-turn behavior (fix-loop batching, turns-per-build, session duration), screenshot / visual-fidelity tests, taste / design quality. Those stay in manual session audits, as noted in the original out-of-scope list.
- **Doesn't modify the prompt itself.** This PR only adds tests.

### Prerequisite

The tests are designed against the rules from [#3207](https://github.com/Automattic/studio/pull/3207) (two-phase build + \`blockify\` skill). If #3207 has not yet merged when this PR is evaluated, several cases (2, 4, 5, 8, 9, 10) will fail on this branch because their rules are not yet in \`trunk\`. Recommended order: **merge #3207 first, rebase this PR, then run \`npm run eval\`.**

## Testing Instructions

1. **Validate YAML**:
   \`\`\`sh
   npx promptfoo@0.121.4 validate -c eval/promptfoo.config.yaml
   \`\`\`
   → \"Configuration is valid.\"

2. **Full eval run** (builds the CLI, then runs all 13 cases — 3 existing + 10 new):
   \`\`\`sh
   studio auth login
   npm run eval
   npm run eval:view
   \`\`\`

3. **Single case** (fast iteration):
   \`\`\`sh
   npm run eval -- --filter-pattern \"wp_cli filter uses eval\"
   \`\`\`

4. **Deliberate regression sanity-check**: temporarily remove the \`Do NOT use --post_content-file\` line from \`apps/cli/ai/system-prompt.ts\`, rebuild, rerun the suite — \`apply-content-abspath-eval\` should fail with a clear \`not-contains\` violation.

## Pre-merge Checklist

- [x] \`promptfoo validate\` passes on the updated config.
- [ ] Run \`npm run eval\` on this branch against an authed Studio session to confirm the new cases behave as expected.
- [ ] Parent PR #3207 is merged (or its rules are otherwise present in trunk) before merging.


youknowriad:
I'm merging cause I need this on a separate PR.

PR #3232, opened 2026-04-24
Share the app.json reader between Studio and the CLI

adamziel:
## Related issues

Unblocks #2939 (gating `studio pull-reprint` behind a beta flag read from `~/.studio/app.json`).

## Proposed Changes

Studio's desktop process owns `~/.studio/app.json` — site metadata, window bounds, beta feature toggles, etc. — via `loadUserData()` / `saveUserData()` in `apps/studio/src/storage/user-data.ts`. That module imports `@sentry/electron/main` to attach breadcrumbs on parse failures, which means anything outside Electron's main process (like the CLI) can't reuse it.

When the CLI needs to read the same file — for example, to check a beta flag set from the desktop app — the options are (a) duplicate the reader, (b) drop Electron/Sentry entirely, or (c) split the module.

This PR does (c): lifts the plain JSON read/write into `tools/common/lib/user-data.ts` and turns `apps/studio/src/storage/user-data.ts` into a thin wrapper that:
- re-exports `lockAppdata` / `unlockAppdata` from the shared module,
- keeps the typed `UserData` return,
- injects the Sentry breadcrumb through a new `onInvalidJson` callback the shared reader exposes.

No behavior change for Studio — same file path, same locking, same ENOENT defaults, same breadcrumbs. The CLI can now do `(await loadUserData()).betaFeatures?.pullReprint` from `@studio/common/lib/user-data` without pulling Electron into `tools/common`.

## Testing Instructions

- `npm run typecheck`
- `npx vitest run apps/studio/src/storage/tests/user-data.test.ts tools/common/lib/tests/user-data.test.ts` — 9 tests covering the existing desktop behavior plus a new integration test suite in `tools/common` (defaults on missing file, version normalization, beta features round-trip, `onInvalidJson` callback on malformed JSON, save/load round-trip).
- Launch the desktop app and confirm `~/.studio/app.json` is still read/written normally (create a site, toggle a setting, relaunch).

## Pre-merge Checklist

- [x] Tests added / updated
- [x] Typecheck clean (pre-existing `ignore` package errors unrelated to this change)
- [x] No behavior change for Studio
- [ ] Reviewed by @wojtekn and @sejas

### DIFF
--- a/apps/studio/src/storage/tests/user-data.test.ts
+++ b/apps/studio/src/storage/tests/user-data.test.ts
@@ -9,18 +9,22 @@ import { vi } from 'vitest';
 import { loadUserData, lockAppdata, unlockAppdata, saveUserData } from 'src/storage/user-data';
 import { UserData } from '../storage-types';
 
-const { getUserDataFilePathMock, getAppConfigLockFilePathMock } = vi.hoisted( () => {
-	return {
-		getUserDataFilePathMock: vi.fn().mockReturnValue( '/path/to/app/.studio/app.json' ),
-		getAppConfigLockFilePathMock: vi.fn().mockReturnValue( '/path/to/app/.studio/app.json.lock' ),
-	};
-} );
+const { getUserDataFilePathMock, getAppConfigPathMock, getAppConfigLockFilePathMock } = vi.hoisted(
+	() => {
+		return {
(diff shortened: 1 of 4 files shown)


adamziel:
I'll just put it behind an env variable

PR #3233, opened 2026-04-24
MVP: Run Studio sites using native PHP binary

fredrikekelund:
## Related issues

- Fixes RSM-1149

## How AI was used in this PR

Extensively, but with quite precise instructions.

## Proposed Changes

- Make `php-server-child.ts` spawn a native PHP process that runs the built-in web server with a custom routing script.
- With this architecture, the process manager daemon spawns `php-server-child.ts`, which in turn spawns a `php` process, meaning we have one additional layer compared to the Playground model.
- stdout and stderr from the PHP process are forwarded to the respective stdio stream on the `php-server-child.ts` process (which is then picked up by the process manager daemon and written to a log file).
- When the `php` process dies, so does `php-server-child.ts`, and when `php-server-child.ts` dies, so does the `php` process.

Here are some quick initial benchmarks comparing the native built-in PHP 8.5 web server (no opcache) with 8 and 1 workers, respectively, to a Studio CLI site on `trunk`. The `siteEditorLoad` and `templateOpen` metrics are especially interesting to us, and they show a healthy improvement.

```
Results Comparison
════════════════════════════════════════════════════════
Metric              8-native-php   1-native-php   studio
────────────────────────────────────────────────────────
blockAdd            2.13s          1.96s          1.56s
siteEditorLoad      858ms          872ms          2.58s
templateOpen        892ms          733ms          2.34s
templateSave        347ms          598ms          712ms
templatesViewLoad   376ms          437ms          457ms
════════════════════════════════════════════════════════
Total               4.63s          4.6s           7.65s
```

## Testing Instructions

1. `npm run download:php-binary`
2. `npm run cli:build`
3. `STUDIO_RUNTIME=native-php node apps/cli/dist/cli/main.mjs site create` and go through the site creation process
4. Ensure that the site is created successfully and opens in the browser

## Pre-merge Checklist


wojtekn:
Improvements look great, but do you have an idea why blockAdd is slower?

fredrikekelund:
> why blockAdd is slower

Not yet, no. I'm assuming it's some kind of lazy loading that behaves differently in the two environments.

We've previously focused primarily on the total time for the benchmark suite relative to other environments, but I think it's actually misleading to treat all metrics equally. `siteEditorLoad` and `templateOpen` are more relevant to perceived speed on the site, IMO.

That said, we plan to look more closely at the difference in `blockAdd` later. I'll share more on this once we've done that 👍

scripts/download-php-binary.ts
fredrikekelund, line 1: The changes in this file are just auto-formatting

apps/cli/php-server-child.ts
```diff
@@ -16,13 +44,334 @@ function errorToConsole( ...args: Parameters< typeof console.error > ) {
...
+				} );
+				stopSignal.addEventListener( 'abort', () => {
+					reject( new DOMException( 'Aborted', 'AbortError' ) );
+				} );
+			} );
+
+			stopSignal.throwIfAborted();
+
+			// There's a brief delay between the PHP process starting and when it accepts connections
+			await new Promise< void >( ( resolve ) => setTimeout( resolve, 500 ) );
```
bcotrim: We should poll the server URL until it replies rather than using an arbitrary delay. I'm guessing the WordPress installation will fail if the server isn't ready so we should make sure it is.
It's OK for this initial version.

apps/cli/php-server-child.ts
```diff
@@ -16,13 +44,334 @@ function errorToConsole( ...args: Parameters< typeof console.error > ) {
...
+
+	const installResponse = await fetch(
+		`http://localhost:${ config.port }/wp-admin/install.php?step=2`,
+		{
+			method: 'POST',
+			headers: {
+				'Content-Type': 'application/x-www-form-urlencoded',
+			},
+			body: new URLSearchParams( {
+				language: 'en',
```
bcotrim: This should use the site's locale from the config rather than hardcoding 'en'. 
Better to tackle in a follow-up issue but let's track it.

apps/cli/php-server-child.ts
```diff
@@ -16,13 +44,334 @@ function errorToConsole( ...args: Parameters< typeof console.error > ) {
...
+	} catch ( error ) {
+		throw new Error(
+			`Failed to set default permalinks: ${
+				error instanceof Error ? error.message : String( error )
+			}`
+		);
+	}
+}
+
+const startServer = wrapWithStartingPromise(
```
bcotrim: Nit: I would write it something like this, just for ease of reading:

```
async function _startServer( config: ServerConfig, signal: AbortSignal ): Promise< void > {
...
}

async function startServer( config: ServerConfig, signal: AbortSignal ): Promise< void > {
	if ( startingPromise ) {
		return startingPromise;
	}
	startingPromise = _startServer( config, signal );
	return startingPromise;
}
```

bcotrim (approved):
Solid start, LGTM 👍 
Left a few minor comments but no blockers, shipping this is important so we can start testing and working around issues.

apps/cli/php-server-child.ts
fredrikekelund: Good catch 👍

PR #3234, opened 2026-04-24
CLI: dev-only PHP binary fetch script for native-php runtime

bcotrim:
## Related issues

- Related to 

## How AI was used in this PR

Built with Claude Code, reviewed manually. Scope was intentionally narrow: only the binary delivery mechanism for local development.

## Proposed Changes

First step of the native-php runtime implementation: a script to download the static PHP binary from [static-php-cli](https://github.com/crazywhalecc/static-php-cli)'s CDN for local dev use. **PHP binaries are not bundled in production builds** — this is dev tooling only.

- **`scripts/download-php-binary.ts`** — downloads PHP 8.4.20 from `dl.static-php.dev` for the host platform/arch:
  - macOS + Linux → `bulk` variant (includes `pdo_sqlite`, `intl`, `opcache`, `sodium`, `imagick`)
  - Windows x64 → `windows/spc-max` variant (includes `pdo_sqlite` + `opcache`; missing `intl`/`sodium`/`imagick`)
  - Maps Node arch names to CDN names (`arm64` → `aarch64`, `x64` → `x86_64`)
  - Windows ARM64 falls back to x64 (no upstream ARM64 binary; runs under OS emulation)
  - SHA-256 verification with hashes pinned for all 5 platform archives
  - Skips download if binary already present
  - Failure is non-fatal in `postinstall` — prints a warning and lets `npm install` finish
  - Args parsed with `z.tuple` + `.default()` for consistency with the rest of the codebase

- **`apps/cli/lib/dependency-management/paths.ts`** — adds `getPhpBinaryPath()` resolving to `~/.studio/php-bin/php[.exe]` via `getConfigDirectory()`, consistent with how Studio locates all other downloaded/generated content

- **`package.json`** — adds `download:php-binary` npm script and wires it into `postinstall` so `npm install` provisions the binary alongside wp-server-files, translations, and agent skills

## Future considerations

We're consuming static-php-cli's pre-built binaries directly from their CDN. This is the simplest path for v1 but has known tradeoffs:

- **Upstream URL stability**: static-php-cli's CDN structure has shifted before (`spc-max` → `windows/spc-max`), so URL drift is a real risk.
- **No upstream checksums**: SHA-256 has to be pinned manually after each version bump.
- **Windows extension gap**: Windows `spc-max` is missing `intl`/`sodium`/`imagick` (macOS/Linux `bulk` has all of these).

The escalation path is to build our own variant via [static-php-cli's `spc` tool](https://github.com/crazywhalecc/static-php-cli) and host the resulting binaries somewhere Automattic-controlled (e.g. WP.com CDN, S3, or GitHub releases). Build time is ~15–30 min per arch on with `--prefer-pre-built`, and they publish a reusable GitHub Actions workflow we can fork. This would give us reproducible URLs + hashes, close the Windows extension gap, and make it easier to ship multiple PHP versions side-by-side later. Out of scope for v1 — would need apps-infra sync for hosting.

## Testing Instructions

```sh
# Download the binary (one-time, ~14 MB compressed → ~92 MB on disk for bulk)
npm run download:php-binary
# Binary lands at ~/.studio/php-bin/php

# Verify it runs
~/.studio/php-bin/php --version

# Verify required extensions are present
~/.studio/php-bin/php -m | grep -iE "intl|opcache|imagick|sodium|gd|sqlite|pdo"
```

Prod builds are unaffected — no binary is downloaded or bundled by the build pipeline.

## Pre-merge Checklist


scripts/download-php-binary.ts
```diff
@@ -0,0 +1,223 @@
...
+// SHA-256 hashes for each platform asset from dl.static-php.dev.
+// Run `sha256sum <downloaded-file>` after first download and populate these before committing.
+// Until populated, the script will warn and skip verification.
+const KNOWN_HASHES: Record< string, string > = {
+	// 'darwin-arm64': '<sha256>',
+	// 'darwin-x64':   '<sha256>',
+	// 'linux-x64':    '<sha256>',
+	// 'linux-arm64':  '<sha256>',
+	// 'win32-x64':    '<sha256>',
+};
```
fredrikekelund, line 41: We should download these and commit the hashes, correct?

scripts/download-php-binary.ts
```diff
@@ -0,0 +1,223 @@
...
+
+type Platform = z.infer< typeof platformSchema >;
+type Arch = z.infer< typeof archSchema >;
+
+const argsSchema = z.tuple( [
+	platformSchema.default( process.platform as Platform ),
+	archSchema.default( process.arch as Arch ),
+] );
+
+const [ platform, arch ] = argsSchema.parse( [ process.argv[ 2 ], process.argv[ 3 ] ] );
```
fredrikekelund:
```suggestion
const args = argsSchema.parse( [ process.argv[ 2 ], process.argv[ 3 ] ] );
```
Nitpicking, but I think this would make it easier to keep track of what `platform` and `arch` refer to.

scripts/download-php-binary.ts
```diff
@@ -0,0 +1,223 @@
...
+		console.log(
+			`\nPHP binary installed: ${ destPath } (${ ( stats.size / 1024 / 1024 ).toFixed( 1 ) } MB)`
+		);
+	} catch ( error ) {
+		console.warn( `Warning: PHP binary download failed — ${ ( error as Error ).message }` );
+		console.warn( 'The native-php runtime will not be available. Run `npm run download:php-binary` to retry.' );
+	}
+}
+
+main();
```
fredrikekelund:
```suggestion
void main();
```

scripts/download-php-binary.ts
```diff
@@ -0,0 +1,223 @@
...
+async function extractBinary( archivePath: string ): Promise< void > {
+	console.log( 'Extracting PHP binary...' );
+
+	if ( isWindows ) {
+		const extractDir = path.join( tmpDir, `php-${ PHP_VERSION }-win` );
+		await extractZip( archivePath, tmpDir );
+		// The zip contains php.exe at the root
+		const src = path.join( extractDir, 'php.exe' );
+		if ( ! fs.existsSync( src ) ) {
+			// Some builds place it directly in the archive root without a subdirectory
```
fredrikekelund: Which other builds? 🤔

scripts/download-php-binary.ts
```diff
@@ -0,0 +1,223 @@
...
+const isWindows = platform === 'win32';
+const ext = isWindows ? 'zip' : 'tar.gz';
+const cdnArch = cdnArchMap[ effectiveArch ];
+const filename = isWindows
+	? `php-${ PHP_VERSION }-cli-win.${ ext }`
+	: `php-${ PHP_VERSION }-cli-${ osSegment[ platform ] }-${ cdnArch }.${ ext }`;
+
+const cdnBase = isWindows
+	? 'https://dl.static-php.dev/static-php-cli/windows/spc-max'
+	: 'https://dl.static-php.dev/static-php-cli/common';
```
fredrikekelund:
```suggestion
	: 'https://dl.static-php.dev/static-php-cli/bulk';
```
Not having `intl` or `opcache` seems like a deal breaker. I would love for `imagick` to be included, too (which [it is](https://static-php-cli.fra1.digitaloceanspaces.com/static-php-cli/bulk/README.txt) in this build)

scripts/download-php-binary.ts
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	await new Promise< void >( ( resolve, reject ) => {
+		file.on( 'finish', resolve );
+		file.on( 'error', reject );
+		file.end();
+	} );
+
+	process.stdout.write( '\n' );
+	console.log( 'Download complete.' );
```
fredrikekelund:
```suggestion
	await pipeline(
		Readable.fromWeb( response.body! as NodeReadableStream< Uint8Array > ),
		fs.createWriteStream( dest )
	);

	console.log( 'Download complete.' );
```
Optional nit, but I would opt to simplify this dance and just stream the response to disk instead of outputting a progress message in the terminal.

fredrikekelund (approved):
Fundamentally really good, except I think we should use the `bulk` build for macOS, and we should add the precomputed sha256 hashes.

scripts/download-php-binary.ts
bcotrim: Nice catch, was a leftover from some earlier explorations

PR #3241, opened 2026-04-24
apps/cli: fix wp_cli post content apply failing silently in WASM

youknowriad:
## Related issues

None. Follows up on the system-prompt portion of #3199; not a revert — a focused fix for one bug surfaced in agent session recordings.

## How AI was used in this PR

The AI agent's own session recordings surfaced this bug (an 18-turn debugging spiral over empty post content on FamilyFarm build). The fix was developed and validated by iterating over session audits to confirm that (a) `wp_cli` runs inside PHP-WASM, (b) host-absolute paths silently resolve to empty content, and (c) `ABSPATH === "/wordpress/"` maps back to the mounted host site directory.

The regression test assertion was also drafted alongside the fix so the bug can't silently creep back via prompt edits.

## Proposed Changes

- Replace the `wp_cli post update --post_content-file=<host path>` apply pattern in the Page-content cadence with `wp_cli eval` + `file_get_contents(ABSPATH . "tmp/page-<slug>.html")`. Host paths do not resolve inside PHP-WASM — `ABSPATH` (== `"/wordpress/"`) does, and the host site directory is mounted there.
- Move the scratch page-content file out of the theme folder into `<site>/tmp/page-<slug>.html` so transient authoring artefacts don't pollute exports / git.
- Add a one-line "why" note in the prompt so future edits don't revert the fix.
- Extend the agent eval runner to capture tool-call `input`, and add a regression assertion on the existing `single-page-build-turn-cadence` test that fails if any `wp_cli` call uses `--post_content-file=`.

## Testing Instructions

1. `npm run cli:build`
2. `npm run typecheck` (CLI workspace clean)
3. `npm test -- apps/cli/ai` — 110 tests should pass.
4. `node apps/cli/dist/cli/main.mjs ai` → prompt: "Build a farm one page site called MyFarm. Use elegant colors." Verify the tool-use trace:
   - `<site>/tmp/page-<slug>.html` is created (NOT `<theme>/page-content.html`).
   - The apply step is `wp_cli eval '$content = file_get_contents(ABSPATH . "tmp/page-<slug>.html"); wp_update_post([...]);'` — not `--post_content-file=`.
   - The final landing page has real content (not empty).
   - The agent does not spiral into a 15-20 turn debugging loop over empty post content.
5. (Optional) `npm run eval -- -n 1 --filter-description single-page-build-turn-cadence` to run the regression-assertion locally.

## Pre-merge Checklist

- [x] TypeScript clean (CLI workspace).
- [x] `npm test -- apps/cli/ai` passes (110/110).
- [x] Manual AI-agent build confirms the empty-post regression is gone.
- [x] Eval regression assertion added; would fail the single-page build test if the broken pattern re-enters the prompt.

### DIFF
--- a/apps/cli/ai/eval-runner.ts
+++ b/apps/cli/ai/eval-runner.ts
@@ -57,9 +57,10 @@ function extractToolCalls( message: SDKMessage ) {
 			} ): block is { type: 'tool_use'; id: string; name: string; input: unknown } =>
 				block.type === 'tool_use'
 		)
-		.map( ( block: { id: string; name: string } ) => ( {
+		.map( ( block: { id: string; name: string; input: unknown } ) => ( {
 			id: block.id,
 			name: normalizeToolName( block.name ),
+			input: block.input,
 		} ) );
 }
 
@@ -140,7 +141,7 @@ async function runEval( input: EvalRunnerInput ) {
 	// Allow running inside a Claude Code session
 	delete env.CLAUDECODE;
 
-	const toolCalls: { id: string; name: string }[] = [];
+	const toolCalls: { id: string; name: string; input: unknown }[] = [];
 	const toolResults: {
 		toolUseId: string | null;
 		toolName: string | null;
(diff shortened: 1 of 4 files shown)


youknowriad:
This is really cool. I was able to confirm the eval test fails in trunk and is solved in this PR :)

PR #3030, opened 2026-04-09
Add visual annotation browser with Agentation

lezama:
## Summary

Adds `/annotate` skill to Studio Code. Users type `/annotate`, a browser opens with the Agentation annotation toolbar, they click elements and leave feedback, and the agent reads those annotations and makes changes.

Tested end-to-end: user annotated "make this red", "change this to H2" → agent read annotations, applied changes via Global Styles custom CSS and WP-CLI.

Alternative approach to the custom element picker in #2936 — instead of building a custom UI, this injects [Agentation](https://www.agentation.com)'s React toolbar into a Playwright browser using esm.sh CDN with `?deps=` pinning for a single React instance.

## How it works

1. `agentation-mcp` bundled as dependency — launches HTTP (localhost:4747) + MCP servers
2. `/annotate` skill clears old annotations, opens headed Playwright browser via `open_annotation_browser` tool
3. Agentation React component injected via dynamic `import()` from esm.sh CDN
4. User clicks elements, adds feedback in the browser
5. Agent reads annotations with `agentation_get_all_pending` (CSS selectors, computed styles, text)
6. Agent makes WordPress changes (Global Styles CSS, template overrides, WP-CLI) and resolves annotations

## Changes

- **`apps/cli/ai/agentation-inject.ts`** (new) — Opens headed Playwright browser, injects Agentation via esm.sh dynamic imports
- **`apps/cli/ai/tools.ts`** — New `open_annotation_browser` tool
- **`apps/cli/ai/plugin/skills/annotate/SKILL.md`** (new) — `/annotate` skill with guided workflow and WordPress-native change patterns
- **`apps/cli/ai/slash-commands.ts`** — Register `/annotate` command
- **`apps/cli/ai/agent.ts`** — Agentation MCP server added alongside Studio tools
- **`apps/cli/package.json`** — `agentation-mcp` dependency

## Testing

1. `npm install && npm run cli:build`
2. `node apps/cli/dist/cli/main.mjs code`
3. Select a running site, type `/annotate`
4. Click the circle icon (bottom-right), click elements, add feedback
5. Type "done" — agent reads and applies changes

## Quality gates

- [x] ESLint: pass
- [x] TypeScript: pass
- [x] Tests: 72/72 pass
- [x] End-to-end: annotations synced and applied to WordPress site

PR #3243, opened 2026-04-25
Add /annotate skill with Studio inspector

youknowriad:
## Related issues

- Builds on #3030 (the original /annotate skill PR by @lezama). This PR rebases that work onto current `trunk` and replaces the third-party agentation toolbar with a self-contained Studio inspector. The first commit on this branch is the original @lezama commit, preserved for attribution; the second commit is the rewrite.

## How AI was used in this PR

Heavy AI involvement — Claude Code drove the rebase, the discovery of the issues with the agentation-based approach (window sizing, dock-icon leak, sync gaps, missing confirmation step), and the rewrite into a vanilla-DOM inspector. End-to-end behaviours were verified with Playwright scripts; screenshots from each iteration are in the session transcript.

Reviewers should focus on:
- The injected page script in `apps/cli/ai/inspector/page-script.ts` — it runs in the host WordPress page so it needs to be defensive against any theme CSS / global handlers.
- The chromium lifecycle in `apps/cli/ai/inspector/inspector-inject.ts` — close-on-page-close + close-on-process-exit hooks are the bits I'd most want a second pair of eyes on.
- The mandatory "summarize and confirm" step in `SKILL.md`. The intent is that the agent never edits files based on annotations without an explicit yes from the user.

## Proposed Changes

This PR is a Proof-of-Concept-grade rework of the annotation skill. **Marking as draft + Proof of Concept label** per AGENTS.md guidance for substantial AI-generated features.

**Removed**
- `agentation` npm package dependency (loaded at runtime from `esm.sh` in #3030)
- `agentation-mcp` dev dependency
- `agentation` MCP server entry in `apps/cli/ai/agent.ts`
- `apps/cli/ai/agentation-inject.ts`

**Added — `apps/cli/ai/inspector/`**
- `page-script.ts` — self-contained vanilla-DOM inspector mounted into a Shadow DOM root. No React, no `esm.sh`, no external runtime deps.
- `inspector-inject.ts` — Playwright launch + inject + `waitForAnnotationsDone` helper.

**Inspector UI (vs. agentation toolbar in #3030)**
- Two-button toolbar: `Annotate` and `Done`, plus a count badge. Layout-mode, settings page, theme picker, colour palette all gone.
- `Annotate` is sticky — after `Save`, picking mode stays on so the user can chain annotations.
- `Done` ships the batch to the CLI via `window.__studioAnnotateDone`, clears local state, plays a live 10-second countdown toast, then the CLI auto-closes the chromium window.
- Picked elements get a blue outline; saved annotations get numbered markers; clicking a marker reopens the popup for edit/delete.
- `Esc` cancels picking / closes the popup.

**CLI lifecycle hardening**
- 1280×800 window + `viewport: null` so the bottom-right toolbar is always inside the visible region regardless of host screen size.
- `bringToFront()` so the macOS window doesn't open behind the terminal.
- `page.on('close')` and `browser.on('disconnected')` tear down the whole chromium process tree.
- `process.once('exit'|'SIGINT'|'SIGTERM'|'SIGHUP')` hooks so quitting `studio ai` doesn't orphan a dock icon.
- 10-second deferred close after `Done` so the user has time to register the confirmation before the window vanishes.

**Skill flow (`SKILL.md`)**
- After `wait_for_annotations` returns, the agent **must** print a numbered summary in plain language and call `AskUserQuestion` to confirm before making any edits.
- Each summary item identifies the element by tag + `nearbyText` (not selector), restates the intent in the agent's own words, and quotes the user's verbatim comment.
- For another annotation round, the user re-runs `/annotate` rather than reusing the (now-closed) window.

## Testing Instructions

1. `npm install && npm run cli:build`
2. `node apps/cli/dist/cli/main.mjs ai`
3. In the chat, run `/annotate` against any local Studio site.
4. Verify the bottom-right `Annotate` / `Done` toolbar appears in the headed chromium window. Click `Annotate`, click any element on the page, type feedback, click `Save`. Picking mode should stay on — click another element and repeat without re-clicking `Annotate`.
5. Click `Done`. Confirm:
   - A toast counts down `10s … 1s … closing now…`.
   - The chromium window closes automatically.
   - The CLI prints a numbered summary (in plain language, with the verbatim comment in parens) and asks for confirmation before making any edits.
6. Confirm the changes get applied. Try `/annotate` again — a fresh window should open with no leftover annotations from the previous session.
7. Close the chromium window manually mid-session — the dock icon should disappear cleanly (no force-quit needed).
8. Quit `studio ai` (Ctrl-C) while the inspector window is open — the chromium process should exit too.

## Pre-merge Checklist

- [x] No TypeScript / lint errors (`npm run typecheck --workspace apps/cli`, `npx eslint --fix` on the modified files).
- [x] CLI builds (`npm run cli:build`).
- [ ] Unit/E2E tests — none added in this PR. The interaction surface lives entirely inside an injected vanilla-DOM script in a headed chromium window; reviewers may want to push for Playwright coverage before landing.
- [ ] Cross-platform (Windows/Linux) verification — only manually tested on macOS.
- [ ] PolyForm-Shield licence concerns from the agentation package are gone, but reviewers should confirm there are no other lingering references.


apps/cli/ai/inspector/page-script.ts
```diff
@@ -0,0 +1,607 @@
...
+ * external imports — written this way so injection is reliable on any
+ * WordPress page regardless of CSP, network access, or theme CSS.
+ *
+ * Storage: a single localStorage key `studio-inspector-annotations-v1`,
+ * scoped per pathname inside the JSON, so refreshing the page or navigating
+ * between site URLs doesn't lose work.
+ *
+ * Hand-off to the agent: clicking "Done" sets `window.__studioAnnotateDone`,
+ * which `waitForAnnotationsDone()` in inspector-inject.ts polls for.
+ */
```
youknowriad, line 15: If this one becomes more complex we could consider extracting it to its own "react app" in the apps folder.

apps/cli/ai/inspector/page-script.ts
```diff
@@ -0,0 +1,607 @@
...
+					.slice( 0, 3 )
+					.map( ( c ) => '.' + CSS.escape( c ) )
+					.join( '' );
+				part += classes;
+			}
+			const parent = node.parentElement;
+			if ( parent ) {
+				const sameTagSiblings = Array.from( parent.children ).filter(
+					( c ) => c.tagName === node.tagName
+				);
```
borkweb, line 74: The host element is `<div id="__studio-inspector-host">`, appended to `<body>`. When the user clicks any top-level <div> child of `<body>`, `parent.children` includes the host, so nth-of-type is off-by-one. The agent looks up the wrong element. This fires on essentially every full-width hero/section on a typical WP theme.


```suggestion
				const sameTagSiblings = Array.from( parent.children ).filter(
					( c ) => c.tagName === node.tagName && c.id !== HOST_ID

				);
```

apps/cli/ai/tools.ts
```diff
@@ -1057,6 +1058,57 @@ const exportSiteTool = tool(
...
+
+const waitForAnnotationsTool = tool(
+	'wait_for_annotations',
+	'Blocks until the user clicks "Done" in the annotation inspector toolbar. ' +
+		'Returns the annotations the user wrote, captured straight from the page. ' +
+		'Call this AFTER `open_annotation_browser`.',
+	{
+		timeoutMinutes: z
+			.number()
+			.optional()
```
borkweb, line 1100: `timeoutMinutes: 0` disables timeout completely. If we want to constrain it, we could do:


```suggestion
			.number()
			.int()
			.min( 1 )
			.max( 120 )
			.optional()
```

The `120` caps it at 2 hours.

borkweb (changes requested):
A couple of small changes

apps/cli/ai/inspector/page-script.ts
youknowriad: Good shout — agreed. Right now the inspector is ~330 lines of vanilla DOM in a Shadow root, which is light enough to live as an injected string. If we start adding things like keyboard navigation, an annotations panel, or theme variants, hoisting it to a proper `apps/inspector` React app (built and shipped as a static bundle the CLI loads as a string) would be the natural next step.

apps/cli/ai/inspector/page-script.ts
youknowriad: Good catch, applied in 99d7d561. Filtered the host element out of `sameTagSiblings`, so we never count our own injected scaffolding when generating `:nth-of-type` indices.

apps/cli/ai/tools.ts
youknowriad: Yep, real footgun — `timeout: 0` in Playwright is "no timeout, block forever" rather than "don't wait". Applied in 99d7d561 with `.int().min(1).max(120).optional()`.

borkweb (approved):

PR #3244, opened 2026-04-25
Add OpenAI/GPT model support to the Studio code agent

youknowriad:
## Related issues

- Related to # 

## How AI was used in this PR

This is a substantial AI-generated exploration. Per AGENTS.md guidance for vibe-coded features, opening as **draft / Proof of Concept** — the architecture and ergonomics are the contribution; tests, polish, and review-driven refactoring are still owed before this could land in trunk.

## Proposed Changes

Adds OpenAI/GPT model support behind the existing WordPress.com provider. Currently exposes **GPT 5.5** alongside the existing Claude models in `/model` and the desktop UI dropdown.

**Architecture**
- Splits `apps/cli/ai/runtimes/` into per-family subfolders: `anthropic/` (Claude Agent SDK) and `openai/` (pi-agent-core + pi-ai). Dispatch in `apps/cli/ai/agent.ts` routes by `getAiModelFamily(model)` so callers don't pick a runtime — picking a model id picks the runtime.
- The OpenAI runtime targets the wpcom AI proxy at `/v1/chat/completions` using pi-ai's `Model<'openai-completions'>`. Custom `baseUrl` + `headers` + `apiKey` are first-class on the Model object, so the proxy slots in cleanly.
- Pro / o-series variants (`*-pro`, `o[1-9]*`) are intentionally *not* exposed yet — they require `/v1/responses` and a longer timeout than our SDK / proxy currently allow. `tools/common/ai/models.ts` carries a comment documenting the trade-off.

**Tools**
- `apps/cli/ai/tools/` is reorganized one-file-per-tool. The OpenAI runtime adapts each tool's zod input-schema to typebox via `pi-tool-adapter.ts` and registers them on pi's `Agent`. Pi's native `Read/Write/Edit/Bash/Grep/Glob/Ls` are mounted under those names so the system prompt is runtime-agnostic.
- `AskUserQuestion` is implemented as a Studio tool factory (`apps/cli/ai/tools/ask-user-question.ts`) registered only when an `onAskUser` callback is provided — the Anthropic side keeps using the SDK preset.

**Session persistence + compaction**
- `apps/cli/ai/runtimes/openai/persistence.ts`: sidecar transcript file (`<jsonl-stem>.openai-state.json`) keeps pi's in-memory `Agent` state alive across the per-turn forks the desktop UI does. Loads on cold start, atomic-writes after each turn.
- `apps/cli/ai/runtimes/openai/compaction.ts`: `transformContext` callback estimates running tokens and summarizes older messages via pi-coding-agent's `generateSummary` helper when context exceeds the threshold. Preserves recent tail at user-message boundaries so tool_use / tool_result pairs stay intact.
- `tools/common/ai/sessions/store.ts` `deleteAiSession` now sweeps any sibling files sharing the JSONL stem, so sidecars don't outlive their session.

**UI**
- Cross-family `/model` swap in apps/ui shows a confirmation dialog and creates a fresh session on confirm — runtimes don't share transcripts, so silently continuing the same JSONL would make the on-screen history disagree with the agent's actual memory.
- The CLI `/model` slash command's matcher was rewritten to use an explicit `label → id` map; the previous `selectedLabel.startsWith(label)` was vulnerable to prefix collisions.
- Desktop now also auto-selects a site after the agent calls `site_create` regardless of the runtime that issued it (`mcp__studio__site_create` from Claude SDK and bare `site_create` from pi both match).

**Bugs fixed during the build-out**
- `getOrCreateAgent` cached the `Agent` per session id forever. Switching models with `/model` mid-session silently kept the original model. Now rebuilds the Agent when the model id changes, preserving the transcript.
- Cross-runtime `resume` ids: Anthropic SDK rejected ids issued by the OpenAI runtime. `agent.ts` now drops the `resume` hint when crossing families so the new runtime starts fresh.

## Testing Instructions

1. `npm run cli:build && node apps/cli/dist/cli/main.mjs code` and pick **GPT 5.5** via `/model`. Send a multi-turn prompt that exercises tools (e.g. "create a site called Foo and add a hero"). Verify each tool call runs and the conversation stays coherent across turns.
2. With both Sonnet 4.6 and GPT 5.5 available, swap mid-session: confirm the new model is actually used (request hits the right endpoint) and the prior transcript is preserved.
3. In apps/ui, pick a model from the composer dropdown across families (Sonnet → GPT 5.5). Confirm the dialog appears, "Cancel" reverts the dropdown, "Start new conversation" creates a fresh session and navigates to it.
4. Empty session — pick a different family before sending the first message: the dialog should NOT appear (nothing to lose).
5. Long session — let GPT-5.5 run enough turns that the running token estimate exceeds threshold; verify the next request includes a single summary message at the head and recent turns intact at the tail.
6. Delete a GPT-backed session: confirm the `.openai-state.json` sidecar is also removed.

## Pre-merge Checklist

- [ ] Have you tested the changes in the desktop UI and CLI separately?
- [ ] Has the wpcom AI proxy been confirmed to support `studio-assistant` for OpenAI traffic on the same access token? (Already does — `studio-assistant` was historically wired for OpenAI.)
- [ ] Removed `gpt-*-pro` re-enablement requires either the Responses-API path or a server-side timeout extension; tracked separately if/when needed.


apps/ui/src/components/session-view/index.tsx
```diff
@@ -224,48 +291,116 @@ export function SessionView( { sessionId }: { sessionId: string } ) {
...
+				pendingModel={ pendingFamilyChange }
+				inFlight={ familySwitchInFlight }
+				onCancel={ cancelFamilyChange }
+				onConfirm={ () => void confirmFamilyChange() }
+			/>
+		</>
+	);
+}
+
+function FamilySwitchConfirmDialog( {
```
youknowriad: I wonder if this is better as part of the composer component, the model switcher is there.

tools/common/ai/models.ts
```diff
@@ -14,6 +21,15 @@ export function isAiModelId( value: string ): value is AiModelId {
 	return Object.prototype.hasOwnProperty.call( AI_MODELS, value );
 }
 
+export type AiModelFamily = 'anthropic' | 'openai';
+
+export function getAiModelFamily( model: AiModelId ): AiModelFamily {
```
youknowriad: To avoid this function based on random strings, maybe the AI_MODELS const should be an array with id, label and family keys?

apps/cli/ai/tools/utils.ts
```diff
@@ -0,0 +1,226 @@
+import path from 'path';
+import { type ValidationReport } from 'cli/ai/block-validator';
+import { readCliConfig, type SiteData } from 'cli/lib/cli-config/core';
+import { getSiteByFolder } from 'cli/lib/cli-config/sites';
+import { getProgressCallback, setProgressCallback } from 'cli/logger';
```
youknowriad, line 4: I wonder if this file is containing utils that are misplaced and might better moved to where they are actually needed.

youknowriad:
Addressed all three in 09269d5e:

**1. Composer owns the dialog now.** `FamilySwitchConfirmDialog` extracted to `composer/family-switch-confirm-dialog.tsx`; the composer holds the pending-pick state and creates the fresh session inline. `SessionView` no longer knows about the dialog at all — it just hands `events` and `ownerSiteId` down. `apps/ui/src/components/session-view/index.tsx` shrank ~60 lines as a result.

**2. AI_MODELS is now `{ id, label, family }[]`.** `getAiModelFamily` is a Map lookup, `getAiModelLabel` replaces every `AI_MODELS[id]` access (added re-export from `cli/ai/agent.ts` for consumers that imported through there), and `Object.entries(AI_MODELS)` becomes plain `AI_MODELS.map(({id, label}) => …)`. Adding a new model is a one-line data change; getting `family` wrong becomes a TS error instead of silently routing to the Anthropic runtime.

**3. `apps/cli/ai/tools/utils.ts` split.**
- `splitCommandArgs` (+ private helpers `splitBasicCommandArgs`, `stripMatchingOuterQuotes`) → moved into `wp-cli.ts` as private. The heuristics are tightly coupled to WP-CLI's parser quirks, no other caller.
- `formatInvalidBlocks` → moved into `validate-blocks.ts` as private.
- `runPreviewCommand` → extracted to a new `apps/cli/ai/tools/preview-helpers.ts`. The four preview tools (`create`, `update`, `delete`, `list`) share it; keeps the preview domain self-contained instead of pooling with general utils.
- What stays in `utils.ts`: `errorResult`, `textResult`, `resolveSite`, `captureCommandOutput`, `captureConsoleOutput` — genuinely shared.

Verified: typecheck clean, 1396 tests pass, CLI builds, lint clean across the touched scope.

youknowriad:
Closing this one, starting with less disruptive #3328

PR #3253, opened 2026-04-27
Support running WP-CLI commands with native PHP runtime

fredrikekelund:
## Related issues

- Fixes RSM-1157

## How AI was used in this PR

I used Codex extensively for the implementation and both Codex and Claude for review.

## Proposed Changes

- Change the `runWpCliCommand()` function to switch between PHP-WASM-powered and native PHP-powered WP-CLI running depending on the `site.runtime` argument.
- Implement separate logic in `apps/cli/commands/wp.ts` for spawning the PHP child process with `stdio: inherit`. This minimizes the number of quirks involved in spawning WP-CLI as a child process of Node.js instead of spawning it directly from a terminal. For example, it makes WP-CLI color its output.
- With native PHP, it doesn't really make sense to send `wp-cli-command` IPC messages to the process manager subprocess, since we can't delegate the WP-CLI command to the running PHP process anyway. Hence, we always spawn the PHP process as an immediate child of the current Node.js process.

## Testing Instructions

1. `npm run download:php-binary`
2. `npm run cli:build`
3. `STUDIO_RUNTIME=native-php node apps/cli/dist/cli/main.mjs site create` and go through the site creation process
4. `node apps/cli/dist/cli/main.mjs wp --path PATH_TO_SITE core version` and ensure that it returns the expected result
5. `node apps/cli/dist/cli/main.mjs wp --path PATH_TO_SITE plugin install wordpress-seo` and ensure that it completes successfully

## Pre-merge Checklist


apps/cli/lib/run-wp-cli-command.ts
```diff
@@ -13,11 +16,116 @@ import { cleanupLegacyMuPlugins, getMuPlugins } from '@studio/common/lib/mu-plug
...
+		child.once( 'exit', ( code ) => resolve( code ?? 1 ) );
+	} );
+
+	return {
+		response: new StreamedPHPResponse(
+			createClosedReadableStream(),
+			toWebReadableStream( child.stdout ),
+			toWebReadableStream( child.stderr ),
+			exitCode
+		),
```
fredrikekelund: Having to construct a `StreamedPHPResponse` instance here really isn't ideal. The `native-php` interfaces should not be using Playground constructs.

I might try to find a neutral interface we can use for both native PHP and Playground before landing this PR.

fredrikekelund: Done in 801ac9f

apps/cli/lib/run-wp-cli-command.ts
```diff
@@ -13,11 +16,132 @@ import { cleanupLegacyMuPlugins, getMuPlugins } from '@studio/common/lib/mu-plug
...
+	get stdoutText(): Promise< string > {
+		this.#stdoutText ??= text( this.stdout );
+		return this.#stdoutText;
+	}
+
+	get stderrText(): Promise< string > {
+		this.#stderrText ??= text( this.stderr );
+		return this.#stderrText;
+	}
+}
```
fredrikekelund, line 63: This is basically a reimplementation of `StreamedPHPResponse`. It exposes `stdout` and `stderr` streams, `stdoutText` and `stderrText` promises, and an `exitCode` promise. 

The idea is to not depend on Playground internals (like `StreamedPHPResponse`) for native PHP code paths.

bcotrim (approved):
Works as expected and code LGTM 👍 
Didn't find any regressions and the approach is very clean.

PR #3257, opened 2026-04-27
Native PHP: on-demand binary downloads for all supported PHP versions

bcotrim:
## Related issues

- Related to RSM-1321

## How AI was used in this PR

Built with Claude Code. I reviewed and validated all code, hashes, and architecture decisions. The SHA-256 hashes were computed by downloading each binary archive from static-php.dev and running `shasum -a 256` locally.

## Proposed Changes

Lays the infrastructure for the native-php runtime to support all PHP versions that Playground does (8.0–8.5). Previously the dev script hardcoded a single version (8.4) with no versioned path, so all versions would've overwritten the same file.

**Approach: on-demand download** (not bundled). Bundling all 7 binaries would add ~210 MB per platform to the app. Instead, each PHP binary is downloaded the first time a site using that version is started, verified with a pinned SHA-256 hash, and cached at `~/.studio/php-bin/<version>/php`.

> Note: PHP 7.4 is excluded — no static binary is available from the upstream `static-php.dev` CDN. It continues to work via the Playground WASM path.

### New files

- **`tools/common/lib/php-binary-metadata.ts`** — Pinned patch versions and SHA-256 hashes for 6 PHP versions (8.0–8.5) × 5 platforms (macOS ARM64/x64, Linux ARM64/x64, Windows x64). Exports `buildPhpBinaryUrl()` and `getPhpBinaryHash()`. Lives in `tools/common` (not `apps/cli`) so both the CLI and the dev script can import it without ESM/CJS conflicts.

- **`apps/cli/lib/dependency-management/php-binary.ts`** — `ensurePhpBinaryAvailable(version, onProgress?)`. Flow: cache hit → in-flight dedup → download → SHA-256 verify (hard fail on mismatch) → extract → install.

### Modified files

- **`paths.ts`** — `getPhpBinaryPath(version)` now includes the version in the path (`~/.studio/php-bin/<version>/php`).
- **`utils.ts`** — `downloadFile()` gains an optional `onProgress(downloaded, total)` callback, reporting every 512 KB.
- **`wordpress-server-manager.ts`** — Calls `ensurePhpBinaryAvailable()` before `startProcess()` when `site.runtime === 'native-php'`, with progress piped through the existing Logger IPC.
- **`logger-actions.ts`** — Adds `ENSURE_PHP_BINARY = 'ensurePhpBinary'` to `SiteCommandLoggerAction`.
- **`scripts/download-php-binary.ts`** — Now accepts `[version]` positional arg (defaults to 8.4), reuses metadata from `php-binary-metadata.ts`, writes to versioned path. Adds `--compute-hashes` flag for re-pinning hashes when patch versions bump.

## Testing Instructions

Build the CLI first:

```bash
npm run cli:build
```

### Binary download on site start (main flow)

The download is triggered when a site with `runtime: 'native-php'` starts. Use `STUDIO_RUNTIME=native-php` to opt in:

```bash
# Remove any cached binary for a clean test
rm -rf ~/.studio/php-bin/8.3

# Create a site with PHP 8.3
STUDIO_RUNTIME=native-php node apps/cli/dist/cli/main.mjs site create --name test-83 --php 8.3

# Start the site — should download + verify the 8.3 binary before spawning the child
STUDIO_RUNTIME=native-php node apps/cli/dist/cli/main.mjs site start test-83
# Expected: progress lines like "Downloading PHP 8.3 (X.X MB / 30.X MB)" then the server starts

# Confirm binary is cached
ls -lh ~/.studio/php-bin/8.3/php

# Start again — should skip download (cache hit)
STUDIO_RUNTIME=native-php node apps/cli/dist/cli/main.mjs site start test-83
# Expected: no download output, starts immediately
```

### Dev script (standalone download helper)

```bash
# Download a specific version
npm run download:php-binary 8.1
# → installs to ~/.studio/php-bin/8.1/php

# Run again → cache hit, prints "already exists" and exits
npm run download:php-binary 8.1

# Defaults to RecommendedPHPVersion (8.4)
npm run download:php-binary

# Compute SHA-256 for updating the hash table (e.g. after a patch bump)
npm run download:php-binary -- --compute-hashes 8.3 darwin arm64
```

## Pre-merge Checklist


scripts/download-php-binary.ts
```diff
@@ -22,173 +23,190 @@ import os from 'os';
...
+const args = z
 	.tuple( [
 		platformSchema.default( process.platform as Platform ),
 		archSchema.default( process.arch as Arch ),
 	] )
-	.transform( ( [ platform, arch ] ) => ( { platform, arch } ) );
-
-const args = argsSchema.parse( [ process.argv[ 2 ], process.argv[ 3 ] ] );
+	.transform( ( [ platform, arch ] ) => ( { platform, arch } ) )
+	.parse( [ positionalArgs[ 1 ], positionalArgs[ 2 ] ] );
```
fredrikekelund:
```ts
const versionSchema = z.enum( SupportedPHPVersions );
const platformSchema = z.enum( [ 'darwin', 'win32', 'linux' ] );
const archSchema = z.enum( [ 'x64', 'arm64' ] );

type Platform = z.infer< typeof platformSchema >;
type Arch = z.infer< typeof archSchema >;

const rawArgs = process.argv.slice( 2 );
const computeHashesMode = rawArgs[ 0 ] === '--compute-hashes';
const positionalArgs = computeHashesMode ? rawArgs.slice( 1 ) : rawArgs;

const args = z
	.tuple( [
		versionSchema.default( RecommendedPHPVersion ),
		platformSchema.default( process.platform as Platform ),
		archSchema.default( process.arch as Arch ),
	] )
	.transform( ( [ version, platform, arch ] ) => ( { version, platform, arch } ) )
	.parse( [ positionalArgs[ 0 ], positionalArgs[ 1 ], positionalArgs[ 2 ] ] );
```
`SupportedPHPVersions` is handy for parsing the `version` argument. Helps us consolidate around zod for parsing the inputs

tools/common/lib/php-binary-metadata.ts
```diff
@@ -0,0 +1,99 @@
...
+// Windows ARM64 falls back to x64, so there is no separate win32-arm64 entry.
+export const PHP_BINARY_HASHES: Record< string, string > = {
+	// PHP 8.5 (8.5.5)
+	'8.5-darwin-arm64': '0974ce32e29d78436fd54c9af4449294f6b138b384c8117a31026d705ac7caa2',
+	'8.5-darwin-x64': '61bb57b22c24f91fd884147533d643cb32d6414b6a1bf848af21504cb3e7b7bf',
+	'8.5-linux-x64': 'b1f5f88287c75432fc8ef256bd05930f156d1f4fb5064f97a2bc59a0b7db10ce',
+	'8.5-linux-arm64': 'b9b7570b586846d19b79077b7dc95ac39a3f8e34a59c01f748672fe65db08c8c',
+	'8.5-win32-x64': 'ddd8098a1e71dfe53c147c392eeaa50eb61259a1c430e3cbe016b0edbdc1cf91',
+	// PHP 8.4 (8.4.20)
+	'8.4-darwin-arm64': 'f0c8e545f77f45d4331e004effa34c863071ea9400c668e4462e41b22ded980d',
```
fredrikekelund: This hash does not match what the download script generates for me locally. @bcotrim, could you verify what you're seeing on your end?

apps/cli/lib/wordpress-server-manager.ts
```diff
@@ -169,6 +172,21 @@ export async function startWordPressServer(
...
+		);
+		await ensurePhpBinaryAvailable(
+			site.phpVersion as SupportedPHPVersion,
+			( downloaded, total ) => {
+				const dl = ( downloaded / 1024 / 1024 ).toFixed( 1 );
+				const tot = total ? ` / ${ ( total / 1024 / 1024 ).toFixed( 1 ) } MB` : '';
+				logger.reportProgress( `Downloading PHP ${ site.phpVersion } (${ dl } MB${ tot })` );
+			}
+		);
+	}
```
fredrikekelund, line 187: Making the native PHP runtime setup "just work" in this way is appealing, but I think we have to view this solution as a stepping stone. For production builds, we should ship the PHP runtimes with the installer. For security reasons, for reliability reasons, and to continue making Studio work offline.

apps/cli/lib/dependency-management/php-binary.ts
```diff
@@ -0,0 +1,148 @@
...
+	NativePhpSupportedVersions,
+	PHP_PATCH_VERSIONS,
+} from '@studio/common/lib/php-binary-metadata';
+import { extract } from 'tar';
+import { getPhpBinaryPath } from './paths';
+import { downloadFile } from './utils';
+import type { SupportedPHPVersion } from '@studio/common/types/php-versions';
+
+// In-flight dedup: if the same version is already downloading, reuse the promise.
+const inFlight = new Map< SupportedPHPVersion, Promise< void > >();
```
fredrikekelund: When would the same CLI process trigger two PHP downloads?

scripts/download-php-binary.ts
fredrikekelund, line 1: Nit, but we might consider using `downloadFile` in this file, too.

apps/cli/lib/dependency-management/php-binary.ts
fredrikekelund: Looks like f5c4ef3 superseded this approach. But still: when is `ensurePhpBinaryAvailable` called more than once in the same process?

fredrikekelund (commented):
I know this isn't ready for review yet, but I took the opportunity to leave some preliminary feedback :)

apps/cli/lib/wordpress-server-manager.ts
bcotrim: That's fair. We already have a similar approach to this one with WordPress versions (although including that would result in a much larger bundle increase I am sure).
I would suggest revisiting this topic at the end of the project, or when we have decided how we are shipping and distributing the binaries.

bcotrim:
thanks for the early review @fredrikekelund 
this should be ready for another go if you can 🙏

apps/cli/lib/wordpress-server-manager.ts
fredrikekelund: > We already have a similar approach to this one with WordPress versions

Good point, but an important addition is that we always ship the latest WP version with the installer. We can consider doing the same for PHP runtimes. We'd win on installer file size, but my instinctive reaction is that we'd pay a relatively high price in complexity.

This question ties into whether and how we should compile PHP ourselves, too.

fredrikekelund (approved):
I made a few changes to bring in the latest changes from trunk and to tighten the types. Fundamentally, this is working really well, though 👍

PR #3254, opened 2026-04-27
deps: bump @wp-playground and @php-wasm packages to 3.1.21

wojtekn:
## Related issues

- https://github.com/WordPress/wordpress-playground/releases/tag/v3.1.21

## How AI was used in this PR

Claude identified and updated all affected package.json files. Changes were reviewed manually.

## Proposed Changes

- Bump `@wp-playground/blueprints`, `@wp-playground/cli`, `@wp-playground/common`, `@wp-playground/storage`, `@wp-playground/wordpress` from `3.1.20` → `3.1.21`
- Bump `@php-wasm/node`, `@php-wasm/universal`, `@php-wasm/util` from `3.1.20` → `3.1.21`
- Updated `package-lock.json` accordingly

Files updated: `apps/cli/package.json`, `apps/studio/package.json`, `apps/ui/package.json`, `tools/benchmark-site-editor/package.json`, `tools/common/package.json`

## Testing Instructions

- Run `npm start` and verify local sites launch correctly
- Run `npm run cli:build && node apps/cli/dist/cli/main.mjs` to confirm CLI works

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #3260, opened 2026-04-27
fix: pin @wp-playground/blueprints to exact version

wojtekn:
## Related issues

Follow-up to #3254.

## How AI was used in this PR

Fix identified by Claude while investigating a 455 MB app size regression shown in CodeVitals on trunk.

## Proposed Changes

Pins `@wp-playground/blueprints` to an exact version in all `package.json` files, consistent with how all other Playground/PHP-WASM packages are already pinned.

**Root cause of the regression:** `^3.1.20` on blueprints meant that when 3.1.21 published to npm, `install:bundle` (which runs `npm install --no-workspaces` inside `apps/cli`) resolved blueprints to 3.1.21. That version of blueprints depends on `@php-wasm/node@3.1.21`, but `apps/cli/package.json` pinned `@php-wasm/node@3.1.20` — so npm installed **both** versions, with the 3.1.21 copies nested under `blueprints/node_modules/`. The `prune-php-wasm` vite plugin only globs top-level `@php-wasm/node-*/asyncify/` paths, so the nested copies (8 PHP versions × ~58 MB each) were never pruned, adding ~466 MB to the app bundle.

## Testing Instructions

- CodeVitals app-size should stay at ~1450 MB, not jump to ~1907 MB after the next Playground release

## Pre-merge Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,10 +25,6 @@ updates:
       - dependency-name: "wpcom"
       - dependency-name: "wpcom-*"
 
-      # WordPress Playground & PHP WASM
-      - dependency-name: "@wp-playground/*"
-      - dependency-name: "@php-wasm/*"
-
       # TypeScript ecosystem
       - dependency-name: "typescript"
       - dependency-name: "typescript-*"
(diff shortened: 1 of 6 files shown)


fredrikekelund (approved):
LGTM 👍 I took the liberty of updating the lockfile

PR #3276, opened 2026-04-29
Remove Studio Code turn cap handling

youknowriad:
## Summary
- stop passing a default `maxTurns` value from the agent wrapper
- remove the special max-turns result path from the JSON adapter, UI, and command flow
- keep explicit `maxTurns` support for callers that provide it
- update unit coverage for the default and explicit SDK passthrough behavior

## Testing
- `npx eslint --fix` on the modified CLI files
- `npm test -- apps/cli/ai/tests/agent.test.ts apps/cli/commands/ai/tests/ai.test.ts`
- `npm run typecheck`


apps/cli/ai/tests/agent.test.ts
```diff
@@ -76,4 +76,28 @@ describe( 'AI agent startup', () => {
 		);
 		expect( result ).toBe( mockQuery );
 	} );
+
```
youknowriad: I think the tests you added are useless just remove them

apps/cli/ai/agent.ts
```diff
@@ -58,7 +58,7 @@ export function startAiAgent( config: AiAgentConfig ): Query {
 		prompt,
 		env,
 		model = DEFAULT_MODEL,
-		maxTurns = 75,
+		maxTurns,
```
youknowriad: Let's remove the config entirely

PR #3280, opened 2026-04-29
Update CLI version to 1.8.1 to prepare for release

youknowriad:


sejas (commented):
@youknowriad , Let's also update the `apps/studio/package.json` so both versions are in sync. Otherwise looks good!

sejas (approved):

PR #3291, opened 2026-04-30
Linux: don't default editor dropdown to VS Code when nothing is installed

gavande1:
## Related issues

- Related to [RSM-870](https://linear.app/a8c/issue/RSM-870/linux-code-editor-dropdown-shows-visual-studio-code-as-selected-when)

## How AI was used in this PR

AI helped trace the dropdown's selection fallback through `PreferencesTab` → `useGetUserEditorQuery` → `getUserEditor()` and confirmed that on Linux with no editor binary on `$PATH`, the IPC layer correctly returns `null` while only the renderer's fallback was forcing `'vscode'`. The fix itself is a one-line change; I reviewed it manually before committing.

## Proposed Changes

- In `apps/studio/src/modules/user-settings/components/preferences-tab.tsx`, stop coercing `editorSelection` to `'vscode'` when neither a dirty edit nor a saved/auto-detected editor exists. Pass `null` through to `EditorPicker` instead so it renders the neutral "Select" placeholder.
- Preserves the existing dirty-state semantics by switching from `??` to an explicit `dirtyEditor !== undefined` check, so a user who actively clears the selection still sees `null` rather than falling back to the saved value.

## Testing Instructions

On Linux, Studio detects editors by walking `$PATH` for any of: `antigravity`, `code`, `phpstorm`, `webstorm`, `windsurf`, `cursor`, `subl`, `zed`. To simulate the "no editor installed" state without uninstalling anything, temporarily rename each detected binary, then revert when done.

1. Identify which editor binaries currently resolve on `$PATH`:

    ```bash
    for cmd in antigravity code phpstorm webstorm windsurf cursor subl zed; do
      printf '%-12s -> %s\n' "$cmd" "$(command -v "$cmd" 2>/dev/null || echo 'not found')"
    done
    ```

2. For each command that resolved to a path, rename it so the lookup fails. Example for VS Code at `/usr/bin/code`:

    ```bash
    sudo mv /usr/bin/code /usr/bin/code.bak
    ```

   Repeat for every editor binary found in step 1 (e.g. `cursor`, `subl`, …).

3. Re-run the loop from step 1 and confirm every command prints `not found`.

4. Ensure `~/.studio/app.json` does not have a `preferredEditor` key set (delete it while Studio is closed if it does), otherwise the saved value will mask the empty-detection state.

5. Launch Studio (`npm start`) and open Settings → General

    - Expected: the Code editor dropdown shows the neutral "Select" placeholder. Opening it shows Visual Studio Code (and any others) under the Not installed group.
    - Before the fix: the dropdown shows "Visual Studio Code" as the selected value even though it's also listed under "Not installed".

6. Revert the renames once testing is complete:

    ```bash
    sudo mv /usr/bin/code.bak /usr/bin/code
    # …and any others you renamed in step 2
    ```

Regression checks (with editor binaries restored):

- Open Settings → General → confirm the auto-detected editor is selected as before.
- Pick an editor, click Save, reopen Settings → confirm the saved selection persists.
- Change the selection without saving, close the panel, reopen → confirm no stale dirty state leaks across sessions.

CleanShot 2026-04-30 at 12 39 05@2x

## Pre-merge Checklist


ivan-ottinger (approved):
Good catch of that issue! The change looks good and works well.

Before (with no editors installed):

Image

After (with no editors installed):

Image

PR #2995, opened 2026-04-07
Add optional Artifact Signing for Windows builds

mokagio:
## Related issues

- AINFRA-2233
- #2709

## How AI was used in this PR

Implementation planned and written with Claude Code (Opus 4.6).
All code reviewed by @mokagio before committing.

## Proposed Changes

Takes the work from #2709 and puts it behind a toggle env var.
The idea being that we can switch back to the previous signing implementation if something turns out to be wrong once released in the wild.

- Add Azure Trusted Signing utilities (`azure-signing.cjs`, `azure-sign-hook.js`) for SHA256-only signing via signtool
- Add `windowsSign.ts` module gated on `USE_AZURE_TRUSTED_SIGNING` env var — returns Azure hook config when set, `undefined` (PFX fallback) otherwise
- Wire `windowsSign` into `forge.config.ts` for both `packagerConfig` and `MakerSquirrel`
- Add conditional signing setup in `build-for-windows.ps1` (Azure or PFX based on toggle)
- Add conditional signing in `package-appx.mjs` for sideload AppX
- Update CI toolkit version to include `setup_azure_trusted_signing.ps1`
- Set `USE_AZURE_TRUSTED_SIGNING=1` in both dev and release Windows build jobs in the pipeline

To revert to PFX signing, remove the `USE_AZURE_TRUSTED_SIGNING` env var from the pipeline YAML files.

## Testing Instructions

I created #2996 to test the changes here and verify that both build paths work.

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors? - N.A.

*Posted by Claude Code (Opus 4.6) on behalf of @mokagio with approval.*

PR #3292, opened 2026-04-30
Fix Windows build script for PowerShell 5.1

mokagio:
## Related issues

- Fixes [AINFRA-2344](https://linear.app/a8c/issue/AINFRA-2344)

## How AI was used in this PR

Investigation, root cause, and one-line fix produced by Claude Opus 4.7. Validated locally before pushing.

## Proposed Changes

The Windows Buildkite agents invoke `.buildkite/commands/build-for-windows.ps1` via `powershell` (Windows PowerShell 5.1), which does not support the `??` null-coalescing operator introduced in PowerShell 7.0.
The operator landed in #2995 and broke every Windows dev build on `trunk` since (15258, 15259, 15260, 15270, 15275 — both `x64` and `arm64`), failing at parser stage with `Unexpected token '??' in expression or statement.`.

This PR replaces the inline `??` with a 5.1-safe pre-assignment.
Stringifying `$env:USE_AZURE_TRUSTED_SIGNING` via `"$(...)"` yields `""` when the variable is unset, so the subsequent `.Trim().ToLower()` is safe without null-coalescing.

The release pipeline (`.buildkite/release-build-and-distribute.yml`) invokes the same script the same way, so this fix unbreaks both dev and release lanes.

## Testing Instructions

Local validation (works from macOS — no Windows VM needed):

`````sh
pwsh -NoProfile -Command "
  Import-Module PSScriptAnalyzer
  Invoke-ScriptAnalyzer -Path .buildkite/commands/build-for-windows.ps1 \
    -Settings @{ Rules = @{ PSUseCompatibleSyntax = @{ Enable = \$true; TargetVersions = @('5.1') } } } |
    Where-Object RuleName -eq 'PSUseCompatibleSyntax'
"
`````

Expected: zero `PSUseCompatibleSyntax` findings.
The same rule on the pre-fix line reports: *"The null-coalescing operator syntax `\$x ?? \$y` is not available by default in PowerShell versions 3,4,5,6"*.

The real proof is CI — the `Windows Dev Build - x64` and `Windows Dev Build - arm64` jobs should now reach the actual build steps instead of failing at script parse.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? — N/A; PowerShell-only change.

### DIFF
--- a/.buildkite/commands/build-for-windows.ps1
+++ b/.buildkite/commands/build-for-windows.ps1
@@ -28,7 +28,8 @@ if ($Architecture -notin $VALID_ARCHITECTURES) {
     Exit 1
 }
 
-If (@('1', 'true') -contains ($env:USE_AZURE_TRUSTED_SIGNING ?? '').Trim().ToLower()) {
+$useAzureTrustedSigning = "$($env:USE_AZURE_TRUSTED_SIGNING)".Trim().ToLower()
+If (@('1', 'true') -contains $useAzureTrustedSigning) {
     Write-Host "--- :lock: Setting up Azure Trusted Signing"
     $setupScript = (Get-Command setup_azure_trusted_signing.ps1 -ErrorAction Stop).Source
     & $setupScript


wojtekn (approved):
Looks good, thanks for the fix.

wojtekn:
I triggered Windows builds for this PR.

PR #3297, opened 2026-04-30
Build static macOS PHP CLI binaries

bcotrim:
## Related issues

- Related to RSM-1650

## How AI was used in this PR

AI helped compare the Buildkite and GitHub Actions approaches, simplify the static-php-cli build, and validate the generated artifacts. I reviewed the changes and tested the resulting binaries locally.

## Proposed Changes

- Add a manual `Build PHP CLI Binaries` GitHub Actions workflow.
- Build PHP CLI binaries with upstream `static-php-cli` instead of custom Buildkite scripts.
- Default to PHP `8.4.20` and static-php-cli `2.8.5`, with workflow inputs to override them.
- Build the currently working targets: `macos-aarch64`, `macos-x86_64`, and `windows-x86_64`.
- Upload packaged archives plus `.sha256` sidecars as GitHub Actions artifacts.
- Leave Apps CDN upload and manifest updates for a follow-up PR.

## Testing Instructions

- Confirm the `macos-aarch64`, `macos-x86_64`, and `windows-x86_64` jobs pass.
- Download the artifacts and verify the inner archive against its `.sha256` file.
- Extract the archive and confirm `php -v` reports PHP `8.4.20`.
- Move the archive to `.studio` folder - `~/.studio/php-bin/8.4` for example
- `npm run cli:build`
- `STUDIO_RUNTIME=native-php node apps/cli/dist/cli/main.mjs site create`
- Confirm the WordPress site is running

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #3301, opened 2026-04-30
Fix E2E tests on macOS when `STUDIO_RUNTIME=native-php`

fredrikekelund:
## Related issues

- Fixes RSM-1725

## How AI was used in this PR

Primarily to help me debug the problem. I alternated between Claude and Codex.

## Proposed Changes

While working on https://github.com/Automattic/studio/pull/3271, @bcotrim discovered that the E2E tests were failing when run with `STUDIO_RUNTIME=native-php`. This PR fixes that by implementing the following fixes:

- When Studio clones a site, explicitly update the site URL. Playground can handle this internally, but the native PHP runtime obviously cannot.
- On Posix platforms, spawn process daemon children with `detached: true` to assign them to a new process group. This is so we can pass a negative PID to `process.kill` to terminate the entire process group (the Node.js process and the associated PHP process). This approach isn't bulletproof, but it's alright, and it works for the purpose of the E2E tests.
- Favor SIGTERM over SIGKILL when killing child processes to give them a chance to clean up.

## Testing Instructions

Run the E2E tests locally on macOS by following these steps:

1. `npm run package`
2. `STUDIO_RUNTIME=native-php npm run e2e`
3. Ensure that they pass successfully

## Pre-merge Checklist


apps/studio/src/ipc-handlers.ts
```diff
@@ -1053,6 +1054,10 @@ export async function copySite(
 		noStart: true,
 	} );
 
+	// Playground sets the correct siteurl internally, but for the native-php runtime, we need to
+	// explicitly update that option
+	await updateSiteUrl( server, `http://localhost:${ details.port }` );
```
bcotrim, line 1059: Should we run this only for `native-php`?

bcotrim (commented):
I had trouble running the e2e tests locally.
First issue was checksum mismatches — looks like upstream static-php-cli re-publishes archives, so the pinned hashes go stale. Should be moot once [#3297](https://github.com/Automattic/studio/issues/3297) lands and we build binaries ourselves.

Second issue: even with hashes fixed, e2e teardown hung for 3 minutes with one PHP child orphaned per session. Root cause is that php-server-child.ts only handles exit/disconnect, not SIGTERM. When the daemon SIGTERMs the Node fork, it dies without firing exit, so killPhpProcess() never runs and the PHP grandchild leaks. Adding these unblocks teardown for me:

```typescript
process.on( 'SIGTERM', () => {
	killPhpProcess();
	process.exit( 0 );
} );
process.on( 'SIGINT', () => {
	killPhpProcess();
	process.exit( 0 );
} );
```

Could you reproduce + confirm the fix on your end?

fredrikekelund:
Sorry about that, @bcotrim. I did confirm this working previously… It seems you're correct that `process.on( 'exit' )` isn't called in response to `SIGINT` or `SIGTERM` signals, so I'm not sure what made the cleanup work before. I may have broken it by messing up a git merge.

In any case, I've updated `php-server-child` to explicitly handle `SIGINT` or `SIGTERM` signals. I've also made it so `process-manager-daemon` sends `SIGTERM` signals to the entire process group (previously we `SIGTERM` only to the Node.js child process and followed up with `SIGKILL` signals to the group if needed).

I've also updated the PHP binary checksums for macOS with ARM chips to make the tests pass.

apps/studio/src/ipc-handlers.ts
fredrikekelund: It's obviously not required for Playground, but that's only because Playground enforces the siteurl internally. I think it makes sense to follow the more standard approach of explicitly updating the siteurl rather than relying on some runtimes to do it for us.

fredrikekelund:
I'm playing around with Windows changes here, too. I could move those commits to a new PR, though. 

@bcotrim, if you review this again, feel free to ignore the changes explicitly targeting Windows. I'll let you know when those are ready for review.

bcotrim (approved):
Thanks for following up on the changes @fredrikekelund 
e2e now run successfully on macOS.

I would suggest moving the Windows to another PR so we can land this one.

Nit: A nice workflow would be to control the `STUDIO_RUNTIME` via a PR label (for example) so the CI runs native PHP e2e. But we can do it in a follow-up PR, and maybe only after we have the binaries in our own CDN.

fredrikekelund:
> A nice workflow would be to control the STUDIO_RUNTIME via a PR label (for example) so the CI runs native PHP e2e

Good idea 👍 I added this in https://github.com/Automattic/studio/pull/3343

PR #3307, opened 2026-04-30
Fix AI screenshot local site resolution

chubes4:
## Summary
- Allow both `take_screenshot` and `share_screenshot` to target local Studio sites with the same `nameOrPath` contract used by the rest of the local-site AI tool surface.
- Add a small `resolveScreenshotUrl()` helper consumed by both tools.
- Preserve existing absolute `url` screenshot behavior; emit a clean error when neither `url` nor `nameOrPath` is provided.

## Why

The local Studio AI prompt tells agents to use `take_screenshot` after building a local site, but the tool originally only accepted `url`. Neighboring local-site tools such as `wp_cli` and `validate_blocks` accept `nameOrPath`, so agents can naturally call `take_screenshot` with `nameOrPath` + `path` and hit MCP schema validation before the screenshot handler runs.

`share_screenshot` (added by #3272) has the same URL-resolution gap: agents that know a local Studio site by name must already know its public URL to share a screenshot of it. Per the standing "fix upstream first, never paper over" rule, fixing one tool but not the other would calcify the workaround pattern. Both tools now share the same resolver and the same fallback behavior — pass `url`, or pass `nameOrPath` (+ optional `path`), or get a clean error.

Closes #3306.

## What changed since the original draft

- Extended `resolveScreenshotUrl()` to `share_screenshot` in addition to `take_screenshot`. Both tools now accept the same `{ url?, nameOrPath?, path? }` shape and call the same resolver.
- Slimmed the test diff to a single regression test that proves `nameOrPath` resolves correctly. The previous draft had four behavior tests plus a mock helper; the regression we need to guard against is specifically "agent passed `nameOrPath`, screenshot resolved against the local site." Types and existing trunk tests cover gating and tool-list shape; explicit-URL / path-composition / missing-input error cases can land in follow-up coverage if needed.

## Tests
- `npx vitest run apps/cli/ai/tests/tools.test.ts` — 22 passed
- `npx eslint apps/cli/ai/tools.ts apps/cli/ai/tests/tools.test.ts` — clean
- `npm run typecheck --workspace apps/cli` — clean

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (GPT-5.5) for the original draft; Claude Code (Sonnet 4.5) for the share_screenshot extension and test slimming.
[...]

PR #3309, opened 2026-04-30
Experiment: use Static Site Importer for generated local sites

chubes4:
## Summary
- Bundles [Static Site Importer](https://github.com/chubes4/static-site-importer) into Studio via the existing `wp-files/` download chain. A fresh `npm install` fetches SSI's `main` zipball, extracts to `wp-files/static-site-importer/`, and the CLI bundle ships it. No environment variables, no separate clone, no setup steps required.
- Reframes local site generation around a normal static HTML/CSS handoff: the agent writes `tmp/static-site/index.html`, then imports it with `static-site-importer import-theme`.
- Updates the deterministic smoke script (`apps/cli/scripts/studio-bfb-smoke.mjs`) to validate the bundled flow end-to-end: SSI generates and activates a block theme with templates, parts, a front-page pattern, and zero fallback events.

## Why
Studio's local site-generation path should not require agents to hand-author Gutenberg block markup or maintain brittle block-format prompt skills. [Static Site Importer](https://github.com/chubes4/static-site-importer) gives Studio a reusable pipeline:

1. The agent writes familiar static HTML/CSS.
2. Static Site Importer converts that HTML into a WordPress block theme.
3. The generated site remains editable in the Site Editor.

This also supports customer-provided static HTML, migrated templates, and AI-generated prototypes as inputs instead of forcing Studio to discard or rewrite them.

Related upstream projects:

- Static Site Importer: https://github.com/chubes4/static-site-importer
- Block Format Bridge: https://github.com/chubes4/block-format-bridge
- HTML to Blocks Converter: https://github.com/chubes4/html-to-blocks-converter

## How SSI gets into the site

```
npm install
  └─ scripts/download-wp-server-files.ts
       └─ fetches https://github.com/chubes4/static-site-importer/archive/refs/heads/main.zip
            └─ extracts into wp-files/static-site-importer/
                 ├─ static-site-importer.php       (plugin entry)
                 ├─ vendor/autoload.php            (composer deps, committed in SSI repo)
                 └─ includes/.../cli-command.php   (registers `wp static-site-importer`)

apps/cli build
  └─ Vite copies wp-files/ into dist/cli/wp-files/

site_create / site_start
  └─ tools/common/lib/mu-plugins.ts
       └─ symlinks dist/cli/wp-files/static-site-importer/ into the per-site mu-plugins temp dir
            └─ a small loader mu-plugin requires static-site-importer.php
                 └─ SSI's CLI command registers → `wp static-site-importer ...` works
```

Path resolution lives in `apps/cli/lib/dependency-management/paths.ts::getStaticSiteImporterPluginPath()`, which mirrors `getPhpMyAdminPath()` / `getSqliteCommandPath()` / `getBlueprintsPharPath()` — all bundled assets resolved against `getWpFilesPath()`.

## Iteration model (draft phase)

While the SSI substrate is still rapidly evolving, the downloader pulls **`main`** rather than a tagged release. Reviewers automatically pick up upstream fixes by re-running `npm install`. There's no per-machine setup and no env-var contract — a fresh clone of this PR plus `npm install` is everything a reviewer needs.

**Before this PR un-drafts**, swap the URL in `scripts/download-wp-server-files.ts` for a pinned tag, e.g.:

```diff
-		getUrl: () =>
-			'https://github.com/chubes4/static-site-importer/archive/refs/heads/main.zip',
+		getUrl: () =>
+			'https://github.com/chubes4/static-site-importer/archive/refs/tags/v0.4.0.zip',
```

That locks the substrate to a known-good release for merge.

## Verification (Riad-equivalent reviewer simulation)

I created a fresh worktree off this branch in a directory that has no relationship to my local SSI development setup, with no environment variables set, and walked it through the reviewer flow:

```
$ git worktree add studio@bfb-reviewer-sim bfb-mu-plugin-agent-output-draft
$ cd studio@bfb-reviewer-sim
$ npm install --no-audit --no-fund
   ...
   [static-site-importer] Downloading Static Site Importer (chubes4/static-site-importer @ main) ...
   [static-site-importer] Extracting files from zip ...
   [static-site-importer] Files extracted
   ...
$ ls wp-files/static-site-importer/static-site-importer.php
   wp-files/static-site-importer/static-site-importer.php
$ ls wp-files/static-site-importer/vendor/autoload.php
   wp-files/static-site-importer/vendor/autoload.php
$ cd apps/cli && npm run build
   ✓ built in 2.43s
$ node apps/cli/scripts/studio-bfb-smoke.mjs
   {
       "importer_loaded": true,
       "bfb_loaded": true,
       "h2bc_loaded": true,
       "file:parts/header.html": true,
       "file:parts/footer.html": true,
       "file:templates/front-page.html": true,
       "file:templates/page-home.html": true,
       "file:patterns/page-home.php": true,
       "file:style.css": true,
       "file:theme.json": true,
       "front_page_id": 4,
       "home_page_id": 4,
       "active_theme": "studio-static-importer-smoke",
       "fallback_count": 0,
       "header_html_blocks": 0,
       "footer_html_blocks": 0,
       "pattern_has_blocks": true
   }
   PASS: Studio Static Site Importer smoke generated an editable block theme.
```

## Tests

- `npm run typecheck` (all workspaces) — clean
- `npx eslint <touched files>` — clean
- `npx vitest run` — 1483 passed across 157 test files, including 9 in `tools/common/lib/tests/mu-plugins.test.ts` (3 new SSI tests cover loader + symlink presence when the plugin path is provided, and absence when it isn't)
- `apps/cli/scripts/studio-bfb-smoke.mjs` — passes end-to-end against a freshly-bundled site

## Notes

- This bundles two unrelated prompt/UI tweaks from the same iteration into the SSI commit: dropping two outdated guidance lines from `apps/cli/ai/system-prompt.ts` (now superseded by the SSI workflow itself), and a welcome-text wording change in `apps/cli/ai/ui.ts` ("block themes" → "WordPress sites"). Both are minor, both ride along.
- Unstacked from #3307 (which fixes `take_screenshot`/`share_screenshot` URL resolution for local sites). The two PRs touch separate concerns; #3309 no longer rides on top of #3307 and can land independently.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Code (Sonnet 4.5)
- **Used for:** Drafted the wp-files download entry, mu-plugin contract refactor (env-var → bundled-path option), CLI entry-point wiring, smoke-script clean-up, and the new mu-plugins tests under Chris's direction and review. The original draft of this PR (commit `2f336e86`) was authored separately by OpenCode/GPT-5.5; the iteration on top is Claude Code's.

### DIFF
--- a/apps/cli/ai/system-prompt.ts
+++ b/apps/cli/ai/system-prompt.ts
@@ -156,8 +156,6 @@ One \`Write\` or \`Edit\` per turn (read-only \`site_info\`, \`site_list\`, \`wp
 
 **After \`site_create\`** (or "redesign"/"rebuild"/"start over" triggers), the next turn MUST be small: \`site_info\` or a single ≤50-line \`Write\`. Never scaffold a whole theme in one turn.
 
-**Long static HTML files (>~200 lines): skeleton first, then fill across Edits.** Create \`index.html\` with the document shell, \`<style>\` block anchors, and body section anchors first; fill one section/anchor per Edit.
-
 ## Available Studio Tools (prefixed with mcp__studio__)
 
 - site_create: Create a new WordPress site (name only — handles everything automatically)
@@ -182,7 +180,6 @@ One \`Write\` or \`Edit\` per turn (read-only \`site_info\`, \`site_list\`, \`wp
 
 ## General rules
 
-- Design quality and visual ambition are not in conflict with Site Editor compatibility. Normal static HTML/CSS targeting semantic classes can achieve any visual design; the importer turns that static site into editable WordPress block theme output.
 - Do NOT modify WordPress core files. Only work within wp-content/.
 - Before running wp_cli, ensure the site is running (site_start if needed).
 - Let Static Site Importer generate and activate the block theme. Do not manually create theme files unless the user explicitly asks for custom plugin/theme development.
--- a/apps/cli/ai/system-prompt.ts
+++ b/apps/cli/ai/system-prompt.ts
@@ -98,7 +98,7 @@ Use \`per_page\` and \`page\` for pagination. Use \`status\` to filter by publis
 1. **Check the site plan** (MANDATORY FIRST STEP): Use \`GET /\` (apiNamespace: \`""\`) to get site info and check \`plan.product_slug\`. Stop and inform the user if they request features unavailable on their plan.
 2. **Understand the site**: Use \`GET /posts\` to list content, \`GET /themes?status=active\` to see the active theme.
 3. **Make changes**: Use POST requests to create/update content, manage templates, switch themes.
-4. **Finish promptly**: Summarize what changed and let the user review the result. Do not run visual screenshot loops unless the user explicitly asks.
+4. **Finish promptly**: Summarize what changed and let the user review the result.
 
 ## General rules
 
@@ -148,7 +148,7 @@ Then continue with:
 2. **Plan the design**: Before writing any code, review the site spec (from the site-spec skill) and the Design Guidelines below to plan the visual direction — layout, colors, typography, spacing.
 3. **Write the static site**: Use Write and Edit to create \`<site>/tmp/static-site/index.html\` as a full static HTML document with semantic body structure and CSS/assets.
 4. **Import into WordPress**: Use \`wp_cli\` to run \`static-site-importer import-theme /wordpress/tmp/static-site/index.html --slug=<theme-slug> --name="<Theme Name>" --activate --overwrite\`. The importer converts the static HTML into an editable block theme and front page. The \`wp_cli\` tool already runs WP-CLI, so do not prefix commands with \`wp\`. It takes one literal WP-CLI command per call, not shell commands: never combine commands with \`&&\`, \`;\`, pipes, redirection, shell substitution such as \`$(cat file)\`, backticks, environment variables, or host temp-file paths. Paths passed to WP-CLI must use WordPress's mounted filesystem (\`/wordpress/...\`), not host paths.
-5. **Check the result**: Use take_screenshot to capture the site's landing page on desktop and mobile and verify the design visually on both viewports, check for wrong spacing, alignment, colors, contrast, borders, hover styles and other visual issues. Fix any issues found. Pay particular attention to the navigation menu and the CTA buttons. The design needs to match your original expectations. **Width check**: any section that was meant to be full-width (heroes, banners, edge-to-edge galleries, full-bleed footers) must visibly span the entire viewport in the desktop screenshot.
+5. **Finish promptly**: Run any cheap static checks that are directly relevant, summarize what changed, and let the user review the result.
 
 ## Working cadence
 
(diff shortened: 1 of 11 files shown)


chubes4:
Testing it with lots of different prompts using automated benchmarks.

Site build prompts being tested: https://github.com/chubes4/homeboy-rigs/tree/main/Automattic/studio/bench/prompts/site-build

Here is a screenshot from the first passing (all valid blocks matching the frontend) generation from last night:

screencapture-localhost-9180-2026-05-02-21_49_16

youknowriad:
I'm giving this PR a try. Quick questions though:

- how does the block conversion work? Is it hard-coded rules? like conceptually how does it work. My guess is that it's only core blocks right?

- why not a nodeJS converter instead of php (to allow for using the same tool for both remote and local sites), we can't install mu plugins on remote sites.

youknowriad:
Not sure what I'm doing wrong, this is not working for me

Screenshot 2026-05-04 at 9 54 06 AM

chubes4:
> how does the block conversion work? 

My initial approach was to mirror the client side rawHandler in Gutenberg, using the core HTML API. Yes only core blocks for now. Non-supported blocks fallback to core/html. 

There is some hardcoding, but I am trying my best to make it more about generic pattern recognition. 

My approach has been to run many different prompts, fix the gaps, and run again. The quality has improved significantly over time. 

Even with hardcoding, there are certain patterns the AI is likely to use. I think we could account for most of them and handle edge cases as they come up. 

> why not a nodeJS converter instead of php 

Honestly, I did not think of that. I had started this experiment a while back for a plugin that creates content at scale. The site editor angle is new. 

Eventually, this could be the responsibility of Gutenberg, instead of Studio. 

> Not sure what I'm doing wrong

You did nothing wrong, it was me who accidentally left a local path in the original draft 🥇. Sorry about that!

youknowriad:
I think this PR is trying to do too much, change how we transform blocks but also change other things in the system prompt... I've also fallen into this trap on previous PRs and it's very hard to gain confidence about whether the change you're making is the right one or not.

I would like to suggestion smaller steps. For the static site import (or block transformation) which is the main thing this PR is trying to achieve. I'm wondering if there's a simple solution that avoid building a php transformer entirely and just rely fully on the raw handler. The advantage is that it will work consistently and support any blocks that is available on the site.  We could achieve it by using an eval in playwright like we do today for block validation, it's just that instead of block validation, we'll be doing block transformation directly.

chubes4:
Thanks for the feedback. I will see if I can come up with a simpler approach. I do want to keep experimenting with this, because I already built the PHP transformer, and have been getting good results with it. 

For the system prompt, I think we should gradually be moving design and workflow constraints out of the system prompt and into a skill. As part of this experiment, I tweaked the prompt to try and speed up generation by reducing the number of steps. 

I am running some benchmarks that track the agent's font choices, color choices, and design decisions in a database over the course of many eval runs with a given system prompt. This can help us measure the impact of prompt changes on output with hard data and make sound decisions. 

The idea is that eventually, Studio Code can become a general-purpose WordPress agent that is powerful enough to be the daily driver for the average WordPress developer.

youknowriad:
> I think we should gradually be moving design and workflow constraints out of the system prompt and into a skill.

I think that makes sense yeah :) 

Looking forward to see where these experiments lead.

chubes4:
Closing this draft as superseded by a cleaner Studio-owned import architecture.

The direction in this PR was useful for proving SSI can power generated local sites, but it couples the work too tightly to the agentic loop and bundled mu-plugin path. The newer direction is to make Studio own a general site-creation/import primitive and use Static Site Importer as the PHP import engine under the hood.

New draft exploration: https://github.com/Automattic/studio/pull/3952

That branch starts with a CLI-first shape:

```bash
studio create --from-artifact ./artifact.json --name "Imported Site"
```

This gives us a generic path for Figma exports, HTML/static artifacts, AI-generated artifacts, and future source adapters without requiring the agent loop to own the import mechanics.

PR #3314, opened 2026-05-01
Animate sidebar thread groups

youknowriad:
Adds a smooth expand and collapse treatment for site conversation threads in the sidebar, plus a hover chevron that reflects each site group's open state.

## Proposed Changes

- Add a right/down chevron to each site row that appears on hover or keyboard focus.
- Keep populated thread lists mounted inside an animated wrapper so collapsing and expanding transitions smoothly.
- Remove collapsed thread links from keyboard tab order while they are visually hidden.
- Respect reduced-motion preferences for the new sidebar transitions.

## Testing Instructions

- Start the new UI and open the dashboard sidebar with at least one site that has conversation threads.
- Hover a site row and confirm the chevron appears pointing down when expanded and right when collapsed.
- Click the site row to collapse and expand its thread list, confirming the list animates smoothly in both directions.
- Confirm collapsed thread links are not reachable with Tab until the site row is expanded again.

## Automated Checks

- `npx eslint --fix apps/ui/src/components/site-list/index.tsx apps/ui/src/components/site-list/style.module.css` (CSS path is ignored by the ESLint config with a warning)
- `npm run typecheck`
- `npm test -- --project ui --passWithNoTests`
- `npm -w @studio/ui run build`

## Pre-merge Checklist

### DIFF
--- a/apps/ui/src/components/site-list/style.module.css
+++ b/apps/ui/src/components/site-list/style.module.css
@@ -36,6 +36,7 @@
 .siteToggle {
 	max-width: 100%;
 	height: auto;
+	gap: var(--wpds-dimension-padding-xs);
 }
 
 .siteName {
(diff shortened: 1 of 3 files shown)


apps/ui/src/components/site-list/style.module.css
```diff
@@ -88,7 +110,31 @@
...
+	transition: grid-template-rows 160ms ease, opacity 120ms ease;
+}
+
+.sessionListFrameOpen {
+	grid-template-rows: 1fr;
+	opacity: 1;
+	pointer-events: auto;
+}
+
+@media (prefers-reduced-motion: reduce) {
```
youknowriad: Is the prefers-reduced-motion style something we should be applied globally to all animations in a generic way instead of having it per animation? WDYT?

PR #3316, opened 2026-05-01
Fix w logo display

mtias:
Correct the rendering of the logo dots effect


youknowriad (approved):

apps/ui/src/components/session-view/empty-background/wordpress-logo.png
youknowriad, line 1: We should remove this one then.

youknowriad: Done in 6500ca493 — thanks for catching it.

youknowriad: Funny, my agent replying to the real me :)

PR #3326, opened 2026-05-04
Add Linux make scripts for cross-architecture builds

ivan-ottinger:
## Related issues

Part of [RSM-1314](https://linear.app/a8c/issue/RSM-1314) — Linux release prep.

## How AI was used in this PR

AI-assisted throughout: investigated why `npm run make -- --arch=x64 --platform=linux` silently produced a host-arch DEB (root cause: `scripts/package-in-isolation.ts` only reads `process.argv[2]` and ignores the trailing flags), drafted the new `make:linux-*` scripts mirroring the Mac/Windows pattern, ran the cross-build, inspected the resulting DEB with `dpkg-deb -x` + `file` to confirm every binary was `x86-64`, and drafted this PR body. Author reviewed each step, ran the builds locally, and pushed back on overstated claims (e.g. "any arch" was tightened to "an arm64 Linux host" since only that direction was directly verified).

## Proposed Changes

Adds `make:linux-x64` and `make:linux-arm64` scripts to `apps/studio/package.json` and matching wrapper entries in the root `package.json`, mirroring the existing `make:macos-*` and `make:windows-*` patterns.

Without these, the only way to build a Linux DEB locally is bare `npm run make`, which always builds for the host CPU.

This is a prerequisite for adding a Linux build group to Buildkite.

## Testing Instructions

From a Linux host:

```
npm run make:linux-x64
npm run make:linux-arm64
```

Output lands at:
- `apps/studio/out/make/deb/x64/studio_<version>_amd64.deb`
- `apps/studio/out/make/deb/arm64/studio_<version>_arm64.deb`

Verified locally on an arm64 Linux host: `npm run make:linux-x64` produces an x86-64 DEB. `file` check on the unpacked contents confirms the main binary, all Electron `.so` files, and the bundled `node` (used by the proxy with `setcap CAP_NET_BIND_SERVICE`) are all `x86-64`. Native modules `koffi` and `fs-ext-extra-prebuilt` ship prebuilt binaries for every arch in the npm package and resolve at runtime, so no per-arch compilation is needed. No workarounds equivalent to Mac's `npm rebuild fs-xattr --cpu universal` are required.

## Pre-merge Checklist


gavande1 (approved):
Nice, the changes look good to me. I was able to build binary locally. ✅

PR #3335, opened 2026-05-04
Fix Studio app metadata for Linux software stores

gavande1:
## Related issues

- Fixes [RSM-1313](https://linear.app/a8c/issue/RSM-1313/fix-studio-app-description-in-the-stores-on-linux-systems)

## How AI was used in this PR

Used AI to:
- Pull the canonical Microsoft Store description copy from the Store API and adapt it for the Linux package metadata.
- Walk through electron-installer-debian's defaults and template logic to identify why the synopsis and extended description were duplicating in the .deb control file.

I reviewed each diff and pushed back on changes I didn't think were right (e.g. an earlier draft used `genericName` for `Name=`, which mismatched the runtime WMClass).

## Proposed Changes

- **`forge.config.ts` MakerDeb options**: set explicit `productName: 'WordPress Studio'` and `description` (synopsis) + `productDescription` (long, mirrors MS Store copy in ASCII-safe punctuation). Without these, both fields fall back to `package.json.description`, producing a duplicated Description block.
- **`desktop.ejs`**: hardcode `StartupWMClass=Studio` so the launcher entry stays linked to Electron's runtime WMClass even though `Name=` is now `WordPress Studio`. Without this, the launcher icon would not group with the running window.

## Testing Instructions

On a Linux VM:

1. Build the .deb: `npm run make`.
2. Purge any prior install (for a clean slate): `sudo apt purge studio`.
3. Verify package metadata:
   - `dpkg-deb -I apps/studio/out/make/deb/<arch>/studio_*.deb`
   - `dpkg-deb --fsys-tarfile apps/studio/out/make/deb/<arch>/studio_*.deb | tar -xO ./usr/share/applications/studio.desktop`
4. Confirm title and description in the output above:
   - control `Description:` — new synopsis ("Meet Studio - ...") followed by the 3-paragraph long description.
   - `.desktop` — `Name=WordPress Studio`, `Comment=Meet Studio - ...`, `StartupWMClass=Studio`.
5. Open the .deb in the app center (double-click in Files, or `xdg-open apps/studio/out/make/deb/<arch>/studio_*.deb`).
6. Confirm title and description in the app center preview match the new copy. (See Known limitation re: title.)

CleanShot 2026-05-04 at 22 37 20@2x

### Known limitation

The pre-install software-store preview shows the lowercase Debian package name (`studio`) as the title, not `WordPress Studio`. PackageKit's local-deb plugin doesn't read AppStream metainfo from inside .deb files, only the control fields, and `Package:` must be lowercase per Debian policy. Renaming the package to `wordpress-studio` would address this but cascades into install paths, postinst probes, and `rc`-state migrations for users with existing `studio` installs — leaving for a follow-up issue.

## Pre-merge Checklist

- [ ] Verified the rebuilt .deb's control + .desktop files
- [ ] Verified software-store listing renders proper metadata after install


apps/studio/forge.config.ts
```diff
@@ -82,10 +82,20 @@ const config: ForgeConfig = {
...
+				// fully-qualified "WordPress Studio" in their menus.
+				productName: 'WordPress Studio',
 				categories: [ 'Utility' ],
 				name: 'studio',
 				bin: 'studio',
+				// Synopsis and extended description shown by package managers and
+				// software stores. Without these, electron-installer-debian falls
+				// back to package.json.description for both, producing a duplicated
+				// Description block. Copy mirrors the Microsoft Store listing.
+				description: 'Meet Studio - a fast, free way to develop locally with WordPress.',
```
gavande1, line 96: Now that `productName` is `WordPress Studio`, should this open with **"Meet WordPress Studio"** for consistency with the launcher/store name? The counter-argument is keeping it verbatim with the Microsoft Store copy. No strong opinion — what do you think? cc @ivan-ottinger

apps/studio/installers/desktop.ejs
```diff
@@ -6,7 +6,7 @@
 Icon=<%= name %>
 <% } %>Type=Application
 StartupNotify=true
-StartupWMClass=<%= productName %>
+StartupWMClass=Studio
```
ivan-ottinger, line 11: What if we added a short comment like this one? Just to ensure this won't get changed accidentally in the future:
```suggestion
# Must match package.json productName, or the launcher icon won't group with the running window.
StartupWMClass=Studio
```

apps/studio/forge.config.ts
ivan-ottinger: Good catch!

I think for now we could keep `Meet Studio - a fast, free way to develop locally with WordPress.` - since otherwise we would have `WordPress` mentioned two times there:

`Meet WordPress Studio - a fast, free way to develop locally with WordPress.`

We could later decide to improve the description and change it on both Microsoft Store and App Center at once.

ivan-ottinger (approved):
Thank you for the fix, Rahul! The changes look good and work as expected. 👍🏼 

Image

One thing I noticed is that the `Developer website` link does not work. What I think we could do is to use the Studio site link there:

```diff
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -5,6 +5,7 @@
      "productName": "Studio",
      "version": "1.8.2",
       "description": "Local WordPress development environment u
Playgrounds",
+     "homepage": "https://developer.wordpress.com/studio/",
      "license": "GPL-2.0-or-later",
      "main": "dist/main/index.js",
```

I left two comments below as well.

apps/studio/forge.config.ts
gavande1: Makes sense — the doubled "WordPress" would read awkwardly. Happy to keep it as `Meet Studio …` and revisit alongside the MS Store copy when we next touch it. Thanks!

apps/studio/installers/desktop.ejs
gavande1: Good idea — pushed the comment on the branch. One small tweak to the wording: the value needs to match Electron's runtime WMClass (derived from `app.getName()`, which defaults to `package.json.name`), not `productName`. That's actually why this had to be hardcoded — when `Name=` flipped to `WordPress Studio`, templating from `productName` would have broken the launcher grouping.

Final form:
```
# Must match Electron's runtime WMClass (derived from app.getName(), i.e. package.json.name).
# Don't template from productName — the launcher icon won't group with the running window.
StartupWMClass=Studio
```

ivan-ottinger:
Thanks for the updates, Rahul!

What do you think about adding that Studio site link as well?

> One thing I noticed is that the Developer website link does not work. What I think we could do is to use the Studio site link there:

```diff
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -5,6 +5,7 @@
      "productName": "Studio",
      "version": "1.8.2",
       "description": "Local WordPress development environment u
Playgrounds",
+     "homepage": "https://developer.wordpress.com/studio/",
      "license": "GPL-2.0-or-later",
      "main": "dist/main/index.js",
```

gavande1:
Thanks for spotting that, @ivan-ottinger — good catch, we missed it in the previous round. Applied in a5728af6 exactly as you suggested. Putting it in `package.json` (vs. `MakerDeb.options`) also means any future Linux packaging — Snap, Flatpak, RPM — will pick the homepage up automatically.

PR #3336, opened 2026-05-04
CLI agent: align pi-tui with other pi-* deps and restore annotation tools

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Authored end-to-end with Claude in an interactive session. I drove the high-level decisions (pi-tui bump, deciding the patch was obsolete, opting in to prompt history, restoring the annotation tools) and Claude handled discovery, edits, and verification. Reviewer focus areas:
- The `CombinedAutocompleteProvider` now receives `process.cwd()` as a basePath — confirm that's the right scope for `@`-mention file completion.
- Verify the new annotation tool definitions in `apps/cli/ai/tools/open-annotation-browser.ts` and `apps/cli/ai/tools/wait-for-annotations.ts` match the originals (they're byte-identical copies of the pre-#3328 versions from `tools.ts`, just split per-file).

## Proposed Changes

- **Bump `@mariozechner/pi-tui` from `0.54.0` → `0.70.2`** so it matches the other `@mariozechner/pi-*` packages (which were already on `0.70.2`). The new `CombinedAutocompleteProvider` constructor takes a `basePath` argument, so we pass `process.cwd()`. This also enables `@`-mention file autocomplete in the chat prompt for free.
- **Wire up prompt history.** pi-tui's `Editor` exposes `addToHistory()`; we now call it on every successful submit (covers both the immediate-resolve and queued-mid-turn paths). `↑` on an empty editor recalls the most recent prompt. The site-picker shortcut on `↓` still wins — pi-tui only navigates history forward once you're already in browse mode.
- **Delete `patches/@mariozechner+pi-tui+0.54.0.patch`.** Its only purpose was dropping `chalk` from pi-tui's declared deps. With `pi-ai` and `pi-coding-agent` 0.70.2 also depending on (and actually using) chalk, patching just pi-tui can no longer keep chalk out of `node_modules` — the patch became a no-op blocker on `npm install`.
- **Restore the annotation agent tools.** PR #3328 split `apps/cli/ai/tools.ts` into one-file-per-tool under `apps/cli/ai/tools/`, but `open_annotation_browser` and `wait_for_annotations` were never migrated. The `/annotate` skill calls those tools by name; without them the agent ends up grabbing at unrelated MCP tools. Re-added as `apps/cli/ai/tools/open-annotation-browser.ts` and `apps/cli/ai/tools/wait-for-annotations.ts`, registered in `tools/index.ts`.

## Testing Instructions

[...]

PR #3338, opened 2026-05-04
CLI agent: scope @-mention autocomplete to the active site path

lezama:
## Related issues

Follow-up to #3336.

## Proposed Changes

When the user picks a different site mid-session via the site picker, the chat editor's `@`-mention file completion now follows. Today, the `CombinedAutocompleteProvider` is constructed once at `AiChatUI` startup with `process.cwd()` and is never updated, so completion stays anchored to the launch directory regardless of which site is active.

- New private `updateAutocompleteBasePath()` on `AiChatUI`. Picks `_activeSite.path` for local sites, falls back to a captured `initialCwd` (the launch cwd) for no-site and remote-site cases — remote sites have no local files to complete against.
- pi-tui's `CombinedAutocompleteProvider` exposes `basePath` only via the constructor — there's no public setter — so the helper replaces the whole provider on each call. The provider has no observable state beyond config, so reconstruction is cheap.
- Hook into the three site-mutation paths: the `activeSite` setter, `setActiveSite`, and `clearActiveSite`. The constructor also now goes through the helper instead of an inline `new CombinedAutocompleteProvider(…, process.cwd())`.

## Testing Instructions

1. `npm install`
2. `npm run typecheck` — passes.
3. `npm test -- apps/cli/ai/tests/ui.test.ts` — 18 tests pass, including 4 new ones covering the basePath logic.
4. Build the CLI and exercise the chat:
   ```
   npm run cli:build
   cd ~                             # launch from somewhere that is NOT a Studio site
   node /path/to/apps/cli/dist/cli/main.mjs
   ```
   - Type `@` → fuzzy completion is scoped to `~` (the launch cwd). Cancel with Esc.
   - Open the site picker, pick a local site (`my-site`).
   - Type `@` → fuzzy completion is now scoped to `~/Studio/my-site/` and lists files inside the site.
   - Open the picker again, pick a different local site (`other-site`).
   - Type `@` → fuzzy completion now lists files in `~/Studio/other-site/`.
   - Open the picker, pick a remote (WordPress.com) site.
   - Type `@` → fuzzy completion falls back to `~`, since remote sites have no local files.

## Pre-merge Checklist

- [x] Tests run and pass locally.


apps/cli/ai/ui.ts
```diff
@@ -910,8 +928,7 @@ export class AiChatUI implements AiOutputAdapter {
 
 	setActiveSite( site: SiteInfo, options: { announce?: boolean; emitEvent?: boolean } = {} ): void {
 		const { announce = true, emitEvent = true } = options;
-		this._activeSite = site;
-		this.editor.activeSiteName = site.name;
+		this.activeSite = site;
```
youknowriad, line 931: Is this like an unrelated cleanup or something?

youknowriad (approved):

apps/cli/ai/ui.ts
lezama: Adjacent dedup, not random — it's part of "funnel site mutations through the `activeSite` setter" in this commit. The setter already does `_activeSite = site` + `editor.activeSiteName = …` + `updateAutocompleteBasePath()`, so `setActiveSite`/`clearActiveSite` were duplicating the first two and the original PR added the third in three places. Routing them through the setter means a single source of truth for "what happens when activeSite changes" — future code paths that mutate it can't forget to update the basePath. Happy to split into its own commit if you'd rather.

PR #3342, opened 2026-05-05
apps/ui: auto-open site preview when the agent navigates

youknowriad:
## How AI was used in this PR

Built end-to-end with Claude Code in an exploratory back-and-forth: first a minimal fix (lift the agent-event subscription out of `SitePreview`), then an architectural pass to land a reusable shape for future *agent → UI* behaviors. Reviewers should sanity-check (a) that `useSessionCommands` is the only place agent events reach UI state, and (b) that the `SessionUIState` schema looks right as the place to grow new slices.

## Proposed Changes

When the agent calls `preview_navigate` / `preview_reload`, the site preview now auto-opens (it was previously dropped if the panel was collapsed) and lands at the requested path.

To get there, the wiring is reorganised into three small layers:

- `SessionUIProvider` + `useSessionPreviewUI` (`apps/ui/src/hooks/use-session-ui.tsx`) — session-scoped UI store backed by a typed reducer. `SessionUIState` is where future agent-driven UI slices plug in. State and dispatch contexts are split so dispatch-only consumers don't re-render on state changes.
- `useSessionCommands` (`apps/ui/src/hooks/use-session-commands.ts`) — a pure bridge: one subscription, no state of its own. Translates `preview.command` events into store actions. Adding a new agent → UI behavior is one `if (event.type === …) dispatch(…)` block.
- `SessionView` is now `<SessionUIProvider><SessionViewContent /></SessionUIProvider>`. `SitePreview` becomes a controlled view of `path` / `reloadNonce` and no longer subscribes to the agent stream itself.

The user-visible bug fix is the auto-open + correct-path behavior; the rest is the structure that makes the next such command (composer focus, modals, …) a one-slice extension.

## Testing Instructions

1. Start Studio with `npm start` and open a chat session whose owner site is local.
2. Collapse the site preview panel.
3. Ask the agent to navigate to a non-root path — e.g. *"open /wp-admin"*.
4. ✅ The preview panel should auto-open and load the requested path (not `/`).
5. Toggle the panel closed again and ask the agent to reload — it should re-open and reload.
6. Switch the session's environment to live (where `canTogglePreview` is false). Trigger a navigate from the agent. The store records the intent but the panel stays hidden; switching back to local opens it at the right path.

## Pre-merge Checklist

### DIFF
--- a/apps/ui/src/components/session-view/index.tsx
+++ b/apps/ui/src/components/session-view/index.tsx
@@ -3,15 +3,7 @@ import { useQueryClient } from '@tanstack/react-query';
 import { __ } from '@wordpress/i18n';
 import { IconButton } from '@wordpress/ui';
 import { clsx } from 'clsx';
-import {
-	useCallback,
-	useLayoutEffect,
-	useMemo,
-	useRef,
-	useState,
-	type ReactNode,
-	type Ref,
-} from 'react';
+import { useCallback, useLayoutEffect, useMemo, useRef, type ReactNode, type Ref } from 'react';
 import {
 	formatAnnotationsAsPrompt,
 	formatAnnotationsSubmittedMessage,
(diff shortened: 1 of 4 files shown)


youknowriad:
That's a nice refactor indeed :)

PR #3343, opened 2026-05-05
Make `native-php` E2E tests pass on Windows

fredrikekelund:
## Related issues

- Fixes RSM-2326

## How AI was used in this PR

Codex was used to help diagnose race condition issues in `collectDirectoryMetadata` and to fix an issue in `blueprints.phar` given error logs from a local Windows VM. I also used it to add the `native-php` E2E testing config for Buildkite. Claude was used to debug the persistently failing E2E tests and came up with the OPcache directory fix.

## Proposed Changes

- Fix an OPcache configuration issue that was causing some E2E tests to fail on Windows.
- Kill the PHP child processes spawned by `php-server-child.ts` on Windows, too.
- Temporarily include a precompiled `blueprints.phar` with a bugfix needed to make `blueprints.test.ts` pass. This can be removed when https://github.com/WordPress/php-toolkit/pull/265 has landed and there's a new php-toolkit release. I advise landing this PR first and reverting this change later.
- Make `collectDirectoryMetadata` more robust against race conditions.

## Testing Instructions

E2E tests should pass on both platforms.

## Pre-merge Checklist


apps/cli/process-manager-daemon.ts
```diff
@@ -175,6 +175,7 @@ export class ProcessManagerDaemon {
 					payload: {},
 				};
 			case 'kill-daemon':
+				await this.beginShutdownByKillingChildren( 'kill-daemon' );
```
fredrikekelund, line 178: Splitting the `shutdown` method into `beginShutdownByKillingChildren` and `finalizeShutdownByClosingSocketServersAndExiting` isn't necessary for the core purpose of this PR: getting E2E tests to pass on Windows. It does help with correctness, though, because it makes the `site stop --all` command complete after the child processes are closed instead of before.

scripts/download-wp-server-files.ts
fredrikekelund, line 1: We should eventually revert this change, but I'm personally open to landing this PR as is and reverting this once https://github.com/WordPress/php-toolkit/pull/265 has landed and we have a new php-toolkit release.

apps/cli/process-manager-daemon.ts
```diff
@@ -454,28 +476,39 @@ export class ProcessManagerDaemon {
...
+		return this.shutdownPromise;
+	}
 
-		await new Promise< void >( ( resolve ) => {
-			void this.controlServer.close().then( () => resolve() );
-		} );
+	private async finalizeShutdownByClosingSocketServersAndExiting(): Promise< void > {
+		await this.controlServer.close();
 		await this.eventsServer.close();
+		process.exit( 0 );
```
fredrikekelund, line 516: I have no indication that the process manager daemon didn't die after executing `site stop --all`, but some defensiveness here doesn't hurt.

fredrikekelund:
The ~250MB increase in installer size must be some kind of fluke… Let's see what CI reports now that I've merged trunk again.

EDIT: Confirmed that this was the case.

apps/cli/process-manager-daemon.ts
```diff
@@ -419,18 +420,39 @@ export class ProcessManagerDaemon {
...
+					windowsHide: true,
+					stdio: 'ignore',
+				} );
+				return;
+			}
+			// Console apps on Windows have no SIGTERM equivalent — `child.kill( 'SIGTERM' )`
+			// maps to TerminateProcess of a single PID, so neither cleanup nor tree-walk runs.
+			// Closing the IPC channel triggers the wrapper's 'disconnect' handler instead, which
+			// kills the PHP child and exits cleanly. Force escalation falls back to taskkill /T.
+			if ( managedProcess.child.connected ) {
```
bcotrim, line 451: Should we add a `disconnect` handler to playground child as well?

bcotrim (approved):
LGTM 👍

apps/cli/process-manager-daemon.ts
fredrikekelund: We always call `child.kill()` anyway, so there's no need for Playground child processes to handle the `disconnect` event (the only logical way for them to handle it would be to exit, as I see it). With native PHP, this approach is necessary for the child process to "forward" the kill signal to the PHP process.

PR #3359, opened 2026-05-06
[WIP] Remove `php-server-child` and manage native PHP child processes directly

fredrikekelund:
## Related issues

- Related to https://github.com/Automattic/studio/pull/3343. Includes the changes from that PR

## How AI was used in this PR

Codex was used extensively to drive the implementation of this PR. 

## Proposed Changes

> [!IMPORTANT]
> Despite not being a draft, this PR isn't fully ready for review yet. That's because E2E tests won't trigger on CI if it's a draft.

This is a large refactor that makes it so the process manager daemon only has to deal with direct child processes. With the previous approach, `php-server-child.ts` would spawn PHP child processes, which became grandchildren of the process manager daemon. These are not as simple to manage as direct child processes.

With this PR, consumers tell the process manager daemon to spawn a _type_ of process rather than a specific JS script. The process types have separate implementations that conform to a `ManagedProcess` interface.

## Testing Instructions

TBD

## Pre-merge Checklist


fredrikekelund:
I embarked on this thinking that the E2E tests on Windows were maybe having trouble killing the PHP grandchild processes. However, ultimately, that wasn't the case. Some parts of this implementation are good, but we could go even further and achieve a clearer system if we only had to deal with a single runtime type. To that end, I'll close this PR for now and revisit this in a month or so.

PR #3337, opened 2026-05-04
Unify CLI agent on a single pi-based runtime

youknowriad:
## Related issues

- Related to #3328 (which landed GPT 5.5 via the pi runtime; this PR retires the parallel Claude Agent SDK runtime so both families share one path)

## How AI was used in this PR

Built end-to-end with Claude Code. Each step was driven by an explicit prompt and validated against the live wpcom proxy with a real Sonnet 4.6 tool call before moving on. The unification was settled with a small spike test ([apps/cli/ai/runtimes/pi/tests/anthropic-pi-spike.test.ts](apps/cli/ai/runtimes/pi/tests/anthropic-pi-spike.test.ts), opt-in via `STUDIO_LIVE_SPIKE=1`) before touching the runtime.

What to verify carefully:
- The wpcom proxy auth path (Bearer via injected `Anthropic` client). The spike covers it; manual `studio code` + `/login` is the most realistic check.
- `studio mcp` still serves Studio's tools to external clients (Claude Desktop / Cursor) after the rewrite to the low-level MCP server API.
- The Skill tool is registered (look for `Skill` in the tool list when running with a local site). The bundled-build static-copy path was a real bug discovered during this work — `dist/cli/skills/site-spec/SKILL.md` must exist after `npm run cli:build`.
- Reasoning is on for both families: Anthropic uses adaptive thinking with effort `'high'`; GPT-5.5 sends `reasoning_effort: 'high'`.

## Proposed Changes

- **Single runtime, both families.** [apps/cli/ai/runtimes/pi/](apps/cli/ai/runtimes/pi/) now serves Anthropic Messages and OpenAI Chat Completions via `@mariozechner/pi-ai`, dispatched off the model's family. The custom Anthropic runtime ([apps/cli/ai/runtimes/anthropic/](apps/cli/ai/runtimes/anthropic/)) is deleted; `pickRuntime`, the cross-family resume guard, and the SDK-cleanup `unhandledRejection` handler are gone.
- **wpcom Bearer auth.** pi-ai's default Anthropic path uses `x-api-key`, which the wpcom proxy rejects. We inject a pre-built `Anthropic` SDK client via a custom `streamFn` only on that path (Anthropic + `ANTHROPIC_AUTH_TOKEN`).
[...]

PR #3360, opened 2026-05-06
AI sessions: adopt pi-coding-agent SessionManager end-to-end

youknowriad:
## Related issues

- Builds on the pi-runtime adoption from #3337

## How AI was used in this PR

Refactor pair-programmed with Claude Code. Reviewed end to end before opening for review.

## Proposed Changes

The pi-runtime adoption (#3337) left two layers of legacy plumbing inherited from the pre-pi runtime:

1. The runtime synthesized Claude-Agent-SDK-shaped `SDKMessage` events from pi's native `AgentEvent`s (~250 lines of `make*` builders) so existing consumers could keep parsing the SDK shape.
2. A sidecar `*.openai-state.json` mirrored the agent's in-memory transcript next to the recorder's JSONL, because the JSONL itself stored synthetic-SDK events instead of native pi messages.

This PR cuts both: the runtime emits a small `AgentRuntimeEvent` union (pi events plus `compaction_start` / `compaction_end` / `turn_completed` — shapes match pi-coding-agent's `AgentSessionEvent` so a future `AgentSession` swap is drop-in), and the on-disk JSONL is now pi-coding-agent's `SessionEntry`-based format end to end. Studio metadata (site selection, agent questions, tool progress, etc.) rides as `studio.*` `CustomEntry` payloads in the same file.

**Migration**: a one-shot eager sweep at app launch (`apps/studio/src/index.ts` `app.on('ready')` and `apps/cli/commands/ai/index.ts` `runCommand` entry) walks `~/.studio/sessions/**` and rewrites any pre-pi `session.started v1` JSONL into pi format in place. Idempotent — pi-format files are skipped, so subsequent boots are no-ops. A lazy fallback in `openStudioSession` catches any file the eager sweep missed.

**Consumers rewritten natively** (no translator at any layer):
- `tools/common/ai/sessions/{summary,filter-events,active-site,store}.ts` and `models.ts.resolveSessionModel` consume `studio.*` custom entries and `model_change` entries directly.
- `apps/cli/ai/sessions/{replay,context}.ts` iterate `SessionManager.getEntries()`.
- `apps/cli/ai/runtimes/pi/index.ts` persists via `SessionManager.appendMessage` — no per-turn legacy-event encoder.
- `apps/cli/commands/ai/index.ts` writes Studio metadata via `appendStudioEntry<T>(sm, customType, data)`.
- `apps/studio` IPC writes `model_change` for UI-driven `/model` swaps via `appendModelChangeEntry` (pi's native discriminator, not a custom entry) so resume-context resolution and pi's own session tooling both pick the override up.
- `apps/ui/src/components/session-view/{conversation,composer,index}.tsx` and `apps/ui/src/data/queries/use-agent-run.ts` consume pi entries; optimistic mutations produce pi entries with synthetic ids that the next `run.exited` refetch replaces with disk-backed reals.

**Cache invalidation**: `LoadedAiSession.events: AiSessionEvent[]` becomes `entries: SessionEntryBase[]`. The renderer's `@tanstack/react-query-persist-client` `buster` key is bumped so persisted localStorage caches from prior builds are dropped on first load after deploy.

**Deletions:**
- `apps/cli/ai/runtimes/messages.ts` (synthetic SDKMessage type)
- `apps/cli/ai/runtimes/pi/persistence.ts` (sidecar)
- `apps/cli/ai/sessions/recorder.ts`
- `tools/common/ai/sdk-messages.ts`
- The `mcp__studio__` prefix-stripping shim in `tools/common/ai/tools.ts`
- Three legacy-paths test files

**What this unlocks (out of scope for this PR):**
- `AgentSession` adoption from pi-coding-agent. The runtime event shapes already match `AgentSessionEvent` and disk format is pi-managed, so the swap drops `apps/cli/ai/runtimes/pi/compaction.ts` (~130 lines) and gives us pi's auto-retry / branching / bash-execution queue for free. Estimated ~150 lines changed when we do it.
- Compaction summaries and `model_change` history now persist on disk via pi's session machinery (the legacy format couldn't represent them).

## Testing Instructions

- [ ] Resume an existing session created before this PR (legacy JSONL on disk). The eager migration should rewrite the file in place on first launch and the conversation should replay correctly. Inspect `~/.studio/sessions/<YYYY>/<MM>/<DD>/*.jsonl` — the first line should change from `{"type":"session.started","version":1,...}` to `{"type":"session","version":3,...}` after first open.
- [ ] Start a new session, exchange a few turns, kill the process, then resume. The transcript should hydrate from disk via pi's `SessionManager.buildSessionContext()` — there's no `*.openai-state.json` sidecar anymore.
- [ ] Open the desktop app's session sidebar with a mix of new (pi-format) and migrated sessions. Both should show populated `firstPrompt`, `ownerSitePath`, `activeEnvironment`, etc. (the symptom of the read-side asymmetry that earlier iterations of this PR had — fixed because every reader now consumes pi entries natively).
- [ ] Run `studio code --json "hello"`. Wire format keeps the legacy `'message'` envelope but the inner payload is the native `AgentRuntimeEvent`.
- [ ] Trigger compaction (long enough conversation to cross the threshold). The "Compacting conversation history…" loader and "Conversation history compacted" info line should still fire.
- [ ] Trigger an AI usage cap on the wpcom provider. The cap message should still appear — detection moved from scanning text blocks to checking pi's `AssistantMessage.errorMessage` for a 429 marker.
- [ ] Open a session that had no turns yet (just a header on disk). The conversation view should render empty without crashing — confirms the `data.entries ?? []` guard.

## Pre-merge Checklist

- [x] TypeScript / React / console errors checked (typecheck clean across all 4 workspaces)
- [x] 1534 vitest tests passing, 0 failures
- [ ] Manual smoke pass on the test plan above
- [ ] Code review


apps/studio/src/index.ts
```diff
@@ -279,6 +281,26 @@ async function appBoot() {
 			await launchExtensionBackgroundWorkers();
 		}
```
youknowriad, line 281: Can we limit the migration to apps/ui runs only an d not the legacy app?

tools/common/ai/sessions/entry-types.ts
```diff
@@ -0,0 +1,96 @@
+// Studio metadata rides as `studio.*` `customType` entries on top of pi's
```
youknowriad, line 1: Curious why all these types are needed, do these existing in pi directly?

apps/studio/src/index.ts
youknowriad: Good call — I dropped the eager sweep entirely in b47331f4b. `readPiFileEntries` already migrates each JSONL on first read, so the upfront pass was redundant work that also blocked startup until every session file had been touched.

tools/common/ai/sessions/entry-types.ts
youknowriad: Good catch — the `SessionEntryBase` mirror was fully redundant. b47331f4b drops it and imports `SessionEntry` (and `CustomEntry` for the `StudioCustomEntry` wrapper) straight from `@mariozechner/pi-coding-agent`. Only the Studio-specific custom-type discriminants and their data shapes live here now.

apps/cli/ai/runtimes/runtime-events.ts
```diff
@@ -0,0 +1,13 @@
...
+// aligned with what the broader pi ecosystem produces.
+
+import type { AgentEvent } from '@mariozechner/pi-agent-core';
+import type { AgentSessionEvent } from '@mariozechner/pi-coding-agent';
+
+export type CompactionStartEvent = Extract< AgentSessionEvent, { type: 'compaction_start' } >;
+export type CompactionEndEvent = Extract< AgentSessionEvent, { type: 'compaction_end' } >;
+export type CompactionReason = CompactionStartEvent[ 'reason' ];
+
+export type AgentRuntimeEvent = AgentEvent | CompactionStartEvent | CompactionEndEvent;
```
youknowriad: Is this override necessary now that we're using the same compaction as pi agentsession?

youknowriad: Good call — collapsed in 8ce192ee6. The whole file is now `export type AgentRuntimeEvent = AgentSessionEvent;` plus a couple of `Extract<>` aliases for the compaction members.

PR #3364, opened 2026-05-06
Fix .deb policykit dependency for Debian trixie / Kali

gavande1:
## Related issues

- Fixes [RSM-2433](https://linear.app/a8c/issue/RSM-2433/studio-deb-declares-retired-policykit-1-dependency-blocking-install-on)

## How AI was used in this PR

Claude Code applied the one-line dependency change proposed in the Linear issue and rewrote the inline comment to explain why the alternative is needed. I verified the build install on both Ubuntu 24.04 and Kali myself.

## Proposed Changes

- Switch the `.deb` `Depends` entry from `policykit-1` to `pkexec | policykit-1` so the package resolves on both legacy (Ubuntu 24.04 / older Debian) and current (Debian trixie / Kali rolling) polkit packaging. Both alternatives ultimately provide the `pkexec` binary that `@vscode/sudo-prompt` needs for hosts-file writes.
- Update the inline comment in `apps/studio/forge.config.ts` to document the split and why the alternative is required.

## Testing Instructions

1. Build the `.deb`: `npm run make` on Linux.
2. Install on **Ubuntu 24.04**: `sudo apt install ./out/make/deb/**/studio_*.deb` — should resolve `policykit-1` and install cleanly. ✅ Verified.
3. Install on **Kali Linux** (and/or Debian trixie): same command — should now resolve `pkexec` and install cleanly (previously failed with `Depends: policykit-1 but it is not installable`). ✅ Verified.
4. On each, run the **Trust Certificate** flow — the polkit prompt should appear and the system-bundle install should succeed.

## Pre-merge Checklist


ivan-ottinger (approved):
Good catch! The change looks good! I am currently building the app and will test it, but already approving the PR as I don't think there will be any issues.

ivan-ottinger:
Just noting that the fix is working well:

Before:

CleanShot 2026-05-11 at 12 22 40@2x

After:

CleanShot 2026-05-11 at 12 23 19@2x

CleanShot 2026-05-11 at 12 26 55@2x

PR #3366, opened 2026-05-06
remote-session: show agent output in attach (STU-1696)

epeicher:
## Related issues

- Related to STU-1696

## How AI was used in this PR

Implementation, tests, and commit messages drafted with Claude Code. I drove the design (what to log, log shape, where the fallback goes), reproduced the empty-SDK-result bug live against the daemon to confirm the diagnosis, and reviewed each diff before committing. Reviewers should focus on the persisted-event log shape (`logger.ts`), the diagnostic fields on `Turn outcome`, and the empty-result workaround in `turn-runner.ts`.

## Proposed Changes

**1. Show agent output on `studio code remote-session attach`** (commit 1)

`RemoteSessionLogger.event()` now persists conversation events (progress, info, error, replies, questions) to the daemon log file in addition to the existing stdout-mirror path. `formatLogLineForStdout()` recognizes the new `{ t, event, content }` shape and renders it as `[time] kind content`. Result: when you attach to a running daemon, the conversation flow is visible alongside the polling logs.

**2. Diagnostics for "no response" turns** (commit 2)

Each non-result `message` event emits one `agent.step` line per content block (text snippet, `tool_use` with input preview, `tool_result`, `thinking`) so you can see exactly what the agent did. The `Turn outcome` info log gained `event_counts`, `result_event_seen`, `result_event_empty_reply`, `turn_completed_seen`, `non_json_stdout_lines`, plus `subtype` / `stop_reason` / `num_turns` on the new `Result event had empty reply` warn. The poll loop logs `No reply produced; posting fallback` before sending the user-facing fallback so the cause is captured in the log file.

**3. Workaround for empty SDK result** (commit 2)

While testing the new attach output, reproduced an issue where the agent finishes cleanly (`status: success`, `stop_reason: end_turn`, `num_turns: 1`) but the SDK `result` event arrives with `result: ""`, even though an assistant text block was streamed earlier in the turn. The remote-session loop was posting "⚠️ Local agent did not return a result." to Telegram even though the agent did answer. Root cause is in the pi runtime (it hardcodes the result text to `''` on `agent_end`). As a guard, the turn runner now captures the last assistant text seen during the turn and uses it as the reply when the result event has empty `replyText` and no error. The proper fix at the runtime source will follow.

## Testing Instructions

Automated:

- `npm test -- apps/cli/remote-session` — 111 tests pass (110 existing + 1 new for the empty-result fallback).
- `npm run typecheck` — clean.

Manual:

1. `npm run cli:build && node apps/cli/dist/cli/main.mjs code remote-session start --detach` (with a configured bot/chat).
2. In a second terminal: `node apps/cli/dist/cli/main.mjs code remote-session attach`.
3. Send a message from Telegram. You should see: `Polled message`, then a stream of `progress` / `agent.step` / `reply` event lines, then `Turn outcome` and `Posting reply`.
4. To exercise the workaround, send a single-tool one-turn question like _"How many local sites are currently running?"_. The reply should land in Telegram (previously it sometimes posted "⚠️ Local agent did not return a result."). The `Turn outcome` line will show `result_event_empty_reply: true` and `used_assistant_text_fallback: true`.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/remote-session/logger.ts
+++ b/apps/cli/remote-session/logger.ts
@@ -129,6 +129,14 @@ interface PersistedLogLine {
 	level?: unknown;
 	msg?: unknown;
 	meta?: unknown;
+	event?: unknown;
+	content?: unknown;
+}
+
+function formatEventForStdout( time: string, kind: string, content: string ): string {
+	const safe = safeForStdout( content ).replace( /\n/g, '\n  ' );
+	const safeKind = stripTerminalControls( kind );
+	return `[${ time }] ${ safeKind } ${ safe }\n`;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


gcsecsey (approved):
Thank you @epeicher for improving the output here and for fixing the "no output" bug, I also ran into that a couple of times. 🙌 

Changes LGTM and test well. 👍 

| telegram | logs |
|--------|--------|
| Image | Image |

PR #3349, opened 2026-05-05
CLI: add eval-runner result diagnostics

chubes4:
## Summary
- Capture final SDK result metadata in eval-runner output: `resultSubtype`, `resultIsError`, `resultStopReason`, `resultText`, and `resultErrors`.
- Treat empty successful completions as failed evals when the SDK reports success but the run produced neither assistant text nor final result text.
- Add opt-in compact transcript diagnostics behind `STUDIO_EVAL_INCLUDE_TRANSCRIPT=1`, with text/tool-result truncation to avoid bloating default artifacts.

## Why
This continues the eval-runner observability work from #3273 and #3330. Those PRs made phase/tool timings, first tool errors, loop exceptions, and timeouts visible. This adds the final SDK result shape and an opt-in turn transcript so eval consumers can distinguish model behavior, PI harness continuation behavior, runner classification, and downstream benchmark quality gates.

The need surfaced while testing the Static Site Importer draft path in #3309 with the Studio site-build benchmark. GPT-5.5 repeatedly produced tool-only runs for built-in prompt variants (`restaurant`, `wordpress-is-dead`): `site_list` / `site_info` returned successfully, then the SDK emitted `subtype: \"success\"`, `stopReason: \"end_turn\"`, and an empty `result`. No assistant text, `Write`, `wp_cli`, or import report was produced, but the eval runner classified the run as successful because it trusted `message.subtype === 'success'`.

With the local transcript diagnostics enabled, that failure shape was clear. The same diagnostics also helped compare Claude Sonnet 4.6 on the same SSI site-build flow: Claude generated source HTML and wrote files, then timed out while repairing generated helper-script errors before reaching import. Different failure mode, same need for better eval evidence.

This PR keeps the transcript opt-in so normal eval artifacts only gain a few scalar fields, while deeper debugging remains available when investigating model/runtime/harness regressions.

## Validation
- `npm install` to bootstrap the clean worktree.
- `npm run cli:build --silent` — passed.
- `npx eslint apps/cli/ai/eval-runner.ts` — passed.
- `npm -w wp-studio run typecheck` — passed.
- `git diff --check` — passed.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (GPT-5.5)
- **Used for:** Diagnosing the eval-runner artifact gap during SSI benchmark runs, drafting the result metadata / opt-in transcript patch, running local validation, and preparing this PR body. Chris reviewed the failure evidence and PR framing.

PR #3273, opened 2026-04-28
CLI: add eval-runner diagnostics

chubes4:
## Summary
- Add phase timing diagnostics to `eval-runner` so benchmark runs can separate provider resolution, runtime startup, first assistant latency, and total eval time.
- Add structured tool timing events and first-tool-error metadata while preserving the existing raw tool call/result/text output.

## Why
This makes Studio agent evals more useful for runtime and tool-path comparisons. The runner now exposes enough structure for benchmark tooling to verify that faster runs still did the expected work, instead of only comparing wall-clock totals.

This supports the eval-runner diagnostics work tracked in #3262.

## Benchmark context
I validated this with [Homeboy](https://github.com/Extra-Chill/homeboy), a local/CI benchmarking CLI I use to run repeatable workloads against pinned git worktrees (“rigs”). The shared Studio benchmark workloads live in the private [`chubes4/homeboy-rigs`](https://github.com/chubes4/homeboy-rigs) repo; the relevant rig update is [`chubes4/homeboy-rigs#7`](https://github.com/chubes4/homeboy-rigs/pull/7).

The important part for this PR is not Homeboy itself, but the contract it exercised: the eval runner now emits phase timings, tool-event counts, and tool durations that external eval tooling can assert on.

## Testing
- `npm run cli:build --silent`
- `npm -w wp-studio run typecheck`
- `homeboy bench --rig studio-agent-sdk-evaltiming --scenario studio-agent-runtime --iterations 5 --runs 3 --warmup 1 --json-summary --ignore-default-baseline`
- `homeboy bench --rig studio-agent-sdk-evaltiming --scenario studio-agent-site-info --iterations 5 --runs 3 --warmup 1 --json-summary --ignore-default-baseline`
- `homeboy bench --rig studio-agent-sdk-evaltiming --scenario studio-agent-runtime --scenario studio-agent-site-info --iterations 2 --runs 2 --warmup 1 --json-summary --ignore-default-baseline`
- `homeboy bench --rig studio-agent-sdk-evaltiming,studio-agent-pi-evaltiming --scenario studio-agent-runtime --iterations 2 --runs 2 --warmup 1 --json-summary --ignore-default-baseline`
- `homeboy bench --rig studio-agent-sdk-evaltiming,studio-agent-pi-evaltiming --scenario studio-agent-site-info --iterations 2 --runs 2 --warmup 1 --json-summary --ignore-default-baseline`

## AI assistance
- **AI assistance:** Yes
[...]

PR #3376, opened 2026-05-07
CLI: add eval-runner result diagnostics

chubes4:
## Summary

- Capture final SDK result metadata in eval-runner output: `resultStopReason`, `resultText`, `resultErrorMessage`, `resultUsage`, and `interrupted`. All derived from `findLastAssistant(event.messages)` on `agent_end` so they reflect the same source of truth as `success`.
- Add `producedNoUsefulOutput`, a heuristic that flags runs where the agent reported `success` but produced neither assistant text nor a successful mutating-tool call. The motivating failure mode is GPT-5.5 ending a turn after read-only tool calls (`site_list`, `site_info`) without writing anything or explaining anything — runs that classify as `success` today but produce no import report, no assistant text, and no `Write` / `wp_cli` activity.
- Add opt-in compact transcript diagnostics behind `STUDIO_EVAL_INCLUDE_TRANSCRIPT=1`. The transcript records every `AgentSessionEvent` (including `compaction_*` and `auto_retry_*`) with text / tool-result truncation, so default artifacts stay small while deeper debugging remains available when investigating model / runtime / harness regressions.

## Why

This continues the eval-runner observability work from #3273 and #3330. Those PRs made phase / tool timings, first tool errors, loop exceptions, and timeouts visible. This adds the final SDK result shape and an opt-in turn transcript so eval consumers can distinguish model behavior, pi-coding-agent harness continuation behavior, runner classification, and downstream benchmark quality gates.

The need surfaced while testing the Static Site Importer draft path in #3309 with the Studio site-build benchmark. GPT-5.5 repeatedly produced tool-only runs for built-in prompt variants (`restaurant`, `wordpress-is-dead`): `site_list` / `site_info` returned successfully, then the agent ended the turn with no assistant text, no `Write`, no `wp_cli`, and no import report. The eval runner classified the run as successful because the run terminated cleanly. With these diagnostics, those runs are now mechanically detectable as `producedNoUsefulOutput: true`.

## Replaces #3349

This PR replaces #3349, which was written against the pre-#3360 `SDKMessage` event model. After #3360 landed (`AI sessions: adopt pi-coding-agent SessionManager end-to-end`), the original PR could not be rebased — every event-handling code path it touched changed shape:

- `startAiAgent()` async iterator → `runStudioAgentTurn()` callback
- `SDKMessage` (`assistant` / `user` / `result`) → `AgentSessionEvent` (`message_end` / `tool_execution_end` / `turn_end` / `agent_end` / `compaction_*` / `auto_retry_*`)
- Final `result` SDKMessage with `subtype` / `stop_reason` / `result` / `is_error` / `errors` → no terminal `result` message at all; the canonical end-state lives on `findLastAssistant(event.messages).stopReason` plus the helper `getAgentEndTurnResult()` exposed by the new `@studio/common/ai/session-events` module

This rewrite is small (+161 / -2 lines, single file) and aligned with the new event surface. Closing #3349 in favor of this PR.

## Validation

- `npx eslint apps/cli/ai/eval-runner.ts` — clean
- `npm -w wp-studio run typecheck` — clean
- `npm run cli:build --silent` — clean, `apps/cli/dist/cli/eval-runner.mjs` produced
- **Direct smoke run** (no tools, identity prompt): all new fields populate correctly; `producedNoUsefulOutput: false` because the assistant produced text.
- **Single-tool eval** (`studio-agent-site-info.bench.mjs`, read-only `site_info`): tool calls and truncated tool results captured in transcript; `producedNoUsefulOutput: false`.
- **Multi-turn mutating run**: `Create a new Studio site named X. Then use wp_cli to set the blog title to Eval Validation. Confirm by saying done.` — agent called `site_create` and `wp_cli` across 3 turns, generated a real local Studio site (cleaned up afterward), `producedNoUsefulOutput: false`, `resultUsage` reported input / output / cache tokens and cost.
- **Read-only tool-only run**: `Use only site_list and respond with nothing else. Do not say anything.` — the agent called `site_list`, produced no text, `success: true`, `producedNoUsefulOutput: true`. This is the exact failure mode the heuristic is designed to flag.
- Existing `studio-agent-runtime.bench.mjs` and `studio-agent-site-info.bench.mjs` ran end-to-end against the new runner with no breakage. The benches don't yet consume the new fields; that's a separate change in the rig.

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** Claude Code (Sonnet 4.5)
- **Used for:** Rewriting the diagnostics layer against the new `AgentSessionEvent` / pi-coding-agent `SessionManager` model that landed in #3360, running the validation matrix above, and drafting this PR description. Chris reviewed the failure-mode evidence and remains responsible for the change.

### DIFF
--- a/apps/cli/ai/eval-runner.ts
+++ b/apps/cli/ai/eval-runner.ts
@@ -19,6 +19,7 @@ import {
 } from 'cli/ai/auth';
 import { runStudioAgentTurn } from 'cli/ai/runtimes/pi';
 import { STUDIO_SITES_ROOT } from 'cli/lib/site-paths';
+import type { StopReason, Usage } from '@mariozechner/pi-ai';
 import type { AgentSessionEvent } from '@mariozechner/pi-coding-agent';
 import type { AiProviderId } from 'cli/ai/providers';
 
@@ -27,6 +28,61 @@ interface EvalRunnerInput {
 	timeoutMs?: number;
 	model?: AiModelId;
 }
+
+// Tool names that indicate the agent did "real work" (mutated state, ran a
+// command, scaffolded files, etc.). Used as a heuristic to detect runs where
+// the agent reported success but produced no assistant text and never invoked
+// a state-changing tool — typically a "false success" failure mode where the
+// model trivially ended a turn without addressing the prompt.
+const MUTATING_TOOL_NAMES = new Set( [
+	'Write',
+	'Edit',
+	'Bash',
+	'wp_cli',
+	'site_create',
+	'site_delete',
+	'site_import',
+	'site_export',
+	'site_pull',
+	'site_push',
+	'site_start',
+	'site_stop',
+	'scaffold_theme',
+	'install_taxonomy_scripts',
+	'preview_create',
+	'preview_delete',
+	'preview_update',
+	'wpcom_request',
+] );
+
+// Cap individual text/tool-result strings inside the opt-in transcript so a
+// single huge tool output doesn't explode the eval artifact.
+const TRANSCRIPT_TEXT_MAX_LENGTH = 4000;
+
+function truncateText( value: string, maxLength = TRANSCRIPT_TEXT_MAX_LENGTH ): string {
+	return value.length > maxLength
+		? `${ value.slice( 0, maxLength ) }…[truncated ${ value.length - maxLength } chars]`
+		: value;
+}
+
+type TranscriptEvent = {
+	index: number;
+	type: AgentSessionEvent[ 'type' ];
+	turnIndex: number;
+	elapsedMs: number;
+	text?: string[];
+	toolCalls?: ToolCallRecord[];
+	toolResult?: {
+		toolUseId: string;
+		toolName: string | null;
+		isError: boolean;
+		text?: string;
+	};
+	stopReason?: StopReason;
+	errorMessage?: string;
+	compaction?: { reason: string; aborted?: boolean };
+	autoRetry?: { attempt: number; success?: boolean; error?: string };
+};
 
 function extractToolCalls( event: AgentSessionEvent ) {
 	if ( event.type !== 'message_end' || event.message.role !== 'assistant' ) {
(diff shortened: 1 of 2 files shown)


apps/cli/ai/eval-runner.ts
```diff
@@ -28,6 +29,61 @@ interface EvalRunnerInput {
...
 
+// Tool names that indicate the agent did "real work" (mutated state, ran a
+// command, scaffolded files, etc.). Used as a heuristic to detect runs where
+// the agent reported success but produced no assistant text and never invoked
+// a state-changing tool — typically a "false success" failure mode where the
+// model trivially ended a turn without addressing the prompt.
+const MUTATING_TOOL_NAMES = new Set( [
+	'Write',
+	'Edit',
+	'Bash',
```
youknowriad, line 37: Makes me think we should normalize how do we name tools, these camel case names are just named like that because we inherited them from claude initially.

chubes4: Seems like a good idea. The Pi harness is a great direction. 

One question i had about this PR was: Do we want to keep the full transcript opt-in, or enable it by default? 

It can be argued that if you are doing evals, the transcript is always helpful.

youknowriad: To be honest, I didn't think about evals too much yet (as much as you did haha), so feel free to go in any direction you think is good.

chubes4: Sounds good. I will try it both ways and see what feels better.

These evals will help a lot as we add more features, more models, and evolve the system prompt.

chubes4:
Added one follow-up commit to make the native eval runner safe for detached/headless benchmark use.

Context: we have been working on a matrix eval runner that can test our different site-generation methods side by side and collect comparable metrics/evidence. The goal is to help us decide with data which approach is most effective across Studio Code, Studio Web, and other generation paths.

The Workflow Bench adapter runs Studio Code's native eval runner from a disposable per-attempt workspace. Before this follow-up, the eval runner still used Studio's default `~/Studio` site root through `STUDIO_SITES_ROOT`, so `site_create` could collide with existing local Studio state during benchmark runs. The new commit keeps normal Studio behavior unchanged, but lets the eval runner opt into the current working directory as its site root when no explicit `STUDIO_SITES_ROOT` is provided.

Verification:
- `npm -w wp-studio run typecheck`
- `npx vitest run apps/cli/ai/tests/pi-runtime.test.ts apps/cli/commands/site/tests/create.test.ts`
- `npm run cli:build --silent`

wojtekn:
Marking as draft until conflicts and tests are fixed.

PR #3393, opened 2026-05-08
Download PHP CLI binaries from Apps CDN

bcotrim:
## Related issues

- Fixes RSM-2592
- Fixes RSM-2133

## How AI was used in this PR

AI helped inspect the existing native PHP binary download path, draft the Apps CDN manifest resolver, and update the focused tests. I reviewed the final diff and trimmed CI publishing changes into the separate upload PR.

## Proposed Changes

- Resolve native PHP CLI archives from the Apps CDN `WordPress.com Studio PHP CLI` manifest instead of the upstream static PHP CDN.
- Require a manifest URL and SHA for the exact PHP patch/platform/arch entry, with a user-friendly error if unavailable.
- Verify downloaded archives against the manifest SHA before extraction.
- Extract archives based on file extension so signed macOS `.zip` artifacts and legacy `.tar.gz` archives both work.
- Update the local PHP binary download helper and native PHP binary docs.

## Testing Instructions

- `npx eslint --fix tools/common/lib/php-binary-metadata.ts apps/cli/lib/dependency-management/php-binary.ts apps/cli/lib/dependency-management/tests/php-binary.test.ts scripts/download-php-binary.ts`
- `npm test -- apps/cli/lib/dependency-management/tests/php-binary.test.ts`
- `npm run typecheck`
- `git diff --check`

## Pre-merge Checklist


apps/cli/lib/dependency-management/tests/php-binary.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+function mockFetchJson( json: unknown ): void {
+	vi.stubGlobal(
+		'fetch',
+		vi.fn().mockResolvedValue( {
+			ok: true,
+			json: vi.fn().mockResolvedValue( json ),
+		} )
+	);
+}
```
fredrikekelund: I say let's use `nock` here instead. This mocking is pretty simple, but using `nock` is equally simple and helps decouple the tests from the implementation.

apps/cli/lib/dependency-management/php-binary.ts
```diff
@@ -104,28 +101,45 @@ async function downloadAndInstall(
...
-				`Cannot verify binary integrity. Run: shasum -a 256 ${ filePath }`
-		);
+): Promise< PhpBinaryDownloadInfo > {
+	const downloadInfo = getPhpBinaryDownloadInfo( version, platform, arch );
+	if ( downloadInfo ) {
+		return downloadInfo;
 	}
 
+	throw new Error(
+		`PHP ${ version } is not available for this device yet. Please try again later.`
```
fredrikekelund:
```suggestion
		`PHP ${ version } is not available for this platform yet.`
```
Minor, but "try again later" seems a bit misleading here. This effectively means that Studio is broken on the current device.

apps/cli/lib/dependency-management/php-binary.ts
fredrikekelund, line 1: We already discussed that, but we should coerce the given PHP version to the `NativePhpSupportedVersion` range somewhere in this file. If a Studio site is configured to use PHP 8.0, we should just log a warning and use PHP 8.1 instead.

apps/cli/lib/dependency-management/php-binary.ts
```diff
@@ -134,37 +146,33 @@ async function extractAndInstall(
...
 	destPath: string,
 	patchVersion: string,
-	platform: NodeJS.Platform,
-	arch: string
+	platform: NodeJS.Platform
 ): Promise< void > {
 	const isWindows = platform === 'win32';
-	const effectiveArch = isWindows ? 'x64' : arch;
 	const tmpDir = os.tmpdir();
+	const binaryName = isWindows ? 'php.exe' : 'php';
```
fredrikekelund: There's a `runtime.json` file in all PHP packages that contains a `binary` key with the executable filename. I think the current solution is fine, but we could also rely on that file.

fredrikekelund (approved):
Great work! This works beautifully 👍 I pushed two commits with some minor tweaks and some type tightening. 

This also effectively resolves RSM-2133, right? Because now, whatever PHP minor version the user has configured for their site, it'll just automatically resolve to a patch version based on the static manifest.

The PR description says this:

> Extract archives based on file extension so signed macOS .zip artifacts and legacy .tar.gz archives both work.

I guess that's a leftover from before, since all install packages are ZIP files now.

PR #3399, opened 2026-05-08
Replace placeholder note widget with desk-style sticky note

youknowriad:
## Summary
- Replaced the placeholder note widget with a desk-style sticky note built on `contentEditable` and `@wordpress/rich-text`
- Kept the existing widget architecture and persistence flow intact, while updating the note schema to use tones instead of the old color enum
- Matched the widget indicator styling to the note tone and updated the default user desk seed plus adapter tests

## Testing
- `npx eslint --fix` on the modified files
- `npm test -- apps/ui/src/ui-desks/user-desk/tldraw-adapter.test.ts`
- `npm run typecheck`
- `npm -w @studio/ui run build`

### DIFF
--- a/apps/ui/src/ui-desks/widgets/note/component.tsx
+++ b/apps/ui/src/ui-desks/widgets/note/component.tsx
@@ -1,6 +1,8 @@
+import { select, type AnyConfig, type StoreDescriptor } from '@wordpress/data';
 import {
 	__unstableUseRichText as useRichText,
 	registerFormatType,
+	store as richTextStore,
 	toggleFormat,
 	type RichTextValue,
 } from '@wordpress/rich-text';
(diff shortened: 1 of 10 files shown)


apps/ui/src/ui-desks/widgets/note/component.tsx
```diff
@@ -1,115 +1,156 @@
+import { create, registerFormatType, toHTMLString } from '@wordpress/rich-text';
```
youknowriad: The code in this component, seems to recreate the RichText component (or useRichText hook) that is available in that wordpress/rich-text package, it feels like it could be simplified a lot with the same result. There's also a format library package that we could be using from wordpress instead of inventing our own formats.

PR #3401, opened 2026-05-08
apps/ui: Fix desks chat model-switch dialog and preserve session expansion

youknowriad:
## Summary
- Fix the model-switch confirmation dialog padding by rendering the message through `Dialog.Content` / `Dialog.Description`.
- Keep the desks chat sidebar expanded after confirming a cross-family model switch and navigate to the newly created session.
- Upsert newly created session summaries into the sessions cache immediately so the UI does not briefly treat the new session as missing.
- Add a focused unit test for the session cache upsert helper.

## Testing
- `npx eslint --fix` on the touched UI files
- `npm test -- apps/ui/src/data/queries/use-sessions.test.ts`
- `npm run typecheck`


apps/ui/src/data/queries/use-sessions.test.ts
```diff
@@ -0,0 +1,49 @@
+import { QueryClient } from '@tanstack/react-query';
```
youknowriad: Remove this test please, feels useless to me.

PR #3403, opened 2026-05-08
Avoid writing new mu-plugins if old ones are still accessible

fredrikekelund:
## Related issues

- Related to RSM-1691

## How AI was used in this PR

Codex implemented the first version, which I then tweaked – specifically to use JSON for stable string de-/serialization between JS and PHP.

## Proposed Changes

With the `native-php` runtime, mu-plugins are loaded like so:

1. Create a temporary directory.
2. Write individual mu-plugin files to that directory.
3. Write a `wp-content/mu-plugins/99-studio-loader.php` file to the site directory, with a path to the temporary directory from before.
4. The loader file searches the temporary directory and includes all the files in it.

Previously, this process was executed not only when a server started, but also on every WP-CLI invocation. This is fine, but it's not very efficient.

This PR prepends the following steps to the previous process:

1. Assert that `wp-content/mu-plugins/99-studio-loader.php` exists.
2. Parse the temporary directory path from it.
3. Assert that the template for the loader file hasn't changed.
4. Assert that the temporary directory path contains the file names we would write.

If all assertions hold, we don't write anything to disk. If they don't, we run the setup process again.

## Testing Instructions

1. Run `STUDIO_RUNTIME=native-php npm start`
2. If you don't already have a site with `runtime: native-php`, create a new one
3. Start that site
4. Open `wp-content/mu-plugins/99-studio-loader.php` and take note of `$studio_mu_plugins_dir`
5. Run `node apps/cli/dist/cli/main.mjs wp core version`
6. Open `wp-content/mu-plugins/99-studio-loader.php` and ensure that `$studio_mu_plugins_dir` is the same as before

## Pre-merge Checklist


bcotrim (approved):
Worked as described, nice improvement!
LGTM 👍

PR #3406, opened 2026-05-08
remote-session: detect stale resume sessions reported via JSON error events

gcsecsey:
## Related issues

- Fixes STU-1711

## How AI was used in this PR

Claude Code investigated the daemon log together with me, traced the regression to PR #3360, and drafted the fix and tests. I drove the diagnosis and reviewed the diff before committing.

## Proposed Changes

After #3360, the session ID stored by the remote-session bridge from a previous build can no longer be resolved by the new SessionManager. `studio code --resume-session <id>` now fails immediately with `error No AI session found for resume ID: ...`. The poll loop already has a recovery path, gated on `outcome.staleSession`. That detection was bypassed because of where the failure surfaces:

- The new pi runtime reports the resume failure as a JSON `error` event on stdout, not on stderr.
- `detectStaleSession()` in `apps/cli/remote-session/turn-runner.ts` only inspected `stderrTail` and `replyText`, so the stale-session pattern (`/no ai session found for resume id/i`) never had a chance to match.
- Result: the poll loop fell through to the generic non-zero-exit branch and posted `Local agent error (no output).` to Telegram. The stale ID stayed in the remote-session state, so every subsequent turn hit the same error.

This PR:

- Captures the message from JSON `error` events into a local `lastErrorEventMessage` in `runTurn()`.
- Feeds it into `detectStaleSession()` alongside `stderrTail` and `replyText`. Cleaned up dead code in that function while there (the previous `isError` short-circuit had no effect on the result).
- Adds a `stale-resume-error-event` scenario to the studio-code mock that emits the failure as a JSON `error` event (matching the real production behavior we observed in the log).
- Adds a unit test asserting `outcome.staleSession === true` when the failure arrives via `error` event with empty stderr.

Once `staleSession` is correctly detected, the existing recovery path in `poll-loop.ts:143-160` clears the saved session, retries without `--resume-session`, and the new sessionId from the retry's `turn.completed` is persisted at `poll-loop.ts:170-172`. The stuck-on-stale-ID loop self-heals on the next turn.

## Testing Instructions

- `npm test -- apps/cli/remote-session` should report 112 passing tests (was 111).
- `npm run typecheck` should be clean.
- Manual:
  - `npm run cli:build && node apps/cli/dist/cli/main.mjs code remote-session start --detach` (with a configured bot/chat).
  - In a second terminal: `node apps/cli/dist/cli/main.mjs code remote-session attach`.
  - Pre-populate the remote-session state file under `~/.studio/` with a non-existent `session_id` (or just have a session_id from before #3360).
  - Send a message from Telegram. Expected before this PR: `Local agent error (no output)`. Expected after: an `info Resume failed; retrying without session_id` line in the log, a `Session expired; started a new one.` message in Telegram, and the agent's actual reply right after.
  - The next turn should resume cleanly against the new sessionId.

| State | Telegram | Daemon log |
| --- | --- | --- |
| Before fix (stale resume) | image | CleanShot 2026-05-08 at 16 17 44@2x |
| After fix (auto-recovery) | CleanShot 2026-05-08 at 16 20 35@2x | CleanShot 2026-05-08 at 16 23 11@2x |

## Pre-merge Checklist


epeicher (approved):
Thanks @gcsecsey, this is working as expected. Changes also LGTM!

Image

PR #3410, opened 2026-05-09
Share the desk shell across user and site desks

youknowriad:
## Summary
- Introduced a shared `Desk` component that handles the full desk shell for both user and site desks.
- Kept the behavioral differences behind `siteId?: string`:
  - user desk persists to `desks.user` and keeps the welcome note default
  - site desks persist to `desks.sites[siteId]` and start empty
  - chats are scoped by `siteId` when present
- Moved the shared desk canvas, action context, and tldraw adapter under `apps/ui/src/ui-desks/desk/`.
- Added site desk IPC/query support plus tests for site desk persistence.
- Kept the header plus menu registry-driven so widgets can be added from the same menu on both desk variants.

## Testing
- `npx eslint --fix` on the touched files passed.
- `npm run typecheck` passed.
- `npm test -- apps/ui/src/ui-desks/widgets/create-widget.test.ts apps/ui/src/ui-desks/desk/tldraw-adapter.test.ts apps/studio/src/modules/desks/lib/ipc-handlers.test.ts` passed.


apps/studio/src/modules/desks/lib/ipc-handlers.test.ts
```diff
@@ -0,0 +1,75 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
```
youknowriad: Let's the same that we had for user desk, if we didn't have any tests for that, remove this file.

apps/ui/src/ui-desks/chrome/index.tsx
```diff
@@ -22,10 +22,14 @@ export function DeskHeader( { children }: DeskHeaderProps ) {
 	);
 }
 
-export function DeskChrome() {
+interface DeskChromeProps {
+	activeSiteId?: string;
+}
+
+export function DeskChrome( { activeSiteId }: DeskChromeProps ) {
```
youknowriad: can we name all these activeSiteId props just siteId?

apps/ui/src/ui-desks/chrome/index.tsx
```diff
@@ -22,10 +22,14 @@ export function DeskHeader( { children }: DeskHeaderProps ) {
...
 	return (
 		<DeskHeader>
-			<DeskMenu />
+			<DeskMenu activeSiteId={ activeSiteId } />
 			<DeskChatsTrigger />
 			<DeskCreateMenu />
```
youknowriad: Should these two above components also get the siteId as prop just like DeskMenu to unify how these components work.

apps/ui/src/ui-desks/desk/actions-context.tsx
```diff
@@ -0,0 +1,71 @@
...
+interface DeskActions {
+	canCreateWidgets: boolean;
+	createWidget: ( type: string ) => boolean;
+	registerEditor: ( editor: Editor | null ) => void;
+}
+
+const DeskActionsContext = createContext< DeskActions | null >( null );
+
+export function DeskActionsProvider( { children }: { children: ReactNode } ) {
+	const [ editor, setEditor ] = useState< Editor | null >( null );
```
youknowriad: I'm not sure I understand what is this "editor" local state for? Can't we instead update our desk objects and let the editor react to the changes there? I'm wondering if this editor local state will force us to do two way sync which is fragile.

apps/ui/src/ui-desks/widgets/note/definition.ts
```diff
@@ -30,4 +32,23 @@ export const noteWidgetDefinition = {
 		cornerRadius: 14,
 		stroke: NOTE_TONE_STROKE[ widgetProps.tone ],
 	} ),
+	creation: {
+		getLabel: () => __( 'New sticky note' ),
+		icon: pencil,
```
youknowriad: I think this property should be moved up, it's a specific creation property but more of a generic "widget type" property.

apps/ui/src/ui-desks/widgets/note/definition.ts
```diff
@@ -30,4 +32,23 @@ export const noteWidgetDefinition = {
 		cornerRadius: 14,
 		stroke: NOTE_TONE_STROKE[ widgetProps.tone ],
 	} ),
+	creation: {
+		getLabel: () => __( 'New sticky note' ),
```
youknowriad: I think we should move this to a "labels" property outside creation, we might have a lot more labels later.

apps/ui/src/ui-desks/widgets/note/definition.ts
```diff
@@ -30,4 +32,23 @@ export const noteWidgetDefinition = {
...
+			shapeProps: {
+				w: 200,
+				h: 200,
+			},
+			widgetProps: {
+				text: '',
+				tone: 'yellow',
+			},
+		} ),
+		getInitialSize: ( shapeProps ) => ( {
```
youknowriad: Let's avoid this, getInitialWidget should be enough and let's move it top level outside creation.

apps/ui/src/ui-desks/widgets/note/definition.ts
```diff
@@ -30,4 +32,23 @@ export const noteWidgetDefinition = {
...
+			w: shapeProps.w,
+			h: shapeProps.h,
+		} ),
+		startEditing: true,
```
youknowriad: I'm not sure we need this at the moment. Let's assume it's true for all widgets

PR #3416, opened 2026-05-09
Generate CLI passwords without window

chubes4:
## Related issues

- Follow-up found while refreshing #3095 and running local Studio CLI benchmarks.

## Proposed Changes

- Replace the browser-only `@automattic/generate-password` wrapper usage with an equivalent runtime-neutral password generator.
- Use `globalThis.crypto.getRandomValues()` so password generation works in Node CLI and browser-capable runtimes.
- Preserve the existing exported `generatePassword()` and `createPassword()` APIs.

## Testing Instructions

- `npx vitest run tools/common/lib/tests/passwords.test.ts`
- `npm run lint -- tools/common/lib/passwords.ts`

## AI assistance
- **AI assistance:** Yes
- **Used for:** Diagnosing the CLI `window is not defined` failure and drafting the small runtime-neutral password generator patch. Chris remains responsible for review and merge.

### DIFF
--- a/tools/common/lib/passwords.ts
+++ b/tools/common/lib/passwords.ts
@@ -1,7 +1,47 @@
-import { generatePassword } from '@automattic/generate-password';
 import { __ } from '@wordpress/i18n';
 
-export { generatePassword };
+const ALPHABET = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
+const DIGITS = '0123456789';
+const SPECIAL_CHARS = '!@#$%^&*()';
+const EXTRA_SPECIAL_CHARS = '-_ []{}<>~`+=,.;:/?|';
+
+/**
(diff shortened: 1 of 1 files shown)


chubes4:
Closing this in favor of fixing the root cause upstream in `@automattic/generate-password`.

The Studio failure is real and tracked in #3417, with Homeboy evidence showing `studio site create --no-start` fails on trunk because the package calls `window.crypto.getRandomValues()` in the Node CLI runtime.

This PR worked around the issue by replacing Studio's package usage with local password generation, but that would fork package behavior inside Studio and leave other Node consumers broken. The better path is:

1. Track/fix `@automattic/generate-password` in `Automattic/wp-calypso`.
2. Add package-level Node-runtime coverage there.
3. Release/update the package.
4. Update Studio to consume the fixed package if needed.

Superseded by the upstream-first fix path.

PR #3420, opened 2026-05-11
apps/ui: Add read-only site map view to desk

youknowriad:
## Summary
- Add a site map mode behind the header button that loads an alternate read-only desk config
- Render pages as page widgets, posts as a single post collection, and synthetic Blog nodes where needed
- Add parent/child connectors, site-map-specific title/count UI, and a centered loading state
- Keep widget selection and stack opening available while blocking moves, adds, removes, and persistence in site map mode

## Testing
- Added and updated unit coverage for site map config, connector rendering, and stack z-index behavior
- Ran the desk UI test suite and typecheck successfully
- Verified linting and formatting for modified files

### DIFF
--- a/apps/ui/src/ui-desks/desk/index.tsx
+++ b/apps/ui/src/ui-desks/desk/index.tsx
@@ -1,6 +1,9 @@
 import { __ } from '@wordpress/i18n';
+import { useState } from 'react';
 import { DeskChats } from '../chats';
 import { DeskChrome } from '../chrome';
+import { LoadingPlaceholder } from '../components';
+import { useSiteMapDeskConfig } from '../site-map/use-site-map-desk-config';
 import { DeskWidgetToolbar } from '../widgets/toolbar';
 import { DeskCanvas } from './canvas';
 import { DeskProvider } from './provider';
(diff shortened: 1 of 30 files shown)


apps/ui/src/ui-desks/desk/index.tsx
```diff
@@ -12,28 +15,88 @@ interface DeskProps {
 }
 
 export function Desk( { siteId }: DeskProps ) {
+	if ( siteId ) {
+		return <SiteDesk siteId={ siteId } />;
+	}
+
 	return (
```
youknowriad, line 26: Let's extract the fallback to a "UserDesk component then and I believe there's no need to pass a siteId to any of these elements since it's undefined.

apps/ui/src/ui-desks/site-map/desk-config.ts
```diff
@@ -0,0 +1,514 @@
+import { getIndicesAbove } from 'tldraw';
```
youknowriad, line 1: The logic in this file (not the line above, the whole file) seems like something that is worth unit testing (by faking the REST API calls)

apps/ui/src/ui-desks/widgets/toolbar.tsx
```diff
@@ -33,7 +34,11 @@ export function DeskWidgetToolbar() {
 	}
 
 	const controls =
```
youknowriad, line 35: I think the controls here should be editable even in readonly, revert this change please.

PR #3346, opened 2026-05-05
Build for Linux in Buildkite + EC2 + Docker

twstokes:
AINFRA-2352

## Related issues

- Related to [RSM-1314](https://linear.app/a8c/issue/RSM-1314) — Linux release prep
- Builds on top of #3326 (`make:linux-x64`/`make:linux-arm64` scripts)

## How AI was used in this PR

AI-assisted throughout. Claude inspected the existing `.buildkite/pipeline.yml`, surveyed the `buildkite-ci` Terraform repo to map available queues, and cross-referenced pipelines in other projects to identify the right pattern. A self-review pass added the Linux ECONNRESET workaround, dynamic Node image derivation from `.nvmrc`, and the shell/apt hardening below.

## Proposed Changes

Adds a \`📦 Build for Linux\` group to \`.buildkite/pipeline.yml\`, mirroring the existing Mac and Windows groups: input gate (skipped on \`trunk\` and tagged release builds), group dependency on the input, and a 2-arch matrix (\`x64\`, \`arm64\`) that runs \`npm run make:linux-{{matrix}}\` and uploads the resulting \`.deb\`.

- **Build runs inside a Node Docker container.** Amazon Linux is RPM-based; \`dpkg\` isn't in its default repos and \`fakeroot\` is only available via EPEL. Both are hard prerequisites of \`electron-forge\`'s \`MakerDeb\`. Wrapping the step in a Debian container — using the Buildkite Docker plugin, an established A8C pattern — sidesteps the AMI mismatch cleanly.

- **Image is derived from \`.nvmrc\`.** \`shared-pipeline-vars\` reads \`.nvmrc\` and exports \`NODE_DOCKER_IMAGE=\"node:\${VERSION}-bookworm\"\`, which \`pipeline.yml\` references as \`\$NODE_DOCKER_IMAGE\`. Bumping \`.nvmrc\` moves the Linux container in lockstep with the NVM-installed Node on Mac/Windows agents, so \`\$NVM_PLUGIN\` isn't needed on this step.

- **\`git remote set-url origin https://...\` inside the container.** The host buildkite-agent clones over SSH and \`prepare-dev-build-version.mjs\` calls \`git fetch --tags --force\`. The container has no \`~/.ssh/known_hosts\` or deploy key, so SSH hangs on the host-key prompt. Studio is a public repo, so swapping to anonymous HTTPS for the in-container fetch is the simplest fix and avoids exposing the agent's deploy key inside the container.

[...]

PR #3334, opened 2026-05-04
Run unit and E2E tests on Linux in Buildkite

ivan-ottinger:
## Related issues

Part of [RSM-1314](https://linear.app/a8c/issue/RSM-1314) — Linux release prep.
Apps Infra request: [AINFRA-2352](https://linear.app/a8c/issue/AINFRA-2352/provision-linux-buildkite-agent-for-studio-dev-builds).

## How AI was used in this PR

AI-assisted: identified the three places that needed updating, drafted the `xvfb-run` wrapper, and drafted this PR body. Author reviewed.

## Proposed Changes

Brings Linux to parity with Mac and Windows in CI:

- `.buildkite/pipeline.yml`:
  - Adds `- linux` to the **Unit Tests** matrix. `run-unit-tests.sh` is platform-agnostic (just runs `npm run test`).
  - Adds `- with: { platform: linux, arch: arm64 }` to the **E2E Tests** matrix.
- `.buildkite/commands/run-e2e-tests.sh`:
  - Adds a `linux)` case to the platform mapping.
  - Wraps `npx playwright test` with `xvfb-run -a` on Linux only — Electron needs a display server, and CI Linux agents are headless. Mac and Windows behavior is unchanged.

E2E runs on `linux-arm64` only (not both arches) to match the established pattern of one native E2E arch per platform (`mac-arm64`, `windows-x64`).

## Testing Instructions

This PR cannot be exercised end-to-end until a Linux Buildkite agent is provisioned and registered to the `linux` queue with `xvfb` available on the agent image (see AINFRA-2352). Once that lands:

1. Push a commit to the branch.
2. Verify the new `Unit Tests on linux` job succeeds.
3. Verify the new `E2E Tests on linux-arm64` job succeeds.

The Linux-side harness changes (so E2E tests actually run on Linux) are already on trunk via #3295.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #3421, opened 2026-05-11
Run E2E tests on Linux in Buildkite

gavande1:
## Related issues

- Implements [RSM-2593](https://linear.app/a8c/issue/RSM-2593/run-e2e-tests-on-linux-in-buildkite)
- Companion to #3370 (Linux unit tests, already merged) — extends the same Docker-on-`default`-queue pattern from #3346 to E2E.
- Earlier draft #3334 bundled unit + E2E; the `linux)` case in `run-e2e-tests.sh` and the `xvfb-run` wrapper are picked up from there.
- Parent: [RSM-1314](https://linear.app/a8c/issue/RSM-1314) — Investigate what it would take to release Studio for Ubuntu officially.

## How AI was used in this PR

Claude drafted the script + pipeline edits from the Linear issue's spec and the existing Linux build / unit-test patterns. Local Docker repro on macOS was attempted but hit a Docker-Desktop bind-mount permission quirk during electron-packager's `extraResource` copy — file mode bits get mangled going through osxfs/virtiofs in a way that won't manifest on the real Linux CI agent. Per the issue's own forecast of "1–2 iteration cycles after first push," we're letting Buildkite exercise the actual unknowns rather than chasing a Docker-Desktop-for-Mac artifact.

## Proposed Changes

- `.buildkite/pipeline.yml`: New `E2E Tests on linux-arm64` step running on `queue: default` inside the Node Debian container, mirroring the `Unit Tests on linux` step from #3370. Kept as a standalone step (not a third matrix entry on `*e2e_config`) because Mac and Windows share `queue` + plugin defaults Linux doesn't. Same `if:` clause as the existing E2E step (skip drafts, tags, native-php-labeled PRs) and same `notify: github_commit_status: context: E2E Tests` so the new step rolls up to a single GitHub status check alongside Mac/Windows.
- `.buildkite/commands/run-e2e-tests.sh`:
  - Added `linux)` case to the platform-to-`FORGE_PLATFORM` map.
  - For `$PLATFORM = linux`: inline `apt-get install` for Electron's runtime libraries + `xvfb` + Playwright's browser deps; inline `npm ci --maxsockets 1` (the a8c-ci-toolkit cache helpers only exist on the host, same reason as the unit-test script).
  - Wrap `npx playwright test` with `xvfb-run -a` on Linux to provide a virtual display (CI agents are headless).

## Testing Instructions

The CI run itself is the test surface — watch the `E2E Tests on linux-arm64` step in this PR's Buildkite build. Expected outcome: the step turns green alongside the existing Mac/Windows E2E steps, all reporting under one `E2E Tests` GitHub check.

Likely follow-up iterations if the first run trips:

- **Electron sandbox error** → add `--cap-add SYS_ADMIN` to the Docker plugin config (preferred over `--no-sandbox` because it preserves real namespace permissions).
- **Missing library at Electron launch** → add the package name to the inline `apt-get install` set.
- **Step timeout** → raise `command_timeout_in_minutes` for this step specifically; E2E + container overhead may exceed the default.

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors? — n/a (CI-config change only)
- [ ] First Buildkite run is green (or has been iterated to green).


.buildkite/pipeline.yml
```diff
@@ -87,6 +85,31 @@ steps:
...
+    artifact_paths:
+      - test-results/**/*.zip
+      - test-results/**/*.png
+      - test-results/**/*error-context.md
+    env:
+      DEBUG: "pw:browser"
+    # TEMP(rsm-2593): if: removed to force E2E on every push while iterating on Linux E2E setup. Revert before merge.
+    notify:
+      - github_commit_status:
+          context: E2E Tests
```
iangmaia, line 111: Just so the status check on GitHub gets a different name and the statuses aren't mixed:
```suggestion
          context: E2E Tests on linux-x64
```

iangmaia: I've updated it on the fix commit 215b5b7cb8d442dbdee0cff4936c6ac98b1edf12.

iangmaia:
@gavande1 based on my tests in https://github.com/Automattic/studio/pull/3466, I've committed https://github.com/Automattic/studio/pull/3421/commits/215b5b7cb8d442dbdee0cff4936c6ac98b1edf12 and https://github.com/Automattic/studio/pull/3421/commits/4f2fa76aec6e3b05be7341281d6e75718913d6db which seemed to have fixed the Linux E2E tests 🎉

gavande1:
Thanks @iangmaia for digging into this and getting Linux E2E green on CI. The fixes around the WP-CLI Blueprint mount conflict and the extra Electron launch stability flags were exactly the missing pieces between the Parallels VM setup and Buildkite’s Docker/xvfb environment.

Really appreciate you taking this over the finish line and leaving the changes clear enough to understand why CI needed different handling.

ivan-ottinger (approved):
Amazing! Thanks for all the work on this PR. Approving. ✅ 

(the failed lint test is unrelated)

gavande1:
@iangmaia (cc @ivan-ottinger) — quick question on the CI checks here: right now this PR is reporting both **`E2E Tests`** and **`E2E Tests on linux-x64`** as two separate GitHub status checks. The original intention was to roll the Linux step up under the existing `E2E Tests` check (same as Mac/Windows) so we'd end up with a single combined status.

Does it make sense to club them into one `E2E Tests` check, or would you prefer to keep `linux-x64` broken out separately? Just want to confirm the direction before adjusting.

ivan-ottinger:
> Does it make sense to club them into one `E2E Tests` check, or would you prefer to keep `linux-x64` broken out separately? Just want to confirm the direction before adjusting.

To me it would look more concise to have them all together. At the same time, it is not 100% clear to me what other consequences that approach brings.

gavande1:
Thanks @ivan-ottinger — same read on my end, a single combined status feels cleaner. My original intent was to have them clubbed under one `E2E Tests` check.

@iangmaia likely split `linux-x64` out separately on purpose while iterating (easier to see Linux pass/fail in isolation while debugging). Let's wait for his take on whether it's safe to roll them back together now that the Linux step is green, or if there's a reason to keep them broken out.

gavande1:
Going ahead with clubbing the Linux E2E status under the shared `E2E Tests` GitHub check (matches Mac/Windows). If we find we want per-platform visibility on the status row later — e.g. to keep a failing Linux run from masking a passing Mac/Windows run — we can split them back out.

PR #3423, opened 2026-05-11
apps/ui: Add media widget and file-drop media uploads to desks

youknowriad:
## Summary
- Add a new `media` desk widget for images and videos, including toolbar actions and canvas rendering
- Support dropping image/video files onto a desk to create placeholder widgets and upload them into the site media library
- Extend the WordPress REST transport to carry binary request bodies so media uploads work through the existing site API path
- Update widget registry, desk/provider plumbing, and adapter coverage for media widgets

## Testing
- `npx eslint --fix` on the touched TypeScript/TSX files passed; CSS module files are ignored by the current ESLint config
- `npm run typecheck` passed
- Focused widget and tldraw adapter tests passed


apps/ui/src/data/wordpress/media.ts
```diff
@@ -0,0 +1,57 @@
+import apiFetch from '@wordpress/api-fetch';
+
+export interface UploadedSiteMedia {
```
youknowriad: Please check whether the types defined here already exist in wordpress/core-data

apps/ui/src/ui-desks/chats/context.tsx
```diff
@@ -0,0 +1,62 @@
+import { createContext, useContext } from 'react';
```
youknowriad, line 1: Can you remove all the chat related changes from this PR, focus it only on the media widget type and uploads.

PR #3426, opened 2026-05-11
Make phpMyAdmin work with native PHP runtime

fredrikekelund:
## Related issues

- Fixes RSM-2697

## How AI was used in this PR

I used Codex to help me understand how Playground makes phpMyAdmin work. Once I had a 90% understanding and a relatively clear idea of what I thought the Studio implementation should look like, I asked Codex for the implementation, which I then tweaked to simplify.

## Proposed Changes

- Make phpMyAdmin work with sites using the `native-php` runtime.
- This is achieved by copying and slightly modifying `DbiMysqli.php` and `config.inc.php` from the Playground repo.
   - The modifications only change where we load files from. With Playground, there's the virtual file system that gives us stable, per-site paths (like `/internal/shared/wp-env.php`). With the native PHP runtime, we obviously don't have this, so we generate temp files and pass the site-specific paths to the PHP process using environment variables.

## Testing Instructions

1. `STUDIO_RUNTIME=native-php npm start`
2. Have a site that uses the `native-php` runtime (if you're unsure, just create a new site – it will automatically be assigned this runtime when Studio is run with `STUDIO_RUNTIME=native-php`)
3. Start that site
4. Click the phpMyAdmin button in the Overview tab in Studio

## Pre-merge Checklist


apps/cli/php/DbiMysqli.php
fredrikekelund, line 1: This file is a straight up copy of the [Playground source code](https://github.com/WordPress/wordpress-playground/blob/dac72f5b40190af049d6a41e4dcc842e8990f64d/packages/playground/tools/src/phpmyadmin/DbiMysqli.php)

bcotrim (approved):
Works as described
LGTM 👍

PR #3435, opened 2026-05-11
Blueprints Gallery Library: Move blueprints to three in a row

senff:
## Related issues

- Fixes [STU-1666
](https://linear.app/a8c/issue/STU-1666/blueprints-gallery-library-move-blueprints-to-three-in-a-row)

## How AI was used in this PR

No AI used.

## Proposed Changes

- Shows 3 blueprints per row instead of 2

image

## Testing Instructions

- Open Studio
- Enable Blueprints Gallery feature flag
- Select ADD SITE → BUILD NEW SITE
- Confirm that there are 3 blueprints per row

## Pre-merge Checklist


katinthehatsite:
I reviewed these changes and I think the alignment in that case does not work well - we would either need to feature more blueprints or we would need to cut them on the top to three. I think we could leave this as is for now e.g. featuring 2 blueprints so that it looks slightly more aligned and easier to read excerpts. We can revisit if we get further feedback from users.

PR #3439, opened 2026-05-12
Build PHP binaries with Xdebug

fredrikekelund:
## Related issues

- Related to RSM-2698

## How AI was used in this PR

Codex and Claude were both used heavily to iterate on the build script, even if we ultimately arrived at a relatively simple change compared to `trunk`.

## Proposed Changes

We use [crazywhalecc/static-php-cli](https://github.com/crazywhalecc/static-php-cli) to build PHP binaries. The big selling point of that project is that it can bundle PHP extensions into the binary itself (normally they're distributed as separate `.so`/`.dll` files, so called "shared" extensions). It turns out that Xdebug cannot be compiled into the static PHP binary ([docs](https://static-php.dev/en/guide/extension-notes.html#xdebug)). AFAICT, this is an inherent limitation in Xdebug.

However, static-php-cli can actually handle shared extensions, too. Adding support for Xdebug specifically seems to have been a key motivation behind implementing this. The only blocker is that this only works on Posix systems, not on Windows. 

So, to get Xdebug support on both platforms, this PR does the following:

1. Continue using static-php-cli to build the macOS PHP binary. Xdebug is built as a shared extension that ends up in `ext/xdebug.so`
2. On Windows, use the official php.net build and repackage it so that extensions are stored in the `ext` directory.
3. Add a `runtime.json` file for both platforms that serves as a reference for the build version, where the binary lives, and where the extension directory is. I'm not sure how useful this will be in practice, but it seems like a decent idea.

## Testing Instructions

I'll run the GitHub Actions workflow for all [supported PHP versions](https://www.php.net/supported-versions.php) and manually confirm the output.

## Pre-merge Checklist


.github/workflows/build-php-cli-binaries.yml
```diff
@@ -49,42 +54,54 @@ jobs:
             runner: macos-15
             artifact_platform: macos
             artifact_arch: aarch64
-            binary_name: php
-            extensions: &macos_php_extensions apcu,bcmath,calendar,ctype,curl,dba,dom,exif,fileinfo,filter,ftp,gd,gettext,iconv,igbinary,imagick,intl,mbregex,mbstring,mysqli,mysqlnd,opcache,openssl,pcntl,pdo,pdo_mysql,pdo_sqlite,phar,posix,readline,redis,session,shmop,simplexml,sockets,sodium,sqlite3,ssh2,tokenizer,xml,xmlreader,xmlwriter,xsl,yaml,zip,zlib
+            extensions: &macos_extensions apcu,bcmath,calendar,ctype,curl,dba,dom,exif,fileinfo,filter,ftp,gd,gettext,iconv,igbinary,intl,mbregex,mbstring,mysqli,mysqlnd,opcache,openssl,pcntl,pdo,pdo_mysql,pdo_sqlite,phar,posix,readline,redis,session,shmop,simplexml,sockets,sodium,sqlite3,ssh2,tokenizer,xdebug,xml,xmlreader,xmlwriter,xsl,yaml,zip,zlib
```
fredrikekelund, line 57: I've removed imagick. I'd love to keep it in, but it's not included in the official php.net Windows build, so I skipped it for now.

PR #3456, opened 2026-05-12
Make site runtime a global feature flag

bcotrim:
## Related issues

- Fixes RSM-1689

## How AI was used in this PR

AI helped trace the existing runtime storage and startup paths, implement the env-driven behavior, and update focused tests. I reviewed the changes and adjusted them based on review feedback so runtime is no longer treated as site data.

## Proposed Changes

- Add a `nativePhpRuntime` beta feature to the existing app menu beta-features flow.
- Apply that beta feature to `STUDIO_RUNTIME` before Studio fetches sites or spawns CLI commands.
- Make CLI site start, blueprint, WP-CLI, and browser-opening behavior resolve runtime from the global env var instead of per-site `runtime`.
- Stop persisting or emitting `runtime` as site metadata.

## Testing Instructions

- Open the app menu and enable the Native PHP runtime beta feature.
- Start an existing site that does not have a per-site `runtime` prop and confirm it uses the native PHP path.
- Disable the beta feature, start a site again, and confirm it uses the Playground path.
- Run CLI commands with `STUDIO_RUNTIME=native-php` and confirm existing sites use the native PHP runtime without requiring stored site metadata.
- Run CLI commands without `STUDIO_RUNTIME` and confirm existing sites default to Playground.
- Confirm site creation and site listing no longer persist or emit `runtime`.

## Pre-merge Checklist


apps/studio/src/lib/beta-features.ts
```diff
@@ -42,10 +61,9 @@ export async function updateBetaFeature(
 		await lockAppdata();
 		const userData = await loadUserData();
 		const betaFeatures = await getBetaFeatures();
-		// @ts-expect-error If `BetaFeatures` is empty, `key` will be `never`, and we cannot use it to
-		// assign to`betaFeatures`.That's fine. Just rely on type checking when this function is called.
```
fredrikekelund, line 46: Can we keep this comment, but just remove the `@ts-expect-error` part? This will become troublesome again once we remove this beta feature.

apps/cli/lib/wordpress-server-manager.ts
```diff
@@ -32,6 +38,10 @@ import type { WordPressInstallMode } from '@wp-playground/wordpress';
 
 export const SITE_PROCESS_PREFIX = 'studio-site-';
 
+export type SiteProcessDescription = ProcessDescription & {
+	runtime: SiteRuntime;
+};
```
fredrikekelund: I don't believe we need this separate type. `ProcessDescription` is inferred from `processDescriptionSchemaBase`, which already has a `runtime` prop. Yes, that prop can be undefined, which this type doesn't allow, but that doesn't appear to be an issue.

fredrikekelund (approved):
I think we can do away with `SiteProcessDescription` for a smaller diff, but functionally, this works great 👍 The logic is clear

apps/cli/process-manager-daemon.ts
```diff
@@ -34,6 +35,7 @@ type ManagedProcessBase = {
 	scriptPath: string;
 	args: string[];
 	env: NodeJS.ProcessEnv;
+	runtime: SiteRuntime;
```
fredrikekelund, line 39: We could leave a comment to highlight that this is only needed to know where WP-CLI commands should be executed.

fredrikekelund (approved):
Splendid! Clean and understandable 👍

PR #3472, opened 2026-05-13
Linux: auto-trust Firefox profiles and surface in-app cert notice

gavande1:
## Related issues

- Fixes [RSM-2943](https://linear.app/a8c/issue/RSM-2943) — Linux: auto-trust the Studio CA in Firefox profiles
- Fixes [RSM-2944](https://linear.app/a8c/issue/RSM-2944) — Linux: auto-install root CA into Firefox profiles

## How AI was used in this PR

Implementation drafted with Claude Code. Tests, types, and the flow described in the testing instructions below were verified manually inside the Linux VM. @ivan-ottinger also drafted #3488 to demonstrate a simpler alternative that drops the in-app Firefox-specific notice and relies on `refetchOnFocus` for the trust-after-first-launch case; that commit has been folded into this PR.

## Proposed Changes

- Extend `trustRootCA()` on Linux to also import the Studio CA into every existing Firefox profile NSS DB (`cert9.db`) across apt (`~/.mozilla/firefox/*.default*`), Snap (`~/snap/firefox/common/.mozilla/firefox/*.default*`), and Flatpak (`~/.var/app/org.mozilla.firefox/.mozilla/firefox/*.default*`) install roots.
- Extend `isRootCATrusted()` on Linux to also verify each existing Firefox profile contains the cert, so the existing **Trust Certificate** notice reappears if a profile drifts. Profiles that don't exist yet are vacuously trusted (Firefox creates `cert9.db` on first launch — nothing to check or import into until then).
- New `tools/common/lib/linux-trust-store.ts` helpers: `getLinuxFirefoxProfileDbDirs`, `areAllFirefoxProfilesTrustedLinux`, `importCAIntoFirefoxProfilesLinux`. They mirror the existing `getLinuxNssDbCandidates` / `isCAImportedInUserNssDbsLinux` / `importCAIntoUserNssDbsLinux` pattern.
- The "Firefox installed but never launched" gap is handled by the existing `refetchOnFocus: true` on `certificateTrustApi` + `setupListeners(store.dispatch)` in `stores/index.ts`. When the user launches Firefox for the first time and returns focus to Studio, the trust check re-runs, finds an untrusted new profile DB, and the existing notice + **Trust Certificate** button reappear. One more click silently imports into the new profile (no sudo prompt — the system bundle install short-circuits when the cert is already in `/etc/ssl/certs`).

## Testing Instructions

Run inside the Linux VM (`ssh parallels@10.211.55.3`, project at `/home/parallels/Desktop/studio-linux`).

**Auto-import into existing Firefox profile**

1. Launch Firefox at least once so a `.default*` profile with `cert9.db` exists.
2. Make sure the Studio CA is currently untrusted (see _Reset_ below).
3. Start Studio, open a site, enable HTTPS. The HTTPS row shows _Disabled / Enabled_ with a **Trust Certificate** button and the untrusted-cert notice.
4. Click **Trust Certificate**, complete the polkit prompt. Button and notice disappear.
5. Verify the cert landed in Firefox: `certutil -d sql:$HOME/.mozilla/firefox/<id>.default-release -L -n 'WordPress Studio CA'` should succeed. Repeat for the snap / flatpak profile roots if you have those installations.
6. Open the site in Firefox — no security warning.

**Firefox installed but never launched (trust-then-launch flow)**

1. Reset trust (see _Reset_ below) and remove any Firefox profile dirs: `rm -rf ~/.mozilla/firefox ~/snap/firefox/common/.mozilla/firefox ~/.var/app/org.mozilla.firefox/.mozilla/firefox`.
2. Start Studio, open a site, enable HTTPS. Click **Trust Certificate**.
3. System bundle + Chromium NSS import happen; Firefox import is a no-op (no profile yet). Button and notice hide.
4. Launch Firefox for the first time, then return focus to Studio.
5. The untrusted-cert notice + **Trust Certificate** button reappear (refetch-on-focus picks up the freshly-created profile DB).
6. Click **Trust Certificate** again — should run without a polkit prompt (system bundle is already trusted; only the NSS imports execute).
7. Reload the site in Firefox — no security warning.

**Trust-state regression**

1. With cert trusted, remove just the Firefox profile copy:
   `certutil -d sql:$HOME/.mozilla/firefox/<id>.default-release -D -n 'WordPress Studio CA'`
2. Return focus to Studio → notice + Trust button should reappear on refetch.

https://github.com/user-attachments/assets/0d4b53f0-e96e-41aa-9518-53741c75dd0f

> Note: the recording above was made before #3488's changes were folded in; it shows the older flow that included a second Firefox-specific notice. The current behavior matches the steps above (one notice, not two).

**No regressions on macOS / Windows**

- On both, the Trust Certificate flow behaves exactly as before; nothing Linux-specific is triggered.

**Reset trusted certificate (for re-testing)**

```bash
sudo rm -f /usr/local/share/ca-certificates/studio-ca.crt && sudo update-ca-certificates --fresh
certutil -d sql:$HOME/.pki/nssdb -D -n 'WordPress Studio CA' || true
certutil -d sql:$HOME/snap/chromium/current/.pki/nssdb -D -n 'WordPress Studio CA' 2>/dev/null || true
for p in $HOME/.mozilla/firefox/*.default* $HOME/snap/firefox/common/.mozilla/firefox/*.default* $HOME/.var/app/org.mozilla.firefox/.mozilla/firefox/*.default*; do
  [ -d "$p" ] && certutil -d sql:"$p" -D -n 'WordPress Studio CA' 2>/dev/null || true
done
```

**Tests**

```bash
npm test -- tools/common/lib/tests/linux-trust-store.test.ts apps/studio/src/components/tests/content-tab-settings.test.tsx
```

## Pre-merge Checklist


apps/studio/src/lib/tests/detect-linux-browsers.test.ts
```diff
@@ -0,0 +1,58 @@
...
+import { vi } from 'vitest';
+import { isFirefoxInstalledOnLinux } from '../detect-linux-browsers';
+import { findOnPath } from '../find-on-path';
+
+vi.mock( 'node:fs', () => ( {
+	default: { existsSync: vi.fn() },
+	existsSync: vi.fn(),
+} ) );
+
+vi.mock( '../find-on-path', () => ( {
```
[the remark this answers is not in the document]
gavande1: Good catch — fixed in a197a6fb. The mock was being applied (the "returns false when no PATH/dirs" test would have failed otherwise on a VM with Firefox installed, since the real findOnPath would resolve it), but the mismatched specifier made the intent ambiguous. Now both the implementation and the test import `findOnPath` via `src/lib/find-on-path` so `vi.mock` is unambiguous.

apps/studio/src/components/content-tab-settings.tsx
```diff
@@ -151,6 +156,18 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
...
+							linuxBrowserCertStatus?.firefoxDetected && (
+								<div className="mt-1 max-w-96" data-testid="trust-cert-firefox-notice">
+									<span className="text-frame-text-secondary mt-1">
+										{ __(
+											'Firefox uses its own certificate store. If Firefox doesn’t recognize the certificate, import it manually to avoid the security warning.'
+										) }
+									</span>{ ' ' }
+									<LearnHowLink docsLinksKey="docsSslLinuxFirefox" />
+								</div>
+							) }
```
ivan-ottinger: What if we kept only the original notice we already have on trunk? Instead of adding this new one we could:
- rely on the new docs section, e.g. "If you use Firefox on Linux" to communicate potential issue to the user
- ensure the existing notice is displayed even in that very edge case where the user has Firefox installed, but haven't launched it yet.

I think that could streamline the user experience and also make the code simpler.

apps/studio/src/lib/get-localized-link.ts
```diff
@@ -37,6 +37,9 @@ const DOCS_LINKS = {
 	docsSslInStudio: {
 		en: 'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/',
 	},
+	docsSslLinuxFirefox: {
+		en: 'https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/#linux-firefox',
```
ivan-ottinger: In relation to https://github.com/Automattic/studio/pull/3472/changes#r3240894626 we could do `https://developer.wordpress.com/docs/developer-tools/studio/ssl-in-studio/#linux`.

That would lead to a new `Linux` heading under the `Certificate Trust by platform` section.

ivan-ottinger (commented):
Thank you for your work on this PR, Rahul!

I am yet to review and test the PR fully and would like to share initial thoughts I have had so far.

apps/studio/src/lib/detect-linux-browsers.ts
```diff
@@ -0,0 +1,18 @@
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { findOnPath } from 'src/lib/find-on-path';
+
+export function isFirefoxInstalledOnLinux( homeDir: string = os.homedir() ): boolean {
```
ivan-ottinger: I might be missing something, but from what I see, if we decide to drop that new UI notice, we could get rid of `isFirefoxInstalledOnLinux` and a few other code parts.

apps/studio/src/components/content-tab-settings.tsx
gavande1: Thanks Ivan, I want to push back a bit here. My biggest concern with falling back to just the existing notice is what it would take to also cover the "Firefox installed but never launched" case you mentioned. Are you comfortable with the notice being persistently visible in the settings tab? That's the part I'm most hesitant about.

The existing notice is gated on `!isCertificateTrusted`. On Linux that already factors in Firefox profile trust via `areAllFirefoxProfilesTrustedLinux`, but it's vacuously true when no profile exists — so to surface the notice in the installed-but-never-launched case, we'd have to gate it on something like `isFirefoxInstalled`, or make `isRootCATrusted` return false in that scenario. Either way, that signal doesn't auto-clear: a user who has Firefox installed (common on Ubuntu, where it ships by default) but never launches it would see the notice sitting in settings indefinitely.

To be upfront about the tradeoff: my current PR doesn't really solve that edge case either — the new Firefox notice is also gated on `!isCertificateTrusted`, so it stays silent until a profile actually exists. The difference is that when a profile *does* show up, the user gets Firefox-specific guidance instead of the generic message, and `isFirefoxInstalledOnLinux` is what lets us know to do that. I'd rather lean on docs for the never-launched case than make the notice permanent for everyone with Firefox preinstalled.

What I was going for overall is the closest analog to the Chrome trust-status check we already have: only surface platform-specific guidance when it's actionable. Happy to trim the supporting code if parts feel heavy — but I'd like to keep the conditional surfacing rather than collapsing it into a docs-only path. WDYT?

ivan-ottinger: Hey Rahul! 👋🏼 I created this draft PR to illustrate what I meant: https://github.com/Automattic/studio/pull/3488. Basically, there would be just one notice that would be there only when it makes sense. So if the user did not start Firefox, the notice would not be rendered.

gavande1: Thanks Ivan — your draft made this much easier to evaluate, and it landed exactly where you said it would: tested the trust-then-launch flow end-to-end on the Linux VM and the existing notice reappears on focus return as you described, then the second Trust click silently imports into the new profile.

Folded your commit into this PR as `1194d83f`, so the new Firefox notice + its detection/IPC/RTK Query plumbing are all gone. Updating the PR description next to reflect the new shape. Happy to leave #3488 to you to close once you've had a chance to look.

apps/studio/src/lib/get-localized-link.ts
gavande1: Resolved by `1194d83f` — `docsSslLinuxFirefox` was removed entirely. The existing notice keeps using `docsSslInStudio`, so the Linux subsection that lands on that page will be reachable from there.

apps/studio/src/lib/detect-linux-browsers.ts
gavande1: Done in `1194d83f` — `isFirefoxInstalledOnLinux`, `detect-linux-browsers.ts` and its test, the IPC handler, the preload binding, and the RTK Query endpoint are all removed.

apps/studio/src/components/content-tab-settings.tsx
ivan-ottinger: Thank you for the updates, Rahul! I will now re-review and test this PR.

ivan-ottinger (approved):
Nice! Looks good and works correctly. 👍🏼 I have tested all scenarios I could think of and tested for regressions. Haven't observed any.

https://github.com/user-attachments/assets/e19231ca-ef4b-4dcf-9e8c-c25cefe4e057

PR #3491, opened 2026-05-14
Fix native php blueprint

bcotrim:
## Related issues

- None.

## How AI was used in this PR

Codex was used interactively to debug the native PHP blueprint startup failure, inspect daemon/PHP behavior on Windows, and draft the focused fix. The behavior was manually reproduced before the fix and validated afterward in both CLI and Studio flows.

## Proposed Changes

- Add a temporary native PHP subprocess `php.ini` for blueprint PHAR execution.
- Load Studio's bundled Windows PHP extensions in that subprocess config, including `pdo_sqlite` and `sqlite3`, so the PHAR can detect the SQLite-backed WordPress install correctly.
- Generate a temporary CA bundle from Node's trusted root certificates and pass it to PHP via `openssl.cafile` and `curl.cainfo`, fixing HTTPS downloads from blueprint steps under native PHP.
- Allow native PHP subprocess callers to pass scoped environment variables and extra PHP args.
- Clean up the temporary blueprint `php.ini` directory after the blueprint run.
- Add focused tests for the native PHP helper behavior.

## Testing Instructions

- Confirm automated checks pass:
  - `npx eslint --fix apps/cli/lib/native-php.ts apps/cli/php-server-child.ts apps/cli/lib/tests/native-php.test.ts`
  - `npm run typecheck`
  - `npm test -- apps/cli/lib/tests/native-php.test.ts apps/cli/lib/tests/wordpress-server-manager.test.ts`
- Build the CLI before manual testing:
  - `npm run cli:build`
- On Windows, create an empty blueprint file containing `{}` and verify native PHP site creation succeeds:
  - `$env:STUDIO_RUNTIME = "native-php"`
  - `node apps/cli/dist/cli/main.mjs --path "$env:USERPROFILE\Studio\native-empty-blueprint-test" site create --name "Empty Blueprint" --wp latest --php 8.4 --blueprint "$env:TEMP\empty-blueprint.json" --skip-browser`
- In the Studio desktop app, create a native PHP site from the Quick Start blueprint and verify the site starts successfully.
- Optional: enable Xdebug for the created native PHP site and confirm the running PHP process loads `php_xdebug.dll`.

## Pre-merge Checklist


wp-files/blueprints/blueprints.phar
fredrikekelund, line 1: Did you patch this file somehow, @bcotrim? php-toolkit has had a new release, so we shouldn't have to commit this file anymore since https://github.com/Automattic/studio/pull/3437

apps/cli/lib/tests/native-php.test.ts
fredrikekelund, line 1: I guess it doesn't hurt, but I'm a little bit allergic to unit tests that LLMs generate out of habit. I think E2E tests serve us much better for truly testing this functionality (we already have E2E tests that do that)

apps/cli/lib/native-php.ts
```diff
@@ -103,10 +104,72 @@ const WINDOWS_PHP_EXTENSIONS = [
...
+	const phpIniDirectory = await fs.promises.mkdtemp( path.join( tempRoot, 'studio-native-php-' ) );
+	const caBundlePath = getNativePhpCaBundlePath( phpIniDirectory );
+	await fs.promises.writeFile( caBundlePath, rootCertificates.join( os.EOL ), 'utf8' );
+	await fs.promises.writeFile(
+		path.join( phpIniDirectory, 'php.ini' ),
+		getNativePhpSubprocessIniContents( phpVersion, caBundlePath ),
+		'utf8'
+	);
+	return phpIniDirectory;
+}
```
fredrikekelund: The `php.ini` contents only change with PHP versions. To keep things simple, I think we should explore writing the `php.ini` file to the PHP executable's directory and using that as the source of truth for the PHP config for all Windows PHP processes. Did you try this already, @bcotrim?

fredrikekelund (approved):
This fixes the problem, which is great! @bcotrim, you mentioned seeing some trouble with images on this branch, but I can upload and view images on the site fine. Let me know what problems you were experiencing. 

Really, the only thing I think we should explore here (although it could be a follow-up PR) is using `php.ini` for all PHP processes on Windows, not just when running blueprints. This would help with simplicity in the implementation and predictability at runtime.

wp-files/blueprints/blueprints.phar
fredrikekelund: I now know you did patch it with the changes from https://github.com/WordPress/php-toolkit/pull/280. I've since also patched it with the changes from https://github.com/WordPress/php-toolkit/pull/281

apps/cli/lib/native-php.ts
fredrikekelund: I implemented this in 3f3977c

apps/cli/lib/tests/native-php.test.ts
fredrikekelund: I went ahead and removed the test file in e834425

fredrikekelund:
I implemented my suggestion to always load the `php.ini` file (that lives next to the binary) and verified that creating a site with a blueprint still works. In fact, I even took it one step further and made it so we create that file on macOS, too. Streamlining the implementation between platforms will make things more predictable.

fredrikekelund:
The E2E tests aren't passing on Windows, but that's expected. We need to build and upload the new PHP packages to the CDN before it'll work

PR #3492, opened 2026-05-14
CLI: Lock language pack setup

### LINKED ISSUE #3489: studio wp can print language-pack ENOENT warnings during concurrent setup
## Summary

Concurrent `studio wp` invocations can race while setting up bundled language packs in `~/.studio/server-files/language-packs`. Some invocations print `Failed to set up dependency language packs` with `ENOENT` from `lstat`, `scandir`, or `utime`, but the WP-CLI command still completes successfully.

## Reproduction

Use a disposable Studio config directory so the repro does not touch a real `~/.studio` cache:

```bash
tmp=\"/tmp/studio-language-pack-repro\"
/bin/rm -rf \"$tmp\"
mkdir -p \"$tmp\"

DEV_CONFIG_DIR=\"$tmp\" studio wp --studio-no-path cli version >/dev/null 2>&1

mkdir -p \"$tmp/server-files/language-packs\"
touch \"$tmp/server-files/language-packs/stale-marker\"

for i in {1..24}; do
  DEV_CONFIG_DIR=\"$tmp\" studio wp --studio-no-path cli version >\"$tmp/out-$i.log\" 2>&1 &
done
wait

for file in \"$tmp\"/out-*.log; do
  /usr/bin/grep -E \"ENOENT|language packs|WP-CLI\" \"$file\"
done
```

## Observed

Multiple commands may print setup warnings while still completing the requested WP-CLI command:

```text
Failed to set up dependency language packs: Error: ENOENT: no such file or directory, lstat '.../server-files/language-packs/es_ES-81d6f084cb273e02e15b01bd9ece87f7.json'
  code: 'ENOENT',
WP-CLI 2.12.0
```

A forced deletion during copy can also reproduce the `utime` variant:

```text
Failed to set up dependency language packs: Error: ENOENT: no such file or directory, utime '.../server-files/language-packs/admin-id_ID.l10n.php'
  code: 'ENOENT',
  syscall: 'utime',
WP-CLI 2.12.0
```

## Expected

Concurrent CLI invocations should not print transient setup warnings for shared language-pack cache races.

## Likely cause

Every CLI invocation runs `setupServerFiles()`. `copyBundledLanguagePacks()` compares the bundled language pack directory with `server-files/language-packs`, then removes and recreates the shared target when it differs:

```ts
await fs.promises.rm( targetLanguagePacksPath, { recursive: true, force: true } );
await recursiveCopyDirectory( sourceLanguagePacksPath, targetLanguagePacksPath );
```

`recursiveCopyDirectory()` uses `fs.cp(..., { preserveTimestamps: true })`, so one process can remove files while another process is copying/statting/preserving timestamps in the same tree.

## Potential fix

Serialize language-pack setup with a lock, or copy into a temporary directory and atomically rename into place so concurrent CLI invocations never delete/read the same target tree mid-copy.

## AI assistance
- **AI assistance:** Yes
- **Used for:** Reproducing the CLI race and drafting this issue report.

chubes4:
## Related issues

- Fixes #3489

## Proposed Changes

Concurrent CLI invocations can all run `setupServerFiles()` and refresh `~/.studio/server-files/language-packs` at the same time. When one process removes the shared target tree while another is copying or preserving timestamps, the losing process prints transient `ENOENT` setup warnings even though the requested WP-CLI command completes.

This PR serializes the language-pack refresh with Studio's existing lockfile helper, and reruns the size/mtime comparison inside the lock so only the first process refreshes the cache while later processes skip once the cache is current.

## Evidence

Reproduction used a disposable `DEV_CONFIG_DIR` with a stale language-pack cache, then ran 24 concurrent `studio wp --studio-no-path cli version` requests.

Before this fix, the repro produced multiple setup warnings while still completing WP-CLI:

```text
Failed to set up dependency language packs: Error: ENOENT: no such file or directory, lstat '.../server-files/language-packs/es_ES-81d6f084cb273e02e15b01bd9ece87f7.json'
  code: 'ENOENT',
WP-CLI 2.12.0
```

I also forced target deletion during copy and reproduced the narrower `utime` variant:

```text
Failed to set up dependency language packs: Error: ENOENT: no such file or directory, utime '.../server-files/language-packs/admin-id_ID.l10n.php'
  code: 'ENOENT',
  syscall: 'utime',
WP-CLI 2.12.0
```

After this fix, the same concurrent stale-cache repro against the patched built CLI completed cleanly:

```text
warning_matches=0
wp_cli_successes=      24
```

## Testing Instructions

1. `npm run cli:build`
2. `npm test -- apps/cli/lib/dependency-management/tests/setup.test.ts`
3. `npx eslint apps/cli/lib/dependency-management/setup.ts`
4. Run the concurrent repro from #3489 against `node apps/cli/dist/cli/main.mjs wp --studio-no-path cli version` and verify there are no `Failed to set up dependency language packs` warnings.

## AI assistance
- **AI assistance:** Yes
- **Used for:** Investigating the race, drafting the fix, running verification commands, and drafting this PR description. Chris remains responsible for review and validation.


fredrikekelund (approved):
Good catch 👍 I would just inline the logic instead of doing the higher-order function, but it's not a big deal either way

PR #3545, opened 2026-05-20
Experimentally remove Playground

fredrikekelund:
## Related issues

- Fixes RSM-3619

## How AI was used in this PR

Claude drove the entire implementation based on a rough spec from me. This is not meant to be merged, only to gather preliminary data.

## Proposed Changes

This PR experimentally removes the Playground PHP runtime for Studio and relies entirely on the native PHP runtime instead. It also removes the bundled Node.js runtime and instead relies on Electron's built-in Node.js runtime.

I'm opening this PR to understand how this change would affect the Studio installer file size.

## Testing Instructions

N/A

## Pre-merge Checklist


fredrikekelund:
The baseline installer file size is 1376 MB on macOS arm64. With this change, we save ~400 MB by bringing it down to 960 MB.

PR #3551, opened 2026-05-20
Exclude unnecessary items from ASAR archive

fredrikekelund:
## Related issues

- Fixes STU-1746

## How AI was used in this PR

Claude did the implementation based on my spec.

## Proposed Changes

While experimenting with https://github.com/Automattic/studio/pull/3545, I got curious about what exactly makes the Studio installer so large. I noticed that the `Contents/Resources/app.asar` archive contained some directories that aren't needed:

- `__mocks__`
- `e2e` 
- `installers`
- `src`

This PR removes them from the output by changing the `ignore` config in `forge.config.ts`. Several entries in that array aren't relevant after the monorepo migration, either, so Claude cleaned up a bunch of additional ones.

## Testing Instructions

CI should pass.

## Pre-merge Checklist

### DIFF
--- a/apps/studio/forge.config.ts
+++ b/apps/studio/forge.config.ts
@@ -35,45 +35,34 @@ const config: ForgeConfig = {
 				return {};
 			},
 		},
+		// Patterns are matched against paths inside the asar root, which is
+		// apps/studio/ (electron-forge packages from this package). Anchor each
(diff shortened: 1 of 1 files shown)


nightnei (approved):
Great catch 👍

PR #3552, opened 2026-05-20
Move front-end-only dependencies to `devDependencies`

fredrikekelund:
## Related issues

- Fixes STU-1747

## How AI was used in this PR

Claude drove the implementation based on my spec.

## Proposed Changes

The ASAR archive in the Electron app build contains a `node_modules` directory. We need that directory, but it would be ideal to exclude front-end-only dependencies, since they will already have been bundled by Vite. The idea is to save on the installer file size.

This PR does just that. 

## Testing Instructions

CI should pass.

## Pre-merge Checklist


nightnei (approved):
I have only concern about version updates, e.g. "react 18.2.0 to 18.3.1" etc. Is it expected or did it happen by mistake? 
But, we are adding it to the first beta, so we can test everything carefully, catch potential errors caused by version upgrades, and fix them before the actual Studio release. So it LGTM.

fredrikekelund:
> version updates

This was a consequence of how Claude did it. I was able to rerun the installation command in a different way that seems to have mostly preserved the package versions, but just moved them to `devDependencies` 👍

PR #3559, opened 2026-05-20
Add Dolly CLI bump stats

gcsecsey:
## Related issues

- Fixes STU-1739

## How AI was used in this PR

Claude implemented the changes following my spec.

## Proposed Changes

- Add Dolly CLI bump-stat groups and metrics for start, attach, turn outcomes, detach reasons, and weekly/monthly unique usage.
- Emit those stats from the remote-session start and poll-loop lifecycle, respecting `--avoid-telemetry` in foreground and daemon modes.
- Add poll-loop tests covering telemetry opt-out, successful attach/turn/detach bumps, turn status mapping, and auth-error detach classification.

## Testing Instructions

- Build the CLI with the NPM config using `npm run cli:build:npm`
- Open `~/.studio/cli.json` and delete the `lastBumpStats` object 
- Run remote sessions in the foreground `NODE_ENV=development node apps/cli/dist/cli/main.mjs code remote-session start --no-detach`
- Check that you see "Would have bumped stat" messages in the log for the newly added stat groups
  image
- Optionally do a turn through Telegram
- Check that in `~/.studio/cli.json`, the stats are populated under `lastBumpStats`

CleanShot 2026-05-20 at 16 08 10@2x

## Pre-merge Checklist


apps/cli/lib/types/bump-stats.ts
```diff
@@ -10,6 +10,13 @@ export enum StatsGroup {
 	STUDIO_CLI_TOTAL_LAUNCHES_APP = 'studio-cli-lch-tot-app',
 	STUDIO_CLI_SITE_CREATE_NPM = 'studio-cli-site-crt-npm',
 	STUDIO_CLI_SITE_CREATE_APP = 'studio-cli-site-crt-app',
+	// Dolly remote-session (Telegram bot bridge) — see STU-1739.
```
epeicher, line 13: Question about naming, are there any other occurrences of `Dolly`, if not, should we adopt the `wp-agent` naming? 🤔

epeicher (approved):
Thanks @gcsecsey for adding those, they will be very helpful to get feedback! I have tested it, and I can see the stats bumped, while they don't appear when using `--avoid-telemetry`. Changes also LGTM! :shipit: 

|Stats in terminal | Stats in config file| Using `avoid-telemetry` flag
|-------|------|------
| Image| Image| Image

apps/cli/lib/types/bump-stats.ts
epeicher: Thinking again about this, I think it's fine to consider Dolly as the internal codename

gcsecsey: I was thinking about this too, and was wondering if we could just call this feature "Studio remote sessions" maybe. But overall, I think it's fine to keep the internal naming for now, too.

PR #3580, opened 2026-05-21
Spawn PHP with multiple workers on Unix

fredrikekelund:
## Related issues

N/A

## How AI was used in this PR

I reasoned with Claude about how to do this, despite being such a minimal change, and ultimately asked Claude to make the edit, too.

## Proposed Changes

Pass the environment variable `PHP_CLI_SERVER_WORKERS=4` to all native PHP processes. On Unix, this allows the PHP server to accept concurrent requests. Unfortunately, this is not true on Windows. 

Per my research, concurrent PHP requests should not be an issue for SQLite, as it has built-in concurrency handling when running natively on "regular" file systems. It's a different story from Playground, which has to work around WASM limitations to achieve the behavior that SQLite has on native platforms.

## Testing Instructions

This should be tested by using Studio sites extensively. Having verified that this is fine at the conceptual level, I'm fine with CI passing and us running beta testing after landing this.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/php-server-child.ts
+++ b/apps/cli/php-server-child.ts
@@ -578,6 +578,10 @@ async function doStartServer(
 				STUDIO_PHPMYADMIN_PATH: getPhpMyAdminPath(),
 				STUDIO_NATIVE_PHPMYADMIN_WP_ENV_PATH: phpMyAdminWpEnvPath,
 				STUDIO_PHPMYADMIN_SESSION_PATH: getPhpMyAdminSessionPath( config ),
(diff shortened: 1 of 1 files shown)


bcotrim (approved):
Couldn't find any regressions, while navigating the site, installing plugins, installing themes, trying out the new WordPress Desktop mode.

Since we are running beta tests and native PHP benchmarks, I suggest approving this and addressing any potential side effects in a follow-up, if necessary.
LGTM 👍

PR #3588, opened 2026-05-21
Stop native PHP server worker processes

bcotrim:
## Related issues

- Fixes RSM-3785

## How AI was used in this PR

AI helped inspect the restart failure logs, identify the native PHP worker cleanup path, and draft the process-group fix. I reviewed the code and verified it with the checks below.

## Proposed Changes

- Spawn the long-running native PHP built-in server in its own Unix process group.
- Send `SIGTERM`/`SIGKILL` to that process group during stop, restart, startup failure cleanup, and wrapper shutdown so PHP worker children do not keep the port bound.
- Keep Windows behavior unchanged and keep short-lived PHP commands, WP-CLI commands, and Blueprint execution on the existing spawn path.

## Testing Instructions

- Create a native PHP site in Studio.
- While the site is running, change its WordPress version.
- Confirm Studio stops the site, updates WordPress, and restarts it on the same port.
- Confirm logs no longer show `Address already in use`.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/php-server-child.ts
+++ b/apps/cli/php-server-child.ts
@@ -604,10 +604,11 @@ async function doStartServer(
 		} );
 		spawnedChild = serverChild;
 		if ( serverChild.pid !== undefined ) {
-			process.send?.( {
+			const message: ChildMessageRaw = {
 				topic: 'server-process-started',
 				data: { pid: serverChild.pid },
-			} );
+			};
+			process.send?.( message );
 		}
 
 		await new Promise< void >( ( resolve, reject ) => {
--- a/apps/cli/php-server-child.ts
+++ b/apps/cli/php-server-child.ts
@@ -166,7 +166,7 @@ function spawnPhpProcess(
 
 type RunPhpCommandOptions = SpawnPhpProcessOptions;
 
-function killPhpServerProcess( child: ChildProcess, signal: NodeJS.Signals ): void {
+function killProcessGroup( child: ChildProcess, signal: NodeJS.Signals ): void {
 	if ( process.platform !== 'win32' && child.pid ) {
 		try {
 			process.kill( -child.pid, signal );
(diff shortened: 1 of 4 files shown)


apps/cli/php-server-child.ts
```diff
@@ -163,6 +166,19 @@ function spawnPhpProcess(
 
 type RunPhpCommandOptions = SpawnPhpProcessOptions;
 
+function killPhpServerProcess( child: ChildProcess, signal: NodeJS.Signals ): void {
```
fredrikekelund:
```suggestion
function killProcessGroup( child: ChildProcess, signal: NodeJS.Signals ): void {
```
Given that there's also a `killPhpProcess` function, it'd be nice to generalize this name to differentiate them

fredrikekelund (commented):
There's a conceptual issue here: `php-server-child` is different from `playground-server-child` in that it's actually a group of two processes instead of one. With `PHP_CLI_SERVER_WORKERS`, the group obviously becomes bigger. 

The problem is that `process-manager-daemon` spawns `php-server-child` with `detached: true` and then relies on the same negative pid signal mechanism to kill that group. However, with `php-server-child` spawning the native PHP process in its own group, those signals from `process-manager-daemon` never reach the PHP process(es) directly.

This is fine for `SIGTERM`, because `php-server-child` can intercept those signals and "forward" them to the PHP process. It's not fine for `SIGKILL`, which cannot be intercepted. So, if `process-manager-daemon` sends a `SIGKILL` to `php-server-child`, the PHP process(es) are orphaned.

fredrikekelund:
How do we fix this? I recommend reporting the PHP server pid back to the `process-manager-daemon` over IPC, storing it in memory, and then also killing that pid (with the negative pid trick) whenever we kill the main `php-server-child` process. Not bullet-proof, but markedly better than what we have now.

fredrikekelund (approved):
LGTM 👍 I made few tweaks. Most importantly, `broadcastEvent` should be called for all event types, not just `process-message`

PR #3601, opened 2026-05-22
Pin Electron to 41.4.0 to fix macOS Keychain prompt regression

wojtekn:
## Related issues

- Related to https://github.com/electron/electron/pull/50419

## How AI was used in this PR

Used Claude Code to research the regression, identify the offending commit, and apply the fix.

## Proposed Changes

- Pins `electron` from `^42.1.0` to exact `41.4.0` in `apps/studio/package.json`
- Updates `package-lock.json` accordingly

## Testing Instructions

1. Build and launch Studio on macOS
2. Confirm the macOS Keychain prompt **no longer appears** on startup
3. Verify normal site operations (create/start/stop sites) still work

## Context

Electron 42 introduced a regression ([electron/electron#50419](https://github.com/electron/electron/pull/50419)) where ESM named imports from `'electron'` touch the OS keychain on `app.ready` even when `safeStorage` is never called by app code. This causes a macOS Keychain permission dialog ("Electron wants to use your confidential information stored in 'Studio Safe Storage'") to appear on every launch.

The fix is merged into Electron's `main` branch (targeting Electron 43). Since no 42.x backport exists, we pin to 41.4.0 until Electron 43 ships.

## Pre-merge Checklist

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -93,7 +93,7 @@
 		"@wp-playground/blueprints": "3.1.34",
 		"cli-table3": "^0.6.5",
 		"date-fns": "^3.3.1",
-		"electron": "^42.1.0",
+		"electron": "41.4.0",
 		"electron-devtools-installer": "^4.0.0",
 		"electron-vite": "^5.0.0",
 		"patch-package": "^8.0.1",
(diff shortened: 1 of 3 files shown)


fredrikekelund (approved):
Installs cleanly. The app starts fine, and E2E tests are passing. LGTM 👍

I removed the `devDependencies` entry from the root package.json file, as that appeared to have been unintentional

wojtekn:
> I removed the devDependencies entry from the root package.json file, as that appeared to have been unintentional

Thanks. I added it as I couldn't start Electron after the downgrade without that - maybe I should just remove node_modules instead. 🤔

PR #3602, opened 2026-05-22
Native PHP worker pool on all platforms

bcotrim:
## Related issues

- Fixes #

## Proposed Changes

Using `PHP_CLI_SERVER_WORKERS` for concurrency is only supported on macOS and Linux. We want concurrency on Windows, too. This PR accomplishes that by implementing a Node.js HTTP load balancer that proxies requests to a fixed pool of 4 `php -S` processes. This works surprisingly well in our testing.

In our benchmarks, this solution yields ~30% performance improvement on Windows compared to Playground.

## Testing Instructions

CI should pass

## Pre-merge Checklist


fredrikekelund (approved):
LGTM 👍 @bcotrim and I worked on this side by side at our meetup last week. Benchmark-wise, we concluded that there's no noticeable drawback to this approach compared to `PHP_CLI_SERVER_WORKERS`, so we're going with this "poor man's php-fpm" approach on all platforms

apps/cli/lib/native-php/config.ts
fredrikekelund, line 1: The changes in this file fix RSM-3965

.buildkite/commands/run-e2e-tests.sh
fredrikekelund, line 1: The changes in this file and in `.buildkite/pipeline.yml` serve to collect all site logs when an E2E test fails. Site logs are typically forwarded to the main process anyway, but this seems like a sensible thing to keep as the default. It will only make debugging easier

fredrikekelund:
The Linux E2E breakage seems to be CI specific and it exists on trunk, too. I'm proceeding to merge this PR

PR #3615, opened 2026-05-26
site-spec: strengthen Round 1 stop instruction to prevent tool calls in same turn

lezama:
## What

Fixes a bug in the `/site-spec` skill where the agent would ask the site name in text output and immediately call `AskUserQuestion` for Round 2 layout choices **in the same turn** — effectively skipping the name question entirely.

## Why

The previous instruction said *"Stop here and wait for their reply — do NOT call any tools"* but the model violated it. The updated wording makes the constraint explicit: the response must end immediately after the name question, with no tool calls and no additional text.

## Changes

- `apps/cli/ai/skills/site-spec/SKILL.md`: Rewrites the Round 1 stop instruction with a **Critical** callout that spells out the constraint unambiguously.

## Testing

1. Run `npm run cli:build`
2. In Studio AI chat, say something like *"ayudame a crear un sitio"* (no site name provided)
3. Agent should ask for the name in text and **stop** — no tool calls, no layout question
4. Reply with the name → agent should proceed to Round 2 (`AskUserQuestion` for layout)

### DIFF
--- a/apps/cli/ai/skills/site-spec/SKILL.md
+++ b/apps/cli/ai/skills/site-spec/SKILL.md
@@ -16,11 +16,13 @@ Gather preferences through 2 rounds. Keep it concise.
 
 ### Round 1 — Name
 
-Ask the user for their business/site name in your text output. **Stop here and wait for their reply** — do NOT call any tools or continue to the next round. The user needs a chance to type their answer in the prompt.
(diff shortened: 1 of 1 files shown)


wojtekn:
@lezama, I marked it as a draft as there is a conflict to solve. Feel free to pass for review when it's fixed.

lezama:
Closing as part of triaging my studio queue (thanks for the nudge, @wojtekn) — pre-pivot site-spec/skills work, superseded by the shared site-builder library direction (Linear BIGR-660).

PR #3618, opened 2026-05-26
Add deeplink for importing a site from a backup URL

borkweb:
## Summary
Adds a `wp-studio://import-backup?url=<url>` deeplink that downloads a backup over https, validates it, and opens the Add Site modal at the import step with the file ready to go. Optional `&name=<filename>` overrides the filename derived from the URL.

## Why
There was no way for an external site (WordPress.com, a third-party host) to hand a user into Studio with a backup queued for import — they had to download it, open Studio, and re-select the file by hand. The deeplink is triggerable from any web page, email, or IM, so the URL is attacker-controlled, and backups run PHP via Playground on import — hence the defensive URL handling below.

## How
- **Main process** (`handlers/import-backup.ts`): parses the URL without re-decoding (`URLSearchParams.get()` already decodes it; a second decode would allow double-encoding past a future allowlist — the CVE-2021-41773 shape), rejects non-`https:` schemes (blocks MITM payload swaps), downloads to a sanitized temp path, validates extension and non-empty size, then emits `import-backup-from-deeplink`. Failures clean up the temp file and show an error dialog.
- **Renderer**: `useImportBackupDeeplink` feeds the file into the Add Site form and routes to `/backup/create`. Since the file never passes through a browser `File`, `use-import-export.tsx` takes an `ImportSource = File | DeeplinkBackupFile` union and passes `removeBackupOnComplete` so the temp download is deleted after import.

## Testing
- [ ] Valid https `.zip`/`.tar.gz` URL → Studio opens at the backup import step with the name shown.
- [ ] `&name=my-export.tar.gz` → supplied name is used.
- [ ] `http://` or `file://` URL → rejected with "Please check the link and try again."
- [ ] Unsupported extension (`.txt`) → same error dialog.
- [ ] Unreachable host → connection-specific error; "Open Studio Logs" works.
- [ ] Minimized Studio → window restored/focused before the modal opens.
- [ ] `npm test -- apps/studio/src/lib/deeplink/tests/import-backup.test.ts` passes.

## Related issues
* https://github.a8c.com/Automattic/telex/pull/1002


lezama:
Nice defensive work on the URL handling. One gap stands out, plus a smaller follow-up — both worth closing before this lands, since an imported backup runs PHP via Playground.

### 1. The `url` host is unrestricted

The deeplink is triggerable from any web page / email / IM, but `import-backup.ts` accepts **any** `https://` host. In the real flow the backup URL always points at a single known origin (the Telex export endpoint), so a host allow‑list costs almost nothing and removes the "any attacker‑chosen URL gets downloaded and queued for a PHP‑executing import" surface entirely.

```ts
const ALLOWED_BACKUP_HOSTS = [ 'telex.automattic.ai', 'telex.localhost' ];

// exact, case-insensitive match — not endsWith, so `telex.automattic.ai.evil.com`
// and `eviltelex.automattic.ai` are both rejected
if ( ! ALLOWED_BACKUP_HOSTS.includes( parsedUrl.hostname.toLowerCase() ) ) {
    throw new Error( `Backup host "${ parsedUrl.hostname }" is not allowed.` );
}
```

(Open question for you: the PR description mentions "a third‑party host" as an intended source — if that's really desired, the allow‑list conflicts with it and we should talk through the threat model. If Telex is the only producer, the allow‑list is the safer default.)

### 2. The https‑only check can be bypassed by a redirect

`download()` uses `follow-redirects`, which by default will follow an `https → http` redirect (it only strips confidential headers on downgrade, it doesn't block it). So `https://allowed/x.zip` → `302 Location: http://…/payload.zip` is fetched over cleartext, re‑opening exactly the MITM payload‑swap window the `protocol !== 'https:'` guard is meant to close. The initial‑URL check isn't enough.

Cleanest fix is in `download()` itself (also hardens the blueprint deeplink, and never affects `http:`‑initiated downloads):

```ts
const requestOptions = urlProtocol === 'https:'
    ? { beforeRedirect: ( options: { protocol?: string } ) => {
          if ( options.protocol !== 'https:' ) {
              throw new Error( `Refusing redirect to insecure protocol "${ options.protocol }".` );
          }
      } }
    : {};
httpModule.get( url, requestOptions, ( response ) => { /* … */ } );
```

I have both changes implemented locally with the test file updated (allow‑list happy/look‑alike/rejected‑host cases added) — 14 tests passing, lint clean. Happy to push them to this branch or hand them over, whichever you prefer.

#### Smaller notes (non‑blocking)
- **Silent drop when a site is processing:** in `use-import-backup-deeplink.ts`, if `isAnySiteProcessing` the handler `return`s, but the backup is already downloaded — orphaned temp file + no user feedback. Worth a toast and/or main‑side cleanup.
- **Abandoned‑import temp leak:** the temp is only removed via `removeBackupOnComplete` on import completion. If the user closes the modal without importing, the (potentially large) zip lingers in `wp-studio-imports/`.
- **Residual after the allow‑list:** an open redirect on an allowed host could still bounce to an untrusted origin; validating the redirect target host (not just the protocol) would close that too.

lezama:
Pushed both changes to this branch in 835c3493 (host allow-list + the `download()` redirect-downgrade guard, with the test updates). 14 tests passing, lint clean.

Heads-up: pushing required the branch to be current with `trunk`, so that commit sits on top of a clean `trunk` merge (no conflicts) — feel free to re-sync however you prefer. The non-blocking notes above (silent drop on `isAnySiteProcessing`, abandoned-import temp leak, redirect-target host validation) I left for you to decide on.

wojtekn:
@borkweb I'm marking this as a draft as there are some conflicts to solve. Feel free to pass for review when those are fixed.

PR #3627, opened 2026-05-27
Fix `npm start` on Windows: spawn via shell so `npx` resolves to `npx.cmd`

ivan-ottinger:
## Related issues

- Fixes [STU-1761](https://linear.app/a8c/issue/STU-1761/windows-npm-start-fails-with-spawn-npx-enoent)

## How AI was used in this PR

Diagnosed and patched with Claude Code. I reviewed the one-line change and verified it locally on Windows.

## Proposed Changes

`scripts/start-studio.mjs` spawns `npx vite ...` and `npx electron-vite ...` with raw `child_process.spawn(...)`. On Windows, `npx` is `npx.cmd` (a batch file), and Node's `spawn` without `shell: true` only resolves real executables — so `npm start` immediately throws `Error: spawn npx ENOENT` and the dev server never starts. macOS/Linux are unaffected because `npx` is a real binary there.

Fix: pass `shell: true` to the default `spawn` options inside `spawnCommand`. The shell resolves `npx` → `npx.cmd` on Windows and finds `npx` the usual way on macOS/Linux.

This was a latent bug in the predecessor `scripts/start-new-ui.mjs` (only reachable via the opt-in `npm run start:new`). #3537 promoted that script to default `npm start`, turning it into a Windows-wide blocker.

### Why this is safe on macOS/Linux

- All spawn args are hardcoded literals (`'vite'`, `'--port'`, `'5200'`, etc.), so there's no shell-quoting risk.
- `stdio: 'inherit'` still works through the shell wrapper.
- `SIGINT`/`SIGTERM` handlers in the script already call `.kill()` on the child references, so Ctrl+C still works.

Caveat (not a regression vs. current state, but worth knowing): with `shell: true` the immediate child is the shell wrapper. If the parent script is killed abruptly, the inner `vite`/`electron-vite` may briefly outlive the wrapper until they notice their parent is gone. The script's own SIGINT/SIGTERM handlers cover the normal Ctrl+C path.

## Testing Instructions

**Windows (the regression):**

1. `git checkout stu-1761-fix-windows-npm-start-enoent`
2. `npm install && npm start`
3. Expected: prestart build runs, then `Starting @studio/ui dev server on port 5200...` followed by `VITE vX ready in ... ms`, then Electron launches.
4. On `trunk` the same command throws `Error: spawn npx ENOENT` and exits with code 1.

**macOS / Linux (regression check):**

1. `npm start` from this branch should behave identically to `trunk` — vite on 5200, then Electron.

## Pre-merge Checklist


fredrikekelund (approved):
Works great 👍 Thanks for fixing this, @ivan-ottinger

PR #3610, opened 2026-05-25
Add creative-direction skill: expand vague site briefs with site-type content

lezama:
## Related issues

N/A — follows up on internal discussion in #studio-code (May 2026 Studio vs Telex comparison).

## How AI was used in this PR

Claude Code drafted the skill content and tests. All code reviewed and edited by author.

## Proposed Changes

- **New skill** `apps/cli/ai/skills/creative-direction/SKILL.md`: when building a new site from a vague prompt, the agent now infers site type, decides which pages/sections/forms to create, commits to a bold design direction, and briefs the user in 2–4 lines before building — without asking follow-up questions.
- **System prompt** `apps/cli/ai/system-prompt.ts`: step 2 (Plan the design) now loads the `creative-direction` skill instead of just reviewing the design guidelines.
- **12 new tests** covering skill discoverability, body content, and system prompt integration.

## Background

Andrei Draganescu ran a head-to-head comparison between Studio Code and Telex using simple prompts ("Make a site for a bar named Boogie Bar"). Telex won because its system prompt makes opinionated content assumptions for the site type — adding menu, events, reservations, and gallery pages without being asked. Studio would only ask for name and one-page vs multi-page, then build a minimal result.

Key finding: *"The main differences appear only because of system prompt."* This is a prompting/skill gap, not an architecture gap.

Studio's `LOCAL_DESIGN_GUIDELINES` already closely mirrors Telex's `design-direction.md` (bold aesthetics, typography, motion). What was missing is **content intelligence** — knowing what a bar site needs vs a SaaS vs a yoga studio.

## Testing Instructions

**Before (trunk):**
```
studio code
> make me a site for a bar called Boogie Bar
```
Studio asks for name (already given), then one-page vs multi-page, then builds a minimal site: a basic theme, generic contact form.

**After (this branch):**
```
npm run cli:build && node apps/cli/dist/cli/main.mjs code
> make me a site for a bar called Boogie Bar
```
Before building, the agent outputs something like:
> *"Building a 5-page site for Boogie Bar: Home, Menu, Events, Gallery, and Contact. Dark & moody aesthetic with jazz-inspired typography. Includes a reservations form and newsletter signup."*

Then builds it — 5 pages, working forms, site-type-appropriate content — without asking further questions.

**Automated tests:**
```
npm test -- apps/cli/ai/tests/creative-direction-skill.test.ts
```
12 tests, all passing. These same tests fail on trunk (skill doesn't exist there).

## Pre-merge Checklist

[...]

PR #3629, opened 2026-05-27
Add creative-direction skill for vague site briefs

lezama:
## What

Adds a `creative-direction` skill that expands a vague site brief into a rich content and structure plan — choosing pages, sections, hero framing, and copy direction — before the agent starts building.

Follows the skill-routing pattern established in #3605.

## Why

When a user says "make a site for my bar", the agent has to guess what pages, sections, and tone to use. Without a planning step it either under-builds (minimal skeleton) or over-asks (wizard-like back-and-forth). This skill lets the agent commit to a concrete creative direction and then build confidently, with no user interruption.

## Changes

- `apps/cli/ai/skills/creative-direction/SKILL.md` — the skill: site-type inference, page/section inventory, hero framing, copy direction, confidence rules
- `apps/cli/ai/system-prompt.ts` — two small routing lines:
  - Invocation trigger in the global skill-routing block (`For new sites built from a vague brief…`)
  - Reference in the "Plan the design" step of the site-building workflow
- `apps/cli/ai/tests/creative-direction-skill.test.ts` — 14 unit tests covering invocation conditions, confidence rules, and system prompt integration (all pass)

## What changed vs the earlier draft (#3610)

#3610 predated #3605 and made large edits to `system-prompt.ts` (+152 / −0 lines) plus touched several other files. This PR is a clean rebase:

- Skill and tests carried over unchanged
- `system-prompt.ts` delta reduced to **+4 / −2 lines** (routing pointer only, no inline spec)
- All other file edits dropped — they're either superseded by #3605 or belong in a separate PR

## Testing

```bash
npm run cli:build
```
Then in Studio AI chat: *"crea un sitio para una hamburguesería"* — agent should load `creative-direction`, plan pages/sections, then start building without asking for approval.


lezama:
Picking up @youknowriad's question from #3610 — marking this as a PoC so the team has something concrete to evaluate.

**This PR implements approach 2 (auto-expansion, like Telex).** Here's how the three approaches play out in practice:

1. **No expansion (trunk today)**: Agent gets `"make a site for my bar"` and has to guess everything — usually produces a minimal skeleton.
2. **Auto-expansion (this PR / Telex)**: Agent infers site type, commits to pages/sections/aesthetic, briefs the user in 2–3 lines, then builds. No user interruption. This is also what Telex does: it generates a structured site spec (layoutMode, heroComposition, typography) before writing a single file.
3. **Guided expansion (BigSky wizard)**: Agent asks targeted questions before building. More control for the user, more friction in the flow.

The Telex observation is relevant here: Telex's better output quality comes specifically from committing to a `heroComposition` (cinematic prose describing the spatial composition) and inferring typography from site type — both happen before any file is written, without asking the user. That's approach 2.

**On Riad's concern about merging with site-spec**: agreed — a separate skill for the same purpose is redundant. A follow-up PR will fold this into `site-spec` so there's one skill that handles both name/layout gathering *and* creative direction expansion when the brief is vague.

PR #3638, opened 2026-05-29
Remove Windows PFX fallback, rename signing gate

mokagio:
## Related issues

Follow-up to the Studio Azure Trusted Signing migration (AINFRA-2233): removing the PFX fallback now that Azure signing is confirmed in shipped builds.

## How AI was used in this PR

Drafted with Claude Code (Opus 4.8). I reviewed the full diff, ran the type checker and lint, and verified the pipeline/PowerShell/JS syntax locally; the Windows signing path itself only runs on CI.

## Proposed Changes

Remove the PFX signing code path.

Use the new `SIGN_WINDOWS_BUILD` env var to toggle opt a Windows build in to signing. The gate stays explicit and opt-in, so package-only jobs (e.g. Windows E2E, which runs `electron-forge package`) keep producing unsigned bundles without requiring Azure setup.

## Testing Instructions

- CI: the Windows dev and release build jobs must produce Azure-signed `studio-setup.exe` / `.appx` artifacts (this is where the signing path actually executes).
- CI: the Windows E2E job must still package an unsigned bundle without Azure setup.
- Local sanity: `npm run typecheck` passes.


apps/studio/windowsSign.ts
```diff
@@ -7,23 +7,19 @@ import type { WindowsSignOptions } from '@electron/packager';
...
-	const useAzureSigning = [ '1', 'true' ].includes(
-		( process.env.USE_AZURE_TRUSTED_SIGNING ?? '' ).trim().toLowerCase()
-	);
-
-	if ( ! useAzureSigning ) {
+	if ( ! process.env.CI || process.platform !== 'win32' ) {
 		return undefined;
 	}
 
 	if ( ! process.env.AZURE_CODE_SIGNING_DLIB || ! process.env.AZURE_METADATA_JSON || ! process.env.SIGNTOOL_PATH ) {
```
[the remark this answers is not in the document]
mokagio: Good catch — confirmed. The Windows E2E job (`run-e2e-tests.sh`) runs `electron-forge package` for an intentionally-unsigned bundle and never runs `setup_azure_trusted_signing.ps1`, so a `CI && win32` gate would have thrown at config import there.

Digging in, `USE_AZURE_TRUSTED_SIGNING` turned out to be load-bearing rather than redundant: it's the "sign this build" signal that the `make` jobs set and E2E leaves unset. So I've kept it as the gate (in `windowsSign.ts` and the pipelines) and removed only the dead PFX fallback branches. Addressed in https://github.com/Automattic/studio/commit/7dea0c60d2c26600fcd6e7c8b6e5810143c58b6d.

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

apps/studio/windowsSign.ts
```diff
@@ -7,16 +7,18 @@ import type { WindowsSignOptions } from '@electron/packager';
...
-// Controlled by the USE_AZURE_TRUSTED_SIGNING env var:
-// - Unset or not '1'/'true': returns undefined, letting Forge use PFX certificate signing.
-// - '1' or 'true': returns the Azure signing hook config, or throws if the
-//   required Azure env vars are missing.
+// Gated on USE_AZURE_TRUSTED_SIGNING, which the signed-build jobs set (see
+// build-for-windows.ps1). Package-only jobs — e.g. the Windows E2E run, which
+// uses `electron-forge package` and never signs — leave it unset and get
+// undefined, so this config (loaded by Forge on every build) stays inert
+// there. The throw fires only for a build that asked to sign but is missing
+// its Azure env, where signtool would otherwise fail with an opaque exit code.
```
mokagio: I don't like this approach. Looking for a better way.

apps/studio/forge.config.ts
```diff
@@ -116,15 +116,9 @@ const config: ForgeConfig = {
...
-				// PFX mode: use the local certificate file and password.
-				...( windowsSign
-					? { windowsSign }
-					: {
-							certificateFile: path.join( repoRoot, 'certificate.pfx' ),
-							certificatePassword: process.env.WINDOWS_CODE_SIGNING_CERT_PASSWORD,
-					  } ),
+				// Sign via the custom Azure Trusted Signing hook (signtool, SHA256-only).
+				// Undefined off Windows CI, where the build is left unsigned.
+				...( windowsSign ? { windowsSign } : {} ),
```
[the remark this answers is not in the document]
mokagio: Already addressed in the current PR head: the comment now describes the Azure and PFX signing modes without saying the hook is only undefined off Windows CI.

*Posted by Codex (GPT-5) on behalf of @mokagio with approval.*

apps/studio/windowsSign.ts
```diff
@@ -5,24 +5,24 @@ import type { WindowsSignOptions } from '@electron/packager';
 //
 // Uses a custom hook module because the default @electron/windows-sign
 // dual-signs (SHA1 + SHA256), but Azure only supports SHA256.
-// The hook calls signtool directly with SHA256-only parameters.
 //
-// Controlled by the USE_AZURE_TRUSTED_SIGNING env var:
-// - Unset or not '1'/'true': returns undefined, letting Forge use PFX certificate signing.
-// - '1' or 'true': returns the Azure signing hook config, or throws if the
-//   required Azure env vars are missing.
+// Gated on SIGN_WINDOWS_BUILD: signed-build jobs set it, package-only jobs
```
iangmaia, line 9: I'm guessing we should add this file to `eslint.config.mjs`?

iangmaia (approved):
I've disabled auto merge so that you can address my comment related to `eslint.config.mjs` if it makes sense, but otherwise the PR looks good 👍

apps/studio/windowsSign.ts
mokagio: Good call — done in 28e51599. `windowsSign.ts` sits at the `apps/studio` root, outside any tsconfig `include`, so typescript-eslint's project service skipped it entirely (same boat as its sibling `forge.config.ts`, already in the list). Adding it brings the file under type-aware lint — which immediately surfaced a `prettier/prettier` wrap on the Azure-env guard, fixed in the same commit.

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

mokagio:
Thanks @iangmaia address the ESLint suggestion in https://github.com/Automattic/studio/pull/3638/commits/28e51599cc838519d9b3b16b278deef75453a65d

Re-enabled auto-merge

PR #3641, opened 2026-05-29
Studio Code: Add Opus 4.8, remove Opus 4.6

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Implemented with Claude. AI updated the agent model registry and the session model-resolution logic, then verified there were no other references to the removed model id across the codebase. Reviewers should pay particular attention to the resume/auto-switch behavior in `resolveSessionModel` and the CLI resume path.

## Proposed Changes

- Adds **Opus 4.8** to the Studio Code agent's selectable models and removes **Opus 4.6**, keeping the model picker aligned with the models we actually support.
- Sessions that were last using a model we no longer offer (e.g. Opus 4.6) now **auto-switch to the default model** when resumed, instead of pinning a model that's gone. The most recently recorded model wins, and an unknown/removed model falls back to the default. Because the fallback points at the default model rather than a hardcoded id, changing the default later automatically updates where these sessions land.
- Both the desktop UI and the CLI resume path share the same resolution logic, so behavior is consistent across surfaces.

## Testing Instructions

- Open a chat in Studio Code and confirm the model picker lists **Sonnet 4.6**, **Opus 4.7**, **Opus 4.8**, and **GPT 5.5**, with **no Opus 4.6** entry.
- Resume an existing session that was last on Opus 4.6 and confirm it comes back on the default model (Sonnet 4.6) rather than erroring or showing a missing model.
- Resume a session on a still-supported model (e.g. Opus 4.7) and confirm it resumes on that same model.
- Start a brand-new session and confirm it uses the default model.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/sessions/context.ts
+++ b/apps/cli/ai/sessions/context.ts
@@ -1,4 +1,4 @@
-import { isAiModelId, type AiModelId } from '@studio/common/ai/models';
+import { resolveSessionModel, type AiModelId } from '@studio/common/ai/models';
 import { isStudioCustomEntryOfType } from '@studio/common/ai/sessions/entry-types';
 import { AI_PROVIDERS, type AiProviderId } from 'cli/ai/providers';
 import type { LoadedAiSession } from '@studio/common/ai/sessions/types';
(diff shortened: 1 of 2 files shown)


youknowriad:
I'm not sure if Opus produces better results than Sonnet. I prefer Sonnet but still this adds the latest Opus at least.

epeicher (approved):
Thanks @youknowriad! I have tested changing between models and resuming sessions started with previous models, and they all work as expected. I have not yet compared Sonnet vs Opus, but I agree rolling up to the latest is a good idea. LGTM!

|Models for a new session | Resuming an Opus 4.6 session| Resuming an Opus 4.7 session
|-------|------|----
|   Image| Image|Image

PR #3648, opened 2026-05-31
Group Linux CI test steps and add Linux native PHP E2E step

bcotrim:
## Related issues

- N/A

## How AI was used in this PR

AI implemented the change, I validated

## Proposed Changes

- Linux Unit and E2E tests were showing up as different steps, this PR aims to group them with the existing ones.
- The new test groups include stable keys so future steps can depend on the whole Unit or E2E group if needed.
- Adds the new `e2e_tests_linux_native_php` Buildkite step so Linux native-PHP E2E runs under the existing `native-php` label gate, matching the macOS and Windows native-PHP coverage.

| Before | After |
|--------|--------|
| image | image | 

## Testing Instructions

- Confirm CI tests ran as expected.
- Confirm the `native-php` label triggers the new `e2e_tests_linux_native_php` step alongside macOS and Windows native-PHP E2E.

## Pre-merge Checklist


.buildkite/pipeline.yml
```diff
@@ -48,77 +48,103 @@ steps:
...
-      - test-results/**/*.zip
-      - test-results/**/*.png
-      - test-results/**/*error-context.md
-    env:
-      DEBUG: "pw:browser"
-    # TEMP(rsm-2593): if: removed to force E2E on every push while iterating on Linux E2E setup. Revert before merge.
-    notify:
-      - github_commit_status:
-          context: E2E Tests
+  - group: Unit Tests
```
gavande1, line 51: Should we add a `key:` to the new `Unit Tests` and `E2E Tests` groups? The existing `Build for Mac/Windows/Linux` groups all declare one, and keeping that consistent would let future steps `depends_on` the whole group if we ever need it. Happy to leave it if you'd rather not expand scope here.

.buildkite/pipeline.yml
```diff
@@ -48,77 +48,103 @@ steps:
...
+          - test-results/**/*error-context.md
+        env:
+          DEBUG: "pw:browser"
+        # TEMP(rsm-2593): if: removed to force E2E on every push while iterating on Linux E2E setup. Revert before merge.
+        notify:
+          - github_commit_status:
+              context: E2E Tests
+
+      - label: E2E Tests on linux-x64 with native PHP
+        key: e2e_tests_linux_native_php
```
gavande1, line 130: I also noticed this PR also adds the `e2e_tests_linux_native_php` step, which isn't mentioned in the description. Does it make sense to split it into a separate PR, or was it added by mistake?

bcotrim: I intentionally added it, but you are right about the description.
Are you ok with pushing it together, with updated description, or would you prefer a separate PR?

gavande1: Clubbing them together is fine by me — would just be nice to mention the new step in the description so it's easy to spot later.

gavande1 (approved):
Thanks for the changes, looks good to me! 🚀

PR #3649, opened 2026-06-01
Use next minor version for trunk dev builds (x.(y+1).0-devN)

wojtekn:
## Related issues

- STU-1759

## How AI was used in this PR

Claude made the one-line change to `scripts/prepare-dev-build-version.mjs`.

## Proposed Changes

- Trunk dev builds were previously versioned as `x.y.z-devN` (e.g. `1.9.0-dev5`), which sorts *below* any `1.10.0-betaN` or the last stable release by semver.
- Now dev builds are versioned as `x.(y+1).0-devN` (e.g. `1.10.0-dev5`), which is unambiguously newer than any `1.9.x` stable or `1.9.x-beta`, and also newer than `1.10.0-betaN` (semver pre-release sorts before the release).

## Testing Instructions

- Run `GITHUB_SHA=abc123 node ./scripts/prepare-dev-build-version.mjs` (with a real tag in the repo) and verify `apps/studio/package.json` shows the bumped minor version with a `-devN` suffix.

## Pre-merge Checklist

### DIFF
--- a/scripts/prepare-dev-build-version.mjs
+++ b/scripts/prepare-dev-build-version.mjs
@@ -24,8 +24,9 @@ if ( ! parsedVersion ) {
 	throw new Error( `Invalid version in latestTag: ${ latestTag }` );
 }
 
-// Create dev version using just the core version numbers (major.minor.patch)
-const devVersion = `${ parsedVersion.major }.${ parsedVersion.minor }.${ parsedVersion.patch }-dev${ commitCount }`;
+// Create dev version targeting the next minor release (major.minor+1.0-devN),
+// so trunk builds sort above any stable or beta of the last release.
+const devVersion = `${ parsedVersion.major }.${ parsedVersion.minor + 1 }.0-dev${ commitCount }`;
 
 packageJson.version = devVersion;
 


wojtekn:
It seems to work fine - for 432e17aa5f6aacd088e7571cea4b95b7e95775a3 it generates `1.10.0-dev167` instead of `1.9.0-dev169`.

sejas (approved):
It looks good to me. We just need to make sure next stable release will bump the minor version by one. Patch versions will be ignored if you have a dev version.

sejas (commented):
We could consider bumping the patch version? 🤔 

`${ parsedVersion.major }.${ parsedVersion.minor }.${ parsedVersion.patch + 1 }-dev${ commitCount }`

wojtekn:
> It looks good to me. We just need to make sure next stable release will bump the minor version by one. Patch versions will be ignored if you have a dev version.

I think it's okay - if a user switches from 1.9.0 to nightly line so they use 1.10.0-dev12, and we release patch version 1.9.1 later, they will already be running 1.10.0-dev25 or whatever dev build was released in the meantime.

wojtekn:
> We could consider bumping the patch version? 🤔
> ${ parsedVersion.major }.${ parsedVersion.minor }.${ parsedVersion.patch + 1 }-dev${ commitCount }

This makes sense, should work too. Changed.

wojtekn:
@sejas actually, if we update the patch version, it won't let users running 1.9.0 update from 1.10.0-beta1 to the dev version which would be 1.9.1.-dev123.

sejas:
> @sejas actually, if we update the patch version, it won't let users running 1.9.0 update from 1.10.0-beta1 to the dev version which would be 1.9.1.-dev123.

I think since the moment we release beta1, then the devs will use that latest version, isn't it? Like 1.10.0-beta1 -> next dev will be 1.10.1-dev123 🤔 

UPDATE: the code reads the latest stable tag, so we would need to modify what tags we get to include betas too.

PR #3651, opened 2026-06-01
Studio Code: use the agentic session UI in the assistant tab

youknowriad:
## How AI was used in this PR

Built end-to-end with Claude Code: ported the agentic `SessionView` subtree from `apps/ui` into the Studio desktop renderer, rewired its data layer to the renderer's existing IPC surface (intentionally **without** the connector abstraction), trimmed it to a single-session experience, and resolved the `@wordpress/ui`-vs-Tailwind-preflight CSS-layering conflicts. Every change was reviewed; `npm run typecheck`, ESLint, and a full `electron-vite` build all pass. Runtime UI was verified manually (see screenshot).

## Proposed Changes

The Assistant tab's experimental "Studio Code" mode (behind the `enableStudioCodeUi` flag) previously used a minimal, bespoke chat that spawned the CLI once per turn and kept history in `localStorage`. This replaces it with the same rich agentic session UI the standalone agentic app uses — streaming transcript, inline tool-call rendering, agent questions answered in place, queued follow-ups, interrupt, and a model picker.

To keep the tab focused, it runs as a **single session per site**: no session list or switching, and "New conversation" simply starts a fresh session. History now persists on disk through the existing CLI session store (the same run-manager backend the agentic app already drives), instead of `localStorage`.

User impact (flag on): a noticeably more capable assistant with live tool/permission handling and persistent history — behaviorally close to the agentic app, minus session management.

## Testing Instructions

1. Run with the flag enabled: `ENABLE_STUDIO_CODE_UI=true npm start`.
2. Open a site → **Assistant** tab.
3. Send a prompt; confirm the streamed transcript, tool-call rows, and thinking indicator render.
4. Trigger an agent question and answer it inline; queue a follow-up while a turn is running; press Stop mid-turn.
5. Switch the model to a different family → confirm the "Start a new conversation?" dialog appears and that confirming starts a fresh session.
6. Click **New conversation** → confirm the transcript resets, and that reopening the tab restores the site's latest session.

## Screenshots

Screenshot 2026-06-01 at 12 56 43 PM

## Pre-merge Checklist


apps/studio/package.json
```diff
@@ -89,9 +92,13 @@
 		"@wordpress/dataviews": "^15.0.0",
 		"@wordpress/element": "^7.0.0",
 		"@wordpress/icons": "13.2.0",
+		"@wordpress/private-apis": "^1.12.0",
 		"@wordpress/react-i18n": "^4.47.0",
+		"@wordpress/theme": "^0.14.0",
+		"@wordpress/ui": "^0.14.0",
 		"@wp-playground/blueprints": "3.1.34",
 		"cli-table3": "^0.6.5",
+		"clsx": "^2.1.1",
```
sejas: We can consider using existing `cx` function ([code](https://github.com/Automattic/studio/blob/trunk/apps/studio/src/lib/cx.ts#L1-L3) ).

See the example: 

https://github.com/Automattic/studio/blob/2a8abbef2fab1d48ce2f07f1e843bcc65d8d842a/apps/studio/src/components/ai-input.tsx#L228

apps/studio/src/components/studio-code-session/markdown/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+			...baseComponents,
+			// Electron swallows plain `target="_blank"` navigations — route clicks
+			// through the IPC bridge so links open in the system browser.
+			a: ( { children: linkChildren, href } ) => {
+				const handleClick = ( event: MouseEvent< HTMLAnchorElement > ) => {
+					if ( ! href ) {
+						return;
+					}
+					event.preventDefault();
+					void getIpcApi().openURL( href );
```
sejas, line 75: We could consider allowing only links pointing to http sites. That way we are sure that displayed links in Markdown won't open other protocols like `vscode://` or `smb://`. Something like:

```diff
  +const SAFE_URL_SCHEMES = new Set( [ 'http:', 'https:' ] );
  +
  +function isSafeExternalUrl( href: string ): boolean {
  +     try {
  +             return SAFE_URL_SCHEMES.has( new URL( href ).protocol );
  +     } catch {
  +             // Relative or malformed — never hand to the OS.
  +             return false;
  +     }
  +}

   a: ( { children: linkChildren, href } ) => {
        const handleClick = ( event: MouseEvent< HTMLAnchorElement > ) => {
  -             if ( ! href ) {
  +             event.preventDefault();
  +             if ( ! href || ! isSafeExternalUrl( href ) ) {
                        return;
                }
  -             event.preventDefault();
                void getIpcApi().openURL( href );
        };
```

sejas (approved):
It works amazing! 🎉
Great work. I tested it with an existing site and I was able to install Jetpack plugin and create a Contact form page. The UI is very polished and the counting time sets a clear expectation that the agent is working.

I left a couple of suggestions for you to consider.

https://github.com/user-attachments/assets/42a496d0-4337-427c-8d1a-940145399878

sejas:
Once this PR is merged I'll create follow-up issues like:
- Stop the autoscroll when the user is reading a previous message.
- Add slash commands autocomplete

youknowriad:
another follow-up I'd like to do is to add prompt suggestions like we have in the current assistant tab or also in desks-ui

apps/studio/package.json
youknowriad: Good call — switched to `src/lib/cx` and dropped the `clsx` dependency (it wasn't in trunk). Done in a02cd9fa4.

apps/studio/src/components/studio-code-session/markdown/index.tsx
youknowriad: Done in a02cd9fa4 — added an `isSafeExternalUrl` guard so only `http:`/`https:` URLs are handed to `openURL`; everything else (relative, malformed, `vscode://`, `smb://`, …) is preventDefaulted and dropped.

PR #3653, opened 2026-06-01
Studio Code: example prompt suggestions for empty conversations

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Built with Claude Code: implemented the empty-state suggestions, the composer hover-preview/draft wiring, the loading state, and the styling. The author reviewed the diff and iterated on the visual design.

## Proposed Changes

When the Studio Code UI is enabled, the assistant tab now uses the agentic session UI but showed nothing for a brand-new conversation or while the session was loading. This adds a friendlier empty state and a loading state:

- On an empty conversation, a short set of example-prompt chips is shown (build a plugin, create a block, fix an error, etc.), mirroring the suggestions in the desks UI.
- Hovering (or focusing) a chip previews its full prompt in the composer in muted text without committing it; moving away restores whatever the user had typed.
- Clicking a chip populates the composer with the full prompt so the user can edit before sending. The suggestions disappear once the conversation has a real prompt or a run is active.
- While the session is loading, a centered spinner is shown in the conversation area instead of a blank tab.
- The assistant tab's grey surface background is dropped when the Studio Code UI is on, so the conversation matches the white background of the other site tabs.

## Testing Instructions

1. Enable the Studio Code UI (`ENABLE_STUDIO_CODE_UI` / `enableStudioCodeUi` feature flag).
2. Open a site and go to the Assistant tab. While the session loads, confirm a centered spinner appears (no blank tab).
3. With no prior conversation, confirm the example-prompt chips appear, are outlined and clearly visible against the white background.
4. Hover a chip: the composer shows the full prompt in muted text. Move away: it clears (and any text you typed is preserved).
5. Click a chip: the composer is populated with the full prompt and focused, cursor at the end. Edit/send as normal.
6. Once you send a prompt (or a run is active), the suggestions no longer show.
7. With the flag off, the legacy assistant tab (including its grey background) is unchanged.

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/components/studio-code-session/index.tsx
+++ b/apps/studio/src/components/studio-code-session/index.tsx
@@ -31,6 +31,7 @@ import { queryClient } from './query-client';
 import { QueuedPrompts } from './queued-prompts';
 import styles from './style.module.css';
 import { AgentRunProvider, useAgentRun } from './use-agent-run';
+import { useExamplePrompts } from './use-example-prompts';
 import { useSession } from './use-session';
 import { useSingleSession } from './use-single-session';
 import buttonDefense from './wp-ui-button-defense.module.css';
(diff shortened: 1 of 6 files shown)


apps/studio/src/components/studio-code-session/index.tsx
```diff
@@ -92,6 +106,93 @@ function UnauthenticatedNotice( { onAuthenticate }: { onAuthenticate: () => void
 	);
 }
 
+const EXAMPLE_PROMPTS: { short: string; full: string }[] = [
```
sejas: Let's move these suggested prompts inside a function/hook so they are translated if the user switches languages.

sejas (approved):
It works well. I like having the suggested prompts, and these very actionable. I like that we don't have one about creating a site from scratch.

https://github.com/user-attachments/assets/fc4b99a9-91b9-42e9-a225-7e49bbd4aa41

I noticed a small edge case issue that can be improved in this PR or as follow up. Here are the steps to reproduce:

1. Type anything in the text input, such as "123456."
2. Move the mouse over the suggested button, "Build a plugin."
3. Start typing "hello."
4. Observe that the "Build a plugin" text has overwritten my previous text without me clicking the button. Also observe that it added one more character at the end of the prefilled text, and not the full word "hello".


https://github.com/user-attachments/assets/39ed34fd-8690-4538-b7c1-a16a0f8c005c

apps/studio/src/components/studio-code-session/index.tsx
youknowriad: I don't understand this, why wouldn't they be translated as is?

sejas: The `EXAMPLE_PROMPTS` is a "global" constant and it will be translated only once when Studio starts. So, if the user installs and starts the app in English, they will be translated into English and the translation will be saved inside the object. If the user then switches to another language, such as Spanish, the constants will still be in English. If we call `__` inside a function, hook or component, then those will be translated in every render or when the language changes if we use `const { __ } = useI18n();`

youknowriad: That makes sense, but I sense that this is not the only place where we have this issue. I wonder if we should have a lint issue about this instead?

youknowriad: Good catch — moved the suggestions into a `useExamplePrompts` hook backed by `useI18n` (615db9826), so they re-translate when the locale changes instead of being frozen at module-load time. Also added stable `id`s for the React keys so a language switch doesn't remount the chips.

sejas: Agreed! A linter would be great. I created the issue and assigned it to myself. [STU-1771](https://linear.app/a8c/issue/STU-1771/add-a-lint-rule-and-an-agentsmd-rule-to-enforce-translating-strings)

PR #3656, opened 2026-06-01
Studio Code: inline slash-command autocomplete

sejas:
## Related issues

- Related to STU-1589

## How AI was used in this PR

AI-assisted implementation and unit tests.

## Proposed Changes

Typing `/` in the Studio Code composer now opens an inline, keyboard-navigable list of the available skill commands, filtered live as you type.
The match is partial not requiring to type the start of the command.
Picking one (↑/↓ then Enter/Tab, or click) inserts `/<command> ` into the textarea — it does **not** send — so you can add context before submitting.
The popup keeps focus in the textarea, closes on Escape, and never shows while an example-prompt preview is active.
Updated the icon `/` inside the text input. It's also inserts the slash char and triggers the popup inside the text input.

This makes the skill commands discoverable from the keyboard.

## Testing Instructions

- Open Studio Code, focus the composer.
- Type `/` → list of commands appears above the textarea.
- Type `/an` → filters to matching commands.
- ↑/↓ to move, Enter or Tab to insert `/<command> ` (verify it does not send).
- Click a command → same insert behavior, focus stays in the textarea.
- Escape closes the popup.
- Confirm the existing `/` button still works. In this case it inserts the slash character and open the modal there, so the user can type or click on the list.

https://github.com/user-attachments/assets/8cadb823-71c4-4503-85da-f8a828355cf6

## Pre-merge Checklist


youknowriad:
I have a small preference for the previous "/" button personally. Why did we move it to the bottom of the prompt, was that a requirement?

sejas:
> I have a small preference for the previous "/" button personally. Why did we move it to the bottom of the prompt, was that a requirement?

Not really. I'm happy to bring the previous button back and keep the automcomplete. The button will just insert the slash character in the text field which will open only one modal.
I prefer if we don't auto-submit the skill when selecting one from the tooltip so the user can be more specific about what they want with that skill. WDYT?

youknowriad:
> I prefer if we don't auto-submit the skill when selecting one from the tooltip so the user can be more specific about what they want with that skill. WDYT?

I don't mind but I think submitting the slash command is a slightly different behavior than actually just typing the text (there's an auto-expansion behavior that might not work properly, we need to confirm how needed is that behavior)

apps/studio/src/components/studio-code-session/composer/index.tsx
```diff
@@ -230,56 +276,99 @@ export function Composer( {
...
+								if ( event.key === 'Enter' && ( event.metaKey || event.ctrlKey ) ) {
+									event.preventDefault();
+									void send();
+								}
+							} }
+							rows={ 2 }
+						/>
+						{ slashOpen ? (
+							<ul
+								className={ `${ styles.autocompletePopup } ${ motionStyles.motion }` }
```
youknowriad: The `motionStyles.motion` class is currently inert here. Its enter/exit transitions are gated entirely on `[data-starting-style]` / `[data-ending-style]` (see `floating-surface-motion/style.module.css`), which Base UI's `Menu.Popup` toggles automatically during mount/unmount. On this plain `<ul>` those attributes are never set, so the popup just snaps in/out — the class only contributes a `transition` declaration and `transform-origin` that never fire.

We don't want to discard the animation, so let's restore it rather than leave a no-op class. Options:
- Drive the attributes manually: render the `<ul>` through a small mount/unmount transition (e.g. set `data-starting-style` on first paint and `data-ending-style` before removal) so the existing motion rules engage, or
- Use the same presence/animation primitive `Menu.Popup` uses under the hood.

As-is, a reviewer reading the code would reasonably assume it animates when it doesn't.

apps/studio/src/components/studio-code-session/composer/index.tsx
```diff
@@ -230,56 +276,99 @@ export function Composer( {
...
+								data-side="top"
+								data-align="start"
+								role="listbox"
```
youknowriad: Accessibility: this is a combobox-style autocomplete but the listbox isn't wired to the textarea. The `<ul>` has `role="listbox"` and the options set `aria-selected`, but the focused element (the textarea) has no `aria-expanded`, `aria-controls`, or `aria-activedescendant` pointing at the active option. As a result, screen readers won't announce that the list opened or which command is highlighted as the user arrows through — the old `Menu.Root` gave us this for free.

Suggested wiring:
- On the textarea: `role="combobox"`, `aria-expanded={slashOpen}`, `aria-controls={listboxId}`, and `aria-activedescendant` set to the highlighted option's id when open.
- Give the `<ul>` a stable `id` and each `<li>` a derived `id` (e.g. `${listboxId}-${command.name}`).

That restores parity with the dropdown we replaced.

apps/studio/src/components/studio-code-session/composer/index.tsx
youknowriad: I wonder if `@wordpress/ui` has an Autocomplete component that helps us here instead of inventing our own and wire up a11y, motion...

I also feel the code of the composer component is becoming a bit confusing and hard to follow, maybe there's some opportunity to extract stuff

youknowriad:
Behavior wise, this seems to work properly.

sejas:
@youknowriad, thanks for pointing out the accessibility and animation issues. I’ll address them now. I brought back the slash button https://github.com/Automattic/studio/pull/3656/changes/e97e04bde93e416c3a0b2049ecaf89ce739a71c3 and updated the PR description and video.

apps/studio/src/components/studio-code-session/composer/index.tsx
sejas: Fixed it in https://github.com/Automattic/studio/pull/3656/commits/98dababc276b41ef7961aed324c2c1ffee23e524

apps/studio/src/components/studio-code-session/composer/index.tsx
sejas: I tried different approaches with existing components, but the issue is that they move the focus from the text input to the popup, so user cannot continue typing.

sejas: I pushed the animation changes here https://github.com/Automattic/studio/pull/3656/commits/fd2bf6b81006360d2caa677ef884a175e40a651d 


https://github.com/user-attachments/assets/dc9258cc-bee4-41d8-9673-dc5a36206404

youknowriad:
One bug I noticed is that I can click the "/" button indefinitely and it keeps adding "/" to the prompt

youknowriad (approved):
LGTM behavior wise, I'm a bit concerned that the composer component is becoming more complex / harder to maintain for side featureS.

sejas:
I added the logic to close the modal when clicking the slash button, so clicking it twice will toggle the modal instead of adding multiple slash commands.

I also refactored the code to extract the logic related to the slash command into a separate hook. That way the composer class will be cleaner.


https://github.com/user-attachments/assets/48f0378b-4206-4d09-aab9-653c204d0759

PR #3659, opened 2026-06-02
Always use default UI mode in production builds

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

Claude identified the fix location and implemented the change.

## Proposed Changes

When a user enables the desk or agentic UI in dev mode, the preference is persisted in `app.json`. Opening a production build would then load that stored mode and render the non-default UI, even though desk/agentic UI is a dev-only feature.

This fix makes `getStudioUiMode` return `'default'` unconditionally when `app.isPackaged` is true, so production builds are never affected by a dev-mode UI preference.

## Testing Instructions

1. Run the app in dev mode
2. Enable the new UI via the feature flag and switch to desk UI
3. Close dev mode
4. Open a production build — it should show the standard (default) UI

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/modules/desks/lib/ipc-handlers.ts
+++ b/apps/studio/src/modules/desks/lib/ipc-handlers.ts
@@ -1,4 +1,4 @@
-import { BrowserWindow, dialog, type IpcMainInvokeEvent } from 'electron';
+import { app, BrowserWindow, dialog, type IpcMainInvokeEvent } from 'electron';
 import fsPromises from 'fs/promises';
 import nodePath from 'path';
 import { assertDeskConfig } from '@studio/common/lib/desk-config';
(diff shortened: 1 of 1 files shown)


apps/studio/src/modules/desks/lib/ipc-handlers.ts
```diff
@@ -51,6 +51,9 @@ export async function getDeskSettings( _event: IpcMainInvokeEvent ): Promise< De
 }
 
 export async function getStudioUiMode( _event: IpcMainInvokeEvent ): Promise< StudioUiMode > {
+	if ( app.isPackaged ) {
```
youknowriad: I think instead of isPackaged, we should be checking the feature flag here. In other words, if the feature flag is disabled, default UI, if the feature flag is on, you can switch. It's more generic and protect us from forgetting this when we make this a beta flag for instance.

wojtekn: But the feature flag menu is not available in the production build. This condition ensures that the new UI is not used in a production build until we make it fully available there. I adjusted it to check for node env, too, so it's aligned with update polling behavior.

I don't think we could forget about this change, as without removing it, we wouldn't be able to see the new UI when we release a beta that makes it available to users.

youknowriad (approved):

PR #3661, opened 2026-06-02
Studio Code: follow the conversation to a newly created site

youknowriad:
## How AI was used in this PR

Built end-to-end with Claude: investigated how the standalone Desks UI (`apps/ui`) handles an agent creating a site mid-conversation, then ported the equivalent behavior to the embedded Studio Code assistant tab. Reviewer should focus on the session-ownership remapping in `use-site-creation-switch.ts` and the open-vs-stay branching.

## Proposed Changes

When you're chatting with Studio Code on one site and the agent creates a **new** site, the session is re-homed to that new site behind the scenes. Previously the assistant tab ignored this: it stayed on the current site and the conversation effectively became orphaned (the new site's tab started fresh, with no way back to the chat where the site was created).

Now the tab reacts to the site being created and asks what you want to do, mirroring the Desks UI's "Continue in the site desk?" prompt:

- **Open {site}** — switches to the new site, whose assistant tab loads the full transcript and keeps streaming the in-flight run. The current site lets go of the moved conversation so it isn't owned in two places.
- **Stay here** — starts a fresh conversation on the current site; the migrated one waits on the new site.

In both cases the new site owns the conversation, so there's no duplication. If the new site hasn't yet propagated into the site list when you choose **Open**, the switch is deferred until it appears (avoids briefly flashing the wrong site).

## Testing Instructions

1. Open the Studio Code assistant tab on an existing site (requires WordPress.com login; `enableStudioCodeUi` feature flag on).
2. Ask the agent to create a new site (e.g. "create a new site called Test").
3. When the site is created, confirm the **Site created** dialog appears.
4. Choose **Open {site}**: the app switches to the new site and the same conversation continues there; returning to the original site shows a fresh chat.
5. Repeat and choose **Stay here**: the current site starts a new chat; opening the new site later shows the migrated conversation.
6. `npm test -- src/components/studio-code-session` — 7 passing tests cover the mapping, session/site filtering, both branches, the deferred switch, and the dismissal guard.

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/components/studio-code-session/index.tsx
+++ b/apps/studio/src/components/studio-code-session/index.tsx
@@ -32,11 +32,13 @@ import { unlock } from './lock-unlock';
 import { queryClient } from './query-client';
 import { QueuedPrompts } from './queued-prompts';
 import { isScrolledToBottom } from './scroll-utils';
+import { SiteCreatedDialog } from './site-created-dialog';
 import styles from './style.module.css';
 import { AgentRunProvider, useAgentRun } from './use-agent-run';
 import { useExamplePrompts } from './use-example-prompts';
 import { useSession } from './use-session';
 import { useSingleSession } from './use-single-session';
+import { useSiteCreationSwitch } from './use-site-creation-switch';
 import buttonDefense from './wp-ui-button-defense.module.css';
 import type { SessionEntry } from '@mariozechner/pi-coding-agent';
 import '@wordpress/theme/design-tokens.css';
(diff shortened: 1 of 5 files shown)


sejas (approved):
Approving as it works as described.
From the user’s perspective, it feels a bit strange that clicking "Stay here" still deletes and creates a new conversation. I know it's explained in the modal.
Maybe we could move the conversation at the same time the modal opens. That way, clicking the modal isn’t perceived as the trigger for deleting the conversation. The modal says, in the past tense, that the conversation has moved, but it doesn’t actually move until we close the modal.
The solution is very strong. Thanks for creating the PR. Another alternative I thought of was having a global chat interface. So only one conversation at the same time, but it also has its challenges.

Image

https://github.com/user-attachments/assets/487e4151-4b09-4d07-9a90-163f8cdbfe85

youknowriad:
Good catch on the timing, @sejas — you're right that the past-tense copy didn't match reality, since the move only happened when the modal was dismissed.

Addressed in f17b81b4c: the conversation now migrates the moment the placement event arrives (the modal opens) — the new site claims the session and the current site resets to a fresh chat right away. So the prompt is now a pure navigation choice:
- **Open {site}** → go to the new site, where the conversation continues.
- **Stay here** → just dismiss; this site already shows its new chat.

This makes "has moved" accurate and means clicking is never the thing that discards the chat. Tweaked the copy to "…or stay here with a fresh chat" to match. Tests updated accordingly.

(Noting your global-single-chat alternative too — agreed it has its own tradeoffs; happy to explore separately if we want to revisit the per-site model.)

PR #3671, opened 2026-06-03
Increase npm dependency cooldown to two days

fredrikekelund:
## Related issues

N/A

## How AI was used in this PR

See https://github.com/Automattic/studio/pull/3436 and https://github.com/Automattic/studio/pull/3440 for prior art

## Proposed Changes

There's no "right value" for `min-release-age`, but the consensus seems to be that `1` is a low value. `7` is regularly recommended. I believe that's a little high for us, though, so I'm bumping this by just one more day to `2`. 

This buys us a little more security while still being low enough that we won't have to think about it very often.

## Testing Instructions

Code review is sufficient

## Pre-merge Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,7 @@ updates:
       interval: "weekly"
 
     cooldown:
-      default-days: 1
+      default-days: 2
 
     allow:
       # Automattic packages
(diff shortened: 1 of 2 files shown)


fredrikekelund:
Approved in p1780472407318689-slack-C0ASQG4MR5Y

PR #3677, opened 2026-06-03
Differentiate preview sites from connected WordPress.com remote sites

sejas:
## Related issues

- Fixes STU-1775

## How AI was used in this PR

Used an AI coding assistant to find the root cause, implement the fix, and write a regression eval. All changes reviewed manually.

## Proposed Changes

Studio Code (the AI agent) was calling a **temporary preview site** a **connected WordPress.com site** when telling users what a local site is linked to.

These are two different things:
- **Connected WordPress.com site** — a real, durable site you push/pull to.
- **Preview site** — a temporary, expiring share link (`*-studio.wp.build`).

The agent had no reliable way to tell them apart, so it guessed from the URL/name and got it wrong.

The fix: each listing now tags every entry with a `type` field (`wpcom-remote` or `preview`), and the preview listing adds an `isExpired` flag. The agent reads the tag instead of guessing, and the tool descriptions no longer call previews "WordPress.com" sites.

**User impact:** when you ask Studio Code what a site is connected to, preview sites and connected WordPress.com sites are now reported correctly, each under its own category.

A regression eval (`scripts/eval/`) seeds both kinds of site and asserts the agent keeps them distinct. The human-facing `studio preview list` output is unchanged.

## Testing Instructions

* Run `npm run cli:build && node apps/cli/dist/cli/main.mjs code`
* Select a local site
* Ask _What wpcom sites do I have?_
* Ask _Publish my site_
* Ask _Create a preview site_
* Confirm responses make sense and target the correct type of site.
* If you have Promptfoo variables configured, run `npm run eval -- --filter-pattern "preview sites"` and confirm the new eval passes correctly.

Screenshot 2026-06-23 at 17 38 58

Screenshot 2026-06-23 at 17 22 13

Screenshot 2026-06-23 at 17 22 56

## Pre-merge Checklist


apps/cli/ai/tools/create-preview.ts
```diff
@@ -6,7 +6,7 @@ import { resolveSite } from './utils';
 
 export const createPreviewTool = defineTool(
 	'preview_create',
-	'Creates a WordPress.com preview site from a local Studio site. Requires WordPress.com authentication. This can take a few minutes, so tell the user to wait after starting it.',
+	'Creates a preview site from a local Studio site. A "temporary site", "temporal site", or "share link" all mean a preview site — use this tool for those requests. When a local site is already selected and the user asks to create a temporary/temporal/preview site, create a preview of the selected site rather than a new local site. Requires WordPress.com authentication. This can take a few minutes, so tell the user to wait after starting it.',
 	{
```
katinthehatsite, line 10: I think it is a good idea to remove the WordPress.com reference from there as this can be quite confusing for the agents. I am wondering if we should mention where the preview sites are hosted or it might not be necessary. What do you think?

katinthehatsite: Although, on a second thought, it might not be necessary for an agent to have that information since it does not really impact user's workflow

katinthehatsite: Additionally, the synonym disambiguation ("temporary site",  "temporal site", "share link") is repeated in both the tool description and the system prompt. Would it make sense to keep the tool description short and defer to the system prompt for synonym mapping?

apps/cli/ai/tests/tools.test.ts
```diff
@@ -790,14 +796,79 @@ describe( 'Studio AI MCP tools', () => {
...
 		const result = await getTool( 'preview_list' ).rawHandler( { nameOrPath: 'My Site' } as never );
 
 		expect( runListPreviewCommand ).toHaveBeenCalledWith( '/sites/my-site', 'json' );
-		expect( getTextContent( result ) ).toBe( '[{"url":"https://demo.wordpress.com"}]' );
+		// No snapshots emitted by the command -> the tool reports an empty list.
+		expect( JSON.parse( getTextContent( result ) ?? 'null' ) ).toEqual( [] );
+	} );
+
+	it( 'enrichPreviewListOutput tags each preview with type "preview" and an expiry flag', () => {
+		const futureDate = 1900000000000;
```
katinthehatsite: Should we do something like `Date.now() + some_large_offset` to avoid having a hard-coded date? I know it is far in the future but who knows 😅

sejas: Thanks for the suggestion. I made the change here https://github.com/Automattic/studio/pull/3677/changes/591adc90c4130eef06548c17457d6370498598c4

apps/cli/ai/tools/create-preview.ts
sejas: I preferred to remove the duplication from the system prompt in favor of the tool description.
https://github.com/Automattic/studio/pull/3677/commits/1e8543a7122998ad5639b38026509f2351870b20

sejas: I confirmed it is still working:
Screenshot 2026-06-25 at 13 39 41

nightnei (approved):
LGTM and works well 👍
Screenshot 2026-06-25 at 13 25 21
Screenshot 2026-06-25 at 13 26 54
Screenshot 2026-06-25 at 13 28 18

Only one thing - should we add mentioning Pressbale (as `Wordpress.com/Pressable`), for example, here `removed site_connected_remote_sites` and in other places where we mention wordpress.com? During my testing, the AI didn't mention Pressable, however, on the screenshots in the "Testing instructions", you had mentioned Pressable, that's why I started wondering - should we mention it directly everywhere, so that the AI always mentions it too in responses?

sejas:
Thank you both for reviewing the PR, and your suggestions.

> During my testing, the AI didn't mention Pressable, however, on the screenshots in the "Testing instructions", you had mentioned Pressable, that's why I started wondering - should we mention it directly everywhere, so that the AI always mentions it too in responses?

That's a good point. I'll create an issue to update skills and docs to take Pressable into consideration.

PR #3682, opened 2026-06-03
Fix screenshot eval assertion to match the tool's JPEG output

lezama:
## Related issues

None.

## How AI was used in this PR

The code change and this PR description were drafted with Claude Code; the diagnosis was confirmed by running the eval suite locally.

## Proposed Changes

The `screenshot-all-timing` agent eval has been failing because its assertion still expected PNG screenshots, while `take_screenshot` now produces JPEG (`screenshot-desktop.jpg` / `screenshot-mobile.jpg`, `image/jpeg`). The agent behavior was correct all along — one fast `viewport: "all"` capture of desktop and mobile — but the check rejected the payload on the file name/format and parsed dimensions as PNG.

This updates the assertion to validate the real invariant rather than the encoding:
- Keys the expected width by viewport (`desktop` → 1040, `mobile` → 390) instead of the file extension, so it no longer breaks if the format changes again.
- Reads pixel dimensions from either a PNG or a JPEG buffer.

The eval suite is green again, so the daily eval reports an accurate signal instead of a standing false failure.

## Testing Instructions

1. From `eval/`, run the screenshot case:
   `npx promptfoo@0.121.4 eval -c promptfoo.config.yaml --no-cache --filter-pattern "screenshot verification"`
2. Confirm it passes (it captures one `viewport: "all"` screenshot and validates the desktop/mobile JPEG dimensions).

## Pre-merge Checklist

- [x] Have you checked for similar PRs that may already address this?
- [x] Verified locally (the case passes).
- [ ] Reviewed by a teammate.

### DIFF
--- a/eval/promptfoo.config.yaml
+++ b/eval/promptfoo.config.yaml
@@ -138,29 +138,53 @@ tests:
               return { pass: false, score: 0, reason: `expected two screenshot media payloads, got ${JSON.stringify(payloads).slice(0, 500)}` };
             }
 
+            // take_screenshot emits JPEG; key the expected width by viewport so
+            // the check survives a future format change rather than the extension.
             const expectedWidths = new Map([
-              ['screenshot-desktop.png', 1040],
-              ['screenshot-mobile.png', 390],
+              ['desktop', 1040],
+              ['mobile', 390],
             ]);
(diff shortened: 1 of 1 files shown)


youknowriad (approved):

sejas (commented):
Thanks for updating the eval 🙌

PR #3684, opened 2026-06-03
Clean up speculative Sentry logging

fredrikekelund:
## Related issues

- Fixes STU-1478

## How AI was used in this PR

N/A

## Proposed Changes

In https://github.com/Automattic/studio/pull/2793, we implemented a data migration from the old `appdata-v1.json` config file to the new decoupled file strategy (living in `~/.studio`). During that data migration, we added Sentry logging to help us catch certain types of errors. 

Per pfHvTO-1ct-p2, we want to avoid leaving this kind of "speculative logging" in the code for longer periods of time, so this PR removes the `Sentry.captureException` calls. These errors would only have occurred if the `appdata-v1.json` file didn't conform to the expected schema. Luckily, we didn't catch any such errors.

## Testing Instructions

Code review is sufficient

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/migrations/02-migrate-to-split-config.ts
+++ b/apps/studio/src/migrations/02-migrate-to-split-config.ts
@@ -13,7 +13,6 @@
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
-import * as Sentry from '@sentry/electron/main';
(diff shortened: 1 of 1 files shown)


wojtekn (approved):
Looks good, thanks for the claenup.

PR #3685, opened 2026-06-03
Data migration to delete `appdata-v1.deprecated.json` and related cruft

fredrikekelund:
## Related issues

- Fixes STU-1494

## How AI was used in this PR

Claude did the implementation over two passes, using a detailed spec from me.

## Proposed Changes

In https://github.com/Automattic/studio/pull/2793 and related PRs, we copied `appdata-v1.json` into three decoupled config files in `~/.studio`. We also copied the `server-files` and `certificates` directories to `~/.studio`.

The migration went off smoothly. This PR adds a new migration that removes the original paths from the user's system.

## Testing Instructions

1. Run `open ~/Library/Application\ Support/Studio` in a terminal.
2. Ensure that you have either an `appdata-v1.deprecated.json` file, a `server-files` directory, or a `certificates` directory. If not, create an empty `server-files` directory.
3. Run `npm start`.
4. Ensure that `appdata-v1.deprecated.json`, `server-files`, and `certificates` are all removed.

## Pre-merge Checklist


bcotrim (commented):
`appdata-v1.json` is only renamed to `appdata-v1.deprecated.json` after the split-config migration completes. If that migration fails before the rename, this cleanup would delete the raw legacy config on the next launch, potentially removing the only recoverable source data.

Should we only delete `appdata-v1.deprecated.json`?

fredrikekelund:
@bcotrim, good call 👍 Thanks for staying alert. I changed the migration condition to only run if `appdata-v1.deprecated.json` exists, as that's our signal that `02-migrate-to-split-config.ts` was successful

PR #3683, opened 2026-06-03
Studio Code Desktop: remove enableStudioCodeUi beta flag and legacy assistant

sejas:
## Related issues

- Fixes STU-1593

## How AI was used in this PR

Used Claude Code to create the code and PR.

## Proposed Changes

The new "Studio Code" desktop assistant has been gated behind the `enableStudioCodeUi` beta flag while the previous WordPress.com chat assistant was the default. This PR ships the new assistant to everyone by removing that flag.

- The assistant tab now always shows the Studio Code experience — there is no longer a "Studio Code Desktop" toggle under Beta Features (Remote Session and Native PHP runtime remain).
- The legacy WPCOM chat assistant and everything only it used is removed, so there is no dead code path or duplicate assistant to maintain.
- No user-visible change for people already on the new assistant. Users still on the old one are moved to the new assistant.

## Testing Instructions

1. `npm start`, open a site, go to the **Assistant** tab.
2. Confirm the Studio Code assistant renders and the old gray panel background is gone.
3. Open **WordPress Studio → Beta Features**: confirm there is no "Studio Code Desktop" entry; Remote Session and Native PHP runtime still listed.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #3686, opened 2026-06-03
Add studio/no-redundant-cx ESLint rule for static class strings

sejas:
## Related issues

- Related to https://github.com/Automattic/studio/pull/3683#pullrequestreview-4417896032

## How AI was used in this PR

Used Claude Code to author the ESLint rule and tests, find and auto-fix existing redundant `cx()` calls across the codebase, and verify the rule end-to-end (RuleTester + real eslint run).

## Proposed Changes

`cx()` exists to join class names conditionally, but it was sometimes used to wrap a single static string — which adds indirection without any benefit. This PR makes that an enforced convention so the codebase stays consistent and the helper signals "there is a condition here".

- Adds a `studio/no-redundant-cx` ESLint rule that flags `cx( 'static string' )` (single static string or non-interpolated template) and auto-fixes it to the bare string. Conditional and multi-argument `cx()` calls are left untouched.
- Documents the convention in `AGENTS.md` so contributors and AI agents follow it.
- Applies the rule across the existing code, removing the now-redundant `cx()` wrappers it found.
- Wires the `eslint-plugin-studio` package into the test runner so its rule tests (this rule plus the pre-existing `require-lock-before-save`) actually run.

No user-facing behavior change — this is an internal code-quality/lint convention.

## Testing Instructions

1. `npm test` — `eslint-plugin-studio` suite passes (rule valid/invalid + autofix cases).
2. `npx eslint .` — no `studio/no-redundant-cx` violations remain.
3. Add a `cx( 'foo' )` call somewhere and run `npx eslint --fix <file>` — it rewrites to `'foo'`; `cx( 'foo', cond && 'bar' )` is left alone.

## Pre-merge Checklist


youknowriad (approved):
Cool, I like the proactiveness

fredrikekelund:
Side note: I'd love to bring in [clsx](https://www.npmjs.com/package/clsx). The cx function doesn't do much for readability, IMO.

sejas:
I don't have a strong opinion as long as we use only one solution. One argument in favor of `cx` is that it reduces  the number of dependencies. clsx has 82M weekly downloads and the last release was 2 years ago. We could pin it to a specific version.

PR #3688, opened 2026-06-03
Enforce lazy translations: add no-module-level-translations ESLint rule

sejas:
## Related issues

- Related to STU-1771

## How AI was used in this PR

AI assistance was used to author the new ESLint rule and to apply the repetitive "wrap the translation in a lazy getter" refactor across the affected files. Every change was reviewed by hand and verified with `npm run lint`, `npm run typecheck`, and `npm test` (1961 tests passing).

## Proposed Changes

Translatable strings are sometimes evaluated at module-load time (e.g. `const LABEL = __( 'Label' )`). In the long-lived desktop renderer the user can change the app language at runtime, but a string captured at import time stays frozen in the language that was active when the module first loaded — so parts of the UI silently fail to translate after a language switch.

This adds a lint rule that catches the mistake at author time and makes the existing offenders lazy so they react to language changes:

- New `studio/no-module-level-translations` ESLint rule flags `__()`, `_x()`, `_n()`, and `_nx()` calls that run at module scope. It allows translations that are wrapped in a function (evaluated at render/call time) and ignores bare discarded statements used only to feed the translation extractor. The one-shot CLI loads its locale before importing modules, so the rule is turned off for `apps/cli`.
- Documents the convention in `AGENTS.md`.
- Makes all flagged strings lazy (widget/control labels, settings, skills constants, shared `tools/common` helpers, etc.), so they now follow the active language after a runtime switch.

No user-visible text changes; the user-facing effect is that affected labels now translate correctly when the language is changed without a restart.

## Testing Instructions

- `npm run lint` — passes; try adding `const X = __( 'Test' )` at module scope in a renderer file and confirm the rule reports it, then wrap it as `const getX = () => __( 'Test' )` and confirm it passes.
- `npm run typecheck` — passes.
- `npm test` — passes (includes the rule's unit tests under `tools/eslint-plugin-studio`).
- Manually: switch the app language in Settings and confirm the affected labels (widget controls, settings options) update without restarting.

## Pre-merge Checklist


sejas:
I think the direction is good. I'll create a fresh PR to cover new code and avoid resolving conflicts.

PR #3690, opened 2026-06-03
Make reprint.phar a dynamically downloaded dependency

epeicher:
## Related issues

- Related to STU-1812

## How AI was used in this PR

Implemented with Claude Code (Opus): updated the build-time download script, the runtime path resolution, and the tests, and removed the earlier runtime-download module. I reviewed the changes.

## Proposed Changes

`reprint.phar` (the PHP archive behind the experimental `pull-reprint` migration) now ships as a fixed, pinned version bundled into every Studio build. A build-time step downloads it into `wp-files/` using the same mechanism as WordPress, wp-cli, and the SQLite plugin (`scripts/download-wp-server-files.ts`), so it is part of the installer instead of being fetched or auto-updated at runtime.

The version is pinned to `v0.8.1`. Bumping it is a one-line change to a constant in the download script.

User impact:

- Every Studio build embeds a known, fixed reprint version (reproducible, no surprise upstream changes).
- No runtime network dependency for `pull-reprint` to obtain the phar.
- The binary is not committed to the repo; it is downloaded at build / postinstall time and is gitignored.

This replaces the earlier runtime auto-download approach, which has been removed.

## Testing Instructions

1. Apply these changes and run `npm ci`. 
2. Confirm `wp-files/reprint/reprint.phar` is downloaded from the `v0.8.1` release. 
CleanShot 2026-06-09 at 16 26 37@2x

3. Build the CLI: `npm run cli:build`. Confirm `apps/cli/dist/cli/wp-files/reprint/reprint.phar` exists and `apps/cli/dist/cli/reprint.phar` does not.
4. Additionally, you can build the application and check that the `reprint.phar` is there by running `npm run make` and then checking `ls -l apps/studio/out/Studio-darwin-arm64/Studio.app/Contents/Resources/cli/wp-files/reprint`

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? (apps/cli and the build script typecheck cleanly; no application console errors introduced.)


apps/cli/lib/dependency-management/reprint.ts
```diff
@@ -0,0 +1,182 @@
...
+import { downloadFile } from '@studio/common/lib/download-file';
+import { lockFileAsync, unlockFileAsync } from '@studio/common/lib/lockfile';
+import { getReprintPharPath } from './paths';
+
+// reprint.phar is published as a release asset on the upstream reprint repository. Instead of
+// bundling a snapshot with the CLI, we download the latest release into the writable cache and
+// refresh it on the shared dependency-update throttle (see `updateServerFiles` in `setup.ts`).
+const REPRINT_REPO = 'adamziel/reprint'; // TODO: Review this if the repo is moved to the Automattic repository
+const REPRINT_ASSET = 'reprint.phar';
+const LATEST_RELEASE_URL = `https://api.github.com/repos/${ REPRINT_REPO }/releases/latest`;
```
wojtekn: Should we use the tagged version link to ensure we intentionally use a specific version?

fredrikekelund:
Echoing @wojtekn's point above and relaying what @epeicher and I discussed in a huddle: we agreed to change this implementation so that every version of Studio ships a fixed version of reprint.

There should be a script that automatically extracts the PHAR from the latest reprint release and puts it in the `wp-files` directory, but this should happen at build time. 

So, we should look to `scripts/download-wp-server-files.ts` more than `apps/cli/lib/dependency-management`. My bad for writing the issue description in a way that made it seem like this auto-update-at-runtime approach was preferable 🙇‍♂️

wojtekn:
> There should be a script that automatically extracts the PHAR from the latest reprint release and puts it in the wp-files directory, but this should happen at build time. 

The download part should be straightforward, e.g.:  https://github.com/adamziel/reprint/releases/download/v0.8.1/reprint.phar

epeicher:
Thanks @wojtekn and @fredrikekelund for your comments. I have addressed them, so now `reprint.phar` is downloaded using a specific version during build time. This is ready for another review. The Unit Tests failure is unrelated and it's being fixed as part of https://github.com/Automattic/studio/pull/3740

scripts/download-wp-server-files.ts
```diff
@@ -77,6 +77,12 @@ type FileToDownload = {
 	destinationPath?: string;
 };
 
+// reprint.phar ships bundled with every Studio build — downloaded here at build time, not at
+// runtime. Pin a specific release so each build embeds a known, fixed version; bump this constant
+// to upgrade. See https://github.com/adamziel/reprint/releases.
```
wojtekn: The code below is self-explanatory, not sure if we need the comment.

wojtekn (approved):
The code change looks good, and works correctly:

```
[reprint] Downloading reprint.phar (v0.8.1) ...
[reprint] Moving reprint.phar to destination ...
[reprint] Files extracted
```

scripts/download-wp-server-files.ts
epeicher: Agreed, removed as part of [6ef73f0](https://github.com/Automattic/studio/pull/3690/commits/6ef73f081cfbfb724a0b5a0c49db4f85ea4d5113)

PR #3691, opened 2026-06-03
Fix stopping sites while Studio quits

bcotrim:
## Related issues

- Fixes STU-1778

## How AI was used in this PR

Used Codex to trace the quit flow, review recent related PRs (#2314, #3350, #3608, #2426, #2683), implement the fix, and add regression tests. I reviewed the final diff and ran the verification commands below.

## Proposed Changes

- Makes the **Stop sites** quit-dialog action reliably stop running sites before Studio exits.
- Keeps the existing safer quit behavior intact: **Leave running** remains the default, and **Cancel** keeps Studio open without entering shutdown cleanup.
- Preserves the ghost-window protection from the recent quit-flow fix by only marking Studio as quitting once the user has committed to quitting.

## Testing Instructions

- Start one or more sites, quit Studio, choose **Stop sites**, and verify Studio exits after the sites stop.
- Start one or more sites, quit Studio, press Enter or choose **Leave running**, and verify Studio exits while the sites keep running.
- Start one or more sites, quit Studio, choose **Cancel**, and verify Studio stays open.
- With sites still starting, quit Studio and choose **Stop sites**; verify Studio quits without reopening ghost windows.

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/index.ts
+++ b/apps/studio/src/index.ts
@@ -429,8 +429,45 @@ async function appBoot() {
 	 */
 	let shouldStopSitesOnQuit = true;
 	let isQuittingConfirmed = false;
+	let isStoppingSitesBeforeQuit = false;
+	let hasCleanedUpBeforeQuit = false;
+
+	function cleanupBeforeQuit() {
+		if ( hasCleanedUpBeforeQuit ) {
+			return;
+		}
+
+		hasCleanedUpBeforeQuit = true;
+		markAppQuitting();
+		globalShortcut.unregisterAll();
+		stopCliEventsSubscriber();
+		stopRemoteSessionStatusPolling?.();
+		stopRemoteSessionStatusPolling = undefined;
+	}
+
+	function stopSitesAndExit() {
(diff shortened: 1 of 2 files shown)


fredrikekelund (changes requested):
I cannot reproduce the problem described in STU-1778. It's also not clear to me what the theory is here for why this isn't working.

Side note: I verified this behavior when testing https://github.com/Automattic/studio/pull/3608. I doubt it's related to that PR.

nightnei:
> I cannot reproduce the problem described in [STU-1778](https://linear.app/a8c/issue/STU-1778). It's also not clear to me what the theory is here for why this isn't working.
> 
> Side note: I verified this behavior when testing #3608. I doubt it's related to that PR.

I reproduced it many times even before, and at some moment I realized that I always clicking "Stop sites" w/o any effect.
Recorded the video and added to the STU-1778

fredrikekelund:
Closing this since we confirmed that the behavior already matches what we want. See the Linear issue

PR #3695, opened 2026-06-04
Distribute: upload Linux .deb as a Full Install build too

gavande1:
Related to STU-1782

## How AI was used in this PR

* Used Claude Code to trace the 404 from the marketing-theme link through the WPCOM CDN download route to the buildkit `distribute_builds` lane, confirm against the `studio-app/updates` endpoint that the `Update` entry is required for the in-app updater, and draft the fix. All changes were reviewed before submitting.

## Proposed changes

* Register the Linux `.deb` on the Apps CDN as `install_type: Full Install` in addition to the existing `Update`, so the public download URL resolves.

## Why are these changes being made?

* `/downloads/wordpress-com-studio/linux-x64/latest` (and `linux-arm64`) returns **404**. The `.deb` is only uploaded as `install_type: Update` (which backs the in-app updater); the public download route defaults to `full-install`, for which no Linux build exists. Mac and Windows already upload both types — Linux was missing the Full Install one.
* Fixes future builds only — v1.10.0 needs a manual backfill.

## Testing instructions

TBD


gavande1:
Closing — no pipeline change needed. The Apps CDN already serves the Linux .deb via the existing keyword path `/downloads/wordpress-com-studio/linux-x64/latest/update` (and `linux-arm64/latest/update`), which 302s to the current build. Consumers should link to `…/latest/update` rather than `…/latest` (full-install). Slightly non-standard but fine, and it avoids uploading the same artifact twice. See STU-1782.

PR #3696, opened 2026-06-04
Default wpcom provider for Studio Code in Studio UI

nightnei:
## Related issues

- Fixes STU-1777

## How AI was used in this PR
I figured out and wrote the code manually. AI just double checked the approach. Since AI was generating overengineered solutions.

## Proposed Changes
In Studio UI we decided to default to wpcom provider, since it's free right now and there is no sense for users to use own keys.

## Testing Instructions
1. `npm run cli:build`
2. `node apps/cli/dist/cli/main.mjs code`
3. Select wpcom
4. Assert that no regresion, it work well as before
5. `npm start`
6. Open Studio Code tab
7. Open `~/.studio/cli.json`
8. Remove `aiProvider`
9. Say `hi` to Assistant
11. Assert that you imiidiatelly see the response from AI. And assert that `aiProvider` has value `wpcom` in `~/.studio/cli.json`

### DIFF
--- a/apps/cli/commands/ai/index.ts
+++ b/apps/cli/commands/ai/index.ts
@@ -315,6 +315,12 @@ export async function runCommand( options: {
 	const config = await readCliConfig();
 	let showCapabilitiesOnConnect = ! config.aiProvider;
 
+	// Since Studio Code inside Studio UI is free right now - we decided to default to WordPress.com provider so far.
+	if ( isJsonMode && showCapabilitiesOnConnect ) {
+		await switchProvider( 'wpcom', false );
+		showCapabilitiesOnConnect = false;
+	}
+
 	if ( showCapabilitiesOnConnect ) {
 		ui.showOnboarding();
 
--- a/apps/cli/commands/ai/index.ts
+++ b/apps/cli/commands/ai/index.ts
@@ -315,7 +315,7 @@ export async function runCommand( options: {
 	const config = await readCliConfig();
 	let showCapabilitiesOnConnect = ! config.aiProvider;
 
-	// Since Studio Code inside Studio UI is free right now - we decided to default to WordPress.com provider so far.
+	// Studio Code Desktop defaults to WordPress.com provider.
 	if ( isJsonMode && showCapabilitiesOnConnect ) {
 		await switchProvider( 'wpcom', false );
 		showCapabilitiesOnConnect = false;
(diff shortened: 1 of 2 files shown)


youknowriad:
Can we instead disable the "api key" provider entirely from Studio App and rely full on the .com provider? Alternative would be to surface the API KEY config in the app settings but I'm not sure it's necessary for a start. I think there's not much value for the moment to provide your own API KEY specially if we're still giving free access.

nightnei:
> Can we instead disable the "api key" provider entirely from Studio App and rely full on the .com provider? Alternative would be to surface the API KEY config in the app settings but I'm not sure it's necessary for a start. I think there's not much value for the moment to provide your own API KEY specially if we're still giving free access.

@youknowriad do you mean to:
1. Remove this:<br />Screenshot 2026-06-04 at 21 07 55

3. If the user opens the Assistant tab and if `aiProvider` is not set yet, then set it to `wpcom` automatically in `cli.json`

Do I understand your suggestion correctly?

I think it doesn't make sense to show only 1 open to chose, like this:<br />Screenshot 2026-06-04 at 21 12 39

sejas (commented):
Vova, thanks for working to add Anthropic API. I tested it and confirmed I was able to add my Anthropic API key and worked as expected.

In the testing instructions there is a small typo with the Studio home folder. The folder is hidden and requires a dot `Open ~/.studio/cli.json`.

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/ac6c5699-0037-43b9-9b8f-5d2e06b6c82e"></video> | <video src="https://github.com/user-attachments/assets/6b563def-d016-480b-9652-c0f77c46380f"></video> |

I agree with Riad and we can default to WPcom provider without asking the user, at least initially.

youknowriad:
Yes, that was my suggestion, I'm hesitant about proposing the removal of the anthropic provider entirely from the the code command as well. (CLI). I think it's still useful for me to check whether an issue is due to our proxy or not, but other than that, I don't see it being useful to users right now.

nightnei:
@sejas @youknowriad done

apps/cli/commands/ai/index.ts
```diff
@@ -315,6 +315,12 @@ export async function runCommand( options: {
 	const config = await readCliConfig();
 	let showCapabilitiesOnConnect = ! config.aiProvider;
 
+	// Since Studio Code inside Studio UI is free right now - we decided to default to WordPress.com provider so far.
```
sejas:
```suggestion
	// Studio Code Desktop defaults to WordPress.com provider.
```

sejas (approved):
@nightnei, thanks for the change. I tested it and works as expected. Let's add or update existing unit tests, so we make sure we don't introduce regressions in the future.
I tested it by renaming my ~/.studio I went through the onboarding , logged in and tested the Studio Assistant/Code  tab. I confirm it responded correctly instead of asking me to select the provider.

Image


Also noting that we don't need to test if it has the token ready, because the UI limits the chat interface to authenticated users.

PR #3698, opened 2026-06-04
Adjust assistant limits in settings

nightnei:
## Related issues

- Fixes STU-1779

## How AI was used in this PR
AI helped to write the code, I checked everything thoroughly.

## Proposed Changes
The new Studio Code Desktop UI doesn't have any limits.
So temporarily, we are updating the limits in Settings to say: Unlimited tokens while Studio Code is in beta.
<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2026-06-04 at 15 39 18
<td>Screenshot 2026-06-04 at 15 38 20
</tr>
</table>

## Testing Instructions
1. `npm start`
2. Open Setting
3. Assert that you see<br />Screenshot 2026-06-04 at 15 38 20


apps/studio/src/stores/wpcom-api.ts
```diff
@@ -119,27 +98,11 @@ function parseResponse< TSchema extends z.ZodType >(
...
-
 export const wpcomApi = createApi( {
 	reducerPath: 'wpcomApi',
 	baseQuery: wpcomBaseQuery,
-	tagTypes: [ 'AssistantQuota', 'SnapshotUsage' ],
+	tagTypes: [ 'SnapshotUsage' ],
 	endpoints: ( builder ) => ( {
-		getAssistantQuota: builder.query< z.infer< typeof assistantQuotaSchema >, void >( {
-			query: () => ( {
-				path: '/studio-app/ai-assistant/quota',
```
nightnei, line 136: @sejas WDYT, should we remove this endpoint from wpcom? I would propose to keep it. And later, when we implement limits for Studio Code UI - we will either adjust or remove this endpoint. Are you ok with it or prefer another way?

sejas: We need to keep the endpoint for a while, as users on older stable releases and previous versions will still use it. We can decide later if it makes sense using same or different endpoint when we implement the limits in Studio Code.

sejas (approved):
Thanks for catching and implementing this change. It works as described. We could consider keeping the progress bar but is not a strong opinion.

Image

epeicher:
FYI, the error in the E2E tests is being fixed as part of https://github.com/Automattic/studio/pull/3699, the `min-release-age=2` was not applied to the studio app

PR #3713, opened 2026-06-05
Fix CLI symlink not updated after Studio app update

wojtekn:
## Related issues

- Fixes STU-1787

## How AI was used in this PR

This PR was developed with Claude Code assistance.

## Proposed Changes

After Studio is updated, the `studio` CLI command could continue running the binary from a previous app bundle. The root cause was that `autoInstallIfNeeded` used a single `cliAutoInstalled` flag to mean both "we've done first-time install" and "user deliberately uninstalled — don't reinstall". Once set, it prevented `installCli()` from ever running again, so a stale symlink would never get fixed.

This PR introduces a separate `cliUserUninstalled` flag that is explicitly set when the user uninstalls the CLI via Settings, and cleared when they reinstall it. `autoInstallIfNeeded` now skips only when `cliUserUninstalled` is true, and otherwise calls `installCli()` on every launch. Since `installCli()` already no-ops when the symlink is correct, this is safe and cheap — it only re-points the symlink when it's actually stale.

## Testing Instructions

- Install Studio, complete CLI installation
- Uninstall CLI via Settings — confirm `studio` command is gone and re-launching Studio does not reinstall it
- Reinstall CLI via Settings — confirm `studio` command works and re-launching Studio keeps it pointing to the correct app

## Pre-merge Checklist


wojtekn:
It seems it works:

1. Remove symlink `~/.local/bin/studio`
2. Build Studio, open from `out/`, enable CLI in settings, close Studio
3.  Confirm the symlink was created and points to location in `out/`

```
% readlink ~/.local/bin/studio
/Users/USER/Sites/studio/apps/studio/out/make/zip/darwin/arm64/Studio.app/Contents/Resources/bin/studio-cli.sh
```

4. Close Studio, move/copy it to `/Applications/`
5. Open Studio from `/Applications/`
6. Confirm the symlink resolves to the default location now:

```
% readlink ~/.local/bin/studio
/Applications/Studio.app/Contents/Resources/bin/studio-cli.sh
```

I thought I may be fixing an edge case that could happen only during development and testing local build, but it seems it can happen for the user, too:

1. Download Studio, open DMG, drag and drop to the Downloads directory
2. Open Studio, enable CLI, save
3. Move Studio to `/Applications/`
5. Symlink now points to a non-existent file and the `studio` command fails

bcotrim (commented):
Works as described.

I found one behavior change that may regress users who explicitly uninstalled the CLI before `cliUserUninstalled` existed.

With this state:
- `cliAutoInstalled: true`
- no `cliUserUninstalled` key
- `~/.local/bin/studio` absent

After launching this branch, Studio recreated `~/.local/bin/studio`.

Was that intentional? This state is ambiguous, but it also matches the old “user uninstalled via Settings” state. The stale-bundle case should still have an existing symlink pointing to an old app path, so I wonder if absent symlink + legacy config should keep the CLI uninstalled.

wojtekn:
@bcotrim thanks for testing it. I added handling for that case.

wojtekn:
I tested the main flow again, and it works fine:

```
% readlink ~/.local/bin/studio
/Users/USER/Sites/studio/apps/studio/out/Studio-darwin-arm64/Studio.app/Contents/Resources/bin/studio-cli.sh
% readlink ~/.local/bin/studio
/Applications/Studio.app/Contents/Resources/bin/studio-cli.sh
```

fredrikekelund:
It took me a second to wrap my head around this, but if I understand it correctly, it's really meant to fix an issue where if you run a local dev build and install it, it might set up a weird symlink at `~/.local/bin` compared to a production install. 

This is kind of an edge case, but I don't have any strong objections to switching the paradigm to running an install on every launch unless we know the user explicitly uninstalled the CLI. However, I think we should consider removing `userData.cliAutoInstalled` entirely. 

It looks like there's some data migration logic in there (and maybe that's the only real reason we kept that prop). If that's the case, I'd ask that we write an explicit data migration in `apps/studio/src/migrations` instead

wojtekn:
@fredrikekelund it could also happen for standard user:

1. Download Studio, open DMG, drag and drop to the Downloads directory
2. Open Studio, enable CLI, save
3. Move Studio to /Applications/
4. Symlink now points to a non-existent file and the studio command fails

fredrikekelund:
Right. I'd appreciate it if you could revisit the `userData.cliAutoInstalled` prop, though. I could be mistaken, but it doesn't appear to be needed anymore, and the CLI installation with the symlink chain is complex already.

wojtekn:
> Right. I'd appreciate it if you could revisit the `userData.cliAutoInstalled` prop, though. I could be mistaken, but it doesn't appear to be needed anymore, and the CLI installation with the symlink chain is complex already.

Good catch. We don't need to write it anymore, but we should support it to ensure we don't have to reinstall the CLI for users who uninstalled it in the past. I opened https://github.com/Automattic/studio/pull/3781

PR #3714, opened 2026-06-05
Studio Code: add read-before-edit hint to scaffold_theme result

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Investigated a real `studio code` session where an Edit failed, identified the root cause with Claude Code, and implemented + tested this fix. The diagnosis came from reading the session transcript (the agent edited a freshly-scaffolded `style.css` without reading it first).

## Proposed Changes

When the agent uses `scaffold_theme` and then edits one of the generated files (commonly `style.css`), the Edit can fail with `Could not find the exact text …`. The cause: the tool result listed only filenames, so the agent guessed a file's contents — e.g. anchoring an Edit on a minimal `Theme Name` header — which never matches the real scaffolded WordPress header (Description, Version, License, Text Domain, Tags, …).

This adds a short hint to the `scaffold_theme` result telling the agent that the generated files already contain standard WordPress headers/starter content and to read a file before editing it. It nudges the agent toward a read-before-edit flow without bloating the result with file contents.

User impact: fewer failed edits and retries right after scaffolding a theme in Studio Code.

## Testing Instructions

1. `npm run cli:build` (Node 24).
2. Unit test (asserts the hint is present in the tool result):
   `npx vitest run apps/cli/ai/tests/tools.test.ts -t scaffold`
3. End-to-end: `node apps/cli/dist/cli/main.mjs code "Scaffold a block theme called Demo Theme in <site>, then add a Text Domain comment to its style.css"` and confirm the agent reads `style.css` before editing (no `Could not find the exact text` error).

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/tests/tools.test.ts
+++ b/apps/cli/ai/tests/tools.test.ts
@@ -1075,6 +1075,15 @@ describe( 'Studio AI MCP tools', () => {
 			expect( getTextContent( result ) ).toContain( 'wp theme activate acme-studio' );
 		} );
 
+		it( 'includes a read-before-edit hint so the agent does not guess file contents', async () => {
+			const result = await getTool( 'scaffold_theme' ).rawHandler( {
+				nameOrPath: scaffoldSite.name,
+				name: 'Acme Studio',
+			} as never );
+
+			expect( getTextContent( result ) ).toContain( 'Read a file before editing it' );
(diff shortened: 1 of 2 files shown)


sejas (approved):
Code LGTM, but I haven’t fully tested it.

PR #3716, opened 2026-06-05
Add visual-polish skill + inspect_design tool to fix unpolished generated designs

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Built with Claude Code (agent prompt + skill engineering). The new `inspect_design` tool, the `visual-polish` skill, and the prompt/skill wiring were authored and iterated with AI; the design of the detection loop and the button/layout root-cause guidance came out of real test sessions on generated sites. Code, tests, and skill copy were human-reviewed.

## Proposed Changes

Sites built by the Studio Code agent frequently come out looking "unpolished" — sections not centered, full-width sections rendering in the narrow content column, buttons with doubled padding and conflicting hover styles, off spacing and column backgrounds. The root cause is that the **rendered DOM differs from the markup the agent intended**: WordPress injects its own layout classes, `validate_and_fix_blocks` re-serializes block markup, and the button block is two nested elements. The existing screenshot-and-fix step let the agent diagnose from pixels alone and guess at causes — so it kept "fixing" the wrong thing.

This gives the agent the ability to see *why* a page looks wrong, and a disciplined workflow for fixing it:

- **`inspect_design` tool** — inspects the rendered DOM for given CSS selectors and returns bounding boxes, curated computed styles, the WordPress block-gap custom property, the ancestor layout-class chain, and optional `:hover` styles. The screenshot is the symptom; this is the cause.
- **`visual-polish` skill** — a diagnose-the-whole-page-first, fix-as-a-batch, verify workflow (capped iterations). It pairs the screenshot with `inspect_design` and carries a checklist of the recurring issues (section width/centering, doubled/mis-targeted button styling, block-gap spacing, column/grid backgrounds, hover), each mapped to exactly what to inspect.
- **Authoring fix** — a `block-content` rule to style buttons via `.wp-element-button` (the inner element WordPress styles) so a custom class on the `.wp-block-button` wrapper doesn't stack padding and double the button's size.
- Wiring: workflow step 6 and skill routing now hand polish work to `visual-polish`; `visual-design` is scoped to creating the design direction. Removed now-redundant verification snippets that duplicated the new skill.

Opening as a **draft / Proof of Concept**: the detection loop still needs more real-site validation (recent tests show the agent sometimes skips the diagnosis gate), so this is intended as a directional starting point for review rather than a finished change.

## Testing Instructions

1. `npm run cli:build`
2. Have the Studio Code agent build or redesign a site, then let it run the polish pass (it loads `visual-polish` at step 6).
3. Verify it calls `inspect_design` to diagnose before editing, and that buttons, full-width/centered sections, spacing, and hover render as intended.
4. Unit tests: `npm test -- apps/cli/ai/tests/tools.test.ts apps/cli/ai/tests/system-prompt.test.ts` (includes new `inspect_design` coverage).

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/skills/block-content/SKILL.md
+++ b/apps/cli/ai/skills/block-content/SKILL.md
@@ -16,6 +16,7 @@ Use this skill before writing or editing page content, post content, templates, 
 - For forms or features core blocks do not cleanly provide, load the `plugin-recommendations` skill and use editable plugin blocks.
 - No decorative HTML comments such as `<!-- Hero Section -->` or `<!-- Features -->`. Only WordPress block delimiter comments are allowed.
 - No custom class names on inner DOM elements. Put custom classes only on the outermost block wrapper via the block `className` attribute.
+- Style buttons via `.wp-element-button` — the inner element WordPress applies the button's padding, background, and border to (shared by the button block and buttons from other blocks). A custom class on a button block sits on the `.wp-block-button` wrapper, so descend to `.your-class .wp-element-button`; never style the wrapper directly, or its padding stacks on top of the default and the button doubles in size.
 - No inline `style` attributes or block `style` attributes for styling. Use `className` plus the theme's `style.css`.
 - Use `core/spacer` for empty spacing elements, not empty `core/group` blocks.
 - No emojis anywhere in generated content.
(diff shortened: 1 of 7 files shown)


sejas (approved):
I tested it and worked great. It felt faster although I didn't measure the time 😅 .

Here's the test I created. I just gave the site name "Cafe Roque Nublo" and selected one page option. I also repeated same prompt with preview version 1.10.1 and the results are better in this PR.

- Before: https://beautifulfullyd86eee3eae-ndekx-studio.wp.build/ 
- After: https://antoniosejas-uwsnq-studio.wp.build/

| Before | After |
|--------|--------|
|  Image | Image |

🚀

PR #3724, opened 2026-06-08
Fix black screen on dev start: pin vite 8 at root, remove vite-plugin-top-level-await

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

Diagnosed and fixed with Claude Code.

## Proposed Changes

When `electron-vite 6.0.0-beta.1` was introduced (to support vite 8), the root `package.json` was never updated with a vite 8 dependency. Since `electron-vite` is installed at root, Node resolves `vite` from the root `node_modules`, which held vite 7.3.3 (hoisted there by vitest's peer dep). This caused `npm start` to build with vite 7 instead of vite 8, producing a black screen and `Missing field moduleType` errors from rolldown's `builtin:vite-react-refresh-wrapper` plugin.

Two fixes:

1. **Add `vite: ^8.0.16` to root `devDependencies`** — ensures electron-vite resolves vite 8, matching the workspace packages.
2. **Remove `vite-plugin-top-level-await`** — this plugin requires the standalone `rollup` package which no longer exists in vite 8 (it's bundled as rolldown). The plugin was unused anyway (no top-level `await` at module scope in the codebase), and vite 8 supports top-level await natively.

## Testing Instructions

- `npm ci && npm start` — app should load without black screen
- Previously this would produce `Missing field moduleType` / `Cannot find module 'rollup'` errors

## Pre-merge Checklist

### DIFF
--- a/apps/studio/electron.vite.config.ts
+++ b/apps/studio/electron.vite.config.ts
@@ -3,7 +3,6 @@ import { createRequire } from 'module';
 import { defineConfig } from 'electron-vite';
 import { normalizePath } from 'vite';
 import react from '@vitejs/plugin-react';
-import topLevelAwait from 'vite-plugin-top-level-await';
 import { viteStaticCopy } from 'vite-plugin-static-copy';
 import wasm from 'vite-plugin-wasm';
 import { sentryVitePlugin } from '@sentry/vite-plugin';
@@ -80,7 +79,6 @@ export default defineConfig( {
 		},
 		plugins: [
 			react(),
-			topLevelAwait(),
 			wasm(),
 			viteStaticCopy( {
 				targets: [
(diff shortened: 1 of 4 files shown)


sejas (approved):
Trunk failed to start the whole app 😓 , with the following error: `2:22:35 PM [vite] Internal server error: Missing field 'moduleType'`. So I couldn't compare the about window.
This branch started the app correctly and the about window looks correct:

Image

wojtekn:
> Trunk failed to start the whole app 😓 , with the following error: 2:22:35 PM [vite] Internal server error: Missing field 'moduleType'. So I couldn't compare the about window.

@sejas, thanks for testing - this PR fixes the issue with the app starting in dev mode, too.

PR #3726, opened 2026-06-08
Fix Native PHP export failures

bcotrim:
## Related issues

- Fixes STU-1784

## How AI was used in this PR

Used Codex to plan and implement the fix and improvement to the export messages. All changes were reviewed and tested manually.

## Proposed Changes

- Prevent export failures from showing raw object text like `[object Object]`; users now see a useful export/CLI message when one is available, or a clear fallback that points them to Studio Logs.
- Fix Native PHP export metadata collection so push/export no longer fails when WP-CLI plugin/theme output is produced through piped stdout.

## Testing Instructions

- `npm start`
- From the menu, enable PHP beta flag
- Create a new site from an import file or create one with a few plugins/themes
- Try to do a full site export
- Export should be successful

## Pre-merge Checklist


apps/cli/lib/import-export/export/exporters/default-exporter.ts
fredrikekelund, line 1: These changes are redundant, right? No functional change, I mean

tools/common/lib/import-export-events.ts
```diff
@@ -170,7 +188,15 @@ export const exportEventTupleSchema = z.union( [
...
 
+export const exportEventTupleSchema = z.union( [
+	exportNonErrorEventTupleSchema,
+	z.tuple( [ z.literal( ExportEvents.EXPORT_ERROR ), exportErrorPayloadSchema ] ),
+] );
+
 export const exportIpcEventSchema = z.object( { event: exportEventTupleSchema } );
 
-export type ExportEventTuple = z.infer< typeof exportEventTupleSchema >;
+export type ExportIpcEventTuple = z.infer< typeof exportEventTupleSchema >;
```
fredrikekelund: This export is no longer used

apps/cli/commands/export.ts
```diff
@@ -21,7 +21,7 @@ import { StudioArgv } from 'cli/types';
 
 const logger = new Logger< LoggerAction >();
 
-function sendIpcEvent( eventTuple: ExportEventTuple ) {
+function sendIpcEvent( eventTuple: ExportIpcEvent[ 'event' ] ) {
```
fredrikekelund, line 20: This change is just a different way of accessing the same type. I don't have a strong opinion either way, but if we stick with this approach, we should remove the `ExportIpcEventTuple` export from `import-export-events.ts` since it's now unused

tools/common/lib/import-export-events.ts
```diff
@@ -170,7 +188,15 @@ export const exportEventTupleSchema = z.union( [
 	z.tuple( [ z.literal( ExportEvents.CONFIG_EXPORT_COMPLETE ), nullOrUndefined ] ),
 ] );
 
+export const exportEventTupleSchema = z.union( [
+	exportNonErrorEventTupleSchema,
+	z.tuple( [ z.literal( ExportEvents.EXPORT_ERROR ), exportErrorPayloadSchema ] ),
+] );
+
```
fredrikekelund: If we called `createExportErrorPayload` when emitting the `ExportEvents.EXPORT_ERROR` events (from `DefaultExporter`, for example), we could undo a lot of the "wiring changes" that we have here and in `apps/cli/commands/export.ts`. 

I would argue that's preferable. It's more predictable if the event payloads stay the same between when they're emitted and consumed

epeicher (approved):
Thanks @bcotrim! I can confirm that after applying the changes, I no longer see the errors in the export. Changes also LGTM, but I'm happy to have @fredrikekelund review the code changes too.

|Before | After|
|-------|------|
| Image| Image|

fredrikekelund (commented):
The error parsing parts look good! I left some suggestions for how to potentially simplify parts of the implementation, but the fundamental logic looks solid. 

> Fix Native PHP export metadata collection so push/export no longer fails when WP-CLI plugin/theme output is produced through piped stdout.

I don't see how the changes in `default-exporter.ts` make any functional difference? 🤔 It would be good to clarify this part before landing the PR.

apps/cli/lib/import-export/export/exporters/default-exporter.ts
fredrikekelund: Or does the fact that we get the `stderrText` and `stdoutText` promises before the `exitCode` promise actually have an impact? If so, can we fix that in `WpCliResponse` somehow?

bcotrim: It does actually make a difference, these changes are what is actually fixing the export, the remaining parts are just to display better errors to the user.
I will look if it's possible to move the fix to `WpCliResponse`

fredrikekelund (approved):
LGTM 👍 I went down a rabbit hole of trying to understand the underlying problem and couldn't help but push two commits. The only substantial one is 4741298. The benefit of that change in approach is that _both_ the `stdout` stream and `stdoutText` promises (and the stderr equivalents) will work on `WpCliResponse` response now. I.e., we could still pipe or read the `stdout` stream now, even if the `stdoutText` also contains the data we want. 

I still don't think this API is very intuitive, but maybe we can iterate on it in the future.

PR #3707, opened 2026-06-05
Migrate @mariozechner/pi-* packages to @earendil-works/pi-* (0.78.0)

wojtekn:
## Related issues

- Closes https://github.com/Automattic/studio/security/dependabot/196

## How AI was used in this PR

Claude Code identified all affected files and performed the migration.

## Proposed Changes

The `@mariozechner/*` npm scope is deprecated — the upstream author has moved to `@earendil-works/*`. This replaces all four packages (`pi-ai`, `pi-agent-core`, `pi-coding-agent`, `pi-tui`) from `0.70.2` to `0.78.0` across all `package.json` files and import statements.

The version jump includes several fixes relevant to Studio:
- Fixed `createAgentSession()` failing with `ENOENT` when `package.json` is missing next to the bundle entrypoint (directly affects bundled CLI)
- Fixed Anthropic-compatible requests incorrectly picking up `ANTHROPIC_AUTH_TOKEN` for non-Anthropic providers
- Fixed HTTP idle timeout for long-running provider streams
- Fixed `AgentSession` retry/compaction/event settlement

Using `0.78.0` rather than `0.78.1` (latest) because `0.78.1` was published June 4 and falls within the dependency cooldown window.

## Testing Instructions

- Run the Studio CLI AI agent flow and verify it starts a session correctly
- Confirm no TypeScript errors: `npm run typecheck`

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #3727, opened 2026-06-08
Fix Windows build: exclude pi-coding-agent from packaged app

wojtekn:
## Related issues

Follows up on #3707 (migrate `@mariozechner/pi-*` to `@earendil-works/pi-*`).

## How AI was used in this PR

Claude Code diagnosed the root cause and wrote the fix.

## Proposed Changes

After #3707 landed, the Windows Squirrel build started failing with:

```
The specified path, file name, or both are too long. The fully qualified file name
must be less than 260 characters, and the directory name must be less than 248 characters.
```

Root cause: `@earendil-works/pi-coding-agent` is a `devDependency` of `apps/studio` (only needed for `import type { SessionEntry }` — erased at compile time). However, after `install:bundle` runs `npm install --no-workspaces`, galactus (electron-packager's `DestroyerOfModules`) doesn't reliably prune the nested `node_modules` under `pi-coding-agent`. Its transitive `@mistralai/mistralai` dependency contains filenames like `getchatcompletionfieldoptionscountsv1observability...post.js` that reach 200+ chars on their own. With the Windows build agent prefix (`C:\buildkite-agent\builds\ci-windows-...\automattic\studio\`) the total path hits ~295 chars — 35 over the 260-char limit.

The fix adds an explicit `ignore` pattern to `forge.config.ts` so electron-forge never copies `pi-coding-agent` into the packaged output, regardless of whether the devDep prune works. This is a belt-and-suspenders approach since the package should already be excluded as a devDep.

## Testing Instructions

- Windows build should pass (Squirrel maker no longer encounters long paths)
- macOS and Linux builds unaffected (no 260-char path limit)
- App functionality unchanged — `pi-coding-agent` was only imported for TypeScript types, not included at runtime

## Pre-merge Checklist

- [x] I have checked this code is correct to the best of my ability
- [ ] I have tested this change on a supported platform
- [x] My changes generate no new deprecation warnings

### DIFF
--- a/.buildkite/commands/build-for-windows.ps1
+++ b/.buildkite/commands/build-for-windows.ps1
@@ -49,6 +49,15 @@ Write-Host "--- :npm: Installing Node dependencies"
 bash .buildkite/commands/install-node-dependencies.sh
 If ($LastExitCode -ne 0) { Exit $LastExitCode }
 
+# The bundled nuget.exe in electron-winstaller is v4.0 which predates long-path support (requires
+# v4.8+). Replace it with the latest stable release so the LongPathsEnabled registry key above
+# actually takes effect during `npm run make`.
+Write-Host "--- :nuget: Upgrading bundled nuget.exe to latest (long-path support requires >= 4.8)"
+$nugetVendorPath = "node_modules\electron-winstaller\vendor\nuget.exe"
+Invoke-WebRequest -Uri "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe" `
+    -OutFile $nugetVendorPath -UseBasicParsing
+If ($LastExitCode -ne 0) { Exit $LastExitCode }
+
 Write-Host "--- :node: Building App for Windows ($BuildType - $Architecture)"
 
 # Run appropriate script based on build type
--- a/.buildkite/commands/build-for-windows.ps1
+++ b/.buildkite/commands/build-for-windows.ps1
@@ -40,6 +40,11 @@ If (@('1', 'true') -contains $useAzureTrustedSigning) {
     If ($LastExitCode -ne 0) { Exit $LastExitCode }
 }
 
+Write-Host "--- :windows: Enabling long path support"
+New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" `
+    -Name "LongPathsEnabled" -Value 1 -PropertyType DWORD -Force
+If ($LastExitCode -ne 0) { Exit $LastExitCode }
+
 Write-Host "--- :npm: Installing Node dependencies"
 bash .buildkite/commands/install-node-dependencies.sh
 If ($LastExitCode -ne 0) { Exit $LastExitCode }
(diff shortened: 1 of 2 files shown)


.buildkite/commands/build-for-windows.ps1
```diff
@@ -40,10 +40,24 @@ If (@('1', 'true') -contains $useAzureTrustedSigning) {
     If ($LastExitCode -ne 0) { Exit $LastExitCode }
 }
 
+Write-Host "--- :windows: Enabling long path support"
+New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" `
+    -Name "LongPathsEnabled" -Value 1 -PropertyType DWORD -Force
+If ($LastExitCode -ne 0) { Exit $LastExitCode }
```
mokagio, line 46: Interesting. The AMI already should enable long paths with these instructions:

```
Write-Output "Enable long path behavior"
# See:
#
# - https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#maximum-path-length-limitation
# - https://support.atlassian.com/bamboo/kb/git-checkouts-fail-on-windows-with-filename-too-long-error-unable-to-create-file-errors/
Set-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem' -Name 'LongPathsEnabled' -Value 1
git config --system core.longpaths true
```

I see this command adds `-PropertyType DWORD -Force` which maybe should be pushed upstream to the AMI, too.

However, it's hard to tell whether the issue is with the original command or with `electron-winstaller` below. That is, I wonder if the latest stable version of that package will work _without_ the additional params in the long-path configuration.

mokagio:
Discovered with Claude: The `LongPathsEnabled` registry key does not seem to reach the binary that's actually throwing.

The failure is in Squirrel's bundled `Update.exe`, which uses the legacy .NET Framework path handling. The stack trace from build [17201](https://buildkite.com/automattic/studio/builds/17201) suggest it:

```
System.AggregateException: One or more errors occurred.
 ---> System.IO.PathTooLongException: The specified path, file name, or both are too long.
      The fully qualified file name must be less than 260 characters, and the directory
      name must be less than 248 characters.
   at System.IO.Path.LegacyNormalizePath(String path, Boolean fullCheck, Int32 maxPathLength, Boolean expandShortPaths)
   at System.IO.Path.InternalGetDirectoryName(String path)
   at Squirrel.ReleasePackage.<>c__DisplayClass14_0.b__0()
   at System.Threading.Tasks.Task.Execute()
   --- End of inner exception stack trace ---
   at Squirrel.ReleasePackage.CreateReleasePackage(String outputFile, String packagesRootDir, ...)
   at Squirrel.Update.Program.Releasify(...)
   at Squirrel.Update.Program.main(String[] args)
```

The key frame is `System.IO.Path.LegacyNormalizePath`.

mokagio:
I researched an alternative with AI and pushed it in #3735

wojtekn:
Closing in favor of https://github.com/Automattic/studio/pull/3735

PR #3728, opened 2026-06-08
Ask for confirmation before creating a local site in pull-reprint

epeicher:
## Related issues

- Fixes STUDIO-9

## How AI was used in this PR

Claude Code assisted with the code. I reviewed the diff myself, simplified, and manually verified the prompt, the `--yes` bypass, and the decline-cleanup path against a real WordPress.com staging site.

## Proposed Changes

Today `studio pull-reprint` immediately starts materializing a brand-new local site — in a `~/Studio` directory unrelated to the current one — and kicks off a full remote pull, with no confirmation. Someone exploring the command, or who mistyped a URL or name, can unexpectedly create a site and start downloading.

This adds a confirmation step on a fresh, interactive pull that spells out exactly what is about to happen before anything is created:

```
This will create a new local site "<name>" at <path>, pulling from <url>. Continue? (Y/n)
```

Declining cancels cleanly and removes the just-created session state, so nothing is left behind for a later run to treat as resumable.

To stay scriptable, the prompt only appears for a brand-new pull on an interactive terminal. A new `--yes`/`-y` flag skips it; resuming an in-progress pull (which already confirmed on its first run) and non-interactive callers (CI, Desktop) keep the existing non-prompting behavior.

## Testing Instructions

Requires a built CLI and the `STUDIO_ENABLE_PULL_REPRINT=true` feature flag.

- **Fresh interactive pull shows the prompt and can be declined:**
   ```
   STUDIO_ENABLE_PULL_REPRINT=true studio pull-reprint --url <your-wpcom-site> --verbose
   ```
   - Confirm the `Continue? (Y/n)` prompt appears with the correct site name, local path, and source URL.
   - Answer `n` → prints `Cancelled.`; verify no site dir is created under `~/Studio` and no session dir is left under `~/.studio/pulls`.
   - Answer `y` → the pull proceeds exactly as before.
- **`--yes` skips the prompt:** add `--yes` to the command above; the pull starts immediately with no prompt.
- **Resume does not re-prompt:** start a pull, interrupt it, then re-run the same command — it resumes without prompting.
- **Non-interactive runs do not prompt:** piping/redirecting stdin (e.g. CI) proceeds without a prompt, as before.

## Pre-merge Checklist


apps/cli/commands/pull-reprint.ts
```diff
@@ -282,6 +291,35 @@ export async function runCommand(
...
+	// resumes already made this choice and non-interactive callers
+	// (CI, Desktop) must keep the current non-prompting behavior.
+	const shouldConfirmSiteCreation = created && !! process.stdin.isTTY && ! yes;
+	if ( shouldConfirmSiteCreation ) {
+		const shouldContinue = await confirm( {
+			message: sprintf(
+				// translators: 1: local site name, 2: local site path, 3: remote source URL.
+				__( 'This will create a new local site "%1$s" at %2$s, pulling from %3$s. Continue?' ),
+				studioMetadata.siteName,
+				studioMetadata.sitePath,
```
fredrikekelund:
```suggestion
				getPrettyPath( studioMetadata.sitePath ),
```
Minor, but this could potentially make the path more readable. This function comes from `cli/lib/utils`

apps/cli/commands/pull-reprint.ts
```diff
@@ -282,6 +291,35 @@ export async function runCommand(
...
+			),
+			default: true,
+		} );
+		if ( ! shouldContinue ) {
+			// `getPullSessionMetadata` already persisted `pull.json` and
+			// created `technicalSiteDirectory`.  Remove that just-created
+			// state so a later run doesn't treat it as a resumable session.
+			// Only safe to delete on a fresh pull (`created === true`); never
+			// touch a pre-existing resumable session's directory.
+			fs.rmSync( studioMetadata.technicalSiteDirectory, { recursive: true, force: true } );
```
fredrikekelund, line 316: If we don't do this, the primary thing that happens is that `getPullSessionMetadata` will return `created: false` on the next run. I'd clarify this in the comment (and maybe slim down the rest of the comment slightly)

fredrikekelund (approved):
LGTM 👍 I left two small comments on the implementation.

apps/cli/commands/pull-reprint.ts
epeicher: Thanks for the suggestion, This looks better now, added as part of [ef9c089](https://github.com/Automattic/studio/pull/3728/commits/ef9c089d1191f87f2d0e0098c1ec1ccff04963f9)

|Before | After|
|-------|------|
|CleanShot 2026-06-09 at 17 12 20@2x | CleanShot 2026-06-09 at 17 11 18@2x|

apps/cli/commands/pull-reprint.ts
epeicher: Updated as part of ef9c089d1 and 50346b546

PR #3729, opened 2026-06-08
Bring the interactive site selector to pull-reprint

epeicher:
## Related issues

- Fixes STU-1811

### Selector compared with previous one
CleanShot 2026-06-09 at 18 16 32@2x

## How AI was used in this PR

Claude Code adapted the existing `pull` site selector (`pickSyncSite`) into a trimmed `pull-reprint` variant per the issue's guidance to copy rather than abstract (since `pull` is being deprecated), added unit tests for both the picker and the source-resolution flow, and verified the change — `cli:build`, `typecheck`, and the new test suites all pass. I reviewed the diff against the reference implementation and confirmed there are no behavior changes on the non-interactive paths.

## Proposed Changes

When you run `pull-reprint` without `--url` and have more than one connected WordPress.com site, the command previously printed the list and aborted, forcing you to re-run with an explicit `--url`. Now, in an interactive terminal, it opens a searchable site picker (type to filter by name or hostname, arrow keys to navigate, Enter to select) — the same friendly selection experience already offered by `pull`. Cancelling with Esc or Ctrl-C is treated as a clean no-op: nothing is downloaded and no local state is created.

Existing behavior is preserved everywhere else: a single connected site is still auto-selected, the zero-site and `--url`/`--secret` paths are unchanged, and non-interactive callers (CI, the Desktop app driving the CLI) keep the original print-list-and-abort behavior so scripted usage is unaffected.

## Testing Instructions

1. Authenticate with a WordPress.com account that has **two or more** connected sites (`studio auth login`).
2. From an interactive terminal, run:
   ```
   STUDIO_ENABLE_PULL_REPRINT=true node apps/cli/dist/cli/main.mjs pull-reprint --verbose
   ```
   (run `npm run cli:build` first)
3. Confirm the interactive picker appears. Type to filter, select a site, and verify the pull proceeds against the chosen site.
4. Re-run and press Esc (or Ctrl-C) at the picker — confirm it exits cleanly with no site created under `~/Studio/` and no state under `~/.studio/pulls/`.
5. Regression checks: with a single connected site it should auto-select without prompting; piping to a non-TTY (e.g. `... pull-reprint | cat`) should fall back to printing the list and aborting.

## Pre-merge Checklist


apps/cli/lib/wpcom-site-picker.ts
```diff
@@ -0,0 +1,117 @@
...
+ * for the `pull-reprint` source-selection flow.  It keeps the
+ * `[staging]` badge but, because `WpComSiteInfo` has no `syncSupport`,
+ * drops the disabled-entry and sync-support logic — every site is
+ * selectable.
+ *
+ * Type-to-filter matches on the site name and normalized hostname.  Esc
+ * cancels the prompt (TTY only, via an `AbortController` wired to the
+ * raw escape byte) and resolves to `undefined`, as does Ctrl-C — so the
+ * caller can treat a cancellation as a clean no-op rather than an error.
+ */
```
fredrikekelund: In my testing, it doesn't appear that Reprint can sync non-Atomic sites. I would think this should be theoretically possible with how the Reprint exporter is deployed through wpcomsh, but I'm not familiar enough with wpcomsh to say for sure. 

In any case, we need to adjust this implementation:

1. We should only display syncable sites in the list. The simple solution to this problem is to change the site fetching to use the same implementation as `apps/cli/commands/pull.ts`
2. If we do this, then there's not really a need for separate implementations between the `pull` and `pull-reprint` commands, as I understand it. Yes, the Jetpack-powered `pull` command is going away soon enough, so this isn't very important, but it'd be a nice touch.
3. We should talk to Adam about whether Reprint is intended to work for simple sites, too, and what adjustments we might need to make that work.

apps/cli/commands/pull-reprint.ts
```diff
@@ -1176,19 +1179,35 @@ async function resolveSourceSite(
...
+			// print-list-and-abort behavior so scripted callers are
+			// unaffected.
+			if ( ! process.stdin.isTTY ) {
+				console.log( __( 'Connected WordPress.com sites:' ) );
+				console.log( formatWpComSitesList( sites ) );
+				console.log( '' );
+				throw new LoggerError(
+					__( 'Multiple WordPress.com sites are available. Re-run with `--url <site-url>`.' )
+				);
+			}
```
fredrikekelund, line 1181: If no `--url` option was provided and this isn't a TTY, I think it's fine to just exit with an error. Realistically, the main case in which this command would be triggered in a non-TTY context is when it is spawned as a child command of Studio. 

If we follow this suggestion, then let's also remove the `formatWpComSitesList` function. 

I could also see us moving the `if ( ! process.stdin.isTTY )` into the `pickWpComSite` implementation, but I'll leave that to you, @epeicher

fredrikekelund (commented):
The control works well, but we need to look into the site list fetching. Right now, non-Atomic sites are displayed in the list, but selecting them leads to an error like this for me:


```
✖ Invalid API response format: Zod validation failed: [
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "body",
      "data",
      "secret"
    ],
    "message": "Invalid input: expected string, received undefined"
  }
]. Raw response: {"code":404,"headers":[{"name":"Content-Type","value":"application/json"}],"body":{"error":"rest_no_route","message":"No route was found matching the URL and request method.","data":{"status":404}}}
```

apps/cli/lib/wpcom-site-picker.ts
fredrikekelund: > separate implementations

I guess I explicitly said not to worry about this in the Linear issue 😅 Let's ignore this unless the implementation end up 1:1

epeicher: Well spotted! I didn't have any simple sites so when comparing with the `pull` command I was obtaining the same results, I will work on that.

epeicher: Addressed as part of [a5ea784](https://github.com/Automattic/studio/pull/3729/commits/a5ea784e510381e398d9b60ad720ecf144d8f43a)

apps/cli/commands/pull-reprint.ts
epeicher: Addressed as part of fc3d1fa01, with regards to moving the condition, I prefer the TTY gate at the call site: "are we allowed to prompt, and what do we do if not?" and keep the picker focused on "given a list, return a choice or a cancel."

epeicher:
Hi @fredrikekelund, after our conversation, I have checked the conditions to pull sites and I've decided to unify them so `pull` and `pull-reprint` use the same logic. The `jetpack` and `is_wpcom_atomic`  conditions are basically the same, I have not been able to simulate an atomic non-jetpack site so I don't think we need to worry about that.

The logic is greatly simplified, can you have a look at it and let me know any issues?

fredrikekelund (approved):
Looks good! 👍 There's more we can do to streamline and simplify site fetching and preflight logic, but we can do that in follow-up PRs

PR #3732, opened 2026-06-08
Add --omit=peer to install:bundle to prevent Windows path-too-long failures

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

Claude Code made the one-line change and ran typecheck to confirm nothing broke.

## Proposed Changes

The `install:bundle` script installs production dependencies into the packaged app before galactus prunes devDependencies. Without `--omit=peer`, npm auto-installs peer dependencies (e.g. `react-native`, `@react-native/*`) that galactus does not prune because they are not marked as `devDependencies`. On Windows these produce deeply nested paths that exceed the MAX_PATH limit and cause the Squirrel maker to fail.

Adding `--omit=peer` prevents npm from installing peer-only packages during the bundle step. There is no overlap between auto-installed peer deps and runtime `dependencies` in `apps/studio/package.json`, so this is safe.

## Testing Instructions

- Run `npm run make:windows-x64` (or trigger a Windows CI build) and confirm the Squirrel maker no longer hits path-too-long errors.
- `npm run typecheck` in `apps/studio` continues to pass.

## Pre-merge Checklist


wojtekn:
Closing in favor of https://github.com/Automattic/studio/pull/3735

PR #3734, opened 2026-06-08
Avoid Node binary downloads for matching runtime

### LINKED ISSUE #3731: Packaging should not require a fresh Node binary download when current Node matches target
## Problem
Lab packaging for Studio repeatedly failed in `scripts/download-node-binary.ts` when nodejs.org fetches timed out or IPv6 was unreachable. This made many-sites benchmark runs fail before the app launched.

## Observed
`npm run package --workspace=studio-app` failed inside Electron Forge with:

```text
Command failed: npx tsx .../scripts/download-node-binary.ts linux x64
Error: fetch failed
```

## Why this matters
The packaging step currently always downloads the Node archive, even when packaging for the same platform/arch/version as the running Node process. That makes CI/lab benchmarks depend on a fresh network fetch every package run.

## Proposed shape
- If `process.platform`, `process.arch`, and `process.version` match the requested target, copy `process.execPath` into `apps/studio/bin/node` instead of downloading.
- Otherwise, use a persistent cache for downloaded Node archives before falling back to the network.

## Acceptance criteria
- Same-platform packaging can complete without network access to nodejs.org.
- Cross-platform packaging still downloads/extracts the correct archive.
- The script logs whether it copied the current binary, used cache, or downloaded.

chubes4:
## Related issues

Fixes #3731.

## Proposed Changes

Studio packaging no longer requires a fresh nodejs.org download when the requested platform, architecture, and Node version match the runtime running the packaging script. In that case, `scripts/download-node-binary.ts` copies `process.execPath` directly into the app bundle.

For cross-target packaging, the script now stores downloaded Node archives in a persistent cache before extracting them, so repeated packaging runs can reuse the archive instead of depending on a fresh network fetch every time. Downloads still use the existing retry helper.

## Testing Instructions

- `npx vitest --project scripts scripts/download-node-binary.test.ts`
- `npx tsc -p scripts/tsconfig.json --noEmit`
- `npx eslint scripts/download-node-binary.ts scripts/download-node-binary.test.ts scripts/vitest.config.ts`
- `npm run typecheck`

The focused tests cover:

- same platform/arch/version copies the current Node binary without downloading
- cache hit extracts without downloading
- cache miss downloads once, stores the archive, then extracts from cache

## AI assistance

- **AI assistance:** Yes
- **Used for:** Investigated the Linux packaging failure, drafted the implementation and targeted tests, and ran verification commands. Chris remains responsible for review and merge decisions.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? N/A, packaging script only.


chubes4:
Local full-unit verification after the Buildkite failure:

- `env -u STUDIO_RUNTIME npm run test` passed: 231 files, 1922 tests.
- `STUDIO_RUNTIME=native-php npm run test` reproduces the same broad existing CLI failure shape locally: `wordpress-server-manager.test.ts` expects Playground runtime but gets native PHP, and `site create` tests reject PHP 8.0 for native PHP before reaching their assertions.

That points at the failing Buildkite Unit Tests status being caused by `STUDIO_RUNTIME=native-php` leaking into the unit-test process, not by the Node binary packaging change itself. The PR-specific focused tests also pass:

- `env -u STUDIO_RUNTIME npm run test -- --project scripts scripts/download-node-binary.test.ts`
- `npx tsc -p scripts/tsconfig.json --noEmit`
- `npx eslint scripts/download-node-binary.ts scripts/download-node-binary.test.ts scripts/vitest.config.ts`

fredrikekelund (commented):
Here's a suggestion for a simpler approach: change `scripts/download-node-binary.ts` to check if `apps/studio/bin/node` already exists. If it does, and if `apps/studio/bin/node --version` outputs what we want, exit the script early.

If an engineer wanted to use the node binary that already exists on their system, they can manually copy it to `apps/studio/bin/node`. 

Would this work for the CI problems you experienced, @chubes4?

chubes4:
Thanks, this makes sense. I revised the PR toward that contract instead of implicitly copying `process.execPath`.

The current shape now:

- checks `apps/studio/bin/node` / `node.exe`
- runs `--version`
- exits early when it matches `.nvmrc`
- otherwise keeps the existing official nodejs.org archive download/extract path

That keeps the downloaded Node archive as the canonical packaging source while avoiding the repeated network fetch during back-to-back packaging/benchmark runs once the bundled binary is already present. That repeated fetch was the actual wrinkle we hit in lab runs: not that Studio downloads Node, but that repeated isolated packaging made the Node download a required network hop every time.

Proof I ran after the revision:

- temporary `apps/studio/bin/node` shim returning `v24.15.0`, then `npx tsx scripts/download-node-binary.ts linux x64` exits early with `Using existing Node.js v24.15.0 binary...`
- `npx tsc -p scripts/tsconfig.json --noEmit`
- `npx eslint scripts/download-node-binary.ts vitest.config.ts`
- `env -u STUDIO_RUNTIME npm run test` passed: 230 files / 1919 tests
- `npm run typecheck`

fredrikekelund (approved):
Nice and simple 👍 Works as expected.

I took the liberty of merging the latest changes from trunk and resolving merge conflicts here. Sorry for not getting back to this for a second review sooner

PR #3735, opened 2026-06-09
Prune unused AI provider SDKs from CLI bundle

mokagio:
> [!NOTE]  
> @wojtekn this JS/TS manipulation of the Electron build structure is outside my area of expertise, so it's possible the result is a big AI slop. At the same time, it does seem to address the build failure issue with the long path, so it might be a good place to start.
>
> I'm having my agent address the Copilot comments now. Next step will be to get a different one to look over the work for possible improvements.

## Related issues

- Related to #3727 (alternative fix for the same Windows `PathTooLongException`)

## How AI was used in this PR

Claude Code (Opus 4.8) diagnosed the root cause from the CI logs, verified the fix empirically against a reproduced bundle tree, and wrote the prune script and tests. I reviewed the analysis and the diff.

## Proposed Changes

The Windows Squirrel maker crashes with `PathTooLongException` while packing the app. The real cause is `@earendil-works/pi-coding-agent`'s transitive `@mistralai/mistralai` — a speakeasy-generated SDK whose ~200-char filenames, nested under `pi-coding-agent/node_modules`, exceed Windows' 260-char path limit. The AWS Bedrock SDK (`@aws-sdk`/`@smithy`) overflows the same way.

These are **lazy, dynamically-imported** `pi-ai` providers, and Studio only ever exposes the **Anthropic and OpenAI** model families (`tools/common/ai/models.ts`). Mistral, Bedrock and Google are therefore unreachable dead weight. This strips them from the bundled CLI during packaging, which:

- fixes the path limit — **183 over-limit files → 0** (worst path drops to 236 / 260), and
- reclaims tens of MB from the installer.

A missing provider degrades gracefully through pi-ai's lazy loader, so removing them is safe.

> Note: this is an alternative to #3727, opened in parallel rather than stacked because that PR's commits (forge `ignore`, registry key, `nuget.exe` swap) are no-ops for this failure — the exception is thrown by Squirrel's legacy `Update.exe` (`Path.LegacyNormalizePath`), which the `ignore`/registry/nuget levers don't reach. See #3727 for that analysis.

## Testing Instructions

- `npm test -- --project scripts` — unit tests for the prune (removes targets in hoisted + nested locations; keeps Anthropic/OpenAI and unrelated `@google/*`).
- Verified end-to-end against a reproduced bundle install: `pruneUnusedProviders` removed every offender at both the top level and under `pi-coding-agent`, taking over-260 files from 183 to 0.
- The Windows build (`make:windows-*`) should now pass the Squirrel maker; macOS/Linux unaffected (and benefit from the smaller bundle).

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? (`npm run typecheck` and `eslint` pass)

---

*Authored by Claude (Opus 4.8) on behalf of @mokagio with approval.*


mokagio:
✅ **Validated on Windows CI.** Triggered the dev Windows build ([Buildkite #17206](https://buildkite.com/automattic/studio/builds/17206)) and both packaging jobs passed:

- 🔨 Windows Dev Build — **x64: passed**
- 🔨 Windows Dev Build — **arm64: passed**

The x64 job cleared the exact `Making a squirrel distributable` step that crashed #3727, with no `PathTooLongException` — the prune removed the over-long provider trees before Squirrel walked the directory. All standard checks (Lint, Unit, E2E, Performance, CodeQL) are green too.

---

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

scripts/remove-unused-ai-providers.ts
```diff
@@ -0,0 +1,93 @@
...
+ * auto-generated filenames that, nested under pi-coding-agent/node_modules, push paths past
+ * Windows' 260-char limit and crash the Squirrel maker. Removing them fixes that and reclaims
+ * tens of MB.
+ *
+ * Removal is keyed on directory name, so it strips both the hoisted copy and any nested
+ * duplicates wherever npm placed them in the tree.
+ */
+
+import { readdirSync, rmSync, statSync } from 'fs';
+import { join } from 'path';
```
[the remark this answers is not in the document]
mokagio: Addressed in https://github.com/Automattic/studio/commit/11f57ba1 — added an `existsSync` guard at the top of `pruneUnusedProviders` so the common missing-`node_modules` case short-circuits with a cheap stat instead of a thrown-and-caught ENOENT.

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

scripts/remove-unused-ai-providers.ts
```diff
@@ -0,0 +1,93 @@
...
+			try {
+				scopedEntries = readdirSync( packageDir, { withFileTypes: true } );
+			} catch {
+				continue;
+			}
+			for ( const scoped of scopedEntries ) {
+				if ( scoped.isDirectory() ) {
+					pruneUnusedProviders( join( packageDir, scoped.name, 'node_modules' ) );
+				}
+			}
```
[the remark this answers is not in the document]
mokagio: Addressed in https://github.com/Automattic/studio/commit/11f57ba1 — added an `existsSync` guard at the top of `pruneUnusedProviders` so the common missing-`node_modules` case short-circuits with a cheap stat instead of a thrown-and-caught ENOENT.

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

scripts/remove-unused-ai-providers.ts
```diff
@@ -0,0 +1,93 @@
...
+			continue;
+		}
+		pruneUnusedProviders( join( packageDir, 'node_modules' ) );
+	}
```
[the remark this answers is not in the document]
mokagio: Addressed in https://github.com/Automattic/studio/commit/11f57ba1 — added an `existsSync` guard at the top of `pruneUnusedProviders` so the common missing-`node_modules` case short-circuits with a cheap stat instead of a thrown-and-caught ENOENT.

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

scripts/remove-unused-ai-providers.ts
```diff
@@ -0,0 +1,93 @@
...
+			console.log( `Removed ${ target } from ${ nodeModulesDir }` );
+		} catch ( e ) {
+			// Tolerate transient failures (Windows AV locks, permissions). This is a size /
+			// path-length optimization; aborting packaging over a stuck file would be worse.
+			console.warn(
+				`Could not remove ${ target } from ${ nodeModulesDir }: ${
+					e instanceof Error ? e.message : String( e )
+				}`
+			);
+		}
```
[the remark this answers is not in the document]
mokagio: Addressed in https://github.com/Automattic/studio/commit/22b72d81 — `pruneUnusedProviders` now returns the directories it couldn't remove, and the Forge hook throws on `win32` when any remain (with the offending paths), so a failed prune fails the build with context instead of degrading into a later `PathTooLongException`. Non-Windows platforms keep tolerating a stuck removal, where the prune is only a size optimization.

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

wojtekn:
Thanks for exploring that @mokagio .

I tried different approaches in https://github.com/Automattic/studio/pull/3727 and https://github.com/Automattic/studio/pull/3732, but none solved the issue. Claude was proposing the pruning way, but I kept trying to find another way.

I think it's reasonable to proceed with that approach, and in the future we could consider dropping such auto-generated dependencies, and/or migrating away from Squirrel - either by using another existing or custom maker for Forge, or better, migrating from Forge to Electron Builder.

scripts/remove-unused-ai-providers.test.ts
fredrikekelund, line 1: Let's remove this test. Unit tests for a build script just doesn't make sense to me

scripts/remove-unused-ai-providers.ts
fredrikekelund, line 1: I would inline this logic in `forge.config.ts` and simplify it by following the `apps/cli/vite.config.prod.ts` example and relying on `globSync` to collect the paths we want to delete, then passing those to `fs.rmSync`

scripts/vitest.config.ts
fredrikekelund, line 1: Per my other comment, let's remove this

fredrikekelund (commented):
I haven't tested this, but fundamentally, the strategy seems sound to me. I think we can compress the actual implementation a bit, though (see my inline comments)

scripts/remove-unused-ai-providers.test.ts
wojtekn: Makes sense, removed.

scripts/remove-unused-ai-providers.ts
wojtekn: Thanks, done.

scripts/vitest.config.ts
wojtekn: Thanks, done.

wojtekn:
@mokagio I simplified the code and merged. Thanks for the help with this.

wojtekn (approved):

PR #3738, opened 2026-06-09
Agentic UI: Application Preferences

shaunandrews:
This PR aims to implement application preferences for Studio. These are global settings across the application, like appearance, Skills, MCP, and more.

image


shaunandrews:
Code review findings (via Claude Code) that apply to this branch:

**1. Cmd+N does nothing on the Desks UI (the default mode)**
`CommandOrControl+N` was removed from the "Add Site" menu item and reassigned to the `new-chat-in-current-site` shortcut, but the handler for that shortcut is only registered in ui-classic's session view. On ui-desks — currently the default mode — Cmd+N silently does nothing. Either register a desks handler or keep a menu accelerator as fallback.

**2. Duplicate subscriptions to `toggle-site-preview`**
Both `apps/ui/src/hooks/use-session-ui.tsx` (~L119) and `apps/ui/src/ui-classic/router/layout-dashboard/index.tsx` (~L98) subscribe to the same IPC event. The provider has a nesting guard, but the layout-dashboard subscription lives outside it — if both end up mounted and active, one menu click toggles twice (net no-op). Worth consolidating into a single owner.

**3. `isApplePlatform` duplicates `isMacPlatform`**
`apps/ui/src/lib/keyboard-shortcuts.ts` adds `isApplePlatform()` while `apps/ui/src/lib/platform.ts` already has `isMacPlatform()`, with slightly different regexes. These should be one function so platform detection can't drift. (The third copy in `inspector-script.ts` is fine — it's serialized into the guest page and can't import.)

**4. `window.confirm()` for destructive preview-site deletion**
`settings-view/index.tsx` uses native `window.confirm()` to confirm deleting a preview site. The app already has a styled Dialog pattern for destructive actions (e.g. `DeleteSiteDialog` in site-list); native confirm breaks the visual language and can behave inconsistently in Electron renderers.

**5. Saved-defaults resolution duplicated in settings form logic**
In `settings-view/index.tsx`, `toFormData` and `diffFromSaved` each independently resolve saved defaults (`defaultSiteDirectory ?? ''`, `messageSendShortcut ?? DEFAULT_MESSAGE_SEND_SHORTCUT`, `studioCliInstalled ?? false`). Changing a default requires updating both in sync — worth extracting a single resolver.

Things checked that look good: i18n coverage in the new settings view is complete, ⌘/Ctrl shortcut display handles Windows/Linux correctly with test coverage, and the settings CSS consistently uses design tokens.

PR #3741, opened 2026-06-09
Make creating WooCommerce shop sites easy with Studio Code

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Built with Claude Code. The skill/prompt changes were driven by analyzing real Studio Code session transcripts of an agent repeatedly failing to build a proper shop ("MyPet") — each fix targets a specific failure observed in those transcripts (trigger never firing, then firing but misclassifying the shop as informational, then the cart link never landing). All changes were reviewed by the author.

## Proposed Changes

When a user asks Studio Code to build a shop, the agent used to produce a static catalog mockup — no WooCommerce, no products, no cart — even for an explicit "pet shop." This makes ecommerce the reliable default and gives the agent a complete happy path for standing up a real store.

- A shop/store request is now treated as ecommerce by default (unless the user says otherwise). The agent provisions WooCommerce, turns off the "Coming soon" placeholder, creates products relevant to the shop's niche with real imported images, and adds a cart (mini-cart) to the header.
- The site planning step now commits to a "Functionality & plugins" decision up front, so plugin needs are decided while gathering the spec rather than being skipped mid-build.
- Design verification now covers every page — including the WooCommerce pages (Shop, Cart, Checkout) — while only iterating on the home page, so secondary pages get checked without inflating build time.
- The Skill tool call renders as a single compact row in the chat transcript instead of dumping the full skill body, which keeps the conversation readable as the number of skills grows.

## Testing Instructions

1. `npm run cli:build`
2. In Studio Code, prompt: *"Build a small pet shop site called MyPet, keep it small, only 2-3 pages. I want to sell dog and cat products."*
3. Verify the agent: installs/activates WooCommerce, creates several products (each with an image), disables the "Coming soon" page, and adds a cart link to the header.
4. Verify visual-polish runs across all pages (Home, Shop, Cart, etc.) but only loops on the home page.
5. Optional — run the new eval: `npm run eval` (look for the `woocommerce-shop-setup` case).

## Pre-merge Checklist

- [x] Logic verified against real agent session transcripts; minimal eval added.

### DIFF
--- a/apps/cli/ai/skills/plugin-recommendations/SKILL.md
+++ b/apps/cli/ai/skills/plugin-recommendations/SKILL.md
@@ -52,7 +52,7 @@ wp_cli plugin install woocommerce --activate
 wp_cli option delete _transient__wc_activation_redirect
 ```
 
-3. Configure store basics. Match the currency and base location to the user's context when known, otherwise use sensible defaults. Also turn off "Coming soon" mode, which fresh WooCommerce installs enable by default - otherwise the storefront shows a "Something big is brewing! Our store is in the works and will be launching soon!" placeholder instead of the real site:
+3. Configure store basics. Match the currency and base location to the user's context when known, otherwise use sensible defaults. Also turn off "Coming soon" mode, which fresh WooCommerce installs enable by default:
 
 ```text
 wp_cli option update woocommerce_coming_soon no
--- a/apps/cli/ai/skills/plugin-recommendations/SKILL.md
+++ b/apps/cli/ai/skills/plugin-recommendations/SKILL.md
@@ -6,7 +6,7 @@ user-invokable: true
 
 # Plugin Recommendations
 
-Use this skill when the user asks for a feature that core WordPress blocks do not cleanly provide, such as forms, slideshows, related content, business hours, ecommerce, events, LMS/course features, or third-party embeds.
+Use this skill when the user asks for a feature that core WordPress blocks do not cleanly provide, such as forms, slideshows, related content, business hours, shops/stores/ecommerce, events, LMS/course features, or third-party embeds.
 
 ## Decision Rules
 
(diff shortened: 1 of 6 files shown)


eval/promptfoo.config.yaml
```diff
@@ -338,3 +338,41 @@ tests:
                 : `no jetpack/* block in any Write tool call; sample: ${(writtenContents[0] || '').slice(0, 300)}`,
             };
           });
+
+  # Regression: a "shop"/"store" request must set up WooCommerce WITH products,
+  # even when framed with scope/style qualifiers ("small", "for inspiration")
```
sejas, line 343: 💖

apps/studio/src/components/studio-code-session/conversation/index.tsx
```diff
@@ -236,7 +236,10 @@ function ToolUseRow( {
 	const detail = getToolDetail( name, input );
 	const [ expanded, setExpanded ] = useState( false );
 	const resultText = result?.text?.trim() ?? '';
-	const hasOutput = resultText.length > 0;
+	// The Skill tool returns the full skill runbook as its result. That body is
+	// guidance for the agent, not something the user needs to read inline, so we
+	// render the call as a single compact row (label + skill name) and omit it.
+	const hasOutput = resultText.length > 0 && name !== 'Skill';
```
sejas: Nit: We could consider moving the string `'Skill'` to a constant, to make sure we are comparing the right value used in `defineTool`: https://github.com/Automattic/studio/blob/516b0a90db36c299e405a05532e80605f3bbfa2c/apps/cli/ai/tools/skill.ts#L17

sejas (commented):
Nice! Thanks for improving the creation of WooCommerce websites.
I tested it by running `npm run eval -- --filter-pattern "shop request sets up WooCommerce"` and also `npm run cli:build && node apps/cli/dist/cli/main.mjs code` with the following prompt: `Create a small shop of surfing boards top brands, keep it small only 2 pages. The name of the shop is CanarySurf`

The shop looked good and it contained 4 products.

Image

The eval failed to install WooCommerce and create the products. It seems we might need to tweak the skill description a bit more.

Image

youknowriad:
I think the latest skill update might fix that eval issue.

sejas:
Yes! It passed correctly after latest changes. Thank you!

sejas (approved):
Working great!

apps/cli/ai/skills/plugin-recommendations/SKILL.md
```diff
@@ -36,6 +36,57 @@ wp_cli eval 'foreach (\WP_Block_Type_Registry::get_instance()->get_all_registere
...
+wp_cli plugin install woocommerce --activate
+```
+
+2. Activation automatically creates the Shop, Cart, Checkout, and My Account pages. Suppress the setup-wizard redirect so the storefront is usable right away:
+
+```text
+wp_cli option delete _transient__wc_activation_redirect
+```
+
+3. Configure store basics. Match the currency and base location to the user's context when known, otherwise use sensible defaults. Also turn off "Coming soon" mode, which fresh WooCommerce installs enable by default - otherwise the storefront shows a "Something big is brewing! Our store is in the works and will be launching soon!" placeholder instead of the real site:
```
lezama: > otherwise the storefront shows a "Something big is brewing! Our store is in the works and will be launching soon!" placeholder instead of the real site:

is this needed?

lezama (approved):

youknowriad:
The unit test failure seems like the one being worked on separately and failing on trunk. So I'm merging this.

PR #3747, opened 2026-06-09
Add Fable 5 model to Studio Code agent

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Authored with Claude Code. I located the central model registry, added the new entry, and verified the change with eslint and the TypeScript type checker. The author reviewed the change.

## Proposed Changes

Enables the **Fable 5** model in the Studio Code agent. It now appears as a selectable option in the chat model picker alongside the existing Anthropic and OpenAI models. Because the model registry is centralized, this single registry entry propagates automatically to the model picker, provider availability lists, the agent runtime (served via the Anthropic-messages path with reasoning enabled), and session model persistence/validation.

It is not the default model — users opt into it via the picker.

> Note: the upstream wire id is `claude-fable-5`, following the existing `claude-<name>-<n>` convention. If the real provider identifier differs, that value should be corrected — a mismatch would surface at request time rather than compile time.

## Testing Instructions

- Open the Studio Code agent chat composer and confirm **Fable 5** appears in the model picker.
- Select it and confirm a turn runs (served through the Anthropic runtime).
- `npm run typecheck` passes.

## Pre-merge Checklist

### DIFF
--- a/tools/common/ai/models.ts
+++ b/tools/common/ai/models.ts
@@ -20,7 +20,6 @@ export interface AiModel {
 // reasoning turns.
 export const AI_MODELS = [
 	{ id: 'claude-sonnet-4-6', label: 'Sonnet 4.6', family: 'anthropic' },
-	{ id: 'claude-opus-4-7', label: 'Opus 4.7', family: 'anthropic' },
 	{ id: 'claude-opus-4-8', label: 'Opus 4.8', family: 'anthropic' },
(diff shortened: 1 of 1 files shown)


sejas (approved):
LGTM! Although we might need to enforce limits pretty soon!
I wouldn't remove the rest of models. The only one I would remove is Opus 4.7 because users can use Opus 4.8 which costs the same and is more capable.

sejas (approved):
Thanks for the changes! 🤞

PR #3744, opened 2026-06-09
Fix Dark theme for Studio Code

nightnei:
## Related issues

- Fixes STU-1809

## How AI was used in this PR
AI adjusted `color` and `background` properties. I tested by applying changes file by file, everything looks good.

## Proposed Changes
Dark theme for Studio Code is broken. With this PR we are fixing it to align with Studio styles.
<table>
<tr>
<td>BEFORE
<td>AFTER
</tr>
<tr>
<td>Screenshot 2026-06-09 at 16 30 08
<td>Screenshot 2026-06-09 at 16 30 49
</tr>
</table>

## Testing Instructions
1. npm start
2. Open Studio Code tab
3. Assert that for Light theme no regressions
4. Set Dark theme
5. Assert that everything looks good

PR #3748, opened 2026-06-09
Document dark-mode color token rule in AGENTS.md

sejas:
## Related issues

- Related to the "Fix Dark theme for Studio Code" (#3744) regression

## How AI was used in this PR

Claude Code analyzed the #3744 dark-theme fix, identified the root cause (renderer components used light-only `--wpds-color-*` tokens), and drafted the documentation guardrails. I reviewed the wording and scope myself before committing.

## Proposed Changes

The Studio Code session UI shipped broken in dark mode because its CSS used `--wpds-color-*` tokens from `@wordpress/ui`, which carry hardcoded light fallbacks and don't respond to `prefers-color-scheme: dark`. Components looked correct in light mode, so the regression slipped through.

This documents the rule so contributors (and AI agents) don't reintroduce it:

- Renderer CSS must use Studio's dark-aware `--color-frame-*` tokens for all colors; `--wpds-color-*` color tokens are banned. Non-color `--wpds-*` tokens (dimension, typography, border-width, elevation, cursor) remain fine.
- UI/CSS changes must be verified in both light and dark schemes before being considered done.
- Adds a CRITICAL pitfall entry referencing the regression for discoverability.

Docs-only change — no runtime impact.

## Testing Instructions

- Read `AGENTS.md`: confirm the new "Theming / colors (renderer CSS)" rule under Conventions, the dark-mode note in Post-Change Verification, and the "Dark Mode Color Tokens" entry under Common Pitfalls.

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors? (N/A — docs only)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,7 +7,7 @@ WordPress Studio - Electron desktop app for managing local WordPress sites. Buil
 **Dev/Build**: `npm start` | `npm run cli:build` | `node apps/cli/dist/cli/main.mjs`
 **Test**: `npm test [-- path/to/test.test.ts]` | `npm run e2e`
 **Quality**: `npx eslint --fix <files>` (lint and format ONLY modified files)
-**IMPORTANT - Post-Change Verification**: After applying code changes, always run the linter and format modified files (`npx eslint --fix <files>`), the type checker (`npm run typecheck`) and run relevant tests (`npm test [-- path/to/test]`) before considering the work complete.
+**IMPORTANT - Post-Change Verification**: After applying code changes, always run the linter and format modified files (`npx eslint --fix <files>`), the type checker (`npm run typecheck`) and run relevant tests (`npm test [-- path/to/test]`) before considering the work complete. For any UI/CSS change, also verify the result in **both light and dark** color schemes.
 **Package**: `npm run make` (builds installers for current platform)
 
 **IMPORTANT - Hot Reload**: Renderer auto-reloads, Main process needs restart (or `rs` in terminal). Changes to Main process IPC handlers require full restart.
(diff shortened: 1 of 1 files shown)


wojtekn (approved):
Looks good.

AGENTS.md
```diff
@@ -58,6 +58,7 @@ WordPress Studio - Electron desktop app for managing local WordPress sites. Buil
...
+**Theming / colors (renderer CSS)**: Studio supports light + dark via `@media (prefers-color-scheme: dark)`. For any **color** (text, background, border, fill, brand/theme, error/running states) **MUST** use Studio's dark-aware `--color-frame-*` tokens defined in `apps/studio/src/index.css` (e.g. `--color-frame-text`, `--color-frame-bg`, `--color-frame-surface`, `--color-frame-border`, `--color-frame-theme`, `--color-frame-error`). **NEVER** use `--wpds-color-*` tokens for color — they ship light-only fallbacks from `@wordpress/ui` and render broken (invisible text, wrong borders) in dark mode. Non-color `--wpds-*` tokens (`--wpds-dimension-*`, `--wpds-typography-*`, `--wpds-border-width-*`, `--wpds-elevation-*`, `--wpds-cursor-*`) are theme-independent and fine to use. When a needed color has no `--color-frame-*` token, add one (with both light and dark values) rather than reaching for `--wpds-color-*`.
```
youknowriad, line 61: Just nothing that this not true for all of studio and more precisely apps/ui. apps/ui uses a ThemeProvider for dark/light mode around the whole canvas which means we should do the opposite actually, always use `--wpds` variables because they adapt to the active theme.

youknowriad: Maybe we should update this to specify that each "app" has its own dark mode rules as it doesn't feel worth our time to refactor apps/studio to use ThemeProvider as there's a chance that we'll replace it entirely.

sejas: Good catch! I created a follow-up Automattic/studio#3814

PR #3750, opened 2026-06-10
Fix ineffective dynamic imports in scratchpad widget

wojtekn:
## Related issues

- N/A

## How AI was used in this PR

Claude Code identified the issue and made the fix.

## Proposed Changes

During `npm run make`, the build emitted two `[INEFFECTIVE_DYNAMIC_IMPORT]` warnings for `@/ui-desks/chats/widget-context` and `@/ui-desks/connectors/editor-commands` in the scratchpad widget definition. These modules are dynamically imported to break a circular dependency (`definition.ts` → `widget-context`/`editor-commands` → `registry` → `definition.ts`), but because they are also statically imported by other modules in the same chunk, the bundler cannot move them into a separate chunk and emits the warning.

The dynamic imports are load-bearing (removing them causes a runtime crash), so the fix is to suppress the warning via an `onwarn` handler in `apps/ui/vite.config.ts` rather than converting them to static imports.

## Testing Instructions

- Run `npm run make` and confirm the two `[INEFFECTIVE_DYNAMIC_IMPORT]` warnings are gone from the build output.
- Verify the scratchpad drag-to-chat and drag-to-connect interactions still work.

## Pre-merge Checklist

### DIFF
--- a/apps/ui/src/ui-desks/widgets/scratchpad/definition.ts
+++ b/apps/ui/src/ui-desks/widgets/scratchpad/definition.ts
@@ -1,16 +1,6 @@
 import { __ } from '@wordpress/i18n';
 import { external } from '@wordpress/icons';
 import {
-	buildWidgetContextDisplayMessage,
-	buildWidgetContextPrompt,
-} from '@/ui-desks/chats/widget-context';
-import {
-	completeConnectorPreview,
-	createConnectorPreview,
-	toPlainPoint,
-	updateConnectorEnd,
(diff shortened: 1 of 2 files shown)


epeicher (approved):
Thanks for removing the warning @wojtekn, I have tested, I can confirm the warning disappears, I have also smoke tested scratchpad and I have not found any regressions. LGTM!

|Before | After|
|-------|------|
| Image|Image   |

PR #3759, opened 2026-06-10
Rename Copy site to Duplicate site

nightnei:
## How AI was used in this PR
Didn't use

## Proposed Changes
NIT Every time I see "Copy site" - I feel confused. Maybe it's personal, but I used to see word "Duplicate" in such cases. And when I see "Copy", I expect to paste it somewhere.
I propose to rename to "Duplicate site".
macOS example:<br/>
Screenshot 2026-06-10 at 14 19 08

## Testing Instructions
Visual review should suffice


epeicher (approved):
Thanks @nightnei! I agree this is slightly clearer. I have reviewed, and this LGTM! But happy to hear other thoughts

Image

fredrikekelund:
I agree "Duplicate" is clearer. With "Copy", I don't know if there will be an intermediate step or not. With "Duplicate", I understand that there will be a new copy of the site available straight away

PR #3646, opened 2026-05-29
Agentic UI: Foundation

shaunandrews:
This PR aims to improve the new Agentic UI for Studio,  including the visual balance, user experience, and usability. Its not meant to be a complete solution; there are still many rough edges and broken flows. But this PR serves as a foundation for a series of other PRs and future work, including: #3717 #3737 #3738 

## Proposed Changes
- Increased base UI font-size to 14px.
- Drag handles for panels (like the sidebar menu and the browser preview) are now integrated into the border for the panel, no longer affecting the layout.
- Added smoother panel behavior, including animated site preview open/close and improved split-view layout.
- Removes the placeholder "Dashboard" and "Skills" menus and screen.
- A redesigned sidebar menu includes larger site icons, compact timestamps, clearer active states, consistent action placement, running-chat indicators, and an Unassigned group for chats that don't belong to a site.
- Clicking on a site in the sidebar shows a new site detail screen. Click the site avatar to manually toggle all active chats for a site.
- Sidebar chats now filter archived conversations out of the regular list.
- Added site overview and unassigned overview routes with active and archive chat lists, and a way to start a new chat.
- Studio's browser gets buttons for back, forward, and refresh, a page title display, and moves the annotation feature into the toolbar. The browser is also now shown by default. The OS View menu includes a toggle, and users can still use the keyboard shortcut to toggle.
- You can now add images to your chat messages.
- Included follow-up compatibility and hardening fixes for React 19 / @wordpress/ui, Base UI submenu behavior, markup sanitization, and CodeQL feedback.

## How to test
Checkout this PR, build the app, and then use the Electron (Studio) -> Beta Features menu to enable the Agentic UI.

## Screenshots
image
image
image
image
image

PR #3760, opened 2026-06-10
Remove placeholder Dashboard and Skills menus and screen

youknowriad:
## Related issues

- Related to #3646 (extracts the Dashboard/Skills removal so it can land on its own)

## How AI was used in this PR

Extracted from #3646 with Claude Code: identified the standalone slice of the larger Agentic UI PR, re-authored it against trunk, and verified with lint, typecheck, and the `apps/ui` test suite.

## Proposed Changes

The Agentic UI sidebar shipped with placeholder navigation that doesn't do anything useful yet: a "Skills" item that goes nowhere and a "Chat" item that opens a placeholder "Dashboard" screen (just a heading and a site count). This removes those placeholders so the UI only offers functional navigation, and replaces the placeholder landing screen with a useful one:

- The "Chat" and "Skills" sidebar items are gone.
- "Settings" — the only nav item left, which felt orphaned at the top — moves to the sidebar footer, sitting right above the user menu and pinned with it as the sidebar scrolls.
- The placeholder Dashboard screen and its `/dashboard` route are removed.
- On launch, the app now drops you straight into a conversation: it opens the most recent chat of the first site, creates a new empty chat for that site if it has none, or opens onboarding when there are no sites at all.
- Closing onboarding (when sites exist) follows the same logic instead of landing on the old placeholder.

This is a standalone extraction from #3646; the redesigned sidebar and the site overview screens will come in follow-up PRs.

## Testing Instructions

- Build the app and enable the **Agentic UI** via the Electron (Studio) → Beta Features menu.
- Confirm the sidebar no longer shows "Chat" or "Skills" at the top; "Settings" now sits at the bottom of the sidebar, above the user menu, and stays pinned there when the site list is long enough to scroll.
- Launch the app with at least one site that has chats: it should open that site's most recent chat (no "Dashboard" heading anywhere).
- Launch with a site that has no chats: a new empty chat is created and opened for it.
- Launch with no sites: onboarding opens.
- With at least one site, open onboarding (add a site) and close it: you should land back in a chat.
- Sessions, site settings, and settings routes still work as before, in both light and dark mode.

## Pre-merge Checklist


apps/ui/src/components/sidebar-layout/style.module.css
```diff
@@ -21,6 +21,17 @@
...
 
+/* Stay pinned at the bottom as the sidebar scrolls; `margin-top: auto`
+   pushes the footer to the bottom in the flex column when content is
+   shorter than the viewport. */
+.sidebarFooter {
+	margin-top: auto;
+	position: sticky;
+	bottom: 0;
+	z-index: 1;
+	background-color: var(--wpds-color-bg-surface-neutral-weak);
```
katinthehatsite, line 32: Should it be `--color-frame-*` instead of `--wpds-color-bg-surface-neutral-weak)`?

katinthehatsite (approved):
I think these changes look good: I tested them and they worked as expected. I left one minor suggestion

apps/ui/src/components/sidebar-layout/style.module.css
youknowriad: No, apps/ui is wrapped in ThemeProvider which provides these "wpds" variables that adapt to both light/dark mode and a "theme" color if we decide to support that at some point. So in apps/ui we should be using wpds variables instead.

PR #3768, opened 2026-06-10
Drive pull-reprint with a single reprint pull

epeicher:
## Related issues

Closes STU-1885

- Required: WordPress/reprint#268 (pull's `flatten_to` -> `flat_document_root` bridge plus the re-pull filter-guard fix that this flow depends on).
- Recommended: WordPress/reprint#264 (skipped-list ENOENT fix) in the bundled reprint.

## How AI was used in this PR

Claude Code measured the single-fork memory profile against a live WP.com Atomic site (26k files / ~570 MB), implemented the migration and tests, and verified the full flow end-to-end (fresh pull plus delta re-pull: remote edit -> re-pull -> local content / auto-login checks). Each change was reviewed and re-tested live.

## Proposed Changes

Re-running `studio pull-reprint` on a completed pull previously printed "pulled successfully" and exited without contacting the remote, so remote edits, inserts, and deletions never reached the local site. With this PR a re-run performs a **delta re-pull**: files re-sync incrementally and the database is refreshed so existing-row edits and deletions propagate.

Studio now drives reprint with a **single `reprint pull`** instead of orchestrating the pipeline itself. One call runs preflight -> files-pull -> db-pull -> db-apply -> flat-docroot -> apply-runtime in one PHP-WASM fork, and reprint resets its own state for a delta re-pull via `prepare_repull()`. This deletes the former Studio orchestration: the five per-sub-command functions, the nine-stage machine, and the `--abort` delta-reset wiring (about 340 lines net). The single fork's `memory_limit` is raised to 1024M because the file-index high-water-mark now persists across phases (measured peak ~510M on the test site).

Fixes that remain part of making re-pulls real:

- **Auto-login keeps working after a re-pull.** The database refresh rebuilds the DB from the remote dump, wiping the local `admin` user and `studio_admin_username` option that `/studio-auto-login` depends on. The pull re-applies the stored credentials over the running site's admin API, and a connection failure on that request doubles as a health check that catches a stale daemon "online" state and falls back to starting the server.
- **Database URLs can't drift to a dead port.** If the Studio site record was re-created with a different id/port, the re-pull adopts the live site record before the pull so the URL rewrite (`--new-site-url`) targets the right port.
- **Site starts no longer crash on stale runtime mounts.** Transient reprint state files mounted for the remote-uploads proxy are dropped when loading the persisted `start-options.json` (previously ENOENT crashed the start).

Bundle the reprint build via `REPRINT_VERSION` in `scripts/download-wp-server-files.ts`; it must include WordPress/reprint#268.

## Testing Instructions

1. Build the reprint phar from a branch that includes WordPress/reprint#268 (and #264): run `composer build:phar` on the reprint repo.
2. On this repo, place `reprint.phar` at `wp-files/reprint/reprint.phar` and run `npm run cli:build`.
3. Pull a site: `node apps/cli/dist/cli/main.mjs pull-reprint --verbose --url <site>` and let it complete. It should invoke a single `reprint pull` (visible with `--verbose`) and start a working site.
4. Edit an existing post on the remote site, then re-run the same command. It should report "Updating ... (delta sync)" and the local post should show the edit.
5. Open the local site's WP Admin via the printed auto-login URL and confirm the edit from step 4.
6. Stop and start the site (`studio site stop` / `site start`). No ENOENT mount errors.

## Pre-merge Checklist


apps/cli/commands/pull-reprint.ts
```diff
@@ -469,16 +545,39 @@ export async function runCommand(
...
+					logger.reportSuccess( __( 'WordPress server started' ) );
+
+					if ( processDesc.status === 'online' ) {
+						await updateSiteLatestCliPid( site.id, processDesc.pid );
+					}
+					await updateSiteAutoStart( site.id, true );
+					studioMetadata.localUrl = getSiteUrl( site );
+					savePullMetadata( studioMetadata );
+					recordCompletedStage( studioMetadata, 'site-started' );
 				}
```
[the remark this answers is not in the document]
epeicher: Addressed as part of [46f1059](https://github.com/Automattic/studio/pull/3768/commits/46f10596bcc6a7e12067ffbcb84487c51f21cc2a)

apps/cli/commands/pull-reprint.ts
```diff
@@ -271,10 +296,33 @@ export async function runCommand(
...
+	// A previously completed pull re-runs as a delta sync: reset the
+	// stage machine so every phase executes again. Reprint does the
+	// incremental work against the preserved state directory — files
+	// re-sync as a delta (re-index + diff), the database is fully
+	// re-downloaded and re-applied (the dump is idempotent, so remote
+	// edits, inserts, and deletes all land locally).
+	const isRepull = studioMetadata.stage === 'completed';
+	if ( isRepull ) {
+		studioMetadata.stage = 'initialized';
+		studioMetadata.hasCompletedOnce = true;
```
adamziel, line 300: Would it be helpful if reprint exposed this information? Such as `php reprint.php import-metadata | jq '.hasCompletedOnce'`? It's already stored there so that would be less state and less opportunity for one of these values to go out of sync.

apps/cli/commands/pull-reprint.ts
```diff
@@ -271,10 +296,33 @@ export async function runCommand(
...
+		savePullMetadata( studioMetadata );
+
+		// Re-verify connectivity (and give the secret-rotation retry path
+		// a chance to run) instead of trusting the cached preflight from
+		// the original pull, which may be days old.
```
adamziel, line 305: Did a cached preflight come up in your testing? In what way did it cause trouble?

apps/cli/commands/pull-reprint.ts
```diff
@@ -383,13 +436,36 @@ export async function runCommand(
 				sourceSite.wpComToken = token;
 				secret = await rotateReprintSecret( matched.id, token.accessToken );
 			}
+			// Rotating the secret does not bump `reprint_exporter_enabled`, so
+			// re-open the gate explicitly; otherwise the retry hits the same
+			// closed window and ?reprint-api falls through to an HTML page.
+			await enableReprintExporter(
```
adamziel, line 431: good catch!

adamziel:
The code change mostly makes sense, thank you @epeicher! I'm only confused by one thing: If we're moving to reliance on `pull`, shouldn't we be able to remove all the manual tool calls, e.g. `applyDownloadedDatabase()`? `pull` covers all that ground in a way we don't need to manage in Studio.

apps/cli/commands/pull-reprint.ts
epeicher: Yes, I think that would be better as reprint "knows better" about the pull, and reducing the state to the minimum seems a good option.

apps/cli/commands/pull-reprint.ts
epeicher: I had errors for sites that were pulled some time before that were "missing multipart boundary" due to HTML being returned. The way to reproduce that is to pull a site, then wait for 60 minutes (or reduce the sliding window), and then pull again

epeicher:
> The code change mostly makes sense, thank you @epeicher! I'm only confused by one thing: If we're moving to reliance on `pull`, shouldn't we be able to remove all the manual tool calls, e.g. `applyDownloadedDatabase()`? `pull` covers all that ground in a way we don't need to manage in Studio.

As part of this PR, I have not moved to relying on `pull`, but to relying on `<sub-command> --abort` to clear the completed state first. I could iterate to rely completely on `pull` (I know we discussed that, but I kept the changes minimal), let me try that and update the PR.

epeicher:
The latest changes [0b27bdf](https://github.com/Automattic/studio/pull/3768/commits/0b27bdff5d7976fef1ffcce5d61d68d65f6a3748), use a composite `pull` command instead of using multiple sub-commands as suggested. After some testing, this is working fine and the code is greatly reduced 🙌 

The changes require https://github.com/WordPress/reprint/pull/268 to be applied and generate a `reprint.phar` from it.

apps/cli/commands/pull-reprint.ts
adamziel: Oh, so this is an issue with the rotating token we  use on WP.com. And it seems like that token is saved in preflight.json, which is why deleting that file helps. 👍

adamziel (approved):
I haven't tested it, but the code change looks sound 👍

epeicher:
After merging of https://github.com/WordPress/reprint/pull/268, these changes have been tested and work as expected. I have been able to pull **both** initial and incremental changes from a remote. Merging this.

CleanShot 2026-06-19 at 18 19 44@2x

PR #3771, opened 2026-06-10
Add eval env overrides: STUDIO_BASE_PORT and STUDIO_EVAL_SESSION_DIR

wesleyfantinel:
## Related issues

- Related to ongoing work on benchmarking and testing the Studio Code agent locally

## How AI was used in this PR

Implemented with Claude Code using subagent-driven development: each change was written TDD-first by an implementation agent, then independently verified by spec-compliance and code-quality review agents, with fixes applied from both. End-to-end behavior was validated by driving the real agent headlessly through parallel runs.

## Proposed Changes

Today the CLI assumes it is the only Studio instance on the machine: site servers always scan for free ports starting at 8881, and the headless eval-runner keeps its agent session in memory and discards it on exit. That makes it hard to run several CLI instances side by side (local automation, stress-testing, A/B-comparing agent behavior between builds) and impossible to inspect what the agent actually did after a headless run.

This PR adds two small, opt-in env overrides that make the CLI flexible for those workflows without changing any default behavior:

- `STUDIO_BASE_PORT` — lets a caller start the port scan from a chosen base so concurrent CLI instances can each work in their own port range instead of racing for 8881. Unset (the desktop app and normal CLI usage), the behavior is exactly as before; zero/negative/invalid values safely fall back to 8881.
- `STUDIO_EVAL_SESSION_DIR` — makes the headless eval-runner persist the agent session transcript (JSONL) to the given directory, so a run can be reviewed or replayed afterwards. The eval-runner result now also reports the resolved model id, so callers can detect a silent fallback to the default model.

No user-visible behavior changes: both variables are unset in the desktop app and normal CLI flows.

## Testing Instructions

- `npm test -- tools/common/lib/tests/port-finder.test.ts` — covers the port override incl. zero/negative/invalid fallbacks.
- `npm test -- apps/cli/ai/tests` — existing runner suite (133 tests) still green.
- End-to-end: `npm run cli:build`, then `STUDIO_EVAL_SESSION_DIR=/tmp/s node apps/cli/dist/cli/eval-runner.mjs "Reply with the single word: pong. Do not use any tools."` → a `*.jsonl` transcript appears under `/tmp/s` and the result JSON includes `"model"`.
- Parallel check: run two eval-runner instances with different `STUDIO_BASE_PORT` bands (e.g. 20000 and 20100) and confirm both create and serve sites without port conflicts.

## Pre-merge Checklist


lezama (approved):
Looks good!

My bots found these minor observations:

  1. Unnecessary await on a synchronous call. SessionManager.create(cwd, sessionDir?,
  options?) returns SessionManager, not a Promise (session-manager.d.ts:294). await
  session = await SessionManager.create(...) works (await on a non-thenable just yields
  the value) but is misleading — it reads as if create is async. Drop the await, or if
  symmetry with the inMemory branch is wanted, leave a comment. Trivial.
  2. Duplicated ?? DEFAULT_MODEL fallback lives in both eval-runner.ts and pi/index.ts.
  They're consistent today, but they can silently drift. Not worth a refactor for one
  line — just flagging the coupling.

PR #3773, opened 2026-06-11
Add bundler block to Dependabot config and route Ruby reviews

mokagio:
Updates Dependabot to check Ruby dependencies daily so we can more easily address security updates. See [AINFRA-2437](https://linear.app/a8c/issue/AINFRA-2437).

Also routes review of the Apps Infra surface (dependency manifests, CI config, toolchain pins) to @Automattic/apps-infra-tooling via CODEOWNERS — GitHub retired the dependabot.yml `reviewers` key, and CODEOWNERS is its designated replacement. The routing is path-based, so it applies to any PR touching those files, not just Dependabot's; deliberate, since no source-scoped alternative exists.

---

Posted by Claude Code (Fable 5) on behalf of @mokagio.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,3 +13,6 @@ fastlane/             @Automattic/apps-infra-tooling
 .ruby-version         @Automattic/apps-infra-tooling
 .xcode-version        @Automattic/apps-infra-tooling
 .nvmrc                @Automattic/apps-infra-tooling
+
+# Secrets
+.configure-files/      @Automattic/apps-infra-tooling
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,15 @@
+# Ruby dependency surface
+Gemfile*              @Automattic/apps-infra-tooling
+.bundle/              @Automattic/apps-infra-tooling
+.rubocop*.yml         @Automattic/apps-infra-tooling
+fastlane/             @Automattic/apps-infra-tooling
+
+# CI and automation
+.buildkite/           @Automattic/apps-infra-tooling
+.github/workflows/    @Automattic/apps-infra-tooling
+.github/dependabot.yml @Automattic/apps-infra-tooling
+
+# Toolchain and environment pins
+.ruby-version         @Automattic/apps-infra-tooling
+.xcode-version        @Automattic/apps-infra-tooling
+.nvmrc                @Automattic/apps-infra-tooling
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -283,3 +283,13 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  # Enable version updates for Ruby gems
+  - package-ecosystem: "bundler"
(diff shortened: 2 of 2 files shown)


.github/dependabot.yml
```diff
@@ -283,3 +283,13 @@ updates:
...
+
+  # Enable version updates for Ruby gems
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    groups:
+      ruby-minor-and-patch:
+        update-types: [minor, patch]
+    open-pull-requests-limit: 5
```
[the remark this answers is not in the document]
mokagio: `update-types` without `patterns` is valid — when `patterns` is omitted the group applies to all dependencies in the ecosystem. Empirically: this exact group config has been live on 8 sibling repos since 2026-05-27 and Dependabot has been opening grouped `ruby-minor-and-patch` PRs there (e.g. pocket-casts-android#5356), so the config parses and groups as intended.

*Posted by Claude Code (Fable 5) on behalf of @mokagio with approval.*

.github/CODEOWNERS
```diff
@@ -0,0 +1,15 @@
+# Ruby dependency surface
+Gemfile*              @Automattic/apps-infra-tooling
+.bundle/              @Automattic/apps-infra-tooling
+.rubocop*.yml         @Automattic/apps-infra-tooling
+fastlane/             @Automattic/apps-infra-tooling
+
+# CI and automation
+.buildkite/           @Automattic/apps-infra-tooling
+.github/workflows/    @Automattic/apps-infra-tooling
+.github/dependabot.yml @Automattic/apps-infra-tooling
```
wojtekn, line 10: Should we add `.configure-files` as well?

It's also worth noting that some build-related scripts also live in `scripts/`.

wojtekn (approved):

.github/CODEOWNERS
mokagio: > Should we add .configure-files as well?

Thanks @wojtekn ! That one had escaped me.

> It's also worth noting that some build-related scripts also live in scripts/.

Unsure about this. I'd say for this repo, the product teams owns most of the build process. Or, at least, we share ownership.

Looking ahead, Apps Infra will need to level up the JS/TS/Electron support, but we are not there yet. So for the moment, I'd leave it out. What do you think?

mokagio: `.configure-files/` addressed in 2fbecbb1b4dfd846a82a7b5c486f0928a0efdbb0

.github/dependabot.yml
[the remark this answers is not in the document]
mokagio: Case in point: https://github.com/Automattic/studio/pull/3802

.github/CODEOWNERS
wojtekn: > Looking ahead, Apps Infra will need to level up the JS/TS/Electron support, but we are not there yet. So for the moment, I'd leave it out. What do you think?

Makes sense!

PR #3776, opened 2026-06-11
Use `Intl.Locale` over `@formatjs/intl-locale`

fredrikekelund:
## Related issues

- N/A

## How AI was used in this PR

To verify that the `Intl.Locale` constructor throws in the same way that the `Locale` constructor from `@formatjs/intl-locale` does.

## Proposed Changes

- Uninstall the `@formatjs/intl-locale` npm module and replace our (very simple) use of the `Locale` constructor with the standard library `Intl.Locale` constructor ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/Locale)). `@formatjs/intl-locale` is indeed a polyfill for `Intl.Locale` ([ref](https://formatjs.github.io/docs/polyfills/intl-locale/)).
- Before this change, the `@formatjs` namespace took up 323kB in the CLI JS bundle. After this change, only `@formatjs/intl-localematcher` remains, which takes up 46kB.
- I also took the opportunity to explicitly list `@formatjs/intl-localematcher` as a dependency in `apps/cli` (and to update that dependency).

## Testing Instructions

CI should pass.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/lib/site-language.ts
+++ b/apps/cli/lib/site-language.ts
@@ -1,6 +1,5 @@
 import fs from 'fs';
 import path from 'path';
-import { Locale } from '@formatjs/intl-locale';
 import { match } from '@formatjs/intl-localematcher';
 import { DEFAULT_LOCALE } from '@studio/common/lib/locale';
 import { getWpFilesPath } from 'cli/lib/dependency-management/paths';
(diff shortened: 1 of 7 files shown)


bcotrim (approved):
LGTM 👍

PR #3780, opened 2026-06-11
Add image attachments to the agentic chat composer and conversation

youknowriad:
## Related issues

- Related to #3646 (extracts the image-attachment slice so it can land independently of the UI overhaul)

## How AI was used in this PR

Claude Code extracted the image-attachment portions of #3646 onto a fresh branch from trunk, leaving out the unrelated interface changes (preview split frame, sidebar rework, composer placeholder animation, tooltips, render-scoping refactor). It then reworked the transcript-rendering approach to use persisted thumbnails (see below). All code was reviewed hunk-by-hunk; lint, typecheck, and tests were run locally.

## Proposed Changes

Users can now attach images (PNG/JPEG) when chatting with the agent in the new UI:

- Attach via the composer's image button (file picker), drag-and-drop onto the composer, or paste from the clipboard.
- Attached images show as removable preview chips above the input; sending with images and no text auto-fills a "Please review the attached image(s)." prompt.
- The conversation renders the user's attached images with the matching user turn, both for live optimistic sends and for transcripts reloaded from disk.
- Client-side validation enforces the existing limits (4 images, 5 MB each, 12 MB total) plus a new 8000px per-side dimension cap, with inline error feedback in the composer.
- The agent run cleanup in the main process is hardened so the temp directory holding attached image payloads is removed eagerly (and exactly once), including when the CLI subprocess fails to start — avoiding leaked user images on disk.

**Transcript rendering uses persisted thumbnails.** Image bytes live in two places in a session: the model's user message (what the agent actually receives — non-negotiable) and the `studio.user_prompt` display entry's attachment summaries (what the transcript renders). Previously the CLI inlined the *full-size* image into the display entry's `previewDataUrl`, doubling each image's on-disk footprint (~13.6 MB of JSONL per 5 MB image); #3646's renderer instead avoided duplication by scanning forward to the model message and stitching bytes back to metadata. This PR replaces both: the composer downscales each image to a ≤256px thumbnail at attach time, the CLI persists that thumbnail on the display entry (falling back to the old full-size inlining only when no thumbnail is provided, e.g. terminal CLI use), and the conversation renders chips purely from the self-contained display entry. Validation caps the thumbnail at 256 KB so a full-size image can't masquerade as a preview. The legacy (feature-flagged) Studio Code composer adopts the same convention via a shared `chat-image-preview` helper, so its sessions also stop persisting full-size duplicates; it additionally now records image dimensions, giving the 8000px cap effect on that path too.

The IPC/CLI transport for images already exists on trunk; this PR adds the new-UI front end and plumbing on top of it.

## Testing Instructions

1. Run `npm start` with the new UI enabled and open a site's agent session.
2. Attach an image via the picker button, drag-and-drop, and paste — verify chips appear, can be removed, and errors show for >4 images, >5 MB files, or non-PNG/JPEG files.
3. Send a message with images attached — verify the images render in the conversation immediately and persist after reloading the session.
4. Inspect the session JSONL (`~/.studio` sessions dir): the `studio.user_prompt` entry's attachment should carry a small `data:` thumbnail, not the full image.
5. Send images with an empty input — verify the auto-filled "Please review the attached image(s)." prompt.
6. Verify both light and dark color schemes.
7. `npm test -- tools/common/ai/tests apps/ui/src/data/queries/use-agent-run.test.tsx`

## Pre-merge Checklist


youknowriad:
Closing — we're going to take a different approach and start this from scratch.

PR #3781, opened 2026-06-11
Stop writing cliAutoInstalled, deprecate in favour of cliUserUninstalled

wojtekn:
## Related issues

- Follow-up to #3713

## How AI was used in this PR

Developed with Claude Code assistance.

## Proposed Changes

Now that `cliUserUninstalled` exists as an explicit signal, `cliAutoInstalled` no longer serves an ongoing purpose. Its original role — preventing re-installation after first launch — is fully handled by `cliUserUninstalled` (absent = not uninstalled).

The only remaining use is a one-time migration: on first launch of the new version, `cliAutoInstalled: true` + absent symlink/directory is treated as a deliberate uninstall and sets `cliUserUninstalled: true`. After that migration runs once per user, `cliAutoInstalled` is never meaningfully consulted again.

This PR stops writing `cliAutoInstalled` on new installs and marks it `@deprecated` in the storage type. The migration block still reads it, so existing users who uninstalled before #3713 shipped are correctly preserved. The flag and migration block can be removed entirely in a follow-up release.

On Windows, this also fixes `updateWindowsCliVersionedPathIfNeeded()` to run on every launch where the CLI is installed (not only when `cliAutoInstalled` was set), keeping the proxy `.bat` path up to date after app updates.

## Testing Instructions

- Launch Studio with `cliAutoInstalled: true` + symlink absent in `app.json` → `cliUserUninstalled` should be set to `true`, CLI not reinstalled
- Launch Studio with `cliAutoInstalled: true` + symlink present but stale → symlink re-pointed to current app
- Fresh install (no flags set) → CLI installs as normal, `cliAutoInstalled` is NOT written

## Pre-merge Checklist


apps/studio/src/modules/cli/lib/windows-installation-manager.ts
```diff
@@ -44,23 +44,21 @@ export class WindowsCliInstallationManager implements StudioCliInstallationManag
...
-				await updateAppdata( { cliUserUninstalled: true } );
-				return;
-			}
+		// Migration: before cliUserUninstalled existed, an absent CLI dir with cliAutoInstalled
+		// set was the only signal that the user had uninstalled the CLI via Settings. Preserve
+		// that intent rather than silently reinstalling on the first launch of this version.
+		if ( userData.cliAutoInstalled && ! ( await this.isCliInstalled() ) ) {
+			await updateAppdata( { cliUserUninstalled: true } );
+			return;
+		}
```
fredrikekelund: Let's put this in a `apps/studio/src/migrations` migration instead.

fredrikekelund (commented):
Nice to see it was such a simple change. We have an explicit framework for migration in `apps/studio/src/migrations`, though, so I advise we put the migration code there.

fredrikekelund (approved):
I tweaked the migration logic so we rely on the canonical `isStudioCliInstalled` function instead of a custom reimplementation. I've tested this manually and reviewed with Claude locally. LGTM 👍

wojtekn:
Thanks for tweaking that @fredrikekelund .

I've merged trunk to the branch and solved the conflict.

PR #3784, opened 2026-06-11
Remove `postinstall` script in `apps/cli` by using new archiver strategy

fredrikekelund:
## Related issues

- Fixes STU-1375

## How AI was used in this PR

Claude was used to iteratively reach the desired state. I bounced around ideas for achieving the goals, iterated on ways to simplify the implementation, and relied on it heavily to simplify tests.

## Proposed Changes

npm v12 will [stop running](https://github.blog/changelog/2026-06-09-upcoming-breaking-changes-for-npm-v12/) `postinstall` scripts by default. The `wp-studio` package currently runs a `postinstall` script to patch the `archiver` npm module using `patch-package`. 

This PR removes the need for that patch by eliminating the use of `archiver.directory()` and instead doing the globbing ourselves and calling `archiver.file()` for each file in our list. The PR also removes the `patch-package` dependency and the actual patch.

Moreover, I took the opportunity to update `archiver` to v8, which is the latest version.

## Testing Instructions

1. Ensure that creating preview sites works as expected
2. Ensure that exporting a site works as expected

## Pre-merge Checklist


bcotrim (approved):
LGTM 👍 

Note:
- An exported site is missing empty folders, unlike trunk.
I don't think it will have any impact, but wanted to flag to raise the flag.
- I fixed unit tests and merge conflicts, feel free to adjust as you see necessary

epeicher (commented):
Thank you very much, @fredrikekelund. Removing the `postinstall` will be a great step! I have tested it, and I have found a couple of issues that I think should be fixed:
1. The preview site does not include the symlink to a single file
2. When exporting the site and importing it as a new site, I am getting the error in the screenshot below

Original site|Preview site | Site created from the Exported site|
---|-------|------|
Image| Image| Image|


Please let me know if you have any questions. In a nutshell, the symlink to folders seems to be working fine, but the symlink to single files does not seem to be working for me.

fredrikekelund:
Kudos for the thorough testing, @epeicher 🙏 I pushed a fix for the file symlink issue. I'll run some more testing tomorrow morning before landing this. It looks like the "Site created from the Exported site" issue you found also came from the file symlink thing

fredrikekelund:
Confirmed that file symlinks are included properly in both types of exports with this new approach ✅

fredrikekelund:
AI review raised a real potential issue: `fs.createReadStream` opens a file descriptor that isn't closed until the archive is finalized. For a large site directory, this means that we could get `EMFILE` errors if the system thinks we're holding too many file descriptors. To solve for this, I changed the strategy from using `Archiver.append()` to using `Archiver.file()` again, but calling realpath on the source path first. 

I've confirmed that the new approach works as expected, too ✅

apps/cli/__mocks__/fs.cjs
fredrikekelund, line 1: This change wasn't crucial to the core purpose of this PR, but it's how Vitest [recommends](https://main.vitest.dev/guide/mocking/file-system) mocking the `fs` module. We don't need to reimplement what `memfs` exports. The point of that module is to provide the same API as `fs`, just backed by a temporary file system.

apps/cli/lib/tests/validation.test.ts
fredrikekelund, line 1: These changes are unrelated to the core purpose of this PR – they're just simplifications I was able to make from tweaking the `fs` mock.

apps/cli/package.json
```diff
@@ -25,31 +23,31 @@
...
-		"@wp-playground/cli": "3.1.36",
-		"@wp-playground/common": "3.1.36",
-		"@wp-playground/storage": "3.1.36",
-		"@wp-playground/wordpress": "3.1.36",
-		"archiver": "^7.0.1",
+		"@wp-playground/blueprints": "3.1.38",
+		"@wp-playground/cli": "3.1.38",
+		"@wp-playground/common": "3.1.38",
+		"@wp-playground/storage": "3.1.38",
+		"@wp-playground/wordpress": "3.1.38",
```
fredrikekelund, line 41: These version changes are unrelated to the core purpose of this PR. https://github.com/Automattic/studio/pull/3829 mistakenly downgraded the Playground dependencies in `apps/cli` to `3.1.36`, so I took the opportunity to reset them back to `3.1.38`.

This is also what caused the change in installer size. Previously, we had different Playground dependency trees: one for `3.1.36` and another for `3.1.38`

PR #3786, opened 2026-06-11
Site PHP runtime & File Access options

bcotrim:
## Related issues

- Fixes RSM-3958

## How AI was used in this PR

Implemented end-to-end with Claude Code, starting from the design agreed in Slack (two per-site settings instead of a 3-option dropdown). I directed the key decisions — per-site `runtime`/`fileAccess`, native PHP as the default, narrowing the PHP version list, and the usage metric — and reviewed the diff. Tests were written/updated by AI and verified locally (typecheck, lint, full suite, built-CLI smoke tests).

## Proposed Changes

Adds **two per-site settings** that control how a site's PHP runs, replacing the old global toggle (and avoiding "safe/unsafe/trusted" wording):

- **PHP Runtime** — `Native` (native PHP binaries) or `Sandbox` (WordPress Playground). **Native is the default for both new and existing sites** — existing sites switch to Native on their next start (clamping any unsupported stored PHP version to the closest supported one); a site explicitly set to Sandbox keeps it.
- **File access** — `Site directory` (default: `open_basedir` jail + `disable_functions`) or `All files` (neither restriction). The invalid "Sandbox + All files" combination is disabled in the UI with an explainer and rejected by the CLI with a clear error.

Both settings are selectable when creating a site and when editing one, shown read-only in the site settings tab (with explainer tooltips), exposed via the CLI (`studio site create|set --runtime <native|sandbox> --file-access <site-directory|all-files>`), and reported by `studio site status`. PHP version choices are unified to **8.2–8.5** for both runtimes. Adoption is tracked with a privacy-preserving daily count of active sites by runtime (no per-site identifiers), recorded in the CLI at site start so it covers every start path — desktop-spawned and terminal alike.

> **On the security promise:** `Site directory` is a **blast-radius guardrail, not a sandbox.** It contains the common WP exploit class (LFI, path traversal, arbitrary file read/delete via PHP file functions), but the exec family stays enabled so deliberately malicious code can still shell out past it — see the [rationale in `config.ts`](https://github.com/Automattic/studio/blob/trunk/apps/cli/lib/native-php/config.ts). The setting governs the served site; WP-CLI is unrestricted by design (trusted, user-initiated). Real confinement would need an OS-level sandbox, which is out of scope here.

| Add site | Edit Site |
|--------|--------|
| image | image | 

⚠️ **Reviewer callout:** PHP 7.4 / 8.0 / 8.1 are dropped from the pickers (all past security EOL). Existing sites stored on those versions keep *running* in Playground — only the pickers narrow; the Edit Site modal pre-snaps to the closest supported version with a "PHP X is no longer supported" warning, and native runtime clamps to the closest supported version at server start.

## Testing Instructions

**Desktop**
1. **Add site** → Advanced settings shows **PHP Runtime** (defaulting to Native) and **File access** dropdowns; **File access** is forced to "Site directory" and disabled while PHP Runtime is Sandbox. Create a site, then check Settings → both values appear as read-only rows with explainer tooltips.
2. **Edit site** → General: switch PHP Runtime / File access and save → the site restarts with the new settings (native PHP binary downloads on demand the first time).
3. **Copy site** keeps the source site's runtime and file access.
4. PHP version select offers 8.2–8.5; a site stored on an older version shows the warning icon and pre-selects the closest supported version.
5. Verify the new UI in **light and dark** color schemes.

**CLI** (`npm run cli:build` first)
- `studio site create --path ~/Studio/test --file-access all-files --no-start` → new site is Native + All files (Native is the default).
- `studio site set --path ~/Studio/test --runtime sandbox` → errors asking to reset file access; add `--file-access site-directory` → succeeds and restarts if running.
- `studio site status --path ~/Studio/test` → shows **PHP runtime** and **File access** rows; `--format json` includes `runtime` / `fileAccess`.
- `--php` choices show `8.5, 8.4, 8.3, 8.2` on `site create`, `site set`, and `blueprint use`.

**Verifying file-access enforcement** — `open_basedir`/`disable_functions` apply to the PHP **server workers**, *not* WP-CLI (intentionally exempt: trusted, user-initiated, and it needs paths outside the site for imports/exports). So probe over HTTP rather than via `studio wp eval`:

1. Create a native site and drop a probe script in its root:
   ```bash
   studio site create --path ~/Studio/test --skip-browser
   cat > ~/Studio/test/probe.php <<'PHP'
   <?php
   header( 'Content-Type: application/json' );
   echo json_encode( [
     'open_basedir'      => ini_get( 'open_basedir' ),
     'disable_functions' => ini_get( 'disable_functions' ),
     'can_read_outside'  => @file_get_contents( '/etc/hosts' ) !== false,
   ] );
   PHP
   ```
2. Probe the default (native + site directory):
   ```bash
   URL=$(studio site status --path ~/Studio/test --format json | jq -r .siteUrl)
   curl -s "${URL}probe.php" | jq
   ```
   → `open_basedir` jailed to the site dir (+ router/phpMyAdmin/tmp), `disable_functions` populated, `can_read_outside: false`.
3. Switch to all files and re-probe → both empty, `can_read_outside: true`:
   ```bash
   studio site set --path ~/Studio/test --file-access all-files
   curl -s "${URL}probe.php" | jq
   ```
4. Switch to the sandbox and re-probe → both empty (Playground doesn't use those directives) but `can_read_outside: false`; the WASM filesystem still confines PHP to the site directory. Clean up with `rm ~/Studio/test/probe.php`:
   ```bash
   studio site set --path ~/Studio/test --runtime sandbox --file-access site-directory
   curl -s "${URL}probe.php" | jq
   ```

## Pre-merge Checklist

---


apps/studio/src/modules/add-site/components/create-site-form.tsx
```diff
@@ -517,6 +528,58 @@ export const CreateSiteForm = ( {
...
+											onChange={ ( value ) => setSelectedFileAccess( value ) }
+											__next40pxDefaultSize
+											__nextHasNoMarginBottom
+										/>
+										<span className="text-frame-text-secondary text-xs">
+											{ selectedRuntime === SITE_RUNTIME_PLAYGROUND
+												? __( 'The sandbox can only access the site directory.' )
+												: usedFileAccess === SITE_FILE_ACCESS_ALL_FILES
+												? __( 'PHP can access any file your user account can access.' )
+												: __( "Restricts the site's file access to the site directory." ) }
```
fredrikekelund: Let's avoid the nested ternary

apps/studio/src/modules/add-site/components/create-site-form.tsx
```diff
@@ -517,6 +528,58 @@ export const CreateSiteForm = ( {
...
+											] }
+											onChange={ ( value ) => setSelectedFileAccess( value ) }
+											__next40pxDefaultSize
+											__nextHasNoMarginBottom
+										/>
+										<span className="text-frame-text-secondary text-xs">
+											{ selectedRuntime === SITE_RUNTIME_PLAYGROUND
+												? __( 'The sandbox can only access the site directory.' )
+												: usedFileAccess === SITE_FILE_ACCESS_ALL_FILES
+												? __( 'PHP can access any file your user account can access.' )
```
fredrikekelund:
```suggestion
												? __( 'PHP can access any file on your system.' )
```
The existing label is not incorrect – I was just a little bit confused by the wording at first. I understand that "your user account" refers to my operating system user account, but it seems like a less discussed concept these days… 

On the other hand, "your system" could arguably mean "the system accessible to your user account". Above all, I think it's more immediately understood.

apps/studio/src/modules/add-site/components/create-site-form.tsx
```diff
@@ -517,6 +528,58 @@ export const CreateSiteForm = ( {
 									/>
 								</div>
 
+								<div className="grid grid-cols-2 gap-4 mt-4">
+									<div className="flex flex-col gap-1.5 leading-4">
+										<label className="font-semibold" htmlFor="php-runtime-select">
+											{ __( 'PHP runtime' ) }
+										</label>
```
fredrikekelund, line 533: We should have a `Learn more` link somewhere in here. That can be a follow-up, though. I added RSM-4363 to track

apps/studio/src/lib/site-runtime-stats.ts
```diff
@@ -0,0 +1,38 @@
...
+			shouldBump = true;
+		}
+	} finally {
+		await unlockAppdata();
+	}
+
+	if ( shouldBump ) {
+		bumpStat( StatsGroup.STUDIO_SITE_RUNTIME_WEEKLY, getSiteRuntimeStat( site ) );
+	}
+}
```
fredrikekelund: This fails to capture cases where the user switches the runtime. Let's say they pick "Native" to start with, then switch to "Sandbox" after 5 mins. The "Sandbox" stat would not be captured until after 7 days. 

The perfect solution to this scenario would be for the "Sandbox" metric to override the original "Native" metric on the remote. I don't see a practical way of achieving that, but what we're looking to understand is how many users make a _persistent_ choice to use the non-default.

If the stats show us that 5% of users are running "Sandbox" sites, but in reality, those 5% just tested "Sandbox" and immediately switched back to "Native", then the stats are misleading.

apps/studio/src/lib/site-runtime-stats.ts
fredrikekelund, line 1: Would it not be better to track this at the CLI level instead of in the app?

apps/studio/src/lib/beta-features.ts
fredrikekelund, line 1: I think we might as well remove the beta feature now and rely on the individual site controls entirely. This also entails defaulting to "Native" and "Site directory"

apps/cli/commands/wp.ts
```diff
@@ -96,11 +99,11 @@ export async function runCommand(
...
 	options: { phpVersion?: string } = {}
 ): Promise< void > {
-	const runtime = getSiteRuntime();
-
 	// Handle global WP-CLI commands that don't require a site path (--studio-no-path)
 	if ( mode === Mode.GLOBAL ) {
-		await using command = await runGlobalWpCliCommand( args, { runtime } );
+		await using command = await runGlobalWpCliCommand( args, {
+			runtime: SITE_RUNTIME_PLAYGROUND,
+		} );
```
fredrikekelund: I'd consider defaulting to native PHP here for the performance improvements. Side note: are we even using the `--studio-no-path` option anywhere these days? It doesn't look like it, and I can't remember exactly what we used it for before.

fredrikekelund (commented):
I need to do some more testing here, but leaving a review with an initial round of feedback. Overall, this is looking really good, but the two bigger questions I think we should still tackle are:

- Improve the metrics tracking
- Remove the beta feature

apps/studio/src/lib/site-runtime-stats.ts
bcotrim: That makes sense.
Here is my new approach:
1. Changed from weekly to daily - increase volume and responsiveness
2. Records unique daily + runtime stats - a site can have both runtime in a given day

These metrics should let us know the amount of sites that run native/sandbox in a given day. It won't track how much time it ran on a given runtime, but over time the counts will be dominated by whatever runtime each site actually runs most, so experiments average out.

apps/studio/src/lib/site-runtime-stats.ts
bcotrim: How do you see that working with `--avoid-telemetry`? The app runs the CLI with it for everything, so CLI-side tracking would be suppressed for app-driven starts unless we ignore the opt-out.

I'd say adding matching CLI-side tracking makes sense, but it could be done in a follow-up. 
What do you think?

fredrikekelund: > unless we ignore the opt-out

The `--avoid-telemetry` option isn't included in the `--help` output, so there's no implicit contract with users that this option exists and should suppress all telemetry. 

In fact, the way `--avoid-telemetry` is used now is more about distinguishing when the CLI was spawned by a user directly from the terminal versus when it's used as the back-end for the app.

> adding matching CLI-side tracking makes sense, but it could be done in a follow-up

Fine by me 👍

apps/studio/src/lib/site-runtime-stats.ts
fredrikekelund: Sounds like a sensible approach 👍

tools/common/lib/site-runtime.ts
```diff
@@ -5,3 +5,22 @@ export const SITE_RUNTIME_NATIVE_PHP = 'native-php';
 
 export const siteRuntimeSchema = z.enum( [ SITE_RUNTIME_PLAYGROUND, SITE_RUNTIME_NATIVE_PHP ] );
 export type SiteRuntime = z.infer< typeof siteRuntimeSchema >;
+
+export function getSiteRuntime( site: { runtime?: SiteRuntime } ): SiteRuntime {
+	return site.runtime ?? SITE_RUNTIME_PLAYGROUND;
+}
```
fredrikekelund, line 11: @bcotrim and I discussed this approach and concluded that we should stick with the previously stated launch plan, which is to default existing sites to native PHP (see pfHvTO-1r1-p2)

apps/studio/src/components/content-tab-settings.tsx
```diff
@@ -202,6 +202,18 @@ export function ContentTabSettings( { selectedSite }: ContentTabSettingsProps )
...
+						<span>{ isNativePhpRuntime ? __( 'Native' ) : __( 'Sandbox' ) }</span>
+					</SettingsRow>
+					<SettingsRow label={ __( 'File access' ) }>
+						{ /* translators: value for the File access setting on the site settings screen */ }
+						<span>
+							{ getSiteFileAccess( selectedSite ) === SITE_FILE_ACCESS_ALL_FILES
+								? __( 'All files' )
+								: __( 'Site directory' ) }
+						</span>
+					</SettingsRow>
```
fredrikekelund, line 252: I would consider adding the same explainer texts here that we have in `EditSiteDetails`. Putting it in a tooltip might be the best strategy, design-wise. 

The reason I think this would be nice is that this is the first place where this new setting is visible for existing sites. We want to give users a clue about what this is without necessarily needing to open the `Edit site` dialog.

apps/studio/src/site-server.ts
```diff
@@ -179,7 +181,12 @@ export class SiteServer {
 		};
 		const server = SiteServer.register( placeholderDetails, meta );
 
-		const result = await createSiteViaCli( { ...options, siteId } );
+		// New sites default to the native PHP runtime; existing sites keep
+		// whatever they have (sandbox when unset, via getSiteRuntime).
+		const runtime = options.runtime ?? SITE_RUNTIME_NATIVE_PHP;
```
fredrikekelund: Just calling out that we need to change this spot, too

apps/cli/commands/site/create.ts
```diff
@@ -520,8 +539,24 @@ export const registerCommand = ( yargs: StudioArgv ) => {
...
+					defaultDescription: RecommendedPHPVersion,
+				} )
+				.option( 'runtime', {
+					type: 'string',
+					describe: __(
+						'Run the site with native PHP ("native") or in the Playground sandbox ("sandbox")'
+					),
+					choices: [ SITE_MODE_NATIVE, SITE_MODE_SANDBOX ],
+					defaultDescription: SITE_MODE_SANDBOX,
+				} )
```
fredrikekelund, line 553:
```suggestion
				.option( 'runtime', {
					type: 'string',
					describe: __(
						'Run the site with native PHP ("native") or in the Playground sandbox ("sandbox")'
					),
					choices: [ SITE_MODE_NATIVE, SITE_MODE_SANDBOX ] as const,
					default: SITE_MODE_SANDBOX,
				} )
```
I think this behavior is already the intent here, but we aren't actually getting the type guarantees because of a few missing tweaks. We need this and a small tweak in `tools/common/lib/site-runtime.ts` – see my comment there

apps/cli/commands/site/create.ts
```diff
@@ -520,8 +539,24 @@ export const registerCommand = ( yargs: StudioArgv ) => {
...
+				.option( 'file-access', {
+					type: 'string',
+					describe: __(
+						'Which files PHP can access with the native PHP runtime: the site directory only, or all files'
+					),
+					choices: [ SITE_FILE_ACCESS_SITE_DIRECTORY, SITE_FILE_ACCESS_ALL_FILES ],
+					defaultDescription: SITE_FILE_ACCESS_SITE_DIRECTORY,
 				} )
```
fredrikekelund, line 561:
```suggestion
				.option( 'file-access', {
					type: 'string',
					describe: __(
						'Which files PHP can access with the native PHP runtime: the site directory only, or all files'
					),
					choices: [ SITE_FILE_ACCESS_SITE_DIRECTORY, SITE_FILE_ACCESS_ALL_FILES ] as const,
					default: SITE_FILE_ACCESS_SITE_DIRECTORY,
				} )
```
Same thing as for `runtime`. Again, small tweak in `tools/common/lib/site-file-access.ts` needed

apps/cli/commands/site/create.ts
```diff
@@ -580,8 +615,23 @@ export const registerCommand = ( yargs: StudioArgv ) => {
...
-			const supportedPhpVersions = getSupportedPhpVersionsForSiteRuntime();
-			const recommendedPhpVersion = getRecommendedPhpVersionForSiteRuntime();
+			const runtime = argv.runtime
+				? siteRuntimeFromMode( siteModeSchema.parse( argv.runtime ) )
+				: undefined;
+			const fileAccess = siteFileAccessSchema.optional().parse( argv.fileAccess );
+			if (
+				fileAccess &&
+				! isFileAccessAllowedForRuntime( runtime ?? SITE_RUNTIME_PLAYGROUND, fileAccess )
+			) {
```
fredrikekelund:
```suggestion
			const runtime = siteRuntimeFromMode( argv.runtime );
			const fileAccess = argv.fileAccess;
			if ( ! isFileAccessAllowedForRuntime( runtime, fileAccess ) ) {
```
With better yargs typings, we can simplify this logic.

tools/common/lib/site-runtime.ts
```diff
@@ -5,3 +5,22 @@ export const SITE_RUNTIME_NATIVE_PHP = 'native-php';
...
+
+// User-facing short names for the runtimes, used by the CLI (--runtime) and the app UI.
+export const SITE_MODE_NATIVE = 'native';
+export const SITE_MODE_SANDBOX = 'sandbox';
```
fredrikekelund:
```suggestion
export const SITE_MODE_NATIVE = 'native' as const;
export const SITE_MODE_SANDBOX = 'sandbox' as const;
```
For improved yargs typings, mainly. Although this is also a good fit here in general

tools/common/lib/site-file-access.ts
```diff
@@ -0,0 +1,24 @@
+import { z } from 'zod';
+import { SITE_RUNTIME_NATIVE_PHP, type SiteRuntime } from '@studio/common/lib/site-runtime';
+
+export const SITE_FILE_ACCESS_SITE_DIRECTORY = 'site-directory';
+export const SITE_FILE_ACCESS_ALL_FILES = 'all-files';
```
fredrikekelund:
```suggestion
export const SITE_FILE_ACCESS_SITE_DIRECTORY = 'site-directory' as const;
export const SITE_FILE_ACCESS_ALL_FILES = 'all-files' as const;
```
Same as in `site-runtime.ts`, this is mainly for improved yargs typings.

apps/cli/commands/site/create.ts
```diff
@@ -87,7 +97,9 @@ export type CreateCommandOptions = {
 	name?: string;
 	siteId?: string;
 	wpVersion: string;
-	phpVersion: SupportedPHPVersion;
+	phpVersion: string;
+	runtime?: SiteRuntime;
+	fileAccess?: SiteFileAccess;
```
fredrikekelund:
```suggestion
	runtime: SiteRuntime;
	fileAccess: SiteFileAccess;
```
These don't need to be optional if we follow my other suggestions

apps/cli/commands/site/create.ts
```diff
@@ -106,8 +118,15 @@ export async function runCommand(
...
-	const phpVersion = validatePhpVersionForSiteRuntime( options.phpVersion );
-	const siteRuntime = getSiteRuntime();
+	const siteRuntime = options.runtime ?? SITE_RUNTIME_NATIVE_PHP;
+	if ( options.fileAccess && ! isFileAccessAllowedForRuntime( siteRuntime, options.fileAccess ) ) {
+		throw new LoggerError(
+			__(
+				'File access "all-files" requires the native PHP runtime. The sandbox only has access to the site directory.'
+			)
+		);
+	}
```
fredrikekelund:
```suggestion
	if ( ! isFileAccessAllowedForRuntime( options.runtime, options.fileAccess ) ) {
		throw new LoggerError(
			__(
				'File access "all-files" requires the native PHP runtime. The sandbox only has access to the site directory.'
			)
		);
	}
```
Again, we can make this simpler if `runtime` and `fileAccess` are required

apps/cli/commands/site/create.ts
```diff
@@ -87,7 +97,9 @@ export type CreateCommandOptions = {
 	name?: string;
 	siteId?: string;
 	wpVersion: string;
-	phpVersion: SupportedPHPVersion;
+	phpVersion: string;
```
fredrikekelund:
```suggestion
	phpVersion: SupportedPHPVersion;
```
This is actually still fine, AFAICT. It'd be good to keep the stricter type if there's no clear reason to drop it.

apps/studio/src/lib/site-runtime-stats.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Counts a site toward the daily active-sites-by-runtime stat. Deduped per site
+ * per day so successive restarts don't inflate the numbers, but re-counted when
+ * the day rolls over or the runtime/file-access choice changes (so a switch is
+ * captured on the next start rather than at the next day boundary). The per-site
+ * marker lives in `app.json`'s site metadata.
+ */
+export async function recordSiteRuntimeUsage( site: SiteRuntimeUsage ): Promise< void > {
```
fredrikekelund: This doesn't get called from the `updateSite` function in `apps/studio/src/ipc-handlers.ts`. Maybe consider dropping a `recordSiteRuntimeUsage()` call into `editSiteViaCli()`?

fredrikekelund (commented):
Fundamentally, this is solid 👍

The big remaining thing I see here is that we should switch the default runtime for existing sites from Playground to native, per our previous discussions on this topic.

Moreover, tracking the runtime stat when a site is edited from Studio is something we should fix.

I left a few more comments that would be nice to address, but they're only about simplifying the code and not about any behavioral change, so no big deal. 

I'll come back for a final approval once you've addressed the high-priority issues, @bcotrim.

apps/cli/lib/cli-config/core.ts
```diff
@@ -45,6 +45,10 @@ const cliConfigSchema = z.object( {
 	lastBumpStats: z
 		.record( z.string(), z.partialRecord( z.enum( StatsMetric ), z.number() ) )
 		.optional(),
+	// Per-site daily dedup markers for the runtime adoption stat (RSM-3958).
+	siteRuntimeStats: z
+		.record( z.string(), z.object( { bumpedAt: z.number(), stat: z.string() } ) )
+		.optional(),
```
fredrikekelund, line 51: Not a show stopper, but we should have defined `cliConfigSchema` with `z.looseObject` from the start 🤦‍♂️ Might as well do it now.

Theoretically, old CLI versions will strip out `siteRuntimeStats` when loading `cli.json`, but this should average out over time.

apps/studio/src/lib/site-runtime-copy.ts
fredrikekelund, line 1: I'd make this a React component instead and load `__` with the standard `useI18n()` hook. We can return a JSX fragment. 

The `Tooltip` component accepts a React component for the `text` prop, so this strategy should work.

fredrikekelund (approved):
Looks great 👍

I left two final comments that would be nice to address (though I'll leave them to your judgement, @bcotrim). I also took the liberty of pushing a small tweak.

PR #3788, opened 2026-06-11
Fetch every page of syncable sites and mark already-connected ones

shaunandrews:
## Related issues

- Part of the site-creation onboarding redesign — see the umbrella PR (linked below once open) for the full plan.

## How AI was used in this PR

Split out of the `workbench/site-creation` exploration branch with Claude Code; designed and tested by Shaun on the original branch.

## Proposed Changes

- Users with many WordPress.com sites were only seeing a subset in the site picker: the unpaginated v1.2 `/me/sites` request silently truncated results. The fetch now drains every page via the v1.3 endpoint, so the full account is always available.
- Sites that are already connected to a local site are now marked as such, so pickers can show an "already connected" state instead of offering them again.

## Testing Instructions

- With a WordPress.com account that has more than 100 sites, open the sync "Connect site" modal — every site should be listed.
- Connect a site, then reopen the picker — the connected site should appear as already connected.

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/modules/sync/lib/ipc-handlers.ts
+++ b/apps/studio/src/modules/sync/lib/ipc-handlers.ts
@@ -524,7 +524,11 @@ export async function fetchSyncableWpcomSites( _event: IpcMainInvokeEvent ): Pro
 	if ( ! token?.accessToken ) {
 		throw new Error( 'Authentication required to fetch WordPress.com sites.' );
 	}
-	return fetchSyncableSites( token.accessToken );
+	// Pass the already-connected remote IDs so the transform can mark those
+	// sites 'already-connected' instead of offering them as syncable again.
+	const connectedSites = await getAllConnectedWpcomSitesForCurrentUser();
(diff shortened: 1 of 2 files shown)


youknowriad:
This doesn't feel great to me. We shouldn't be fetching unpaginated results, there's a risk here depending on the number of sites of a given user. Instead I believe it's better to have a UI that is either paginated or load on scroll or things like that.

katinthehatsite:
> This doesn't feel great to me. We shouldn't be fetching unpaginated results, there's a risk here depending on the number of sites of a given user. Instead I believe it's better to have a UI that is either paginated or load on scroll or things like that.

Another point to add to this is that I find that the previews are generating for quite some time when all the sites are first fetched so as a user, you are stuck on a screen with all the sites with loading previews and it is hard to see which site is which from a quick glance.

shaunandrews:
Closing this split PR for now while I rethink and regroup the site-creation onboarding approach. The umbrella PR #3797 will stay open as the place to track the broader exploration.

PR #3790, opened 2026-06-11
Add onboarding UI primitives: dot grid, illustrations, footer, busy overlay

shaunandrews:
## Related issues

- Part of the site-creation onboarding redesign — see the umbrella PR (linked below once open) for the full plan.

## How AI was used in this PR

Split out of the `workbench/site-creation` exploration branch with Claude Code; designed and tested by Shaun on the original branch.

## Proposed Changes

Self-contained building blocks for the redesigned site-creation onboarding (nothing renders them yet — the flow PRs that follow wire them up):

- An interactive dot-grid canvas backdrop and a set of onboarding illustrations that give the new flows their visual identity.
- A shared onboarding footer and a busy overlay that shields the window during long-running actions like site creation.
- Keyboard/a11y groundwork: arrow-key grid navigation, keyboard-only focus rings, and an `@wordpress/a11y` dependency for screen-reader announcements.
- Small hooks for offline detection and mouse back/forward navigation buttons.

## Testing Instructions

- Nothing is user-visible yet; this lands dormant code. Verify the app builds and existing UI is unaffected (focus rings on buttons should only show for keyboard focus, in both light and dark mode).

## Pre-merge Checklist

### DIFF
--- a/apps/ui/package.json
+++ b/apps/ui/package.json
@@ -17,6 +17,7 @@
 		"@tanstack/react-query": "^5.75.5",
 		"@tanstack/react-query-persist-client": "^5.96.2",
 		"@tanstack/react-router": "^1.120.14",
+		"@wordpress/a11y": "^4.47.0",
 		"@wordpress/api-fetch": "^7.47.0",
 		"@wordpress/components": "^34.0.0",
 		"@wordpress/core-data": "^7.47.0",
--- a/apps/ui/src/components/busy-overlay/index.tsx
+++ b/apps/ui/src/components/busy-overlay/index.tsx
@@ -1,0 +1,15 @@
+import styles from './style.module.css';
+
+/**
+ * Transparent full-window shield that blocks pointer interaction while a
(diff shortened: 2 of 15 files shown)


youknowriad:
This seems fine but it's a bit weird to have a number of components and tools on a PR that no one uses.

My reading is that you split the big PR in terms of "code" instead of splitting the PR in terms of behavior/features which is better.

For instance, I'd have instead of having a hook to handle keyboard navigation backward/forward... here and use it later in another unrelated PR. I'd have implemented this navigation back/forth directly on top of the current design/steps in its own PR, and later the redesign would just change the steps... but won't impact the back/forward navigation. This would make things easy to ship because we can actually test what we're shipping while right now we can only review in the abstract which is not great.

shaunandrews:
Closing this split PR for now while I rethink and regroup the site-creation onboarding approach. The umbrella PR #3797 will stay open as the place to track the broader exploration.

PR #3819, opened 2026-06-12
fix(cli): restore the session's recorded model when resuming by id

wesleyfantinel:
Resuming a session with \`--resume-session\` was silently dropping the model: the session file records every model change, but the resume-by-id path never read it, so every resumed turn ran the default model no matter what you were using before. The interactive resume path already restores it via \`resolveResumeSessionContext\` — this brings the by-id path in line.

One line of behavior: after finding the session, resolve the recorded model and use it for the turn. \`resolveSessionModel\` already falls back to the default for sessions with no recorded model or a model we no longer offer, so nothing changes for those.

Found this while wiring multi-turn benchmarks in studio-eval: we resume sessions headlessly with \`--json --resume-session\`, and model comparisons were quietly all running the default model on the follow-up turns. Verified end to end there too: a resumed turn on \`claude-sonnet-4-6\` now stays on sonnet (the session JSONL shows no model flip back to the default).

## Testing
- New test in \`apps/cli/commands/ai/tests/index.test.ts\`: resuming by id a session whose JSONL records a \`model_change\` to a non-default model asserts \`runStudioAgentTurn\` receives that model.
- \`npx vitest run apps/cli/commands/ai/tests/index.test.ts\` — 3 passed.
- \`npm run typecheck\` in \`apps/cli\` — clean.

### DIFF
--- a/apps/cli/commands/ai/index.ts
+++ b/apps/cli/commands/ai/index.ts
@@ -4,7 +4,7 @@ import {
 	type StudioChatFileAttachment,
 } from '@studio/common/ai/chat-files';
 import { type StudioChatImage } from '@studio/common/ai/chat-images';
-import { DEFAULT_MODEL, type AiModelId } from '@studio/common/ai/models';
+import { DEFAULT_MODEL, resolveSessionModel, type AiModelId } from '@studio/common/ai/models';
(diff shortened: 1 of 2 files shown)


sejas (approved):
Thanks! I confirmed that the model is preserved when resuming a session.

PR #3836, opened 2026-06-15
fix: allow selecting and copying Studio Code conversation text

sejas:
## Related issues

- Fixes [STU-1831](https://linear.app/a8c/issue/STU-1831/allow-selecting-and-copying-conversation-text)

## How AI was used in this PR

Used to trace why conversation text was unselectable and to pinpoint the inherited `select-none` on the app shell as the root cause. The diagnosis and the one-line fix were reviewed manually.

## Proposed Changes

Studio Code users couldn't select or copy any text from a conversation — user prompts, assistant replies, tool output, or agent questions. The app shell intentionally marks the whole window non-selectable for a native feel (`select-none` on the root), and that `user-select: none` was cascading into the conversation.

This re-enables text selection just for the Studio Code conversation, so users can highlight and copy anything the agent says (or that they said) — e.g. to reuse a command, save a snippet, or share a reply. The rest of the app chrome stays non-selectable, as before.

## Testing Instructions

1. Open Studio Code tab.
2. Send a message
3. Drag-select across the user prompt and the assistant's markdown reply
4. Observe the text should highlight and can be copied.

https://github.com/user-attachments/assets/3608e965-04f5-4928-b7f0-9f9a2c665373

As a follow-up we could add copy button to copy the message in markdown text.

## Pre-merge Checklist


nightnei (commented):
LGTM and works as expected 👍

nightnei (approved):

PR #3844, opened 2026-06-16
Fix dark mode in Studio Code conversation dialogs

sejas:
## Related Issues

- Fixes [STU-1835](https://linear.app/a8c/issue/STU-1835/fix-dark-mode-in-start-new-conversation-modal)

## How AI was used in this PR

Claude Code located the affected dialogs, traced the token cascade, and implemented the fix.

## Proposed Changes

In dark mode the Studio Code dialogs — "Start a new conversation?" (shown when switching to a different model family) and "Site created" — appeared as a white box with dark, barely-readable text while the rest of Studio Code followed the dark scheme. They now match dark mode: dark surface, light text, and a correct border.

These dialogs come from `@wordpress/ui`, which colors itself from `--wpds-color-*` design tokens. `apps/studio` has no app-wide color `ThemeProvider`, so those tokens keep their light-only fallbacks, and the dialog popup portals out to `<body>` beyond Studio's frame styling — so it stayed light regardless of scheme. The fix remaps the few color tokens the dialog chrome reads onto Studio's existing dark-aware `--color-frame-*` tokens, mirroring the existing `wp-ui-button-defense` pattern.

## Testing Instructions

1. Run `npm start`
2. Set Studio App  Dark mode in Settings -> General
3. Open Studio, go to the **Studio Code** tab.
4. Send a message
5. Change the model from Claude to GPT, or vice versa. 
6. Confirm the dialog uses a dark background with readable text
7. Repeat in Light appearance to confirm no regression.

| Before | After |
|--------|--------|
|  Screenshot 2026-06-16 at 13 35 51 | Screenshot 2026-06-16 at 13 32 25 |

## Pre-merge Checklist

- [x] Linted modified files
- [x] Type checks pass (`apps/studio`)
- [x] Existing `studio-code-session` tests pass
- [ ] Visually verified in both light and dark

### DIFF
--- a/apps/studio/src/components/studio-code-session/composer/family-switch-confirm-dialog.tsx
+++ b/apps/studio/src/components/studio-code-session/composer/family-switch-confirm-dialog.tsx
@@ -2,6 +2,7 @@ import { getAiModelLabel } from '@studio/common/ai/models';
 import { __, sprintf } from '@wordpress/i18n';
 import { Button, Dialog } from '@wordpress/ui';
 import buttonDefense from '../wp-ui-button-defense.module.css';
+import dialogDefense from '../wp-ui-dialog-defense.module.css';
 import type { AiModelId } from '@studio/common/ai/models';
 
 /**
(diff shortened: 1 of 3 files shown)


nightnei (changes requested):
It LGTM 👍 Screenshot 2026-06-16 at 14 39 19
Screenshot 2026-06-16 at 14 39 35
Only one thing - I see that in Studio we use native dialogs, for example: <br />Image

I think would be better to use native dialogs everywhere, but to avoid breaking Studio Agent mode styles, I think we can go with different dialogs for now.

nightnei (approved):

PR #3852, opened 2026-06-16
Remove ui-desks interface

shaunandrews:
## Proposed Changes

- Removes the experimental ui-desks interface from Studio so the app no longer exposes or carries the alternate desk/canvas experience.
- Keeps the supported Studio renderer paths intact: default Studio continues to use the legacy renderer, while the agentic UI flag uses the maintained `@studio/ui` renderer.
- Cleans up now-unused desk state so existing app configs stop carrying obsolete desk data after migration.
- Drops the desk-only dependency footprint, including `tldraw`, along with stale UI-mode flags and user-facing desk wording.

---

Most of this PR is deletion. The main new logic is the `06-remove-desks-config` migration, which removes obsolete top-level `desks` data from the Desktop app config after the ui-desks interface is gone.

The migration:

- Checks whether `app.json` exists and has an own top-level `desks` property.
- Treats missing, invalid, or non-object config as "nothing to migrate".
- Locks app config with `lockAppdata()` before writing.
- Re-reads the config while locked, removes only `desks`, and preserves the rest of the file.
- Always calls `unlockAppdata()` in `finally`, including write failures.

The focused tests cover:

- `needsToRun()` returning true only when `desks` exists.
- Missing or invalid config safely skipping the migration.
- Removing only `desks` while preserving unrelated app config.
- Unlocking app data if the write fails.

---

### Do we want the migration at all?

Keeping it makes the removal tidy by deleting orphaned ui-desks state from `app.json`. Removing it would make this PR closer to a pure code deletion, and may be reasonable because only a small number of people likely used ui-desks and any saved desk data is unlikely to be worth preserving or explicitly cleaning up. The stale `desks` key would be inert once the interface and IPC/storage paths are gone.


wojtekn:
@shaunandrews, this cleanup looks good.

I merged the trunk and fixed one regression.

I think we may need to handle one more thing in a follow-up. Removing Desks leaves most of `studio_present` widget types unused - the Desks canvas was their only renderer, and `run-manager.ts` already consumes only site-preview, which drives the agentic UI's live preview. It makes sense to open a separate PR to reduce the prompt manifest to site-preview and remove the "desk widgets" wording.

sejas (approved):
Thanks for this clean up. I started existing and agentic UI and didn't catch any regression. Changes make sense to me.

PR #3859, opened 2026-06-16
Agentic UI: Combine create and blueprint onboarding into Build a new site

shaunandrews:
This PR simplifies the Agentic UI Add Site flow by combining the Empty Site and Blueprint options into a single “Build a new site” path.

Previously, users had to choose between “Create new” and “Start from a Blueprint” from the first screen. That split made the Empty Site path feel like a different kind of action from the Blueprint path, even though both are really ways to start building a new local WordPress site.

With this change, the first screen asks the broader question: do you want to build a new site or bring in an existing one? Once the user chooses “Build a new site,” the next screen lets them pick the starting point: an empty install, an uploaded Blueprint, or a featured Blueprint.

This keeps the decision closer to the available starting content, makes “Empty site” feel like one starting point among several, and reduces the number of top-level choices in the onboarding home.

The PR also fixes the Blueprint picker layout so long featured Blueprint lists can scroll properly. The onboarding shell now grows beyond the viewport instead of trapping overflow below the fold, while keeping the close control available during scrolling.

## Proposed Changes

- Combines the separate “Create new” and “Start from a blueprint” onboarding choices into one “Build a new site” entry point.
- Moves the blank-site choice into the blueprint/start-point selector as an “Empty site” option, so users pick the kind of new site after choosing to build.
- Keeps the existing create and blueprint flows intact; this only changes how users enter them. It does include one fix for scrolling the blueprint gallery.

image

## Testing Instructions

- Open the app UI onboarding home.
- Confirm the home screen offers “Build a new site” instead of separate create/blueprint cards.
- Choose “Build a new site” and confirm the selector includes “Empty site” plus featured blueprints.
- Choose “Empty site” and confirm it opens the normal create-site form.

## Related issues

- Related to #3797

### DIFF
--- a/apps/ui/src/components/blueprint-selector/index.tsx
+++ b/apps/ui/src/components/blueprint-selector/index.tsx
@@ -26,6 +26,7 @@ interface BlueprintSelectorProps {
 	featured: FeaturedBlueprint[] | undefined;
 	isFeaturedLoading: boolean;
 	onPick: ( blueprint: PickedBlueprint ) => void;
+	onPickEmpty: () => void;
 }
 
 const FILE_ACCEPT = 'application/json,.json,application/zip,.zip';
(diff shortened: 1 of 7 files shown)


shaunandrews:
Closing this split PR while we rethink the onboarding strategy around #3797.

PR #3870, opened 2026-06-17
Annotate: remove the confirmation step before addressing annotations

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Used Claude Code to locate every place the annotate workflow was defined (the CLI skill and the two `apps/ui` prompt builders), remove the confirmation gate consistently, and update the affected unit test.

## Proposed Changes

When a user submits visual annotations from the site preview, the agent previously stopped to restate every change and ask for explicit confirmation (via `AskUserQuestion`) before touching the site. That extra round-trip slowed the loop down for what is usually a direct "fix this" request.

This removes that validation step across all three definitions of the workflow — the CLI `/annotate` skill, the agentic UI (desks) prompt, and the classic UI prompt. The agent now creates a todo list for the submitted annotations and proceeds straight to making the changes.

Retained the lightweight guidance that was still useful: tracking work with `TodoWrite`, identifying elements by what's visible rather than by selector, and the re-annotate fallback in the CLI skill.

Trade-off worth noting for reviewers: without the gate, the agent edits the site immediately on submit, so there's no built-in checkpoint to catch a misread annotation before changes land.

## Testing Instructions

- Run the unit test: `npm test -- apps/ui/src/ui-classic/components/session-view/annotations.test.ts` (passes; now asserts the prompt is TodoWrite-first and contains no `AskUserQuestion`).
- In the app, submit one or more annotations from the site preview and confirm the agent goes straight to making the changes rather than asking for a yes/no confirmation first.

## Pre-merge Checklist


draganescu:
I tested this and indeed it did not ask me to confirm anymore. However I saw this:

```
user: 1 annotation submitted
agent:
I'll set up the todo and fix the showtimes grid responsive breakpoints. Note: the previous navigation work was interrupted — I'll focus on this annotation now.

Update todo list
	Tool TodoWrite not found
The Todo tool isn't available in this environment, so I'll just track the single task inline and complete it. Let me look at the current responsive CSS for the showtimes grid.

```

youknowriad:
Ah yes, how did I miss that haha

youknowriad:
Good catch — fixed in 2fc33ff3e.

The prompt was instructing the agent to call `TodoWrite`, but that tool isn't registered in the pi runtime the CLI and app share (the session is given only Studio's custom tools, and pi itself has no todo built-in), so it failed with "Tool TodoWrite not found". The original prompt referenced it too, but the old confirmation gate ran first and masked it; once the gate was removed, `TodoWrite` became the agent's first action and surfaced the error.

The agent now acts on the submitted annotations directly (no `TodoWrite`, no confirmation gate). Removed the tool reference from the desks prompt, the classic prompt, and the CLI `/annotate` skill, and updated the unit test to assert the prompt no longer mentions `TodoWrite`.

katinthehatsite (commented):
I was just testing it as it was merged but wanted to mention that it worked correctly for me: 

Image

It would be also cool for the preview to refresh once the annotation notes were implemented so that the user can see the changes right away instead of having to manually refresh the preview.

PR #3871, opened 2026-06-17
Studio Code: prefer the theme color palette over hardcoded colors

youknowriad:
## Related issues

This is based on user feedback.

## How AI was used in this PR

Authored with Claude Code. The agent investigated how Studio Code is instructed around color, wrote a failing eval, made the prompt/skill changes, and verified the before/after by running the eval end-to-end against the real agent (WordPress.com AI proxy). All changes were reviewed by the author.

## Proposed Changes

When asked to add sections or build sites, Studio Code often invented **custom hex colors** instead of using the site theme's existing color palette. That detaches the result from Global Styles, theme switching, and light/dark variations, and tends to look off-brand.

This change makes **palette-first** the default behavior:

- Color sections from the active theme's palette by slug — block color attributes like `{"backgroundColor":"accent-1","textColor":"base"}` or `var(--wp--preset--color--*)` in CSS — rather than hardcoding hex.
- When building a new theme, define the palette once in `theme.json` and reference it everywhere; when editing a site that already has a theme, inherit its palette.
- Custom hex becomes a deliberate exception (a color the palette genuinely lacks — add it to the palette first), not the default.

It also fixes a skeleton-CSS example in the `block-content` skill that previously taught hardcoded `#ffffff`/`#111111` custom properties, and adds an eval (`section-uses-theme-palette`) that guards the behavior.

User impact: sections the agent adds stay consistent with the theme, respond to Global Styles edits, and adapt to theme/color-scheme changes — without constraining genuinely needed custom colors.

## Testing Instructions

Requires Node ≥22 with npm 11 (e.g. `nvm use 24`).

Run the new eval case (from the eval config dir, since the provider path is relative):

```
npm run cli:build
cd scripts/eval
npx promptfoo@0.121.4 eval -c promptfoo.config.yaml --filter-pattern "theme color palette" --no-cache
```

Verified before/after against real agent runs:
- **Before (trunk prompt):** fails — agent built the hero/CTA with hardcoded hex CSS (`#2d3a8c`, `#ffffff`, `#c7d0f8`, …) and zero palette references.
- **After (this PR):** passes — agent colored both sections from the palette (`{"backgroundColor":"accent-3","textColor":"base"}`, `{"backgroundColor":"accent-1","textColor":"contrast"}`) with no hardcoded hex.

To compare manually, `git stash` the three prompt/skill files, rebuild, and re-run the case to see it fail, then restore and re-run to see it pass.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? (lint + `npm run typecheck` pass)


apps/cli/ai/skills/block-content/SKILL.md
```diff
@@ -18,6 +18,7 @@ Use this skill before writing or editing page content, post content, templates,
...
 - No inline `style` attributes or block `style` attributes for styling. Use `className` plus the theme's `style.css`.
+- Color from the theme palette, not hardcoded hex. Apply block colors with palette **slug** attributes — `{"backgroundColor":"accent-1","textColor":"base"}` — and in `style.css` reference palette colors as `var(--wp--preset--color--<slug>)`. Never write raw hex color values (e.g. `#1a1a1a`) into block markup or section CSS. Discover the available slugs from the active theme's `theme.json` `settings.color.palette` (for a theme you are building, the palette you defined there). If a section needs a color the palette lacks, add an entry to the `theme.json` palette and reference its new slug — do not inline a hex value. This keeps sections in sync with Global Styles, theme switching, and light/dark variations. Custom hex is a deliberate exception, not the default.
```
youknowriad: "Never write raw hex color values (e.g. `#1a1a1a`) into block markup or section CSS"

That feels too strong for me, sometimes a custom color is warranted for one offs... So maybe remove this sentence or say something like "prefer block colors from palette"...

apps/cli/ai/system-prompt.ts
```diff
@@ -116,7 +116,7 @@ For any request that involves a WordPress site, you MUST first determine which s
 Then continue with:
 
 1. **Get site details**: Use site_info to get the site path, URL, and credentials.
-2. **Plan the design**: Before writing any code, review the site spec (from the \`site-spec\` skill) and load the \`visual-design\` skill to plan the visual direction: layout, colors, typography, and spacing.
+2. **Plan the design**: Before writing any code, review the site spec (from the \`site-spec\` skill) and load the \`visual-design\` skill to plan the visual direction: layout, colors, typography, and spacing. Decide the color palette up front, define it in \`theme.json\`, and drive every section from those palette colors by slug rather than hardcoding hex (reuse the existing palette when the site already has a theme).
```
youknowriad: I think the change here is probably redundant/not needed.

apps/cli/ai/system-prompt.ts
```diff
@@ -161,6 +161,7 @@ ${ studioPresentToolBullet }${ automaticArtifactSection }
 ## General rules
 
 - Design quality and visual ambition are not in conflict with using core blocks. Custom CSS targeting block classNames can achieve any visual design. The block structure is for editability; the CSS is for aesthetics.
+- Color sections from the theme's palette by default. Define colors once in \`theme.json\` (\`settings.color.palette\`) and reference them by slug — block color attributes like \`{"backgroundColor":"accent-1","textColor":"base"}\` or \`var(--wp--preset--color--<slug>)\` in CSS. Do NOT hardcode hex color values in block markup or section CSS. When working on a site that already has an active theme, reuse its existing palette; add a new palette entry only when a needed color is genuinely missing, then reference its slug. (See the \`block-content\` skill.)
```
youknowriad: This feels redundant with what we have in the skills no?

apps/cli/ai/system-prompt.ts
```diff
@@ -199,6 +200,7 @@ const REMOTE_CONTENT_GUIDELINES = `## Block content guidelines
 
 - Use only core WordPress blocks. No custom HTML blocks except for inline SVGs.
 - No decorative HTML comments (e.g. \`<!-- Hero Section -->\`). Only block delimiter comments are allowed.
+- Color content from the active theme's palette using block color-slug attributes (e.g. \`{"backgroundColor":"primary","textColor":"base"}\`) rather than hardcoded hex values; only introduce a custom color when the palette genuinely lacks one.
```
youknowriad, line 202: Is this necessary? maybe it is if remote system prompt don't rely on the design or block content skill? Can you check

apps/cli/ai/skills/block-content/SKILL.md
youknowriad: Agreed — softened in 72c33ef. Now reads "Prefer theme palette colors over hardcoded hex … a raw hex value is fine for a deliberate one-off, but it should be the exception, not the default." The absolute "never" was overreach.

apps/cli/ai/system-prompt.ts
youknowriad: Agreed, removed. The visual-design skill is loaded at this exact step and already carries it.

apps/cli/ai/system-prompt.ts
youknowriad: Agreed, removed. For local sites the block-content/visual-design skills are always loaded for design/markup work, so this was redundant — palette guidance now lives only in the skills. Re-ran the eval after removing it: still passes (agent loads block-content and uses palette slugs without this reinforcement).

apps/cli/ai/system-prompt.ts
youknowriad: Checked: the remote intro (buildRemoteIntro) only loads the wpcom-remote-management skill — never block-content or visual-design — and that skill has no color/palette guidance. So unlike the local path, nothing else covers this for remote sites; this bullet is the only place a WordPress.com build gets the nudge. Kept it (already phrased as "prefer + escape hatch", consistent with the block-content change).

lezama (approved):

sejas (commented):
Thanks for improving the generation of color vars and adding a promptfoo 🙏!

PR #3872, opened 2026-06-17
Remove the dead TodoWrite rendering subsystem from the CLI

youknowriad:
## Related issues

- Related to #

## How AI was used in this PR

Used Claude Code to trace `TodoWrite` end-to-end — confirming via `buildAgentTools` and the pi runtime that it is not a registered tool, mapping every reference across the CLI TUI, and removing the wiring plus the state/tests it orphaned.

## Proposed Changes

`TodoWrite` is not a tool the agent can call. `buildAgentTools` (`apps/cli/ai/runtimes/pi/index.ts`) constructs the full tool list explicitly — `Read`/`Write`/`Edit`/`Bash`/`Grep`/`Glob`/`Ls` plus the Studio tools, ask-user, and skill — and the pi runtime ships no todo built-in. So a `TodoWrite` call always fails with "Tool not found", and the CLI's todo diff/render machinery (which only does anything on a *successful* `TodoWrite`) has been dead since the pi-coding-agent migration (#3360) stopped surfacing synthetic todo messages. A code comment in the deleted `todo-stream.ts` already noted that the runtime "no longer surfaces" the todo message type.

This removes that unreachable subsystem: the todo diff engine, the snapshot rendering, the per-instance todo state on the CLI UI, and the now-unused display label. If a model still emits a stray `TodoWrite` out of habit, it now renders through the generic tool path (a normal "tool not found" line) instead of dedicated dead code — no user-facing behavior is lost, since the dedicated path could only ever render the same failure.

This is pure dead-code removal: no behavior that was actually reachable changes. Scope was deliberately limited to `TodoWrite` and the state/tests it orphaned; no broader dead-code sweep was done.

Companion to the annotate-workflow PR (#3870), which stops *instructing* the agent to call `TodoWrite`. The two are independent — this one removes the rendering, that one removes the prompt instructions.

## Testing Instructions

- `npm run cli:build && npm run typecheck` — passes.
- `npm test -- apps/cli/ai/tests/ui.test.ts` — all 12 tests pass (the removed lines were leftover state setup, not todo assertions).
- Optional manual check: run the CLI agent and confirm tool rows (Read/Edit/Bash/Skill/etc.) still render correctly under their matching rows.

## Pre-merge Checklist


youknowriad:
This is legacy handling of the todo tool that don't exist anymore.

PR #3875, opened 2026-06-17
Studio Code: treat shop/store/sell requests as ecommerce deterministically

youknowriad:
The main reason I started this PR is that I noticed that woo commerce was not consistently used when a user asked for a shop. The eval was passing most of the time but not all the time.

## Related issues

- Related to the `woocommerce-shop-setup` eval case in `scripts/eval/promptfoo.config.yaml`.

## How AI was used in this PR

Studio Code (the in-CLI agent) is the subject of the change — this PR tunes the `plugin-recommendations` skill it loads. The change was developed and verified with Claude: I reproduced the flaky behavior, strengthened the skill, and confirmed determinism by running the agent end-to-end against the WordPress.com AI proxy across multiple consecutive runs.

## Proposed Changes

When a user asks the agent to build a shop or to "sell" something, they expect a working storefront — WooCommerce installed with representative sample products — not a few hand-built informational pages. Two intermittent failures caused the opposite:

1. **Misclassification.** Requests carrying scope/style qualifiers like "small", "just a couple of pages", or "for inspiration" were sometimes treated as informational content sites, so WooCommerce was skipped entirely.
2. **Empty catalog under time pressure.** Even when WooCommerce was set up, the agent often did theme scaffolding and visual polish first and ran out of its turn budget before creating any products — leaving a shop with nothing to sell.

This keeps the guidance where the WooCommerce logic already lives — the `plugin-recommendations` skill — without touching the core system prompt:

- The skill's trigger and description now make it unambiguous that **selling anything** ("I want to sell X", a shop, store, storefront, or ecommerce site, or any design implying products/prices/a catalog/a Shop page) is always an ecommerce request that requires WooCommerce **and** products — and that size/style qualifiers never downgrade it.
- The skill now tells the agent to **front-load** WooCommerce setup and product creation: install, configure, and populate the catalog *before* the slower theme-design and polish phase, and never let image sourcing block products from being created. The catalog is the core of the site, not a finishing touch.

Net effect for users: shop/store/"I want to sell …" prompts deterministically produce a real storefront with products, regardless of how small or stylistically framed the request is.

## Testing Instructions

- From `scripts/eval`, run the regression case (Node ≥22 / npm ≥11):
  ```
  npx promptfoo@0.121.4 eval -c promptfoo.config.yaml --filter-pattern "WooCommerce" --no-cache
  ```
  The `woocommerce-shop-setup` case asserts WooCommerce is installed and ≥1 product is created.
- Verified on **3 consecutive promptfoo runs (all passed)** with the final skill; each run installed WooCommerce and created 6 products, with product creation occurring before the design/polish phase. (An earlier skill-only attempt without front-loading flaked 1-in-3 by timing out after categories but before products — the front-loading change fixes that failure mode.)
- `npm run typecheck` passes.

## Pre-merge Checklist


apps/cli/ai/skills/plugin-recommendations/SKILL.md
```diff
@@ -38,7 +38,11 @@ wp_cli eval 'foreach (\WP_Block_Type_Registry::get_instance()->get_all_registere
...
+Selling means ecommerce, and ecommerce means WooCommerce with products. Any request to **sell** something - "I want to sell X", "a shop that sells …", "a store for …" - or any request that names a shop, store, storefront, or ecommerce site is an ecommerce request, full stop. You MUST set up WooCommerce AND create products for it. This also applies whenever the planned design includes products, product categories or ranges, prices, a catalog, a "Shop" page, or add-to-cart, even if the user never names WooCommerce and never explicitly asks for products. Treating such a request as an informational/content site and hand-building plain pages instead is wrong - a storefront without WooCommerce and without a populated catalog does not let anyone sell or buy anything.
```
katinthehatsite: >  a storefront without WooCommerce and without a populated catalog does not let anyone sell or buy anything

I am not sure this bit is needed as everything before that looks clear and explanatory. It is not a strong opinion so we can also leave it as is

katinthehatsite (approved):
I think the changes look good 👍 

> Empty catalog under time pressure. Even when WooCommerce was set up, the agent often did theme scaffolding and visual polish first and ran out of its turn budget before creating any products — leaving a shop with nothing to sell.

I ran into this multiple times, especially with the first page having products but then clicking each product led to empty URLs. This was also a common issue with product categories. I think it is a good direction to prioritize making full products first before setting up everything else to make the user experience more complete 👍

apps/cli/ai/skills/plugin-recommendations/SKILL.md
youknowriad: I can try to leave it out and run the eval to confirm.

youknowriad: Made the change, I'm running the eval before merging.

PR #3878, opened 2026-06-17
Studio Web: SecexRuntime — run the agent in a hosted SecEx sandbox

lezama:
## Related issues

- **Stacked on #3877** (the `AgentRuntime` seam — review/merge that first; this PR targets its branch).
- Part of the Studio Web → hosted effort. Drives the wpcom `studio-code` endpoint (Automattic/wpcom#212555) server-side.

## How AI was used in this PR

The runtime and this description were drafted with Claude Code under close direction, written to the `studio-code` endpoint's documented contract. I reviewed the diff and ran the local verification (typecheck, lint, CLI build, and a smoke test of the failure path + that the default local runtime is unaffected).

## Proposed Changes

#3877 put the agent's execution behind an `AgentRuntime` seam. This adds the second implementation: instead of forking the agent as a local child process, `SecexRuntime` runs it inside a hosted **per-user SecEx sandbox** by driving the wpcom `POST /wpcom/v2/studio-code/run` endpoint and relaying the sandbox's streamed events through the same callbacks the local runtime uses. The run-manager, the HTTP/SSE API, and the browser don't change — only *where the agent runs*.

The model mirrors Studio App: one sandbox per user is the cloud analog of "your laptop", holding the user's sites and sessions, resolved from the caller's WordPress.com token. It's opt-in — nothing changes unless you start the backend with `STUDIO_WEB_BACKEND=secex`; the default local runtime is untouched.

Scope is deliberately narrow — **running a turn** (send prompt → stream the agent's events back). Two things are out of scope and tracked as follow-ups:
- **Session listing/history from the sandbox.** In SecEx mode the agent's sessions live inside the sandbox, not the web-server's local store, so the existing `/sessions` listing won't reflect them yet.
- **Interrupt/answer semantics.** The endpoint has no interrupt route, so interrupt closes the stream (the endpoint kills the CLI and snapshots partial state); answer is a no-op because the endpoint runs the CLI with `--auto-approve`.

> ⚠️ **Not yet verified against the live endpoint.** The code is written to the endpoint's documented contract, but an end-to-end run needs the `studio-code` endpoint deployed and a built SecEx template. The local failure path (auth missing / endpoint unreachable → `error` + `run.exited`, no crash) is verified; the happy path is not.

Known risk to handle later (flagged on the endpoint side too): long agent turns can drop the HTTP/2 connection mid-stream — needs heartbeats/resume.

## Testing Instructions

Local (failure path, no endpoint needed):
1. `npm install && npm run cli:build`.
2. `STUDIO_WEB_BACKEND=secex STUDIO_SECEX_RUN_URL=http://127.0.0.1:9/run node apps/cli/dist/cli/main.mjs web-server`.
3. Send a prompt; confirm the run streams an `error` then `run.exited` and the server stays up.
4. Without the env var, confirm the local runtime behaves exactly as before.

End-to-end (needs the live endpoint + a signed-in user):
5. `studio auth login`, then `STUDIO_WEB_BACKEND=secex node apps/cli/dist/cli/main.mjs web-server`, and send a prompt — the agent runs in the sandbox and its events stream into the browser.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/web-server/agent-runs.ts
+++ b/apps/cli/web-server/agent-runs.ts
@@ -1,22 +1,25 @@
-import { fork, type ChildProcess } from 'node:child_process';
 import crypto from 'node:crypto';
+import { localRuntime } from './local-runtime';
+import type { AgentProcess, AgentRuntime } from './runtime';
 import type { ActiveAgentRun, AgentEvent, AgentRunEvent } from '@studio/common/ai/agent-events';
-import type { JsonEvent } from '@studio/common/ai/json-events';
 
 /**
  * Headless analog of the desktop `run-manager` (apps/studio/src/modules/
- * ai-agent/run-manager.ts). It forks the exact same CLI subcommand the desktop
- * forks — `code sessions resume <id> <prompt> --json` — relays the child's
- * `JsonEvent`s as `AgentRunEvent`s, and synthesizes the same lifecycle events
- * (`run.started`, `run.exited`, ...). The only difference is the sink: instead
- * of `webContents.send`, events go to a broadcaster (SSE) injected via
- * `setBroadcast`.
+ * ai-agent/run-manager.ts). It runs the same `studio code sessions resume <id>
+ * <prompt> --json` agent the desktop runs, relays the agent's `JsonEvent`s as
+ * `AgentRunEvent`s, and synthesizes the same lifecycle events (`run.started`,
+ * `run.exited`, ...). The sink is a broadcaster (SSE) injected via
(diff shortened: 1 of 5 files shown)


lezama:
Closing: Studio Web is paused and this prototype is superseded — the SecEx endpoint work shipped via the merged series in wpcom (secure foundation → POST /run → session broker → /export → CLI bump), full inventory in Linear BIGR-604. See BIGR-660 for the site-generation decision (wpcom-native).

PR #3881, opened 2026-06-18
Make the `pull-reprint` command work with native PHP

fredrikekelund:
## Related issues

- Fixes STU-1814

## How AI was used in this PR

I used Claude to help me with the initial implementation. Then I had to slow down to understand the `pull-reprint` architecture and the changes that Claude made. I went through a few rounds of iterations by hand and with Claude before reaching the final state.

## Proposed Changes

The `pull-reprint` command comes with more assumptions about the PHP runtime than most other Studio features do. The prime example is how it uses PHP-WASM-based parsing to extract constants from `runtime.php`.

Studio now has two PHP runtimes, and native PHP is slated to become the new default shortly. This PR makes the `pull-reprint` command work with native PHP. This implementation does not cover all bases, but it works when pulling a typical WordPress.com site. We'll keep iterating to refine the experience.

## Testing Instructions

1. Run `npm run cli:build`
2. Run `STUDIO_ENABLE_PULL_REPRINT=1 STUDIO_RUNTIME=native-php node apps/cli/dist/cli/main.mjs pull-reprint`
3. Ensure that you can successfully complete a pull, that the site starts and works

## Pre-merge Checklist


wp-files/reprint/reprint.phar
fredrikekelund, line 1: This file, the changes in `scripts/download-wp-server-files.ts` and `.gitignore` are temporary. I needed the changes from https://github.com/WordPress/reprint/pull/261 to make this work, and there isn't yet a Reprint release that includes those changes. 

I'll mark this PR as ready for review once there is a Reprint release with those changes and I've reverted these parts.

apps/cli/playground-server-child.ts
```diff
@@ -222,28 +222,6 @@ async function getBaseRunCLIArgs(
...
-					config.mounts = saved.mounts;
-					config.wordpressInstallMode = saved.wordpressInstallMode ?? config.wordpressInstallMode;
-					config.useExactMountLayout = true;
-					logToConsole( `Loaded persisted start options from ${ optionsPath } before startup` );
-				}
-			}
-		} catch {
-			// Ignore missing or invalid start options and continue with the provided config.
-		}
-	}
```
fredrikekelund, line 245: We get this data straight from the `config` object sent over IPC. No need to re-read `start-options.json`

apps/cli/lib/types/wordpress-server-ipc.ts
```diff
@@ -1,19 +1,20 @@
...
+import type { WordPressInstallMode } from '@wp-playground/wordpress';
 
 // Zod schemas for validating IPC messages from wordpress-server-manager
 const mountSchema = z.object( {
 	hostPath: z.string(),
 	vfsPath: z.string(),
 } );
 
-const wordpressInstallModeSchema = z.enum( [
+const wordpressInstallModeSchema: z.ZodType< WordPressInstallMode > = z.enum( [
```
fredrikekelund, line 11: We cannot derive a schema from the `WordPressInstallMode` type, but we can ensure that our manually defined schema matches the given type.

scripts/download-wp-server-files.ts
fredrikekelund, line 1: I had to upgrade Reprint to the latest release to make my changes work. This new approach ensures we still ship a fixed Reprint version with every release (i.e. it's not updated on the user's machine), but we also don't have to explicitly change our code to get the latest version. I think that's a good approach

epeicher: I agree with this approach. It was originally asked by @wojtekn [here](https://github.com/Automattic/studio/pull/3690#discussion_r3379141510) if we should use a tagged version, but given the fact that Reprint is a known dependency, I think it is safe to get the latest version, and we don't have to change our code to upgrade the version.

apps/cli/lib/native-php/php-process.ts
```diff
@@ -150,35 +158,43 @@ export function reapPhpTreeOnInterrupt( child: ChildProcess ): () => void {
...
 		} );
 		phpScriptProcess.once( 'close', ( code ) => {
 			if ( code === 0 ) {
-				resolve( { stdout } );
+				resolve( { stdout, stderr } );
 				return;
 			}
 
-			reject( new Error( `PHP command failed (code: ${ code })` ) );
+			reject( new Error( `PHP command (${ args.join( ' ' ) }) failed (code: ${ code })` ) );
```
epeicher: In case of an error, is the `stderr` logged somewhere after these changes?

epeicher (commented):
Thanks for solving this @fredrikekelund! This is now required after merging #3786 because the `pull-reprint` is now failing. With this solution, we would not need https://github.com/Automattic/studio/pull/3922. 

I have added an early review, and the changes look good to me. I have just left some questions/comments. Please let me know when you resolve the merging conflicts and I can give some testing.

apps/cli/php/router.php
fredrikekelund, line 1: The changes in this file are necessary to prevent `auto_prepend_file` from firing twice for requests that target PHP files directly (like `/wp-admin/admin-ajax.php`).

If `auto_prepend_file` fires twice, it could cause classes to be redeclared, for example, which triggers an error.

apps/cli/lib/native-php/php-process.ts
fredrikekelund: This change was actually a leftover that I reverted.

What happens to the output of the child process depends on the `mode` option passed to the `runPhpCommand()` function

apps/cli/commands/pull-reprint.ts
```diff
@@ -434,7 +450,13 @@ export async function runCommand(
 				sourceSite.wpComToken.accessToken,
 				verbose
 			);
-			preflight = await runPreflight( studioMetadata, apiUrl, secret, verbose );
+			preflight = await runPreflight(
+				SITE_RUNTIME_NATIVE_PHP,
```
epeicher, line 454: Should we use a dynamic value for this `SITE_RUNTIME_NATIVE_PHP`? I'm wondering what happens if a user selects the _playground_ runtime in Studio. Not a blocker, just a question for consideration.

scripts/download-wp-server-files.ts
```diff
@@ -121,8 +118,17 @@ const FILES_TO_DOWNLOAD: FileToDownload[] = [
...
 		name: 'reprint',
-		description: `reprint.phar (${ REPRINT_VERSION })`,
-		getUrl: () => REPRINT_PHAR_URL,
+		description: `reprint.phar`,
+		getUrl: async () => {
+			const release = await fetchLatestGithubRelease( 'WordPress/reprint' );
+			const asset = release.assets.find( ( a ) => a.name === 'reprint.phar' );
+			if ( ! asset ) {
+				throw new Error(
+					`No asset found in latest wp-cli-sqlite-command release ${ release.tag_name }`
```
epeicher: Wrong message from another Error probably

epeicher (commented):
Thanks for working on this @fredrikekelund! I have tested, and the `pull-reprint` works, but when the site is going to start, I'm getting the following error in the logs:
```
Error: Error establishing a database connection.
```
I am testing this by running:
```
node apps/cli/dist/cli/main.mjs pull-reprint --url <site-url>
```
This happens consistently when pulling a new site and creating a new site in Studio:

CleanShot 2026-06-24 at 16 18 21@2x

epeicher:
After some investigation, I have created this draft PR in Reprint that seems to fix the issue described [here](https://github.com/Automattic/studio/pull/3881#pullrequestreview-4562067923):

- https://github.com/WordPress/reprint/pull/278

To test it, what I do is:
1. On that branch, run `composer build:phar`
2. Copy `reprint.phar` to the Studio repo `wp-files/reprint/`
3. Run `npm run cli:build` on Studio
4. Test by running `node apps/cli/dist/cli/main.mjs pull-reprint --url mypurewebsite5.wpcomstaging.com`

Hope that helps

apps/cli/commands/pull-reprint.ts
fredrikekelund: We need to iterate on this in another PR. I'm trying to think of ways that we could tie the `pull-reprint` state management closer to the regular Studio state management (which is site-based rather than pull-based). That would help address this issue

fredrikekelund:
> `Error: Error establishing a database connection.`

I could not reproduce this exact error, but the root cause of this was a late change on my part from passing `--runtime=nginx-fpm` to `--runtime=php-builtin`. I tested this, but clearly not thoroughly enough.

[Your PR](https://github.com/WordPress/reprint/pull/278) seems worth considering, but I changed the option back to `--runtime=nginx-fpm` for now so that we can land this PR and proceed with the other items on our list.

apps/cli/commands/pull-reprint.ts
epeicher: Great

epeicher (approved):
Thanks @fredrikekelund for addressing the comments. This is working perfectly now, I have tested a new `pull` and an incremental `pull` and everything works as expected. LGTM!

Image

epeicher:
> [Your PR](https://github.com/WordPress/reprint/pull/278) seems worth considering, but I changed the option back to `--runtime=nginx-fpm` for now so that we can land this PR and proceed with the other items on our list.

I think that PR is not required eventually, your changes fix the original issue, I have closed it and [linked back here for the reasoning.](https://github.com/WordPress/reprint/pull/278#issuecomment-4797310675)

Issue #28, opened 2024-04-25
Dragging doesn't work for the right (white) part of the initial app window


zdenys:
### Quick summary

Dragging the initial app screen works only for the left part (blue) but not the entire right one (white). 

https://github.com/Automattic/studio/assets/39308239/c3e8dcff-0608-42f4-ad86-8efdefa8a3fb

### Steps to reproduce

1. Start Studio App for the first time
2. At the initial screen, try clicking the right (white) part of the window and dragging it

### What you expected to happen

The window should drag just like it happens for the left (blue) part of the window.

### What actually happened

No dragging happens. 

### Impact

All

### Available workarounds?

Yes, easy to implement

### Platform

Mac Silicon

wojtekn:
> The window should drag just like it happens for the left (blue) part of the window.

@zdenys, thanks for your report. I'm unsure why you would expect the whole area to be a dragging trigger. On this screen, the left blue part is a trigger, and on the screen shown after onboarding, the top, right, and bottom borders are used for that.

I've checked a few other apps, and I couldn't see examples of using areas considered the main part of the app, especially forms, as draggable areas.

zdenys:
Not a designer/developer, so I don't know the best practices honestly. Just like with the other bug I reported https://github.com/Automattic/studio/issues/29 my main point is that the dragging area is not the one I expect.

By default it's always the entire top part of an app. It works for Chrome, Alfred, Safari, Calculator, etc.

And in this initial app screen that is not the case which confused me as a user. I expect this area to trigger the drag:

![Screenshot on 2024-04-25 at 14:26:25](https://github.com/Automattic/studio/assets/39308239/e053fa23-dc63-4348-8b7a-25d6272caa48)

Not this one:

![Screenshot on 2024-04-25 at 14:26:44](https://github.com/Automattic/studio/assets/39308239/b25ffe99-d784-4d11-b26a-2aabb33199ee)

I guess if the expected area was triggering the drag for me, I may not even have noticed that some much wider area is triggering the drag too. But the very fact that the expected one was not triggering the dragging confused me. 

Seeing that the entire left part of the app screen triggers the drag caused me to question "why the right side part is not doing that?".

Now I am genuinely curious about what reasoning is behind not following the default practice of making the entire top part of the app to be a dragging trigger? 

That said, my distilled ask is to make the entire top part of the app to be a dragging trigger, in alignment with common app behavior.

matt-west:
[...]

PR #3882, opened 2026-06-18
Studio Code: split provisioning from content apply around the validation gate

youknowriad:
## Related issues

- Follow-up to #3874 (made `validate_blocks` a mandatory gate; now merged into trunk).

## How AI was used in this PR

Follow-up refinement to the Studio Code agent's local workflow prompt, prompted by review of #3874. Verified the provisioning-before-validation ordering by driving a plugin-block build (a slideshow, whose block must be registered to validate) through `apps/cli/ai/eval-runner.ts` and confirming the agent installs/activates the plugin before it calls `validate_blocks`.

## Proposed Changes

#3874 left provisioning and content-apply bundled in one "Configure WordPress" step, even though they belong on opposite sides of the validation gate. `validate_blocks` runs in the live block editor, so a plugin-provided block (form, slideshow, Woo) only validates correctly once its plugin is active — bundled wording could read as "install plugins after validating," which would flag those blocks as invalid.

This splits the workflow into ordered steps — provision → validate → apply — so prerequisites are active before validation and only validated content is applied. It also drops the redundant "(MANDATORY GATE)" label from the step header; the enforcing language ("MUST pass validate_blocks… Never apply unvalidated block content") stays in the body.

User impact: more reliable builds for pages using plugin-provided blocks — the agent activates what those blocks depend on before validating them, avoiding false "invalid block" loops.

## Testing Instructions

1. `npm run cli:build` (Node ≥22 / npm ≥11).
2. Drive a plugin-block build and inspect tool ordering:
   `node apps/cli/dist/cli/eval-runner.mjs 'Create a site named "Slideshow Test" and add a slideshow of three placeholder images to its homepage.'`
   In the result JSON's `toolCalls`, the plugin install/activate `wp_cli` call should precede `validate_blocks`, and validation should pass.
3. Re-run the `single-page-build-turn-cadence` eval case to confirm the core-block path still passes.

Verified: in a clean slideshow run the agent installed/activated the plugin at tool call #5 and validated at #28 (provision well before validation), validation passing. Clean re-runs of the core-block eval case on this change are still pending — a transient WP.com AI backend slowdown interfered with that batch; will confirm green before merge.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? (`eslint --fix` + `npm run typecheck` for `apps/cli` pass)


youknowriad:
I actually thought this change was merged with the previous PR but I had forgotten to commit and I was testing locally with this change applied (run multiple evals)

PR #3884, opened 2026-06-18
CLI: promote site subcommands to top level and add config get/set

youknowriad:
## Related issues

Related to STU-1850

## How AI was used in this PR

This PR was implemented with Claude Code. The CLI command surface was mapped, the registration refactor and the new `config get` command were authored, tests were written and run, and the docs were updated. The author reviewed the diff, verified the help output and `config get` against a real local site, and ran the CLI test suite (167 passing).

## Proposed Changes

The Studio CLI buried its most common operations under a `site` sub-group, so users had to type `studio site start`, `studio site list`, and so on. This promotes those verbs to the top level — `studio create`, `studio list`, `studio start`, `studio stop`, `studio delete`, `studio status` — making the everyday commands shorter and more discoverable.

Site settings move into their own `config` group:

- `studio config set …` replaces `studio site set` (same flags).
- `studio config get` is new: with no argument it prints all settable settings (`--format table|json`); with a key (`studio config get php`) it prints that single value raw, which is convenient for scripting.

For backward compatibility the old `studio site …` group still works exactly as before — it's just hidden from help. Existing scripts and muscle memory keep working; new output and docs steer people to the top-level forms.

The top-level help is also ordered to group related commands sensibly (site management first, then sync, then the rest).

## Testing Instructions

1. `npm run cli:build`
2. `node apps/cli/dist/cli/main.mjs --help` — verify the top-level commands are listed (and `site` is no longer shown).
3. Run a few flattened verbs against a local site, e.g. `studio list`, `studio status --path ~/Studio/<site>`.
4. `studio config get --path ~/Studio/<site>` (table), `--format json`, and a single key (`studio config get php --path …`).
5. `studio config set --path ~/Studio/<site> --php 8.4`.
6. Confirm the hidden alias still works: `studio site --help`, `studio site list`, `studio site set …`.
7. `npm test -- apps/cli/commands/config apps/cli/commands/site` (167 passing).

## Pre-merge Checklist


bcotrim (approved):
Tested and works as described, didn't find any regressions.

One important fix we should make before landing this is to add `runtime` and `file-access` to `studio config get`. `studio config set` and `studio status` already have both.
Note: we should be careful with the translation from `getSiteRuntime` and `mode`, the option we present to the user is `native`/`sandbox`, while internally we have `native-php`/`playground`

Also while testing, `studio config get` outputted a site with an empty domain which I know it is related to custom domain, it seems the label already exists in the code, but it's not being used.

Image

youknowriad:
Thanks for the review! Addressed in 3250f1e07:

- **Added `runtime` and `file-access` to `config get`**, matching `config set` and `status`. `runtime` is reported as the user-facing mode (`native`/`sandbox`) via `siteModeFromRuntime( getSiteRuntime( site ) )` rather than the internal `native-php`/`playground`, so it round-trips with `config set --runtime`. `file-access` uses `getSiteFileAccess()` (`site-directory`/`all-files`). Added tests covering the translation and defaults.
- **Empty `domain` / unused label**: you were right that the `label` field was dead — the table was rendering the raw key. I removed `label` entirely, since the keys are the right thing to show here (they map 1:1 to `config set --<key>` and `config get <key>`). I kept the empty `domain` row visible on purpose: `config get` is a full inventory of settable keys, so an empty cell means "settable but currently unset" (the JSON output shows it as `null`). Happy to filter empty rows from the table instead if you'd prefer that — let me know.

Updated the studio-cli SKILL.md keys list too.

PR #3885, opened 2026-06-18
Fix Studio Code syntax error in Node.js 22

fredrikekelund:
## Related issues

- Fixes STU-1857

## How AI was used in this PR

Claude was used to help identify and fix the issue.

## Proposed Changes

We recently updated to Vite 8 (see https://github.com/Automattic/studio/pull/3674 and related PRs). Previously, Vite transformed the JS output to support our targeted Node.js version. It still can, but Vite 8 wasn't picking up the transform target based on our existing config. This PR adds the necessary config to transform the output JS code so Node.js 22 supports it.

All of this is obviously related to Vite 8 swapping out esbuild/Rollup for Rolldown. 

## Testing Instructions

1. `npm run cli:build`
2. `nvm use 22`
3. `node apps/cli/dist/cli/main.mjs code`
4. Ensure Studio Code starts

## Pre-merge Checklist

### DIFF
--- a/apps/cli/vite.config.base.ts
+++ b/apps/cli/vite.config.base.ts
@@ -33,6 +33,9 @@ const phpSourceCodePath = resolve( __dirname, 'php' );
 const skillsSourcePath = resolve( __dirname, 'ai/skills' );
 
 export const baseConfig = defineConfig( {
+	oxc: {
+		target: `node${ semver.major( minimumNodeVersion ) }`,
+	},
 	plugins: [
 		{
 			name: 'write-dist-extras',


fredrikekelund:
A quick look at the Vite [docs](https://vite.dev/config/build-options#build-target) suggests that `build.target` _should_ be enough, but it clearly isn't in our case.

sejas (approved):
Looks good. Thanks for fixing it. I confirm I was able to execute `nvm use 22 && node apps/cli/dist/cli/main.mjs code` after running `nvm use && npm i && npm run cli:build`.

We could consider using node 22 CI unit test or similar.

fredrikekelund:
> We could consider using node 22 CI unit test or similar.

@youknowriad also suggested some smoke tests (in the true "does this thing even work?" sense) that we could run for different Node.js versions. I'll defer to @bcotrim on this, because I think he might be driving changing the E2E test suite to run on the CLI at some point soon

PR #3886, opened 2026-06-18
UX: clarify naming between site-level and app-level Settings

katinthehatsite:
## Related issues

Fixes STU-1694

## How AI was used in this PR

It was used to update the tests and do the implementation.

## Proposed Changes

* Confirm that the `Settings` tab is now changed to `Site Settings`
* Confirm that all the tests are passing

## Pre-merge Checklist


apps/studio/e2e/page-objects/site-content.ts
bcotrim, line 66: Finding the tabs by visible label caused this update into a bigger change. 
`TabPanel` already ids each tab by its internal `name` (`tab-panel-<n>-settings`), which is less prone to failure due to localization/renaming issues. Something like:

```ts
getTabButton( tab ) {
  return this.locator.locator( `[role="tab"][id$="-${ tab }"]` );
}
```

Callers become `navigateToTab( 'settings' )` and never break on a relabel. Non-blocking.

bcotrim (commented):
Changes LGTM and work as described

Image

apps/studio/e2e/page-objects/site-content.ts
katinthehatsite: Made updates in several places in https://github.com/Automattic/studio/pull/3886/commits/422824818d3739daf44e6cdb84fb746f1b2bedc8

epeicher (approved):
Thanks @katinthehatsite! I have tested it, and I can see the new `Site Settings` tab as expected. Also navigating to other tabs work as expected. I would say that once the E2E tests pass, this could be merged. LGTM!

Image

PR #3887, opened 2026-06-18
Limit the number of sites the process manager keeps running

katinthehatsite:
## Related issues

Fixes STU-1841

## How AI was used in this PR

* It was used to plan and implement the feature

## Proposed Changes

This PR limits the amount of sites that can be started at the same time to avoid draining memory. We can start: 

* 24 native-php sites (24 × 1 = 24)                                                                          
* 4 playground sites (4 × 6 = 24)                                                                            
* Or any mix, e.g. 3 playground + 6 native-php (18 + 6 = 24) 

Screenshot 2026-06-18 at 4 56 11 PM

## Testing Instructions

* Build CLI `npm run cli:build`
* Run `node apps/cli/dist/cli/main.mjs site stop --all`
* Try starting 5 Plaground sites and confirm that on the fifth one, you get the warning about the running limit as indicated on the screenshot above

## Pre-merge Checklist


apps/cli/process-manager-daemon.ts
```diff
@@ -29,6 +33,16 @@ const STOP_TIMEOUT_MS = 2_500;
...
 
+// Weighted capacity limit for site processes. Playground (PHP WASM) sites use ~6x more memory
+// than native PHP sites (~720 MB vs ~120 MB), so they carry a heavier weight. The cap of 24
+// allows up to 24 native-PHP sites or 4 Playground sites (or a mix).
+const SITE_PROCESS_PREFIX = 'studio-site-';
+const CAPACITY_WEIGHTS: Record< SiteRuntime, number > = {
+	[ SITE_RUNTIME_NATIVE_PHP ]: 1,
+	[ SITE_RUNTIME_PLAYGROUND ]: 6,
+};
+const MAX_WEIGHTED_CAPACITY = 24;
```
fredrikekelund: I know the weight numbers came from my description in the issue, but seeing the total being capped at 24, I feel like that might be a bit too limiting. For a hard limit like this, maybe we should go a bit higher… Six Playground sites in parallel, perhaps?

apps/cli/process-manager-daemon.ts
```diff
@@ -214,6 +238,16 @@ export class ProcessManagerDaemon {
 			return this.toProcessDescription( existing );
 		}
 
+		if ( processName.startsWith( SITE_PROCESS_PREFIX ) ) {
+			const weight = CAPACITY_WEIGHTS[ runtime ] ?? CAPACITY_WEIGHTS[ SITE_RUNTIME_PLAYGROUND ];
+			const currentUsage = this.getWeightedCapacityUsage();
+			if ( currentUsage + weight > MAX_WEIGHTED_CAPACITY ) {
+				throw new Error(
+					`CAPACITY_LIMIT_REACHED: Cannot start site — the weighted capacity limit of ${ MAX_WEIGHTED_CAPACITY } has been reached (current usage: ${ currentUsage }, requested: ${ weight }). Stop some running sites first.`
+				);
```
fredrikekelund:
```suggestion
				const errorMessage =
					runtime === SITE_RUNTIME_PLAYGROUND
						? `Cannot start site. The maximum number of running sites has been reached (${ currentUsage }/${ MAX_WEIGHTED_CAPACITY }). Sandbox sites count as ${ weight } units. Stop some running sites first.`
						: `Cannot start site. The maximum number of running sites has been reached (${ currentUsage }/${ MAX_WEIGHTED_CAPACITY }). Stop some running sites first.`;

				throw new Error( `CAPACITY_LIMIT_REACHED: ${ errorMessage }` );
```
This message is displayed to users when starting a site using the CLI. I think we can simplify it along these lines. "Weighted capacity" doesn't mean much to a regular user.

apps/studio/src/hooks/use-site-details.tsx
```diff
@@ -565,7 +576,17 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
+						for ( const site of autoStartSites ) {
+							if ( cancel ) {
+								break;
+							}
+							const { capacityLimitReached } = await startServer( site );
+							if ( capacityLimitReached ) {
+								break;
+							}
+						}
+					} )();
```
fredrikekelund: This switches from parallel startup to sequential, which is much slower. I suggest reverting.

We have to be more clever about how we display the error modal if we keep the parallel startup behavior, but that's still fully doable.

apps/studio/src/hooks/use-site-details.tsx
```diff
@@ -594,7 +615,12 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 
 	const startAllStoppedSites = useCallback( async () => {
 		const stoppedSites = sites.filter( ( site ) => ! site.running && ! site.isAddingSite );
-		await Promise.allSettled( stoppedSites.map( ( site ) => startServer( site ) ) );
+		for ( const site of stoppedSites ) {
+			const { capacityLimitReached } = await startServer( site );
+			if ( capacityLimitReached ) {
+				break;
+			}
+		}
```
fredrikekelund: Again, this switches from parallel to sequential. I suggest reverting.

apps/studio/src/hooks/use-site-details.tsx
```diff
@@ -404,9 +404,10 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 	}, [] );
 
 	const startServer = useCallback(
```
fredrikekelund, line 411: An idea: add a new `startServers` function that starts a list of sites in parallel and can keep track of local state to prevent multiple `The maximum number of running sites has been reached` modals from being displayed if there's an error.

fredrikekelund (commented):
The basics are solid here 👍 Still, there are a few things I think we should tackle before landing this. Aside from the inline comments, I think we should also update `apps/cli/commands/site/start.ts` to write `autoStart: false` for the site if `startWordPressServer` throws.

fredrikekelund:
> `autoStart: false`

Just noting that https://github.com/Automattic/studio/pull/3837 just landed, which changes where the autoStart props are stored.

katinthehatsite:
Thanks for the review @fredrikekelund - I will take a look at the comments now.

apps/cli/process-manager-daemon.ts
katinthehatsite: I made adjustment for this in https://github.com/Automattic/studio/pull/3887/changes/a34787130e56487b7ab0b2668ae8190170977a8d to increase the amount of Playground sites

apps/cli/process-manager-daemon.ts
katinthehatsite: Made adjustments in https://github.com/Automattic/studio/commit/4f1be5af8f81e6b98e3d21f87b3802fa7a070733

apps/studio/src/hooks/use-site-details.tsx
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/3887/changes/ad7126d27126735a8c047a036681825124cc0835

apps/studio/src/hooks/use-site-details.tsx
katinthehatsite: Reverted in https://github.com/Automattic/studio/pull/3887/commits

apps/studio/src/hooks/use-site-details.tsx
katinthehatsite: Added in https://github.com/Automattic/studio/pull/3887/changes/a140e48017385e222d974aa3790c217dc2422a53

fredrikekelund (approved):
LGTM 👍 I pushed a fix for the thing I mentioned in my previous review about setting `autoStart: false` on sites that fail to start

katinthehatsite:
> LGTM 👍 I pushed a fix for the thing I mentioned in my previous review about setting autoStart: false on sites that fail to start

Thank you 🙇

PR #3816, opened 2026-06-12
Studio Web: run the portable UI in a browser via a web connector and a CLI web-server command

lezama:
## Related issues

- Related to the Studio Web exploration (running Studio's agent from a browser). This PR extracts the minimal foundation from the broader exploration in #3746 to start implementing Studio Web for real.

## How AI was used in this PR

The code and this PR description were drafted with Claude Code under close direction, starting from the working exploration in #3746. I reviewed the diff, and every verification step (typecheck, lint, full `apps/ui` test suite, both UI build targets, and a live smoke test of every endpoint) was run on this branch.

## Proposed Changes

Studio's portable renderer (`apps/ui`) was built so the same React app can run against different backends through the `Connector` interface. This PR exercises that promise end to end: it lets the full Studio UI — site list, AI sessions, agent runs — run in a plain browser tab, with no Electron involved.

It adds two opt-in pieces:

- **`studio web-server`** — a CLI command that serves the browser UI and exposes the capabilities the desktop app reaches over IPC as a small HTTP + SSE API under `/api` (sessions, sites, agent runs). It forks the exact same `code sessions resume … --json` subcommand the desktop app forks, and reads the same session store, so sessions started on desktop show up in the browser and vice versa. It binds to loopback only and rate-limits requests, since it serves the local user's data without authentication. One command, one origin: `studio web-server` → open `http://localhost:8088`.
- **A web connector + browser entry for `apps/ui`** — `STUDIO_TARGET=web` builds/serves the same React app wired to the HTTP/SSE connector instead of the Electron IPC connector. The site list shows the signed-in user's workable WordPress.com sites.

Nothing changes for existing users: the desktop renderer build, dev server port, and CLI commands are untouched unless you explicitly run the new command or the `:web` scripts. This is the foundation increment for Studio Web; follow-ups (WordPress.com login in the browser, session workspaces with draft/publish) build on this API surface.

**Bundle**: the web build is the same portable renderer the desktop ships — `dist-web` totals 22 MB, identical to the desktop renderer's `dist` (main chunk 8.24 MB / 2.21 MB gzipped vs desktop's 8.21 MB / 2.20 MB; the 13.4 MB vips worker is fetched on demand, not on initial load). The web target adds no bundle weight.

See `apps/cli/web-server/README.md` for the process topology and how the local pieces map onto a hosted deployment.

## Testing Instructions

1. `npm install && npm run cli:build`
[...]

PR #3890, opened 2026-06-18
Studio Web: run the agent in a hosted SecEx sandbox (opt-in runtime)

lezama:
## Related issues

Continues the Studio Web effort that landed the web-server + web connector (#3816) and the `AgentRuntime` seam.

## How AI was used in this PR

The runtime, the SSE reader, the pause/answer/resume handling, and this description were drafted with Claude Code, then reviewed; the change typechecks and builds locally.

## Proposed Changes

Studio Web can now run the agent in a **hosted SecEx sandbox** instead of a local child process, so the browser build of Studio works without anything running on the user's machine. Each user gets one sandbox — the cloud analog of "your laptop", with the same `~/Studio/` sites and `~/.claude` sessions inside it, like the desktop app.

It plugs into the existing `AgentRuntime` seam: a new runtime drives the wpcom `studio-code` endpoint (`POST /wpcom/v2/studio-code/run`), which runs `studio code --json` in the sandbox and streams its events back over SSE. The web-server relays those through the same callbacks the local runtime already uses, so the run-manager and the browser UI are untouched.

**Interactive questions and multi-turn work like the desktop.** There's no live process in the sandbox to answer over IPC, so the runtime mirrors the proven headless pattern: when a turn pauses on a question the agent process stays "connected" instead of exiting, and answering re-drives the endpoint with the user's choice as the next message on the same sandbox session (resolved from `turn.completed.sessionId`). To the web-server and the UI it looks like an ordinary live run that asks and gets answered.

It's **opt-in and zero-risk by default**: the local child-process backend stays the default, and the SecEx runtime is only installed when `STUDIO_WEB_BACKEND=secex` is set.

## Testing Instructions

```bash
npm run cli:build
STUDIO_WEB_BACKEND=secex node apps/cli/dist/cli/main.mjs web-server
# (sign in first with `auth login`; open the printed URL and send a prompt;
#  when the agent asks a question, answer it in the UI and the turn resumes)
```

Without the env var, `web-server` behaves exactly as before (local agent).

> Note: full end-to-end depends on the wpcom `studio-code` `/run` route being deployed; the runtime is written to that endpoint's contract.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? (`npm run typecheck` and `npm run cli:build` pass.)

### DIFF
--- a/apps/cli/web-server/agent-runs.ts
+++ b/apps/cli/web-server/agent-runs.ts
@@ -1,4 +1,5 @@
 import crypto from 'node:crypto';
+import { wdbg } from './debug';
 import { localRuntime } from './local-runtime';
 import type { AgentProcess, AgentRuntime } from './runtime';
 import type { ActiveAgentRun, AgentEvent, AgentRunEvent } from '@studio/common/ai/agent-events';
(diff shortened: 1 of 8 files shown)


lezama:
Closing: Studio Web is paused and this prototype is superseded — the SecEx endpoint work shipped via the merged series in wpcom (secure foundation → POST /run → session broker → /export → CLI bump), full inventory in Linear BIGR-604. See BIGR-660 for the site-generation decision (wpcom-native).

PR #3893, opened 2026-06-18
Add e2e coverage for the Sandbox (Playground) runtime

bcotrim:
## Related issues

_No linked issue._ Addresses the gap where the Sandbox (WordPress Playground) runtime had no end-to-end coverage after Native PHP became the default site runtime in #3786.

## How AI was used in this PR

Investigated the e2e harness, the Add Site modal UI, and the runtime plumbing, then wrote the test and page-object helpers with Claude Code (Opus 4.8). Selectors were checked against the real components (the `#php-runtime-select` dropdown and the Settings "PHP runtime" row), the additive `completeOnboarding` change was confirmed backward-compatible with all existing callers, and ESLint + `tsc` were run clean on the changed files.

## Proposed Changes

Studio supports two per-site PHP runtimes: Native PHP (the default since #3786) and the Sandbox (WordPress Playground / PHP-WASM). Every existing e2e test runs against the default runtime, so the Sandbox path had **zero** end-to-end coverage — and site/blueprint creation on Playground lost the implicit coverage it had back when Playground was the default.

This adds a single e2e test that exercises the Sandbox runtime through the real UI:

- Creates a site via onboarding with **Sandbox** selected under Advanced settings.
- Verifies the site actually boots under Playground (waits for the "Running" state).
- Verifies the Settings tab reports the site's PHP runtime as **"Sandbox"**.

It runs as part of the normal e2e suite on every platform/arch — no dedicated CI job and no gating. The supporting page-object additions (runtime selection, the runtime display row) are reusable for future runtime-specific tests. No product code is changed.

## Testing Instructions

1. Build the app and run just this test:
   ```
   npm run e2e -- sandbox.test.ts
   ```
   (or `npx playwright test sandbox.test.ts` against an existing build)
2. The test creates a Playground site, waits up to ~3 min for first-run asset downloads, then asserts the Settings tab shows **PHP runtime = Sandbox**.
3. Requires network access — Playground downloads the PHP-WASM build and WordPress on first run.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? — ESLint `--fix` and `tsc` are clean for the changed e2e files. The test is statically verified but **not yet run green**; please run it locally before merging.

### DIFF
--- a/apps/studio/e2e/sandbox.test.ts
+++ b/apps/studio/e2e/sandbox.test.ts
@@ -34,5 +34,13 @@ test.describe( 'Sandbox runtime', () => {
 		// The Settings tab reports the site as running on the Sandbox runtime.
 		const settingsTab = await siteContent.navigateToTab( 'Settings' );
 		await expect( settingsTab.phpRuntimeDisplay ).toContainText( 'Sandbox' );
+
+		// The Sandbox site actually serves its home page over HTTP.
+		await expect( siteContent.frontendButton ).toBeVisible();
+		const frontendUrl = await siteContent.frontendButton.textContent();
+		expect( frontendUrl ).not.toBeNull();
+		const response = await fetch( `http://${ frontendUrl }` );
+		expect( [ 200, 302 ] ).toContain( response.status );
+		expect( response.headers.get( 'content-type' ) ).toMatch( /text\/html/ );
 	} );
 } );
--- a/apps/studio/e2e/sandbox.test.ts
+++ b/apps/studio/e2e/sandbox.test.ts
@@ -1,0 +1,38 @@
+import { test, expect } from '@playwright/test';
+import { SITE_RUNTIME_PLAYGROUND } from '@studio/common/lib/site-runtime';
+import { E2ESession } from './e2e-helpers';
+import Onboarding from './page-objects/onboarding';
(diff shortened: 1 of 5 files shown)


apps/studio/e2e/sandbox.test.ts
```diff
@@ -0,0 +1,38 @@
...
+		await onboarding.closeWhatsNew();
+
+		// The site boots under the Playground runtime.
+		const siteContent = new SiteContent( session.mainWindow, siteName );
+		await expect( siteContent.runningButton ).toBeAttached( { timeout: 180_000 } );
+		await expect( siteContent.siteNameHeading ).toHaveText( siteName );
+
+		// The Settings tab reports the site as running on the Sandbox runtime.
+		const settingsTab = await siteContent.navigateToTab( 'Settings' );
+		await expect( settingsTab.phpRuntimeDisplay ).toContainText( 'Sandbox' );
```
fredrikekelund, line 36: Requesting the site home page and looking for a 200 HTTP status code response seems like a nice addition here.

fredrikekelund (approved):
Good idea 👍 I suggest adding two HTTP requests to the site home page (one for each runtime), but other than that, this LGTM

PR #3894, opened 2026-06-18
Agentic Sidebar: Polish Site Rows, Chat Indicators, and Action Alignment

shaunandrews:
| Before | After |
| ------------- | ------------- |
| image|image|

image

## How AI was used in this PR

Used Codex to iteratively implement requested visual refinements to the new agentic sidebar and run verification after the changes.

### Sidebar Header
- Removed the "Studio" label; we don't need it, as the application menu and icon exist at the OS level.
- Reducing the padding on the right to align actions with those below for sites and appearance. I also removed the dimming behavior, so the buttons are the same color as all the others.

### Site List
- Increased the row height to better fit the icon buttons and other indicators with equal spacing all around.
- Site icons are now always visible and slightly bigger. The icons allow for customization and help with scanning long lists.
- Site Icons now dim when the site is stopped.
- Adjusted row spacing so site icons, chat status indicators, and action buttons align consistently. Running and pending chat indicators now use a smaller shared status slot, with the pending state shown as a compact triangle indicator with improved light/dark contrast.
- Tightened chat status indicators so they feel less visually heavy next to site icons and chat titles.
- Reduced font-weight for chat titles.

### Sidebar Footer
- Adjusted the spacing to align the Gravatar with the site icons.

## Testing Instructions

- Run `npm run typecheck`.
- Run `npm test -- apps/ui/src`.
- In the new agentic UI sidebar, verify site icons appear for all sites, stopped sites render grayscale, site/chat/user text columns align, top action buttons align with site row actions, and pending/running chat indicators have comfortable spacing.
- Check the sidebar in both light and dark color schemes.

## Pre-merge Checklist

### DIFF
--- a/apps/ui/src/components/sidebar-header/index.tsx
+++ b/apps/ui/src/components/sidebar-header/index.tsx
@@ -16,7 +16,6 @@ export function SidebarHeader( { onToggleSidebar }: Props ) {
 	const navigate = useNavigate();
 	return (
 		<div className={ `${ styles.root } ${ isFullscreen ? styles.fullscreen : '' }` }>
-			<span className={ styles.title }>{ __( 'Studio' ) }</span>
 			<div className={ styles.actions }>
 				<Menu.Root modal={ false }>
 					<Menu.Trigger
--- a/apps/ui/src/components/sidebar-header/style.module.css
+++ b/apps/ui/src/components/sidebar-header/style.module.css
@@ -1,11 +1,15 @@
 .root {
+	--sidebar-header-inline-end: calc(
+		var(--wpds-dimension-padding-2xl) - var(--wpds-dimension-padding-xs)
+	);
+
 	display: flex;
 	align-items: center;
 	gap: var(--wpds-dimension-padding-sm);
 	/* Leave room for macOS traffic lights at {20, 20} plus a visible gap.
 	   min-height is tuned so the flex-centered content lines up with the
 	   vertical center of the traffic lights. */
-	padding: var(--wpds-dimension-padding-sm) var(--wpds-dimension-padding-2xl)
+	padding: var(--wpds-dimension-padding-sm) var(--sidebar-header-inline-end)
 		var(--wpds-dimension-padding-sm) 94px;
 	min-height: 46px;
 	-webkit-app-region: drag;
@@ -18,37 +22,22 @@
 }
 
 /* In macOS fullscreen the traffic lights are hidden, so reclaim the
-   padding that was reserved for them and align the title with the site
-   list entries below. The site list outer padding is `2xl` and its rows
-   now start flush with that edge (no row/button inline-start padding), so
-   the site label text starts at `2xl` from the sidebar edge. */
+   padding that was reserved for them. */
 .fullscreen {
-	padding-left: var(--wpds-dimension-padding-2xl);
-}
-
-.title {
-	flex: 1;
-	min-width: 0;
-	font-size: var(--wpds-typography-font-size-md);
-	font-weight: 600;
-	color: var(--wpds-color-fg-content-neutral);
-	overflow: hidden;
-	text-overflow: ellipsis;
-	white-space: nowrap;
+	padding-left: var(--wpds-dimension-padding-lg);
 }
 
 .actions {
 	display: flex;
 	align-items: center;
+	margin-inline-start: auto;
 	/* Match the site-row action buttons (`.siteActions` in site-list), which
 	   sit flush with no gap. */
 	gap: 0;
-	opacity: 0.65;
-}
-
-.actions:hover,
-.actions:focus-within {
-	opacity: 1;
+	/* The site rows have their action rail tucked into the row padding. Pull
+	   the header buttons over by the row's inner padding so the create/toggle
+	   controls share those same trailing columns. */
+	margin-inline-end: calc(-1 * var(--wpds-dimension-padding-sm));
 }
 
 .popup {
--- a/apps/ui/src/components/site-list/style.module.css
+++ b/apps/ui/src/components/site-list/style.module.css
@@ -1,15 +1,27 @@
 .root {
-	--site-list-row-height: 32px;
+	--site-list-row-height: 34px;
 	--site-list-row-background-hover: rgb(0 0 0 / 4%);
 	--site-list-row-background-active: rgb(0 0 0 / 7%);
+	--site-list-row-leading-inset: var(--wpds-dimension-padding-xs);
+	--site-list-row-outer-inset: calc(
+		var(--wpds-dimension-padding-lg) - var(--site-list-row-leading-inset)
+	);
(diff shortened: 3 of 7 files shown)


apps/ui/src/components/site-list/style.module.css
```diff
@@ -388,15 +381,18 @@
 	--wp-ui-button-border-color-active: transparent;
 	--wp-ui-button-height: var(--site-list-row-height);
 
+	position: relative;
```
bcotrim, line 392: Moving `position: relative` onto `.sessionLink` unconditionally is a nice catch — the `?` indicator now anchors correctly even when a chat has a pending question but is not running.

Since `.sessionLinkRunning` is gone, its reference in `index.tsx:170` is now dead (`styles.sessionLinkRunning` resolves to `undefined`). Can simplify to:
```
className={ styles.sessionLink }
```

apps/ui/src/components/site-list/style.module.css
```diff
@@ -72,30 +75,20 @@
...
 
-.siteActive .siteIconSlot {
-	inline-size: 16px;
-	/* Gap between the site icon and its title. Kept in sync with the chat-row
-	   indent below (`.sessionLink` padding-inline-start), which carries the same
-	   +4px so chat labels stay aligned under the site title. */
-	margin-inline-end: calc(var(--wpds-dimension-padding-xs) + 4px);
-	opacity: 1;
-	transform: translateX(0) scale(1);
+.siteIcon {
```
bcotrim, line 94: NIT / robustness: this sizes `SiteIcon` by overriding its own `.root { width: 16px }` from a different CSS module at equal specificity, so the winner depends on stylesheet order. I checked the built bundle and our override does win (the parent module CSS lands after the child), so it works today — just a touch load-order-fragile. Optional: give `SiteIcon` a size prop, or bump specificity like the `.siteStatusGlyph.siteStatusGlyph` trick already used below. Same idea applies to `.sessionInlineSpinner` sizing `Spinner`.

apps/ui/src/components/site-list/style.module.css
```diff
@@ -434,24 +427,29 @@
...
+/* The running spinner and question indicator sit inside the active row shape,
+   inset from its edge while the chat label keeps its shared title column. */
 .sessionInlineSpinner {
+	box-sizing: border-box;
 	position: absolute;
 	inset-block-start: 50%;
-	inset-inline-start: calc(var(--wpds-dimension-padding-sm) + 6px);
-	margin-block-start: -6px;
+	inset-inline-start: var(--site-list-chat-status-offset);
+	inline-size: var(--site-list-chat-status-size);
```
bcotrim, line 449: The inline spinner grows 12px → 18px here to fill the icon column — worth a glance that an 18px ring doesn't read too heavy as a 'working' indicator.

NIT: setting both `inline-size`/`block-size` and `width`/`height` to the same value is redundant, though it mirrors `.siteStatusGlyph`, so consistent enough to leave.

apps/ui/src/components/sidebar-header/index.tsx
```diff
@@ -16,7 +16,6 @@ export function SidebarHeader( { onToggleSidebar }: Props ) {
 	const navigate = useNavigate();
 	return (
 		<div className={ `${ styles.root } ${ isFullscreen ? styles.fullscreen : '' }` }>
```
bcotrim, line 18: Confirming intent: the header no longer renders the 'Studio' title, so the app name does not appear in the sidebar anymore — just the create/hide buttons now. 👍 if deliberate (it is not called out in the PR description, so flagging).

apps/ui/src/components/sidebar-header/style.module.css
```diff
@@ -18,37 +22,22 @@
...
-	overflow: hidden;
-	text-overflow: ellipsis;
-	white-space: nowrap;
+	padding-left: var(--wpds-dimension-padding-lg);
 }
 
 .actions {
 	display: flex;
 	align-items: center;
+	margin-inline-start: auto;
```
bcotrim, line 33: NIT: along with the alignment change, the action buttons lost their resting `opacity: 0.65` (and hover/focus → 1) and are now always full opacity. Assuming that's intended alongside 'Align Actions' — just confirming the dimming was dropped on purpose.

bcotrim (commented):
Reviewed locally on the branch — really nice polish 👏 The alignment system holds up: site icons + footer gravatar share the leading column, and site titles + chat labels + display name land on the next.

Ran it: eslint ✅, typecheck ✅, site-list test ✅ (2/2), `vite build` ✅. **Dark mode looks great.**

**RTL doesn't mirror** — but this is a pre-existing `apps/ui` gap, not this PR. `apps/ui` loads translations (`setLocaleData` in `main.tsx`) yet never sets `document.documentElement.dir`, the way `apps/studio` does in `use-localization-support.ts`. So strings localize but the layout stays LTR. This PR is actually RTL-ready (logical properties throughout — the only physical bit is the macOS traffic-light padding, which should stay physical). Worth a separate issue to wire `dir=rtl` across `apps/ui`; shouldn't block here.

Left a few inline notes — one cleanup plus a couple of NITs/confirmations. Nothing blocking.

> ⚠️ Visual change: still needs a light-mode pass (dark ✅ confirmed).

apps/ui/src/components/sidebar-header/index.tsx
katinthehatsite: I actually quite liked having the title there but that might be a personal preference

apps/ui/src/components/site-list/style.module.css
```diff
@@ -434,24 +427,29 @@
...
+	block-size: var(--site-list-chat-status-size);
+	width: var(--site-list-chat-status-size);
+	height: var(--site-list-chat-status-size);
```
katinthehatsite: I think the width/height here might not be needed

apps/ui/src/components/site-list/style.module.css
shaunandrews: Good catch, this wasn't intentional! I've cleaned it up.

apps/ui/src/components/sidebar-header/index.tsx
shaunandrews: Yup, this was an intentional design decision to remove the Studio title. Its not very common to see apps that have this on macOS, as the app's name appears in the Dock and the Menubar. Windows and Linux both show a title bar, which includes the "Studio" label.

apps/ui/src/components/sidebar-header/style.module.css
shaunandrews: Also intentional. With the opacity, the top action buttons were lighter than all the other actions in the sidebar. Removing the opacity makes all actions the same color.

bcotrim (approved):
LGTM 👍

katinthehatsite (approved):
The changes look good 👍

PR #3903, opened 2026-06-19
STU-1873: Improve tool call descriptions

shaunandrews:
## Related issues

- Related to https://linear.app/a8c/issue/STU-1873/studio-agentic-ui-improve-tool-call-presentation

## How AI was used in this PR

I used AI assistance to inspect the existing tool-call rendering surfaces, port the shared description changes from the older exploration branch, and add focused tests around the shared helper behavior. I reviewed the resulting diff and ran the checks listed below.

## Proposed Changes

- Makes Studio agent tool calls easier to understand in compact surfaces by giving common WordPress and Studio tools clearer action-oriented labels.
- Keeps terminal and WP-CLI command details available for surfaces like Studio Code and the CLI, where users may not have a richer details panel.
- Preserves safe fallbacks for unknown tools so existing transcripts still render.

## Testing Instructions

- Run `npx eslint --fix apps/cli/ai/ui.ts apps/studio/src/components/studio-code-session/conversation/index.tsx apps/ui/src/ui-classic/components/session-view/conversation/index.tsx tools/common/ai/tools.ts tools/common/ai/tests/tools.test.ts`.
- Run `npm test -- tools/common/ai/tests/tools.test.ts apps/cli/ai/tests/ui.test.ts apps/studio/src/components/studio-code-session/conversation/index.test.tsx`.
- Run `npm run typecheck`.
- Trigger a Studio Code or Agentic UI run that uses WP-CLI or Bash and confirm the tool row uses a clearer label while still showing the command detail.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #3904, opened 2026-06-19
STU-1873: Add tool call detail panel

shaunandrews:
image
image

## Related issues

- Related to https://linear.app/a8c/issue/STU-1873/studio-agentic-ui-improve-tool-call-presentation
- Follow-up to #3903

## How AI was used in this PR

I used AI assistance to port the tool-row presentation from the older exploration branch onto current trunk, keep the scope limited to the classic Agentic UI transcript, and add regression tests. I reviewed the final diff, updated the branch after #3903 landed, ran automated checks, and verified the UI in light and dark modes with a temporary Vite harness.

## Proposed Changes

- Makes classic Agentic UI tool calls easier to scan by showing an icon, a human-readable label, and a compact detail.
- Lets users expand a tool row to inspect the command/input and output without long output taking over the transcript.
- Keeps Ask User questions from rendering a duplicate raw tool row, so the custom question UI remains the only visible prompt.

## Testing Instructions

- Run `npm test -- tools/common/ai/tests/tools.test.ts apps/ui/src/ui-classic/components/session-view/conversation/index.test.ts`.
- Run `npm run typecheck`.
- In classic Agentic UI, trigger tool calls such as WP-CLI and Bash, then click the tool rows to confirm input/output appears inside a bounded scrollable details panel.
- Check the tool rows and expanded details in both light and dark color schemes.

## Pre-merge Checklist


apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
```diff
@@ -232,39 +495,77 @@ function ToolUseRow( {
...
+					data-expanded={ expanded }
+					onClick={ () => {
+						if ( expanded ) {
+							setExpanded( false );
+							return;
+						}
+						setDetailsMounted( true );
+						setExpanded( true );
+					} }
+					title={ expanded ? __( 'Hide tool details' ) : __( 'Show tool details' ) }
```
katinthehatsite, line 515: Do you think the contrast might be too low for accessibility for the tool row for when it is not expanded with the light mode?

Image

I am finding that it works out better in the dark mode where the contrast reads better.

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
```diff
@@ -232,39 +495,77 @@ function ToolUseRow( {
...
-							className={ styles.toolOutputToggle }
-							onClick={ () => setExpanded( ( prev ) => ! prev ) }
-						>
-							{ expanded ? __( 'Show less' ) : __( 'Show more' ) }
-						</button>
-					) : null }
+			{ hasExpandableDetails ? (
+				<button
+					type="button"
+					className={ clsx( styles.toolRow, styles.toolRowButton ) }
```
katinthehatsite, line 502: I think we can add an aria-label there for screen readers:

```diff --git a/apps/ui/src/ui-classic/components/session-view/conversation/index.tsx b/apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
index aa55be7ea..444212916 100644
--- a/apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
+++ b/apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
@@ -524,6 +524,7 @@ function ToolUseRow( {
                                <button
                                        type="button"
                                        className={ clsx( styles.toolRow, styles.toolRowButton ) }
+                                       aria-label={ display.detail ? `${ display.label } ${ display.detail }` : display.label }
                                        aria-expanded={ expanded }
                                        aria-controls={ detailsId }
                                        data-expanded={ expanded }
```

So that it would look something like this:

Image

Otherwise, it ends up being broken into two different rows:

Image

bcotrim (commented):
Solid, well-scoped PR — does exactly what STU-1873 part 2 calls for, and I didn't find any functional issues. A few non-blocking nits below; NIT 1 (helper duplication) is the only one I'd nudge on. 🙌

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
```diff
@@ -221,7 +262,229 @@ function AssistantText( { text }: { text: string } ) {
...
+		return value;
+	}
+	return value.slice( 0, maxLength - 1 ).trimEnd() + '…';
+}
+
+function shortPath( value: string ): string {
+	return value.split( '/' ).filter( Boolean ).slice( -2 ).join( '/' ) || value;
+}
+
+function splitCommandArgs( command: string ): string[] {
```
bcotrim: **NIT** `getInputString` (L273–276) and `splitCommandArgs` (L289–296) are byte-for-byte copies of the module-private versions in `tools/common/ai/tools.ts` (L3–6 and L8–15). `splitCommandArgs` especially is a non-trivial shell tokenizer — I'd rather not maintain two copies that can drift.

Suggest exporting both from `tools.ts` and importing here:

```diff
 import {
 	getToolDetail,
 	getToolDisplayName,
+	getInputString,
+	splitCommandArgs,
 	type NormalizedToolResult,
 } from '@studio/common/ai/tools';
```

…then drop the two local defs (L273–296).

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
```diff
@@ -221,7 +262,229 @@ function AssistantText( { text }: { text: string } ) {
...
+			return trash;
+		case 'wp_cli': {
+			const command = getInputString( input, 'command' );
+			return command ? getWpCliToolIcon( command ) : code;
+		}
+		case 'open_annotation_browser':
+			return pencil;
+		case 'wait_for_annotations':
+			return pending;
+		case 'AskUserQuestion':
```
bcotrim: **NIT:** `AskUserQuestion` is in `HIDDEN_TOOL_ROWS` (L106), so its row never renders and `getToolIcon` is never reached for it — this branch is dead. Safe to drop, along with the now-unused `help` import (L27). Remove both together or eslint will flag the unused import:

```diff
-		case 'AskUserQuestion':
-			return help;
```
```diff
 	globe,
-	help,
 	image,
```

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
```diff
@@ -221,7 +262,229 @@ function AssistantText( { text }: { text: string } ) {
...
+			display.detail = '';
+			break;
+		case 'Bash':
+			display.label = __( 'Run terminal command' );
+			display.detail = '';
+			break;
+		case 'Read':
+		case 'Write':
+		case 'Edit':
+			display.detail = genericDetail ? shortPath( genericDetail ) : '';
```
bcotrim: **NIT (lowest priority — fine to skip):** `getToolDetail` already trims Read/Write/Edit paths to their last two segments (`tools.ts` L403–411), and `display.detail` is initialized to that `genericDetail` (L331). So `shortPath()` here re-trims an already-trimmed value — effectively a no-op. The only input it changes is a file one level below root (e.g. `/wp-config.php`, where it strips the leading slash), which won't happen for real project paths.

If that edge isn't intentional, this case and the `shortPath` helper (L285–287) can both go:

```diff
-		case 'Read':
-		case 'Write':
-		case 'Edit':
-			display.detail = genericDetail ? shortPath( genericDetail ) : '';
-			break;
```

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
```diff
@@ -81,75 +81,184 @@
...
 	padding: 2px 0;
 	min-width: 0;
+	overflow: hidden;
+	white-space: nowrap;
+}
+
+.toolRowButton {
+	--wp-ui-button-background-color: transparent;
+	--wp-ui-button-border-color: transparent;
+	--tool-row-active-background: var(--wpds-color-bg-interactive-neutral-weak-active, #ededed);
```
bcotrim, line 113: **NIT:** This is a plain `<button>`, not a wpds `<Button>`, so nothing consumes `--wp-ui-button-*` — the active background comes from `--tool-row-active-background` / `background-color` below. These two lines (and the matching `--wp-ui-button-background-color` on L129 in the `:hover`/`:focus-visible` block) can go:

```suggestion
.toolRowButton {
	--tool-row-active-background: var(--wpds-color-bg-interactive-neutral-weak-active, #ededed);
```

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
shaunandrews: Good catch; I've made it darker so it passes WCAG contrast ratios.

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
shaunandrews: Done in 05447d2d1. I added the combined aria-label using the label plus detail when present, and added a regression test for the accessible name.

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
shaunandrews: Done in 05447d2d1. I exported getInputString and splitCommandArgs from tools/common/ai/tools.ts, imported them here, and removed the local copies.

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
shaunandrews: Done in 05447d2d1. Removed the unreachable AskUserQuestion icon case and the unused help import.

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
shaunandrews: Done in 05447d2d1. I removed the extra shortPath() helper and the Read/Write/Edit special case, since getToolDetail() already returns the trimmed path.

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
shaunandrews: Done in 05447d2d1. Removed the unused --wp-ui-button-* vars and kept the active background driven by --tool-row-active-background.

bcotrim (approved):
LGTM 👍

PR #3905, opened 2026-06-19
STU-1874: Improve Ask User option rendering

shaunandrews:
image
image
image
image

## Related issues

- Fixes [STU-1874](https://linear.app/a8c/issue/STU-1874/studio-agentic-ui-improve-ask-user-option-rendering)
- Stacked on #3904

## How AI was used in this PR

Codex implemented the scoped Ask User rendering changes, added tests for historical answer handling, and ran the verification commands below. I reviewed the diff and adjusted the live answer cache update to avoid touching older duplicate questions.

## Proposed Changes

- Makes Ask User choices read as an ordered list with visible option descriptions, so users can compare the available choices without relying on hidden title text.
- Adds a batched-question flow that shows one active question at a time, keeps answered questions as compact summaries, and lets users revisit earlier answers while the batch is still pending.
- Keeps selected Ask User answers visible after submission and when reviewing historical sessions, including multi-question batches whose answers are stored as `ask_user` prompts.
- Improves the batched-question handoff with selected-answer acknowledgement, reduced layout bounce, and scroll clearance so the next question and options stay comfortably above the composer.
- Preserves accessible button semantics with `aria-pressed` for selected choices and description text connected via `aria-describedby`.

## Testing Instructions

- `npx eslint --fix apps/ui/src/data/queries/use-agent-run.tsx apps/ui/src/ui-classic/components/session-view/conversation/index.tsx apps/ui/src/ui-classic/components/session-view/conversation/index.test.ts tools/common/ai/sessions/entry-types.ts apps/ui/src/ui-classic/components/session-view/conversation/style.module.css`
  - CSS module reports the expected "File ignored because no matching configuration was supplied" warning.
- `npm run typecheck`
- `npm test -- tools/common/ai/tests/tools.test.ts apps/cli/ai/tests/ui.test.ts apps/ui/src/ui-desks/chats/conversation/index.test.ts apps/ui/src/ui-classic/components/session-view/conversation/index.test.ts`
- Verified the Ask User option list in both light and dark themes with a temporary Vite harness and Playwright screenshots.

## Pre-merge Checklist


apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
```diff
@@ -181,6 +232,8 @@ export function entriesToRenderItems( entries: SessionEntry[] ): RenderItem[] {
 				key: `${ entryIndex }:question`,
 				question: data.question,
 				options: data.options,
+				pickedLabel:
```
katinthehatsite: From what I understand, it seems that the `selectedLabel` is for what was picked right now and the second one from `findAskUserAnswerAfter` is for looking up answers that were saved before?

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
```diff
@@ -281,26 +281,113 @@
...
+	background-color: var(--wpds-color-bg-surface-neutral);
 	color: var(--wpds-color-fg-content-neutral-weak);
+	font-size: var(--wpds-typography-font-size-xs);
+	line-height: 1;
+	transition:
+		border-color 120ms ease,
+		color 120ms ease;
+}
+
+.questionOptionCheck {
```
katinthehatsite: For me, it is not present on hover when I hover over the option:

https://github.com/user-attachments/assets/837c80f0-b86e-42d3-8077-0b71ee1f7255

It happens when the assistant is asking multiple questions in one batch.

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
```diff
@@ -281,26 +281,113 @@
...
+	width: 14px;
+	height: 14px;
+	color: currentColor;
+	fill: currentColor;
+}
```
katinthehatsite: We could potentially save some code lines here:

```diff --git a/apps/ui/src/ui-classic/components/session-view/conversation/index.tsx b/
apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
index 06ef148fc..37620e377 100644
--- a/apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
+++ b/apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
@@ -643,7 +643,7 @@ function AgentQuestion( {
        <Icon
                icon={ check }
                size={ 14 }
-                className={ styles.questionOptionCheck }
+                style={ { fill: 'currentColor' } }
                aria-hidden="true"
        />
                                                                                ) : (
diff --git a/apps/ui/src/ui-classic/components/session-view/conversation/style.module.css b/apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
index 0097ce25c..b770af479 100644
--- a/apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
+++ b/apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
@@ -358,17 +358,6 @@
                color 120ms ease;
 }
 
-.questionOptionCheck {
-       width: 14px;
-       height: 14px;
-       color: currentColor;
-       fill: currentColor;
-}
-
-.questionOptionCheck :where(path:not([fill='none'])) {
-       fill: currentColor;
-}
-
 .questionOptionCopy {
        display: flex;
        flex-direction: column;
```

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
```diff
@@ -281,26 +281,113 @@
 	margin: 0;
 	padding: 0;
 	display: flex;
-	flex-wrap: wrap;
-	gap: var(--wpds-dimension-padding-md);
+	flex-direction: column;
+	gap: var(--wpds-dimension-padding-lg);
+	border-radius: var(--wpds-border-radius-lg);
```
bcotrim: NIT: no background/border/`overflow` here, so this `border-radius` doesn't render anything — looks like a leftover.

```suggestion
```

bcotrim (commented):
Reviewed end-to-end, dark and light mode, closed and re-opened the app. Everything works as described.
Just a few non-blocking nits inline. Nice work on this! 🚀

| Light | Dark |
|--------|--------|
| image | image |

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
```diff
@@ -281,26 +281,113 @@
...
+}
+
+.questionOptionCheck :where(path:not([fill='none'])) {
+	fill: currentColor;
+}
+
+.questionOptionCopy {
+	display: flex;
+	flex-direction: column;
+	gap: 0;
```
bcotrim: NIT (trivial): `gap: 0` is the flex default, can drop.

```suggestion
```

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
```diff
@@ -281,26 +281,113 @@
...
 }
 
 .questionOption:hover:not(:disabled),
 .questionOption:focus-visible:not(:disabled) {
-	text-decoration: underline;
+	outline: none;
 }
 
-.questionOption:disabled {
+.questionOption:focus:not(:focus-visible) {
```
bcotrim, line 472: NIT (non-blocking): these focus rules overlap with the global `button:focus:not(:focus-visible)` in `index.css` — a couple look redundant here (local rule already wins, button has `border: none`). Could slim down + a one-line comment on what they're taming.

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
```diff
@@ -106,6 +107,56 @@ interface PiToolResultLike {
 
 const HIDDEN_TOOL_ROWS = new Set( [ 'studio_present', 'AskUserQuestion' ] );
 
+function findAskUserAnswerAfter(
```
bcotrim: NIT: scans backward (batch position) *and* forward (answers), so `…After` undersells it — maybe `resolveBatchedAnswerForQuestion`. A one-liner on the "questions then answers, in order" assumption would help too.

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
bcotrim: I can replicate the same issue

shaunandrews: Good find!

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
shaunandrews: Yep, exactly. `selectedLabel` is the value stored on the question entry; the fallback handles older batched transcripts where answers are stored afterward. I renamed the helper and added a comment for that path.

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
shaunandrews: Fixed. The batched flow now keeps the selected state visible during acknowledgement/fold, then summarizes answered questions. Added coverage around that path too.

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
shaunandrews: Done. Swapped the icon to `fill: currentColor` inline and dropped the extra CSS.

apps/ui/src/ui-classic/components/session-view/conversation/style.module.css
shaunandrews: Kept the local focus rule because it prevents the global mouse-click background while preserving keyboard focus. Added a short comment so the intent is clearer.

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
shaunandrews: Done. Renamed this to `resolveBatchedAnswerForQuestion` and documented the positional lookup.

shaunandrews:
Thanks for the review, folks. I've fixed everything here. I also added some more work around batch questions. Appreciate another review. Thanks!

PR #3906, opened 2026-06-19
Fix native PHP site icon paths

shaunandrews:
## Related issues

- Fixes [STU-1886](https://linear.app/a8c/issue/STU-1886/native-php-sites-can-cache-invalid-duplicated-site-icon-paths)

## How AI was used in this PR

Codex investigated why a configured WordPress Site Icon was rendering as the generated fallback in the agentic UI sidebar, traced the issue to malformed cached native PHP icon paths, implemented the fix, and added regression coverage. I reviewed the diff and verified it locally with targeted tests and typecheck.

## Proposed Changes

- Fixes site icons for native PHP sites whose cached icon path points at a duplicated site root, so stopped sites with an existing WordPress Site Icon can show the real image again instead of the generated fallback.
- Prevents future malformed icon paths by resolving the WordPress attachment path relative to the active WordPress root rather than assuming the Playground `/wordpress` mount.
- Uses the same normalization for refreshed icon paths and cached icon paths, so existing broken appdata entries can recover without requiring a site start first.

## Testing Instructions

- `npx eslint --fix tools/common/lib/mu-plugins.ts apps/studio/src/site-server.ts apps/studio/src/lib/resolve-site-icon-path.ts apps/studio/src/ipc-handlers.ts apps/studio/src/tests/site-server.test.ts apps/studio/src/tests/ipc-handlers.test.ts`
- `npm test -- apps/studio/src/tests/site-server.test.ts apps/studio/src/tests/ipc-handlers.test.ts`
- `npm run typecheck`
- `git diff --check`

Manual check:

- Use a site with a configured WordPress Site Icon and a cached malformed `siteIconPath` that duplicates the site root.
- Restart Studio so the main-process IPC change is active.
- Confirm the agentic UI sidebar renders the configured Site Icon instead of the generated fallback, even while the site is stopped.

## Pre-merge Checklist


fredrikekelund:
Closing this in favor of https://github.com/Automattic/studio/pull/3945

PR #3912, opened 2026-06-22
Add IPv6 loopback hosts entry for custom domains to fix slow .local DNS on macOS

### LINKED ISSUE #3655: .local DNS resolution slow on MacOS
### Quick summary

There is a known issue with how MacOS resolves the .local domain which can add on average around 3 - 6 seconds to the load time.

[This answer on StackOverflow](https://stackoverflow.com/a/45331805) explains it well:

> ...is most likely the result of Bonjour IPv6 lookups being issued for DNS lookups on .local domains. These IPv6 lookups create a timeout delay until the original IPv4 DNS lookup is issued.

An easy fix would be to add an IPv6 entry to the etc/hosts file in addition to the IPv4 one:
```
127.0.0.1 my-project.local
::1 my-project.local
```

As far as I'm aware this is also the same thing that LocalWP does in the etc/hosts file when a new site is created.

### Steps to reproduce

1. Create a new site in Studio on MacOS
2. Make sure to use a custom domain name (e.g.: my-project.local)
3. Create a second site
4. This time, make sure **not** to use a custom domain name
5. Run the following command in a terminal to measure the time it takes to get a response from the first site:
`time curl -I http://my-project.local/ > /dev/null` (replace my-project.local with whatever domain name you chose at step 2)
6. Run the following command in a terminal to measure the time it takes to get a response from the second site:
`time curl -I http://localhost:8882/ > /dev/null` (replace 8882 with the actual port of your second site)
7. Compare the results

I have also tried the proposed fix by editing my etc/hosts file to add the IPv6 entry. For me, this has reduced the time for the curl command from 5.6s to 0.6s.

The DNS resolution should be fast.

### What actually happened

The DNS resolution was slow.

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Platform

Mac Silicon

### Logs or notes

_No response_

youknowriad:
## Related issues

- Fixes #3655

## How AI was used in this PR

Claude Code investigated the reported `.local` slowdown, located the hosts-file logic, implemented the IPv6 entry plus the migration path for existing sites, and wrote the tests. The author reviewed the diff and reasoning.

## Proposed Changes

On macOS, resolving a `.local` custom domain that only has an IPv4 entry in the hosts file can stall on an IPv6 (AAAA) lookup that has to time out before the IPv4 result is used — adding several seconds to every request for affected users (see #3655). This writes a matching `::1` loopback entry alongside the existing `127.0.0.1` entry for each custom domain, which is the standard remedy for this behavior.

Users with existing custom-domain sites are migrated automatically: the IPv6 entry is added the next time the site starts. Already-correct entries are left untouched, so no spurious elevated-permission prompt appears on startup.

The slowdown is environment-dependent (the linked issue tags it as affecting "Some (< 50%)" of users), but the extra entry is harmless for unaffected users and is the cure for those who hit it.

## Testing Instructions

On a macOS machine that exhibits the slowdown:

1. Create a Studio site with a custom `.local` domain.
2. Measure: `time curl -sI http://your-domain.local/ > /dev/null` — note the time.
3. With this change, restart the site so the hosts file is rewritten.
4. Confirm `/etc/hosts` now contains both `127.0.0.1 your-domain.local` and `::1 your-domain.local` inside the `# BEGIN/END WordPress Studio` block.
5. Re-measure with the same `curl` command — the time should drop substantially.
6. Delete the site and confirm both entries are removed (no orphaned `::1` line).

## Pre-merge Checklist

### DIFF
--- a/apps/cli/lib/hosts-file.ts
+++ b/apps/cli/lib/hosts-file.ts
@@ -64,16 +64,31 @@ export const writeHostsFile = async ( content: string ): Promise< void > => {
 };
 
 /**
- * Create a regular expression matching the hosts entry for a given domain:
+ * Create a regular expression matching the hosts entries for a given domain,
+ * covering both the IPv4 and IPv6 loopback addresses:
  *
  * 	127.0.0.1 foo.wp.cloud # Port 8000
+ * 	::1 foo.wp.cloud # Port 8000
  *
  * 	Remove backslashes as a security measure and escape regex special characters.
  */
 export function createHostsEntryPattern( domain: string ): RegExp {
 	const sanitizedDomain = domain.replace( /\\/g, '' );
 	const escapedDomain = escapeRegex( sanitizedDomain );
-	return new RegExp( `127\\.0\\.0\\.1\\s+${ escapedDomain }(\\s|$)`, 'i' );
+	return new RegExp( `(?:127\\.0\\.0\\.1|::1)\\s+${ escapedDomain }(\\s|$)`, 'i' );
+}
+
+/**
+ * Build the canonical hosts entries for a domain. We add both an IPv4 and an
+ * IPv6 loopback entry: without the explicit `::1` line, macOS issues an IPv6
+ * (AAAA) lookup for `.local` domains that has to time out before the IPv4
+ * result is used, adding several seconds to every request.
+ */
+function createHostsEntries( encodedDomain: string, port: number ): string[] {
+	return [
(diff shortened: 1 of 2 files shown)


wojtekn:
I couldn't notice any difference either, but the fix seems reasonable and harmless.

wojtekn (approved):

youknowriad:
I noticed local does the same as well.

PR #3913, opened 2026-06-22
Add MailPit email catching integration

### LINKED ISSUE #24: Feature Request: Catch all emails
### What

I would like Studio to catch all emails sent from it

### Why

Test form plugins etc

### How

You could use https://mailpit.axllent.org/

It could be as easy as setting the SMTP port and host to the one mailpit is using, and using mailpit as is.
> Mailpit listens by default on port 0.0.0.0:8025 for the web UI, and port 0.0.0.0:1025 for SMTP.

youknowriad:
## Related issues

Closes #24 

## How AI was used in this PR

This PR was opened with AI assistance to surface and revive existing work: an agent located the dormant `add-mailpit-integration` branch, mapped the implementation against Studio's two PHP runtimes, identified gaps, and drafted this description. **The implementation itself is entirely @ryanr14's** — all commits are authored by him and unchanged. No AI-generated code has been added on top.

## Proposed Changes

Adds local **email testing** to Studio, so messages sent by a site never leave the machine and can be inspected in a real inbox UI — the capability Studio is missing compared with tools like Local and Cove.

- Studio bundles and runs [Mailpit](https://github.com/axllent/mailpit) per site, exposing a web inbox plus an SMTP endpoint for manual testing.
- Outgoing `wp_mail()` messages (recipients, CC/BCC/Reply-To, custom headers, HTML/text body, attachments) are captured and routed to the local Mailpit inbox instead of being delivered.
- Capture works in **both** the WordPress Playground (PHP-WASM) and native-PHP runtimes by forwarding mail to Mailpit's HTTP send API from a must-use plugin, rather than relying on outbound SMTP (which PHP-WASM cannot do).
- The desktop app gains an "Email Inbox" shortcut on the site Overview and an "Email testing" section in Settings (open inbox, copyable SMTP host/port).

> ⚠️ Visual change: needs human review in light + dark mode.

### Known gaps before this is merge-ready

- Mailpit is currently enabled for **every new site** with no opt-in/opt-out flag, and existing sites are not backfilled.
- No CLI surface yet (no `studio mail` command; the inbox isn't referenced in `site status`) — capture works headless but is only discoverable from the desktop app.
- Branch predates ~1 month of `trunk` movement (incl. CLI subcommand restructuring) and needs a rebase + full typecheck/test pass.

## Testing Instructions

1. `npm run download-mailpit` (or a full `npm install`) to fetch the Mailpit binary for your platform.
2. Create a new site and start it.
3. Trigger an email (e.g. trigger a password reset, or `studio wp eval 'wp_mail("to@example.com","Hello","Body");'`).
4. Open the site's **Email Inbox** (Overview shortcut or Settings → Email testing) and confirm the message appears with correct subject/recipients/body.
5. Repeat under both runtimes (Playground and native PHP) if possible.

## Pre-merge Checklist

---


youknowriad:
For context, I think this is a strong feature, I didn't implement this myself, it's based on @ryanr14's work but I believe we should consider helping moving it forward. I've seen a lot of people request this. It's a gap compared to other tools.

fredrikekelund:
Acknowledging that I've seen this, but also that I won't be able to review it until next week.

sejas (commented):
Riad, thanks for moving this integration forward. I agree that this is a feature many users have requested and it will be useful for them.

I just noticed a couple of things:
- When deleting a site we might want to remove the emails related to it in ~/.studio/mailpit/${siteId}.db and similar files
- Running `nvm use && npm i && npm start` was not enough and resulted in `http://127.0.0.1:8025/` returning an error/empty page. It worked correctly after running `npm run download-mailpit` once.

I tried reseting the password of admin@localhost.com:
Image
Image

I'll be AFK tomorrow and next week, so I won't be able to review or help with this PR until I come back.

apps/cli/lib/mailpit.ts
```diff
@@ -0,0 +1,173 @@
...
+} from 'cli/lib/cli-config/core';
+import { ensureMailpitBinaryAvailable } from 'cli/lib/dependency-management/mailpit-binary';
+
+function reserveConfiguredPorts( sites: SiteData[] ): void {
+	for ( const site of sites ) {
+		portFinder.addUnavailablePort( site.port );
+		portFinder.addUnavailablePort( site.mailpit?.httpPort );
+		portFinder.addUnavailablePort( site.mailpit?.smtpPort );
+	}
+}
```
fredrikekelund, line 26: I understand the need for this, but IMO it should sit much higher in the call hierarchy. Anytime `portFinder` tries to find a new port, we should ensure we've loaded the entire site list first and called `addUnavailablePort()` for all site ports and Mailpit ports.

tools/common/lib/mu-plugins.ts
```diff
@@ -36,6 +38,296 @@ const NATIVE_PHP_EXCLUDED_MU_PLUGINS = new Set( [
...
+				return null;
+			}
+
+			$status = wp_remote_retrieve_response_code( $response );
+			if ( $status >= 200 && $status < 300 ) {
+				return true;
+			}
+
+			return null;
+		}, 10, 2 );
```
fredrikekelund, line 326: Is it better for Studio to parse the incoming message and pass it to Mailpit via the HTTP API, rather than adding a `phpmailer_init` filter here to configure the site to send email through Mailpit's SMTP port? 

Using the `phpmailer_init` filter would be a lot more compact. Parsing the message like we do here also comes with an inherent risk of misrepresenting something. Leaving the message intact and just changing the SMTP sender seems preferable.

fredrikekelund (commented):
A few thoughts:

1. This is pretty great overall 👍 Kudos @ryanr14. There are a few rough edges in the implementation that I would love to pick up and sand down, but the big picture is looking really good to me: Mailpit looks like a solid choice, it's easy for us to distribute, the implementation for spinning it up is decent and it seems to integrate well in my quick testing. 
2. There needs to be a way to access Mailpit in the Studio UI. In Studio's agentic UI, the `PreviewAddressBar` component (at the top of the browser pane) seems like a logical choice. In the classic UI, we can add a button in the Overview tab (the classic UI is likely being phased out soon, yes, but it's a very simple change). 
3. I left an inline comment about the mu-plugin logic that actually passes emails on to Mailpit. The bottom line is that I think it would be preferable to use the `phpmailer_init` filter instead. 
4. I tried sending an email with Mailpoet, and it never showed up in Mailpit (confusingly similar names, btw). This may be a problem with the mu-plugin logic, or some Mailpoet-specific fluke. I'm happy to dig deeper.
5. This is ultimately mostly a cosmetic concern, but it'd be nice if Mailpit was available at `http://localhost:SITE_PORT/mailpit` (or whatever custom domain the site is using). This is easy for native PHP sites or any site with a custom domain, but Playground sites without a custom URL are trickier. Not something we have to block this PR (or even releasing Mailpit) on.

PR #3920, opened 2026-06-22
Fix import e2e test broken by add-site modal redesign

gcsecsey:
## Related issues

- Related to STU-1863

## How AI was used in this PR

Claude fixed the test while measuring e2e runtimes for STU-1863. The import spec was the only one that couldn't be timed because it failed before reaching the import. All changes were reviewed by a human.

## Proposed Changes

`import.test.ts` skips unless a `jetpack-backup.tar.gz` fixture is present, so this breakage was invisible in CI. With the fixture supplied locally, the test still failed, it's been broken since the add-site modal redesign.

This PR updates the test so that, when the fixture is present, it exercises the full import flow. On CI, this test is still skipped.

## Testing Instructions

1. Download `jetpack-backup.tar.gz` from 39997-pb into `apps/studio/e2e/imports/`.
2. Build the app: `npm run package`
3. Run: `npx playwright test apps/studio/e2e/import.test.ts`, the test should pass

| Before | After |
|--------|--------|
| CleanShot 2026-06-22 at 16 11 33@2x | CleanShot 2026-06-22 at 16 07 59@2x | 

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? (lint clean; test passes)
- [ ] Have you considered if this change needs tests, and added them if appropriate? (this *is* a test fix)


katinthehatsite (approved):
The changes look good to me and make sense. I noticed that some tests are failing so I rebased trunk and hopefully this will resolve the errors.

apps/studio/e2e/page-objects/add-site-modal.ts
```diff
@@ -19,7 +19,11 @@ export default class AddSiteModal {
 	}
 
 	get importButton() {
-		return this.page.locator( 'button:has-text("Import from a backup")' ).first();
+		// The "Import from a backup" option is a drop-zone card (a <div>), not a
```
katinthehatsite: I think these comments can be skipped but it is not a strong preference. We can leave them in if you prefer

gcsecsey: I was also hesitant about whether to commit these comments. I'll just remove them.

PR #3922, opened 2026-06-22
Set Playground runtime on reprint-pulled sites to fix startup DB error

epeicher:
## Related issues

- Related to STU-1828

## How AI was used in this PR

Claude Code diagnosed the failure from the Studio + daemon logs, reproduced the root cause with the bundled PHP binary, implemented the fix and a regression test, and verified it end-to-end with a real `pull-reprint`. I reviewed the diagnosis and change.

## Proposed Changes

Sites imported with `studio pull-reprint` failed to start, dying with **"Error establishing a database connection."**

The bug: a reprint-pulled site's files are laid out for the **Playground** runtime — VFS mounts place `wp-content` (and its SQLite `db.php` drop-in) under `/wordpress/wp-content`. But the site was registered without a runtime, so it defaulted to **native-php**. On that runtime PHP resolves the symlinked layout to the real WordPress core path, `WP_CONTENT_DIR` no longer points at the SQLite drop-in, WordPress falls back to MySQL, and the connection fails.

The fix: register reprint-pulled sites with the Playground runtime (the runtime they're already built for and the one the command uses for the auto-login URL), so they start correctly.

## Testing Instructions

1. `npm run cli:build`
2. `node apps/cli/dist/cli/main.mjs pull-reprint --url <your-site>.wpcomstaging.com -y`
3. The site starts and serves HTTP 200. 
4. Stop it by running `studio stop --path ...` and start it by running `studio start --path ...`
5. Verify it starts. On `trunk` the site does not start successfully

## Pre-merge Checklist


wojtekn:
Is it the correct approach? What if we remove support for Playground?

epeicher:
> Is it the correct approach? What if we remove support for Playground?

Good catch! I came here to comment that this approach will be superseded by https://github.com/Automattic/studio/pull/3881 - [comment here](https://github.com/Automattic/studio/pull/3881#pullrequestreview-4551773602) so I am closing this PR

PR #3923, opened 2026-06-22
E2E tests: Add tests to cover IPC bridge

gcsecsey:
## Related issues

- Related to STU-1864

## How AI was used in this PR

I used Claude to investigate the IPC architecture, confirm the IPC surface is type-derived from a single source of truth, and scaffold the smoke-test PoC. I steered the investigation and verified the tests.

## Proposed Changes

This is part of the STU-1864 spike on how to keep the IPC layer covered once the E2E suite moves to the CLI and the remaining UI tests get mocked at the IPC boundary.

During the spike, I found that:

- The IPC surface is already a single source of truth enforced by the type checker. The renderer, preload, and main-process registration all derive from the handler exports.
- The only thing types cannot prove is runtime wiring (contextBridge plus the `invoke`/`handle` round-trip and serialization across the process boundary), which a couple of round-trip smoke tests cover cheaply.

My recommendation is to rely on the type checker as the primary defense, and add a tiny IPC smoke suite for runtime wiring. This PR adds the small smoke test suite.

## Testing Instructions

- CI should pass, running `npm run typecheck` and the new test suite

### Optionally, break the test locally

- Apply this diff (`pbpaste | git apply`) that's type-valid but crashes at runtime on a Mac:
```diff
diff --git a/apps/studio/src/ipc-handlers.ts b/apps/studio/src/ipc-handlers.ts
index f6a13ac62..cb5886a57 100644
--- a/apps/studio/src/ipc-handlers.ts
+++ b/apps/studio/src/ipc-handlers.ts
@@ -1331,7 +1331,7 @@ export function copyText( event: IpcMainInvokeEvent, text: string ) {
 
 export function getAppGlobals(): AppGlobals {
 	return {
-		platform: process.platform,
+		platform: 'linux',
 		appName: app.name,
 		appVersion: app.getVersion(),
 		arm64Translation: app.runningUnderARM64Translation,
EOF
```
- Package the app with `npm run package`
- Run the tests with `npx playwright test ipc-bridge`
- Confirm the test fails

| With HEAD version | With the diff applied |
|--------|--------|
| CleanShot 2026-06-23 at 11 48 23@2x | CleanShot 2026-06-23 at 12 16 48@2x | 

## Pre-merge Checklist


gavande1 (approved):
Thanks Gergely for working on this, and great investigation. The minimal change is great for us. LGTM 🚀

PR #3926, opened 2026-06-23
Learn more links under "PHP runtime" controls

fredrikekelund:
## Related issues

- Fixes RSM-4475

## How AI was used in this PR

I asked Claude to do the implementation mostly because I didn't know by heart where the site controls live.

## Proposed Changes

Add "Learn more" links to the create site and edit site modals:

| Add site modal | Edit site modal |
| - | - |
| Screenshot 2026-06-23 at 13 58 56 | Screenshot 2026-06-23 at 13 59 07 |

## Testing Instructions

1. `npm start`
2. Add a new site
3. Open the advanced settings control
4. Ensure that the "Learn more" link under "PHP runtime" leads to the correct documentation page (privately published for now)
5. Open the "Site settings" tab for an existing site
6. Click the "Edit site" button
7. Ensure that the "Learn more" link under "PHP runtime" leads to the correct documentation page (privately published for now) 

## Pre-merge Checklist

### DIFF
--- a/apps/studio/src/modules/add-site/components/create-site-form.tsx
+++ b/apps/studio/src/modules/add-site/components/create-site-form.tsx
@@ -37,7 +37,7 @@ import { SiteFormError } from 'src/components/site-form-error';
 import TextControlComponent from 'src/components/text-control';
 import { WPVersionSelector } from 'src/components/wp-version-selector';
 import { cx } from 'src/lib/cx';
-import { FileAccessDescription } from 'src/lib/site-runtime-copy';
+import { FileAccessDescription, RuntimeDescription } from 'src/lib/site-runtime-copy';
 import { useCheckCertificateTrustQuery } from 'src/stores/certificate-trust-api';
 import type { BlueprintPreferredVersions } from '@studio/common/lib/blueprint-validation';
 import type { CreateSiteFormValues, PathValidationResult } from 'src/hooks/use-add-site';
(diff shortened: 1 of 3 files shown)


apps/studio/src/modules/add-site/components/create-site-form.tsx
```diff
@@ -545,7 +545,10 @@ export const CreateSiteForm = ( {
 										<span className="text-frame-text-secondary text-xs">
 											{ selectedRuntime === SITE_RUNTIME_NATIVE_PHP
 												? __( 'Runs the site with native PHP for the best performance.' )
-												: __( 'Runs the site in an isolated WordPress Playground sandbox.' ) }
+												: __(
+														'Runs the site in an isolated WordPress Playground sandbox.'
+												  ) }{ ' ' }
+											<LearnMoreLink docsLinksKey="docsPhpRuntimes" />
```
wojtekn: Let's use `createInterpolateElement` to avoid interpolating translatable texts.

fredrikekelund: We aren't interpolating anything into the `__()` call here. I see we're using `createInterpolateElement` elsewhere in this file, but the practical implications of both approaches seem identical to me (apart from `<lean_more_link />` being included in the translatable string with `createInterpolateElement`).

wojtekn: My bad, I meant concatenating. See PCYsg-3hI-p2

> but the practical implications of both approaches seem identical to me 

Not really - if we interpolate instead concatenate, translator could reorder elements if particular language requires that, could add text after "Learn more", wrap it with some characters etc.

fredrikekelund: That's actually a good point 👍 I pushed a fix

wojtekn: Thanks! 🙌

wojtekn (approved):

PR #3927, opened 2026-06-23
Build gd with additional libraries

fredrikekelund:
## Related issues

- Fixes RSM-4476

## How AI was used in this PR

Scarcely

## Proposed Changes

static-php-cli builds gd without libjpeg by default. See [docs](https://github.com/crazywhalecc/static-php-cli/blob/b4ed673261548dcc0caa5b61ed782de1ccbe971f/docs/en/guide/extension-notes.md#gd). This means WordPress can't edit these images, which is a critical bug.

This PR fixes the problem by passing `--with-libs=freetype,libjpeg,libwebp` to static-php-cli when building the binaries.

## Testing Instructions

I'll manually verify that this gives the result we want.

## Pre-merge Checklist

### DIFF
--- a/.github/workflows/build-php-cli-binaries.yml
+++ b/.github/workflows/build-php-cli-binaries.yml
@@ -175,6 +175,8 @@ jobs:
           )
           build_args=(
             "$static_extensions"
+            # Give gd JPEG/WebP image support and FreeType (TrueType text).
+            --with-libs=freetype,libjpeg,libwebp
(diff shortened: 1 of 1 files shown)


fredrikekelund:
I've verified that [the gd build](https://github.com/Automattic/studio/actions/runs/28029716706) now supports JPEG, like so:

```
./php -r 'echo gd_info()["JPEG Support"];'
```

PR #3928, opened 2026-06-23
Dev-stamp the standalone CLI version in nightly builds

bcotrim:
## Related issues

- Related to [STU-1772](https://linear.app/a8c/issue/STU-1772) — the standalone CLI update workflow depends on this.

## How AI was used in this PR

Caught during real-world testing: installing a published nightly CLI showed `studio --version` → `1.11.0`, even though the Apps CDN publishes that same nightly as `1.12.0-dev103`. Traced it to `prepare-dev-build-version.mjs` dev-stamping only `apps/studio/package.json`. I reviewed the change myself.

## Proposed Changes

`prepare-dev-build-version.mjs` rewrites the version to the `X.(Y+1).0-devN` dev form for nightly/dev builds — but only for the desktop app (`apps/studio/package.json`). The standalone CLI (`apps/cli/package.json`) was left at the static base version, so its baked `__STUDIO_CLI_VERSION__` (what `studio --version` reports) didn't match the version the bundle is actually published under on the CDN.

This stamps `apps/cli/package.json` with the same dev version. After it:
- `studio --version` on a nightly reports its real version (e.g. `1.12.0-dev103`) instead of a static, meaningless `1.11.0`.
- It unblocks the standalone CLI update notifier (STU-1772): the CLI derives its update channel from its own version (`-devN` → nightly), so it needs the real dev version baked in to check the right channel. Without this, a nightly install looks like "production" to the update endpoint and never sees nightly updates.

## Testing Instructions

```sh
GITHUB_SHA=$(git rev-parse HEAD) node ./scripts/prepare-dev-build-version.mjs
grep '"version"' apps/studio/package.json apps/cli/package.json
# → both show the same X.(Y+1).0-devN (e.g. 1.12.0-dev103), matching the CDN-published nightly
git checkout -- apps/studio/package.json apps/cli/package.json   # revert the local mutation
```

The CLI build (`cli:bundle`) runs after this script in the build pipelines, so the bundle bakes the stamped version.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? — build script only; ran it and verified both files stamp to the same dev version.

### DIFF
--- a/scripts/prepare-dev-build-version.mjs
+++ b/scripts/prepare-dev-build-version.mjs
@@ -20,7 +20,9 @@ if ( ! commitCount && commitCount !== 0 ) {
 // Build a dev version targeting the next minor release (major.(minor+1).0-devN) from a base
 // version, so trunk builds sort above any stable or beta of that base. Strips a leading 'v'.
 function toDevVersion( baseVersion, source ) {
-	const parsed = semver.parse( baseVersion.startsWith( 'v' ) ? baseVersion.slice( 1 ) : baseVersion );
+	const parsed = semver.parse(
+		baseVersion.startsWith( 'v' ) ? baseVersion.slice( 1 ) : baseVersion
+	);
 	if ( ! parsed ) {
 		throw new Error( `Invalid version in ${ source }: ${ baseVersion }` );
 	}
--- a/scripts/prepare-dev-build-version.mjs
+++ b/scripts/prepare-dev-build-version.mjs
@@ -17,27 +17,26 @@ if ( ! commitCount && commitCount !== 0 ) {
 	throw new Error( 'Missing commit count' );
 }
 
-// Use version from latestTag (strip leading 'v' if present)
-const tagVersion = latestTag.startsWith( 'v' ) ? latestTag.slice( 1 ) : latestTag;
-const parsedVersion = semver.parse( tagVersion );
-if ( ! parsedVersion ) {
-	throw new Error( `Invalid version in latestTag: ${ latestTag }` );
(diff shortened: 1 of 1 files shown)


scripts/prepare-dev-build-version.mjs
```diff
@@ -32,3 +32,12 @@ packageJson.version = devVersion;
...
 const packageJsonPath = path.resolve( 'apps', 'studio', 'package.json' );
 await fs.writeFile( packageJsonPath, JSON.stringify( packageJson, null, '\t' ) + '\n' );
+
+// Also stamp the standalone Studio CLI so its baked `__STUDIO_CLI_VERSION__` matches this dev
+// build. The CLI's version drives its update channel (a `-devN` version → nightly), so without
+// this a nightly bundle reports the static base version, looks like production to the update
+// endpoint, and never sees nightly updates.
+const cliPackageJsonPath = path.resolve( 'apps', 'cli', 'package.json' );
+const cliPackageJson = JSON.parse( await fs.readFile( cliPackageJsonPath, 'utf8' ) );
+cliPackageJson.version = devVersion;
```
fredrikekelund: There's no guarantee that the `apps/cli` version is the same as the `apps/studio` version. 

Let's read the current version from `apps/cli/package.json` and add to it the same way we do with the version from `apps/studio/package.json`

fredrikekelund: I realize that this may not have been that big a concern given our versioning strategy (CLI-only releases are allowed to bump the patch version, and app releases always bump the minor version). Still, it makes sense for the code to be resilient

fredrikekelund (approved):
LGTM 👍

Issue #3766, opened 2026-06-10
Fatal errors after cloning sites from Dotcom


Nyiriland:
I've been trying to work locally on Team 51 sites, using Studio to clone the Staging site from Dotcom.  <br>I'm *not* using the nightly build version of Studio.

### Problem 1: database suddenly disconnected upon opening app

I had been working on a clone just fine until one day I opened up Studio and the site started throwing a "Error establishing a database connection" message.

Tried to use Claude to remedy:

```
- Found the issue — the db.php drop-in is missing. Let me verify the full picture.
- The db.php drop-in is missing from wp-content/ — that’s the file that intercepts WordPress’s database calls and routes  them to SQLite instead of MySQL. Without it, WordPress tries to connect to a MySQL server that doesn’t exist. I’ll create it by copying the template and filling in the correct path.
```

... but it reset the database instead of fixing it, so I bailed and decided to re-clone. (At the time, I didn't know I could have just tried pulling the database again from Dotcom.)

### Problem 2: fatal errors

Next, I thought it would make sense to clone Production, as I wanted to work with actual content, not placeholder.

It was super slow. Front-end eventually loaded, but wp-admin timed out:

```
Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 1323008 bytes) in /wordpress/wp-content/plugins/jetpack/vendor/jetpack-autoloader/class-php-autoloader.php on line 102

Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 1323008 bytes) in /wordpress/wp-includes/sodium_compat/autoload-php7.php on line 27
```

### Problem 3: export compatibility

I was curious if it was just the database size that Studio couldn't handle, so I tried to use its native Export feature and import into Local. This didn't work either (I now understand the formats are not compatible). As a user, I would expect the export format to work in other commonly used WordPress tools, but maybe this is asking too much.

### Success: re-cloning staging again

Creating a fresh clone of the Staging site worked and I was able to get up and running again.

### Problem 4: more fatal errors

After Staging had launched, we updated it to match Production's database. When I tried to re-clone Staging again, I ran into the same slowness I experienced in Problem 2.

* The front-end loads, although it takes a *very* long time (around a minute)
* Loading `wp-admin` causes my laptop fan to run excessively hard (which I've never heard before on my M4 mac) and gives me a critical error:

[...]

PR #3936, opened 2026-06-23
Increase Playground runtime PHP `memory_limit` to 512M

gcsecsey:
## Related issues

- Related to #3766
- Related to STU-1821

This PR addresses "Problem 2 (fatal errors / memory exhaustion)" from that report.

## How AI was used in this PR

Used Claude Code to reproduce the wp-admin OOM locally, trace it to the Playground runtime's memory cap, and verify the fix against a heavy plugin stack.

## Proposed Changes

The reported issue is that when a site with numerous plugins is cloned into Studio, and then run with the Playground runtime, the front-end loads but wp-admin dies with `Fatal error: Allowed memory size of 268435456 bytes exhausted` (256 MB).

The PHP-WASM HTTP server serves every request with a hard-coded PHP `memory_limit` of 256 MB. A realistic production plugin stack (WooCommerce, WooPayments, Jetpack, Jetpack Boost, Jetpack CRM, MailPoet, Site Kit, etc.) already peaks at ~224 MB on an empty database, so wp-admin tips over 256 MB when the site's plugins and data load.

This PR raises the Playground serving `memory_limit` to 512 MB, matching the native runtime, so the two behave consistently, and heavier sites can open wp-admin.

| trunk | this branch |
|--------|--------|
| CleanShot 2026-06-23 at 17 02 48@2x | CleanShot 2026-06-23 at 16 59 14@2x | 

## Testing Instructions

- Visual review should be enough

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? (`npm run typecheck` passes; eslint clean on the changed file.)

### DIFF
--- a/apps/cli/playground-server-child.ts
+++ b/apps/cli/playground-server-child.ts
@@ -469,10 +469,6 @@ const startServer = wrapWithStartingPromise(
 			await setPhpIniEntries( server.playground, {
 				'openssl.cafile': '/internal/shared/ca-bundle.crt',
 				'curl.cainfo': '/internal/shared/ca-bundle.crt',
-				// Match the native PHP runtime's 512M (apps/cli/lib/native-php/config.ts). The
-				// PHP-WASM default of 256M is exhausted by real WooCommerce + Jetpack sites when
-				// loading wp-admin (STU-1821). memory_limit is a per-request ceiling, not a
-				// reservation, so this does not increase baseline host memory use.
 				memory_limit: '512M',
 			} );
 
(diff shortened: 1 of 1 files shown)


gavande1 (approved):
Thanks, Gergely. The code changes look good. I am curious: Do we really need to increase memory? 256 MB should be enough.

I am approving the PR to unblock you.

gcsecsey:
> Thanks, Gergely. The code changes look good. I am curious: Do we really need to increase memory? 256 MB should be enough.
> 
> I am approving the PR to unblock you.

Thanks! Yes, I think the setup I used was pretty reasonable. A normal Woo site would also have this many plugins and might hit local memory limits. The original reporter also hit the memory limit when cloning their site to Studio. Also, just as a side note, raising the limit to 512M for Playground is matching the limit we already have in place for native PHP.

PR #3940, opened 2026-06-23
Studio Code: Add data liberation slash command 

nightnei:
## Related issues

- Fixes STU-1510

## How AI was used in this PR

The solution is completely mine and is written partially manually.
AI helped with writing code.

## Proposed Changes
We are adding `/liberate` command for Studio Code. Which, under the hood, uses https://github.com/Automattic/data-liberation-agent
It allows the creation of websites based on closed platforms. See example below.
Note: I have created dozens of sites, and I have occasionally encountered imperfect results with broken pages, but those cases were rare. Keep in mind that if the result doesn’t look right, it may be worth trying again. For example, in the screenshot below, the page generated through `Block → data-liberation-agent → /liberate → Home page` doesn’t look as polished as the result generated with Studio Code.
Also, it's not connected to the PR - `Theme replication` works much better than `Blocks`. JFYI.

### Theme replication
<table>
<tr>
<td valign="top" width="15%">Studio Code /liberate
<td width="35%">localhost_8883_111111
<td width="50%" valign="top">localhost_8883_blog_ (5)
</tr>
<tr>
<td valign="top">data-liberation-agent /liberate
<td>gfgfgf
<td valign="top">localhost_8883_blog_ (3)
</tr>
</table>

### Blocks
<table>
<tr>
<td valign="top" width="15%">Studio Code /liberate
<td width="35%">localhost_8883_ (3)
<td width="50%" valign="top">localhost_8883_blog_ (1)
</tr>
<tr>
<td valign="top">data-liberation-agent /liberate
<td>localhost_8883_ (1)
<td  valign="top">localhost_8883_blog_
</tr>
</table>

## Testing Instructions
1. `node apps/cli/dist/cli/main.mjs code`
2. `/liberate https://neicss.wixsite.com/sunflower-valley-exp` or you can use own site. I was testing with this my sunflower-valley-exp site.
3. Choose `Migrate content into blocks`
4. Verify that the created site looks decent and that you didn't have errors during the process of site creation
5. Remove the site in Studio and also remove the folder from `~/Studio/_liberations/<your_site>`
6. Run `/liberate https://neicss.wixsite.com/sunflower-valley-exp` again
7. This time choose `Theme replication`
8. Assert the same


apps/cli/ai/tools/data-liberation.ts
```diff
@@ -0,0 +1,244 @@
...
+import path from 'path';
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';
+import { Type } from 'typebox';
+import { STUDIO_SITES_ROOT } from 'cli/lib/site-paths';
+import { defineTool } from './define-tool';
+import { textResult } from './utils';
+
+const ENGINE_DIR = path.join( STUDIO_SITES_ROOT, '_liberations', 'data-liberation' );
+const ENGINE_REPO = 'https://github.com/Automattic/data-liberation-agent.git';
```
youknowriad: IS this how we want to consume this, feels very fragile to me. I'd have preferred if that liberation agent becomes just a package in the Studio repository personally.

youknowriad: (shouldn't stop this PR from shipping if we think it's producing good results but I believe it's something important to solve. @borkweb was also thinking about these things.

apps/cli/ai/tools/data-liberation.ts
```diff
@@ -0,0 +1,244 @@
...
+
+	await fs.mkdir( path.dirname( ENGINE_DIR ), { recursive: true } );
+
+	await runProcess(
+		'git',
+		[ 'clone', '--depth', '1', '--branch', 'main', ENGINE_REPO, ENGINE_DIR ],
+		STUDIO_SITES_ROOT
+	);
+
+	await runProcess( 'npm', [ 'ci' ], ENGINE_DIR );
```
sejas: What if the user doesn't have `git` or `npm` installed?

sejas (commented):
Vova, thanks for exploring the integration of the Liberate skill into Studio ✨.

On Riad’s note, I see that we require users to have `git` and `npm` installed on users' machines, and running `npm ci` could also introduce Node version compatibility issues. Do you think we could explore adding Liberate as a package?

apps/cli/ai/tools/data-liberation.ts
youknowriad: Also related to this, "data-liberation" is not the only "tool/package" that would benefit from this, we have the taxonomist too which is very similar.

nightnei:
> feels very fragile to me

Makes sense

> I'd have preferred if that liberation agent becomes just a package in the Studio repository personally.

@youknowriad, what do you mean by "package"? To download the GitHub repo during build time, build it, and include it in the Studio bundle?

youknowriad:
> @youknowriad, what do you mean by "package"? To download the GitHub repo during build time, build it, and include it in the Studio bundle?

Ideall for me is a "packages" folder in Studio repo like we have "apps" with reusable npm packages, that we can ship independently if needed but also bundle in other "apps" (like CLI)

nightnei:
> > @youknowriad, what do you mean by "package"? To download the GitHub repo during build time, build it, and include it in the Studio bundle?
> 
> Ideall for me is a "packages" folder in Studio repo like we have "apps" with reusable npm packages, that we can ship independently if needed but also bundle in other "apps" (like CLI)

@youknowriad do you propose to remove https://github.com/Automattic/data-liberation-agent github repo and move all the code into <this_studio_github_repo>/packages/data-liberation-agent folder?

youknowriad:
Yes

nightnei:
@youknowriad @sejas @borkweb I propose to align and agree the solution.

Is everybody okay with archiving https://github.com/Automattic/data-liberation-agent and moving the code into `studio/packages/data-liberation-agent`?
Personally, I am ok, but my initial intention was - keep the repo separately and download it (or npm package) at runtime when somebody runs `/liberate` for the first time. The main goal is to avoid increasing Studio's bundle size.

youknowriad:
Do we know how much kb would add the addition of the data liberation agent to Studio CLI. My initial impression is that it should be negligible but I could be wrong.

nightnei:
> Do we know how much kb would add the addition of the data liberation agent to Studio CLI. My initial impression is that it should be negligible but I could be wrong.

There are many dependencies - https://github.com/Automattic/data-liberation-agent/blob/main/package.json
I started testing the approach by cloning the repo to `studio/packages/...`.
And at a glance, it's heavy:
Screenshot 2026-06-24 at 17 15 23

I asked AI to analyze deps and there is good news about `@wp-playground/cli` (I checked - it's indeed unused):
Screenshot 2026-06-24 at 17 16 53
So, it's difficult to say the end result, but it will definitely be mb, not kb.

Screenshot 2026-06-24 at 17 23 10

youknowriad:
My guess is that the biggest chunk of these dependencies is already a dependency of Studio CLI

sejas:
> Is everybody okay with archiving https://github.com/Automattic/data-liberation-agent and moving the code into `studio/packages/data-liberation-agent`?

I'm ok with that, although I can only answer with the part of accepting the package into the Studio repository. I don't have the context to talk about the future of the https://github.com/Automattic/data-liberation-agent repository.

nightnei:
@youknowriad, one more question - do you think Matthew Batchelder’s approval is enough for us to move the repo into Studio, or can we proceed without waiting for a response? Also, do we need Matt’s approval since he created the repo?

youknowriad:
@nightnei Let's do a P2 post explaining the decision and the reasoning... and ping both Matt, @borkweb and any stakeholders that might need to jump in.

apps/cli/ai/skills/liberate/SKILL.md
```diff
@@ -0,0 +1,31 @@
+---
+name: liberate
+description: Import and rebuild a website from a closed platform (Wix, Squarespace, Webflow, Shopify, GoDaddy, Hostinger, HubSpot, Weebly) into a Studio WordPress site. Extracts pages/posts/products + media, then reconstructs the design as editable blocks + WooCommerce OR as a high-fidelity replica theme. Invoke when the user wants to migrate, import, liberate, or rebuild a site from one of these platforms.
+---
+
+# Liberate a website into Studio
+
+This skill is only a **redirect**. The real, always-up-to-date pipeline lives in the Data Liberation engine's own skill, which Studio downloads on first use. Do NOT re-plan or summarize the steps here — defer to the engine skill so its updates take effect automatically. Your job is just to (1) stand the engine up and (2) follow its skill, translating its tool calls into Studio's bridge.
```
borkweb:
```suggestion
This skill is only a **redirect**. The real, always-up-to-date pipeline lives in the Data Liberation engine's own skill. Do NOT re-plan or summarize the steps here — defer to the engine skill so its updates take effect automatically. Your job is just to (1) stand the engine up and (2) follow its skill, translating its tool calls into Studio's bridge.
```

borkweb (changes requested):
Just one small change with one small thought:

In the data-liberation-agent, there's an alias to the `/liberate` skill called `/migrate` that I added because the word liberate is probably not something that folks would easily remember. The migrate Skill merely invokes the liberate Skill.

I wonder if we should do similarly here—adding a `/migrate` slash command that merely calls liberate.

Not 100% necessary, but might be a good usability boost.

nightnei:
> Just one small change with one small thought:
> 
> In the data-liberation-agent, there's an alias to the `/liberate` skill called `/migrate` that I added because the word liberate is probably not something that folks would easily remember. The migrate Skill merely invokes the liberate Skill.
> 
> I wonder if we should do similarly here—adding a `/migrate` slash command that merely calls liberate.
> 
> Not 100% necessary, but might be a good usability boost.

I completely agree. I think /migrate is easier to remember and understand at a glance than /liberate.
UPDATE: I created linear issue (STU-1973) to unblock this PR. @sejas @youknowriad share please your opinions.

youknowriad:
I have no opinion on the name, I'm also ok with adding "aliases" support to slash commands.

nightnei:
> I have no opinion on the name, I'm also ok with adding "aliases" support to slash commands.

Thanks @youknowriad

JFYI, data-liberation-agent has already been [merged](https://github.com/Automattic/studio/tree/trunk/packages/data-liberation-agent) into the Studio repo. It now reuses the shared node_modules with Studio CLI, so there’s no duplication anymore. There are still a few issues in the flow that I’m investigating, but other than that, it works and this PR is ready for review, when you have time.

youknowriad:
I'm testing and I'll share my findings as I go through it.

The first thing is that after running the liberate_extract, I got a timeout and it told me, it's fine because it's running in the background and then every second I'm seeing

> ⏺ Data Liberation liberate_status                                      
{"outputDir":"/Users/riad/Studio/_liberations/www.studio-chocol… (0.0s) 
   ⎿ {                                                                  
       "running": true,                                                 
       "processed": 6,                                                  
       "remaining": 0,                                                  
       "failed": 0,                                                     
     ... 4 more lines · ctrl+o to expand 
     
That doesn't feel like a great user experience, instead I'd prefer a loader/spinner like we do for import/export.

youknowriad:
And that tool call on loop is probably consuming a lot of tokens as well.

nightnei:
> I'm testing and I'll share my findings as I go through it.
> 
> The first thing is that after running the liberate_extract, I got a timeout and it told me, it's fine because it's running in the background and then every second I'm seeing
> 
> > ⏺ Data Liberation liberate_status
> > {"outputDir":"/Users/riad/Studio/_liberations/www.studio-chocol… (0.0s)
> > ⎿ {
> > "running": true,
> > "processed": 6,
> > "remaining": 0,
> > "failed": 0,
> > ... 4 more lines · ctrl+o to expand
> 
> That doesn't feel like a great user experience, instead I'd prefer a loader/spinner like we do for import/export.

Thanks, yeah, right now I am working on this and will improve/fix a few more.
I was thinking maybe to start with the initial PR and then work on follow-up PRs to fix such issues like this. Just to move things forward and avoid hanging PR.

youknowriad:
I'm totally ok with merging and iterating 👍

youknowriad:
I noticed the agent trips often on random WP CLI commands and things like that. 

Also for some reason (tried liberating this randomly found wix site https://www.studio-chocolate.co.uk) the result has nothing to do with the site (I picked block based recreation). Even the colors are different. I'll try again with "custom HTML option"

nightnei:
> I'm totally ok with merging and iterating 👍

Then, review please the approach (code) and assert that it works.
And I will work on follow up PR to polish everything.

nightnei:
I am also testing this website - https://www.studio-chocolate.co.uk/
It's huge, extraction takes a lot of time.

youknowriad:
Ok I tried the site again, this time I picked "custom css, theme fidelity" option and the result is very close to the original site, it's night and day compared to the other option.

youknowriad:
> Conversation history compacted 
> ⏺ Cannot continue from message role: assistant  

I also get this at the end, right after the summary message. It might not be specific to the data liberation though.

nightnei:
> Ok I tried the site again, this time I picked "custom css, theme fidelity" option and the result is very close to the original site, it's night and day compared to the other option.

The "blocks" reconstruction underperforms even in the standalone data-liberation-agent (without Studio Code as a proxy). It's the blocks path itself, which I've reproduced locally (example in the PR description). But I see ways to improve it.

apps/cli/ai/skills/liberate/SKILL.md
```diff
@@ -0,0 +1,31 @@
...
+
+Call the `data_liberation` tool with `{ tool: "setup" }`. It returns `{ engineDir, liberateSkill, skillsDir }` — the paths to the engine's skill files. The engine ships prebuilt with Studio, so this is instant — just proceed.
+
+## Step 2 — Load the engine's tool catalog
+
+Call `data_liberation` with `{ tool: "list" }` to load the engine's full tool catalog (every tool name + its argument JSON-Schema). Treat that catalog as the **source of truth** for tool names and arguments — the engine skill names the tools to call, but the catalog tells you their exact arguments. If you are ever unsure of an argument, re-read the catalog rather than guessing.
+
+## Step 3 — Follow the engine's real liberate skill
+
+`Read` the `liberateSkill` path from Step 1 (the engine's own `SKILL.md`) and execute its workflow verbatim, applying these Studio bindings wherever it instructs you:
```
youknowriad, line 20: This is a very weird way to use the data_liberation. If the data liberation packages bundles skills, can't we actually "import" them directly where we define the skills of Studio Code instead of having a "wrapper" skill?

youknowriad (commented):
I think it's ok to ship this PR as is but I'd love to know if there's better ways to actually embed the data liberation engine into studio Code, importing tools and skills directly instead of having "wrappers".

Like ideally, there skills and tools are no different than any other first party Studio Code tools and skills.

borkweb:
@youknowriad - I still think the right solution for this is to support AI plugins in Studio Code. No that we are on pi.dev, maybe implementing something like including https://pi.dev/packages/pi-claude-marketplace in order to consume the data-liberation stuff as a plugin - then we _shouldn't_ need a wrapper of any sort (though I haven't tested this)

youknowriad:
What kind of plugins do we expect to have outside data liberation and taxonomist?

nightnei:
> I think it's ok to ship this PR as is but I'd love to know if there's better ways to actually embed the data liberation engine into studio Code, importing tools and skills directly instead of having "wrappers".
> 
> Like ideally, there skills and tools are no different than any other first party Studio Code tools and skills.

@youknowriad, the single-wrapper approach here (`/liberate` as the one entrypoint, with the `data_liberation` bridge tool underneath) is a deliberate choice, for two reasons:

1. The engine's tools and skills aren't standalone — they only work as an orchestrated batch. The engine exposes ~40 MCP tools and ~24 skills that are meaningless individually: they must run in a specific sequence, share a common output directory, and depend on each other's artifacts. The only unit that stands on its own — the thing a user actually wants — is the full migration (`/liberate` command).
2. Merging them directly would flood Studio Code's tool/skill surface with ~60 interdependent pieces that no one can call in isolation. That's harder to maintain and more confusing for end users, who'd see dozens of liberate_* tools that only function as a group.

So I see two real options: the wrapper approach in this PR, or something like Matthew's plugin-system idea (AFAIU). A plugin system is the cleaner long-term answer if we expect several such embedded engines — but for a 1 (or maybe 2, also taxonomist ) engine today it feels like overkill. My proposal: ship the wrapper now, and revisit the plugin approach later if a second use case makes it worth the investment.

One more thing worth noting: the integration in this PR is thin — it's a small amount of glue code, and it requires almost no changes to the data-liberation-agent's own codebase. We consume it close to as-is, which means upstream updates flow in with minimal friction and we keep a clean seam between the two projects. Merging the tools/skills directly would break that seam: we'd have to reshape the engine's internals to fit Studio Code's conventions, which makes both codebases harder to maintain — the engine included.

borkweb:
> What kind of plugins do we expect to have outside data liberation and taxonomist?

@youknowriad - the site-builder library/plugin we're working on (extracting Telex site builder logic into something reusable).

Re: @nightnei - 
> The engine exposes ~40 MCP tools and ~24 skills that are meaningless individually.

Definitely meaningless outside the concept of liberation/migration. But definitely meaningful calling them individually if the user stops the liberation process partway and wants to resume. Or wants to retry certain steps. Most of those pieces are resumable and a handful of them are callable as single items in isolation to redo portions of the flow.

Re-calling portions to pick up from the middle saves a lot of time and tokens than always starting from the beginning from my experience!

youknowriad:
Sounds good to me but I think it's not a clear cut to be honest. For instance, we now have two "scaffold theme" skills, one within the data liberation and one outside and they probably duplicate some stuff, and do some other stuff differently. How can we avoid this

nightnei:
> But definitely meaningful calling them individually if the user stops the liberation process partway and wants to resume. Or wants to retry certain steps. Most of those pieces are resumable and a handful of them are callable as single items in isolation to redo portions of the flow.

Yeah, great point 👍 It should be documentet somehow, to understand the UX and work flows. But maybe it's already documentet inside the data-liberation-agent, I haven't read all *.md files 😅

@youknowriad @borkweb I have the next proposition/plan:

1. This PR is literally low-hanging fruit, since it's already done, tested, and I have clear picture of the next steps to polish it in the next week. Additionally, the approach here is not invasive, there are changes just to a few files and they are straightforward. So we can start with this approach and merge this PR as the first step.
2. Next week I will polish it, also review [Matthew's PR](https://github.com/Automattic/studio/pull/4011) and we will aim to archive data-liberation-agent github repo.
3. Two previous points will be done in the context of "Studio Code: Power Tools" linear project (studio-code-power-tools-45e9dd76a8ba). And later, we can reconsider proposed by Matthew's "plugins" aproach, if necessary - implement and it will be relativelly easy to migrate data-liberation-agent to connect it as a "plugins". Meanwhile, we can continue the discussion of plugins in parallel.

My goal is not to block "Studio Code: Power Tools" linear project (studio-code-power-tools-45e9dd76a8ba) and move forward with `/liberate` command delivery.

WDYT?

youknowriad (approved):
Personally, I'm not convinced by the plugin approach, I think a first-party import approach would result in a better product. But this shouldn't stop neither this PR from landing nor explorations of these approaches.

sejas (approved):
Vova, Thanks for adding the Liberation skill to Studio code.  I tested it with `Theme Replication` and it worked quite good. I noticed that the skill is not available in the Desktop app. I'll create an issue to add it to the composer.

Image

PR #3945, opened 2026-06-24
Make the `wp studio get-site-icon` command runtime-agnostic

fredrikekelund:
## Related issues

- Fixes STU-1886

## How AI was used in this PR

N/A

## Proposed Changes

The new agentic UI fetches the WordPress site icon (if there is one) by calling the custom `studio get-site-icon` WP-CLI command. This command is supposed to return a relative path to the site icon, but it implicitly assumes a Playground PHP runtime by always looking for `/wordpress` at the start of the site icon path. This breaks the logic for native PHP sites, where the site path is not virtualized.

This PR fixes the problem by stripping `ABSPATH` from the start of the site icon path, rather than `/wordpress`.

## Testing Instructions

1. Have a site with a site icon that uses the native PHP runtime
2. Stop Studio
3. Delete `wp-content/mu-plugins/99-studio-loader.php` for that site
4. Run `ENABLE_AGENTIC_UI=true npm start`
5. Ensure that the site icon shows up in the sidebar

## Pre-merge Checklist

### DIFF
--- a/tools/common/lib/mu-plugins.ts
+++ b/tools/common/lib/mu-plugins.ts
@@ -497,10 +497,14 @@ function getStandardMuPlugins( options: MuPluginOptions ): MuPlugin[] {
 				return;
 			}
 
-			// get_attached_file() returns paths rooted at /wordpress
-			// (Studio's PHP-runtime mount point). Strip the leading
-			// /wordpress so the host can resolve against the site dir.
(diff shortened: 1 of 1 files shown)


epeicher (approved):
Thanks @fredrikekelund! I have tested it, and I can see the site icon now. LGTM!

|Before | After|
|-------|------|
| Image| Image|

PR #3924, opened 2026-06-22
Add Studio Code slide to What's New modal

sejas:
## Related issues

- Related to STU-1896

## How AI was used in this PR

AI assistance was used to scaffold the new slide by following the existing `WhatsNewPage` pattern (imports, slide entry, localized link). All copy, the docs URL, and placement were reviewed and adjusted by hand.

## Proposed Changes

Studio Code — the agentic WordPress expert previously available only via the CLI/terminal — is now part of the Studio desktop app, replacing v1 of the AI assistant. This adds a "What's New" slide so users discover it on launch.

- New slide leads the What's New modal, with a Studio Code illustration and a "Learn more" link to the Studio Code docs.

## Testing Instructions

- Run `npm start`.
- Open the modal via **Help → What's New** (or clear `lastSeenVersion` to hit the first-launch path).
- Confirm the Studio Code slide appears **first**, the illustration renders without cropping, and **Learn more** opens the [Studio Code docs](https://developer.wordpress.com/docs/developer-tools/studio/studio-code/) page.
- Confirm the text reads correctly
- Check appearance in **both light and dark** mode.

The image is temporal. I'll replace it once we have a final one in STU-1896

| Light | Dark |
|--------|--------|
|  Screenshot 2026-06-24 at 06 58 42 | Screenshot 2026-06-24 at 06 58 54 |

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #3915, opened 2026-06-22
Add native PHP runtime page to What's New modal

bcotrim:
## Related issues

- Related to [RSM-4414](https://linear.app/a8c/issue/RSM-4414/update-the-whats-new-modal-with-details-about-the-native-php-runtime)

## How AI was used in this PR

Used Claude Code (Opus 4.8) to draft the slide copy, generate the placeholder illustration, and implement the change. I reviewed the wording for accuracy against the shipped **PHP runtime: Native / Sandbox** site settings (`site-runtime-copy.tsx`) and ran ESLint + the `whats-new`/`app-version-api`/`get-localized-link` unit tests locally.

## Proposed Changes

Adds a **native PHP runtime** slide as the lead item in Studio's What's New modal, so users discover that native PHP is now the **default runtime** for the best performance, and that they can switch any site between the Native and Sandbox runtimes in its settings. Includes a **Learn more** link to the PHP runtimes docs.

To keep the carousel to four slides, the last one (Blueprints) was retired. The order is now: native PHP → dark mode → phpMyAdmin → CLI. The modal is also set to surface once for existing users on this release.

> ⚠️ Visual change: needs human review in light + dark mode.
>
> 🖼️ The illustration is a **temporary placeholder** — final artwork (design in progress) will replace it later. The import path stays the same, so it's a one-file swap.

## Testing Instructions

1. Run Studio (`npm start`).
2. Trigger the What's New modal from the app menu (What's New).
3. Confirm the first slide is **"Run your sites with native PHP"**, with the placeholder illustration, the "What's new" pill, and a working **Learn more** link (opens the PHP runtimes docs).
4. Page through with Next / Previous / Done — confirm the dark mode, phpMyAdmin, and CLI slides follow and navigation works.
5. Verify in **both light and dark** appearance.
6. Optional: with no stored "last seen version", confirm the modal auto-shows; onboarding should still suppress it.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors?

PR #1128, opened 2025-03-25
Studio: Add new modal design

katinthehatsite:
## Related issues

STU-226

## Proposed Changes

This PR adds a modal to Studio according to Figma: RToz6tIuQ7nlZrikBte4GU-fi-7169_205567

Screenshot 2025-03-26 at 9 57 33 AM

## Testing Instructions

* Pull the changes from this branch
* Start the app with `STUDIO_WHATS_NEW_SECTION=true npm start`
* Click on the `Help > What's New` menu button in the topbar
* Observe that the modal appears
* Click through the modal and confirm that you can see intro page + three other pages with different features

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #3946, opened 2026-06-24
Fix truncated native PHP copy in What's New modal

bcotrim:
## Related issues

- Follow-up to #3915 — fixes the truncation reported in [review feedback](https://github.com/Automattic/studio/pull/3915#issuecomment-4781997343)
- Related to [RSM-4414](https://linear.app/a8c/issue/RSM-4414/update-the-whats-new-modal-with-details-about-the-native-php-runtime)

## How AI was used in this PR

Used Claude Code (Opus 4.8) to diagnose the truncation, trace the history of the modal's line-clamp behavior, and implement the fix. I confirmed the modal renders every slide's copy in full (no truncation, no scrollbar) in the running app, and ran ESLint + a typecheck on the changed file.

## Proposed Changes

The **"Faster local sites with native PHP"** slide in the What's New modal was cutting its description off mid-sentence (*"…switch between Native…"*).

The root cause was a fragile rule, not the copy itself: the modal clamped the **first** slide's description to 5 lines but **every other** slide to 3. The native PHP copy was written while it was the lead slide; once a Studio Code slide became the new lead (#3924), native PHP dropped into a 3-line slot and its (unchanged) text no longer fit.

Rather than trimming the copy, this gives **every slide the same 5-line clamp** the intro slide already used — which the fixed-height modal accommodates without scrollbars. That removes the intro-vs-regular distinction that caused the regression, so future copy tweaks or slide reordering won't silently truncate. It also drops a leftover wrapper element that had no styling. The slide copy is unchanged from `trunk`.

Thanks to @fredrikekelund for [suggesting](https://github.com/Automattic/studio/pull/3946#pullrequestreview-4561383809) the all-slides approach over a copy-only fix.

## Testing Instructions

1. Run Studio (`npm start`).
2. Open the What's New modal (app menu → **What's New**).
3. Page through every slide and confirm each description shows in full with no truncation/ellipsis — in particular, the **"Faster local sites with native PHP"** slide should end with *"…in your site settings."*
4. Confirm no scrollbar appears on any slide and the footer buttons stay in place.
5. Verify in **both light and dark** appearance.

## Pre-merge Checklist


fredrikekelund (approved):
The difference in line clamping between the first slide and the subsequent ones seems to have been introduced to prevent scrollbars from appearing. See https://github.com/Automattic/studio/pull/1128#issuecomment-2761194225

I can't really trigger a scrollbar even if I change the line clamping to 5 lines. I could see us changing that setting to 5 lines for all slides. This seems like an issue that could easily trip us up in the future, too.

I'll leave that call to you, @bcotrim. This simplistic solution works, too

bcotrim:
@fredrikekelund I also couldn't replicate the scrollbar, so I removed the conditional number of lines.
Let's keep an eye for any regressions.

I reverted the change back to the original text.

PR #3947, opened 2026-06-24
Add CLI integration test for site create (custom name, domain, HTTPS)

gavande1:
## Related issues

- Related to STU-1867

## Proposed Changes

The end-to-end suite drives every site action through the Studio UI, which is slow, flaky, and ties coverage to the UI. This adds the first real CLI integration test for `studio site create`: it runs the actual command against an isolated config directory and asserts on the persisted result, covering a site with a custom name and a site with a custom domain plus HTTPS. The tests are tagged `e2e` so they run in the slower release or manual suite rather than on every pull request.

## Testing Instructions

1. Build the CLI with `npm run cli:build`.
2. Run `npm test -- --tagsFilter='e2e'` and confirm both tests pass.
3. Run `npm test -- --tagsFilter='!e2e'` and confirm these tests are skipped, so the fast suite is unaffected.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/site/tests/create.e2e.test.ts
+++ b/apps/cli/commands/site/tests/create.e2e.test.ts
@@ -1,0 +1,117 @@
+/**
+ * @vitest-environment node
+ *
+ * Real end-to-end test for `studio site create`. Unlike create.test.ts (which
+ * mocks the command's dependencies), this spawns the built CLI binary and
+ * creates an actual site, verifying the real persisted state on disk.
+ *
+ * Requires the CLI to be built first (`npm run cli:build`); the suite skips
+ * itself otherwise. Tagged `e2e` so it runs in the slower (release/manual)
+ * suite rather than on every PR — run with `npm test -- --tagsFilter='e2e'`.
+ */
+import fs from 'fs';
+import path from 'path';
+import { afterEach, describe, expect, it } from 'vitest';
+import {
+	cleanupCliEnv,
+	cliE2ePrerequisitesMet,
+	readCliConfig,
+	runCli,
+	setupCliEnv,
+	type CliEnv,
+} from './helpers/cli-e2e';
+
+describe.skipIf( ! cliE2ePrerequisitesMet() )( 'CLI e2e: studio site create', () => {
+	let env: CliEnv | undefined;
+
+	afterEach( () => {
+		if ( env ) {
+			cleanupCliEnv( env );
+			env = undefined;
+		}
+	} );
+
+	it( 'creates a site with a custom name', { tags: [ 'e2e' ], timeout: 120_000 }, async () => {
+		env = setupCliEnv();
+		const siteName = 'Custom E2E Site';
+		const sitePath = path.join( env.sitesDir, 'custom-e2e-site' );
+
+		const result = await runCli(
+			[
+				'site',
+				'create',
+				'--name',
+				siteName,
+				'--path',
+				sitePath,
+				'--wp',
+				'latest',
+				'--no-start',
+				'--skip-browser',
+				'--skip-log-details',
+			],
+			env
+		);
+
+		expect( result.code, result.stderr ).toBe( 0 );
+
+		// The site is persisted to the real cli.json with the custom name.
+		const config = readCliConfig( env );
+		expect( config.sites ).toHaveLength( 1 );
+		const [ site ] = config.sites;
+		expect( site.name ).toBe( siteName );
+		expect( site.path ).toBe( sitePath );
+		expect( site.phpVersion ).toBeTruthy();
+		expect( site.running ).toBe( false );
+
+		// Real WordPress core files were copied into the site directory.
+		// (wp-config.php is generated at server start, which --no-start skips.)
+		expect( fs.existsSync( path.join( sitePath, 'wp-load.php' ) ) ).toBe( true );
+		expect( fs.existsSync( path.join( sitePath, 'wp-includes', 'version.php' ) ) ).toBe( true );
+	} );
+
+	it(
+		'creates a site with a custom domain and HTTPS',
+		{ tags: [ 'e2e' ], timeout: 120_000 },
+		async () => {
+			env = setupCliEnv();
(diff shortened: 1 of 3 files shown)


apps/cli/commands/site/tests/helpers/cli-e2e.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * Whether the prerequisites for spawning the CLI are present: the built binary
+ * and the bundled WordPress files. Used to skip the suite with a clear signal
+ * when the CLI hasn't been built (run `npm run cli:build` first).
+ */
+export function cliE2ePrerequisitesMet(): boolean {
+	return fs.existsSync( CLI_MAIN ) && fs.existsSync( BUNDLED_LATEST_WP );
+}
```
gcsecsey, line 44: Just noting here that these prerequisites make these tests currently auto-skipped on CI, because we're not building the CLI in the Unit Test job. We can tackle how to run these in CI in a separate issue.

gcsecsey (approved):
Nice work @gavande1, spawning the CLI and asserting against files on disk is a good direction for these tests IMO.

The two tests cases added here LGTM, and we can add more incrementally as needed.

apps/cli/commands/site/tests/helpers/cli-e2e.ts
gavande1: Good point. I opened STU-1937 to run these in CI, either by building the CLI in the Unit Tests job or folding them into the e2e suite.

apps/cli/commands/site/tests/create.e2e.test.ts
```diff
@@ -0,0 +1,117 @@
...
+		const [ site ] = config.sites;
+		expect( site.name ).toBe( siteName );
+		expect( site.path ).toBe( sitePath );
+		expect( site.phpVersion ).toBeTruthy();
+		expect( site.running ).toBe( false );
+
+		// Real WordPress core files were copied into the site directory.
+		// (wp-config.php is generated at server start, which --no-start skips.)
+		expect( fs.existsSync( path.join( sitePath, 'wp-load.php' ) ) ).toBe( true );
+		expect( fs.existsSync( path.join( sitePath, 'wp-includes', 'version.php' ) ) ).toBe( true );
```
fredrikekelund, line 70: I'd add an HTTP request that verifies that we get a 200 OK HTTP response from the site. Either we need to parse the regular stdout output from the `site create` command, or we add a `--format=json` parameter (that's the kind of change I was talking about in our huddle where working on the test suite might uncover flaws in the CLI implementation)

apps/cli/commands/site/tests/create.e2e.test.ts
```diff
@@ -0,0 +1,117 @@
...
+import {
+	cleanupCliEnv,
+	cliE2ePrerequisitesMet,
+	readCliConfig,
+	runCli,
+	setupCliEnv,
+	type CliEnv,
+} from './helpers/cli-e2e';
+
+describe.skipIf( ! cliE2ePrerequisitesMet() )( 'CLI e2e: studio site create', () => {
```
fredrikekelund, line 24: It seems more helpful if the test would break rather than gracefully skip if the CLI artifacts don't exist.

apps/cli/commands/site/tests/create.e2e.test.ts
```diff
@@ -0,0 +1,117 @@
...
+			// (--no-start skips the hosts-file / certificate setup that running would do.)
+			const config = readCliConfig( env );
+			expect( config.sites ).toHaveLength( 1 );
+			const [ site ] = config.sites;
+			expect( site.name ).toBe( siteName );
+			expect( site.customDomain ).toBe( customDomain );
+			expect( site.enableHttps ).toBe( true );
+			expect( site.running ).toBe( false );
+
+			expect( fs.existsSync( path.join( sitePath, 'wp-load.php' ) ) ).toBe( true );
```
fredrikekelund, line 114: Same thing here about adding an HTTP request for verification

apps/cli/commands/site/tests/helpers/cli-e2e.ts
```diff
@@ -0,0 +1,110 @@
...
+	// CLI skips its 24h WordPress-version update: keeps the run offline and
+	// deterministic, and avoids writing through the server-files symlink.
+	const cliConfigPath = path.join( configDir, 'cli.json' );
+	fs.writeFileSync(
+		cliConfigPath,
+		JSON.stringify( {
+			version: 1,
+			sites: [],
+			snapshots: [],
+			lastDependencyCheckTime: Date.now(),
```
fredrikekelund, line 70: Nice 👍

fredrikekelund (commented):
I left some comments about doing HTTP requests to the site's homepage after running the `site create` commands. The fundamentals here look really good, though 👍

apps/cli/commands/site/tests/create.e2e.test.ts
gavande1: Done in #4000. The create test now starts the site and asserts a 200 with HTML from the homepage.

apps/cli/commands/site/tests/create.e2e.test.ts
gavande1: I kept the custom-domain and HTTPS case as persistence-only. Serving it needs the full stack: the proxy on ports 80 and 443, a trusted root CA, and a write to the shared /etc/hosts behind an elevated-privileges prompt, so it cannot run in an isolated, headless test. The live check stays on the plain localhost site in #4000.

gavande1:
Thank you for the review @fredrikekelund. I have addressed your comments in #4000.

apps/cli/commands/site/tests/create.e2e.test.ts
gavande1: We already have open [issue](https://linear.app/a8c/issue/STU-1937/run-cli-e2e-tests-in-ci) to handle CLI tests.

PR #3948, opened 2026-06-24
Ensure that we display an appropriate error message for when there is .maintenance file

katinthehatsite:
## Related issues

Fixes STU-1847

## How AI was used in this PR

It was used to implement the solution.

## Proposed Changes

This PR ensures that when there is a .maintenance file and a fresh lock, the appropriate message is displayed to the user if they try to stop and start the site. It detects the lock that is fresher than 10 minutes and after 10 minutes, WordPress will ignore the stale lock file. 

## Testing Instructions

* Pull the changes from this branch and start Studio with `npm start`
* Create a site (or use an existing one) and start it
* Test fresh lock — create a .maintenance file with a recent timestamp by running in your Terminal the following (make sure to replace <your-site-name> with the name of your site):

```
  echo '<?php $upgrading = '$(date +%s)'; ?>' > ~/Studio/<your-site-name>/.maintenance
```
* Stop the site in Studio, then try to start it. You should see a message saying WordPress is in maintenance
   mode and the lock will expire within 10 minutes:

Screenshot 2026-06-25 at 12 03 28 PM

* Test no lock — delete the file:
  `rm ~/Studio/<your-site-name>/.maintenance`
* Start should work normally.

## Pre-merge Checklist


katinthehatsite:
This is the behaviour compared to trunk:

Screenshot 2026-06-25 at 2 16 25 PM

bcotrim (changes requested):
I suggest moving the `checkMaintenanceFile` to the CLI so it covers the app, cli and agent workflows. 
The app can still catch the error to display the correct Maintenance file error message.
What do you think?

katinthehatsite:
Thanks for the review @bcotrim - I applied your suggestion!

apps/studio/src/hooks/use-site-details.tsx
```diff
@@ -504,6 +504,24 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
...
+							'WordPress is currently performing an update. The maintenance lock should expire automatically within 10 minutes. Please wait for the update to finish and try starting the site again.'
+						),
+						showOpenLogs: false,
+					} );
+				} else {
+					const errorToShow = simplifyErrorForDisplay( error );
+					getIpcApi().showErrorMessageBox( {
+						title: sprintf( __( "Failed to start '%s'" ), siteName ),
+						message: __(
+							"Please verify your site's local path directory contains the standard WordPress installation files and try again. If this problem persists, please contact support."
```
gavande1: It might be silly question, Is there an option to contact support for Studio?

apps/studio/src/hooks/use-site-details.tsx
```diff
@@ -504,6 +504,24 @@ export function SiteDetailsProvider( { children }: SiteDetailsProviderProps ) {
 							showOpenLogs: true,
 						} );
 					}
+				} else if ( error instanceof Error && error.message.includes( 'maintenance mode' ) ) {
```
gavande1: The renderer matches `error.message.includes( 'maintenance mode' )`, but the CLI throws it through [__()](https://github.com/Automattic/studio/blob/5395ead4e7869707f89e60afcd28b52e832e23ef/apps/cli/commands/site/start.ts#L84), so would this still match in a non-English locale?

apps/studio/src/hooks/use-site-details.tsx
katinthehatsite: I think so, we have it in a couple of places: 

Screenshot 2026-07-07 at 2 37 55 PM

We can make the link to go there directly and open that page.

apps/studio/src/hooks/use-site-details.tsx
katinthehatsite: I made some adjustments for this and moved it to a constant similar to other errors in https://github.com/Automattic/studio/pull/3948/commits/1edcf9dd6834173162aed9812d5168949b348f7e

apps/studio/src/hooks/use-site-details.tsx
katinthehatsite: Actually, I cleaned that up and adjusted the message. I think the link won't work in the native Electron dialog as it does not allow rich text. I think the message should be fine as is

gavande1 (approved):
Thanks for addressing the concerns. I tested this, and it looks good to me. ✅

katinthehatsite:
Thanks for the review @gavande1  🙇

PR #1774, opened 2025-09-12
Extend the plan and include other property types returned by the wpcom sites endpoint

sejas:
## Related issues

- Fixes STU-673 and this [Sentry error](https://a8c.sentry.io/issues/6315200823/?project=4506612776501248).

## Proposed Changes

- Make the wpcom sites type more flexible to reduce the number of errors on Sentry, and make those sites visible to connect to them.

## Testing Instructions

- Run `npm start`
- Go to the sync tab
- Click on connect a site
- Observe that the modal opens and all your sites are still there

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #1669, opened 2025-08-12
Add offline site creation entry to release notes

sejas:
## Related issues

- STU-550
-  https://github.com/Automattic/studio/pull/1594

## Proposed Changes

- Add an release notes entry mentioning offline site creation, and upgrade Playground dependencies.

## Testing Instructions

- Confirm that the new entry is accurate and clearly explains what has changed for users.

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #3952, opened 2026-06-25
Route studio create --from URLs through Data Liberation

chubes4:
## Proposed Changes

`studio create --from <url>` captures a portable source artifact with the vendored Data Liberation CLI, then imports it through Static Site Importer. A capture with declared failures now stops before creating a WordPress site and points to its retained diagnostics. Missing or malformed capture receipts also stop the import.

The Data Liberation source/runtime is pinned to integration `eac207469d01c75d3eb27557563005b4a6734923`, based on refreshed upstream `main` at `678c74bf6ee4bbd01d6c6b5bd3a71feac0aab858`. It combines [DLA #255](https://github.com/Automattic/data-liberation-agent/pull/255), which preserves localized background media in captured stylesheets, and [DLA #254](https://github.com/Automattic/data-liberation-agent/pull/254), which simplifies the evidence-slug regex. This includes [DLA #251](https://github.com/Automattic/data-liberation-agent/pull/251): automatic disclosure expansion skips navigation links and popup controls, preventing capture from leaving the requested page and losing its cleanup state. The upstream npm cache is excluded, preserving the existing source/runtime-only vendor policy.

The incremental CLI build regression now invokes its fixed npm command through a shell so Windows can execute npm.cmd. Existing structured SSI quality rejection messages and strict quality gates remain in place.

The branch remains a draft integration candidate. These fixes do not establish complete source-to-WordPress visual parity.

## Contributor Attribution

Aagam Shah (@aagam-shah) authored the original Studio sitemap fix (`459cf87fa`) and is the author of the merged upstream DLA #207 commit (`eb0e348fea63156a800b2d2acdb350921d50ac48`).

Aagam also authored the transform-URL safety fix and regression in Studio #4784 (`2b9000c9a9687b551fc9dd95fc7caaad72a0bec7`) and the comma-safe srcset fix and regression in Studio #4785 (`cdd4723ad00ad275d4abc4e9358ac0e7b4763320`). Those fixes were ported into DLA #170 and #171. Their original commits and Claude co-author trailers remain in this branch history.

## Current Verification

- Studio client, create-command and incremental-build tests: **3 files, 96 tests passed**. Full typecheck and CLI rebuild passed after vendor synchronization. Vendored source retains exact upstream formatting.
- All **311 upstream source/runtime files** match the recorded DLA integration archive, preserving symlinks and excluding the upstream npm cache. All **487 built DLA dist files** match Studio's packaged runtime.
- The fresh SMPL URL import completed in **114.554 seconds**: 13/13 captures, zero capture failures, cleanup complete, 13 imported pages, 999 compiled blocks, zero reported fallback/invalid blocks. Paired SSI: `ad26f768c5608f956fa906ee99ede60e39ddf535`; Blocks Engine: `98c7345494da98a6922d9ac4e3ad1fd544fe28b1`.
- The paired package includes scoped upstream fixes [SSI #1668](https://github.com/Automattic/static-site-importer/pull/1668), [#1669](https://github.com/Automattic/static-site-importer/pull/1669), [#1670](https://github.com/Automattic/static-site-importer/pull/1670), and [Blocks Engine #1773](https://github.com/Automattic/blocks-engine/pull/1773), [#1774](https://github.com/Automattic/blocks-engine/pull/1774), [#1775](https://github.com/Automattic/blocks-engine/pull/1775). The formerly intermittent worker failure was reproduced as a concurrent directory-creation race and has a regression-tested fix in SSI #1670.
- Headless Playwright verified restored Home background images, Home visibility across six widths, correct navigation padding, matching service-card resting/hover states, and correct Unicode browser titles. Three editors load with no invalid/unregistered blocks. A Home paragraph edit/save/restore round trip passed.
- Full parity is **not** reached. Home's editor layout remains distorted; SVG icons and other layout/form/carousel differences remain. Ten routes already return source-side PHP errors. DLA compare checks all 13 routes offline with zero findings, but its four-route source sample has six failed checks involving external requests, captured menu interaction and Projects geometry. Import quality metrics are not a visual-parity claim.
- On Studio head `2e4f6d3c8da4a3572972a40a318d221d6eff5ff4`, Data Liberation, Lint, Unit and CodeQL pass; required CLI E2E is still running on Buildkite #22072. The previous head's CLI E2E passed in 35m44s. No PR has been merged.

## AI Assistance

The earlier branch integrations used GPT-6 Astra and GPT-5.6 Terra via OpenCode. This repair iteration used Codex to coordinate diagnosis, targeted fixes, review, local verification, and CI monitoring. Aagam Shah retains the original contributor attribution above and merge/release authority.


chubes4:
CI follow-up pushed in `7d7f45d8c`:

- Unit lane now runs `npm run test -- --tagsFilter='!e2e'`, keeping tagged CLI integration suites in the dedicated serialized CLI E2E lane.
- Filtered Unit verification passed: 211 files, 2747 passed, 60 skipped.
- Focused `start-stop.e2e.test.ts` passed: 2/2.
- The full local CLI E2E lane remains machine-sensitive under concurrent Playground/PHP load; isolated lifecycle coverage passes and the existing dedicated lane remains serialized.

AI assistance: openai/gpt-5.6-sol via OpenCode diagnosed the CI lane overlap, drafted the one-line test-selection change, and ran verification. Chris Huber remains responsible for the change.

chubes4:
## 1,145-route scale validation

The latest branch head (`ca6191a59`) is pushed and the end-to-end Richlynn run completed successfully through this Studio CLI path.

- Source: `https://www.richlynngroup.com/`
- Routes: 1,145/1,145 completed
- Batches: 124/124 completed
- Final state: zero failed or pending batches
- WordPress routes: 1,148 unique page URIs, no duplicate paths
- Cache: 5,364 network requests avoided
- Recovery: resumed across compiler interruption, encoded Unicode paths, transient HTML classification, and canonical route collisions

This proves the Studio orchestration and dependent SSI batching path can handle the target scale. It does **not** mark the generated site as fidelity-complete: Richlynn still has 13,055 `core/html` blocks, 1,235 reported fallbacks, 396 retained external assets, and no exact visual/editor parity proof. The generated site remains evidence under the zero-fallback gate.

Dependency PRs:
- Automattic/static-site-importer#798
- Automattic/blocks-engine#776

Validation on the pushed Studio head: `npx vitest run apps/cli/commands/site/tests/create.test.ts apps/cli/lib/tests/run-wp-cli-command.test.ts` (66/66 passed).

**AI assistance:** openai/gpt-5.6-sol via OpenCode was used for scale-run diagnosis, implementation support, verification, and drafting this evidence update. Chris Huber remains responsible for the submitted changes.

borkweb:
The `/liberate` command already imports static sites into Studio via the `blocks-engine` 🕺 

You can do so via: `/liberate ~/path/to/static/site`

That being said, switching out liberate's tooling to a different set of direct calls would make sense if the `php-transformer` has caught up to the JS approach! I also like the direct CLI command that you have for it.

chubes4:
@borkweb I have not compared the outputs side by side, but I was hoping the approaches could coexist initially. This PR intentionally avoids disrupting existing behavior. 

Longer term, I think Data Liberation could own the browser-dependent stages: client-side rendering, runtime behavior, and the visual feedback loop, and use the Studio CLI to materialize the site. 

The shared contract would be a clean HTML-based website artifact, regardless of the original source. 

I think that refactoring can be explored separately from this PR. Redirecting `/liberate` is not necessary for this primitive to exist.

chubes4:
Local SSI development-package flow is now implemented in `793ec7c2b`.

- `site create` accepts `--static-site-importer-path` as an explicit alternative to the release URL.
- Studio copies the ZIP beside its temporary Blueprint and installs it through a `bundled` resource; no permanent hosting or loopback HTTP server is required.
- Cleanup runs after successful and failed Blueprint execution without touching the caller's ZIP.
- Focused verification: 81 tests, TypeScript, ESLint, and CLI build pass.
- Real native-PHP validation used an SSI package built from SSI `aa55d6cf9e08` and Blocks Engine `f9de01f5e8d4`; import completed successfully with eight authored routes and an active generated block theme.

Reproduction shape:

```sh
studio site create \
  --path /tmp/fig-import \
  --from /path/to/source.fig \
  --static-site-importer-path /path/to/static-site-importer-dev.zip \
  --no-start --skip-browser
```

This keeps the published release URL as the production default while allowing rapid cross-repository iteration against immutable source commits.

AI assistance: OpenAI GPT-5.6 Sol via OpenCode implemented and tested the local bundled-package path and ran the native Figma integration. Chris Huber remains responsible for the changes.

jordesign:
## `studio create --from <url>` fails by default: bundled SSI release predates the request-bundle resolver this PR's own request-building code depends on

`DEFAULT_STATIC_SITE_IMPORTER_PLUGIN_URL` (`apps/cli/commands/site/create.ts`) points at the published `v1.8.5` release zip. `resolveStaticSiteImporterSource()`/`buildStaticSiteImporterRequest()` in this same file build directory-sourced import requests as `{ type: 'files', ref: 'request-bundle:source' }` — an opaque reference that only resolves if the installed SSI plugin registers a `static_site_importer_resolve_source_reference` filter for it via `static_site_importer_cli_prepare_request_bundle()` in `includes/cli.php`.

That resolver mechanism was added in Automattic/static-site-importer#1433 / #1448 / #1459 — all merged **2026-08-31**. `v1.8.5` predates it. I confirmed directly: the plugin actually installed from the default URL has no `static_site_importer_cli_prepare_request_bundle` function at all.

## Reproduction

```
studio create --path ./imported-site --name "Imported Site" \
  --from https://www.arrowmartialarts.com/ --skip-browser
```

Liberation and WordPress-server startup succeed; the static-site import step fails immediately:

```
{"success":false,"error":{"code":"static_site_importer_source_reference_unresolved","message":"The opaque source reference was not resolved by a server-owned resolver."}}
```

This isn't specific to this site — it's guaranteed for **every** `--from <url>` import against the default plugin URL, for any user who doesn't manually pass `--static-site-importer-path` with a build newer than `v1.8.5`.

The code comment above `DEFAULT_STATIC_SITE_IMPORTER_PLUGIN_URL` already documents the intended workaround (`npm run build:dev-package -- --blocks-engine-path <path>`), but that path is currently blocked by Automattic/static-site-importer#1272 (open) — `homeboy` rejects the reconstructed snapshot for lacking a `.git` root. I worked around both by hand-building a plugin zip from SSI `main` (`composer install --no-dev` + zip) and passing it via `--static-site-importer-path`; that build has the resolver and imports proceed past this step.

## Suggested direction

Either (a) cut a new SSI release that includes the resolver and bump `DEFAULT_STATIC_SITE_IMPORTER_PLUGIN_URL` to it, or (b) have Studio build/vendor SSI from a pinned source ref the way it already vendors `data-liberation-agent`, rather than depending on an external release's cadence for a feature this PR's own request-building code assumes exists. Not proposing a specific fix here since it's a distribution/architecture call — flagging with full repro so it's visible before this comes out of draft.

## AI assistance

Claude Code (Sonnet 5) reproduced this while running the loop described in this PR's own workflow against a live external site, traced it to the resolver/version mismatch, and verified the fix path by hand-building a compatible local package. Jordan Gillman orchestrated the investigation and is responsible for this report.

chubes4:
Synced the DLA iPhone mobile capture identity repair into the existing draft, resolving the generated-bundle-only rebase conflict by rebuilding from the rebased vendor source. Verification: vendor targeted test and full vendor suite passed (87 files, 907 tests); Studio CLI build, lint, and typecheck passed. Full Studio test has 14 environment failures after the fresh dependency setup because the bundled SQLite integration was initially absent; postinstall then completed the runtime asset setup. AI assistance: OpenAI gpt-5.6-terra via OpenCode performed source-level vendoring, regeneration, rebase, and verification.

ellenbauer:
### Interrupted `create --from` imports were restarting from zero

Pushed two small commits from a Data Liberation capture-engine run. Context for the first one, since the reason lives in the importer rather than here.

**What we saw.** Two Wix capture runs kept losing whole import cycles to ordinary interruptions. On `karate-dojo-muellheim.de` a ~65 minute cycle lost a race with machine interruptions three times in a row, leaving 93 compiled pages unverified and four separate run directories of receipts on disk. On `kdclegal.com` the run only finished because an operator went around this command and called the importer directly with `wp static-site-importer import --request=… --import-id=…`, which picked the dead run back up and finished it in 24 minutes.

**Why.** This command already resumes correctly: when the site is registered and the staged `.studio-import/request.json` still matches, `create` reruns that request against the existing site instead of reprovisioning. The missing half was in the importer. In v1.9.5, which this command pins, `Static_Site_Importer_Direct_Artifact_Import::resume()` is only reachable when `source.import_id` is set, so a rerun without `--import-id` always fell through to `start()` — a fresh random run id, a fresh workspace, and every page recompiled. Automattic/static-site-importer#1524 is that exact symptom; #1528 fixed it in v1.9.6 by discovering a still-running workspace whose binding matches the request it was handed.

So the rerun this command already performs now continues the work instead of repeating it. Unmatched, expired or mismatched runs still restart deliberately, so it stays self-healing rather than becoming a new failure mode.

**The pin**, moved to the current release:

- asset `static-site-importer-html-site-import.zip`, 1597202 bytes
- sha256 `c049712bb018da1085f13c05ba56ff9b0f2a1f03d9148ede2956e0a9424d16f1`
- vendored blocks-engine php-transformer 0.13.0, was 0.9.3
- verified the published zip carries `find_resumable` in both the direct artifact import and the canonical import service, and that v1.9.5 carries it in neither

The transformer moves four minors with it, so that side deserves your eyes more than the resume behaviour does. Happy to pin to v1.9.6 or v1.9.7 instead if you would rather take the resume fix with a smaller runtime delta.

**Second commit** is unrelated housekeeping: `CreateCommandOptions` never declared the `reportFiles` that `buildCreateFromSourceBlueprint` has been returning, so typecheck was failing with three TS2339 errors and Lint was red on the branch head before this push. Lint is green again now.

Leaving this as a draft and not marking it ready.

chubes4:
## Closes the visual-parity loop: Studio now feeds SSI's own oracle (#1707)

Static Site Importer merged [#1707](https://github.com/Automattic/static-site-importer/pull/1707), a section-geometry oracle (`Static_Site_Importer_Visual_Parity_Oracle::evaluate()`) that compares captured-vs-imported section geometry and fails `quality_pass` on a real disagreement — but nothing ever populated `validation_artifacts.source_pages`/`imported_pages`, so it always ran `skipped`. This commit wires it up.

### What's wired, and where

- **Source side** (`source_pages`): read directly from Data Liberation's own `sections/*.json` capture (`apps/cli/lib/visual-parity.ts: loadCapturedSectionPages`) — no re-extraction. DLA's capture is the trusted, pixel-exact stage.
- **Imported side** (`imported_pages`): measured from the *running* WordPress site after `wp static-site-importer import` completes, one Playwright page per captured route, using a polled readiness predicate (stable DOM node/image-decode signature across 6 ticks, then a full scroll pass) — never `networkidle`, no swallowed timeouts. A route that can't be measured is dropped from *both* sides rather than reported as a pass; a route that errors (HTTP 4xx/5xx) is kept on the source side only, so the oracle reports a real `missing_imported_page`.
- **Evaluation**: `Static_Site_Importer_Visual_Parity_Oracle::evaluate()` — the real, already-merged class — is called via `wp eval-file` (`apps/cli/php/visual-parity-eval.php`) right after materialization, before the plugin's own post-import cleanup removes it. This is one extra, cheap WP-CLI call, not a second import: no re-materialization, no risk of duplicate content. A real disagreement fails the CLI command with the oracle's own diagnostic quoted; a measurement failure never reads as a pass.

### How Studio obtains the extractor (and what's duplicated, disclosed)

- SSI's oracle module (`tools/visual-parity-oracle.mjs`) ships inside the installed plugin, but its only extraction entry point (`extractImportedPage`) bundles its own Playwright navigation *and* hardcodes `waitUntil: 'networkidle'` with no lower-level exported primitive to reuse just the DOM query. That's incompatible with the polled-readiness requirement here (verified live: `networkidle` on this corpus doesn't reliably wait for below-the-fold lazy images to decode).
- Given that, `extractImportedSectionPage` in `visual-parity.ts` is a **disclosed, minimal port** of the oracle's `EXTRACT_SECTIONS` selector heuristic — same selectors, same section-boundary logic — with two fixes required to get correct results against *real* WordPress block output (confirmed empirically, not assumed):
  1. A `<main>` landmark that contains other qualifying `<section>` elements is excluded from `sections[]` (kept only in `landmarks[]`). Unfixed, `<main>` (blocks-engine's standard content wrapper) swallowed every nested `<section>` into one page-spanning blob, since it's visited before its children in document order.
  2. `<header>`/`[role="banner"]` is never counted as a section. DLA's own `sections[]` never includes it, so counting it created an off-by-one index shift that manufactured disagreements for content that actually matched.
- The captured side is read from a second, independent producer (DLA) against the *same* documented schema, which is exactly what the design note anticipates ("SSI evaluates section records when a runtime provides them") — not a copy of the oracle's code.
- One data-shape reconciliation, not extraction logic: DLA tracks inline SVG icons separately from raster images (`section.icons` vs `section.images`); blocks-engine's SVG→`` conversion is documented as healthy (EVIDENCE.md), so `loadCapturedSectionPages` folds icon counts into the image/landmark-media counts it hands to the oracle, so a healthy conversion doesn't read as an "extra image."

### Verification

**Negative control — the known regression is caught.** Package built from SSI `main` (`475c0af0`) + blocks-engine trunk (`7d5499343`, the commit with the `/work` filter-chip regression):
```
Failed to import static site: Static site import failed visual parity validation:
work#0 (section_height): Section height disagrees with the capture. Section top disagrees with the capture.
```
Full diagnostic context: `work` section 0 (`section.py-20.lg:py-30.px-6`, the filter-chip/project-grid section), `source: 2068`, `imported: 2316` — **a +248px delta**, the exact magnitude and section EVIDENCE.md documented for this regression. On a package built from the same SSI ref against a pre-regression blocks-engine commit (`afa7f593`, ancestor of `7d5499343`, before PRs #1983/#1986 introduced the regression), that same section reports **zero** top/height disagreement — a clean A/B on the exact target page.

**Positive control — mixed, reported honestly.** With the pre-regression blocks-engine build, `visual_parity_artifacts` is no longer `pending` (real `browser_render`/`visual_diff` refs populate), and several pages (`labour`, `process`, `safety`, `about`, `renovations`) come back clean except for three uniform, template-level deltas that repeat identically on *every* page: a header nav icon WordPress renders natively that the source lacks, and a footer social-icon count drop that is EVIDENCE.md's own **Finding 4, explicitly flagged there as out of scope for this wave**. `contact` and the home page (`index`) show additional, larger structural disagreements that line up with this PR's own stated gaps ("Home's editor layout remains distorted") and EVIDENCE.md's Finding 1 (form fallback). None of this is a false positive from the wiring — the oracle's strict, index-based, no-tolerance-for-count comparator is measuring real, already-documented gaps in the current pipeline's fidelity that are independent of the specific regression this task targets. I did not loosen tolerances or add omissions to force a clean run; `fail_on_quality`'s pre-existing gate reported `quality_pass: true, fallback_count: 0` unaffected in both runs, exactly as before this change.

**Wall-clock overhead**: measured overhead for 8 routes, one Playwright measurement pass on the imported site plus one `wp eval-file` call, is roughly **15–20s** on top of the ~85–90s baseline import (no second import pass — the two runs above measured 82s and 86s / 119s total including a same-page 30s readiness timeout that degraded gracefully instead of failing the run).

### AI assistance disclosure

This PR iteration (wiring `validation_artifacts` for #1707) was produced with **opencode** running **anthropic/claude-sonnet-5**, operating autonomously end-to-end: reading SSI's oracle/consumer PHP and JS source to determine the exact input contract, choosing the extraction/evaluation architecture, writing the TypeScript/PHP glue, building SSI dev packages against two blocks-engine refs, running real `studio site create --from` imports against the live source, diagnosing and fixing the two selector-heuristic gaps and the icon/image data-shape mismatch from real output (not assumption), and re-verifying against a live site before reporting results. All findings and numbers quoted above come from real command output during this session, not simulated.

chubes4:
**CI status note**: Lint, CodeQL, Data Liberation, CLI E2E Tests, and fastlane Helper Tests are green on `2bf57f4`. One of the two "Unit Tests" shards (mac/windows/linux — Buildkite posts all three under the same `Unit Tests` GitHub status context) is currently red: https://buildkite.com/automattic/studio/builds/22169#01a0b510-c5ac-43ef-a7b1-e64e5d21a2bc

I don't have Buildkite credentials in this sandbox, so I can't open that log directly. What I've verified instead:
- `npm run test -- --tagsFilter='!e2e'` (the exact command `run-unit-tests.sh` runs) passes 100% locally: 3579 passed, 44 skipped, 0 failed, across the full workspace including `create.test.ts`'s 85 tests.
- Every commit earlier in this PR's history shows both `Unit Tests` shard postings passing; this is the first red one, so I can't confidently write this off as pre-existing flakiness either.
- My new code (`buildVisualParityValidationArtifacts` / Playwright launch) only executes when a Data Liberation `sections/` capture directory is present, which no existing test fixture provides — so no existing unit test should reach that path — but I can't rule out a platform-specific issue (e.g. a network-dependent Playwright browser install attempt somewhere in that path) without seeing which platform/test actually failed.

Flagging this for a maintainer to check the actual Windows/Linux Buildkite log rather than asserting it's a flake. Happy to keep iterating once I know what failed.

chubes4:
## Section-extractor dedup: investigated, blocked, evidence below

**Goal**: make `static-site-importer`'s `tools/visual-parity-oracle.mjs` the single source of truth for section extraction, and have this file (`apps/cli/lib/visual-parity.ts`) import it instead of maintaining its own copy.

**Outcome**: exported the extractor on the SSI side ([Automattic/static-site-importer#1716](https://github.com/Automattic/static-site-importer/pull/1716)), but did **not** switch this file over to it, and did **not** delete the local copy. Direct reuse isn't safe without either changing what this file measures or exceeding the intended scope of a dedup. Comment-only commit here (`deaab48b1`) updates the two stale in-file comments to reflect what's below; no functional change.

### What I found

Static Site Importer merged **two** PRs affecting this oracle, not one:

- **#1707** (`f61e5a9b`) — what this file (`2bf57f495`) was built against: `EXTRACT_SECTIONS`, schema `visual-parity-oracle-input/v1`, `source_pages`/`imported_pages`, fields like `top`, `headingSizes`, `images[].displayWidth`.
- **#1710** (`06dfc6b5`) — already merged to SSI's `main`, one commit before its current tip — **fully replaced** the same `Static_Site_Importer_Visual_Parity_Oracle` class and the same `.mjs` file with an incompatible contract: `EXTRACT_LAYOUT`, schema `layout-baseline/v1`, `source_reports.layout_baseline`/`imported_render`, `offset.top`, `headings[].font_size` objects, a broader `media` query (img+svg+video, not img-only), `forms[].fields[].padding`. No backward-compat shim.

This means the oracle class this file's WP-CLI shim (`apps/cli/php/visual-parity-eval.php`) calls has **already drifted out from under this integration**, independent of anything in this dedup task — the exact class of silent-drift bug this whole visual-parity effort exists to catch, just a bigger instance of it than expected.

### Empirical proof (not just reading code)

Built a dev package from SSI `main` (current, unmodified), built this CLI (`2bf57f495`, unmodified), and ran a real `site create --from` against the J&T source with blocks-engine pinned to `7d5499343` (the known chip-regression commit):

- `/work` measured `docHeight: 2713` — confirms the regression is present in this build (source capture is 2465).
- Top-level import: `quality_pass: true` — **not currently caught**.
- This file's own visual-parity WP-CLI call returned:
  ```json
  {
    "status": "not_verified",
    "reason": "Layout baseline contract is absent or malformed.",
    "missing_data_contract": [
      "source_reports.layout_baseline schema static-site-importer/layout-baseline/v1",
      "source_reports.layout_baseline.viewports",
      "source_reports.layout_baseline.pages",
      "source_reports.layout_baseline.intentional_omissions"
    ]
  }
  ```
  i.e. it **never evaluates real geometry today, regardless of my changes** — the schema mismatch degrades it to `not_verified` on every run.

Re-ran the identical import against a dev package built from SSI's `feat/export-visual-parity-extractor` (the export-only PR above): **byte-identical result**, same `not_verified` reason, same measured section/landmark geometry (diffed both `visual-parity-input.json` files, normalizing only site-specific asset URLs/theme slugs — geometry is byte-identical). The PHP oracle class itself is byte-identical between the two SSI packages; only the exported `.mjs` differs, and this file doesn't consume that path anyway.

### Why a drop-in swap isn't safe

Even setting the schema mismatch aside, `EXTRACT_LAYOUT` (SSI's current, canonical extractor) is not behavior-equivalent to `extractImportedSectionPage` in this file:

- This file explicitly excludes `<header>` as its own section, and excludes a wrapping `<main>` that already contains other qualifying sections — two fixes added here for confirmed gaps against real WordPress block output. `EXTRACT_LAYOUT` has neither; swapping to it would change section counts/boundaries for the exact output this measures, which is a real extraction-behavior change, not a refactor.
- Field shapes differ throughout (see above) — the oracle expects the newer shape SSI now emits; this file's `VisualParityArtifacts` type, `loadCapturedSectionPages`'s icon-folding logic, and the `visual_parity` envelope sent to the WP-CLI shim would all need to change to consume it.
- Separately: this file's own comment already documents that the plugin's files aren't available as a build-time npm dependency; for PHP-WASM/Playground-runtime sites (one of the two runtimes `site create --from` supports) the plugin also lives inside a virtual filesystem this Node process can't read directly, so even a schema-compatible extractor couldn't be reliably `import()`-ed at runtime across both supported runtimes without further work.

Doing all of that (schema migration + runtime-availability handling) to land a "simple" extractor swap is a materially bigger, riskier change than "delete the duplicated extractor," and out of the stated scope of keeping the capture/imported orchestration, the WP-CLI shim, and the `validation_artifacts` plumbing untouched. Forcing it now risks a half-migration that trades one drift bug for a different, harder-to-spot one.

### Recommendation

Track a follow-up to migrate this file's `visual_parity` contract to `static-site-importer/layout-baseline/v1` end-to-end (source-side too — Data Liberation's own `sections/*.json` capture is still on the *original* `schema: 9` shape from #1707, so it would also need to change or be adapted). Once that's done, the extractor swap becomes safe and mechanical.

### AI disclosure

Investigated and authored with AI assistance — model `anthropic/claude-sonnet-5`, tool `opencode`. Used for cross-repo git archaeology (`git log`/`git show` across both repos to find #1707 vs #1710), building/running the dev packages and real `site create --from` imports quoted above, diffing the resulting artifacts, and drafting this comment and the comment-only commit. All claims above were verified by actually running the code (real builds, real Playwright-driven imports, real diffs), not inferred from reading source alone.

chubes4:
## Static import quality-failure reporting

When SSI's quality gate fails after a site has already been created, the CLI still throws and still keeps the site. It now also tells the operator what exists and how to continue.

**Before**
```
Failed to import static site: Static site import failed quality validation: SSI reported 10 core_html_block failures. Review the importer diagnostics and retry.
```

**After** (failure signal unchanged; retained site is now described)
```
The site is running and can be inspected.
Path: /path/to/site
URL: http://localhost:8881
Re-run the same command to resume the import.

Failed to import static site: Static site import failed quality validation: SSI reported 10 core_html_block failures. Review the importer diagnostics and retry.
```

Credentials are not printed (`logSiteDetails` remains unused on this path). The site is not rolled back.

**Exit code:** Empirically, `runCommand` only throws and does not set `process.exitCode`. The yargs handler catches that `LoggerError` and `reportError` sets `process.exitCode = 1`. Kept non-zero: the quality gate failed. The objection was a message implying nothing was created, not the failure signal.

Verified with `npx eslint --fix` on the touched files, `npm run typecheck`, and `npx vitest run apps/cli` (1061 passed).

AI assistance: xai/grok-4.6 via OpenCode.

chubes4:
## Capture sidecars now reach the importer (`948924726`)

`--from <url>` still stages **`website/`** as the importer source (`liberateWebsite()` return value). SSI prefixes that tree as `website/index.html`, `website/contact/index.html`, `assets/website/...`. That contract is unchanged.

The gap was that Data Liberation writes compiler-evidence JSON **one directory above** `website/`. Studio already looked up there for `sections/` and for `capture-receipt.json` in the client, but `collectArtifactRootReports` only ran when the caller passed the capture root. The URL flow never does, so `interaction-states.json` (and siblings) never entered `.studio-import/source` or `metadata.reports`. SSI #1735 and blocks-engine #2031 therefore saw an empty artifact file set.

**Approach:** keep `website/` as `sourcePath`. When the source *is* the website directory, look at the parent capture root (same parent lookup `resolveDataLiberationSectionsDir` already does) and copy a whitelist of JSON sidecars into the staged source, declaring them in `metadata.reports` so SSI does not prefix them under `website/`. Passing the whole capture root would have ballooned the payload (`screenshots/`, `media/`, `resources/`, `layout-geometry/`) and reshaped source paths.

**Whitelist** (present files only; absence is not an error):

`capture-receipt.json`, `asset-evidence.json`, `breakpoints.json`, `cleanup-evidence.json`, `computed-styles.json`, `css-variables.json`, `diagnostics.json`, `interaction-states.json`, `layout-geometry-proof.json`, `layout-geometry-report.json`, `palette.json`, `scroll-states.json`, `source-profile.json`, `typography.json`

Excluded on purpose: `screenshots/`, `media/`, `resources/`, `layout-geometry/` (large), and `sections/` (already forwarded separately).

### Evidence

- `npm run typecheck` — pass
- `npm run lint` — pass (two pre-existing unused-var warnings, not in this change)
- `npm test -- apps/cli` — 1065 passed, 38 skipped
- Focused coverage: sidecars forwarded when the source is `website/`; sourcePath remains the website tree; no-sidecar / plain HTML still import; excluded directories are not in `reportFiles`; copy lands in `.studio-import/source/<name>`

Could **not** run a full `site create --from https://cerberusmed.base44.app/` import: no SSI dev package was available in this environment. Payload presence is proven from the constructed blueprint / staged copy, not from an end-to-end SSI run.

PR 3952 remains a **draft**. Not merged.

**AI assistance:** model `xai/grok-4.6`, tool `opencode`. Used to trace `staticSiteImport.sourcePath` into the staged importer payload, implement the parent-root sidecar forward, add focused tests, and run local gates. Chris Huber remains responsible for the change.

chubes4:
## Visual parity gate was inert: Studio wrote a payload shape the current SSI oracle rejects

`32bfb44a4` updates Studio to emit the contract the currently-merged oracle actually consumes, and fails closed when a real payload still comes back `not_verified`.

### Contract drift

Studio's check measured the imported site correctly, then threw the result away. Every real-import defect shipped with a silent pass, including an HTTP 404 page, because `Static_Site_Importer_Visual_Parity_Oracle::layout_baseline_from()` never found a baseline and `evaluate()` degraded to `not_verified`. Studio treated anything other than `failed` as success.

I ran SSI's current oracle against two payloads describing the same condition (two captured pages, one of which 404s on the imported site):

| | Studio wrote (pre-fix) | Oracle expects |
|---|---|---|
| envelope | `visual_parity.source_pages` / `.imported_pages` | `source_reports.layout_baseline` + `imported_render` |
| schema | `static-site-importer/visual-parity-oracle-input/v1` | `static-site-importer/layout-baseline/v1` |
| stage | `import_vs_capture` | `import_vs_baseline` |
| page shape | keyed map of `{ page, sourceUrl, sections[], landmarks[] }` | `pages[]` array of `{ id, viewport, page_height, sections[], landmarks[] }` |
| required top-level | — | `viewports[]`, `pages[]`, `intentional_omissions[]` |

**A)** Studio PR 3952 payload as shipped → `status=not_verified`, `disagreements=0`, reason: *Layout baseline contract is absent or malformed.*

**B)** Current oracle contract, page missing → `status=failed`, `disagreements=1` (`missing_imported_page`).

The oracle logic is correct. Only the contract had drifted (SSI refactored after `#1707`; Studio was never updated).

### What changed

- `apps/cli/lib/visual-parity.ts` still measures with the existing Playwright extractor (header exclusion + wrapping-`<main>` unwrap unchanged). It now serializes those measurements as `source_reports.layout_baseline` + `imported_render` in `static-site-importer/layout-baseline/v1`, with `viewports[]`, `pages[]`, `intentional_omissions[]`, and per-field names the oracle compares (`top` / `offset.top`, `height`, `page_height`, `headings[].font_size`, `media[].display_width` / `display_height`, forms).
- `runVisualParityCheck` now uses `visualParityGateFailure()`: oracle `failed` still fails the import; oracle `not_verified` on a **populated** Studio payload also fails, naming `missing_data_contract` paths. Genuine could-not-measure cases (`artifacts.status === 'not_verified'`, wp-cli/oracle unavailable) stay non-fatal.
- `apps/cli/php/visual-parity-eval.php` is unchanged in behavior (still `evaluate()` of the JSON envelope).

### Tests

`apps/cli/lib/tests/visual-parity.test.ts` asserts:

- the emitted envelope satisfies the oracle's required contract paths (`schema`, `viewports`, `pages`, `intentional_omissions`, per-page `id` / `viewport` / `sections`)
- a captured page missing from the import is kept on the baseline so the oracle can report `missing_imported_page`
- section field names match what the oracle reads (`media` + `display_width` / `display_height`, not `images` + `displayWidth`)
- the **pre-migration** `source_pages` / `imported_pages` envelope fails those contract paths (this is the test that would have caught the drift)
- `visualParityGateFailure` fails `ready` + `not_verified` with the missing contract paths, and does not fail a genuine could-not-measure case

CLI unit tests, lint, and typecheck passed locally (`406` tests in the CLI lib + create suites; `npm run typecheck` green).

### End-to-end / oracle evidence

Direct evaluation of the current SSI oracle class:

| payload | oracle `status` | notes |
|---|---|---|
| A) pre-fix Studio envelope | `not_verified` | `missing_data_contract` lists `source_reports.layout_baseline` schema / `viewports` / `pages` / `intentional_omissions` |
| B) new envelope, `/social-kit` absent from `imported_render` | `failed` | disagreement `missing_imported_page` |
| C) new envelope, matching pages | `passed` | no disagreements |

Live browser measurement of the District 21 capture's `sections/` against a running WordPress site where `/` is 200 and `/social-kit` is HTTP 404 produced `artifacts.status=ready` with baseline pages `[index, social-kit]` and imported pages `[index]`. Feeding that payload to the same oracle returned **`status=failed`**, including:

```
social-kit missing_imported_page Imported render did not include this baseline page.
```

Full `site create --from` of `https://beneficial-district-twenty-one.base44.app/social-kit` and of the root URL, using the specified SSI zip, both failed **before** the visual-parity gate with SSI materialization `WordPress site plan contains unresolved local browser reference "/"`. The gate never ran on those commands because SSI `quality_pass` already failed. That is independent of this serialization fix; the oracle evidence above is what this change actually unblocks.

### AI assistance

Model `xai/grok-4.6`, tool `opencode`. Used to read the SSI oracle + Studio visual-parity path, reshape the payload, add contract tests, run lint/typecheck/unit tests, and attempt the two `site create --from` acceptance imports.

PR remains a draft. Do not merge.

chubes4:
## CI on `32bfb44a4`

Green: Lint, CLI E2E Tests, Data Liberation, fastlane Helper Tests, CodeQL.

Unit Tests: GitHub shows one failed Buildkite job (`22180#01a0b6b1-ab81`). Buildkite logs are not readable from this sandbox (403). Local `npm test -- --tagsFilter='!e2e'` on this commit: 390 files, 3601 passed, 0 failed. This matches the earlier shard-red pattern on this PR (mac/windows/linux share the `Unit Tests` context).

PR remains a draft.

ellenbauer:
**Finding from a Wix test run: the final visual check fails every free wixsite.com import, even when the site was built correctly**

Source: https://ellenbauer37.wixsite.com/my-site-5 (my own Wix test site, 3 pages)

What a customer sees: the import runs for several minutes, WordPress ends up with all 3 pages looking right, and then the command ends with "Failed to import static site". The log shows why:

```
Visual parity: http://localhost:8900/my-site-5 responded with HTTP 404; treating "my-site-5" as missing on the imported site.
```

What broke: `routeForCapturedPage()` in `apps/cli/lib/visual-parity.ts` probes the imported site at the path of the source URL. For a source that lives under a subpath, that path is the Wix folder name (`/my-site-5`), not a page. The capture exports the homepage to `website/index.html`, so WordPress serves it at `/`. Every free Wix address has this shape (`https://<user>.wixsite.com/<site>`), so every one of them would report a missing homepage.

Fix pushed to this branch as one commit, 7fc0dc6a: the check now reads the capture receipt next to `sections/` and probes the route each captured URL actually occupies in the export (`website/index.html` gives `/`, `website/privacy-policy/index.html` gives `/privacy-policy/`). When the receipt has no matching entry, the old behavior is unchanged. Two regression tests added in `apps/cli/lib/tests/visual-parity.test.ts`. I also checked it against the real capture: the three pages now resolve to `/`, `/privacy-policy/` and `/accessibility-statement/`, all of which answer 200 on the imported site.

Run details: Studio 805a9b6c, Data Liberation Agent 4bf6019 (vendored copy synced locally from b6a6483), Static Site Importer f0210e84, Blocks Engine 12c93f35.

Side note: the imported site is still named after the Studio site ("wix-run-my-site-5") instead of "My Site 5". That is already tracked in #4720.

This PR stays a draft; nothing here asks for a merge.

chubes4:
## Question on scope: does visual parity belong in this branch at all?

Raising this here rather than as a Studio issue, since the code only exists on this branch.

`apps/cli/lib/visual-parity.ts` (839 lines, added 2026-09-18 in `2bf57f495`) is a browser-driven geometry measurement layer: it launches a page, waits for readiness, walks section geometry, and compares captured against imported. `trunk` has no equivalent. It was introduced by this migration loop.

Two implementations of that concern already existed:

- **Data Liberation Agent**, `src/lib/fidelity/` — a deliberate two-tier gate. A cheap offline self-consistency tier, plus a browser-backed source-fidelity tier that its own AGENTS.md describes as checking "text, geometry and reflow at widths the capture never sampled, plus dialogs." It has `check.ts`, `checks.ts`, `rendered-contract-checks.ts`, `score.ts`, `evidence.ts`, each with tests.
- **Static Site Importer**, `tools/visual-parity-oracle.mjs` — which the Studio file's own header comment says it matches the schema and viewport of.

Today's run produced concrete evidence that the duplication is already costing us. The Studio gate excluded the home page with `never settled to a stable, fully-decoded render`. Root cause: its readiness predicate requires `completeImageCount === imageCount` before it scrolls, but 238 of the page's 278 images are `loading="lazy"` and are never requested until scrolled. `completeImageCount` sits at 37/275 forever. Only the two short legal pages were measurable.

DLA already solved exactly this. `triggerLazyLoad` exists in its screenshot helpers and is applied before measurement. The newest implementation reimplemented readiness without the lesson the oldest one had already learned — which is the predictable failure mode of having three of these.

The effect on this loop is not academic: visual parity is currently **unmeasurable** on the page that matters most, for any source with a long, image-heavy landing page. That describes most real Wix marketing sites.

So before fixing the deadlock, the question worth answering is whether Studio should be measuring parity at all, or whether it should invoke DLA's `compare` and SSI's oracle and report their verdicts. Fixing it in place would entrench a third copy.

Studio #4907 was opened against the wrong repository for this and has been closed in favour of this comment.

## AI assistance

Investigated with OpenCode using Anthropic Claude Sonnet 5 during an end-to-end import verification run; earlier phases of the same session used OpenAI GPT-5.6 Luna and GPT-6 Astra via OpenCode. Chris Huber raised the ownership question and identified the misfiling.

chubes4:
**Moved: Studio's visual-parity duplication → orchestrating `data-liberation compare`**

Generated with OpenCode using Anthropic Claude Sonnet 5, finalized outside Homeboy with operator authorization (Homeboy's control plane was unavailable).

This branch had grown `apps/cli/lib/visual-parity.ts` (839 lines): a full second implementation of readiness-waiting, section-geometry extraction, and captured-vs-imported comparison, feeding its own bundled PHP script that called into `static-site-importer`'s oracle class. `data-liberation-agent` already owns that exact concern end to end and exposes it as `data-liberation compare <dir>` (readiness, geometry, and the comparison against the live source — see its `src/lib/fidelity/`).

**Removed** (now redundant with DLA, or dependent on what was removed):
- `apps/cli/lib/visual-parity.ts` and its test — readiness polling, section/landmark extraction, `layout-baseline/v1` serialization, and the captured-source-URL → imported-permalink mapping. All of this existed only to feed a Studio-side re-measurement of the *imported WordPress render*, which DLA's `compare` does not do (it compares its own exported capture against the live source, not an arbitrary rendered destination) — DLA deliberately keeps destination knowledge out of its capture/compare path.
- `apps/cli/php/visual-parity-eval.php` — the bundled script that called `Static_Site_Importer_Visual_Parity_Oracle` via `wp eval-file`. No longer invoked.
- `getBundledVisualParityEvalScriptPath()` in `dependency-management/paths.ts`.

**Added**: `compareLiberatedCapture()` in `apps/cli/lib/data-liberation-client.ts`, extending the existing subprocess pattern (`liberateWebsite()` already spawns the vendored DLA CLI) to also run `data-liberation compare <dir>`. `create.ts` now calls this right after SSI materializes the site — same timing as before — and relays DLA's own pass/fail verdict and report text; a disagreement still fails the import.

**Kept as Studio's job** (per the operator's framing of "when Studio invokes the gate is Studio's business even if how it measures is not"):
- Resolving which on-disk directory is the DLA capture for a given import source (`resolveDataLiberationCaptureDirectory`, replacing the old `sections/`-specific resolver — DLA's own `compare` resolves either the capture root or its `website/` directory itself, so Studio just hands back whichever one it already has).
- Running the gate on **every** `create` — fresh or resumed. The old gate read `site.url` directly, which is only ever set in memory on a fresh create and never persisted, so it silently skipped on every resumed import (the bug `aff1e33fa` partially patched). The replacement gates on the capture directory alone — resolved fresh from the source path on both paths — and needs no site URL at all, since DLA compares the capture to its live source rather than to the running WordPress site. Verified with a regression test that fails if that guard is reverted to depend on the site URL again.
- Surfacing the verdict and deciding import pass/fail.

**Scope narrowing, flagged honestly**: the old check's real subject was the *materialized WordPress render* (post-SSI, at the live permalink); DLA's `compare` verifies the *DLA capture itself* against its live source (pre-import). These are different guarantees — narrower than before. I judged this as the operator's explicit intent ("Studio should orchestrate DLA's compare and report its verdict, not perform its own measurement") rather than an oversight, but it's worth a second look: if imported-WordPress-render fidelity still needs checking, that capability doesn't exist in DLA today and isn't something I ported ahead of the concurrent DLA work.

**Verified**:
- `npx eslint --fix`, `npm run typecheck`, and the affected `apps/cli` unit tests (`create.test.ts`, `data-liberation-client.test.ts`) all pass; full `apps/cli` suite (1083 tests) passes.
- Confirmed the resume regression test is meaningful: reverting the guard to depend on the site URL again fails exactly that one new test.
- Ran Studio's real (non-mocked) `compareLiberatedCapture()` against a retained real capture (Wix source) and its live source over the network: produced a genuine pass/fail verdict for all three captured routes, **including the home page** — the old code's readiness predicate deadlocked on lazy-loaded images and silently excluded the home page from the gate entirely; this replacement has no such predicate to deadlock.
- A full `studio site create --from <capture>` end-to-end run against that same real capture was blocked before reaching this gate by an unrelated, pre-existing `static-site-importer` materialization failure (an unresolved same-page anchor, `#Section 10`) — not caused by, or fixable within, this change.
- `packages/data-liberation-agent` stays a verbatim upstream copy (confirmed via `scripts/check-data-liberation-agent-upstream.mjs`); nothing in it was patched.

chubes4:
**CI status on `b795083`**

- Lint, Data Liberation, fastlane Helper Tests, CLI E2E Tests, and one of the three "Unit Tests" matrix legs: **passing**.
- "Unit Tests" shows one failing leg (~6m39s) and the Windows leg is stuck pending. Both pre-date this change: the identical failing leg, down to the same 6m39s duration, already fails on `aff1e33fa` and `fc5ca4911` (the two commits immediately before mine on this branch), and the Windows leg is stuck pending on those same commits too. Confirmed via `gh api repos/Automattic/studio/commits/<sha>/statuses` on all three commits.
- The separate CodeQL failure is also pre-existing: it already fails on `fc5ca4911` with the same 4 alerts, all inside vendored `packages/data-liberation-agent/` files this change does not touch (and which must stay a verbatim upstream copy).
- Locally, `npm ci` followed by the exact CI command (`npm run test -- --tagsFilter='!e2e'`) passes: 390 test files / 3613 tests, 0 failures.

Not something this change introduced or can fix from here — flagging per the verification standard rather than merging past it silently.

gmovr:
https://github.com/Automattic/studio/issues/4919 Raising this here :)

chubes4:
Fixed the parallel-install race reported in #4919 (thanks @gmovr) in eff964fec, along with two related defects found while reproducing it.

**Races in `apps/cli/lib/import-runtime.ts`**
- **Engine install (the reported `ENOTEMPTY`).** `loadCaptureEngine` removed the target, then renamed its staging directory into place with no coordination, so a second process could fail or delete an engine another import was using. The new `publishStagedInstall()` renames atomically. If another process published first, its complete install is kept and this process's staging directory is discarded. A complete install is never removed; an incomplete directory at the target is moved aside, then replaced. Staging names are now unique per process.
- **Asset download.** Parallel downloads of the same Static Site Importer ZIP shared one `<destination>.partial` path, so one process renamed it away under the other (`ENOENT … .partial`). This was the most common failure in practice. The partial file is now unique per download.
- **Playwright link.** A concurrent `symlink` now tolerates `EEXIST`.

**GitHub rate limit.** Each `studio create --from` makes two unauthenticated `releases/latest` calls, and the cache was only in memory for one process. The unauthenticated limit is 60 per hour per IP, so a shared network or a burst of imports hit `HTTP 403` and failed the import outright, even with the runtime already installed. Resolved releases are now recorded on disk:
- Processes share a lookup for the existing 5-minute window.
- If GitHub is unreachable or rate-limited, the last release this machine verified is used, with a notice on stderr. Nothing unverified is ever used.

**Verification**
- Real reproduction with no mocks: six processes resolving and installing both runtimes in parallel into an empty runtime directory, repeated five rounds.
  - Before: 3 of 30 succeeded, failing with `ENOENT … static-site-importer-html-site-import.zip.partial`.
  - After: 30 of 30 succeeded, with no staging, partial or stale directories left behind.
- Rate-limit fallback, live: with an expired lookup record and GitHub returning `HTTP 403` unauthenticated, the import resolved DLA 0.5.3 and SSI 1.15.3 from the record and continued.
- New unit tests: parallel same-asset download, a second publisher keeping the first install (same inode, loser cleaned up), incomplete-target replacement, cross-process lookup sharing, and fallback on both 403 and network failure. The download and publish tests fail against the previous code.
- `npm run typecheck` passes. `npm run test -- --tagsFilter='!e2e'` passes: 3,768 tests.

AI assistance: Claude (Anthropic) via OpenCode reproduced the race, wrote the fix and tests, and ran the verification above. Directed by @chubes4.

chubes4:
Simplification and reliability pass (e758ba54a, da24e12db), with `trunk` merged in.

**Removed**
- **The in-import fidelity check** (`compareLiberatedCapture`, `reportDataLiberationFidelity`, and the `captureDirectory` plumbing through three call sites). It re-ran Data Liberation's live-source check on every URL import, only reported, and duplicated `data-liberation compare`. Studio still prints the more useful `compare --candidate` command, which compares the new site with the original.
- **The `onCompareCommand` / `onPartialCapture` callbacks.** `liberateWebsite` now returns `{ websiteDir, compareCommand, partialCapture }`.
- Net: −373 lines, +55.

**Fixed (found by a real end-to-end run)**
- The runtime cache's `node_modules/playwright` link pointed into a Studio checkout that no longer existed, so every URL import failed with `Cannot find package 'playwright'`. Any Studio update or moved install hits this. The link is now re-pointed when it does not target the running Studio's Playwright, and swapped in atomically. The regression test fails without the fix.

**Kept, deliberately**
- Studio's partial-capture policy reads the capture receipt's `route_capture_failed` entries. DLA's returned `failures` count per viewport and include 404s, so the receipt is the precise signal and this is not duplication.
- The SSI quality-failure message remains as a defensive check against SSI's contract.

**Verification**
- `npm run typecheck` and lint pass. `npm run test -- --tagsFilter='!e2e'` passes: 3,774 tests.
- Real `studio create --from https://tommerritt.com/ --keep-source` using the built CLI, DLA 0.5.3 and SSI 1.15.5 resolved from releases: exit 0 in **60 s**, compared with 250–270 s for the earlier import flow. Full-page screenshots against the live site are pixel-identical at 1600 and 390 px, with equal page heights.
- The printed `compare --candidate` command ran end to end (6 routes in 4 min). Its remaining findings are WordPress core additions (the hidden "Skip to content" link, the `s.w.org` emoji script) and duplicate font-face registrations, not visible differences.

AI assistance: Claude (Anthropic) via OpenCode did the simplification, found and fixed the Playwright link bug, and ran the verification above. Directed by @chubes4.

PR #3963, opened 2026-06-25
Add copy button and selectable text to agent code blocks

youknowriad:
## Related issues

- Related to: none — surfaced while testing the Studio Code agent; no tracking issue.

## How AI was used in this PR

Authored end-to-end with Claude Code. It diagnosed two separate root causes (the Electron renderer denies `navigator.clipboard`, so the agentic-UI copy silently failed; and code blocks needed an explicit `user-select` since the desks canvas/app shell opt out of selection) and implemented the fix across both markdown renderers plus the connector. Reviewers should sanity-check the new `StudioConnector.copyText` method and confirm the visual result in light + dark (see note below).

## Proposed Changes

When the Studio Code agent renders a code snippet, you previously couldn't reliably select or copy it. This makes code blocks first-class to copy from, in **both** surfaces that render agent output — the agentic UI conversation and the Studio Code tab:

- Code blocks are now explicitly **selectable** (text selection no longer suppressed by the surrounding canvas/app-shell).
- Each fenced code block gets an always-visible **Copy** button (top-right) that flips to “Copied” for a moment after use.
- Copying is routed through the **host clipboard** via a new `StudioConnector.copyText` (desktop → main-process clipboard; web → `navigator.clipboard`). This is why the agentic-UI button now actually copies and updates — the renderer's own `navigator.clipboard` is denied in the Electron desktop.

## Testing Instructions

1. Open a Studio Code agent session and ask for a code snippet (e.g. “give me a small CSS snippet”).
2. In the **Studio Code tab** and in the **agentic UI** conversation:
   - Select text inside the code block with the mouse — it should highlight.
   - Click the **Copy** button — it should switch to **Copied**, and pasting elsewhere should yield the code.
3. Verify both **light and dark** appearance of the button.

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? (typecheck, ESLint, and unit tests pass; visual/dark-mode pass still needs a human.)

### DIFF
--- a/apps/studio/src/components/studio-code-session/markdown/index.test.tsx
+++ b/apps/studio/src/components/studio-code-session/markdown/index.test.tsx
@@ -22,9 +22,16 @@ describe( 'Studio Code Markdown', () => {
 		fireEvent.click( button );
 
 		expect( copyText ).toHaveBeenCalledWith( 'body { color: red; }' );
-		await waitFor( () =>
-			expect( screen.getByRole( 'button', { name: 'Copied' } ) ).toBeInTheDocument()
-		);
+		await waitFor( () => expect( screen.getByRole( 'status' ) ).toHaveTextContent( 'Copied' ) );
+		expect( screen.getByRole( 'button', { name: 'Copy code' } ) ).toBeInTheDocument();
+	} );
+
+	it( 'shows a tooltip for the copy button', async () => {
+		render( <Markdown>{ '```css\nbody { color: red; }\n```' }</Markdown> );
+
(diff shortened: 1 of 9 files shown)


shaunandrews:
Love it. Made some tweaks to it. removed the label. Made it a more neutral button. I also added a background so that if text ran underneath it, it didn't obscure.

image

So perhaps we should add some sort of spacing to the top, or reposition the button so it doesn't obscure content. Or maybe we only show it when hovering.

youknowriad:
It's funny you're always on dark mode, I'm always on light, so I don't see the issues that you see.

shaunandrews:
I switch constantly throughout the day!

PR #3972, opened 2026-06-26
Consolidate Unix CLI installation managers into a shared base

bcotrim:
## Related issues

- N/A

## How AI was used in this PR

Claude Code traced the three platform installation managers and the shared dispatch in `ipc-handlers.ts`, then hoisted the duplicated Unix logic into a base class. All changes were reviewed manually; type-checking and the existing CLI unit test were run locally.

## Proposed Changes

The macOS and Linux CLI installation managers were ~85% identical — shell-profile editing, symlink management, the standalone-install detection helpers, and the install/uninstall error-dialog handling were duplicated nearly verbatim, and all three platforms repeated the same auto-install try/catch shape.

This consolidates that shared logic into a single Unix base class that macOS and Linux extend, overriding only what is genuinely platform-specific (shell-profile filenames, default profile, the production CLI path, and macOS's legacy `/usr/local/bin` symlink cleanup). Windows keeps its registry/path mechanics but now reuses the shared error/info dialog and auto-install helpers.

Pure internal refactor: **no behavior change** — same dialogs, same shell-profile edits, same symlink handling, same Sentry/error reporting. Net ~240 fewer lines, making future changes to install behavior a single edit instead of three. No user-visible impact.

## Testing Instructions

This is a behavior-preserving refactor, so existing CLI install/uninstall flows should work exactly as before:

- macOS/Linux: install the CLI from the app, confirm `studio` resolves and `~/.local/bin` is added to the shell profile; uninstall and confirm the symlink is removed (and, on macOS, the legacy `/usr/local/bin/studio` is cleaned up).
- Windows: install/uninstall and confirm the PATH entry + proxy bat file behave as before.
- Standalone (curl) installs must still be detected and left untouched on all platforms.

## Pre-merge Checklist


gavande1:
Thanks for working on this, the refactor makes sense.

One thought: since macOS and Linux differ only by config and one post-uninstall step, we could drop the two subclasses and pass everything through the config on a single class. It removes the `abstract`/`extends` layer while keeping the same behavior. For example:

```ts
new UnixCliInstallationManager( {
	platform: 'darwin',
	shellProfiles: { bash: '.bash_profile', zsh: '.zshrc' },
	defaultProfile: '.zshrc',
	prodCliPackagedPath,
	onUninstalled: uninstallLegacyCliIfNeeded, // was the subclass override
} );
```

Not a blocker, more of a preference for composition over inheritance here.

bcotrim:
Thanks for the review @gavande1 could you take another look please? 🙇

gavande1:
Thanks, this is exactly what I had in mind. The composition reads much cleaner here, and the behavior looks preserved across all three platforms. LGTM 👍

One optional nit: with the subclasses gone, `cli-installation-manager-base.ts` is not really a base anymore, so I think `unix-installation-manager.ts` might read better. Not a blocker.

gavande1 (approved):

PR #3973, opened 2026-06-26
Studio sync: handle 413 push failures gracefully

gavande1:
## Related issues

- Fixes STU-1843

## How AI was used in this PR

AI assistance was used to draft the retry-classification helper, the HTTP status extraction, and the user-facing error messages. I reviewed every change, restored the repository code formatting after an unintended reformat, and confirmed the new messages are translatable.

## Proposed Changes

When a site push produces an archive larger than the WordPress.com upload (host site) limit, the server responds with 413. Before this change, the client kept retrying that response through every retry delay and then reported a vague "Unknown error" or a connection failure, so users could not tell what went wrong or how to fix it.

Now the upload stops on the first non-retryable client error (403, 413, and similar) and shows a clear, translatable message: a "site archive is too large" message for 413, or the HTTP status for other client errors. Transient failures (5xx, network drops, and the resumable 409 and 423 statuses) still retry as before, so normal pushes are unaffected.

## Testing Instructions

Manual code review should be enough!

## Pre-merge Checklist


epeicher (approved):
Thanks for handling this gracefully @gavande1! I have reviewed the changes and they LGTM!

PR #3974, opened 2026-06-26
Add studio uninstall command for the standalone CLI

bcotrim:
## Related issues

- Fixes STU-1938

## How AI was used in this PR

Used Claude Code to confirm the standalone install layout against the live wpcom-hosted `install.sh`/`install.ps1` (paths aren't in this repo), scaffold the `uninstall` subcommand and tests, and diagnose the stale cross-channel update banner (live-probed the updates endpoint to confirm the server was correct and the cache was at fault). All code was reviewed by the author.

## Proposed Changes

The standalone curl/irm one-liner installs the Studio CLI but there was no matching way to remove it (raised during nightly testing). This adds a discoverable, offline, cross-platform `studio uninstall`:

- Removes the installed bundle and takes `studio` off PATH. On macOS/Linux it deletes its own runtime in place and removes the `~/.local/bin/studio` symlink (leaving the shell-profile PATH line, which the desktop app also relies on). On Windows — where the running `node.exe` can't delete itself — a detached helper finishes the removal after exit and strips the PATH registry entry the installer added.
- Stops running sites and the daemon first so nothing holds open handles on the runtime being removed.
- Preserves user data (`~/.studio` — sites, config) by default; `--purge`/`--all` removes it too, with a confirmation prompt.
- Only acts on standalone installs; npm and desktop-embedded installs get pointed at the right uninstall path instead.

It also fixes a related bug surfaced while testing: a production CLI could show an update banner offering a `-dev` nightly. The standalone update-check cache lives in the shared `~/.studio` config, so reinstalling a different channel onto the same directory reused the previous channel's cached result for up to 24h. The cache is now ignored when its channel differs from the running CLI's, so the server stays the source of truth.

## Testing Instructions

Build a standalone bundle and install it locally:

```bash
npm run cli:bundle -- darwin arm64   # or your platform/arch
curl -fsSL https://public-api.wordpress.com/wpcom/v2/studio-app/install.sh \
  | STUDIO_CLI_URL="$(pwd)/standalone-bundles" bash
```

Then, in a new shell:

- `studio uninstall` — confirm `bin/`, `cli/`, and the `~/.local/bin/studio` symlink are removed, `studio` is gone from PATH, and `~/.studio` config/sites are kept.
- Reinstall, then `studio uninstall --purge` — confirm it prompts before deleting `~/.studio`, and removes it on confirmation.
- Start a site, then `studio uninstall` — confirm sites/daemon stop before removal.
- Run `studio uninstall` from an npm install (`npm i -g wp-studio`) and a desktop-embedded build — confirm it does nothing destructive and prints the correct guidance.
- **Windows** (needs manual cross-platform verification): confirm the detached helper removes the bundle after the process exits and strips the `<install>\bin` entry from the user PATH registry.

Unit tests: `npm test -- apps/cli/commands/tests/uninstall.test.ts apps/cli/lib/tests/update-notifier.test.ts`

> Note: this PR also carries a small drive-by fix to the standalone update notifier (the cross-channel cache guard), which is STU-1772 scope — bundled here because it's a one-line correctness fix surfaced during this work.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/commands/uninstall.ts
+++ b/apps/cli/commands/uninstall.ts
@@ -1,0 +1,170 @@
+import { spawn } from 'node:child_process';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { confirm } from '@inquirer/prompts';
+import { getConfigDirectory } from '@studio/common/lib/well-known-paths';
+import { __, sprintf } from '@wordpress/i18n';
+import { Mode, runCommand as stopSites } from 'cli/commands/site/stop';
+import { getCliInstallKind } from 'cli/lib/update-notifier';
+import { StudioArgv } from 'cli/types';
+
+// The launcher execs `<installDir>/bin/node`, so the running binary's grandparent
+// is the standalone install root (handles a STUDIO_CLI_HOME override too — no need
+// to assume ~/.studio).
+function getInstallDir(): string {
+	return path.dirname( path.dirname( process.execPath ) );
+}
+
+// Remove the `~/.local/bin/studio` PATH symlink, but only if it still points at
+// the bundle we're removing. The shared `export PATH=…/.local/bin` rc line is left
+// alone — the desktop app relies on it too.
+function removePosixSymlink( installDir: string ): string | null {
+	const symlink = path.join( os.homedir(), '.local', 'bin', 'studio' );
+	try {
+		if ( ! fs.lstatSync( symlink ).isSymbolicLink() ) {
+			return null;
+		}
+		const target = path.resolve( path.dirname( symlink ), fs.readlinkSync( symlink ) );
+		if ( target !== path.join( installDir, 'bin', 'studio' ) ) {
+			return null;
+		}
+		fs.unlinkSync( symlink );
+		return symlink;
+	} catch {
+		return null;
+	}
+}
+
+// Windows locks bin\node.exe while the CLI runs, so we can't delete our own
+// runtime in-process. Hand the destructive work to a detached PowerShell helper
+// that waits for this PID to exit, then removes the bundle dirs and strips the
+// installer's PATH registry entry (mirroring install.ps1's raw-registry write).
+function scheduleWindowsCleanup( installDir: string ): void {
+	const q = ( p: string ) => p.replace( /'/g, "''" );
+	const binDir = path.join( installDir, 'bin' );
+	const script = `
+$ErrorActionPreference = 'SilentlyContinue'
+while (Get-Process -Id ${
+		process.pid
+	} -ErrorAction SilentlyContinue) { Start-Sleep -Milliseconds 200 }
+Remove-Item -LiteralPath '${ q( binDir ) }' -Recurse -Force
+Remove-Item -LiteralPath '${ q( path.join( installDir, 'cli' ) ) }' -Recurse -Force
+Remove-Item -LiteralPath '${ q( installDir ) }' -Force
+$key = [Microsoft.Win32.Registry]::CurrentUser.OpenSubKey('Environment', $true)
+if ($key) {
+	$raw = [string]$key.GetValue('Path', '', [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames)
+	$parts = $raw -split ';' | Where-Object { $_ -ne '' -and $_ -ne '${ q( binDir ) }' }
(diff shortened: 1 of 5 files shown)


apps/cli/commands/uninstall.ts
```diff
@@ -0,0 +1,170 @@
...
+		return;
+	}
+	if ( installKind !== 'standalone' ) {
+		console.log(
+			__(
+				'This Studio CLI is bundled with the Studio desktop app. Uninstall the app to remove it.'
+			)
+		);
+		return;
+	}
```
fredrikekelund, line 88:
```suggestion
	if ( installKind === 'npm' ) {
		console.log(
			__( 'This Studio CLI was installed via npm. Remove it with: npm rm -g wp-studio' )
		);
		process.exitCode = 1;
		return;
	}
	if ( installKind !== 'standalone' ) {
		console.log(
			__(
				'This Studio CLI is bundled with the Studio desktop app. Uninstall the app to remove it.'
			)
		);
		process.exitCode = 1;
		return;
	}
```
We should signal that the uninstallation did not complete successfully.

apps/cli/commands/uninstall.ts
```diff
@@ -0,0 +1,170 @@
...
+		const symlink = removePosixSymlink( installDir );
+		if ( symlink ) {
+			removed.push( symlink );
+		}
+	}
+
+	if ( removeConfig ) {
+		fs.rmSync( configDir, { recursive: true, force: true } );
+		removed.push( configDir );
+	}
```
fredrikekelund, line 135: Let's trash it instead.

apps/cli/commands/uninstall.ts
```diff
@@ -0,0 +1,170 @@
...
+	if ( ! removeConfig ) {
+		console.log(
+			sprintf(
+				/* translators: %s is the config directory path */
+				__( '\nYour sites and config were kept in %s. Re-run with --purge to remove them.' ),
+				configDir
+			)
+		);
+	}
```
fredrikekelund, line 149: You can only rerun `studio uninstall` by reinstalling the CLI. I think it makes more sense to tell them which directory to look at

apps/cli/commands/uninstall.ts
```diff
@@ -0,0 +1,170 @@
...
+
+export const registerCommand = ( yargs: StudioArgv ) => {
+	return yargs.command( {
+		command: 'uninstall',
+		describe: __( 'Uninstall the standalone Studio CLI' ),
+		builder: ( yargs ) => {
+			return yargs.option( 'purge', {
+				type: 'boolean',
+				alias: 'all',
+				describe: __( 'Also delete your Studio config and site list (~/.studio)' ),
```
fredrikekelund: "site list" could potentially imply site data and is kind of redundant in this sentence. @bcotrim, do you agree that just `Also delete your Studio config (~/.studio)` might actually be clearer?

apps/cli/lib/update-notifier.ts
```diff
@@ -194,7 +194,16 @@ async function notifyStandalone( currentVersion: string ): Promise< void > {
 
 	let latestVersion: string | null = null;
 
-	if ( cached && now - cached.lastChecked < UPDATE_CHECK_INTERVAL_MS ) {
+	// Ignore a cached check from a different channel than the running CLI. The cache is
+	// keyed only by config dir (~/.studio), which is shared across installs — reinstalling
+	// a different channel (e.g. nightly → production) onto it would otherwise surface a
+	// stale, cross-channel banner (a `-dev` build offered to a production install) until the
+	// TTL elapses. On a mismatch we refetch and let the server decide.
```
fredrikekelund: Not totally redundant, but distinctly AI-smelly

apps/studio/package.json
```diff
@@ -90,7 +90,7 @@
 		"@wordpress/element": "^7.0.0",
 		"@wordpress/icons": "13.2.0",
 		"@wordpress/private-apis": "^1.12.0",
-		"@wordpress/react-i18n": "^4.47.0",
+		"@wordpress/react-i18n": "4.47.0",
```
fredrikekelund, line 93: I guess this is the change from https://github.com/Automattic/studio/pull/3985?

fredrikekelund (approved):
I haven't tested on Windows yet, but this looks good to me and works as expected on macOS 👍

I left a few comments that'd be nice to address, but I'll let you address the ones you think are relevant without waiting for another review from me, @bcotrim

PR #3981, opened 2026-06-27
Invert dark mode tooltip colors

shaunandrews:
## Proposed Changes

- In dark mode, tooltips now use an inverted treatment: white background with black text, improving contrast against Studio's dark UI.
- The local `@wordpress/ui` patch exposes tooltip background and foreground CSS custom properties, so Studio can control tooltip colors without brittle generated selectors or wrapper-only workarounds.

image
image

## Testing Instructions

- Run `ENABLE_AGENTIC_UI=true npm start`.
- Switch Studio to dark mode, then hover icon-only controls that render tooltips.
- Confirm tooltips use a white background with black text in dark mode.
- Switch to light mode and confirm the existing light-mode tooltip colors are unchanged.


katinthehatsite (commented):
The changes look good, I have one note regarding the testing instructions:

Image

I had to first run:

` rm -rf node_modules && npm install`

to ensure that the patch was installed correctly. In case this helps other testers

katinthehatsite (approved):

katinthehatsite:
Looks like some tests are failing so I retriggered those and rebased trunk.

katinthehatsite:
Everything is passing, I will merge this now

PR #3982, opened 2026-06-28
Allow agent to control light/dark color theme, and display screenshots in the chat

shaunandrews:
## Proposed Changes

- Makes screenshots captured by Studio Code appear inline in the conversation as media artifacts, so users can see the captured image instead of only reading a text summary — in both the agentic UI and the legacy Studio Code session view.
- Adds light/dark color scheme arguments to screenshot and visual-inspection tools, making it easier for agents to verify theme-aware pages in the intended color scheme.
- Screenshots are emitted as structural chat artifacts only — the raw `mediaWidgetPayload=` text lines are gone from tool output (a shared stripper keeps old transcripts clean), and the `studio_present` screenshot rule was replaced so the agent no longer re-presents the same capture, which caused the duplicate image reported in review.
- Captures are stored in a per-session `<session>.screenshots/` directory next to the session file instead of the OS temp dir, so inline images keep working in older transcripts; deleting a session removes its screenshots.
- Remote-site sessions now emit chat artifacts for all Studio tools (previously only `take_screenshot` was wrapped, so e.g. `site_create` never emitted its site-preview artifact).
- Browser builds (hosted / `studio ui`) can't read local files, so local-only screenshot artifacts are hidden there via a new `readLocalMedia` connector capability instead of rendering guaranteed "Image unavailable" boxes.

## Screenshots

image
image
image

## Testing Instructions

- Run `npm test -- apps/cli/ai/tests packages/common/ai apps/ui/src/ui-classic/components/session-view/conversation apps/studio/src/components/studio-code-session/conversation`
- Run `npm run typecheck`
- In Studio Code, ask the agent to take a screenshot and confirm exactly **one** captured image appears inline in the conversation (the duplicate is gone).
- Confirm the capture is saved under `~/Library/Application Support/Studio/sessions/<date>/<session>.screenshots/` and still renders after restarting the app and reopening the session.
- Ask for light and dark captures with `colorScheme: "all"` and confirm both artifacts are produced, labeled `(desktop light)` / `(desktop dark)`.
- Open the same session in the legacy Studio Code view and confirm the screenshot renders inline there too (check light + dark mode).
- Delete the session and confirm its `.screenshots/` directory is removed.
- Open a pre-existing session whose transcript contains `mediaWidgetPayload=` lines and confirm they stay hidden in the expanded tool details.


apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
```diff
@@ -185,6 +191,75 @@ function resolveBatchedAnswerForQuestion(
...
+			reject( new Error( 'Unable to read local media file as a data URL.' ) );
+		} );
+		reader.addEventListener( 'error', () => {
+			reject( reader.error ?? new Error( 'Unable to read local media file as a data URL.' ) );
+		} );
+		reader.readAsDataURL( new Blob( [ file.data ], { type: file.mimeType } ) );
+	} );
+}
+
+function stripScreenshotMediaPayloadLines( text: string ): string {
```
bcotrim: `stripScreenshotMediaPayloadLines` is duplicated in a`pps/studio/src/components/studio-code-session/conversation/index.tsx:68`, and the `mediaWidgetPayload=`/`mediaWidgetPayloads=` prefixes are hardcoded here, there, and in the producer (`apps/cli/ai/tools/take-screenshot.ts:146`). If the CLI renames the prefix, both strippers silently break. Lift the prefixes + helper into tools/common? Non-blocking.

apps/ui/src/ui-classic/components/session-view/conversation/index.tsx
```diff
@@ -185,6 +191,75 @@ function resolveBatchedAnswerForQuestion(
 	return optionLabels.has( answer ) ? answer : undefined;
 }
 
+function isRecord( value: unknown ): value is Record< string, unknown > {
```
bcotrim: Dupes the unexported `isRecord` in `tools/common/ai/chat-artifacts.ts:38` — export that one and import it instead?

bcotrim (commented):
While testing I noticed the agent appears to be taking 2 screenshots, or printing the screenshot twice.

| Attempt 1 | Attempt2  |
|--------|--------|
| Image | Image |

shaunandrews:
> While testing I noticed the agent appears to be taking 2 screenshots, or printing the screenshot twice.

Its taking the screenshot in light and dark mode, but since your site doesn't change anything, it looks weird.

shaunandrews:
> Its taking the screenshot in light and dark mode, but since your site doesn't change anything, it looks weird.

Hmmmm, maybe not. Looks like it might be showing them twice for some reason. Looking at a fix now.

shaunandrews:
👋 Update on the review feedback — a batch of fixes just landed on this branch (rebased onto trunk first). Here's the tour:

**The double screenshot @bcotrim spotted** turned out not to be two captures at all — it was one capture presented twice. The tool now auto-emits its artifact, but the old `screenshot-local-media` prompt rule was *also* telling the model to re-present the same image via `studio_present`. Fixed at the source (4bbec3cb2): the `mediaWidgetPayload=` lines are gone from tool output entirely, and the rule now tells the model screenshots are already shown inline. One screenshot, one image.

**Which also resolves the two duplication comments:** with the producer no longer writing marker lines, the strippers only exist for old transcripts — there's now a single `stripMediaWidgetPayloadLines` (plus the marker constants) in `@studio/common/ai/chat-artifacts`, used by both renderers (cf66b7307). `isRecord` is exported from there too, and the third private copy in `studio-widgets.ts` now imports it as well.

**On Copilot's data-URL performance note:** the conversion now happens once inside a cache layer — react-query in the classic UI, a per-path map in the legacy view — so each file crosses IPC and gets encoded a single time per app lifetime. (Fun detour: we tried `blob:` object URLs first, but the desktop CSP only allows `data:` in `img-src`, and revoke-on-cleanup broke under StrictMode remounts. Data URLs it is — a6c3f8ffc.)

A few more things surfaced while testing that seemed worth fixing in the same pass:

- **Screenshots now live next to the session file** in a `<session>.screenshots/` folder instead of `os.tmpdir()` (686cb2510). Before, macOS would purge the temp files within a few days and every screenshot in an older transcript became a permanent "Image unavailable" box. Deleting a session cleans up its screenshots too.
- **The legacy Studio Code view renders screenshots inline now as well** (85b2b8109) — previously it hid the payload line but showed nothing in its place, so the capture was unreachable there.
- **Remote sessions were only wrapping `take_screenshot`** for artifact emission — `site_create` returns a site-preview artifact too, and unwrapped it never fired (so session placement never happened for remote-created sites). All the remote studio tools are wrapped uniformly now, and the wrapper forwards `signal`/`onUpdate` and keeps `details` intact (df716ffae).
- Guarded the artifact render path against malformed persisted entries (one bad JSONL line used to be able to take down the whole conversation view), hid local-file artifacts in the browser builds where the connector can't read disk, and a handful of small dedups (shared `colorScheme` schema, shared `emulateMedia` helper, test mock factory).

Typecheck is clean and the affected suites pass (~235 tests, including new coverage for the storage, delete sweep, and malformed-entry cases). Still worth a human eyeball on the inline images in light + dark mode.

---
*Written by Claude (working with @shaunandrews), who also did most of the implementation in this batch.*

bcotrim (approved):

PR #3642, opened 2026-05-29
Upgrade React to 19 and WordPress packages to React 19-compatible versions

wojtekn:
## Related issues

Closes #3590
Closes #3637

## How AI was used in this PR

This PR was created with Claude Code. All changes were reviewed and verified with passing typecheck and full test suite.

## Proposed Changes

This PR combines two Dependabot PRs (#3590 and #3637) that have a circular dependency — the new `@wordpress/components@34`, `@wordpress/element@7`, `@wordpress/compose@8`, and `@wordpress/dataviews@15` all require React 19 as a peer dependency, making them impossible to land independently.

**Package upgrades:**
- `react` / `react-dom`: 18.2.0 → 19.2.6
- `@types/react` / `@types/react-dom`: 18.x → 19.x
- `@wordpress/components`: 33.0.0 → 34.0.0
- `@wordpress/compose`: 7.36.0 → 8.0.0
- `@wordpress/element`: 6.39.0 → 7.0.0
- `@wordpress/dataviews`: 14.2.0 → 15.0.0
- `@wordpress/icons`: 13.1.0 → 13.2.0 (unpinned — the prior pin was a pre-React 19 workaround)
- `@wordpress/i18n`, `@wordpress/react-i18n`, and several `apps/ui` WP packages bumped
- `@rive-app/react-canvas`: 4.12.0 → 4.18.0 (adds React 19 peer dep support)
- `date-fns`: ^3.3.1 → ^4.1.0 in `tools/common` and `apps/studio` — required so `@wordpress/components@34` 

**Patch file:**
- Renamed `@wordpress+components+33.0.0.patch` → `@wordpress+components+34.0.0.patch`

**TypeScript fixes for React 19 breaking changes:**
- `useRef<T>()` → `useRef<T>(undefined)` (no-arg overload removed for non-DOM refs)
- `ForwardRefRenderFunction` second param updated to `React.ForwardedRef<T>`
- `RefObject<T | null>` propagated through prop types (React 19 `useRef(null)` is now nullable)
- Dropped `ref` from react-markdown `ExtraProps` spreads to avoid v18/v19 `LegacyRef` mismatch
- `allowpopups="true"` → `allowpopups={true}` (boolean attribute)
- `Tooltip.Popup side` prop migrated to `positioner={<Tooltip.Positioner side={...} />}` (@wordpress/ui 0.14 API change)
[...]

PR #3985, opened 2026-06-29
Pin @wordpress/react-i18n to 4.47.0 to fix trunk bundling install (ERESOLVE)

bcotrim:
## Related issues

- Related to STU-1938 (surfaced while working on it; this is a standalone trunk fix)

## How AI was used in this PR

Used Claude Code to diagnose the CI failure: probed the npm registry to find the offending `@wordpress/react-i18n` version, reproduced the `install:bundle` ERESOLVE locally via a dry-run, and confirmed via GitHub commit statuses that trunk itself is red. The diff was reviewed by the author.

## Proposed Changes

**Trunk is currently red** (E2E + Performance Metrics) and has been since `167141e6f` — every build fails in the `prePackage` hook with:

```
npm error ERESOLVE
Found: react@19.2.7
peer react@"^18.0.0" from @wordpress/react-i18n@4.49.0
```

Root cause: `apps/studio`'s `install:bundle` runs `npm install --no-package-lock`, so it ignores the lockfile and re-resolves `@wordpress/react-i18n: ^4.47.0` against the live registry on every build. `react-i18n@4.48.0`/`4.49.0` added a `react@^18` peer, which conflicts with the repo's react 19 (since #3642). `4.47.0` is the last version with no react peer; the lockfile already pins it, which is why dev, lint, and unit tests pass — only the lockfile-less bundling install floats and breaks.

This pins `@wordpress/react-i18n` to exact `4.47.0` in `apps/studio` and `apps/ui` (matching the exact-pin convention already used there for other `@wordpress/*` packages) so the bundling install can no longer float to the broken version. Behavior-neutral — dev already runs 4.47.0.

> Note: the **Performance Metrics** check on *this* PR will still show red, because that job builds the current (still-broken) trunk as its comparison baseline. That clears automatically once this lands on trunk. Since trunk is already red, merging this is strictly an improvement.

## Testing Instructions

- `git checkout` this branch and run the exact bundling install that CI runs:
  ```
  cd apps/studio && npm install --omit=dev --no-audit --no-fund --no-package-lock --no-progress --install-links --no-workspaces --dry-run
  ```
  It should exit 0 with no `npm error ERESOLVE` (only benign peer-override warnings).
- On `trunk` the same command fails with the ERESOLVE above.

## Pre-merge Checklist


epeicher (approved):
Thanks @bcotrim for fixing this! I can confirm that I can run the `cd apps/studio && npm install --omit=dev --no-audit --no-fund --no-package-lock --no-progress --install-links --no-workspaces --dry-run` script without any errors, while I have the `ERESOLVE` error in `trunk` for the same command. I agree that this PR can be merged with the error in the `Performance Metrics` check as it runs against `trunk` that errors. LGTM!

Image

PR #3988, opened 2026-06-29
Fix native PHP sites loading assets from remote siteurl/home (STU-1925)

bcotrim:
## Related issues

- Fixes STU-1925

## How AI was used in this PR

Claude Code investigated the root cause, implemented the fix, and verified it end-to-end against a pulled site (cotrim.dev). I reviewed the diff and tested manually via the CLI.

## Proposed Changes

When a site pulled from a remote (e.g. a WordPress.com staging site) runs on the native PHP runtime, it kept loading assets from the remote URL because the database's `siteurl`/`home` were served as-is. Native PHP now defines `WP_HOME`/`WP_SITEURL` before WordPress boots — derived from the request, so it survives dynamic ports and custom domains and never mutates the database (a push still sends the original remote URL). This also fixes a latent issue: PHP's built-in server ignores `auto_prepend_file` when a router script is used, so reprint's `runtime.php` never ran for imported native sites — the router now applies it explicitly.

## Testing Instructions

1. On the **native PHP** runtime, use a site whose database holds a remote `siteurl` (pull a WP.com staging site, or set `siteurl`/`home` to a remote URL in the DB).
2. Start the site, open it, and check DevTools → Network.
3. **Before:** theme/plugin assets (e.g. `style.css`) load from the remote domain. **After:** all assets load from `http://localhost:<port>`, and `wp-json` reports the local `url`/`home`.
4. `wp-cli` still reports the stored URL (intentionally untouched).

## Pre-merge Checklist


apps/cli/php/router.php
```diff
@@ -10,6 +10,15 @@
  * WP::parse_request() can resolve it against the rewrite rules.
  */
 
+// PHP's built-in server ignores the auto_prepend_file ini directive when a
+// router script is in use, so apply it ourselves before dispatching. This is
+// how Studio's pre-boot prepend (local WP_HOME/WP_SITEURL) and reprint's
+// runtime.php for imported sites get a chance to run.
```
fredrikekelund: This is not true.

bcotrim: I tested this further and it seems to be related with PHP version.
Same Studio site, without the fix in `router.php`

| PHP 8.3 | PHP 8.4 |
|--------|--------|
| image | image | 

Can you see that same behavior?

fredrikekelund: Yep, I see the same thing. Some PHP 8.4 release seems to have changed this behavior. 

`require_once` seems to work really well in this context. With this change, `auto_prepend_file` as a config directive just keeps working, regardless of PHP version. Nice 👍

apps/cli/lib/native-php/site-setup.ts
```diff
@@ -65,6 +67,51 @@ $transformer->to_file( $wp_config_path );
...
+
+	// Define WP_HOME/WP_SITEURL before WordPress boots so the site serves from
+	// the local URL regardless of the siteurl/home in the DB (e.g. after pulling
+	// a remote site — STU-1925). Pre-boot so derived URLs (WP_CONTENT_URL, etc.)
+	// resolve locally too. Taken from the request to survive dynamic ports/custom
+	// domains; wp-cli is left untouched.
+	return `<?php
+if ( PHP_SAPI !== 'cli' && ! empty( $_SERVER['HTTP_HOST'] ) ) {
+	$studio_is_https = ( ! empty( $_SERVER['HTTPS'] ) && $_SERVER['HTTPS'] !== 'off' )
+		|| ( ! empty( $_SERVER['HTTP_X_FORWARDED_PROTO'] ) && stripos( $_SERVER['HTTP_X_FORWARDED_PROTO'], 'https' ) !== false );
```
fredrikekelund: The PHP child processes never use HTTPS. Only the HTTPS proxy does that in in Studio. Might as well remove this part

apps/cli/lib/native-php/site-setup.ts
```diff
@@ -65,6 +67,51 @@ $transformer->to_file( $wp_config_path );
...
+	$studio_local_url = ( $studio_is_https ? 'https' : 'http' ) . '://' . $_SERVER['HTTP_HOST'];
```
fredrikekelund: I mentioned this on Slack, too, but I think we should generate this URL deterministically instead of looking at the `Host` header. There's just no reason to use a dynamic value here, since we already know the complete URL we want for `siteurl` and `home`

fredrikekelund (approved):
Good catch @shaunandrews and nice work, @bcotrim 👍 This is a nice and targeted fix, and I'm glad we found the `auto_prepend_file` discrepancy between PHP versions.

Let's generate the `siteurl` / `home` URL deterministically (since we already know which value we want). Other than that – this LGTM

apps/cli/lib/native-php/site-setup.ts
```diff
@@ -65,6 +68,48 @@ $transformer->to_file( $wp_config_path );
...
+ * URL. Returns the path to use as auto_prepend_file. For imported sites it
+ * chains to reprint's own runtime.php so that prepend still runs.
+ */
+export function writeSiteUrlPrependFile(
+	sitePath: string,
+	siteUrl: string,
+	originalAutoPrependFile?: string
+): string {
+	const hash = crypto.createHash( 'sha1' ).update( sitePath ).digest( 'hex' ).slice( 0, 16 );
+	const prependPath = path.join( os.tmpdir(), `studio-siteurl-prepend-${ hash }.php` );
```
fredrikekelund: This is fine, but we could use `fs.mktempd` for a similar effect, instead of generating a hash ourselves https://nodejs.org/docs/latest/api/fs.html#fspromisesmkdtempprefix-options

apps/cli/lib/native-php/site-setup.ts
```diff
@@ -65,6 +68,48 @@ $transformer->to_file( $wp_config_path );
...
+	const chained = originalAutoPrependFile
+		? `require '${ escapePhpSingleQuotedString( originalAutoPrependFile ) }';`
+		: '';
+
+	// Define WP_HOME/WP_SITEURL before WordPress boots so the site serves from
+	// the local URL regardless of the siteurl/home in the DB (e.g. after pulling
+	// a remote site — STU-1925). Pre-boot so derived URLs (WP_CONTENT_URL, etc.)
+	// resolve locally too. The URL is the site's configured local URL (custom
+	// domain or http://localhost:PORT), matching the Playground runtime's
+	// --site-url, rather than the request Host header.
```
fredrikekelund: I think we could slim this down, or indeed even remove it. Some of this is encoding the AI agent session history

fredrikekelund (approved):
Confirming that the latest changes look good and test well 👍

gcsecsey (approved):
I can also confirm that the changes work as described with a freshly cloned site, I get no requests targeting the remote.

Image

PR #3993, opened 2026-06-29
Recreate missing db.php drop-ins and preserve custom SQLite drop-ins

gcsecsey:
## Related issues

- Related to STU-1821 (fixes Problem 1: database suddenly disconnected after cloning)
- Fixes STU-1571

## How AI was used in this PR

I used Claude to investigate both issues, trace the shared root cause in the SQLite drop-in management policy, implement the fix, and write the unit and end-to-end tests. I reviewed the diagnosis and verified every branch of the new policy against real `db.php` files from local sites.

## Proposed Changes

The way Studio managed the `wp-content/db.php` SQLite drop-in was wrong in two ways. A configured site whose `db.php` went missing was never restored, leaving the site stuck on "Error establishing a database connection" (STU-1821 Problem 1). At the same time, a present `db.php` was reinstalled on every operation, breaking custom SQLite drop-ins like markdown-database-integration (STU-1571).

- Studio now ensures the SQLite drop-in on every site operation and recreates `db.php` whenever it is missing, so a cloned or repaired site can connect to its database again.
- The drop-in is now classified by its contents instead of merely by presence: our stock file is refreshed, a missing or foreign (non-SQLite) `db.php` is replaced so the site can boot, and a custom SQLite-compatible drop-in (or one marked `@studio-keep`) is preserved.
- After a pull, the drop-in is reinstalled even when the site server was stopped, because WordPress.com backups never bundle it.

## Testing Instructions

Core fix (STU-1821 Problem 1):

- Open a working site and confirm the front end and wp-admin both load.
- Stop the site, delete `wp-content/db.php`, then start the site again.
- Expect the site to boot with no `Error establishing a database connection` error, and to recreate `db.php`.

`db.php` policy (STU-1571):

For each case below, drop the file into `~/Studio/<site>/wp-content/db.php`, then stop and start the site. A unique comment that survives means the file was kept, if it is gone, the file was replaced.

- Missing, gets recreated: delete the file entirely.
- Foreign, gets replaced:

```php
<?php
// Query Monitor drop-in, not a SQLite drop-in
require_once 'qm-db.php';
```

- Custom SQLite drop-in, gets kept:

```php
<?php
// My custom SQLite drop-in
define( 'SQLITE_DB_DROPIN_VERSION', '1.8.0' );
require_once __DIR__ . '/mu-plugins/sqlite-database-integration/wp-includes/sqlite/db.php';
```

- Marked with @studio-keep, gets kept:

```php
<?php
// @studio-keep
define( 'SQLITE_DB_DROPIN_VERSION', '1.8.0' );
require_once __DIR__ . '/mu-plugins/sqlite-database-integration/wp-includes/sqlite/db.php';
```

- Stock drop-in (edited), gets replaced:

```php
<?php
/**
 * Plugin Name: SQLite integration (Drop-in)
 *
 * This file is auto-generated and copied from the sqlite plugin.
 */
define( 'SQLITE_DB_DROPIN_VERSION', '1.8.0' );
// EDIT ME, gone after next start
```

Pull path:

- With a connected site, stop the local site, then pull it from WP.com.
- Expect `db.php` to be present after the pull.

## Pre-merge Checklist


chubes4 (approved):
I tested the old MDI failure mode against this branch: a configured site with MDI's custom `wp-content/db.php`, with `@studio-keep` removed, running the Studio SQLite setup path via `keepSqliteIntegrationUpdated() -> installSqliteIntegration()`. The markerless MDI drop-in was preserved, stayed identifiable via `MARKDOWN_DB_DROPIN`, did not become the stock drop-in, and the SQLite mu-plugin was still refreshed.

Given that, this fixes the issue I cared about. One cleanup suggestion: if `SQLITE_DB_DROPIN_VERSION` is now the preservation contract for SQLite-compatible custom drop-ins, I think we can remove the `@studio-keep` escape hatch to avoid carrying two contracts. MDI can align to the new contract without the marker. Not blocking from me.

fredrikekelund (changes requested):
The `needsSqliteSetup` logic is meant to allow users to use MySQL instead of SQLite if they want, which I would think is a more common use case than bringing custom `db.php` drop-ins (although we can support both). See [documentation](https://developer.wordpress.com/docs/developer-tools/studio/frequently-asked-questions/#use-studio-with-mysql-server) about MySQL support for more context

wojtekn:
> The needsSqliteSetup logic is meant to allow users to use MySQL instead of SQLite if they want, which I would think is a more common use case than bringing custom db.php drop-ins (although we can support both). See [documentation](https://developer.wordpress.com/docs/developer-tools/studio/frequently-asked-questions/#use-studio-with-mysql-server) about MySQL support for more context

Agreed, this is a use case that allows users to use Studio with a MySQL server of their choice. Let's ensure that all the test cases from the spreadsheet are still supported.

gcsecsey:
> > The needsSqliteSetup logic is meant to allow users to use MySQL instead of SQLite if they want, which I would think is a more common use case than bringing custom db.php drop-ins (although we can support both). See [documentation](https://developer.wordpress.com/docs/developer-tools/studio/frequently-asked-questions/#use-studio-with-mysql-server) about MySQL support for more context
> 
> Agreed, this is a use case that allows users to use Studio with a MySQL server of their choice. Let's ensure that all the test cases from the spreadsheet are still supported.

@fredrikekelund @wojtekn Thanks for the catch, and for the MySQL docs link. I restored needsSqliteSetup with checks according to the docs, and updated the tests. @wojtekn could you share the spreadsheet you mentioned?

gcsecsey:
> I tested the old MDI failure mode against this branch: a configured site with MDI's custom `wp-content/db.php`, with `@studio-keep` removed, running the Studio SQLite setup path via `keepSqliteIntegrationUpdated() -> installSqliteIntegration()`. The markerless MDI drop-in was preserved, stayed identifiable via `MARKDOWN_DB_DROPIN`, did not become the stock drop-in, and the SQLite mu-plugin was still refreshed.
> 
> Given that, this fixes the issue I cared about. One cleanup suggestion: if `SQLITE_DB_DROPIN_VERSION` is now the preservation contract for SQLite-compatible custom drop-ins, I think we can remove the `@studio-keep` escape hatch to avoid carrying two contracts. MDI can align to the new contract without the marker. Not blocking from me.

Good call, thanks @chubes4! I checked that we don't have any public-facing docs about it, then removed `@studio-keep`. A custom db.php is now preserved if it defines `SQLITE_DB_DROPIN_VERSION` (and isn’t our stock drop-in, which we still refresh). Thanks for testing the markerless drop-in against the branch!

packages/common/lib/sqlite-integration.ts
```diff
@@ -78,12 +114,14 @@ export abstract class SqliteIntegrationProvider {
 		} );
 	}
 
-	async keepSqliteIntegrationUpdated( sitePath: string ): Promise< boolean > {
+	async keepSqliteIntegrationUpdated( sitePath: string ): Promise< void > {
+		// Skip SQLite for sites using their own database server (see needsSqliteSetup).
+		// Otherwise (re)install: this restores a missing db.php drop-in (otherwise the
+		// site reports "Error establishing a database connection") and keeps the bundled
+		// mu-plugin current, while installSqliteIntegration() preserves custom drop-ins.
```
fredrikekelund:
```suggestion
```
I've noticed that Claude has been getting very verbose recently. I'd argue this is a superfluous comment. It's fair to put a docblock above `needsSqliteSetup()`, but we don't need to re-explain those effects here, too.

packages/common/lib/sqlite-integration.ts
```diff
@@ -12,25 +17,56 @@ export abstract class SqliteIntegrationProvider {
...
+	 *
+	 * A Studio site can only boot through a drop-in the local SQLite runtime understands,
+	 * so a db.php is preserved exactly when it is a custom SQLite-compatible drop-in,
+	 * identified by defining SQLITE_DB_DROPIN_VERSION (how the SQLite plugin recognizes it):
+	 *  - missing/unreadable → don't keep (recreate it so the site can connect)
+	 *  - Studio's own stock drop-in → don't keep (refresh its path and version)
+	 *  - a custom drop-in defining SQLITE_DB_DROPIN_VERSION → keep (e.g. markdown-database-integration)
+	 *  - anything else, e.g. a plugin-owned db.php restored from a WordPress.com backup → don't keep (replace)
+	 */
 	async shouldKeepExistingDbDropin( sitePath: string ): Promise< boolean > {
```
fredrikekelund:
```suggestion
	async isExistingDbDropinAutoGenerated( sitePath: string ): Promise< boolean > {
```
Optional. I found the negated logic here somewhat difficult to reason about. Flipping the logic (and changing the method name) could help there.

apps/cli/commands/pull.ts
```diff
@@ -188,6 +188,16 @@ export async function runCommand(
...
+			// A WordPress.com backup never bundles the SQLite drop-in (Jetpack excludes
+			// wp-content/database/), so (re)install it after every import — even when the
+			// site server was stopped — so the freshly pulled site can connect to its database.
+			logger.reportStart(
+				LoggerAction.INSTALL_SQLITE,
+				__( 'Setting up SQLite integration, if needed…' )
+			);
+			await keepSqliteIntegrationUpdated( siteFolder );
+			logger.reportSuccess( __( 'SQLite integration configured as needed' ) );
+
```
fredrikekelund: This is sensible, but we also call `keepSqliteIntegrationUpdated()` when starting the site, so I don't really see where the previous logic would have failed…

fredrikekelund (approved):
LGTM 👍 This works as expected. 

I left a comment about changing the logic of `shouldKeepExistingDbDropin()` from negative to affirmative that'd be nice to address, but I'll let you decide about that, @gcsecsey.

packages/common/lib/sqlite-integration.ts
gcsecsey: I agree with flipping the logic, but I think `isExistingDbDropinAutoGenerated` is narrowing the name down a bit too much. We have three states: missing, stock (auto-generated), and custom. So I'd go with something like `shouldReplaceDbDropin` to still do the flip but keep the naming broader.

gcsecsey: I went with `shouldReplaceDbDropin `, and updated the call site and tests in [2f949be](https://github.com/Automattic/studio/pull/3993/commits/2f949be24d12218e33a1f0079137fb960bd070c2)

packages/common/lib/sqlite-integration.ts
gcsecsey: I agree this comment is redundant here, thanks for the suggestion!

apps/cli/commands/pull.ts
gcsecsey: Good catch, thanks, this is redundant with the start call.

Unfortunately, I don't have the exact logs from the reporter, but my suspicion is that the issue was the old `needsSqliteSetup` logic, it returned `hasDbPhp || !hasWpConfig`, so a pulled site which has `wp-config.php` and `wp-content/database/.ht.sqlite`, but no `db.php` returned `false`, and both the pull-time and start-time calls skipped it.

Now that `needsSqliteSetup` is also checking `.ht.sqlite` and the mu-plugin, `start.ts` already recreates the drop-in on the next start. I'll remove the post-import call.

PR #3994, opened 2026-06-29
Fix `studio push` for native reprint-pulled sites

epeicher:
## Related issues

- Related to STU-1950

## How AI was used in this PR

Claude Code was used throughout: to reproduce and diagnose the failure, to discover that there were **two distinct blockers** (a DB-connection error and a separate "could not locate the SQLite integration plugin" error), to validate the fix design with controlled experiments against a real reprint-pulled site, to implement the changes, and to verify the full push end-to-end. All generated code was reviewed.

## Proposed Changes

Makes `studio push` work for sites created by `studio pull-reprint` on the **native-php** runtime. The export step previously failed — first with "Error establishing a database connection", then "Database export failed". The same WP-CLI path backs the desktop "get theme details" / "get site icon" features, which failed identically for these sites and are fixed too.

**Why it failed:** a reprint pull wires SQLite only through a generated `runtime.php`, which the web server applies as a PHP `auto_prepend_file`; these sites have no `wp-content/db.php` drop-in. The server reaches the database, but standalone WP-CLI — which push uses for `wp plugin list`, `wp theme list`, and `wp sqlite export` — never loads `runtime.php`. That produced two separate failures:

1. `wp plugin list` / `theme list` fell back to MySQL → "Error establishing a database connection".
2. `wp sqlite export` additionally needs the SQLite integration plugin discoverable in `wp-content`, which reprint sites don't have → "Could not locate the SQLite integration plugin".

WASM/Playground imported sites were never affected because their pull installs a real `db.php` drop-in; native imported sites never did.

**The fix** is two coordinated changes:
- The native WP-CLI runner loads the site's `runtime.php` as `auto_prepend_file` for imported sites, restoring the database connection for general WP-CLI.
- The sqlite-command (`wp sqlite export`/`tables`) is the exception: it loads its own integration copy and reads the database file directly, so it runs **without** the prepend, and push installs the SQLite integration into the imported site's `wp-content` so the command can find it.

Safe on both ends: `runtime.php` takes precedence over the drop-in on the running server (`require_wp_db()` returns early when `$wpdb` is set, so the drop-in is inert), and the exporter already excludes `db.php` and `sqlite-database-integration` from the upload, so neither reaches the MySQL remote.

## Testing Instructions

1. Apply the changes on this branch and run `npm run cli:build`
2. Create a Studio site or use an existing Studio site
3. Run `STUDIO_ENABLE_PULL_REPRINT=true node apps/cli/dist/cli/main.mjs pull-reprint --path <path-to-Studio-site>` to `pull` a remote site using Reprint, e.g. `STUDIO_ENABLE_PULL_REPRINT=true node apps/cli/dist/cli/main.mjs pull-reprint --path ~/Studio/my-shiny-website`. Select the remote site to pull and wait for completion
4. Open Studio and connect to the remote site selected in previous step, or use the following CLI command on the Studio folder `cd <path-to-Studio-site> && node ~/github/studio/apps/cli/dist/cli/main.mjs push`, e.g. `cd ~/Studio/my-shiny-website && node ~/github/studio/apps/cli/dist/cli/main.mjs push`
5. Verify that the `Push` complete successfully


apps/cli/commands/push.ts
```diff
@@ -57,6 +61,14 @@ export async function runCommand(
...
 	);
 	await keepSqliteIntegrationUpdated( siteFolder );
+	// Reprint-pulled (imported) sites wire SQLite through runtime.php and ship no db.php
+	// drop-in, so the step above skips them. The database export below uses `wp sqlite
+	// export`, which requires the SQLite integration to be discoverable in wp-content.
+	// Install it for imported sites; the exporter excludes db.php and the integration from
+	// the upload, so it never reaches the remote.
+	if ( site.runtimeBlueprintPath && ! ( await isSqliteIntegrationInstalled( siteFolder ) ) ) {
+		await installSqliteIntegration( siteFolder );
+	}
```
fredrikekelund: Could we if/else here?

epeicher: Done as part of 1ce73b210

apps/cli/lib/run-wp-cli-command.ts
```diff
@@ -188,10 +189,24 @@ async function runNativeWpCliCommand(
...
+	//
+	// Exception: the sqlite-command (`wp sqlite export`/`tables`, requireSqliteCliCommand)
+	// loads its own copy of the SQLite integration and reads the database file directly.
+	// Prepending runtime.php there loads a second integration copy and fatals, so it runs
+	// without the prepend and relies on the integration installed in wp-content (see the
+	// imported-site setup in push.ts).
+	const autoPrependFile = options.requireSqliteCliCommand
+		? undefined
+		: getImportedSiteAutoPrependFile( site );
+	const prependArgs = autoPrependFile ? [ '-d', `auto_prepend_file=${ autoPrependFile }` ] : [];
```
fredrikekelund: Can we move this into `getDefaultPhpArgs()` and edit `spawnPhpProcess()` in `apps/cli/lib/native-php/php-process.ts` accordingly?

apps/cli/lib/pull/runtime-start-options.ts
```diff
@@ -191,6 +191,29 @@ export function loadImportedRuntimeStartOptionsNative(
...
+ * connection". Loading the same file the server uses gives WP-CLI matching SQLite wiring.
+ *
+ * Only reprint/imported sites have `runtimeBlueprintPath`; `runtime.php` is its sibling
+ * (the same directory the native server-start path resolves it from above). Returns
+ * undefined — never throws — for normal `studio create` sites or when the file is absent,
+ * so those are unaffected.
+ */
+export function getImportedSiteAutoPrependFile( site: {
+	runtimeBlueprintPath?: string;
+} ): string | undefined {
```
fredrikekelund:
```suggestion
export function getImportedSiteAutoPrependFile( site: SiteData ): string | undefined {
```
We might as well use the canonical type here, unless we have good reason to take a partial type

apps/cli/lib/pull/runtime-start-options.ts
```diff
@@ -191,6 +191,29 @@ export function loadImportedRuntimeStartOptionsNative(
...
+ * Reprint-pulled sites wire SQLite only through `runtime.php`, which the web server
+ * applies as a PHP `auto_prepend_file`. A standalone native WP-CLI process never loads
+ * it, so WordPress falls back to MySQL and fails with "Error establishing a database
+ * connection". Loading the same file the server uses gives WP-CLI matching SQLite wiring.
+ *
+ * Only reprint/imported sites have `runtimeBlueprintPath`; `runtime.php` is its sibling
+ * (the same directory the native server-start path resolves it from above). Returns
+ * undefined — never throws — for normal `studio create` sites or when the file is absent,
+ * so those are unaffected.
+ */
```
fredrikekelund, line 187: I think this can be shortened quite a bit. It's just unnecessarily verbose

apps/cli/commands/push.ts
```diff
@@ -56,7 +60,16 @@ export async function runCommand(
...
+	// Reprint-pulled (imported) sites wire SQLite through runtime.php and ship no db.php
+	// drop-in, so keepSqliteIntegrationUpdated skips them. The database export below uses
+	// `wp sqlite export`, which requires the SQLite integration to be discoverable in
+	// wp-content. Install it for imported sites; the exporter excludes db.php and the
+	// integration from the upload, so it never reaches the remote.
+	if ( site.runtimeBlueprintPath && ! ( await isSqliteIntegrationInstalled( siteFolder ) ) ) {
+		await installSqliteIntegration( siteFolder );
+	} else {
+		await keepSqliteIntegrationUpdated( siteFolder );
+	}
```
fredrikekelund: I noticed the `export` CLI command failing for a Reprint-pulled site on the database step, and I assume that's because of a missing "regular" SQLite integration. That's arguably a separate issue, but it seems small enough that we might fix it here. Either by moving this SQLite integration updating/installation into the exporter, or by copying this logic to the `export` command.

We should also consider preview site creation.

fredrikekelund (commented):
Nice and targeted fix 👍 Looks good overall, although it'd be nice to address the SQLite integration issue in the `export` command here, too. Let me know your thoughts there, @epeicher

apps/cli/commands/push.ts
epeicher: Sure, let me test and address those commands, if it turns out to be more complex, I might create a follow-up

apps/cli/lib/pull/runtime-start-options.ts
epeicher: Done as part of b4a1fee

apps/cli/lib/pull/runtime-start-options.ts
epeicher: Done as part of b4a1fee and 6b2dec0

apps/cli/lib/run-wp-cli-command.ts
epeicher: Ok, that reduces the code a bit, I have done the refactor as part of 22f127b

apps/cli/commands/push.ts
epeicher: > I noticed the export CLI command failing for a Reprint-pulled site on the database step

@fredrikekelund, how can I reproduce the error? I am testing with these changes and I'm able to export a site using the CLI command successfully 
CleanShot 2026-07-01 at 17 49 27@2x

fredrikekelund: I double checked these steps just now:

1. `npm run cli:build`
2. `studiodev site create` (that's the alias I use for the CLI dev build)
3. `cd` into the site directory
4. `STUDIO_ENABLE_PULL_REPRINT=1 studiodev pull-reprint`
5. `studiodev site stop` (still needed until https://github.com/Automattic/studio/pull/4007 lands) 
6. `studiodev site start`
7. `studiodev export`

The last step fail like so: `✖ Export failed: Database export failed`

epeicher: @fredrikekelund I have created a follow-up https://github.com/Automattic/studio/pull/4038 to fix the `create preview` error as it doesn't seem to be related to these changes. Also, I have suggested to fix the root cause on Reprint as part of https://github.com/WordPress/reprint/pull/296

epeicher: Thanks for double checking! By following those exact steps I have reproduced the issue, I have fixed it as part of 40c61b0 so this is ready for another review 
CleanShot 2026-07-02 at 09 41 03@2x

apps/cli/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -282,6 +283,18 @@ export class DefaultExporter extends ImportExportEventEmitter implements Exporte
...
+		// below requires the SQLite integration to be discoverable in wp-content, so install
+		// it for imported sites. It's excluded from the archive (see isExactPathExcluded),
+		// so it never reaches the backup or the remote.
+		if (
+			this.options.site.runtimeBlueprintPath &&
+			! ( await isSqliteIntegrationInstalled( this.options.site.path ) )
+		) {
+			await installSqliteIntegration( this.options.site.path );
+		}
+
```
fredrikekelund: It'd be nice to defer this "is this a reprint site" check to the lower-level API here eventually. We can tackle that in another PR, though

apps/cli/lib/pull/runtime-start-options.ts
```diff
@@ -191,6 +192,24 @@ export function loadImportedRuntimeStartOptionsNative(
...
+ * Returns undefined — never throws — for normal `studio create` sites or when the file
+ * is absent, so existing non-reprint sites are unaffected.
+ */
+export function getImportedSiteAutoPrependFile( site: SiteData ): string | undefined {
+	if ( ! site.runtimeBlueprintPath ) {
+		return undefined;
+	}
+	const runtimePhpPath = path.join( path.dirname( site.runtimeBlueprintPath ), 'runtime.php' );
+	return fs.existsSync( runtimePhpPath ) ? runtimePhpPath : undefined;
+}
```
fredrikekelund, line 204: This can also be a follow-up, but this is essentially duplicating the responsibility of `loadImportedRuntimeStartOptionsNative()`, so we might want to look into folding those two functions into one in another PR

apps/cli/lib/run-wp-cli-command.ts
```diff
@@ -186,8 +187,19 @@ async function runNativeWpCliCommand(
...
 	await writeStudioMuPluginsForNativePhpRuntime( site.path, site.isWpAutoUpdating );
 
+	// Reprint-pulled sites wire SQLite through runtime.php (loaded as auto_prepend_file),
+	// so load it here too. No-op for normal sites (helper returns undefined).
+	//
+	// Exception: the sqlite-command (`wp sqlite export`/`tables`, requireSqliteCliCommand)
+	// loads its own copy of the SQLite integration and reads the database file directly.
+	// Prepending runtime.php there loads a second integration copy and fatals, so it runs
+	// without the prepend and relies on the integration installed in wp-content (see the
+	// imported-site setup in push.ts).
```
fredrikekelund: Documenting why the behavior changes depending on the value of `options.requireSqliteCliCommand` is sensible, but Claude (which I assume this is) is really being very verbose with comments lately. Over time, too verbose comments makes the code more difficult to understand rather than the opposite. 

Your call here, @epeicher, but I wanted to mention it

fredrikekelund (approved):
Clean fix 👍 Works great. I'll take a look at https://github.com/Automattic/studio/pull/4038 next

apps/cli/lib/run-wp-cli-command.ts
epeicher: Thanks for mentioning it. I guess we need to find a middle ground here, let me rephrase this to be less verbose.

apps/cli/lib/pull/runtime-start-options.ts
epeicher: This has been addressed, now we only have `loadImportedRuntimeStartOptionsNative`

epeicher:
After all changes, the following tests have been done successfully:
- Create a site
- Pull a remote by using `pull-reprint`
- Stop/start the site - Verified it works fine
- Pull again so it is an incremental pull
- Stop/start the site - Verified it works fine
- Push the site - Verified the remote site works fine
- Exported the site

PR #3995, opened 2026-06-29
Adjust native PHP vs Playground agent expectations

katinthehatsite:
## Related issues

Fixes STU-1840

## How AI was used in this PR

It was used to plan and implement the feature. 

## Proposed Changes

This PR ensures that when the site is using the native PHP runtime, the agents are using the instructions that are tailored to it. When a site is using a Playground sandbox, the agents are given the correspoding instructions. 

## Testing Instructions

* Run `npm install`
* Rebuild CLI with `npm run cli:build`
* Create a site with native PHP by running:

`node apps/cli/dist/cli/main.mjs site create --name test-native --runtime native`

* Create a site with Playground sandbox:

`node apps/cli/dist/cli/main.mjs site create --name test-sandbox --runtime sandbox`
* Once done, run `open ~/Studio/test-native/STUDIO.md  ` in your terminal and confirm that you can see references such as `Core edits are overwritten on every WordPress update, so extend through hooks instead`
* Repeat the same step for `open ~/Studio/test-sandbox/STUDIO.md  ` and confirm the references are correct for Playground

## Pre-merge Checklist


apps/cli/ai/runtimes/pi/index.ts
```diff
@@ -255,6 +261,26 @@ async function runAgentSessionTurn(
...
+// or unreadable sites.
+async function resolveActiveSiteRuntime(
+	activeSite: SiteInfo | null | undefined
+): Promise< SiteRuntime > {
+	if ( ! activeSite || activeSite.remote || ! activeSite.path ) {
+		return SITE_RUNTIME_NATIVE_PHP;
+	}
+	try {
+		const config = await readCliConfig();
+		const site = config.sites?.find( ( candidate ) => candidate.path === activeSite.path );
```
fredrikekelund:
```suggestion
		const site = await getSiteByFolder( activeSite.path );
```
This isn't just shorter, but the difference in how paths are compared can actually be relevant depending on the underlying file system

packages/common/lib/agent-skills.ts
```diff
@@ -95,6 +133,7 @@ export async function installSkillToSite(
 	sitePath: string,
 	bundledPath: string,
 	skillId: string,
+	runtime: SiteRuntime,
 	overwrite: boolean
```
fredrikekelund: `runtime` could be derived from `sitePath`. Or even better, we could take a `SiteData` object as the first argument and derive the runtime from it instead of calling `getSiteByFolder` again.

katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/3995/changes/2206cc5901572b95c487f6168fd9db8114379e11 - thanks for the suggestion 👍

apps/cli/ai/runtimes/pi/index.ts
katinthehatsite: Adjusted in https://github.com/Automattic/studio/pull/3995/changes/97a5b6cf33204dd7ba184650e655dc64f86fea4c

katinthehatsite:
@fredrikekelund I addressed the comments, it is ready for another review

nightnei:
> Once done, run open ~/Studio/test-native/STUDIO.md   in your terminal and confirm that you can see references such as "Core edits are overwritten on every WordPress update, so extend through hooks instead"

Checked, but I don't see it inside `test-native/STUDIO.md`.

UPDATE: oh, I see, there is important step missed to do `npm install`, otherwise it bundles the stale STUDIO.md
Works well 👍

katinthehatsite:
> UPDATE: oh, I see, there is important step missed to do npm install, otherwise it bundles the stale STUDIO.md

Thanks for the review @nightnei ! I will update the testing instructions 👍 

Do you think we are good to merge?

apps/cli/ai/system-prompt.ts
```diff
@@ -72,7 +77,29 @@ IMPORTANT: ${ PLAN_DATA_GUARDRAIL }
...
+// Guidance for delivering `--post_content` to `wp_cli`. The shared part applies
+// to both runtimes (the tool never runs a shell). The runtime-specific part
+// differs: Playground runs in a WASM sandbox that cannot see the host
+// filesystem, so content must be passed inline and Studio rewrites large
+// content to a virtual temp file. The native PHP runtime reads the real
+// filesystem, so a scratch file is allowed and is the better choice for large
+// content (inline args can hit the OS command-length limit).
+function getPostContentGuidance( runtime?: SiteRuntime ): string {
+	const shared =
+		'The `wp_cli` tool takes literal arguments, not shell commands — never use shell substitution or shell syntax such as `$(cat file)`, backticks, pipes, redirection, or environment variables to provide post content.';
```
youknowriad, line 89: Is this needed for the php runtime?

katinthehatsite: From my testing, it seems to be needed: The `wp_cli` tool uses spawn() without a shell on both runtimes, so shell syntax like `$(cat file)` is passed as literal text. I verified this by creating a post with $(echo INJECTED) as content on a native PHP site and it was stored verbatim, not executed.

fredrikekelund: If this is an issue, consider changing how WP-CLI is spawned and passing `shell: true`. This might be more complex for the Playground-powered WP-CLI, but for the native version, it should be relatively straightforward

PR #3997, opened 2026-06-29
Add MySQL proof of concept

shaunandrews:
## Proposed Changes

- Adds a proof-of-concept MySQL path for Studio native PHP sites while keeping SQLite as the default database engine.
- Lets users opt into MySQL from advanced site creation settings or the CLI, with Playground sites rejecting MySQL clearly.
- Downloads and verifies a managed MySQL runtime on demand instead of increasing every Studio install size.
- Starts and provisions a per-site MySQL database so native PHP, WordPress install, Blueprint execution, and WP-CLI can run against real MySQL.
- Gates unsupported MySQL workflows in the desktop UI where the current experience is still SQLite-specific.
- Documents the strategy, binary delivery model, implementation plan, and remaining product work.

image

## What Works Now

- Create a native PHP site with `databaseEngine=mysql`.
- Download MySQL on demand into `~/.studio/mysql-bin/8.4.10/`.
- Store per-site data under `~/.studio/mysql-data/<site-id>/`.
- Start and stop MySQL with the native PHP site.
- Generate `wp-config.php` with real MySQL credentials.
- Apply empty-site Blueprints against MySQL after the follow-up fix.
- Block MySQL for sandbox/Playground sites.
- Disable Sync UI for MySQL sites.

## Big Remaining Work

1. **Platform binaries**
   - Current metadata only includes `darwin-arm64`.
   - Missing: macOS Intel, Windows x64, and Linux.
   - Windows code paths partially exist for `.exe` and `.zip`, but are untested.
   - Linux needs archive selection and extractor/runtime validation.

2. **Binary delivery policy**
   - Decide final source: Oracle MySQL vs MariaDB.
   - Add metadata with URL, SHA, archive type, and root dir for each platform.
   - Confirm licensing and redistribution requirements.
   - Keep on-demand download as default unless product wants an offline-first bundle.

3. **Lifecycle hardening**
   - Handle orphaned `mysqld` after app crashes.
   - Add install locks so two sites cannot race while downloading/extracting MySQL.
   - Improve cleanup of runtime socket/PID dirs.
   - Decide per-site MySQL process vs shared global server. Current PoC is effectively per-site.

4. **Feature boundaries**
   - Sync is disabled for MySQL right now.
   - Import/export/backup workflows need review.
   - Pull from WordPress.com/Pressable into MySQL is not solved.
   - Site duplication, deletion, reset, and migration need explicit MySQL behavior.

5. **User experience**
   - Improve first-use messaging: “Downloading MySQL, about 160-270 MB.”
   - Improve failure messages for unsupported platform, offline state, and hash mismatch.
   - Consider hiding MySQL unless native PHP is selected.
   - Add a visible “Database engine: MySQL” indicator in site details/status.

6. **WP-CLI / developer tooling**
   - Internal WP-CLI paths can work if MySQL is running.
   - Public CLI does not currently expose a clean `studio wp ...` command in this branch.
   - Add or restore a supported WP-CLI command path that starts MySQL when needed.

7. **Tests and CI**
   - Add unit tests for metadata/platform resolution.
   - Add tests for MySQL download/extract paths, including Windows archive shape.
   - Add lifecycle tests for start/stop/restart.
   - Add at least one e2e smoke test per supported OS if CI can tolerate it.

8. **Upgrade story**
   - Decide what happens when MySQL 8.4.11 ships.
   - Decide whether old runtimes stay installed.
   - Define how existing data dirs are migrated or verified.
   - Define recovery behavior for failed runtime upgrades.


chubes4:
## Fanout concurrency: proxy POST-pin fixed; a second upstream constraint isolated

### Fix in this commit (proxy layer)
`shouldUsePrimaryWorker()` in `apps/cli/php-server-child.ts` pinned **every** non-GET/HEAD/OPTIONS request to worker 0. That method-based rule was an over-broad stand-in for "stateful admin request" and swept in all of WordPress's own POST loopbacks — including Action Scheduler's async queue runner (`admin-ajax.php?action=as_async_request_queue_runner`) and wp-cron. Async fanout fires N concurrent loopback POSTs to wake N workers; with the method pin all N serialized on worker 0.

Now affinity is pinned by **path** only. phpMyAdmin (the one genuinely stateful route — file-based session store scoped to one worker) keeps its existing `/phpmyadmin` pin, which already covers all its methods. WordPress is stateless across the pool (sessions live in shared MySQL), so its POSTs load-balance like GETs.

**Proxy-level proof (8-worker pool, `getmypid` probe):**
| requests | before | after |
|---|---|---|
| 8 concurrent POSTs | 1 distinct worker | **8 distinct workers** |
| 8 concurrent GETs | 8 distinct workers | 8 distinct workers |

### Second constraint isolated (upstream — NOT the proxy, NOT this repo)
With the proxy fixed, a 5-page `site-forge generate` still drained its 5 branches **serially at the 60s WP-Cron heartbeat cadence** (peak 2 concurrent), not in parallel. Root cause is in **agents-api**, not Studio: `WP_Agent_Workflow_Action_Scheduler_Branch_Executor::loopback_dispatch_target()` rewrites the loopback host to `127.0.0.1` but keeps the `https://` scheme and default :443. This HTTPS `.local` runtime's worker-pool proxy is **plain HTTP on `localhost:<port>`**; there is no HTTPS listener on `127.0.0.1:443`, so every concurrent async-runner POST fails the TLS handshake and the branches fall back to the serial WP-Cron drain.

**Isolation proof (same runtime, same 8-worker pool, `WpOrg\Requests::request_multiple`):**
- Current target `https://127.0.0.1/wp-admin/admin-ajax.php…` → **0/8 succeed** (`SSL routines::ssl/tls alert handshake failure`)
- Correct target `http://localhost:<port>/…` + canonical `Host:` header → **8/8 succeed (200)**, and the 8 runners actually claim + drain branches

So the branch-fanout mechanism, the AS concurrency policy, and this proxy fix are all correct; the remaining cap is agents-api constructing an HTTPS/:443 loopback target that doesn't match a plain-HTTP native-PHP worker-pool runtime. That fix belongs upstream in agents-api's `loopback_dispatch_target` (derive scheme/port from the actual local server, or reuse the plain-HTTP proxy port), and is deliberately not papered over here.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Code (Opus 4.8)
- **Used for:** root-cause analysis, the proxy fix, and the proxy-level + end-to-end concurrency proofs

PR #4000, opened 2026-06-30
CLI test: add HTTP response check for created site

gavande1:
## Related issues

- Related to STU-1867

## How AI was used in this PR

Claude was used to write the code and explore the solution. The author reviewed it manually.

## Proposed Changes

Follow-up to the CLI e2e tests in #3947. The `site create` test only asserted persisted state on disk. This adds a real check: after creating and starting a site, the test requests its homepage and confirms a 200 with HTML, so it now fails if a created site cannot be served.

This does not cover the custom-domain and HTTPS case, which cannot run in an isolated, headless test. See the discussion in #3947.

## Testing Instructions

1. Build the CLI with `npm run cli:build`.
2. Run `npm test -- apps/cli/commands/site/tests/create.e2e.test.ts --tagsFilter='e2e'` and confirm both tests pass.
3. Run the same with `--tagsFilter='!e2e'` and confirm the tests are skipped, so the fast suite is unaffected.

## Pre-merge Checklist


apps/cli/commands/site/tests/create.e2e.test.ts
```diff
@@ -18,57 +18,67 @@ import {
...
+					'create',
+					'--name',
+					siteName,
+					'--path',
+					sitePath,
+					'--wp',
+					'latest',
+					'--skip-browser',
+					'--skip-log-details',
+				],
```
gcsecsey, line 60: I got this flagged by Claude:

> Because this command no longer passes `--runtime sandbox`, site create uses the CLI default native runtime. The e2e harness only provisions bundled WordPress, not php-bin, so this test can download PHP or fail in offline/release environments instead of staying hermetic like the other CLI e2e tests.

As a temporary measure, we could adjust the test to pass the `--runtime sandbox` parameter, but as a long-term solution, I think we should extend the e2e harness to download PHP as well.

gcsecsey (commented):
Thanks for adding the HTTP checks @gavande1! 🙌 

The new tests are working well:
Image

And they're skipped when filtered out as expected:
Image

I added a note about the PHP binary being available on CI.

apps/cli/commands/site/tests/create.e2e.test.ts
gavande1: Added `--runtime sandbox` in 18a7bb3.

PR #4002, opened 2026-06-30
Add data-liberation-agent into Studio

nightnei:
## Related issues

Fixes STU-1965

## How AI was used in this PR
Just to review and confirm that we can merge it without changes as the initial step.

## Proposed Changes
The discussion - pfHvTO-1uh-p2#comment-1341
Let's keep the cloned codebase intact in the initial commit.
Later we will submit follow-up PRs to clean it up and adapt it to our use case (More details - pfHvTO-1uh-p2#comment-1341).

The reason is that the PR is already huge. If we start modifying the code as part of the migration, it will become nearly impossible to review.

## Testing Instructions
* Assert that CI is green


borkweb (approved):
I just did a file diff and things look good!

PR #4001, opened 2026-06-30
Refactor `pull-reprint` to require an existing local site

fredrikekelund:
## Related issues

- Related to STU-1930

## How AI was used in this PR

I relied on Claude to first understand the existing `pull-reprint` state management, then to plan a new `cli.json`-based approach where Reprint state is stored on individual site objects. Claude produced a work specification split across three PRs, and then I ran the agent for each of those specs in new sessions.

## Proposed Changes

`pull-reprint` state revolves around pulls, which is a foreign concept in Studio. In other areas, Studio state revolves heavily around sites. There's not really a good reason for this fundamental discrepancy in `pull-reprint`, so this PR begins making the state more site-centric by requiring an existing Studio site when `pull-reprint` is triggered. There will be at least two more PRs on this train.

Concretely, this PR does the following:

- Rely on the `--path` CLI option to load a Studio site first thing in `pull-reprint`
- Remove the `--abort` option. `studio site delete` does the same thing.

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node apps/cli/dist/cli/main.mjs site create`. Step through the flow and take note of the site path.
3. Run `STUDIO_ENABLE_PULL_REPRINT=1 node apps/cli/dist/cli/main.mjs pull-reprint --path PATH_TO_SITE`
4. You will need to restart the site after the pull completes. This is temporary and will be fixed in a follow-up PR.
5. Ensure that the site loads successfully after the pull.

## Pre-merge Checklist

- [ ] Have you checked for TypeScript, React or other console errors?

PR #4003, opened 2026-06-30
Store most `pull-reprint` state in `cli.json`

fredrikekelund:
## Related issues

- Related to STU-1930

## How AI was used in this PR

I relied on Claude to first understand the existing `pull-reprint` state management, then to plan a new `cli.json`-based approach where Reprint state is stored on individual site objects. Claude produced a work specification split across three PRs (this is the second), and then I ran the agent for each of those specs in new sessions.

## Proposed Changes

This PR follows up on https://github.com/Automattic/studio/pull/4001 by moving all `pull.json` state but the `stage` field to the site object in `cli.json`.

As mentioned in the previous PR, `pull-reprint` state revolves around pulls, which is a foreign concept in Studio. In other areas, Studio state revolves heavily around sites. The end goal with this PR train is to remove `pull.json` entirely and instead assume that there's always a Studio site to work with when `pull-reprint` is called (i.e., `site create` was previously called) and that the `pull-reprint`-related state is stored in `cli.json`, along with other site state.

Concretely, this PR does the following:

- Add the `reprintOrigin` and `importComplete` fields to the site schema for `cli.json`
- Instead of storing `PullSessionMetadata` in `pull.json`, load and derive it from site object properties in `cli.json`

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node apps/cli/dist/cli/main.mjs site create`. Step through the flow and take note of the site path.
3. Run `STUDIO_ENABLE_PULL_REPRINT=1 node apps/cli/dist/cli/main.mjs pull-reprint --path PATH_TO_SITE`
4. You will need to restart the site after the pull completes. This is temporary and will be fixed in a follow-up PR.
5. Ensure that the site loads successfully after the pull.

## Pre-merge Checklist


apps/cli/commands/pull-reprint.ts
```diff
@@ -679,54 +657,76 @@ function hasPullCompletedStage( metadata: PullSessionMetadata, stage: PullStage
...
+	const reprintState = readReprintState( stateDirectory );
+	if ( reprintState?.filter && reprintState.filter !== 'essential-files' ) {
+		writeReprintState( stateDirectory, {
+			...reprintState,
+			status: 'complete',
+			stage: null,
+			filter: 'essential-files',
+		} );
+	}
+}
```
epeicher, line 677: Thanks for fixing this @fredrikekelund! I have had that issue in previous testing.

With regards to this function that updates the `Reprint` state, what do you think of moving that to the [reprint-state.ts](https://github.com/Automattic/studio/pull/4003/changes#diff-fe3af5f3dd5d6df7bd68e8d39018ca19b9635e9ddd3834a13314ee51b09e9b5f) component? That component claims to be "the single seam where Studio couples to reprint's internal schema"

epeicher (approved):
I have reviewed and tested the code, and it works as expected. I haven't found any issues, and I can pull the site successfully. I just left a non-blocking question, but this LGTM!

Image

apps/cli/commands/pull-reprint.ts
fredrikekelund: Sure, I can do that 👍

I'm looking to remove all Studio CLI logic that directly modifies the Reprint state on disk (due to the obvious associated risks). https://github.com/WordPress/reprint/pull/292 would allow us to remove this logic and the initial pieces in `downloadSkippedFiles()`. No harm in moving this function in the meantime, though

epeicher: That sounds like a great plan, and yes, the function `downloadSkippedFiles()` also modifies the state on disk, and it caught my attention too. If that function is going to be removed on future refactors I don't mind if it stays there temporarily 👍. Thanks for the extra context!

PR #4005, opened 2026-06-30
Inject a DB_NAME fallback for native WP-CLI commands

youknowriad:
## Proposed Changes

Full-site export builds a `meta.json` by running native WP-CLI commands (e.g. `plugin list`) that boot WordPress. On sites whose `wp-config.php` omits `DB_NAME` — common for pulled/imported sites, or sites only ever run under Playground (which supplies the DB connection at runtime) — the v3+ SQLite integration drop-in requires a non-empty `DB_NAME`, so those commands fail with “Error establishing a database connection.”

This injects the same fallback the Playground path already defines (`define('DB_NAME','wordpress')`) into the native WP-CLI invocation via `--exec` — but only when `wp-config.php` doesn't already define `DB_NAME`, so a real value is never redefined and sites that define it are untouched.

## Testing Instructions

- `npm test -- apps/cli/lib/tests/run-wp-cli-command.test.ts` (4 tests).
- Full-export a site whose `wp-config.php` has no `DB_NAME` and confirm it no longer errors on the DB connection.

### DIFF
--- a/apps/cli/lib/run-wp-cli-command.ts
+++ b/apps/cli/lib/run-wp-cli-command.ts
@@ -1,4 +1,5 @@
 import { ChildProcess, spawn } from 'node:child_process';
+import { readFileSync } from 'node:fs';
 import path from 'node:path';
 import { PassThrough, Readable } from 'node:stream';
 import { buffer, text } from 'node:stream/consumers';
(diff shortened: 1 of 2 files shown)


youknowriad:
While working on something unrelated, my agent suggested this fix. I have to admin I have no idea if this is valid or not. cc @bcotrim @sejas just in case? Otherwise we can just close this PR.

PR #4006, opened 2026-06-30
Add design doc: Studio apps & surfaces

youknowriad:
## Proposed Changes

Adds a design doc (`docs/design-docs/studio-apps-and-surfaces.md`) clarifying the different Studio apps and surfaces — desktop, hosted, local (`studio ui`), and CLI — the architecture of each, the desktop/local backend convergence in `@studio/common`, and the shared `apps/ui` layer that serves all three browser surfaces.

Documentation only; no code changes.

### DIFF
--- a/docs/design-docs/studio-apps-and-surfaces.md
+++ b/docs/design-docs/studio-apps-and-surfaces.md
@@ -1,0 +1,188 @@
+# Studio Apps & Surfaces
+
+## About this doc
+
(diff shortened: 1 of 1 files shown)


wojtekn:
Looks clear, thanks for documenting it.

However, it's unclear why users would need a local web variant - it would make sense to explain the reasoning behind it.

wojtekn (approved):

youknowriad:
> However, it's unclear why users would need a local web variant - it would make sense to explain the reasoning behind it.

There are number of users that are using Studio CLI today without the app, including with custom agents and things like that. Studio local UI is a way to meet these users where they are and provide them with an on demand UI.

PR #4007, opened 2026-06-30
Fully remove `pull.json` state

fredrikekelund:
## Related issues

- Fixes STU-1930

## How AI was used in this PR

I relied on Claude to first understand the existing `pull-reprint` state management, then to plan a new `cli.json`-based approach where Reprint state is stored on individual site objects. Claude produced a work specification split across three PRs (this is the third), and then I ran the agent for each of those specs in new sessions.

## Proposed Changes

As mentioned previously (see https://github.com/Automattic/studio/pull/4001 and https://github.com/Automattic/studio/pull/4003), `pull-reprint` state revolves around pulls, which is a foreign concept in Studio. In other areas, Studio state revolves heavily around sites. 

This PR completes the `pull-reprint` state refactor by completely removing `pull.json` in favor of a `status` prop on the site object in `cli.json`. This new `status` prop is a simplified version of the previous `PullStage` concept in `pull-reprint`. 

Moreover, this PR adds logic to restart the site after the pull if this was the first pull (which we need, or there will be `open_basedir` errors) and to also only pass the `--force` parameter to Reprint on the first pull.

There's also a bit of general cleanup here now that `pull.json` is gone, and we rely a little less on the staging logic that used to be more central to `pull-reprint`. 

## Testing Instructions

1. Run `npm run cli:build`
2. Run `node apps/cli/dist/cli/main.mjs site create`. Step through the flow and take note of the site path.
3. Run `STUDIO_ENABLE_PULL_REPRINT=1 node apps/cli/dist/cli/main.mjs pull-reprint --path PATH_TO_SITE`
4. Ensure that the site loads successfully after the pull, without needing to restart.

## Pre-merge Checklist


docs/design-docs/cli.md
```diff
@@ -54,6 +54,30 @@ We could almost ship the CLI source code as-is. We know which Node.js version in
 
 Long-term, we might want to move in that direction, but for now, we are still bundling. It offers us some flexibility around which exact code we ship to users (by allowing us to define globals that act as feature flags), and we've seen in testing that bundled code uses less memory, presumably because of code splitting and tree shaking. 
 
+### Pulling a remote site (`pull-reprint`)
```
epeicher, line 57: Thanks for updating the docs! 🙌

apps/cli/commands/pull-reprint.ts
```diff
@@ -293,6 +277,30 @@ export async function runCommand(
...
+	// until this run finishes the site is not a healthy install, so other
+	// commands (`site start`, `status`, `list`) surface it rather than treat
+	// it as normal; persisting `technicalSiteDirectory` now (rather than only
+	// at the later linking step) means `studio delete` can always trash the
+	// scratch, even if the pull dies before linking. A crash here simply
+	// leaves the site `pulling`; a re-run resumes.
+	site.status = 'pulling';
+	site.technicalSiteDirectory = studioMetadata.technicalSiteDirectory;
+	await updateSiteRecord( site.id, ( record ) => {
+		record.status = 'pulling';
```
epeicher: I think we should update the status to `pulling` after `runPreflight` to ignore potential errors while doing the preflight, like a deleted remote site. That way, we don't affect the local site until the actual moving of files start.

apps/cli/commands/pull-reprint.ts
```diff
@@ -501,32 +524,36 @@ export async function runCommand(
...
-		const resumeCommand = [
-			'studio pull-reprint',
-			`--path "${ localPath }"`,
-			`--url ${ normalizedRemoteUrl }`,
-		];
+		// The pull errored or was killed mid-flight: the site directory may
+		// be half-written, so mark it `pull-failed`. It is never trusted as
+		// healthy again until an idempotent re-run restores it to `ready`
+		// (or `site delete` removes it).
+		await setSiteStatus( site, 'pull-failed' );
```
epeicher: In line with the [comment above](https://github.com/Automattic/studio/pull/4007/changes#r3511885584) about ignoring errors on the `runPreflight` for the local site, what do you think about only updating the status if the status is not `ready` so we don't update the status if there is an error on `preflight` and we leave the local site working if the remote has any sort of errors. This is just for the scenario where we have NOT yet started pulling files. Something like this:
```js
    if ( getSiteStatus( site ) !== 'ready' ) {
        await setSiteStatus( site, 'pull-failed' );
    }
```

That change should be applied only if the [suggestion above](https://github.com/Automattic/studio/pull/4007/changes#r3511885584) is applied.

epeicher (approved):
Thanks @fredrikekelund for this great refactor! I have tested it, and it works as expected. I can see that the status of the `pull` is now stored in `cli.json,` and the `pull.json` is not saved anymore. The pull and re-pull work great 👍 

I have left some questions for consideration, but they are not blocking, as they cover a specific use case of errors during the preflight phase. The changes overall LGTM!

epeicher:
I think the [Unit Test failure](https://buildkite.com/automattic/studio/builds/18550/list?jid=019f1ee4-705b-4df0-b69f-fa5a5cdc98a7&tab=output) should be fixed with a merge with trunk as it seems related to this fix https://github.com/Automattic/studio/pull/4025

apps/cli/commands/pull-reprint.ts
```diff
@@ -530,11 +530,14 @@ export async function runCommand(
...
 		process.exit( 0 );
 	} catch ( error ) {
-		// The pull errored or was killed mid-flight: the site directory may
-		// be half-written, so mark it `pull-failed`. It is never trusted as
-		// healthy again until an idempotent re-run restores it to `ready`
-		// (or `site delete` removes it).
-		await setSiteStatus( site, 'pull-failed' );
+		// Mark the site `pull-failed` if the error was thrown after it was
+		// marked `pulling`.
+		if ( site.status === 'pulling' ) {
```
epeicher, line 538: 👍

PR #4011, opened 2026-06-30
Set up Studio as a Claude and Codex marketplace for data-liberation

borkweb:
## Related issues

- Related to #4002 (this PR is stacked on top of that branch)

## How AI was used in this PR

Authored with Claude Code. The marketplace/plugin configs were written and then verified against the actual Claude Code plugin spec and the on-disk `bork`/`vercel` reference plugins to confirm path-resolution behavior (Claude vs. Codex both resolve a plugin `source` relative to the marketplace repo root). The MCP-registration, relative-path, and license corrections came out of that verification pass.

## Proposed Changes

Promotes the Studio repo root to a plugin **marketplace** for both Claude (`.claude-plugin/marketplace.json`) and Codex (`.agents/`), and rehomes the `data-liberation` plugin metadata under the Studio identity. With this, the bundled `packages/data-liberation-agent` workspace installs as a first-class plugin in both Claude Code and Codex — surfacing its skills, commands, and the data-liberation MCP server — without consumers needing to add a separate marketplace.

Both marketplaces point their plugin `source` directly at `./packages/data-liberation-agent`, resolved from the repo root, so no plugin-locating symlink is required. This also corrects configuration that would otherwise have broken installation:

- The Claude plugin now registers its MCP server via the valid `mcpServers` key (pointing at the canonical `./.mcp.json`) instead of an unrecognized `mcp` field.
- The Codex plugin's `skills`/`mcpServers` paths are plugin-root-relative rather than double-prefixed with the package path.
- Both plugin manifests declare `GPL-2.0-or-later` to match the host repo.

No application code changes; this is marketplace/plugin configuration only.

## Testing Instructions

1. **Claude:** From a checkout of this branch, add the local marketplace (`/plugin marketplace add .`) and install `data-liberation`. Confirm the plugin's skills/commands appear and the `data-liberation` MCP server registers and starts.
2. **Codex:** Add the marketplace from `.agents/` and install `data-liberation`; confirm the plugin resolves to `packages/data-liberation-agent` and its skills + MCP server load.
3. Validate the JSON manifests parse (all four marketplace/plugin files are valid JSON).


nightnei (changes requested):
@borkweb I tested and everything is good except one thing -  Studio is an npm workspace now, so dependencies are hoisted to the Studio repo root node_modules — they never make it into the copy from `packages` folder to `~/.claude/plugins/cache/studio/data-liberation`.
AI recommendation is - `bundle the MCP server into a self-contained file (esbuild) and point .mcp.json at node dist/..., or ship a package-lock.json inside packages/data-liberation-agent so the plugin installer restores deps in the cache.` I think that bundling is the right direction.

borkweb:
@nightnei re:

> AI recommendation is - bundle the MCP server into a self-contained file (esbuild) and point .mcp.json at node dist/..., or ship a package-lock.json inside packages/data-liberation-agent so the plugin installer restores deps in the cache. I think that bundling is the right direction.

Yeah, I agree with you! That's the approach I took with the blocks-engine for WP dependencies and it seems to work well. I'll make that tweak

borkweb:
@nightnei - Alrighty, the MCP dist bundle is implemented. The unfortunate state of Claude Code and Codex plugins is that we cannot keep the dist bundle as part of a CI process for building the dist. Technically, there is a way, but the experience for people adding the plugins to their own harness would be poor. Committing the dist allows someone to:

```bash
claude plugin marketplace add Automattic/studio
```

There's a couple of notes on the dev experience for us:

1. Buildkite has a check that ensures that the data-liberation-agent dist is not stale
2. We have a `scripts/mcp-launcher.mjs` that ensures that a dev checkout of the repository will use the `src/` directory via `tsx` rather than deferring to the dist.

packages/data-liberation-agent/scripts/build-mcp-bundle.mjs
```diff
@@ -0,0 +1,81 @@
+// Bundle the MCP server into a single self-contained dist/mcp-server.bundle.mjs.
+//
+// Why: the Claude/Codex plugin installer copies this package's files into
+// ~/.claude/plugins/cache/... WITHOUT node_modules — and inside the Studio npm
+// workspace the dependencies are hoisted to the repo root anyway — so a
+// `npx tsx src/mcp-server.ts` entry has nothing to import once installed as a
```
nightnei, line 6: Hm, the same problem we shoudl expect in prod for `npx tsx scripts/...` in [skills/replicate-theme/SKILL.md](https://github.com/Automattic/studio/blob/trunk/packages/data-liberation-agent/skills/replicate-theme/SKILL.md). It seems it's necessary to migarte runnign npx to MCP or some another way

nightnei (changes requested):
> @nightnei - Alrighty, the MCP dist bundle is implemented. The unfortunate state of Claude Code and Codex plugins is that we cannot keep the dist bundle as part of a CI process for building the dist. Technically, there is a way, but the experience for people adding the plugins to their own harness would be poor. Committing the dist allows someone to:
> 
> ```shell
> claude plugin marketplace add Automattic/studio
> ```
> 
> There's a couple of notes on the dev experience for us:
> 
> 1. Buildkite has a check that ensures that the data-liberation-agent dist is not stale
> 2. We have a `scripts/mcp-launcher.mjs` that ensures that a dev checkout of the repository will use the `src/` directory via `tsx` rather than deferring to the dist.

I see your points, let's start with commiting it, it isn't huge and isolated inside packages, so I don't see problems 👍

1. ✅ Claude Code - everythign went great, also I tried Theme replication and teh site looks great.
2. ⛔️ Codex - the mcp isn't starting.  Here is the explanation from AI, so we shoudl fix it in this PR:
```
Root cause: Codex doesn't expand ${CLAUDE_PLUGIN_ROOT}. The shared .mcp.json uses it, and Codex passes it through literally — codex mcp get data-liberation shows:

args: ${CLAUDE_PLUGIN_ROOT}/scripts/mcp-launcher.mjs
cwd:  ~/.codex/plugins/cache/studio/data-liberation/1.0.0/${CLAUDE_PLUGIN_ROOT}   ← doesn't exist

Codex's own bundled plugins (codex-security, openai-developers) use plugin-root-relative paths instead: "args": ["./mcp/server.mjs"], "cwd": ".". The launcher itself is fine — invoked relatively from the Codex plugin cache, it completes the MCP initialize handshake immediately.

Suggested fix: Claude needs ${CLAUDE_PLUGIN_ROOT} (no cwd guarantee), Codex needs relative paths — and the plugin manifests already diverge per harness, so point .codex-plugin/plugin.json's "mcpServers" at a Codex-specific file (e.g. ./.mcp.codex.json):

{
  "mcpServers": {
    "data-liberation": {
      "command": "node",
      "args": ["./scripts/mcp-launcher.mjs"],
      "cwd": "."
    }
  }
}

…leaving the existing .mcp.json for Claude untouched.
```
3. I think this PR can be merged, as soon as we address Codex. But as a follow up thing necessary to take a look at [this](https://github.com/Automattic/studio/pull/4011#discussion_r3536442279) and most likelly address in the follow-up PR. Could you please take a look at it?

nightnei:
@borkweb I went ahead and pushed the fix to your PR to move things forward. Hope that’s okay :) 
I tested both Claude and Codex - workspace added successfully, plugin installs w/o issues too, mcp connects, and I see skills, `/liberate` / `$liberate` are available, so I approve the PR. 
Take a look, please, at my commit, and if it looks good to you, then feel free to merge :) 

The last issue that should be investigated/fixed outside this PR is the third point from [here](https://github.com/Automattic/studio/pull/4011#pullrequestreview-4645133679).

nightnei (approved):

borkweb:
Looks great! Merging

borkweb:
@nightnei - here's the fix for item 3: https://github.com/Automattic/studio/pull/4126

PR #4021, opened 2026-07-01
Remove `tsx` in favor of `node --experimental-strip-types`

fredrikekelund:
## Related issues

- Fixes STU-1376

## How AI was used in this PR

Codex was used to execute the implementation from start to finish. I drove the process, iterating over several rounds, and identified a few key fixes.

## Proposed Changes

Our build scripts in the `./scripts` directory are written in Typescript. This is obviously nice and desirable, but it has the downside of requiring a compilation step to run on Node.js. Except that's not actually true, because Node.js has an `--experimental-strip-types` option ([ref](https://nodejs.org/learn/typescript/run-natively)).

This PR drops the [tsx](https://www.npmjs.com/package/tsx) dependency we were using to execute those scripts in favor of `node --experimental-strip-types`.

Unfortunately, this PR can no longer remove the `tsx` dependency from the repo after https://github.com/Automattic/studio/pull/4002 landed, since `packages/data-liberation-agent` now specifies `tsx` as a dependency.

## Testing Instructions

CI should pass

## Pre-merge Checklist


apps/studio/electron.vite.config.ts
```diff
@@ -159,7 +159,7 @@ export default defineConfig( {
 			// Force CSS extraction instead of inlining
 			cssCodeSplit: true,
 			// Enable minification and compression
-			minify: 'esbuild',
+			minify: 'oxc',
```
fredrikekelund, line 162: This was needed because esbuild was only indirectly part of our dependency tree, via tsx. The Vite 8 default is oxc, so this is what we want anyway.

wojtekn (approved):
The change looks good. I updated the branch based on trunk, and triggered Mac builds.

PR #4022, opened 2026-07-01
Teach Studio Code to recommend Automattic plugins for more features

katinthehatsite:
## Related issues

- Related to: Studio Code should suggest Automattic's own products/plugins (Newsletters, Forms, courses, polls, etc.) instead of re-inventing the wheel.

## How AI was used in this PR

The skill/instruction copy was drafted with AI. Every factual claim — plugin slugs, block names, custom post types, and Jetpack module behavior — was verified against real local Studio sites (created via the CLI, installed each plugin, enumerated registered blocks/modules/CPTs, then deleted the probe sites). Reviewers should sanity-check the wording and the Jetpack Newsletter caveat.

## Proposed Changes

Studio Code already routes feature requests to a `plugin-recommendations` skill, but it only covered WooCommerce and Jetpack Forms — so for newsletters, courses, and polls the agent had no named Automattic product to reach for and could fall back to hand-building markup.

This broadens that guidance so the agent reuses Automattic's own products first:

- A feature → Automattic-plugin map (WooCommerce, Jetpack Forms, Jetpack Newsletter, Sensei LMS, Crowdsignal, Jetpack Social, Akismet) as a single source of truth.
- Per-product guidance for the previously-missing gaps: **Jetpack Newsletter**, **Sensei LMS** (courses/lessons/quizzes), **Crowdsignal** (polls/surveys/ratings), and **Akismet** (spam).
- The system prompt's skill-routing triggers now include newsletters, online courses/LMS/quizzes, polls/surveys, and social auto-posting, so these requests actually load the skill.

User impact: when someone asks Studio Code for a newsletter signup, a course, a poll, etc., it recommends and installs the matching Automattic plugin and builds with its editable blocks, rather than generating a bespoke HTML/JS equivalent.

One deliberate honesty note in the skill: Jetpack Newsletter's `subscriptions` module cannot activate on a purely local site (it depends on WordPress.com to manage subscribers/send email — confirmed it fails even in Jetpack offline mode). The skill still recommends Jetpack Newsletter as the right tool and explicitly tells the agent not to substitute a hand-built form when the module is unavailable locally.

## Testing Instructions

- `npm run cli:build`
- `npm test -- apps/cli/ai/tests/` — all AI tests pass (skill routing + "references only bundled skills" covers the new copy).
- `npm run typecheck` — clean.
- Optional manual check: `studio code` → ask for a contact form (installs Jetpack Forms), a course site (installs Sensei LMS), or a poll (installs Crowdsignal) and confirm it reaches for the plugin rather than hand-building.

## Pre-merge Checklist


katinthehatsite:
Superseded by #4023, which moves the same changes onto `fix/improve-plugin-recommendations` (rebased on current trunk).

katinthehatsite:
I am just experimenting with conductor, ignore this. It did this without my permissions 🙈

PR #4025, opened 2026-07-01
Fix flaky WordPress server exit tests on Windows

epeicher:
## Related issues

- Related to the recurring Windows unit-tests CI failures (build 18486)

## How AI was used in this PR

Claude diagnosed the failing Windows CI log, traced the timeout to a listener-attach race, implemented the fix, and empirically proved both the root cause and the fix by injecting a simulated slow-subscribe delay (old pattern reproduces the exact `Test timed out in 5000ms`; new pattern passes).

## Proposed Changes

The Windows unit-tests job intermittently failed with `Test timed out in 5000ms` on `wordpress-server-manager.test.ts` — most recently on *"should append PHP errors captured during the start attempt to the failure"*, though several sibling tests share the same latent flaw.

These tests simulate a server crash by emitting a one-shot `exit` event on a fixed 10 ms timer. But `startWordPressServer` only subscribes to the daemon bus *after* awaiting some filesystem setup (clearing and statting the PHP error log). On a slow, heavily loaded Windows agent that setup can outlast the 10 ms timer, so the one-shot event fires before the listener exists and is dropped — there's no buffering for pre-subscription events. The call then hangs on the 2-minute inactivity timeout, so the 5 s test timeout trips first. That's why only one test hung (~5 s) while its near-identical siblings passed in ~15 ms.

This makes the exit emission deterministic: it fires once the server has actually attached its `process-event` listener (via the `EventEmitter` `newListener` signal), instead of guessing with a fixed delay. This mirrors the existing `setupIpcMocks` workaround already used for the `ready` message, and — because subscription happens after the baseline log size is captured — keeps the PHP-error assertions correct. No production code changes; test-only.

## Testing Instructions

- `npm test -- apps/cli/lib/tests/wordpress-server-manager.test.ts` passes (22 tests), repeatably.
- To reproduce the original failure and confirm the fix: temporarily add `await new Promise( ( r ) => setTimeout( r, 60 ) );` right before `subscribeForReadyOrExit` in `apps/cli/lib/wordpress-server-manager.ts`. With the old one-shot `setTimeout(…, 10)` pattern the exit tests time out at 5000 ms; with this change they pass. Remove the injected delay afterward.

## Pre-merge Checklist


bcotrim (approved):
LGTM 👍

PR #4026, opened 2026-07-01
Add weekly GitHub action cron job to check for new PHP versions

fredrikekelund:
## Related issues

- Fixes RSM-4591

## How AI was used in this PR

Codex executed the implementation with iterative driving instructions from me. 

## Proposed Changes

The `Build PHP CLI Binaries` GitHub Actions workflow gives us a really smooth way of building new PHP binaries for all platforms. Input a PHP version and set `apps_cdn_visibility` to `external`, and you get a PR back that modifies the `php-binary-cdn-metadata.mjs` file. The missing piece is automation that looks for new PHP versions and automatically triggers those builds. This PR adds it as a weekly GitHub Actions cron job. 

For added safety, I've changed the `error_on_duplicate` parameter in the `Build PHP CLI Binaries` job to true. Overwriting existing appscdn packages would lead to errors in some local Studio CLI installations. We should only do that if we really know what we're doing. 

## Testing Instructions

1. Generate a GitHub access token: https://github.com/settings/personal-access-tokens
2. Run `GH_TOKEN=xxxxxx node scripts/check-php-cli-versions.mjs` (this will only generate output and not trigger a CI workflow run)
3. Ensure that the script runs successfully (right now, you'll see that there are new PHP versions available)

## Pre-merge Checklist


bcotrim (commented):
Great improvement, some suggestions:
1. **Skip dispatch if the version’s already on the CDN**. If the bump PR sits unmerged, the next cron re-builds all 4 platforms and only fails at upload so the whole build is wasted.
2. **Alert on failure.** Let's add a Slack ping if this fails, we already do it on other CI checks, so it should be easy to add. Could be a follow-up if it's not as straightforward as I am assuming.
3. **Expose error_on_duplicate as a dispatch input (default true)**. Just to make it easier for a manual run to allow an overwrite. Let's say we notice an issue in a nightly build, we could manually trigger an overwrite without any side effects to our users.

fredrikekelund:
> Skip dispatch if the version’s already on the CDN

Good idea 👍 Updated.

> Alert on failure

Ditto 👍

> Expose error_on_duplicate as a dispatch input (default true)

I disagree here. I think we should be super careful about updating existing packages on appscdn, because local Studio installs expect a certain checksum for those packages. If the checksum doesn't match, those Studio installs will stop being able to install those PHP versions.

You and I discussed this a while back and concluded that we should probably get creative about the version numbers we use for PHP packages to mitigate this (i.e., make it so that we are able to rebuild PHP versions that are already on the CDN without breaking existing Studio installs).

bcotrim:
> You and I discussed this a while back and concluded that we should probably get creative about the version numbers we use for PHP packages to mitigate this (i.e., make it so that we are able to rebuild PHP versions that are already on the CDN without breaking existing Studio installs).

That's fair, I was being optimistic about the runner but you make a very good point, a mistake on an updated version would result in a breaking state for our users. 👍

bcotrim (approved):

PR #4027, opened 2026-07-01
Add refresh_browser tool so the agent can reload the site preview

bcotrim:
## Related issues

- Fixes STU-1904

## How AI was used in this PR

Built with Claude Code: explored the agent tool + site-preview architecture, implemented the tool and event wiring, and wrote the unit tests. I reviewed every change myself and validated the behavior manually in the app. No CSS/color changes, so there is nothing to review in dark mode.

## Proposed Changes

The Studio agent had no way to refresh the in-app site preview after making changes, so it fell back to stopping and starting the whole site just to force a fresh webview mount — slow, disruptive, and impossible when the site was already running (the preview would simply stay stale). This adds a `refresh_browser` tool that reloads the preview **in place** via a dedicated `preview.reload` event, and steers the agent through its system prompt to use it after user-visible changes instead of restarting the site. Users now see their agent-made changes reflected immediately, without a server restart.

## Testing Instructions

1. Start Studio, open the agentic UI, and select a **running** local site with the preview panel open.
2. Ask the agent to change something visible without navigating, e.g. `Run wp-cli option update blogname "Reloaded-1" on this site`. The preview still shows the old title.
3. Ask the agent to `refresh the browser preview`. It should call `refresh_browser` (shown as "Refresh preview") and the header should update in place — with **no** site stop/start.
4. Automated: `npm test -- apps/cli/ai/tests/tools.test.ts` and `npm test -- apps/ui/src/hooks/use-session-commands.test.tsx`.

## Pre-merge Checklist

### DIFF
--- a/apps/cli/ai/tools/refresh-browser.ts
+++ b/apps/cli/ai/tools/refresh-browser.ts
@@ -1,0 +1,13 @@
+import { emitEvent } from 'cli/ai/json-events';
+import { defineTool } from './define-tool';
+import { textResult } from './utils';
+
+export const refreshBrowserTool = defineTool(
+	'refresh_browser',
+	'Reloads the site preview browser in the Studio app so the user sees your latest changes. Call this after a change that alters what the site renders (content, options/settings, theme, plugins, activation). It reloads in place — never stop/start the site (site_stop/site_start) just to refresh the preview.',
+	{},
+	async () => {
+		emitEvent( { type: 'preview.reload', timestamp: new Date().toISOString() } );
+		return textResult( 'Reloaded the site preview.' );
+	}
+);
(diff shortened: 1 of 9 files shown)


apps/cli/ai/tools/refresh-browser.ts
```diff
@@ -0,0 +1,13 @@
+import { emitEvent } from 'cli/ai/json-events';
+import { defineTool } from './define-tool';
+import { textResult } from './utils';
+
+export const refreshBrowserTool = defineTool(
+	'refresh_browser',
+	'Reloads the site preview browser in the Studio app so the user sees your latest changes. Call this after a change that alters what the site renders (content, options/settings, theme, plugins, activation). It reloads in place — never stop/start the site (site_stop/site_start) just to refresh the preview.',
```
katinthehatsite, line 7: Nice, this is a much better experience than having to reload things manually which I was doing before

katinthehatsite (approved):
I think this looks good 👍 I asked to make some visual changes like adjusting colors and I noticed that the preview refreshed by itself once the agent was done without me having to do it manully. I think this is a much better experience.

PR #4009, opened 2026-06-30
Adjust data-liberation-agent as an npm workspace

nightnei:
## Related issues

Fixes STU-1967

## How AI was used in this PR

AI helped to investigate and apply changes

## Proposed Changes
It's the second step after https://github.com/Automattic/studio/pull/4002, to add data-liberation-agent into the Studio repo - clean up unnecessary things.

## Testing Instructions
* Assert that CI is green
* `npm i`
* `npm run cli:build`
* `npm start`
* Assert that Studio still works well
* `node apps/cli/dist/cli/main.mjs code`
* Assert that Studio Code still works well

PR #4028, opened 2026-07-01
Restore blocks-engine registry dependency (drop temporary vendor)

nightnei:
## How AI was used in this PR
AI helped to revert.

## Proposed Changes
Follow up to https://github.com/Automattic/studio/pull/4009
Specifically, revert tyhis commit - https://github.com/Automattic/studio/pull/4009/changes/8753283865299463ae7458071e24903f09c14747

`@automattic/blocks-engine@0.2.2` was published less than two days ago, so studio's [min-release-age=2](https://github.com/Automattic/studio/blob/trunk/.npmrc) supply-chain guard rejected it in CI.<br />
Screenshot 2026-07-01 at 13 57 01<br />
As a temporary solution to unblock the PR I vendored the package as a committed tarball referenced via `file:`.
Tomorrow, 0.2.2 age will past the two-day window, so we can revert that temporary solution with this PR.

## Testing Instructions
Visual review should suffice


katinthehatsite (approved):
Everything is passing so approving 👍

PR #4035, opened 2026-07-01
Disable Jetpack Stats for local sites

shaunandrews:
## Related issues

- Related to https://linear.app/a8c/issue/STU-1945/investigate-localhost-traffic-inflating-wordpresscomjetpack-stats

## How AI was used in this PR

Codex investigated the localhost Stats repro, confirmed that local Studio-rendered pages can emit Jetpack Stats pixels with live blog IDs, and prepared this scoped fix. I reviewed the generated change and kept the PR limited to the Studio MU plugin behavior plus a regression test.

## Proposed Changes

- Prevent Studio-served local sites from reporting page views to the remote WordPress.com/Jetpack Stats site when an imported or synced local database still contains a live Jetpack connection.
- Preserve the existing local-development behavior that disables Jetpack Protect for auto-login, while also disabling Jetpack Stats through the same module filter path.
- Keep the fix server-side in the local WordPress runtime so it applies to Studio previews, thumbnail captures, and local sites opened in an external browser.

## Testing Instructions

- `npx eslint --fix tools/common/lib/mu-plugins.ts tools/common/lib/tests/mu-plugins.test.ts`
- `npm test -- tools/common/lib/tests/mu-plugins.test.ts`
- `npm run typecheck`

## Pre-merge Checklist


sejas:
@shaunandrews I thought about implementing this same solution when I found you (and your agents) already created the PR. 💫

sejas (commented):
The code looks good to me, but I'm still seeing the pixel and the stats module active. I'll investigate more next week. Happy to take over this PR.

Image

Image

shaunandrews:
Please do!

sejas (approved):
I confirmed this is the right fix and works for new sites. It doesn't work for old sites because there is a bug where the mu-plugins don't get new changes. I'll create a separate PR to fix that.

Thanks Shaun!

sejas (approved):
Image

PR #4037, opened 2026-07-01
[Tooling] Secure Claude workflows

iangmaia:
## Summary
- Allow `@claude` review comments only from trusted commenters and skip Claude workflows on fork/external PRs before checkout or Claude execution.
- Pin mutable action references and preserve the Claude action's required OIDC permission for GitHub App token exchange.
- Keep review feedback inline-only where the workflow does not grant issue-comment permissions.


mokagio (approved):
Ship it.

bcotrim:
LGTM, thanks for the improvements!

PR #4038, opened 2026-07-01
Skip dangling symlinks when archiving site content

epeicher:
## Related issues

- Related to: reprint-pulled (WP Cloud) sites shipping a dangling `advanced-cache.php` drop-in symlink.

## How AI was used in this PR

Claude Code reproduced the failure, traced it to `fs.realpathSync` on a dangling symlink aborting the archive, wrote the guard and the unit test, and verified the fix end-to-end (a real preview-site creation against an affected site). I reviewed the diff.

## Proposed Changes

Creating a preview site — and pushing — from a reprint-pulled WP Cloud site failed with:

`Failed to create preview site: ENOENT: no such file or directory, stat '.../wp-content/advanced-cache.php'`

Those sites keep the WP Cloud `advanced-cache.php` cache drop-in as a symlink pointing into `wordpress/drop-ins/`, which the pull doesn't download — so it dangles. When archiving `wp-content`, `fs.realpathSync` on that broken link threw `ENOENT` and aborted the whole archive.

Now a dangling symlink is skipped (with a warning) instead of failing the whole operation, so preview create/update and push succeed. This makes the archiver robust to any broken symlink — which matters for sites already pulled, since they keep the dangling drop-in until re-pulled.

The reprint side — not creating the dangling symlink in the first place — is fixed separately in `WordPress/reprint`.

## Testing Instructions

1. On a reprint-pulled WP Cloud site (its `wp-content/advanced-cache.php` is a broken symlink), rebuild the CLI (`npm run cli:build`) and run:
   `node apps/cli/dist/cli/main.mjs preview create --path ~/Studio/<site>`
2. Before: fails with `ENOENT ... advanced-cache.php`. After: the preview site is created successfully.
3. Unit tests: `npm test -- apps/cli/lib/tests/archive.test.ts` (includes a dangling-symlink case).

## Pre-merge Checklist

- [x] Have you checked for TypeScript, React or other console errors? (`tsc -p apps/cli` clean, ESLint clean)


apps/cli/lib/archive.ts
```diff
@@ -57,7 +57,15 @@ export async function archiveSiteContent(
...
+			try {
+				resolvedPath = fs.realpathSync( path.join( wpContentPath, relativePath ) );
+			} catch ( error ) {
+				// Dangling symlink. Skip it rather than aborting the whole archive.
+				console.warn( `Skipping ${ archiveEntryPath }: ${ error }` );
+				continue;
+			}
+			archiveBuilder.file( resolvedPath, {
 				name: archiveEntryPath,
 			} );
```
fredrikekelund:
```suggestion
			try {
				const resolvedPath = fs.realpathSync( path.join( wpContentPath, relativePath ) );
				archiveBuilder.file( resolvedPath, { name: archiveEntryPath } );
			} catch ( error ) {
				// Dangling symlink. Skip it rather than aborting the whole archive.
				console.warn( `Skipping ${ archiveEntryPath }: ${ error }` );
			}
```
Optional nit, but it doesn't seem necessary to call `fs.realpathSync()` in a different scope to `archiveBuilder.file()`

apps/cli/lib/import-export/export/exporters/default-exporter.ts
```diff
@@ -271,7 +271,15 @@ export class DefaultExporter extends ImportExportEventEmitter implements Exporte
...
+			try {
+				resolvedPath = fs.realpathSync( fullEntryPathOnDisk );
+			} catch ( error ) {
+				// Dangling symlink. Skip it rather than aborting the whole archive.
+				console.warn( `Skipping ${ entryPathRelativeToArchiveRoot }: ${ error }` );
+				continue;
+			}
+			this.archiveBuilder.file( resolvedPath, {
 				name: entryPathRelativeToArchiveRoot,
 			} );
```
fredrikekelund:
```suggestion
			try {
				const resolvedPath = fs.realpathSync( path.join( wpContentPath, relativePath ) );
				archiveBuilder.file( resolvedPath, { name: archiveEntryPath } );
			} catch ( error ) {
				// Dangling symlink. Skip it rather than aborting the whole archive.
				console.warn( `Skipping ${ archiveEntryPath }: ${ error }` );
			}
```
Optional nit. Same thing here

fredrikekelund (commented):
Good fix regardless of whether the site is Reprint-pulled or not 👍

One thing worth looking into is the `calculateDirectorySizeForArchive()` function. I'm getting a warning like this from there:

```
Error processing /Users/fredrik/Studio/my-true-website/wp-content/advanced-cache.php: Error: ENOENT: no such file or directory, stat '/Users/fredrik/Studio/my-true-website/wp-content/advanced-cache.php'
    at async Object.stat (node:internal/fs/promises:1038:18)
    at async file:///Users/fredrik/Code/studio/apps/cli/dist/cli/rewrite-wp-cli-post-content-wf2YZvuH.mjs:44:22
    at async Promise.all (index 0)
    at async calculateSize (file:///Users/fredrik/Code/studio/apps/cli/dist/cli/rewrite-wp-cli-post-content-wf2YZvuH.mjs:36:5) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/Users/fredrik/Studio/my-true-website/wp-content/advanced-cache.php'
```

This looks like the exact same underlying problem, so it seems sensible to address in this PR

epeicher:
> Good fix regardless of whether the site is Reprint-pulled or not 👍

I agree, the user could create dangling symlinks and we don't want to break some flows.

> One thing worth looking into is the `calculateDirectorySizeForArchive()` function. I'm getting a warning like this from there:
> This looks like the exact same underlying problem, so it seems sensible to address in this PR

Yes, that's the same error, let me look at it

epeicher:
Thanks for the review @fredrikekelund, I have addressed the other functions that were causing noisy warnings. I have turned down the warning so it's still logged that the symlink is skipped, but in a controlled way. Please let me know your view on that, the commit is [4ab9f7a](https://github.com/Automattic/studio/pull/4038/commits/4ab9f7a3274f59805fcec7c61aa1ed2cbc66e5ed)

apps/studio/src/ipc-handlers.ts
```diff
@@ -1815,7 +1815,15 @@ export function getFileSize( _event: IpcMainInvokeEvent, siteId: string, filePat
...
-	return fs.statSync( nodePath.join( site.details.path, ...filePath ) ).size;
+	const fullPath = nodePath.join( site.details.path, ...filePath );
+	try {
+		return fs.statSync( fullPath ).size;
+	} catch ( error ) {
+		// Dangling symlink or unreadable entry. It's skipped when archiving,
+		// so count it as zero rather than failing the size check.
+		console.warn( `Skipping ${ fullPath }: ${ error }` );
+		return 0;
+	}
```
fredrikekelund, line 1826: Nice addition 👍

fredrikekelund (approved):
https://github.com/WordPress/reprint/pull/296 already clears the `advanced-cache.php` symlink specifically, but I tested this with a manually created dangling symlink. Works great 👍

PR #4042, opened 2026-07-02
Include Visual C++ DLLs with Windows PHP builds and add `packageVersion` field to identify PHP packages

fredrikekelund:
## Related issues

- Related to STU-1972

## How AI was used in this PR

Codex drove the implementation based on previous diagnosis.

## Proposed Changes

Sentry shows that some users are running into `PHP command failed (code: 3221225781)` errors. This is because of missing DLLs on their system – very likely the Visual C++ Redistributable ones, as PHP explicitly [names that dependency](https://www.php.net/downloads.php?os=windows#instructions). This PR updates the native PHP build workflow to include the necessary DLLs next to `php.exe`. 

Because there aren't new PHP patch versions available yet for every minor PHP version we offer, I also had to find a way to rebuild PHP versions that already exist on appscdn. This cannot be done today without breaking local Studio installations, because they expect the appscdn package to have a predefined checksum. If the checksum of the appscdn package changes, the Studio client will refuse to install it.

The solution is to add a `packageVersion` field that identifies the package version separately from the PHP version. The workflow sets this to `studio-1` by default – meaning a typical workflow run (that builds a new PHP version) can just use the default. Moreover, I've set `error_on_duplicate` to always be true (same as in https://github.com/Automattic/studio/pull/4026, which I mistakenly merged to another branch than trunk)

## Testing Instructions

I'll manually verify that these builds do what we want on Windows.

## Pre-merge Checklist

### DIFF
--- a/.github/workflows/build-php-cli-binaries.yml
+++ b/.github/workflows/build-php-cli-binaries.yml
@@ -75,7 +75,10 @@ jobs:
             artifact_arch: aarch64
             extensions: *unix_extensions
           - target: windows-x86_64
-            runner: windows-latest
+            # PHP 8.4+ Windows builds use the VS17 toolchain. Pin the runner so
+            # the matching VC143 app-local runtime remains available even as
+            # windows-latest moves to newer Visual Studio releases.
+            runner: windows-2022
             artifact_platform: windows
             artifact_arch: x86_64
             extensions: apcu,bcmath,calendar,ctype,curl,dba,dom,exif,fileinfo,filter,ftp,gd,iconv,igbinary,intl,mbregex,mbstring,mysqli,mysqlnd,opcache,openssl,pdo,pdo_mysql,pdo_sqlite,phar,redis,session,shmop,simplexml,sockets,sodium,sqlite3,ssh2,tokenizer,xdebug,xml,xmlreader,xmlwriter,yaml,zip,zlib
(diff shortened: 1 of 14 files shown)


.github/workflows/build-php-cli-binaries.yml
```diff
@@ -459,15 +477,50 @@ jobs:
...
+          $requiredVcRuntimeDlls = @('vcruntime140.dll', 'vcruntime140_1.dll')
+          foreach ( $dll in $requiredVcRuntimeDlls ) {
+            if ( -not (Test-Path (Join-Path $vcRuntimeDir $dll)) ) {
+              throw "Required Visual C++ runtime DLL not found: $dll"
+            }
+          }
+
+          Copy-Item -Path (Join-Path $vcRuntimeDir '*.dll') -Destination $stagingDir
+          $vcRuntimeVersion = (Get-Item (Join-Path $vcRuntimeDir 'vcruntime140.dll')).VersionInfo.FileVersion
+          Set-Content -Path (Join-Path $stagingDir 'vc-runtime.version') -Value $vcRuntimeVersion -NoNewline
```
fredrikekelund, line 510: Copying the DLLs from the CI Windows system was the method recommended by Codex for putting them next to the `php.exe`. AFAICT, Microsoft only offers an installation wizard for the Visual C++ Redistributables – not static DLLs that we can download and put where we want them.

fredrikekelund:
Confirmed that https://github.com/Automattic/studio/pull/4045 fixes the issue on a Windows VM that has no system-wide Visual C++ Redistributable DLLs 👍

bcotrim (approved):
Code and fix LGTM 👍
The true test will happen when we distribute the new build, let me know if you want an extra validation then.

Suggestion:

- Let's add a migration to remove superseded PHP package dirs (older patch/package revisions of each minor, and minors we no longer support). Keeping them in `~/.studio/php-bin/` just takes space for no purpose. It's ok to tackle as a follow-up.

fredrikekelund:
> add a migration to remove superseded PHP package dirs

Agreed that this is a good idea, although I'd prefer to do it in a follow-up. Not critical yet, either, as this is only the second version of each PHP package that we ship to users.

PR #4046, opened 2026-07-02
Add WordPress export (WXR/.xml) import support

wojtekn:
## Related issues

- Fixes [STU-1497](https://linear.app/a8c/issue/STU-1497)

## How AI was used in this PR

Authored with Claude Code: codebase exploration, the importer/handler/validator implementation, the PHP driver, tests, and this description. All changes were reviewed by the author, and the flows were manually tested in the app (see Testing Instructions).

## Proposed Changes

Studio's Import/Export screen previously accepted full-site backups (Jetpack, Local, Playground, `.wpress`) and raw `.sql` databases, but not a WordPress export file — the `.xml` (WXR) produced by **Tools → Export**. That importer is a first-class option in the WordPress dashboard, and users reasonably expect it in Studio too.

This adds `.xml` as a supported import format in both the **Import/Export** tab (existing site) and the **Add-Site → Import from a backup** flow (new site). The content — posts, pages, terms, authors, and media — is imported the same way the dashboard does it: via the official `wordpress-importer` plugin. The plugin is installed into the target site before the import runs, so the import works offline with no wordpress.org fetch at runtime.

WXR import differs from the other formats in that it *merges content into an existing WordPress install* rather than replacing files/database. A few things the reviewer should know about:

- **Add-Site ordering:** for `.xml` specifically, the new site is now created and started once (installing WordPress) *before* the import runs, so `wp-config.php` and the database exist when the importer needs them. Other backup types keep their existing skip-start behavior.
- **Media caveat (intentional):** attachments are downloaded from the URLs baked into the WXR, exactly like the dashboard importer. Exports from a publicly reachable site import media correctly. Exports from *another local Studio site* do not — their `.wp.local` URLs resolve only via the OS hosts file + Studio's proxy and serve a self-signed cert, none of which the import runtime has, so posts import but images keep pointing at the source site. This is documented in the PHP driver and left as a possible follow-up.
- **Plugin bundling:** the `wordpress-importer` plugin is not vendored in the repo. It's downloaded at install time (pinned version) via the existing `FILES_TO_DOWNLOAD` registry in `scripts/download-wp-server-files.ts`, alongside WP core, SQLite, WP-CLI, etc., and ships in the CLI bundle under `wp-files/`. The runtime install into the site is a local copy — no network needed at import time. **CI must run `postinstall` (or the download step) before packaging** so the plugin is present in the bundle; the sibling downloads already rely on this.

## Testing Instructions

1. In any WordPress site, go to **Tools → Export → All content** to produce a `.xml` (WXR) file. Prefer a public source site if you want to verify media import.
2. **Import/Export tab:** open an existing Studio site → **Import/Export** → drop the `.xml`. Confirm the progress bar completes, then open the site and verify posts/pages imported. With a public source, media appears in the Media Library.
3. **Add-Site flow:** **Add site → Import from a backup** → select the `.xml`. Confirm the new site is created and the content imported.
4. **Offline:** disable networking and confirm the import still runs (the `wordpress-importer` plugin installs from the bundled copy). Media that must be fetched from origin will be missing offline, which is expected.
5. Automated: `npm run cli:build && npm test -- apps/cli/lib/import-export` (adds validator, backup-handler-factory, and importer-selection tests).

## Pre-merge Checklist


apps/cli/php/import-wxr.php
```diff
@@ -0,0 +1,119 @@
...
+	}
+}
+if ( ! class_exists( 'WP_Import' ) ) {
+	WP_CLI::error( 'WP_Import class still not defined after loading the plugin. Did you remember to pass --skip-plugins=wordpress-importer on the wp-cli invocation?' );
+}
+
+// Minimum boilerplate to drive wordpress-importer headlessly — mirrors what the
+// plugin's own CLI path does.
+kses_remove_filters();
+$admins = get_users( array( 'role' => 'Administrator' ) );
```
sejas: Nit, let's use the lowercase slug
```suggestion
$admins = get_users( array( 'role' => 'administrator' ) );
```

gcsecsey (approved):
I tested importing multiple sites with publicly accessible media, and the import consistently worked well. Importing in Offline mode also works as described.

| Online | Offline |
|--------|--------|
| Image | Image | 

Nit: maybe we could add `.xml` to the list of extensions here?

Image

sejas (approved):
I tested it in the classic UI, and it worked well. I was able to import a WordPress xml export in an existing site or in a new site.

Image

We could consider adding XML or WXR in the new site creation box.

Image

sejas (commented):
@wojtekn , we can also modify Studio Code tool descriptions.

https://github.com/Automattic/studio/blob/a88eb4955325e0503aedd7b541581f411dea7208/apps/cli/ai/tools/import-site.ts#L8

**`apps/cli/ai/tools/import-site.ts`**
```diff
-	'Imports a backup file into a local WordPress site. Supports .zip, .tar.gz, .sql, and .wpress formats. ' +
+	'Imports a backup file into a local WordPress site. Supports .zip, .tar.gz, .sql, .wpress, and WordPress export (WXR) .xml formats. ' +
```

https://github.com/Automattic/studio/blob/a88eb4955325e0503aedd7b541581f411dea7208/apps/cli/ai/system-prompt.ts#L200

**`apps/cli/ai/system-prompt.ts`**
```diff
-- site_import: Import a backup file (.zip, .tar.gz, .sql, .wpress) into a local site.
+- site_import: Import a backup file (.zip, .tar.gz, .sql, .wpress, .xml WordPress export) into a local site.
```

**`apps/cli/README.md`** the supported-formats paragraph should mention `.xml` as well.

https://github.com/Automattic/studio/blob/a88eb4955325e0503aedd7b541581f411dea7208/apps/cli/README.md?plain=1#L109

wojtekn:
Thanks @gcsecsey and @sejas - I applied proposed updates.

PR #4062, opened 2026-07-02
Fix broken Central Kurdish text on macOS by bundling Vazirmatn font

wojtekn:
## Related issues

- Fixes STU-1962

## How AI was used in this PR

Used Claude Code to trace the font stack, confirm the failure mode, select and verify the font, and wire up the locale-scoped CSS. I reviewed the diff, the font license, and the specificity reasoning myself. AI-verified items: the bundled font actually contains the reported glyphs (ێ ۆ ڵ), TypeScript passes, and Vite emits + references the hashed woff2. It could **not** verify actual glyph rendering or dark mode — that needs a human on macOS (see below).

## Proposed Changes

On macOS, the UI renders in the system font (San Francisco), which has no glyphs for several Central Kurdish letters such as ێ ۆ ڵ. As a result, Kurdish words and sentences break and become hard to read once the app language is set to کوردیی ناوەندی. Linux is unaffected because its default fonts cover these letters.

This bundles **Vazirmatn** — a free, SIL Open Font License font that fully covers the Central Kurdish alphabet (and the exact font the reporter confirmed fixes the problem) — and applies it **only** when the UI language is Central Kurdish. Every other language keeps today's system-font appearance unchanged.

For Kurdish users on macOS, text now renders correctly and legibly across the app, including form fields, tooltips and dropdowns (which otherwise keep their own hardcoded font).

> ⚠️ Visual change: needs human review in light + dark mode.

|**Before**|**After**|
|---|---|
|kurdish-before|kurdish-after|

## Testing Instructions

1. On **macOS**, start Studio.
2. Go to **Settings → Language** and switch to **کوردیی ناوەندی**.
3. Confirm Kurdish text throughout the app (settings, buttons, labels, inputs, tooltips, dropdowns) renders correctly — the letters ێ ۆ ڵ should display without breaking words.
4. Switch back to another language and confirm its appearance is unchanged.
5. Check both **light and dark** appearance (System Settings → Appearance).

## Pre-merge Checklist


apps/studio/src/index.css
```diff
@@ -48,8 +60,13 @@
 	}
 
 	.skeleton-bg {
-		animation: pulse 2s cubic-bezier(0.4, 0, 0.6, 1) infinite;
-		background-image: linear-gradient(to right, var(--color-frame-surface), var(--color-frame-surface-alt), var(--color-frame-surface));
+		animation: pulse 2s cubic-bezier( 0.4, 0, 0.6, 1 ) infinite;
```
katinthehatsite, line 63: This does not seem to be related to the PR? Looks like formatting? 😅

katinthehatsite (approved):
I think the solution makes sense, I looked into alternatives (e.g. loading the font from the network which adds network dependency) but this seems to be the best option for the issue that makes user's experience smoother 👍 

I left one minor comment but it can be merged with it.

It was a bit hard to review since I don't know for sure what I am looking for 😅 but it seems correct after the change:

Image

apps/studio/src/index.css
wojtekn: Yep, linter fixed it.

